PR #1136, opened 2024-08-26
Hide non-necessary symbols from shared object

ksivaman:
# Description

`libtransformer_engine` currently exposes all symbols globally, which can lead to symbol resolution conflicts when certain libraries are exported after TE's shared object file is loaded dynamically, for .e.g.,

```
pip install rdkit==2024.3.3
python -c "import transformer_engine; from rdkit import Chem"
```
 Leads to the following error:
```
terminate called after throwing an instance of 'std::bad_alloc'
  what():  std::bad_alloc
Aborted (core dumped)
```

## Type of change

- [x] Infra/Build change
- [ ] Code refractor

## Changes

- Limit exposing unnecessary symbols in `libtransformer_engine`.

# Checklist:

### DIFF
--- a/transformer_engine/common/libtransformer_engine.version
+++ b/transformer_engine/common/libtransformer_engine.version
@@ -1,0 +1,4 @@
+{
+	global: *nvte*; *transformer_engine*;
+	local: *;
+};
--- a/transformer_engine/common/CMakeLists.txt
+++ b/transformer_engine/common/CMakeLists.txt
@@ -15,6 +15,10 @@ if (CMAKE_BUILD_TYPE STREQUAL "Debug")
   set(CMAKE_CUDA_FLAGS_DEBUG "${CMAKE_CUDA_FLAGS_DEBUG} -G")
 endif()
 
+# Hide non-necessary symbols in shared object.
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/libtransformer_engine.version")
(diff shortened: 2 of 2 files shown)


timmoon10 (approved):

transformer_engine/common/libtransformer_engine.version
```diff
@@ -0,0 +1,4 @@
+{
+	global: *nvte*; *transformer_engine*;
```
ptrendx, line 2: why do we need `*` before `nvte` or `transformer_engine`? Shouldn't it be `nvte_*` instead (since that is the format of our C APIs)? I'm also not convinced we need to expose anything with `transformer_engine`, could you give examples of the APIs we would want to expose that way?

ksivaman: 1. The * before `nvte` and `transformer_engine` is needed because some symbols are mangled, for e.g. `_Z14nvte_unpermuteP...`.
2. `transformer_engine` is necessary because we use some functionality such as transformer_engine::getenv in the frameworks, so we need to define the symbol for it to load the framework shared objects.

Given how both `*transformer_engine*` and (to a lesser extent) `*nvte*` patterns are specific to TE, I think it's safe to continue exposing these globally.

ksivaman:
Pipeline 17865970

transformer_engine/common/libtransformer_engine.version
ptrendx: 1. Anything that is mangled is a bug since we advertise C API. In the case of permute it is because there is no `extern`"C"` decorator around those functions. Also we are using `nullptr` instead of `NULL` or `0` - apparently C23 introduced it but I don't think we want to expect people to have such new C standards. @phu0ngng could we have it fixed?
2. I thought we do not do this anymore and instead compile that function in each framework extension separately. @timmoon10 could you comment?

PR #1140, opened 2024-08-27
Norms Refractor

phu0ngng:
# Description

Layernorm and RMS refractor, preparing for TE/cuDNN Norms Integration.

## Type of change

- [ ] Infra/Build change
- [x] Code refractor

## Changes

# Checklist:


transformer_engine/common/layer_norm/norms.h
```diff
@@ -0,0 +1,334 @@
...
+
+template <NVTE_NORM_TYPE NormEnum, bool IF_TUNED, typename Enable = void>
+struct RegistryType {};
+
+template <NVTE_NORM_TYPE NormEnum, bool IF_TUNED>
+struct RegistryType<NormEnum, IF_TUNED, typename std::enable_if<IF_TE_NORMS<NormEnum>()>::type> {
+  using type = std::conditional_t<
+      IF_TUNED, std::conditional_t<IF_TE_FWD_NORMS<NormEnum>(), FwdTunedRegistry, BwdTunedRegistry>,
+      std::conditional_t<IF_TE_FWD_NORMS<NormEnum>(), FwdGeneralRegistry, BwdGeneralRegistry>>;
+};
```
timmoon10, line 181: While templating on `IF_TUNED` makes sense for our current kernel implementation, I don't think it's a good idea going forward. The reason TE compilation is so slow (and is unusable on some systems) is because we statically compile so many tuned LayerNorm/RMSNorm kernels. If we choose to keep the TE tuned kernels, we should port them to NVRTC so we can compile them at runtime, similar to how we handle the transpose kernels. This would completely remove the need for this static registry for tuned kernels.

transformer_engine/common/layer_norm/norms.h
```diff
@@ -0,0 +1,334 @@
...
+
+  virtual std::vector<size_t> get_workspace_shape() { return {0}; };
+
+  virtual std::vector<size_t> get_barrier_shape() { return {0}; };
+};
+
+template <NVTE_NORM_TYPE NormEnum>
+class NormFwdTe : public NormBase {
+ public:
+  NormFwdTe();
```
timmoon10, line 293: I see that calling this constructor is considered an error:
https://github.com/NVIDIA/TransformerEngine/blob/41273ebec7941ddf12c785cd6f7113cf8c9062c0/transformer_engine/common/layer_norm/norms.cpp#L147-L152
We can catch this at compile-time with:
```suggestion
  NormFwdTe() = delete;
```

transformer_engine/common/layer_norm/norms.cpp
```diff
@@ -0,0 +1,350 @@
...
+  uint64_t type_key = get_type_id(wtype) | (get_type_id(itype) << 2) | (get_type_id(otype) << 4) |
+                      (get_type_id(ctype) << 6);
+  uint64_t launcher_key = (type_key << 32) | hidden_size;
+  return launcher_key;
+}
+
+template <NVTE_NORM_TYPE NormEnum>
+FwdFunction& get_fwd_launcher(DType wtype, DType itype, DType otype, DType ctype,
+                              const FwdParams& params) {
+  if constexpr (!IF_TE_FWD_NORMS<NormEnum>()) NVTE_ERROR("Unexpected NVTE_NORM_TYPE!");
```
timmoon10, line 72: We can catch these errors at compile-time with a `static_assert`:
```suggestion
  static_assert(IF_TE_FWD_NORMS<NormEnum>(), "Unexpected NVTE_NORM_TYPE");
```
I see this `if constexpr (...) { NVTE_ERROR(...) }` pattern used in several other places.

PR #1149, opened 2024-08-29
[PyTorch/C] Exposed Userbuffers configuration option to control comm and compute stream priorities

denera:
# Description

Userbuffers configuration dictionary now has new `"comm_priority": -1` and `"gemm_priority: -1` options with default values of -1 for both.

Attn. @rachitgarg91 

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Updated `tex.UbufCommOverlap` and `tex.UbufP2PCommOverlap` constructors with optional arguments for comm and GEMM stream priorities.
- Updated `te.module.base.initialize_ub()` to read `"comm_priority"` and `"gemm_priority"` options in the config dictionary.

# Checklist:

### DIFF
--- a/transformer_engine/common/util/cuda_runtime.cpp
+++ b/transformer_engine/common/util/cuda_runtime.cpp
@@ -145,6 +145,31 @@ const std::string &include_directory(bool required) {
   return path;
 }
 
+std::pair<int, int> stream_priority_range(int device_id) {
+  int ori_dev;
+  if (device_id >= 0) {
+    ori_dev = current_device();
+    NVTE_CHECK_CUDA(cudaSetDevice(device_id));
+  }
+  int min_priority, max_priority;
+  NVTE_CHECK_CUDA(cudaDeviceGetStreamPriorityRange(&min_priority, &max_priority));
+  if (device_id >= 0)
+    NVTE_CHECK_CUDA(cudaSetDevice(ori_dev));
+  return std::make_pair(min_priority, max_priority);
+}
+
+bool supports_multicast(int device_id) {
+  int result;
+  CUdevice cudev;
+  if (device_id < 0)
+    NVTE_CHECK_CUDA(cudaGetDevice(&device_id));
+  NVTE_CALL_CHECK_CUDA_DRIVER(cuDeviceGet, &cudev, device_id);
+  NVTE_CALL_CHECK_CUDA_DRIVER(cuDeviceGetAttribute, &result,
+                              CU_DEVICE_ATTRIBUTE_MULTICAST_SUPPORTED, cudev);
+
+  return static_cast<bool>(result);
+}
+
 }  // namespace cuda
 
 }  // namespace transformer_engine
--- a/transformer_engine/common/util/cuda_runtime.cpp
+++ b/transformer_engine/common/util/cuda_runtime.cpp
@@ -102,8 +102,7 @@ void stream_priority_range(int *lowest_priority, int *highest_priority, int devi
 bool supports_multicast(int device_id) {
   static std::vector<bool> cache(num_devices());
   static std::vector<std::once_flag> flags(num_devices());
-  if (device_id < 0)
-    device_id = current_device();
+  if (device_id < 0) device_id = current_device();
 
   NVTE_CHECK(0 <= device_id && device_id < num_devices(), "invalid CUDA device ID");
   auto init = [&]() {
--- a/transformer_engine/common/util/cuda_runtime.cpp
+++ b/transformer_engine/common/util/cuda_runtime.cpp
@@ -153,16 +153,14 @@ std::pair<int, int> stream_priority_range(int device_id) {
   }
   int min_priority, max_priority;
   NVTE_CHECK_CUDA(cudaDeviceGetStreamPriorityRange(&min_priority, &max_priority));
-  if (device_id >= 0)
-    NVTE_CHECK_CUDA(cudaSetDevice(ori_dev));
+  if (device_id >= 0) NVTE_CHECK_CUDA(cudaSetDevice(ori_dev));
   return std::make_pair(min_priority, max_priority);
 }
 
 bool supports_multicast(int device_id) {
   int result;
   CUdevice cudev;
-  if (device_id < 0)
-    NVTE_CHECK_CUDA(cudaGetDevice(&device_id));
+  if (device_id < 0) NVTE_CHECK_CUDA(cudaGetDevice(&device_id));
   NVTE_CALL_CHECK_CUDA_DRIVER(cuDeviceGet, &cudev, device_id);
   NVTE_CALL_CHECK_CUDA_DRIVER(cuDeviceGetAttribute, &result,
                               CU_DEVICE_ATTRIBUTE_MULTICAST_SUPPORTED, cudev);
--- a/transformer_engine/common/util/cuda_runtime.cpp
+++ b/transformer_engine/common/util/cuda_runtime.cpp
@@ -80,6 +80,51 @@ int sm_count(int device_id) {
   return cache[device_id];
 }
 
+void stream_priority_range(int *lowest_priority, int *highest_priority, int device_id) {
+  static std::vector<std::vector<int>> cache(num_devices());
+  static std::vector<std::once_flag> flags(num_devices());
+
+  auto init = [&]() {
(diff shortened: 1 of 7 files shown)


transformer_engine/pytorch/csrc/comm_gemm_overlap.h
```diff
@@ -233,8 +236,6 @@ struct UbufCommOverlap : torch::CustomClassHolder, UbufBase {
     cudaEventDestroy(_stop_compute);
     cudaEventDestroy(_start_compute);
 
-    for (size_t i = 0; i < _stream_compute.size(); i++) cudaStreamDestroy(_stream_compute[i]);
```
timmoon10, line 236: Are we intentionally leaking these `cudaStream_t`s? I see that PyTorch leaks intentionally because it can't guarantee that the streams are destroyed before the CUDA environment is destroyed: https://github.com/pytorch/pytorch/blob/cf11fc0dcbb9c907cf6e851109b92f4157e445c9/c10/cuda/CUDAStream.cpp#L27-L39

transformer_engine/pytorch/csrc/comm_gemm_overlap.h
```diff
@@ -245,6 +246,22 @@ struct UbufCommOverlap : torch::CustomClassHolder, UbufBase {
     }
   }
 
+  void _set_stream_priority(at::cuda::CUDAStream stream, int priority) {
+    cudaLaunchAttributeValue attributes = {.priority = priority};
```
timmoon10, line 250: Should this be:
```suggestion
    cudaLaunchAttributeValue attributes;
    attributes.priority = priority;
```

transformer_engine/pytorch/csrc/comm_gemm_overlap.h
```diff
@@ -791,6 +814,25 @@ struct UbufP2PCommOverlap : torch::CustomClassHolder, UbufBase {
...
```
timmoon10, line 818:
```suggestion
    cudaLaunchAttributeValue attributes;
    attributes.priority = priority;
```

timmoon10 (approved):
LGTM

transformer_engine/common/util/cuda_runtime.cpp
```diff
@@ -145,6 +145,29 @@ const std::string &include_directory(bool required) {
...
+  return std::make_pair(min_priority, max_priority);
+}
+
+bool supports_multicast(int device_id) {
+  int result;
+  CUdevice cudev;
+  if (device_id < 0) NVTE_CHECK_CUDA(cudaGetDevice(&device_id));
+  NVTE_CALL_CHECK_CUDA_DRIVER(cuDeviceGet, &cudev, device_id);
+  NVTE_CALL_CHECK_CUDA_DRIVER(cuDeviceGetAttribute, &result,
+                              CU_DEVICE_ATTRIBUTE_MULTICAST_SUPPORTED, cudev);
```
ptrendx, line 166: This seems to be introduced in CUDA 12.1 and we support CUDA 12.0 - should be guarded.

ptrendx: Also most probably for both of these functions we should not be calling those functions every single time, but rather cache the results.

transformer_engine/pytorch/csrc/comm_gemm_overlap.h
```diff
@@ -192,12 +187,17 @@ struct UbufCommOverlap : torch::CustomClassHolder, UbufBase {
       printf("!!! [UB] Register UBuf %d\n", _ub_reg);
     }
 
+    // if priorities are zero, use max priority for comm and min priority for GEMM
+    if (gemm_priority == 0 && comm_priority == 0)
+      std::tie(gemm_priority, comm_priority) = transformer_engine::cuda::stream_priority_range();
```
ptrendx, line 192: Huh? How can we use those functions in the framework plugin? We should not be exposing C++ symbols from transformer_engine.so. Most probably (for now, until the refactor) those functions should be part of this extension.

transformer_engine/pytorch/csrc/comm_gemm_overlap.h
```diff
@@ -192,12 +187,17 @@ struct UbufCommOverlap : torch::CustomClassHolder, UbufBase {
...
     at::cuda::CUDAStream stream_main = at::cuda::getCurrentCUDAStream();
     for (int i = 0; i < std::min(num_max_streams, num_splits); i++) {
-      cudaStream_t stream;
-      cudaStreamCreateWithPriority(&stream, cudaStreamNonBlocking, -1);
+      cudaStream_t stream_tmp;
+      NVTE_CHECK_CUDA(
+          cudaStreamCreateWithPriority(&stream_tmp, cudaStreamNonBlocking, gemm_priority));
       _stream_compute.push_back(
-          at::cuda::getStreamFromExternal(stream, stream_main.device_index()));
+          at::cuda::getStreamFromExternal(stream_tmp, stream_main.device_index()));
```
ptrendx, line 200: Why do we need to tell pyTorch about our temporary stream?

transformer_engine/pytorch/csrc/comm_gemm_overlap.h
```diff
@@ -207,7 +207,7 @@ struct UbufCommOverlap : torch::CustomClassHolder, UbufBase {
 
     // Set the number of SMs for GEMM with margin
     cudaDeviceProp prop;
-    cudaGetDeviceProperties(&prop, 0);
+    NVTE_CHECK_CUDA(cudaGetDeviceProperties(&prop, 0));
```
ptrendx, line 210: This call is expensive. Is this constructor running only once or during every communication? In the layernorm we take that info from pyTorch because it caches that information. We should cache it as well.

transformer_engine/pytorch/csrc/comm_gemm_overlap.h
```diff
@@ -302,12 +311,15 @@ struct UbufCommOverlap : torch::CustomClassHolder, UbufBase {
     if (B_scale_inverse.numel()) B_scale_inverse = B_scale_inverse[B_fp8_tensor];
 
     assert(pre_gelu_out.numel() == 0);
+    at::cuda::setCurrentCUDAStream(_stream_compute[0]);
```
ptrendx, line 314: Why do we need that?

ptrendx: Ok, nvm - I see that the te_gemm uses current Torch stream. We could probably change that.

transformer_engine/pytorch/csrc/comm_gemm_overlap.h
denera: The constructor is executed only once per overlap type (e.g. "qkv_fprop", "fc1_dgrad", etc...) when we call `te.module.base.initialize_ub()`.

It's still not a bad idea for our own `cuda_driver.h` in TE/common to provide a cached version of this call. I'll add it to the PR.

ksivaman: @denera `at::cuda::getCurrentDeviceProperties()` could be used, framework already caches these.

transformer_engine/pytorch/csrc/comm_gemm_overlap.h
denera: These aren't temporary. We maintain them throughout the lifetime of the Userbuffer object. That said, I don't know why the original code was written to wrap them into `at::cuda::CUDAStream` objects before storing into the vector of compute streams. I suspect that it might be necessary to expose the streams to the PyTorch profiler, but I don't think it makes any performance or functionality difference otherwise whether we do this or not.

transformer_engine/pytorch/csrc/comm_gemm_overlap.h
denera: Thanks for the heads up! I'm replacing this with a pure C signature.

transformer_engine/pytorch/csrc/comm_gemm_overlap.h
denera: I just replaced this chunk of code with `transformer_engine::cuda::sm_count()`, which is already a cached implementation, instead of manually getting it from device props.

PR #921, opened 2024-06-13
[PyTorch] reverting autocast API back to PyTorch v2.3.1 and below

denera:
# Description

PR #916 prematurely updated PyTorch autocast API to v 2.4.0-RC. This PR reverts it back to the API in v2.3.1 and older.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Changes

- Revert autocast API changes in PR #916.

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1167, opened 2024-09-06
Fix autocast deprecation warning.

jondeaton:
Fixes warning about deprecation of `torch.get_autocast_gpu_dtype`

### DIFF
--- a/transformer_engine/pytorch/module/base.py
+++ b/transformer_engine/pytorch/module/base.py
@@ -575,7 +575,7 @@ class TransformerEngineBaseModule(torch.nn.Module, ABC):
         """Get activation data type for AMP."""
         # Native AMP (`torch.autocast`) gets highest priority
         if torch.is_autocast_enabled():
-            self.activation_dtype = torch.get_autocast_gpu_dtype()
+            self.activation_dtype = torch.get_autocast_dtype('cuda')
             return
 
         # All checks after this have already been performed once, thus skip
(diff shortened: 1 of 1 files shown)


timmoon10 (approved):
Thanks for the fix, this looks good to me. Can you [sign your commit](https://github.com/NVIDIA/TransformerEngine/blob/bdea56fc023014eaf52a171047b641b3bfdded70/CONTRIBUTING.rst#sign-your-work) to get past the DCO check?

jondeaton:
> Can you [sign your commit](https://github.com/NVIDIA/TransformerEngine/blob/bdea56fc023014eaf52a171047b641b3bfdded70/CONTRIBUTING.rst#sign-your-work) to get past the DCO check?

Yep sure thing, I think I've signed off properly now.

ksivaman:
Looks like the first commit is still not signed off, could you try to sign-off all commits? Thanks! @jondeaton

jondeaton:
Yeah, sorry about that - new to the process. I think I've signed off on all of them now?

ksivaman (changes requested):
On second thought, we had this change previously in a PR and ended up reverting it [here](https://github.com/NVIDIA/TransformerEngine/pull/921). Since this API is a relatively new addition to pytorch, we would want it to be accompanied by a pytorch version check so that versions 2.3.1 or older aren't broken!

PR #1174, opened 2024-09-10
[PyTorch] Miscellaneous fixes for FA3 attention

cyanguwa:
# Description

This PR makes a few changes to the FA3 attention path.
- Adds `descale_q`, `descale_k` and `descale_v` to FA3 FP8 call. This allows for custom descaling factors instead of the default 1s for `q`, `k` and `v`. This requires FA3 PR [1210](https://github.com/Dao-AILab/flash-attention/pull/1210) to be in your FA3 installation.
- Restricts FA3 path to `flash_attn_func` for FP8 since `flash_attn_varlen_func` does not support FP8 yet.
- Fixes the transposes in the `qkv_format=sbhd` case when `fp8_mha=true`.
- Enables sliding window support for FP16/BF16 from FA3 (no FP8 support yet).
- Improves the messaging when FA3 is not installed, and when it's missing some kwargs. We provide some installation instructions, and remind users to update their installation, in those two cases. This PR targets FA 3.0.0b1, and it's rapidly developing, with more kwargs being added to the API.
- Casts FA3 output to the same type as cuDNN attention in unit tests.
- Enables INFO level messaging in CI test to help check correct backends are used for different tests.

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- See description above.

# Checklist:


transformer_engine/pytorch/attention.py
```diff
@@ -5053,6 +5073,7 @@ def forward(
                         softmax_scale=self.softmax_scale,
                         causal="causal" in attn_mask_type,
                         deterministic=self.deterministic,
+                        **fa_optional_forward_kwargs_fp8,
```
ptrendx, line 5076: We should guard this behind a version check - is there already a published version of FA that contains your PR?

cyanguwa: There isn't. The latest release is 2.6.3 and my PR is just a commit after that on main. I can guardrail it with `>2.6.3`, but if people have used my method (`pip install "git+https://github.com/Dao-AILab/flash-attention.git#egg=flashattn-hopper&subdirectory=hopper"`) to install it, it doesn't matter what the base FA2 version is. That method just installs from the latest FA/hopper directory.

cyanguwa:
FA3: pipeline 18489052

transformer_engine/pytorch/attention.py
cyanguwa: Added try/except to deal with the default/custom descales in [0553a83](https://github.com/NVIDIA/TransformerEngine/pull/1174/commits/0553a8317f538129ee31086824336dd7f5924b16).

cyanguwa:
FA3 pipeline 18528229

cyanguwa:
FA3 pipeline 18528479

cyanguwa:
FA3 pipeline 18528978

transformer_engine/pytorch/attention.py
```diff
@@ -5088,8 +5132,12 @@ def forward(
         if qkv_format == "sbhd":
             # (bs)hd -> bs(hd) -> sb(hd)
             if fp8 and fp8_meta["recipe"].fp8_mha:
-                output.reshape(batch_size * max_seqlen_q // cp_size, -1).transpose_2d()
-                output = output.reshape(batch_size, max_seqlen_q // cp_size, -1)
+                output = Float8Tensor.make_like(
+                    output,
+                    data=output._data.reshape(batch_size, max_seqlen_q // cp_size, -1)
+                    .transpose(0, 1)
+                    .contiguous(),
```
ptrendx, line 5139: Could you explain the problem with the previous logic?

cyanguwa: The `.transpose_2d()` was wrong - we were trying to transpose the `batch_size` and `max_seqlen_q//cp_size` dimensions, not `batch_size * max_seqlen_q // cp_size` with `-1`.

Also, the `.make_like()` was necessary because otherwise the `Float8Tensor` would still have the old shape while its `._data` has been transposed.

xrennvidia:
LGTM. Only one small question or comment: seems like FA3 only can support FP8 with BSHD/SBHD format, THD format is not supported with FP8. Should we added an assert message for this in TE? Anyway, this will finally trigger error in Tri Dao's code, but I think better to tell users this in TE also.

xrennvidia (approved):

cyanguwa:
FA3: 19002912

cyanguwa:
@xrennvidia do you mind taking another look at the PR? I made a couple more changes after your last review. Thanks!

transformer_engine/pytorch/attention.py
```diff
@@ -436,6 +436,9 @@ def get_attention_backend(
                 "padding between sequences, i.e. [a, a, PAD, b, b, b, PAD, c, PAD]"
             )
             use_flash_attention = False
+        if use_flash_attention and _use_flash_attn_3 and fp8 and fp8_meta["recipe"].fp8_dpa:
+            logger.debug("Disabling FlashAttention 3 for FP8 and qkv_format = thd")
+            _use_flash_attn_3 = False
```
xrennvidia, line 441: qkv_format == sbhd/bshd and "padding" in attn_mask_type also cannot run FP8, right?
and at here, I think you can directly set `use_flash_attention = False`, because FA2 cannot run FP8 anyway.

transformer_engine/pytorch/attention.py
```diff
@@ -5013,7 +5015,7 @@ def forward(
                 if _flash_attn_2_4_1_plus:
                     fa_optional_forward_kwargs["deterministic"] = self.deterministic
                 fa_optional_forward_args_thd = []
-                if qkv_format in ["bshd", "sbhd"] and "padding" not in attn_mask_type:
+                if (qkv_format in ["bshd", "sbhd"] and "padding" not in attn_mask_type) or fp8:
```
xrennvidia, line 5018: if fp8 is true, I think `qkv_format in ["bshd", "sbhd"] and "padding" not in attn_mask_type` is supposed to be true for sure? so adding `or fp8` seems redundant?

transformer_engine/pytorch/attention.py
xrennvidia: Maybe just add `if use_flash_attention and _use_flash_attn_3 and (qkv_format == "thd" or "padding" in attn_mask_type): use_flash_attention = False` after [this line](https://github.com/cyanguwa/TransformerEngine/blob/add_descales/transformer_engine/pytorch/attention.py#L395).

cyanguwa: Thanks for helping me figure out the logic here :) It's a bit of a brain teaser I guess.

I almost went with what you said, but on second thought, I feel my current logic is right. Hear me out: 
- For F16 and the `qkv_format == sbhd/bshd and "padding" in attn_mask_type` case, we allow the `FlashAttention` backend (i.e. `use_flash_attention=True`), but not the `flash_attn_func` call from FA2. Because `FlashAttention` can pack up the useful tokens through `pack_tensors()` and then call `flash_attn_varlen_func`.
- For FP8, we can't even do that because `flash_attn_varlen_func` doesn't support FP8 yet.

Let me know if this makes sense. Thanks.

xrennvidia: Sorry, I do not totally understand this.

You are checking under `if qkv_format == "thd"`,  and you have `and fp8 and fp8_meta["recipe"].fp8_dpa`, which means only FP8 with THD format running can come to the code at here, right? neither FA2 nor FA3 can support FP8 with THD format, right? so I think we can directly set `use_flash_attention=False` at here (not just disable FA3)?

`qkv_format == sbhd/bshd and "padding" in attn_mask_type` will skip the code at here because `qkv_format != thd`. I think we mainly want to check if the FP8 running case can be supported by FlashAttention, so I [made this recommendation](https://github.com/NVIDIA/TransformerEngine/pull/1174/files#r1786920097). It's a check added under `if fp8 and fp8_meta["recipe"].fp8_dpa`, so should not impact BF16 at all, right?

cyanguwa: We disable FP8 for FA2 [here](https://github.com/cyanguwa/TransformerEngine/blob/b765f3d5c042b7adc79fdff4a29a66fbd20c68c1/transformer_engine/pytorch/attention.py#L388) under `Filter: Execution type`, and it'll never support FP8.

We disable FP8 THD for FA3 [here](https://github.com/cyanguwa/TransformerEngine/blob/b765f3d5c042b7adc79fdff4a29a66fbd20c68c1/transformer_engine/pytorch/attention.py#L439) under `Filter: QKV layout`, but FA3 might add the support soon, given how fast things are moving.

I'd like to keep things separate under different `Filter`s and also give FA3 the room to change (without us changing too much).

With `Filter: attn_mask_type`, we do support `qkv_format == sbhd/bshd and "padding" in attn_mask_type` with the `FlashAttention` path, hence `use_flash_attention` can be True. But within it, we have to distinguish the functionality of `flash_attn_func` and `flash_attn_varlen_func`, and their difference in FA2 and FA3 as well.

cyanguwa: I expect to remove [this](https://github.com/cyanguwa/TransformerEngine/blob/b765f3d5c042b7adc79fdff4a29a66fbd20c68c1/transformer_engine/pytorch/attention.py#L439) section, and the `or fp8` [here](https://github.com/cyanguwa/TransformerEngine/blob/b765f3d5c042b7adc79fdff4a29a66fbd20c68c1/transformer_engine/pytorch/attention.py#L5014), when `flash_attn_varlen_func` in FA3 supports FP8. Nothing else should need to be changed (none of the mask related code).

cyanguwa: Resolved offline.

cyanguwa:
FA3: 19123312

PR #359, opened 2023-08-04
Optimize flash-attention transposes

ksivaman:

PR #1176, opened 2024-09-11
[PyTorch] Relax the contiguous check for flash attention

yaox12:
# Description

The continuity requirements and behavior of FlashAttention are:
- q/k/v are contiguous **at the last dimension**. [[ref 1](https://github.com/Dao-AILab/flash-attention/blob/bdf733be55f0b323a8cf7cc6745a81c3f43cd7f0/flash_attn/flash_attn_interface.py#L15)][[ref 2](https://github.com/Dao-AILab/flash-attention/blob/bdf733be55f0b323a8cf7cc6745a81c3f43cd7f0/csrc/flash_attn/flash_api.cpp#L379-L381)]
- out has the same layout (stride) as q. [[ref](https://github.com/Dao-AILab/flash-attention/blob/bdf733be55f0b323a8cf7cc6745a81c3f43cd7f0/csrc/flash_attn/flash_api.cpp#L441)]

So we can just remove the `.contiguous()` calls on q/k/v and rely on FlashAttention doing the minimum continuity check. For `sbhd` formats, all the four `.contiguous()` calls on q/k/v/out are removed.

For context parallel, we still force q/k/v to be contiguous.

Closes https://github.com/NVIDIA/TransformerEngine/pull/359.

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/attention.py
+++ b/transformer_engine/pytorch/attention.py
@@ -4881,11 +4881,19 @@ class FlashAttention(torch.nn.Module):
                     query_layer, key_layer, value_layer = [
                         x.transpose(0, 1) for x in (query_layer, key_layer, value_layer)
                     ]
+            if context_parallel:
+                query_layer, key_layer, value_layer = [
+                    x.contiguous() for x in (query_layer, key_layer, value_layer)
+                ]
         else:
             if qkv_format == "sbhd":
                 query_layer._data, key_layer._data, value_layer._data = [
                     x.transpose(0, 1)
                     for x in (query_layer._data, key_layer._data, value_layer._data)
+                ]
+            if context_parallel:
+                query_layer._data, key_layer._data, value_layer._data = [
+                    x.contiguous() for x in (query_layer._data, key_layer._data, value_layer._data)
                 ]
 
         batch_size = query_layer.shape[0]
--- a/transformer_engine/pytorch/attention.py
+++ b/transformer_engine/pytorch/attention.py
@@ -4879,22 +4879,13 @@ class FlashAttention(torch.nn.Module):
                     )
                 else:
                     query_layer, key_layer, value_layer = [
-                        x.transpose(0, 1).contiguous()
-                        for x in (query_layer, key_layer, value_layer)
+                        x.transpose(0, 1) for x in (query_layer, key_layer, value_layer)
                     ]
-            elif qkv_format in ["bshd", "thd"]:
-                query_layer, key_layer, value_layer = [
-                    x.contiguous() for x in (query_layer, key_layer, value_layer)
-                ]
         else:
             if qkv_format == "sbhd":
                 query_layer._data, key_layer._data, value_layer._data = [
-                    x.transpose(0, 1).contiguous()
+                    x.transpose(0, 1)
                     for x in (query_layer._data, key_layer._data, value_layer._data)
-                ]
-            elif qkv_format in ["bshd", "thd"]:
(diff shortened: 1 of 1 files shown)


yaox12:
@cyanguwa Can you review?

transformer_engine/pytorch/attention.py
```diff
@@ -5090,19 +5089,15 @@ def forward(
...
+                output = output.view(batch_size, max_seqlen_q // cp_size, -1).transpose(0, 1)
         elif qkv_format == "bshd":
             # (bs)hd -> bs(hd)
             output = output.reshape(batch_size, max_seqlen_q // cp_size, -1)
         elif qkv_format == "thd":
             # thd -> t(hd)
             output = output.reshape(output.shape[0], -1)
 
-        return output
+        return output.contiguous()
```
cyanguwa, line 5100: Wouldn't this force the other paths to call `.contiguous()` as well? Should we remove the last two changes?

transformer_engine/pytorch/attention.py
```diff
@@ -5090,19 +5089,15 @@ def forward(
                 output.reshape(batch_size * max_seqlen_q // cp_size, -1).transpose_2d()
                 output = output.reshape(batch_size, max_seqlen_q // cp_size, -1)
```
cyanguwa, line 5090: I think the reshaping/transpose here is a bit fishy - I'll have to take another look at it (It's from my FA3 PR).

cyanguwa, line 5090: I think the logic for FP8 MHA is probably wrong here. Could you change it to something like this:
```
output = Float8Tensor.make_like(output, data=output._data.transpose(0,1).contiguous())
```

cyanguwa: It's ok, I'll address this in a different PR.

transformer_engine/pytorch/attention.py
yaox12: I removed `.contiguous()` calls on q/k/v for all paths except for CP, so we need a `.contiguous()` here to ensure the continuity of output.

I list the number of `.contiguous()` calls under all situations before and after this change as below:
1. If input q/k/v is contiguous
  a. Before: 4 (q/k/v/out) for `sbhd`, 0 for `bshd` and `thd`.
  b. After: 0 for all paths.
2. If input q/k/v is contiguous on the last dimension but not others
  a. Before: 4 (q/k/v/out) for `sbhd` and 3 (q/k/v) for `bshd` and `thd`.
  b. After: 1 (out) for all paths.
3. If input q/k/v is not contiguous on the last dimension
  a. Before: 4 (q/k/v/out) for `sbhd` and 3 (q/k/v) for `bshd` and `thd`.
  b. After: 4 (q/k/v in FA, out in our code) for `sbhd`, and 3 (q/k/v in FA) for `bshd` and `thd`.

Though not all the cases would happen, moving the `.contiguous()` calls from q/k/v to out does benefit the above case 2).

cyanguwa: I guess I wasn't sure if every `.contiguous()` call translates to work. Looks like it doesn't: see [here](https://github.com/pytorch/pytorch/blob/e037bb326fdafca243bdb08023bbef93b29a4513/aten/src/ATen/native/TensorProperties.cpp#L109). Then I agree with your assessment.

cyanguwa (approved):

PR #1178, opened 2024-09-11
Allow to pass architectures like 90a, without being overriden

aurianer:
In case the architecture 90a is passed, it's overridden to 90. This also avoid hardcoding the architectures to avoid mistakes.


build_tools/pytorch.py
```diff
@@ -80,10 +80,10 @@ def setup_pytorch_extension(
...
         )
 
-        if "80" in cuda_architectures:
-            nvcc_flags.extend(["-gencode", "arch=compute_80,code=sm_80"])
-        if "90" in cuda_architectures:
-            nvcc_flags.extend(["-gencode", "arch=compute_90,code=sm_90"])
+        for arch in cuda_architectures:
+            if arch_nb != "70":
+                arch_nb = arch.replace("sm_", "")
+                nvcc_flags.extend(["-gencode", f"arch=compute_{arch_nb},code={arch}"])
```
timmoon10, line 86: `cuda_architectures` is a semicolon-separated string:
https://github.com/NVIDIA/TransformerEngine/blob/0ee5ccda2eacf96fbc9c7ef1f7c084a71e0df7a6/build_tools/utils.py#L192-L193
We should change the logic:
```suggestion
        for arch in cuda_architectures.split(";"):
            if arch == "70":
                continue  # Already handled
            nvcc_flags.extend(["-gencode", f"arch=compute_{arch},code=sm_{arch}"])
```

timmoon10 (approved):

PR #1033, opened 2024-07-22
[PyTorch] Normalization ops

timmoon10:
# Description

This PR extends the operation-based API (see https://github.com/NVIDIA/TransformerEngine/pull/707) with LayerNorm, RMSNorm, and FP8 cast operations.

Compare with the existing module-based API:
```python
# Module-based API
module1 = te.LayerNormLinear(...)

# Operation-based API
module2 = te.ops.Sequential(
    te.ops.LayerNorm(...),
    te.ops.Linear(...),
)
```

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refractor

## Changes

Please list the changes introduced in this PR:

- LayerNorm operation
- FP8 cast operation
- RMSNorm operation

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1191, opened 2024-09-18
[PyTorch] Minor optimizations to reduce CPU overheads in modules

timmoon10:
# Description

We have observed that TE modules experience non-trivial CPU overhead, which often becomes a performance bottleneck in the forward pass of small models. For example, measuring the CPU runtime for Megatron-core modules with BF16 compute and TP=1:
- [`ColumnParallelLinear`](https://github.com/NVIDIA/Megatron-LM/blob/6b35ca80e8baca6f357b97304979e6b1c9a31899/megatron/core/tensor_parallel/layers.py#L649): 74 us per forward pass
- [`TEColumnParallelLinear`](https://github.com/NVIDIA/Megatron-LM/blob/6b35ca80e8baca6f357b97304979e6b1c9a31899/megatron/core/extensions/transformer_engine.py#L354): 140 us per forward pass

Unfortunately this overhead is distributed throughout the forward pass. Many basic PyTorch operations, e.g. getting attributes from `torch.Tensor`, involve O(1 us) overhead, so even basic checks to handle all of our advanced features will eventually add up to something non-trivial.

This PR makes a few minor optimizations:
- Avoid importing from `te.pytorch.cpu_offload` in every forward pass
- Memoize NCCL process group properties
- Avoid custom logic in `torch.nn.Module.__setattr__` when possible
- Avoid custom logic for accessing params in `torch.nn.Module` when possible
- Avoid accessing tensor attrs more than necessary

I see a 1.22x speedup, with 115 us per forward pass.

## Type of change

- [ ] Infra/Build change
- [x] Code refractor

## Changes

- Avoid importing from `te.pytorch.cpu_offload` in every forward pass
- Memoize NCCL process group properties
- Avoid custom logic in `torch.nn.Module.__setattr__` when possible
- Avoid custom logic for accessing params in `torch.nn.Module` when possible
- Avoid accessing tensor attrs more than necessary

# Checklist:


yaox12 (commented):
Can you propagate the CPU offloading importing fix to GroupedLinear as well? https://github.com/NVIDIA/TransformerEngine/blob/c0caadbe1cdba42efb09aba205f4de0905739f26/transformer_engine/pytorch/module/grouped_linear.py#L783

transformer_engine/pytorch/module/base.py
```diff
@@ -632,9 +649,9 @@ def _get_fp8_params(self) -> Union[List[torch.Tensor], None]:
     # assume FP8 execution.
     def init_fp8_metadata(self, num_gemms: int = 1) -> None:
         """Initialize fp8 related metadata and tensors during fprop."""
-        self.fp8_parameters = FP8GlobalStateManager.with_fp8_parameters()
-        self.fp8 = FP8GlobalStateManager.is_fp8_enabled()
-        self.fp8_calibration = FP8GlobalStateManager.is_fp8_calibration()
+        self._fast_setattr("fp8_parameters", FP8GlobalStateManager.with_fp8_parameters())
```
ptrendx, line 652: I wonder if we couldn't instead just do something like
```python
te_params = self.get_te_params()  # calls _fast_getattr internally, te_params is a normal object
te_params.fp8_parameters = FP8GlobalStateManager.with_fp8_parameters()
```

ptrendx: Otherwise it will be hard to enforce everybody using only the `_fast_get/setattr` I think.

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -369,7 +371,7 @@ def forward(
             out, _ = allreduce(out, tp_group)
 
         # [*, in_features] -> [*, out_features] except first dimension changes for SP
-        out = out.view(-1, *inp.shape[1:-1], out.shape[-1])
+        out = out.view(-1, *inp_shape[1:-1], out_features)
```
ptrendx, line 374: Why can't we just create out with the right shape in the gemm call instead?

transformer_engine/pytorch/module/base.py
timmoon10: Even better would be storing these attrs in `fp8_meta` or some other dict.

timmoon10: I tried wrapping these attrs in a property that internally calls `_fast_setattr`, but that added ~10 us overhead (probably from the extra indirection when getting the attrs). I think it's a good idea to refactor these frequently changed attrs so they are not held directly by the module, but I think that would be beyond the scope of this PR.

transformer_engine/pytorch/module/layernorm_linear.py
timmoon10: I think it would be a good idea if the GEMM extension implicitly flattened all input tensors to 2D. I'd prefer only including Python changes in this PR and leaving that change to a follow-up PR.

transformer_engine/pytorch/module/base.py
```diff
@@ -406,6 +406,23 @@ def __init__(self) -> None:
         self.fsdp_wrapped = False
         self.fsdp_group = None
         self._fp8_workspaces: Dict[str, Float8Tensor] = {}
+        self.activation_dtype: Optional[torch.dtype] = None
+
+        # Fast variant of __setattr__
+        # Note: torch.nn.Module has custom __setattr__ that handles
+        # modules, parameters, and buffers. This is unnecessary
+        # overhead when setting plain attrs.
+        self._fast_setattr: Callable[Tuple[str, Any], None]
```
ksivaman, line 415: What is the purpose of this?

transformer_engine/pytorch/module/layernorm.py
```diff
@@ -186,8 +187,21 @@ def reset_parameters(self, defer_init=False) -> None:
...
+            self.activation_dtype = torch.get_autocast_gpu_dtype()
+        elif self.activation_dtype != inp.dtype:
+            dtype = inp.dtype
+            for name, param in self.named_parameters():
+                if param is not None:
+                    assert dtype == param.dtype, (
+                        "Data types for parameters must match when outside of autocasted region. "
+                        f" Found input dtype: {dtype} and {name!r} dtype: {param.dtype}"
+                    )
+            self.activation_dtype = dtype
```
ptrendx, line 204: Shouldn't this use the fast version?

timmoon10: This is going away soon with https://github.com/NVIDIA/TransformerEngine/pull/1033, so I figure it's better not to put much effort into it.

transformer_engine/pytorch/module/base.py
timmoon10: I like to have a type hint to make usage clearer, but it's too long to fit in one line.

transformer_engine/pytorch/module/base.py
timmoon10: Moving this logic into the `__setattr__` function makes things a little cleaner. It adds ~2 us overhead, but it's still a win of ~6 us compared to the baseline.

transformer_engine/pytorch/utils.py
```diff
@@ -218,8 +218,12 @@ def safely_set_viewless_tensor_data(tensor: torch.Tensor, new_data_tensor: torch
 
 def cast_if_needed(tensor: torch.Tensor, dtype: torch.dtype) -> torch.Tensor:
     """Cast tensor to dtype"""
+    if tensor is None:
+        return None
+    if tensor.dtype == dtype:
+        return tensor
```
ptrendx, line 224:
```suggestion
    if tensor is None or tensor.dtype == dtype
        return tensor
```

ptrendx (approved):

PR #1196, opened 2024-09-23
Tests for distributed

pggPL:
# Description

I am working on debug API. Before it can be merged, it needs to be tested. We need to ensure that all the new layers will work properly in the distributed case also. The tests present in the repo focus on testing gemm/comm overlapping - the tests I want to add focus more on checking numerical corectness of multiple configurations of TE layers. Moreover, current 
## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- added numerical tests for distributed training,

# Checklist:


timmoon10:
As mentioned by @ptrendx, we'll need to include these tests in one of the QA scripts (see [`qa`](https://github.com/NVIDIA/TransformerEngine/tree/main/qa)) so that they are included in the CI pipelines. [`L1_pytorch_distributed_unittest`](https://github.com/NVIDIA/TransformerEngine/tree/main/qa/L1_pytorch_distributed_unittest) is simplest, but we could also make a new L0 distributed test. If we go that route, it might be worth including the [`te.Sequential` distributed tests](https://github.com/NVIDIA/TransformerEngine/blob/main/tests/pytorch/test_fusible_ops_distributed.py) as well.

tests/pytorch/distributed/run_numerics.py
```diff
@@ -0,0 +1,701 @@
...
+
+def main(argv=None, namespace=None):
+    global WORLD_RANK, WORLD_SIZE, NCCL_WORLD, FP8
+
+    dist.init_process_group(backend="nccl")
+    NCCL_WORLD = dist.new_group(backend="nccl")
+
+    WORLD_RANK = dist.get_rank()
+    WORLD_SIZE = dist.get_world_size()
+    assert WORLD_SIZE == 2, "This test uses 2 GPUs. Run with torchrun --nproc_per_node=2."
```
timmoon10, line 41: `test_numerics.py` can launch with a different number of GPUs:
https://github.com/NVIDIA/TransformerEngine/blob/fd67d8f3c3d32ba0f149ed0b0286bfd17e9db678/tests/pytorch/distributed/test_numerics.py#L32
```suggestion
```

tests/pytorch/distributed/run_numerics.py
```diff
@@ -0,0 +1,701 @@
...
+def _gather(tensor, dim=0):
+    """
+    Gathers tensors and concats them. Since torch.distributed.nn.functional.all_gather
+    multiplies gradients by WORLD_SIZE, those gradiedts are rescaled.
+    """
+
+    class HalfGradient(torch.autograd.Function):
+        @staticmethod
+        def forward(ctx, input):
+            ctx.save_for_backward(input)
```
timmoon10, line 94: Is this needed?
```suggestion
```

tests/pytorch/distributed/run_numerics.py
```diff
@@ -0,0 +1,701 @@
...
+
+
+def _check_outputs(output_single_node, output_distributed):
+    numerics_failed = torch.tensor([0], dtype=torch.uint8, device="cuda")
+    rtol = 0.125 if FP8 else 0.025
+    atol = 0.0625 if FP8 else 0.00125
+    output_failed, output_info = _compare_tensors(
+        "outputs", output_distributed, output_single_node, rtol, atol
+    )
+    dist_print(output_info, src=WORLD_RANK, error=output_failed)
```
timmoon10, line 124: Printing successful checks might be too verbose:
```suggestion
    if output_failed:
        dist_print(output_info, src=WORLD_RANK, error=True)
```

tests/pytorch/distributed/run_numerics.py
```diff
@@ -0,0 +1,701 @@
...
+    stream = sys.stderr if error else sys.stdout
+    if WORLD_RANK == (0 if src is None else src):
+        stream.write(f"[rank{WORLD_RANK}] {msg}{end}\n")
+    dist.barrier()
+
+
+def _check_outputs(output_single_node, output_distributed):
+    numerics_failed = torch.tensor([0], dtype=torch.uint8, device="cuda")
+    rtol = 0.125 if FP8 else 0.025
+    atol = 0.0625 if FP8 else 0.00125
```
timmoon10, line 120: Ideally we would use the tightest tolerances possible for the dtype, like in [`torch.testing.assert_close`](https://pytorch.org/docs/stable/testing.html#torch.testing.assert_close). I see that the tensor dimensions are small (~64), so we should be able to get away with this.

tests/pytorch/distributed/run_numerics.py
```diff
@@ -0,0 +1,701 @@
...
+            # Update the slicing indices for the current dimension
+            indices[i] = src_slice
+
+    # Slice single_param to obtain the output matching dist_param's shape
+    to_output = single_param[tuple(indices)]
+
+    return to_output
+
+
+def _check_gradients(model_distributed, model_single, main_grad_check=False):
```
timmoon10, line 169: Similar questions as in `_check_outputs`.

tests/pytorch/distributed/run_numerics.py
```diff
@@ -0,0 +1,701 @@
...
+        {"bias": False},
+        {"init_method": _constant},
+        {"fuse_wgrad_accumulation": True},
+        {"return_bias": True},
+        {"params_dtype": torch.float16},
+    ]
+    for kwargs in kwargs_list:
+        for parallel_mode in ["column", "row"]:
+            for sequence_parallel in [False, True]:
+                _test_linear(parallel_mode, sequence_parallel, **kwargs)
```
timmoon10, line 334: I find it helpful to print the test configs so we can narrow down in the case there's a test failure:
```suggestion
                dist_print(f"_test_linear with {kwargs=}, {parallel_mode=}, {sequence_parallel=}")
                _test_linear(parallel_mode, sequence_parallel, **kwargs)
```
We'd want to do similar logging in all other layer tests.

tests/pytorch/distributed/test_numerics.py
```diff
@@ -0,0 +1,87 @@
...
+        or "NUMERICAL CHECK PASSED" not in result.stdout.decode()
+    ):
+        raise AssertionError(result.stderr.decode())
+
+
+all_boolean = [True, False]
+
+
+@pytest.mark.parametrize("fp8", all_boolean)
+def test_linear(fp8):
```
timmoon10, line 56: Having a separate test for each layer is nice for error reporting, but it adds a few seconds overhead to launch the parallel jobs. It's more scalable to launch a single parallel job and to test all layers and layer configurations internally.

tests/pytorch/distributed/run_numerics.py
pggPL: I changed test_numerics.py to always run with 2 processes - it makes tests simpler.

tests/pytorch/distributed/test_numerics.py
pggPL: Changed. It takes 1 minute not 3 minutes as before.

tests/pytorch/distributed/run_numerics.py
pggPL: changed

tests/pytorch/distributed/run_numerics.py
pggPL: I use the same values as test_comm_gemm_overlap. Tolerances you suggested are not sufficient enough.

tests/pytorch/distributed/run_numerics.py
pggPL: It's already printed - look at the wrapper run_distributed_test

tests/pytorch/distributed/run_numerics.py
```diff
@@ -0,0 +1,701 @@
...
+
+
+def _check_gradients(model_distributed, model_single, main_grad_check=False):
+    for i, ((name, param_d), param_s) in enumerate(
+        zip(model_distributed.named_parameters(), model_single.parameters())
+    ):
+        numerics_failed = torch.tensor([0], dtype=torch.uint8, device="cuda")
+        grad_failed, grad_info = None, None
+        rtol = 0.125 if FP8 else 0.025
+        atol = 0.0625 if FP8 else 0.00125
```
ptrendx, line 175: Could we get those tolerances in 1 place rather than having copies in multiple functions?

tests/pytorch/distributed/run_numerics.py
```diff
@@ -0,0 +1,701 @@
...
+
+    # Prepare input tensors
+    input_single_node = torch.randn((BATCH_SIZE, HIDDEN_SIZE)).cuda().to(params_dtype)
+
+    if parallel_mode == "row":
+        # Split input across GPUs for row parallelism
+        split_size = HIDDEN_SIZE // WORLD_SIZE
+        input_distributed = input_single_node[:, :split_size].clone()
+        if WORLD_RANK == 1:
+            input_distributed = input_single_node[:, split_size:].clone()
```
ptrendx, line 275: Why can't we just have
```suggestion
            input_distributed = input_single_node[:, split_size*WORLD_RANK:split_size*(WORLD_RANK+1)].clone()
```
?
Then we could maybe lift the 2 rank only restriction?

tests/pytorch/distributed/run_numerics.py
```diff
@@ -0,0 +1,701 @@
...
+    elif parallel_mode == "column":
+        if sequence_parallel:
+            # Duplicate input for sequence parallelism
+            input_single_node = (
+                torch.empty((WORLD_SIZE * BATCH_SIZE, HIDDEN_SIZE)).cuda().to(params_dtype)
+            )
+            input_distributed = torch.randn((BATCH_SIZE, HIDDEN_SIZE)).cuda().to(params_dtype)
+            input_single_node[:BATCH_SIZE, :] = input_distributed
+            input_single_node[BATCH_SIZE:, :] = input_distributed
```
ptrendx, line 284: I don't get it. Is this emulating the allgather?

tests/pytorch/distributed/run_numerics.py
ptrendx: Also, these tolerances seem high. Since the inputs are the same the tolerances should only depend on the output type, rather than FP8 which is an input type. Do you really need those tolerances to be this high?

tests/pytorch/distributed/run_numerics.py
timmoon10: What layers are you seeing these numerical errors? The reason we have such loose tolerances in the other tests is because the tensor sizes are so large. For the record I think this was a mistake, and it gives us much more confidence if we have tight tolerances with small tensor sizes. We also test many of the same kernels with tight tolerances in [`test_fusible_ops.py`](https://github.com/NVIDIA/TransformerEngine/blob/main/tests/pytorch/test_fusible_ops.py) and [`test_fusible_ops_distributed.py`](https://github.com/NVIDIA/TransformerEngine/blob/main/tests/pytorch/test_fusible_ops_distributed.py), so I wouldn't expect us to have significant correctness issues.

tests/pytorch/distributed/run_numerics.py
timmoon10: Seems like we're missing an `if WORLD_RANK == 0`/`else`?

pggPL: I changed to gather

tests/pytorch/distributed/run_numerics.py
pggPL: nice

tests/pytorch/distributed/run_numerics.py
pggPL: answered in discussion above

tests/pytorch/distributed/run_numerics.py
pggPL: For FP8 layer norm_linear, layernorm_mlp and transformer_layer cast some values to FP8, so I guess that is the reason. I obtain different outputs in Linear with row parallelism. It's behavior of FP8 gemm, which I do not fully understand. It's not because of communication - sum of outputs of these gemms does not equal output of the gemm in non-distributed case by margin 10^-4.

For FP32/FP16 I made all tensor smaller and it seems to be working. I was still getting problem with transformer_layer with layer_type=encoder, so I removed it from the tests.

tests/pytorch/distributed/run_numerics.py
```diff
@@ -112,12 +127,24 @@ def dist_print(msg, src=None, end="\n", error=False):
...
 
 
+def _get_tolerances(dtype):
+    if FP8:
+        return {"rtol": 0.125, "atol": 0.0625}
+
+    if dtype == torch.float16:
+        return {"rtol": 1e-3, "atol": 1e-5}
+    if dtype == torch.float32:
+        return {"rtol": 1.3e-6, "atol": 1e-5}
```
timmoon10, line 137: Anticipating BF16 tests in the future:
```suggestion
    if dtype == torch.float16:
        return {"rtol": 1e-3, "atol": 1e-5}
    if dtype == torch.bfloat16:
        return {"rtol": 1.6e-2, "atol": 1e-5}
    if dtype == torch.float32:
        return {"rtol": 1.3e-6, "atol": 1e-5}
    raise ValueError(f"Unsupported dtype ({dtype})")
```

tests/pytorch/distributed/run_numerics.py
timmoon10: Makes sense, if there's FP8 compute then that is the limiting factor for numerical accuracy. Having tight tolerances for the non-FP8 case is the main thing.

timmoon10 (approved):
Overall LGTM. These tests are going to be very helpful.

ptrendx (approved):

PR #1208, opened 2024-09-26
Improve CP P2P efficiency

yenchenlin:
# Description

The original implementation saves each attention layer's whole `K`, `V`.
Instead, we can discard used `K`, `V` to ensure each GPU only holds the current and the next `K`, `V`.

### Memory profiling results using the toy unit-test

Original:
Screenshot 2024-09-26 at 1 13 04 PM

This MR:
Screenshot 2024-09-26 at 1 13 16 PM

One can see that the unused `K`, `V` are immediately released.

## Type of change

- [ ] Infra/Build change
- [x] Code refractor


MaciejBalaNV:
Hey @xrennvidia did you have a chance to take a look at this? Perhaps it would help with large-scale inference (>= 32 GPUs) P2P perf issues we've observed.

xrennvidia:
> Hey @xrennvidia did you have a chance to take a look at this? Perhaps it would help with large-scale inference (>= 32 GPUs) P2P perf issues we've observed.

Sorry for the delayed reply, I was OOTO. Not sure if @yenchenlin still woks on the PR, probably no :) but this is to reduce memory overheads, not about perf, I think it probably will not change perf.

And current CP implementation in TE was mainly developed for training. The fwd function can be used for inference prefill phase, but cannot work for generation phase.

What's the perf issue that you think this could be helpful for?

MaciejBalaNV:
> but this is to reduce memory overheads, not about perf, I think it probably will not change perf.
I guess that's right.

> The fwd function can be used for inference prefill phase, but cannot work for generation phase.
I think @yenchenlin was using this code in the context of diffusion models, so only "prefill" phase applies. 

> What's the perf issue that you think this could be helpful for?
I was talking about the P2P not overlapping compute with communication effectively for very large scale inference (32 GPUs), even with CUDA Graphs. Not sure if you have any ideas for improving the performance in such case.

xrennvidia:
This cannot solve the issue.

That just means you over-scaled the system with P2P and CP32 is too big. You need to think about smaller CP size for P2P, or uses other cp_comm_type, such as a2a+p2p.

PR #1220, opened 2024-09-30
Removed the unused options from GroupedLinear docs and fixed the bug with offsets

ptrendx:
# Description

Some options in the GroupedLinear were unused (like parallel_mode) and tensor-parallelism related. Can't fully remove them now since MCore apparently sets them, but at least removing them from documentation to not cause confusion and false expectations.

Also, the offsets for the FP8 meta were set as global Python variables. This is not correct, since there could be multiple instances of the GroupedLinear with different number of GEMMs inside. Changed those to be per-module parameters.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactor

### DIFF
--- a/transformer_engine/pytorch/module/grouped_linear.py
+++ b/transformer_engine/pytorch/module/grouped_linear.py
@@ -44,6 +44,7 @@ from ..export import is_in_onnx_export_mode
 
 __all__ = ["GroupedLinear"]
 
+
 class _GroupedLinear(torch.autograd.Function):
     """GroupedLinear semi-top level module
     Calls custom cuda extensions.
@@ -303,7 +304,12 @@ class _GroupedLinear(torch.autograd.Function):
                         )
                 else:
                     if not ctx.fp8_meta["recipe"].override_linear_precision.wgrad:
-                        indices = list(range(ctx.offsets["grad_output"], ctx.offsets["grad_output"] + ctx.num_gemms))
+                        indices = list(
+                            range(
+                                ctx.offsets["grad_output"],
+                                ctx.offsets["grad_output"] + ctx.num_gemms,
+                            )
+                        )
                         grad_output_c, grad_output_t = fp8_multi_cast_transpose_fused(
                             grad_output_mats,
                             ctx.fp8_meta["scaling_bwd"],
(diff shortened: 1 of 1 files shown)


transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -536,23 +518,6 @@ class GroupedLinear(TransformerEngineBaseModule):
...
-             used as TP (tensor parallel) world size when TP groups are not formed during
-             initialization. In this case, users must call the
-             `set_tensor_parallel_group(tp_group)` method on the initialized module before the
-             forward pass to supply the tensor parallel group needed for tensor and sequence
-             parallel collectives.
-    parallel_mode : {None, 'column', 'row'}, default = `None`
-                   used to decide whether this GroupedLinear layer is Column Parallel Linear or Row
-                   Parallel Linear as described `here <https://arxiv.org/pdf/1909.08053.pdf>`_.
-                   When set to `None`, no communication is performed.
-
```
ksivaman, line 555: We can remove them from `_GroupedLinear`, but we still need them in the `GroupedLinear` as they are used, right? In this case we should retain the docs.

ptrendx: Well, that's the point - they are not actually used in the code and do not really work (e.g. there is no communication happening), so I do not want people to think that they can use them.
I need to keep the actual parameters in the constructor to not break MCore but at the very least do not want to advertise functionality that is not there.

ksivaman (approved):
LGTM

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -74,12 +62,9 @@ def forward(
         fp8_meta: Dict[str, Any],
         fuse_wgrad_accumulation: bool,
         cpu_offloading: bool,
-        tp_group: Union[dist_group_type, None],
-        tp_size: int,
         sequence_parallel: bool,
-        tensor_parallel: bool,
         activation_dtype: torch.dtype,
-        parallel_mode: Union[str, None],
+        offsets: Dict[str, int],
```
timmoon10, line 67: Something like `fp8_meta_offsets` may be more readable:
```suggestion
        fp8_meta_offsets: Dict[str, int],
```

transformer_engine/pytorch/module/grouped_linear.py
timmoon10: We could set the kwarg defaults to `None` and print a warning if they are set.

timmoon10 (approved):
LGTM aside from some stylistic suggestions.

transformer_engine/pytorch/module/grouped_linear.py
ptrendx: Agreed, but at this point I want to just minimize the exposure rather than properly fix this (also I do not know what exactly is passed to those modules inside MCore, so there is a possibility that such warning would be triggered by them even if they are otherwise fine).

PR #1229, opened 2024-10-08
[Pytorch] Check gradient in test numerics

pggPL:
# Description

I added checking gradient numerical correctness to the test_numerics.py. 

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- checking gradient numerical correctness to the test_numerics.py

# Checklist:


tests/pytorch/test_numerics.py
```diff
@@ -67,6 +67,10 @@ def __init__(self, hidden_size, eps, num_attention_heads, embed, num_layers, seq
     "126m": ModelConfig(768, 1e-5, 12, 64, 12, 2048),
 }
 
+model_configs_small = {
+    "small": ModelConfig(128, 1e-5, 8, 36, 4, 128),
+}
```
timmoon10, line 72: I see we're using an idiom of `model_configs | model_configs_small` throughout the code. It seems more natural if we just add `"small"` to `model_configs`:
```suggestion
    "126m": ModelConfig(768, 1e-5, 12, 64, 12, 2048),
    "small": ModelConfig(128, 1e-5, 8, 36, 4, 128),
```
Anyways, most of the tests shouldn't be iterating over all the model configs. We should explicitly pick the configs:
```python
@pytest.mark.parametrize("model", ["small"])
```

tests/pytorch/test_numerics.py
```diff
@@ -110,14 +114,12 @@ def dtype_tols(dtype: torch.dtype) -> Dict[str, float]:
...
-    l1: List[torch.Tensor],
-    l2: List[torch.Tensor],
-    atol: float,
+    l1: List[torch.Tensor], l2: List[torch.Tensor], atol: float, rtol: float = None
 ) -> bool:
     """Ensures two lists are equal."""
     assert len(l1) == len(l2), "Unequal number of outputs."
     for i, (t1, t2) in enumerate(zip(l1, l2)):
-        result = torch.allclose(t1, t2, atol=atol)
+        result = torch.allclose(t1, t2, atol=atol, **({"rtol": rtol} if rtol is not None else {}))
```
timmoon10, line 122: Nit: I think we're trying too hard to make this a one-liner.
```suggestion
        tols = dict(atol=atol)
        if rtol is not None:
            tols["rtol"] = rtol
        result = torch.allclose(t1, t2, **tols)
```

timmoon10 (approved):
Overall this is good, although I have some stylistic comments.

One thing to keep in mind is to be judicious about adding tests. If we have `n` options, parametrizing naively them will result in `O(2^n)` tests:
```python
@pytest.mark.parametrize(opt1, [True, False])
...
@pytest.mark.parametrize(optn, [True, False])
def test_something(opt1, ..., optn):
    ...
```
We only have `O(2*n)` tests if we break this up into smaller pieces:
```python
def _test_something(opt1=default1, ..., optn=defaultn):
    ...

@pytest.mark.parametrize(opt1, [True, False])
def test_something_opt1(opt1):
    _test_something(opt1=opt1)

...
```
As a warning, we run >10k tests in `test_sanity.py`.

tests/pytorch/test_numerics.py
```diff
@@ -868,11 +870,25 @@ def test_gpt_accuracy(dtype, bs, model, parallel_attention_mlp):
...
+
     # Check output.
-    if dtype == torch.float32:
-        assert_allclose(te_outputs[0], torch_outputs[0], 5e-3)
-    else:
-        assert_allclose(te_outputs[0], torch_outputs[0], 5e-2)
+    assert_allclose(te_outputs[0], torch_outputs[0], atol[dtype])
+
+    # Check gradients, only for small model
+    if model == "small":
```
ksivaman, line 883: Were these tolerances too tight for the `126m` model?

ksivaman (commented):
While you're at it, could you just for future reference document the needed tolerances to pass the larger model, even if we don't end up including it?

ksivaman (approved):

tests/pytorch/test_numerics.py
pggPL: Yes, there were differences, such as 0.2 versus 0.5. These were exceptions; for most entries in the considered tensors, the distances were very close, but for a few entries, the difference was significant.

tests/pytorch/test_numerics.py
pggPL: I changed that. Only 3 models iterate through "small" and "126m".

pggPL: Ok, I need to add small model to more tests. To not increase number of tests I removed 126m run from *_accuracy test which needed to run small.

tests/pytorch/test_numerics.py
```diff
@@ -119,17 +116,26 @@ def assert_allclose(
...
-                f"Outputs not close enough in tensor at idx={i}. "
-                f"Location of the maximum difference: {m.item()} "
-                f"with {t1.flatten()[m].item()} vs {t2.flatten()[m].item()} "
-                f"(diff {diff[m].item()})."
-            )
-            raise AssertionError(msg)
+            diff = torch.abs(t1 - t2)
+            tol = atol + (rtol * torch.abs(t2))
+            exceed_mask = diff > tol
+            if exceed_mask.any():
```
timmoon10, line 127: `exceed_mask` will miss NaNs. To handle weird edge cases like this, we should always throw an error if `allclose` fails.

tests/pytorch/test_numerics.py
```diff
@@ -1544,7 +1573,7 @@ def _generate_random_numbers(n, total_sum):
 @pytest.mark.parametrize("dtype", param_types)
 @pytest.mark.parametrize("num_gemms", [3, 6])
 @pytest.mark.parametrize("bs", batch_sizes)
-@pytest.mark.parametrize("model", model_configs.keys())
+@pytest.mark.parametrize("model", "126m")
```
timmoon10, line 1576:
```suggestion
@pytest.mark.parametrize("model", ["126m"])
```
There are a few other places this is needed.

tests/pytorch/test_numerics.py
pggPL: I added throwing error  always.

timmoon10 (approved):

PR #1230, opened 2024-10-08
Fused Attention Support 64-bit Ragged Offsets for Large THD Tensors

mgoldfarb-nvidia:
# Description

This PR updates the data type of the ragged offset tensors, e.g. [offset_q](https://github.com/NVIDIA/TransformerEngine/blob/3b89c36f0e7427199e4e87076b8a5e1545d70346/transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu#L144), in the THD path of cuDNN attention. For long sequences, these offset tensors could have `int32` overflow. This PR updates them to `int64` for cuDNN 9.5+.

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

Add logic to determine if cuDNN 9.5+ is available to support 64-bit offset.

# Checklist:

- [X] I have made corresponding changes to the documentation

### DIFF
--- a/transformer_engine/common/fused_attn/fused_attn.cpp
+++ b/transformer_engine/common/fused_attn/fused_attn.cpp
@@ -80,7 +80,18 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
   const int sm_arch_ = cuda::sm_arch(device_id);
   NVTE_CHECK(q_dtype == kv_dtype, "Q and KV must have the same data type.");
   NVTE_QKV_Format qkv_format = nvte_get_qkv_format(qkv_layout);
+  NVTE_QKV_Layout_Group layout_group = nvte_get_qkv_layout_group(qkv_layout);
   auto cudnn_runtime_version = cudnnGetVersion();
+
+  // For ragged offsets we only support 32-bit prior to cuDNN 9.5
+  // Only used when THD format is requested.
+  const bool requires_64bit_ragged_offset =
+      (qkv_format == NVTE_THD && fused_attn::get_ragged_offset_dtype(
+                                     layout_group, num_attn_heads, num_gqa_groups, max_seqlen_q,
+                                     max_seqlen_kv, head_dim_qk, head_dim_v) == DType::kInt64);
+  const bool supported_ragged_offset_size =
+      (!requires_64bit_ragged_offset || cudnn_runtime_version >= 90500);
+
   if (((q_dtype == NVTEDType::kNVTEFloat8E4M3) || (q_dtype == NVTEDType::kNVTEFloat8E5M2)) &&
       (sm_arch_ >= 90) && (bias_type == NVTE_Bias_Type::NVTE_NO_BIAS) &&
       (((cudnn_runtime_version >= 8900) && (qkv_layout == NVTE_QKV_Layout::NVTE_T3HD) &&
@@ -91,7 +102,8 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
         ((qkv_format == NVTE_QKV_Format::NVTE_BSHD) ||
          (qkv_format == NVTE_QKV_Format::NVTE_SBHD)) &&
         ((attn_mask_type == NVTE_Mask_Type::NVTE_CAUSAL_MASK) ||
-         (attn_mask_type == NVTE_Mask_Type::NVTE_NO_MASK))))) {
+         (attn_mask_type == NVTE_Mask_Type::NVTE_NO_MASK)))) &&
+      !requires_64bit_ragged_offset) {
     if (cudnn_runtime_version >= 8900) {
       backend = NVTE_Fused_Attn_Backend::NVTE_FP8;
     } else {
@@ -203,6 +215,9 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
            flag_arb)) {
         backend = static_cast<NVTE_Fused_Attn_Backend>(env_backend);
       }
+    }
+    if (!supported_ragged_offset_size) {
+      backend = NVTE_Fused_Attn_Backend::NVTE_No_Backend;
     }
     if (cudnn_runtime_version < 8901 &&
         backend == NVTE_Fused_Attn_Backend::NVTE_F16_max512_seqlen) {
--- a/transformer_engine/common/fused_attn/fused_attn.cpp
+++ b/transformer_engine/common/fused_attn/fused_attn.cpp
@@ -130,7 +130,8 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
          (qkv_layout == NVTE_QKV_Layout::NVTE_BS3HD) ||
          (qkv_layout == NVTE_QKV_Layout::NVTE_BSHD_BS2HD) ||
          (qkv_layout == NVTE_QKV_Layout::NVTE_BSHD_BSHD_BSHD)) &&
-        ((window_size_left == -1) && (window_size_right == -1 || window_size_right == 0))) {
+        ((window_size_left == -1) && (window_size_right == -1 || window_size_right == 0)) &&
+        !requires_64bit_ragged_offset) {
       flag_m512 = true;
     }
     if (  // architecture
@@ -195,7 +196,9 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
               max_seqlen_q == max_seqlen_kv)) &&
             dropout == 0.0 && bias_type == NVTE_Bias_Type::NVTE_NO_BIAS &&
             (qkv_format == NVTE_QKV_Format::NVTE_BSHD ||
-             qkv_format == NVTE_QKV_Format::NVTE_SBHD)))))) {
+             qkv_format == NVTE_QKV_Format::NVTE_SBHD))))) &&
+        // check 64-bit ragged offset support
+        (supported_ragged_offset_size)) {
       flag_arb = true;
     }
     if (((max_seqlen_q > 512) || (max_seqlen_kv > 512)) && (flag_arb == true)) {
@@ -215,9 +218,6 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
            flag_arb)) {
         backend = static_cast<NVTE_Fused_Attn_Backend>(env_backend);
       }
-    }
-    if (!supported_ragged_offset_size) {
-      backend = NVTE_Fused_Attn_Backend::NVTE_No_Backend;
     }
     if (cudnn_runtime_version < 8901 &&
         backend == NVTE_Fused_Attn_Backend::NVTE_F16_max512_seqlen) {
--- a/transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
+++ b/transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
@@ -58,6 +58,7 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
     cudnn_frontend::DataType_t tensorType, void *workspace, size_t *workspace_size,
     cudaStream_t stream, cudnnHandle_t handle) {
   using namespace transformer_engine;
+
   bool is_bias = (bias_type == NVTE_Bias_Type::NVTE_POST_SCALE_BIAS);
   bool is_alibi = (bias_type == NVTE_Bias_Type::NVTE_ALIBI);
   bool is_causal = ((mask_type == NVTE_Mask_Type::NVTE_CAUSAL_MASK) ||
@@ -75,7 +76,8 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
   if (is_ragged) {
     NVTE_CHECK(is_padding, "Ragged QKV input requires padding or padding_causal mask!");
   }
-  auto cudnn_runtime_version = cudnnGetVersion();
+  const auto cudnn_runtime_version = cudnnGetVersion();
+  const DType ragged_offset_type = cudnn_runtime_version >= 90500 ? DType::kInt64 : DType::kInt32;
 
   try {
     FADescriptor_v1 descriptor{b,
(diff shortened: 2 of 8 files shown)


transformer_engine/common/fused_attn/utils.cu
```diff
@@ -337,7 +337,7 @@ cudnn_frontend::Operation ternary_pw_op_create(cudnn_frontend::Tensor const &xDe
 }
 
 // convert cu_seqlens_q to qkv/o_ragged_offset and actual_seqlens_q
-__global__ void cu_seqlens_to_offsets(size_t b, size_t h, size_t d, int32_t *cu_seqlens_q,
+__global__ void cu_seqlens_to_offsets(int64_t b, int64_t h, int64_t d, int32_t *cu_seqlens_q,
                                       int32_t *actual_seqlens_q, int32_t *qkv_ragged_offset,
                                       int32_t *o_ragged_offset) {
```
cyanguwa, line 342: Do `qkv_ragged_offset` and `o_ragged_offset` need to be `int64_t`?

mgoldfarb-nvidia: Good catch, indeed they do.

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -145,22 +145,22 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
                                        .set_name("offset_q")
                                        .set_dim({b + 1, 1, 1, 1})
                                        .set_stride({1, 1, 1, 1})
-                                       .set_data_type(fe::DataType_t::INT32));
+                                       .set_data_type(fe::DataType_t::INT64));
```
zlsh80826, line 148: Is this compatible with old cuDNN? If not, we need a variable for different cuDNN runtime version.

```
auto offset_type = fe::DataType_t::INT64;
if (cudnn run time version < ???) { offset_type = fe::DataType_t::INT32; }
```

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -353,15 +353,15 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
...
+      void *devOffsetsO = static_cast<int8_t *>(devOffsetsV) + (b + 1) * sizeof(int64_t);
       NVTE_QKV_Layout_Group layout_group = nvte_get_qkv_layout_group(layout);
       cu_seqlens_padded_to_offsets<<<grid, nthreads_per_block, 0, stream>>>(
           layout_group, b, h, hg, d_qk, d_v, static_cast<int32_t *>(devPtrSeqOffsetsQ),
-          static_cast<int32_t *>(devPtrSeqOffsetsKV), static_cast<int32_t *>(devOffsetsQ),
-          static_cast<int32_t *>(devOffsetsK), static_cast<int32_t *>(devOffsetsV),
-          static_cast<int32_t *>(devOffsetsO));
+          static_cast<int32_t *>(devPtrSeqOffsetsKV), static_cast<int64_t *>(devOffsetsQ),
+          static_cast<int64_t *>(devOffsetsK), static_cast<int64_t *>(devOffsetsV),
+          static_cast<int64_t *>(devOffsetsO));
```
zlsh80826, line 364: Same, we need to extend cu_seqlens_padded_to_offsets for both `int32_t` and `int64_t` dtype and selecting the kernel according to the runtime cuDNN version.

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
mgoldfarb-nvidia: Asked the cudnn folks and indeed this feature is not in cuDNN <9.5 so I will refactor to check. We will also need to add some more logic to warn the user if they use configs that require too large of size.

cyanguwa:
I think we can just use int32 for cuDNN <= 9.5, and int64 for cuDNN >= 9.6. Detecting the maximum required from batch to batch using `get_ragged_offset_dtype` might be unnecessary?

mgoldfarb-nvidia:
> I think we can just use int32 for cuDNN <= 9.5, and int64 for cuDNN >= 9.6. Detecting the maximum required from batch to batch using `get_ragged_offset_dtype` might be unnecessary?

One benefit I see is that if a caller provides a large size that exceeds int32 they get an error. I think currently the behavior is undefined if the int32 overflows. Thoughts?

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -80,7 +80,18 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
   const int sm_arch_ = cuda::sm_arch(device_id);
   NVTE_CHECK(q_dtype == kv_dtype, "Q and KV must have the same data type.");
   NVTE_QKV_Format qkv_format = nvte_get_qkv_format(qkv_layout);
+  NVTE_QKV_Layout_Group layout_group = nvte_get_qkv_layout_group(qkv_layout);
   auto cudnn_runtime_version = cudnnGetVersion();
+
+  // For ragged offsets we only support 32-bit prior to cuDNN 9.5
```
cyanguwa, line 86: "prior to cuDNN 9.6" maybe?

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -91,7 +102,8 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
         ((qkv_format == NVTE_QKV_Format::NVTE_BSHD) ||
          (qkv_format == NVTE_QKV_Format::NVTE_SBHD)) &&
         ((attn_mask_type == NVTE_Mask_Type::NVTE_CAUSAL_MASK) ||
-         (attn_mask_type == NVTE_Mask_Type::NVTE_NO_MASK))))) {
+         (attn_mask_type == NVTE_Mask_Type::NVTE_NO_MASK)))) &&
+      supported_ragged_offset_size) {
```
cyanguwa, line 106: This section is for FP8. We have two versions of FP8 implementation. The older implementation supports T3HD but int32 regardless of cuDNN version. The newer implementation doesn't support THD. So I think we can maybe do this here:
```&& !requires_64bit_ragged_offset```

File: transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu [hunk 6]
```diff
@@ -77,6 +79,13 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
   }
   auto cudnn_runtime_version = cudnnGetVersion();
 
+  // We choose between 32-bit and 64-bit offsets depending on need.
+  // This allows us to support older cuDNN runtimes gracefully.
+  const NVTE_QKV_Layout_Group layout_group = nvte_get_qkv_layout_group(layout);
+  const DType ragged_offset_type = ç(layout_group, h, hg, s_q, s_kv, d_qk, d_v);
```
cyanguwa, line 85: Typo for the function name?

cyanguwa: Or maybe this is not the latest version of your code? :)

transformer_engine/common/fused_attn/utils.cu
```diff
@@ -393,6 +412,60 @@ __global__ void cu_seqlens_padded_to_offsets(NVTE_QKV_Layout_Group layout_group,
...
+      offsets_qkvo[0] = num_attn_heads * head_dim_qk * max_seqlen_q;
+      offsets_qkvo[1] = 2 * num_gqa_groups * head_dim_qk * max_seqlen_kv;
+      offsets_qkvo[2] = offsets_qkvo[1];
+      break;
+  }
+
+  offsets_qkvo[3] = num_attn_heads * head_dim_qk * max_seqlen_q;
+
+  size_t max_offset = *std::max_element(offsets_qkvo.begin(), offsets_qkvo.end());
+  if (max_offset > std::numeric_limits<int32_t>::max()) {
```
cyanguwa, line 462: I think we should set the required type to INT64 when cuDNN >= 9.5. Otherwise, if some batches require INT32 and others INT64, we'd have to recompile the graph between batches because the `.set_data_type()` attribute has changed.

mgoldfarb-nvidia: Wouldn't a change to the parameters passed to this function also indicate a change the dimensions of the tensors requiring a new graph to be generated?

mgoldfarb-nvidia: As discussed we will use 64-bit for 9.5+ so the calls to cuDNN will only check the version. We will keep this function to report to the user that their config requires 64 bit if not available and recommend upgrading.

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
mgoldfarb-nvidia: Fixed

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -204,6 +216,11 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
         backend = static_cast<NVTE_Fused_Attn_Backend>(env_backend);
       }
     }
+    if (!supported_ragged_offset_size && backend != NVTE_Fused_Attn_Backend::NVTE_No_Backend) {
+      backend = NVTE_Fused_Attn_Backend::NVTE_No_Backend;
+      std::cout << "Warning: Tensor sizes are too large for 32-bit THD/ragged offsets."
+                << " Please upgrade your cuDNN version to 9.5.0+ if possible." << std::endl;
+    }
```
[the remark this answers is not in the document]
mgoldfarb-nvidia: SG I can remove the error message.

I agree we should overhaul the checking here its has low readability and also hard to debug when we get an unexpected no backend.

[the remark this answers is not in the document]
cyanguwa: FE is developing some `check_support()` mechanism, which returns error codes and messages, so we're planning to adopt that when it's ready.

transformer_engine/common/fused_attn/utils.cu
cyanguwa: Good point - different parameters would cause a new graph to be generated too.

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -204,6 +216,9 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
         backend = static_cast<NVTE_Fused_Attn_Backend>(env_backend);
       }
     }
+    if (!supported_ragged_offset_size) {
+      backend = NVTE_Fused_Attn_Backend::NVTE_No_Backend;
+    }
```
cyanguwa, line 221: I think you might have to separate the `NVTE_F16_max512_seqlen` backend and `NVTE_F16_arbitrary_seqlen` backend. `NVTE_F16_max512_seqlen`  doesn't support INT64 regardless of the cuDNN version, so it should have `!requires_64bit_ragged_offset` (like with the FP8 backend), whereas `NVTE_F16_arbitrary_seqlen` should have `!supported_ragged_offset_size`.

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -311,10 +313,14 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
           offset_v, offset_o, dropout_seed, dropout_offset] =
         get_graph(sdpa_f16_fprop_cache, descriptor);
 
-    auto plan_workspace_size = mha_graph->get_workspace_size();
     // Exit to request upper level API to allocate memory if needed
-    size_t actual_seqlen_workspace_size = 2 * b * sizeof(int32_t);
-    size_t seqlen_offsets_workspace_size = 4 * (b + 1) * sizeof(int32_t);
+    // n.b. Care should be taken to align each of the added worksapce tensors to their type.
```
cyanguwa, line 317: Is the alignment value based on the type, or is it fixed for all types because of some cuDNN implementation requirements?

"worksapce -> workspace" :)

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -311,10 +313,14 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
...
+    // We do this by adding padding at the end of each separate allocation.
+    auto plan_workspace_size = alignTo<16>(mha_graph->get_workspace_size());
+    const size_t actual_seqlen_workspace_size = alignTo<16>(2 * b * sizeof(int32_t));
```
cyanguwa, line 320: How come we don't pad each of the two actual_seqlen tensors, like we did with the 4 ragged_offset tensors below?

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -693,11 +703,14 @@ void fused_attn_arbitrary_seqlen_bwd_impl(
...
+    const size_t num_bytes_per_ragged_offset =
+        alignTo<16>((b + 1) * typeToSize(ragged_offset_type));
+    const size_t seqlen_offsets_workspace_size = 4 * num_bytes_per_ragged_offset;
```
cyanguwa, line 713: Same comments as for fwd above.

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
mgoldfarb-nvidia: Alignment was chosen to be 16 since that is the largest alignment noted in https://docs.nvidia.com/cuda/cuda-c-programming-guide/#built-in-vector-types Its not a requirement but seemed like a natural value to use.

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
mgoldfarb-nvidia: Its a good point I'll fix it so that each tensor gets aligned individually to be consistent and the most "safe".

transformer_engine/common/fused_attn/fused_attn.cpp
mgoldfarb-nvidia: Fixed.

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
zlsh80826: I think actually 8 bytes alignment is enough for int64_t, but 16 bytes is also fine.

zlsh80826 (approved):
LGTM!

PR #1233, opened 2024-10-09
Support CUDA Graph for MoE models

buptzyb:
# Description

Different from non-MoE models like llama2, MoE models have dynamic shaped activations in FFN layers, so one cudagraph can only capture a part of one transformer layer, instead of covering the whole layer. We call this a "breaking-layer" cudagraph mode. This PR adds breaking-layer cudagraph supports for MoE models on the TE side, and fixes several related bugs in TE.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Add `is_initialized()` method in CudaRNGStatesTracker to align with what is already done in MCore.
- Fix wrong `per_callable_module_params` order bug in `_make_graphed_callables` when `_order` is given.
- Fix warmup argument mismatch bug in `_make_graphed_callables` when `_order` is given.
- Fix fp8 accuracy issue by adding `fp8_group` argument to `make_graphed_callables()` and modifing `is_first_microbatch`, `skip_fp8_weight_update` and `fp8_meta` code.
- Support MoE models cudagraph by filtering graphed TE modules and model weights during warmup.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/graph.py
+++ b/transformer_engine/pytorch/graph.py
@@ -240,20 +240,7 @@ def _make_graphed_callables(
         if isinstance(module, TransformerEngineBaseModule):
             visited_te_modules.add(module)
 
-    # Filter the weights without gradients in backward. These weights are not in the computation graph so can be removed from cudagraph inputs.
-    no_grad_weights = None
-
-    def get_no_grads(static_input_surface, grad_inputs, func_idx):
-        grad_index = 0
-        none_grads = []
-        for i in range(len(static_input_surface)):
-            if static_input_surface[i].requires_grad:
-                if grad_inputs[grad_index] is None and i >= len(flatten_sample_args[func_idx]):
-                    none_grads.append(i)
-                grad_index += 1
-        return set(none_grads)
-
-    # Run warmup and do the above two filtering.
+    # Run warmup and do the above filtering.
     with torch.cuda.stream(torch.cuda.Stream()):
         for func_idx, func in zip(warmup_func_idx, warmup_func):
             args = sample_args[func_idx]
@@ -274,26 +261,12 @@ def _make_graphed_callables(
                     only_inputs=True,
                     allow_unused=allow_unused_input,
                 )
-                if no_grad_weights is None:
-                    no_grad_weights = get_no_grads(static_input_surface, grad_inputs, func_idx)
                 del outputs, grad_inputs
             # The following code is added specifically for MCore's special requirements,
             # aimed at preventing warmup from altering the control flow.
             for module in func.modules():
                 if hasattr(module, "is_first_microbatch"):
                     module.is_first_microbatch = True
-            if len(no_grad_weights) > 0:
-                per_callable_static_input_surfaces[func_idx] = tuple(
-                    inp
-                    for i, inp in enumerate(per_callable_static_input_surfaces[func_idx])
-                    if i not in no_grad_weights
-                )
-                per_callable_module_params[func_idx] = tuple(
-                    param
-                    for i, param in enumerate(per_callable_module_params[func_idx])
-                    if i + len(flatten_sample_args[func_idx]) not in no_grad_weights
-                )
-            no_grad_weights = None
     torch.cuda.synchronize()
 
     # All captures here share a mempool. To avoid replays corrupting each other's memory,
--- a/transformer_engine/pytorch/graph.py
+++ b/transformer_engine/pytorch/graph.py
@@ -233,8 +233,9 @@ def _make_graphed_callables(
         set(warmup_func_idx)
     ), f"Warmup runs {len(warmup_func)} but only {len(set(warmup_func_idx))} are unique."
 
-    # Filter the TE modules that cudagraph can access.
+    # Filter the TE modules and parameters that cudagraph can access.
     visited_te_modules = set()
+    visited_params = set()
 
     def hook_fn(module, input, output):
         if (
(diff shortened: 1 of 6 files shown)


transformer_engine/pytorch/graph.py
```diff
@@ -200,21 +204,96 @@ def _make_graphed_callables(
...
+
+    def get_no_grads(static_input_surface, grad_inputs, func_idx):
+        grad_index = 0
+        none_grads = []
+        for i in range(len(static_input_surface)):
+            if static_input_surface[i].requires_grad:
+                if grad_inputs[grad_index] is None and i >= len(flatten_sample_args[func_idx]):
+                    none_grads.append(i)
+                grad_index += 1
+        return set(none_grads)
```
timmoon10, line 256: If I understand correctly, you are checking for params that recieve no grads and removing them from the static input surface. Is this safe? A param might still be required in the forward compute even if it is disconnected from the autograd graph, e.g. by calling `tensor.detach()` or using `torch.no_grad()`.

transformer_engine/pytorch/graph.py
```diff
@@ -459,6 +538,17 @@ def new_fwd(*user_args, **user_kwargs):
...
+                                    # Only Set the FP8 meta for the modules included by forward
+                                    continue
+                                fp8_recipe = FP8GlobalStateManager.get_fp8_recipe()
+                                if (
+                                    not fp8_recipe.fp8_mha
+                                    and not fp8_recipe.fp8_dpa
+                                    and hasattr(m, "attention_dropout")
+                                    and m.deterministic
+                                ):
+                                    continue
```
timmoon10, line 551: Why are we skipping the FP8 scale update logic for this case?
```suggestion
```

timmoon10 (commented):
Overall I don't see glaring issues, although I have questions and stylistic suggestions.

@ptrendx @ksivaman @sbhavani What is our priority for this feature? The custom Mcore logic in `make_graphed_callables` is already messy and fragile, and this PR does exacerbate those problems. We should explicitly make a decision before committing to maintain it.

transformer_engine/pytorch/graph.py
yifeis-nv: This is for a deterministic test with FP8. Even without CUDA graphs, we find that FP8_DPA leads to random output, even when set to deterministic mode. With CUDA graphs, we not only need to set DPA to BF16, but also skip the FP8_meta update for DPA. After that, we will finally obtain a stable output under the deterministic test.

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -1206,7 +1209,7 @@ def forward(
                 self.apply_bias and not self.gemm_bias_unfused_add,
                 self.eps,
                 is_first_microbatch,
-                self.fp8,
+                False if self.fp8 is None else self.fp8,
```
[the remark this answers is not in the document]
buptzyb: Thanks, removed.

transformer_engine/pytorch/graph.py
buptzyb: Make sense, thanks for pointing out.

TE always adds all parameters in callable as cudagraph inputs, so here I want to filter out the parameters that are not really covered by cudagraph.

Now I changed to using hooks to record the parameters that are actually accessed in warmup, together with the TE modules recorder hook.

transformer_engine/pytorch/graph.py
```diff
@@ -200,21 +204,96 @@ def _make_graphed_callables(
...
                     grad_outputs=tuple(torch.empty_like(o) for o in outputs if o.requires_grad),
                     only_inputs=True,
                     allow_unused=allow_unused_input,
                 )
+                if no_grad_weights is None:
+                    no_grad_weights = get_no_grads(static_input_surface, grad_inputs, func_idx)
                 del outputs, grad_inputs
+            for module in func.modules():
+                if hasattr(module, "is_first_microbatch"):
+                    module.is_first_microbatch = True
```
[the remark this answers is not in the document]
yifeis-nv: Thank you for the reminder. I also believe that modifications are needed here. The change we originally intended to address is the issue where the fp8 weight caching behavior in MoE leads to different behaviors for the first microbatch compared to other microbatches. If we do not include this piece of code, the warmup process will update `is_first_microbatch` to False, causing all captured graphs to exhibit non-first microbatch behavior, which does not align with our requirements. Therefore, we chose to reset this parameter after warmup.
In summary, our requirement is either to prevent `is_first_microbatch` from being updated during warmup or to reset `is_first_microbatch` after warmup. Choosing the former may require adding a flag to inform MoE that it is currently in the warmup phase, while choosing the latter might necessitate making this code a bit more general. Do you have any input on the modification plan that could serve as a reference for us?

buptzyb:
> Have you tested that it works with Mcore?
Yes, we also made some changes in Mcore to enable MoE cudagraph. You can refer to issue 193 in our Megatron-LM repo.

transformer_engine/pytorch/graph.py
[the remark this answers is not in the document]
timmoon10: I see, so this is Mcore-specific logic. It's uncomfortable that it's made its way into TE, but it's a tricky problem and I can't think of a better solution either, at least without significant changes in Mcore.

We should document what this is doing, especially for TE developers with no knowledge of Mcore.

transformer_engine/pytorch/graph.py
timmoon10: It seems like we're covering up a correctness bug. This may also affect convergence since we are no longer doing amax reductions within MHA (e.g. for the FP8 cast before `qkv` and `proj`).

transformer_engine/pytorch/graph.py
timmoon10: The new approach is much cleaner and easier to understand.

transformer_engine/pytorch/graph.py
yifeis-nv: It makes sense, and I think we should '`continue`' without fp8_mha and fp8_dpa, regardless of whether the test is deterministic or not.

timmoon10: Actually, this change will cause correctness issues. We do a max all-reduce on the amaxes so that FP8 scaling factors are synchronized over the TP group. If we skip the amax reduction, then the MHA's TP communication will be wrong. In particular, we'll all-gather FP8 inputs to the `qkv` GEMM, but the scaling factors will be different for each TP rank.

timmoon10: As far as I can tell, this logic is only relevant to `DotProductAttention`. We could make intent much more obvious with:
```suggestion
                                if (
                                    isinstance(m, DotProductAttention)
                                    and not fp8_recipe.fp8_mha
                                    and not fp8_recipe.fp8_dpa
                                ):
                                    # Don't need to update FP8 meta for non-FP8 DPA
                                    continue
```

transformer_engine/pytorch/graph.py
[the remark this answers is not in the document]
yifeis-nv: Done. I try to explain the reason for this MoE specific logic.

transformer_engine/pytorch/graph.py
yifeis-nv: Thanks! I have modified this.

buptzyb:
Hi @timmoon10 , do you have more suggestions on this PR?

timmoon10 (approved):
LGTM

yaox12:
Merge this PR since pipeline 20710146 passed and Tim approved.

PR #1234, opened 2024-10-09
Save CUDA Graph memory by reusing input and output tensors

buptzyb:
# Description

TE allows for passing multiple callables into `make_graphed_callables()` to share cudagraph pool to save memory. However, each cudagraph has its own input and output data buffer. This causes large memory overhead especially when `_order` is specified so that we must capture graphs for all microbatches. This PR allows for reusing the input or output tensors among cuda graphs captured in one `make_graphed_callables()` according to the pattern of `_order`. Note that reusing output tensors will cause extra DtoD memory copy in cudagraph execution.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Add `reuse_graph_inputs` option to `make_graphed_callables()` API.
- Add `reuse_graph_outputs` option to `make_graphed_callables()` API.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/graph.py
+++ b/transformer_engine/pytorch/graph.py
@@ -60,6 +60,8 @@ def _make_graphed_callables(
     allow_unused_input: bool = False,
     fp8_weight_caching: bool = False,
     sample_kwargs: Optional[SingleOrTuple[Dict[str, Any]]] = None,
+    reuse_graph_inputs=False,
+    reuse_graph_outputs=False,
     _order: Optional[List[int]] = None,
     pool: Optional[Tuple[int, ...]] = None,
 ) -> SingleOrTuple[Callable]:
@@ -87,6 +89,17 @@ def _make_graphed_callables(
         callables = (callables,)
         sample_args = (sample_args,)
         sample_kwargs = (sample_kwargs,)
+    if reuse_graph_inputs:
+        len_args = len(sample_args[0])
+        for arg in sample_args:
+            assert len_args == len(arg), f"Arguments must have same length and shape for reusing."
+        sample_args = list(sample_args)
+        len_kwargs = len(sample_kwargs[0])
+        for kwarg in sample_kwargs:
+            assert len_kwargs == len(
+                kwarg
+            ), f"Keyword arguments must have same length and shape for reusing."
+        sample_kwargs = list(sample_kwargs)
 
     # Check sizes of args
     if _order is None:
@@ -228,8 +241,26 @@ def _make_graphed_callables(
         per_callable_static_grad_inputs = [None] * len(flatten_sample_args)
         fwd_idx = [0] * num_model_chunks
         bwd_idx = [0] * num_model_chunks
-        for c_id in _order:
+        # Following variables are for input/output reusing to save memory.
+        fwd_order_recorder = {}
+        fwd_order_accu = 0
+        per_callable_fwd_idx_recorder = []
+        static_grad_outputs = None
+        static_grad_inputs = []
+        static_grad_inputs_exists = False
+        for idx, c_id in enumerate(_order):
             if c_id > 0:
+                if reuse_graph_inputs or reuse_graph_outputs:
+                    # Record the fwd order pattern for input data reusing.
+                    if c_id in fwd_order_recorder:
+                        fwd_order_recorder[c_id].append(fwd_order_accu)
+                    else:
+                        fwd_order_recorder[c_id] = [fwd_order_accu]
+                    fwd_order_accu += 1
+                    if idx > 1 and _order[idx - 1] < 0:
+                        # It can use the tensor buffer of a previous one.
+                        reuse_fwd_idx = fwd_order_recorder[abs(_order[idx - 1])].pop(0)
+
                 # Capture forward graph for model chunk c_id, microbatch fwd_idx[c_id-1]
                 m_chunk = c_id - 1
                 for l_no in range(num_layers):
@@ -237,13 +268,54 @@ def _make_graphed_callables(
                     per_callable_fwd_idx = (m_chunk * num_microbatches * num_layers) + (
                         fwd_idx[m_chunk] * num_layers + l_no
                     )
+                    if reuse_graph_inputs or reuse_graph_outputs:
+                        per_callable_fwd_idx_recorder.append(per_callable_fwd_idx)
+                        if idx > 1 and _order[idx - 1] < 0:
+                            # It can use the tensor buffer of a previous one.
+                            reuse_per_callable_fwd_idx = per_callable_fwd_idx_recorder[
+                                reuse_fwd_idx * num_layers + l_no
+                            ]
+                            if reuse_graph_inputs:
+                                sample_args[per_callable_fwd_idx] = sample_args[
+                                    reuse_per_callable_fwd_idx
+                                ]
+                                sample_kwargs[per_callable_fwd_idx] = sample_kwargs[
+                                    reuse_per_callable_fwd_idx
+                                ]
+                                flatten_sample_args[per_callable_fwd_idx] = flatten_sample_args[
+                                    reuse_per_callable_fwd_idx
+                                ]
+                                per_callable_static_input_surfaces[per_callable_fwd_idx] = (
+                                    per_callable_static_input_surfaces[reuse_per_callable_fwd_idx][
+                                        : len(flatten_sample_args[per_callable_fwd_idx])
+                                    ]
+                                    + per_callable_static_input_surfaces[per_callable_fwd_idx][
+                                        len(flatten_sample_args[per_callable_fwd_idx]) :
+                                    ]
+                                )
+                            if reuse_graph_outputs:
+                                static_outputs = per_callable_static_outputs[
+                                    reuse_per_callable_fwd_idx
+                                ]
+                                detached_static_outputs = tuple(
+                                    so.detach() for so in static_outputs
+                                )
                     args = sample_args[per_callable_fwd_idx]
                     kwargs = sample_kwargs[per_callable_fwd_idx]
                     fwd_graph = fwd_graphs[per_callable_fwd_idx]
                     with torch.cuda.graph(fwd_graph, pool=mempool):
                         outputs = func(*args, **kwargs)
-                    flatten_outputs, spec = _tree_flatten(outputs)
-                    per_callable_static_outputs[per_callable_fwd_idx] = tuple(flatten_outputs)
+                        flatten_outputs, spec = _tree_flatten(outputs)
+                        if reuse_graph_outputs and idx > 1 and _order[idx - 1] < 0:
+                            for i, static_output in enumerate(detached_static_outputs):
+                                static_output.copy_(flatten_outputs[i])
+                            per_callable_static_outputs[per_callable_fwd_idx] = (
+                                detached_static_outputs
+                            )
+                        else:
+                            per_callable_static_outputs[per_callable_fwd_idx] = tuple(
+                                flatten_outputs
+                            )
                     per_callable_output_unflatten_spec[per_callable_fwd_idx] = spec
                     graph_callables[per_callable_fwd_idx] = func
                 fwd_idx[m_chunk] += 1
@@ -258,9 +330,10 @@ def _make_graphed_callables(
                     static_outputs = per_callable_static_outputs[per_callable_bwd_idx]
                     bwd_graph = bwd_graphs[per_callable_bwd_idx]
                     # For now, assumes all static_outputs require grad
-                    static_grad_outputs = tuple(
-                        torch.empty_like(o) if o.requires_grad else None for o in static_outputs
-                    )
+                    if not reuse_graph_inputs or static_grad_outputs is None:
+                        static_grad_outputs = tuple(
+                            torch.empty_like(o) if o.requires_grad else None for o in static_outputs
+                        )
                     with torch.cuda.graph(bwd_graph, pool=mempool):
                         grad_inputs = torch.autograd.grad(
                             outputs=tuple(o for o in static_outputs if o.requires_grad),
@@ -269,21 +342,29 @@ def _make_graphed_callables(
                             only_inputs=True,
                             allow_unused=allow_unused_input,
                         )
-                    # Constructs a tuple suitable for returning from Graphed.backward:
-                    # Pads out the actually-needed grads with Nones in gradient slots for inputs
-                    # that don't require grad. I couldn't think of a one-liner for this pattern.
-                    static_grad_inputs = []
-                    grad_idx = 0
-                    for arg in static_input_surface:
-                        if arg.requires_grad:
-                            static_grad_inputs.append(grad_inputs[grad_idx])
-                            grad_idx += 1
-                        else:
-                            static_grad_inputs.append(None)  # type: ignore[arg-type]
-                    static_grad_inputs = tuple(static_grad_inputs)  # type: ignore[assignment]
+                        # Constructs a tuple suitable for returning from Graphed.backward:
+                        # Pads out the actually-needed grads with Nones in gradient slots for inputs
+                        # that don't require grad. I couldn't think of a one-liner for this pattern.
+                        if not reuse_graph_outputs:
+                            static_grad_inputs = []
+                        grad_idx = 0
+                        for input_idx, arg in enumerate(static_input_surface):
+                            if arg.requires_grad:
+                                if reuse_graph_outputs and static_grad_inputs_exists:
+                                    if static_grad_inputs[input_idx] is not None:
+                                        static_grad_inputs[input_idx].copy_(grad_inputs[grad_idx])
+                                else:
+                                    static_grad_inputs.append(grad_inputs[grad_idx])
+                                grad_idx += 1
+                            elif not reuse_graph_outputs or not static_grad_inputs_exists:
+                                static_grad_inputs.append(None)  # type: ignore[arg-type]
+                    if reuse_graph_outputs:
+                        static_grad_inputs_exists = True
 
                     per_callable_static_grad_outputs[per_callable_bwd_idx] = static_grad_outputs
-                    per_callable_static_grad_inputs[per_callable_bwd_idx] = static_grad_inputs
+                    per_callable_static_grad_inputs[per_callable_bwd_idx] = tuple(
+                        static_grad_inputs
+                    )
                 bwd_idx[m_chunk] += 1
     else:
         # Capture forward graphs
@@ -514,6 +595,8 @@ def make_graphed_callables(
     num_warmup_iters: int = 3,
     allow_unused_input: bool = False,
     sample_kwargs: Optional[SingleOrTuple[Dict[str, Any]]] = None,
+    reuse_graph_inputs: bool = False,
+    reuse_graph_outputs: bool = False,
     fp8_enabled: bool = False,
     fp8_calibrating: bool = False,
     fp8_recipe: Optional[DelayedScaling] = None,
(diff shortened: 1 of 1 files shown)


transformer_engine/pytorch/utils.py
```diff
@@ -616,3 +616,114 @@ def torch_get_autocast_gpu_dtype() -> torch.dtype:
...
+    ret = _torch_dtype_to_np_typestr_dict.get(dtype)
+    assert ret is not None, f"Unsupported dtype: {dtype}"
+    return ret
+
+
+def volume(d: Sequence[int]):
+    """Calculate the volume of a tensor."""
+    return np.prod(d)
+
+
```
timmoon10: This "volume" terminology is non-standard (PyTorch uses ["numel"](https://docs.pytorch.org/docs/stable/generated/torch.numel.html#torch.numel) and NumPy uses ["size"](https://numpy.org/doc/2.1/reference/generated/numpy.size.html#numpy.size)). Also, this function is only used within `TensorWrapper`, so might as well just use `np.prod` there.
```suggestion
```

transformer_engine/pytorch/utils.py
```diff
@@ -616,3 +616,114 @@ def torch_get_autocast_gpu_dtype() -> torch.dtype:
...
+class TensorWrapper:
```
timmoon10: We only use this class in `make_weak_ref` and I don't think it's worth the effort to make this a robust, externally-visible class. We should give it a name that reflects it's an internal implementation:
```suggestion
class _WeakRefTensor:
```
Also, `torch_dtype_to_np_typestr` and `convert_to_torch_tensor` should be implemented as member functions.

transformer_engine/pytorch/graph.py
```diff
@@ -739,17 +827,27 @@ def forward_func(*args, **kwargs):
...
+            num_warmup_iters=num_warmup_iters,
+            allow_unused_input=allow_unused_input,
+            fp8_weight_caching=fp8_weight_caching,
+            sample_kwargs=sample_kwargs,
+            _order=_order,
+            _num_layers_per_chunk=_num_layers_per_chunk,
+            pool=pool,
+            retain_graph_in_backward=retain_graph_in_backward,
+            io_memory_reduction=io_memory_reduction,
+        )
```
timmoon10: I don't think this is correct. The `fp8_autocast` context has some logic when it starts and finishes (e.g. all-reduce amaxes). This behavior must be captured within the CUDA Graph or else the user must manually reimplement it (breaks existing users and makes this much more difficult to use). 

Can you clarify how this fixes the Mcore bug you mentioned?

> Also fixes a bug that the fp8_autocast context is added inside the module forward function. In MCore, the forward function is not called in cudagraph capturing path, so cudagraph would eventually capture only bf16 ops. Fixed by wrapping the _make_graphed_callables function with a fp8_autocast context.

We replace all the module forwards using the `wrap_autocast` function, so the graph capture should correctly capture FP8 ops.

transformer_engine/pytorch/graph.py
```diff
@@ -110,29 +113,105 @@ def _make_graphed_callables(
...
         num_model_chunks = max(_order)
         num_microbatches = len(_order) // num_model_chunks // 2
         assert num_model_chunks * num_microbatches * 2 == len(_order)
-        assert len(sample_args) * 2 >= len(_order) and (
-            len(sample_args) * 2 % len(_order) == 0
-        ), f"{len(sample_args)} >= {len(_order)} and {len(sample_args)} % {len(_order)} == 0"
-        num_layers = len(sample_args) // num_model_chunks // num_microbatches
-        assert len(callables) == num_model_chunks * num_layers, (
-            f"Callables should have ({num_model_chunks * num_layers}) "
+        if _num_layers_per_chunk is None:
```
timmoon10, line 129:
```suggestion
        
        # Determine number of layers in each model chunk
        if _num_layers_per_chunk is None:
```

transformer_engine/pytorch/graph.py
```diff
@@ -110,29 +113,105 @@ def _make_graphed_callables(
...
+
+        # The starting index of each chunk in callables.
+        _prefix_num_layers = [0]
+        for m_chunk in range(num_model_chunks):
+            num_layers = _num_layers_per_chunk[m_chunk]
+            _prefix_num_layers.append(_prefix_num_layers[-1] + num_layers)
+        assert len(sample_args) == _prefix_num_layers[-1] * num_microbatches, (
+            f"Expected {_prefix_num_layers[-1] * num_microbatches} args tuple, but got"
+            f" {len(sample_args)}."
+        )
```
timmoon10: This check is a bit redundant since we already check that `sample_args` matches `total_num_layers`.

transformer_engine/pytorch/graph.py
```diff
@@ -634,8 +734,10 @@ def make_graphed_callables(
     fp8_group: Optional[dist_group_type] = None,
     fp8_weight_caching: bool = False,
     _order: Optional[List[int]] = None,
+    _num_layers_per_chunk: Optional[List[int]] = None,
     pool: Optional[Tuple[int, ...]] = None,
     retain_graph_in_backward: bool = False,
+    io_memory_reduction: bool = False,
```
timmoon10: This feature is too complicated to expose to regular users.
```suggestion
    _io_memory_reduction: bool = False,
```
It's also a confusing name since I/O makes me think it's related to the filesystem. Something like `_reuse_static_input_surface` is clearer.

transformer_engine/pytorch/graph.py
```diff
@@ -110,29 +113,105 @@ def _make_graphed_callables(
...
+        len_kwargs = len(sample_kwargs[0])
+        assert isinstance(
+            sample_kwargs, list
+        ), "sample_kwargs must be a list for io_memory_reduction."
+        for i in range(len(sample_kwargs)):
+            assert len_kwargs == len(
+                sample_kwargs[i]
+            ), f"Keyword arguments must have same length and shape for io_memory_reduction."
+
+        # Reorganize args and kwargs for input tensor reuse
```
timmoon10: We should expand this comment. The code is not intuitive enough to easily understand.
```suggestion
        # Reorganize args and kwargs for input tensor reuse
        # Note: When capturing a graph, we hold onto the args and kwargs so we have static buffers when the graph is replayed. If two model chunk microbatches have no overlap between their forward and backward, then we can reduce memory usage by reusing the same static buffers.
```
For future reference: We are blindly replacing the args and kwargs in order to reuse the static input surface multiple times. The tensor values will be wrong, but that's fine during graph capture. When the graph is replayed, the args and kwargs will be copied into the static input surface.

transformer_engine/pytorch/graph.py
```diff
@@ -350,6 +435,21 @@ def hook_fn(module, inputs, outputs):  # pylint: disable=unused-argument
...
+                            static_outputs
+                        )
+                        if previous_per_callable_bwd_idx is not None:
+                            per_callable_static_grad_inputs[previous_per_callable_bwd_idx] = (
+                                make_weak_ref(
+                                    per_callable_static_grad_inputs[previous_per_callable_bwd_idx]
+                                )
+                            )
+                        previous_per_callable_bwd_idx = per_callable_bwd_idx
+                        del static_outputs, static_grad_inputs, grad_inputs
```
timmoon10: We should expand this comment. The code is not intuitive enough to easily understand.

For future reference: Once a model chunk microbatch has finished its backward, we assume its output tensor is no longer needed. We can deallocate and let PyTorch's memory allocator reuse it elsewhere.

timmoon10: Is this grad input logic safe? We are assuming that the grad input is safe to deallocate as soon as the next microbatch backward is finished. This is fine if `_order` looks like:
```
[1, 2, 3, 4, 1, 2, 3, 4, -4, -3, -2, -1, -4, -3, -2, -1]
```
But what if it looks like:
```
[1, 1, 2, 2, 3, 3, 4, 4, -4, -4, -3, -3, -2, -2, -1, -1]
```
or
```
[1, 2, 1, 2, 3, 4, 3, 4, -4, -3, -4, -3, -2, -1, -2, -1]
```

transformer_engine/pytorch/graph.py
```diff
@@ -314,17 +398,18 @@ def hook_fn(module, inputs, outputs):  # pylint: disable=unused-argument
...
                     static_outputs = per_callable_static_outputs[per_callable_bwd_idx]
                     bwd_graph = bwd_graphs[per_callable_bwd_idx]
                     # For now, assumes all static_outputs require grad
-                    static_grad_outputs = tuple(
-                        torch.empty_like(o) if o.requires_grad else None for o in static_outputs
-                    )
+                    if not io_memory_reduction or static_grad_outputs is None:
+                        static_grad_outputs = tuple(
+                            torch.empty_like(o) if o.requires_grad else None for o in static_outputs
+                        )
```
timmoon10: Future reference: grad output is only used within backward, so we can reuse the same static buffers every time.

transformer_engine/pytorch/graph.py
buptzyb: The problem is that in mcore cudagraph capturing, we won't go into the TransformerLayer forward function. It hits [`self._cuda_graph_capture`](https://github.com/NVIDIA/Megatron-LM/blob/core_r0.12.0/megatron/core/transformer/transformer_layer.py#L781) and then call [self._forward_attention](https://github.com/NVIDIA/Megatron-LM/blob/core_r0.12.0/megatron/core/transformer/transformer_layer.py#L655) and [self._forward_mlp](https://github.com/NVIDIA/Megatron-LM/blob/core_r0.12.0/megatron/core/transformer/transformer_layer.py#L661) on demand. So, by wrapping the forward function, we actually do nothing about fp8, and cudagraph just captures bf16 kernels. I also tried to follow your logic to wrap `__call__()` instead of `forward()`, but it didn't work, it's still bf16 kernels in cudagraph, I didn't look into why.

After I changed to this method to wrap the whole `_make_graphed_callables`, it works correctly and the loss curve is fully aligned with non-cudagraph. The issue you are concerned about doesn't happen because TransformerBlock always makes an fp8 wrapper before [calling in TransformerLayer](https://github.com/NVIDIA/Megatron-LM/blob/core_r0.12.0/megatron/core/transformer/transformer_block.py#L529). Therefore, you don't necessarily need another wrapper inside TransformerLayer.

transformer_engine/pytorch/graph.py
buptzyb: With this argument we can reuse `static_input`, `static_output`, `static_grad_input`, and `static_grad_output`. So only `_reuse_static_input_surface` doesn't seem to be precise. I'm not good at naming a thing... Maybe `_reuse_graph_input_output_buffers` or something?

transformer_engine/pytorch/utils.py
buptzyb: Agree. All the weird writing styles you saw in `utils.py` are because I just borrowed them from the TRTLLM code https://github.com/NVIDIA/TensorRT-LLM/blob/v0.20.0/tensorrt_llm/_utils.py#L842-L921

Let me refactor my code.

transformer_engine/pytorch/utils.py
buptzyb: I'm not sure whether implementing `convert_to_torch_tensor` as a member function is a good idea or not. It works on top of the `_WeakRefTensor`. How about moving it inside `make_weak_ref()` as a nested function?

transformer_engine/pytorch/graph.py
buptzyb: Thinking of your question, I find that the grad input is safe to deallocate as soon as we get it, not necessarily waiting for the next microbatch backward to finish. So the code can be simpler, like
```
  if io_memory_reduction:
      # Weak ref the static outputs and static grad inputs.
      per_callable_static_outputs[per_callable_bwd_idx] = make_weak_ref(static_outputs)
      per_callable_static_grad_inputs[per_callable_bwd_idx] = make_weak_ref(static_grad_inputs)
  else:
      per_callable_static_grad_inputs[per_callable_bwd_idx] = static_grad_inputs
```
So generally, the `static_outputs` and `static_grad_inputs` both have done their duty and can be freed once we finish its backward. All we need to do is to hold their weak ref for cudagraph replay.

timmoon10: Don't we need `static_grad_inputs` in the previous layer's backward? If we deallocate quickly, there's a risk the previous layer might overwrite the data before it consumes the grad input. This is worse if we don't do the previous layer backward immediately, e.g. if we interleave forwards and backwards.

transformer_engine/pytorch/graph.py
timmoon10: Non-Mcore users will experience this as a breaking change. Currently `make_graphed_callables` creates CUDA Graphs that include `fp8_autocast` behavior (e.g. the amax reduction), but this PR creates CUDA graphs without `fp8_autocast`.

In general, TE tries to match PyTorch's design as much as possible to make it easier for users to transition from plain PyTorch code. Mcore is doing something non-standard with `__call__`, so we don't want that to affect TE's API and make it non-standard.

timmoon10: How about doing something like this to use the custom `__call__` impl within the graph capture:
```python
def wrap_autocast(block: torch.nn.Module) -> Callable:
    old_call = block.__call__

    def __call__(*args, **kwargs):
        with fp8_autocast(
            enabled=fp8_enabled,
            calibrating=fp8_calibrating,
            fp8_recipe=fp8_recipe,
            fp8_group=fp8_group,
            _graph=True,
        ):
            return old_call(*args, **kwargs)

    block.__call__ = __call__
    return old_call

# Wrap callable with fp8_autocast
old_call_funcs = []
for module in modules:
    assert isinstance(module, torch.nn.Module), f"Graphing for {type(module)} is not supported."
    old_call_funcs.append(wrap_autocast(module))

# Capture CUDA Graph
graphed_callables = _make_graphed_callables(
    modules,
    ...
)

# Restore original callable
for module, call_func in zip(modules, old_call_funcs):
    module.__call__ = call_func
```

buptzyb: My concern is that when using in mcore, this leads to a double-autocast, one from TransformerBlock calling TransformerLayer, and one captured in cudagraph. I'm not sure whether this will lead to a precision issue or performance regression. I can have a try.

transformer_engine/pytorch/graph.py
buptzyb: I think I'm wrong in the above statement... We still need to keep the previous layer's `static_grad_inputs`, and this is sufficient. We don't need to keep more. In your example

```
[1, 1, 2, 2, 3, 3, 4, 4, -4, -4, -3, -3, -2, -2, -1, -1]
```

I think you want to show that the first `-3` should take the `static_grad_inputs` of the first `-4` as `grad_outputs`, so that we cannot deallocate it right after the second `-4`, right? But what actually happen is, the `static_grad_inputs` of the first `-4` will be transferred to the next VPP rank for computation, and this transmission finishes no later than the next microbatch in local rank (i.e., the second `-4`) finishes even if we have PP overlapping enabled, that's why we should keep the value for one microbatch long. The same reason, what the first `-3` take as input is also the `static_grad_inputs` transferred from the previous VPP rank, not from the first `-4` on local rank.

Does this make sense to you?

transformer_engine/pytorch/graph.py
buptzyb: Done. Wrapping the `__call__` method works well on my side. Note that we should wrap the class method instead of the instance method: `type(block).__call__`.

transformer_engine/pytorch/graph.py
buptzyb: Added more comments to explain how the `static_outputs` and `static_grad_inputs` are reused.

transformer_engine/pytorch/graph.py
buptzyb: done

transformer_engine/pytorch/graph.py
buptzyb: removed

transformer_engine/pytorch/graph.py
buptzyb: Added comments to explain this.

transformer_engine/pytorch/graph.py
```diff
@@ -664,6 +783,10 @@ def make_graphed_callables(
           this graph may share memory with the indicated pool.
     retain_graph_in_backward: bool, default = `False`
                               Whether to set retain_graph=True in backward graph capture.
+    _reuse_graph_input_output_buffers: bool, default = `False`
+                         Whether or not to reduce memory usage by optimizing input/output
+                         data buffer between graphs. Only available when all callables in
+                         `modules` have the same input/output dtype and shape.
```
timmoon10:
```suggestion
    _reuse_graph_input_output_buffers: bool, default = `False`
        Reduce memory usage by reusing input/output data buffers between 
        graphs. Only supported with Mcore interleaved pipeline parallelism, i.e.
        when `_order` is provided. All callables in `modules` are assumed to have
        inputs and outputs with the same dtype and shape.
```

transformer_engine/pytorch/graph.py
```diff
@@ -758,6 +890,10 @@ def forward_func(*args, **kwargs):
     else:
         torch.cuda.set_rng_state(original_rng_states)
 
+    # Restore FP8 wrapper.
```
timmoon10:
```suggestion
    # Remove FP8 wrapper.
```

transformer_engine/pytorch/graph.py
```diff
@@ -436,19 +446,26 @@ def hook_fn(module, inputs, outputs):  # pylint: disable=unused-argument
...
+                    # reuse them elsewhere.
+                    if _reuse_graph_input_output_buffers:
+                        # Weak ref the static outputs of the forward pass of this backward. It's
+                        # no longer needed after the corresponding backward graph is built up.
                         per_callable_static_outputs[per_callable_bwd_idx] = make_weak_ref(
                             static_outputs
                         )
+                        # Weak ref the static grad inputs of the previous backward pass. The static
+                        # grad inputs should be kept alive for one more backward pass because its
+                        # value is still needed as the input of its next layer backward.
```
timmoon10: Based on the discussion at https://github.com/NVIDIA/TransformerEngine/pull/1234#discussion_r2170504859:
```suggestion
                        # Weak ref the static grad inputs of the previous backward pass. 
                        # Note: After a backward pass, we assume Mcore will send the
                        # grad input to another pipeline parallel rank and that the 
                        # communication is finished before the end of the next backward
                        # pass.
```

timmoon10 (dismissed):
Overall looks good. I have some tweaks to the comments, but I'm happy merging if CI passes.

transformer_engine/pytorch/graph.py
timmoon10: Makes sense, I've updated the comment: https://github.com/NVIDIA/TransformerEngine/pull/1234#discussion_r2172636224

transformer_engine/pytorch/graph.py
```diff
@@ -271,6 +361,34 @@ def hook_fn(module, inputs, outputs):  # pylint: disable=unused-argument
                         only_inputs=True,
                         allow_unused=allow_unused_input,
                     )
+
+                    # Filter module params that get None grad from grad_inputs and remove them
```
buptzyb, line 365: @timmoon10 Hope you don't mind I add another fix to this PR... These lines remove module params that get None grad from `static_input_surface`. This is usually unnecessary in normal running. But if the user adds some backward hooks like `register_hook()` or `register_post_accumulate_grad_hook()`, these hooks will be triggered unexpectedly in model backward.

One case is that in MCore MoE training, we only capture the attention part. If we further enable recomputation to the moe part, it will cause the moe params backward grad-reduce hook to be triggered twice - one by cudagraph, one by recomputation. Therefore it fails this [assertion](https://github.com/NVIDIA/Megatron-LM/blob/core_r0.12.0/megatron/core/distributed/distributed_data_parallel.py#L409).

timmoon10 (approved):
I'm happy with this PR, but @ksivaman would like to check that it doesn't break some existing workflows.

buptzyb:
Hi @ksivaman , could you help check?

PR #970, opened 2024-06-26
[PyTorch] Runtime lookup for CUDA Driver API calls in Userbuffers

denera:
# Description

PR #901 made Userbuffers a permanent part of the TE/PyTorch extension, which necessitated linking to `libcuda.so` at compile time to support CUDA multicast functionality (e.g. calls to `cuMemXYZ` and `cuMulticastXYZ`). This seems to have broken minor version compatibility with older drivers, causing missing symbol errors when loading the extension library.

This PR avoids linking to `libcuda.so` at compile time by recovering the memory addresses for the necessary CUDA Driver functions from the CUDA Runtime API via [`cudaGetDriverEntryPoint()`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DRIVER__ENTRY__POINT.html#group__CUDART__DRIVER__ENTRY__POINT) instead.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Removed compile-time linking to `libcuda.so` for the TE/PyTorch extension.
- Implemented a templated `cudaCallDriverEntryPoint()` tool for recovering, caching and executing CUDA driver functions using symbol name and argument types.
- Fixed missing `initialize_ub` and `destroy_ub` imports at the top of TE/PyTorch.
- Userbuffer communicator structure is now initialized with `new` instead of `malloc` to avoid a segfault on uninitialized `std::function` callbacks.

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #138, opened 2023-04-07
Support for NVRTC kernels

timmoon10:
This PR adds basic support for run-time compilation of kernels with the goals of reducing binary size, reducing compilation time, and enabling more configurable kernel optimizations.

### Changes:
- NVRTC infrastructure (influenced by [MXNet](https://github.com/apache/mxnet/blob/master/src/common/cuda/rtc.cc) and [nvFuser](https://github.com/pytorch/pytorch/blob/master/torch/csrc/jit/codegen/fuser/cuda/fused_kernel.cpp))
- NVRTC kernel for `transpose` using similar perf optimizations as https://github.com/NVIDIA/TransformerEngine/pull/101
- Convenience functions for string manipulation, accessing environment variables, CUDA API calls, and more

### Tasks:
- [x] Implementation
- [x] Evaluate performance
- [x] Documentation
- [x] Unit tests for NVRTC infrastructure
- [x] Unit tests for convenience functions

### Future work:
- NVRTC kernels for other transpose operations
- NVRTC kernels for layer norm
- Refactor error checking infrastructure to use convenience functions

PR #1240, opened 2024-10-10
Do not link against CUDA driver when building

timmoon10:
# Description

It is bad practice to link against the CUDA driver library (`libcuda.so.1` on Linux) at build time because it may differ from the library at run time. In particular, the CUDA Toolkit provides stubs in case CUDA code is being compiled on a system without GPUs. TE already has infrastructure to access with CUDA driver library indirectly via [`cudaGetDriverEntryPoint`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DRIVER__ENTRY__POINT.html#group__CUDART__DRIVER__ENTRY__POINT_1gcf55d143722ccfa9252758181701c876) (see https://github.com/NVIDIA/TransformerEngine/pull/970 and https://github.com/NVIDIA/TransformerEngine/pull/138). However, we should avoid explicitly linking against the CUDA driver in CMake to reduce the risk of linker snafus, e.g. if a future developer accidentally adds a directly call to a CUDA driver function.

## Type of change

- [x] Infra/Build change
- [ ] Code refractor

## Changes

- Do not link against CUDA driver when building

# Checklist:

### DIFF
--- a/transformer_engine/common/CMakeLists.txt
+++ b/transformer_engine/common/CMakeLists.txt
@@ -88,7 +88,6 @@ target_include_directories(transformer_engine PUBLIC
 # Configure dependencies
 target_link_libraries(transformer_engine PUBLIC
                       CUDA::cublas
-                      CUDA::cuda_driver
                       CUDA::cudart)
 target_include_directories(transformer_engine PRIVATE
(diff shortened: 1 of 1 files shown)


ksivaman (approved):
LGTM

PR #1246, opened 2024-10-12
[Bugfix] Fix bias for 0-dim tensors in gemm

yaox12:
# Description

For `te_gemm` and `te_grouped_gemm`,
- If `grad == false`, we should do nothing to `bias`.
- If `grad == true`, we should calculate `grad_bias` here.

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

# Checklist:


transformer_engine/pytorch/csrc/extensions/gemm.cu
```diff
@@ -17,7 +17,8 @@ void te_gemm(at::Tensor A, at::Tensor A_scale_inverse, transformer_engine::DType
   using namespace transformer_engine;
   if (A.data_ptr() == nullptr || B.data_ptr() == nullptr) {
     if (D.data_ptr() != nullptr && !accumulate) D.zero_();
-    if (bias.data_ptr() != nullptr) bias.zero_();
+    // torch.sum is able to handle 0-dim tensors
+    if (bias.data_ptr() != nullptr && grad) bias.copy_(B.sum(0));
```
timmoon10, line 21: We should handle the case where `B` has no data:
```suggestion
    if (bias.data_ptr() != nullptr && grad) {
      if (B.data_ptr() == nullptr) {
        bias.zero_();
      } else {
        bias.copy_(B.sum(0))
      }
    }
```
It seems we are checking `data_ptr() == nullptr` in order to check for empty tensors. As an alternative, it may be better to make our intention explicit:
```c++
  if (A.numel() == 0 || B.numel() == 0) {
    if (D.numel() != 0 && !accumulate) D.zero_();
    // torch.sum is able to handle 0-dim tensors
    if (bias.numel() != 0 && grad) bias.copy_(B.sum(0));
    if (pre_gelu_out.numel() != 0) pre_gelu_out.zero_();
    return;
  }
```

yaox12: Updated.

ksivaman (approved):
LGTM, thanks!

yaox12:
I'll merge this PR since the CI passes except for a doc test failing with irrelevant error.

PR #1250, opened 2024-10-14
[PyTorch] Replace views with reshapes and update PyTorch autocast API

### LINKED ISSUE #1247: FSDP with FP8 is not working
Image

layernorm needs to have reshape instead of view.
In pytorch base module, torch.get_autocast_gpu_dtype() is deprecated.

eljandoubi:
# Description

Update torch.get_autocast_gpu_dtype to torch.get_autocast_dtype("cuda") for all the pytorch.
Migrate from torch.view to torch.reshape. Works with non-contiguous tensors, can copy data. It is handful for distributed training.

@timmoon10 @ksivaman @ptrendx @cyanguwa

Fixes #1247 

## Type of change

- [x] Code refractor


transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -234,9 +234,9 @@ def forward(
         if isinstance(tensor, Float8Tensor):
             return Float8Tensor.make_like(
                 tensor,
-                data=tensor._data.view(*shape),
+                data=tensor._data.reshape(*shape),
             )
-        return tensor.view(*shape)
+        return tensor.reshape(*shape)
```
timmoon10, line 239: This logic is for `Float8Tensor.view`, so using `reshape` could cause correctness problems:
```suggestion
                data=tensor._data.view(*shape),
            )
        return tensor.view(*shape)
```

transformer_engine/pytorch/csrc/extensions/attention.cu
```diff
@@ -39,7 +39,7 @@ __global__ void __launch_bounds__(block_size)
 // fast zero-fills of tensors
 void mha_fill(const at::Tensor &self, const at::Tensor &start_index) {
   auto max_tokens = self.size(0);
-  auto self_2d = self.view({max_tokens, -1});
+  auto self_2d = self.reshape({max_tokens, -1});
```
timmoon10, line 42: `mha_fill_kernel` outputs into `self_2d`, so reshaping can cause correctness problems:
```suggestion
  auto self_2d = self.view({max_tokens, -1});
```

tests/paddle/test_operators.py
```diff
@@ -114,7 +114,7 @@ def convert_float_to_uint16(float_list):
     new_output = []
     for x in np.nditer(float_list):
         new_output.append(np.uint16(copy_bits_from_float_to_uint16(x)))
-    new_output = np.reshape(new_output, float_list.shape).view(np.uint16)
+    new_output = np.reshape(new_output, float_list.shape).reshape(np.uint16)
```
timmoon10, line 117: [`np.ndarray.reshape`](https://numpy.org/doc/2.0/reference/generated/numpy.ndarray.reshape.html) does not support dtype conversion:
```suggestion
    new_output = np.reshape(new_output, float_list.shape).view(np.uint16)
```

timmoon10 (commented):
Naively running `sed 's/view/reshape/g'` results in several correctness issues.

Overall I agree that `reshape` is preferable to `view`.

transformer_engine/pytorch/tensor/float8_tensor.py
eljandoubi: I see. I have made the changes for Float8Tensor.

transformer_engine/pytorch/attention.py
timmoon10, line 1: I think it's too dangerous to blindly replace `view` in this file. Context parallelism has a lot of multi-stream logic and asynchronous communication, so the memory copy in `reshape` can result in race conditions.

We should revert the changes in this file, and only make minimal changes that you can verify are safe.

timmoon10 (changes requested):

timmoon10 (approved):
Overall LGTM. `reshape` is generally preferable to `view`, although many of these changes are redundant since the tensors are already contiguous.

Please [sign your commits](https://github.com/NVIDIA/TransformerEngine/blob/a488b8b127da6743465ee928450f7683fc6d52b5/CONTRIBUTING.rst#sign-your-work) to pass the DCO check, and we'll merge if there are no concerning test failures.

timmoon10 (changes requested):
The changes to `transformer_engine/pytorch/attention.py` have snuck back in. It can be reverted with:
```bash
git checkout main -- transformer_engine/pytorch/attention.py
git commit -m "Revert changes to transformer_engine/pytorch/attention.py" --signoff
```

The commit history is mangled and there are still some unsigned commits. At this point it's better to squash your commits:
```bash
# Merge main branch
git checkout main
git pull origin main
git checkout fix_layernorm_fsdp
git merge main

# Squash all changes into a single commit
git reset main
git commit -a -m 'Use reshape instead of view and update PyTorch autocast API' --signoff

# Force-push to your GitHub
git push eljandoubi fix_layernorm_fsdp --force
```
Just in case, I've made a copy of your branch: https://github.com/timmoon10/TransformerEngine/tree/eljandoubi/fix_layernorm_fsdp

timmoon10:
I've manually squashed the commit (original branch is at https://github.com/timmoon10/TransformerEngine/tree/eljandoubi/fix_layernorm_fsdp-20241018), but bugs have crept back in (https://github.com/NVIDIA/TransformerEngine/pull/1250#discussion_r1802184372, https://github.com/NVIDIA/TransformerEngine/pull/1250#discussion_r1803888790, https://github.com/NVIDIA/TransformerEngine/pull/1250#discussion_r1803889235).

At the moment I think this PR is riskier and more of a hassle than it is worth it. If this is important for your use-case, I suggest breaking this up into two more manageable PRs:
- Update the PyTorch autocast API. This should be a fairly easy and safe change.
- Thoughtfully replace `view` with `reshape`. The current approach (blindly replace all `view`s and changing back to `reshape` when necessary) was dangerous, mostly redundant, and exposed a large surface area for merge conflicts. It would be much better to do the opposite approach and only replace `reshape`s that you know are safe and effective.

PR #1251, opened 2024-10-14
[pyTorch] Infrastructure for C++ QuantizedTensor

ptrendx:
# Description

The goal of this PR is to bring the QuantizedTensor (e.g. Float8Tensor) closer to the C++ level in order to minimize overheads, while keeping the same functionality and Pythonic nature of it. It also includes changes to the GEMM call (once closer to completeness most probably going to be taken out as separate PR) to test the functionality.

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/cpp_extensions/gemm.py
+++ b/transformer_engine/pytorch/cpp_extensions/gemm.py
@@ -9,7 +9,6 @@ import torch
 import transformer_engine_torch as tex
 from ..constants import TE_DType
 from ..utils import assert_dim_for_fp8_exec
-from ..tensor import Float8Tensor
 
 
 __all__ = [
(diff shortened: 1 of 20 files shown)


transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -25,6 +27,143 @@ def _empty_tensor() -> torch.Tensor:
...
+    gelu: bool = False,
+    accumulate: bool = False,
+    out: Optional[torch.Tensor] = None,
+    bias: Optional[torch.Tensor] = None,
+    use_split_accumulator: bool = False,
+    D_dtype: Optional[tex.DType] = None,
+    ub_algo: tex.UbufOverlapAlgo = None,
+    ub: Union[tex.UbufCommOverlap, tex.UbufP2PCommOverlap] = None,
+    extra_output_tensor: Optional[torch.Tensor] = None,
+) -> torch.Tensor:
```
timmoon10, line 43: While we're reworking this API, perhaps we should call it `matmul` since it's less ambiguous (e.g. with column-major/row-major order). We should also keep the core API simple like [`torch.matmul`](https://pytorch.org/docs/stable/generated/torch.matmul.html) and [`np.matmul`](https://numpy.org/doc/2.0/reference/generated/numpy.matmul.html), and leave our non-standard options as kwargs:
```python
def matmul(
    A: torch.Tensor,  # maybe QuantizedTensor
    B: torch.Tensor,  # maybe QuantizedTensor
    /,
    out: Optional[torch.Tensor] = None,  # maybe QuantizedTensor
    *,
    transa: bool = False,
    transb: bool = False,
    out_dtype: Optional[tex.DType] = None,
    accumulate_out: bool = False,  # alternatively: alpha and beta
    bias: Optional[torch.Tensor] = None,
    activation: Optional[str] = None,  # more general than gelu
    workspace: torch.Tensor,  # consider allocating in C++
    use_split_accumulator: bool = False,  # maybe hide within cublas_options kwarg?
    userbuffers_options: Optional[dict] = None,  # minimize impact of unstable UB API
) -> torch.Tensor:  # maybe QuantizedTensor
```

transformer_engine/pytorch/quantization_params.py
timmoon10, line 1: I think we could clean up the API for the quantization params. As I understand it, we currently have three classes:
- `QuantizationParams`: Holds state needed for quantization (scale, amax). It is used in C++ functions.
- `QMeta`: Builder class for `QuantizedTensor` that holds an FP8 recipe. It mainly constructs a default `QuantizationParams` and calls `QuantizedTensor.quantize`.
- `QuantizationParamsProxy`: Holds `fp8_meta` and converts it to `QuantizationParams`.

I propose unifying these APIs into a builder class:
```python
class Quantizer(abc.ABC):
    @abc.abstractmethod
    def quantize(self, tensor: torch.Tensor) -> QuantizedTensor:
        ...

class Float8Quantizer(Quantizer):
    def __init__(self, scale, amax, fp8_dtype, dtype, roles, ...):
        self.scale = scale
        ...

    def quantize(self, tensor):
        return tex.cast(tensor, self)

    # Maybe a direction for the future
    # def update_(self):
    #     tex.fused_amax_and_scale_update_after_reduction(...)

class FP8MetaQuantizer(Quantizer):
    def __init__(self, fp8_meta, fp8_meta_index, ...):
        ...

    def quantize(self, tensor):
        quantizer = Float8Quantizer(fp8_meta[...].scale, ...)
        return quantizer.quantize(tensor)
```
My thought is for `Quantizer` to replace `fp8_meta` over time. Instead of modules holding a complicated `dict` with random tensors, it can just hold a few `Quantizer`s (for input, params, grad output, etc). Instead of `fp8_autocast` needing to dig through the `fp8_meta`s to update `amax_history` and `scale`, we can encapsulate it in an `update_` function. If we go in this direction, we'll eventually be able to get rid of `FP8MetaQuantizer`.

PR #1256, opened 2024-10-16
[PyTorch] Upgrade to Pylint 3.3.1 in lint test

timmoon10:
# Description

We have experienced failures in the linter since Pylint 2.13.5 doesn't support Python 3.12 (see https://github.com/pylint-dev/astroid/issues/2201, https://github.com/pylint-dev/pylint/issues/8782). This can be fixed by upgrading to Pylint 3.3.1, but that also introduces several new checks. This PR fixes those issues, with no substantial functional changes.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refractor
- [x] Testing

## Changes

- Upgrade to Pylint 3.3.1 in PyTorch lint test
- Fix linter warnings from Pylint 3.3.1

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1257, opened 2024-10-16
Upgrade pylint to 3.3.1

ksivaman:
# Description

Upgrades pylint to 3.3.1

## Type of change

- [ ] Infra/Build change
- [x] Code refractor

## Changes

Upgrades pylint to 3.3.1 and makes all related formatting fixes.

# Checklist:


transformer_engine/pytorch/utils.py
```diff
@@ -12,6 +12,11 @@
 import transformer_engine.pytorch.cpp_extensions as ext
 
 
+def maybe_contiguous(tensor: torch.Tensor) -> torch.Tensor:
+    """Make tensor contiguous if final stride is not 1."""
+    return tensor.contiguous() if tensor.stride(-1) != 1 else tensor
```
timmoon10, line 17: Is `maybe_contiguous` used outside of `attention.py`? It seems weird and non-general, so I'd prefer not exposing it in the utilities if possible.

timmoon10 (changes requested):
This is an alternative to https://github.com/NVIDIA/TransformerEngine/pull/1256.
 
The biggest difference is how we handle `missing-function-docstring`: https://github.com/NVIDIA/TransformerEngine/pull/1257 adds trivial docstrings while https://github.com/NVIDIA/TransformerEngine/pull/1256 manually suppresses the warning. I don't like the trivial docstrings since they actually remove useful documentation (there are tools like [`inspect.getdoc`](https://docs.python.org/3/library/inspect.html#inspect.getdoc) that can get docstrings from the parent class). My approach of manually adding suppressions is very annoying though and could encourage bad habits. I'm kind of leaning toward suppressing `missing-function-docstring` entirely.

I am also opposed to suppressing `possibly-used-before-assignment`. It's showing up in places with obfuscated and confusing code (especially related to UB), so I think it helps enforce discipline.

timmoon10 (approved):

PR #946, opened 2024-06-19
[TE/JAX] Prototype for New XLA Custom Calls with FFI

phu0ngng:
# Description

This PR introduced a prototype for converting TE/JAX custom calls to the new XLA Custom Call with FFI.
These primitives are implemented with the new custom calls:
- `CastTranspose`
- `ActLu`
- `DActLu`

`FFI Datatype` with FP8 types has been added to XLA in this PR (openxla/xla#13856). 
All related tests passed.

 Verified: custom calls are captured in CudaGraph by default.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refractor

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [ ] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #1263, opened 2024-10-16
[JAX] XLA Custom Calls with FFI for FusedAttnFwd, Quantize, Transpose, ActLuFP8, LayerNormForwardFP8FFI, and LayerNormBackwardFFI

huanghua1994:
# Description

This PR introduced the following primitives implemented with the new custom calls:

* `FusedAttnFwd`
* `Quantize`
* `Transpose`
* `ActLuFP8`
* `LayerNormForwardFP8FFI`
* `LayerNormBackwardFFI`

New unit tests are also introduced to cover the FFI implementations of:

* `Quantize`
* `Transpose`
* `CastTranspose` (introduced in #946 but no unit test cover)

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

# Checklist:


transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -335,6 +333,124 @@ void FusedAttnForward(cudaStream_t stream, void **buffers, const char *opaque, s
...
+  const CustomCallFusedAttnDescriptor descriptor = {
+      input_batch,   bias_batch,       q_max_seqlen,
+      kv_max_seqlen, attn_heads,       num_gqa_groups,
+      bias_heads,    head_dim,         max_segments_per_seq,
+      wkspace_size,  scaling_factor,   dropout_probability,
+      bias_type,     mask_type,        qkv_layout,
+      dtype,         wkspace_dtype,    is_training,
+      deterministic, window_size_left, window_size_right};
+
+  FusedAttnForwardImpl(stream, buffers, descriptor);
```
phu0ngng, line 412: Hi, I think we don't need to pack `buffers` and `descriptors` here to call the `FusedAttnForwardImpl`.
Since `FusedAttnForwardImpl` is only used for this `FusedAttnForwardFFI()` as `FusedAttnForward()` will be removed in the future, we can insert the computations into `FusedAttnForwardFFI` directly. 

When we will switch to new APIs with FFI for all custom calls, then all old custom call related functions will be removed, including the current `FusedAttnForward`.

huanghua1994: Revised as suggested

zlsh80826 (commented):
Overall looks good to me on fused attn, but do we have fused_attn FFI unit tests?
Since we only have FFI forward, will it cause issue when mixing FFI forward + legacy custom call backward?

phu0ngng:
>  Since we only have FFI forward, will it cause issue when mixing FFI forward + legacy custom call backward?
This does not cause an issue. 
I confirmed it by testing the ActLuFFI + legacy DActLu before; it works without an issue, and cudaGraph was also captured correctly.

zlsh80826:
Sounds good.
Since the FFI is enabled by default. I believe the fused_attn FFI can be covered by the current test_fused_attn.py, right?

phu0ngng:
> Sounds good. Since the FFI is enabled by default. I believe the fused_attn FFI can be covered by the current test_fused_attn.py, right?

Yes, FFI is enabled by default thus the new custom call is selected at runtime and gets tested.

zlsh80826 (approved):
LGTM

PR #1277, opened 2024-10-21
[PyTorch] Fix autocast deprecation warnings

yaox12:
# Description

Fix `autocast` deprecation warnings.

Starting from PyTorch 2.4, use
```python
torch.get_autocast_dtype("cuda")
torch.amp.autocast("cuda")
```
instead of device-specific APIs.

Closes https://github.com/NVIDIA/TransformerEngine/pull/1167.

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

# Checklist:


transformer_engine/pytorch/utils.py
```diff
@@ -305,3 +305,18 @@ def devices_match(device1: torch.device, device2: torch.device) -> bool:
             index2 = torch.cuda.current_device()
         return index1 == index2
     return device1 == device2
+
+
+TORCH_MAJOR = int(torch.__version__.split(".")[0])
+TORCH_MINOR = int(torch.__version__.split(".")[1])
+if TORCH_MAJOR == 2 and TORCH_MINOR >= 4:
```
ksivaman, line 312: I think it would be cleaner to define a single function and do the torch version check inside it to call either `get_autocast_dtype` or `get_autocast_gpu_dtype` as opposed to doing the version check separately and defining 2 separate functions like this.

transformer_engine/pytorch/distributed.py
```diff
@@ -245,26 +245,59 @@ def in_fp8_activation_recompute_phase() -> bool:
...
 
-def _get_active_autocast_contexts():
-    """
-    Returns new CPU and GPU torch.amp.autocast(..) contexts that match the active autocast state
-    at the time of this function's execution.
-    """
-    autocast_cached = torch.is_autocast_cache_enabled()
+TORCH_MAJOR = int(torch.__version__.split(".")[0])
+TORCH_MINOR = int(torch.__version__.split(".")[1])
+if TORCH_MAJOR == 2 and TORCH_MINOR >= 4:
```
ksivaman, line 250: Same as previous comment. Better to have single function with version check inside.

transformer_engine/pytorch/utils.py
yaox12: Updated.

transformer_engine/pytorch/jit.py
```diff
@@ -33,6 +34,11 @@
         # no "recursive" option in pyTorch 2.0 - it acts as if recursive was True
         no_torch_dynamo = lambda recursive=True: torch._dynamo.disable
 
+if torch.__version__ >= "2.4":
+    gpu_autocast_ctx = partial(torch.amp.autocast, device_type="cuda")
+else:
+    gpu_autocast_ctx = torch.cuda.amp.autocast
```
ksivaman, line 40: This should be moved to `utils.py`

transformer_engine/pytorch/utils.py
```diff
@@ -305,3 +305,12 @@ def devices_match(device1: torch.device, device2: torch.device) -> bool:
             index2 = torch.cuda.current_device()
         return index1 == index2
     return device1 == device2
+
+
+def torch_get_autocast_gpu_dtype() -> torch.dtype:
+    """Get PyTorch autocast GPU dtype."""
+    TORCH_MAJOR = int(torch.__version__.split(".")[0])
+    TORCH_MINOR = int(torch.__version__.split(".")[1])
+    if TORCH_MAJOR == 2 and TORCH_MINOR >= 4:
```
ksivaman, line 314: Could we do a simpler >= "2.4" similar to the one currently in jit.py. This method isn't future proof and will break for 3+

transformer_engine/pytorch/distributed.py
```diff
@@ -252,17 +252,38 @@ def _get_active_autocast_contexts():
...
     autocast_cached = torch.is_autocast_cache_enabled()
 
-    gpu_autocast_enabled = torch.is_autocast_enabled()
-    gpu_autocast_dtype = torch.get_autocast_gpu_dtype()
-    gpu_autocast_ctx = torch.cuda.amp.autocast(
-        gpu_autocast_enabled, gpu_autocast_dtype, autocast_cached
-    )
+    TORCH_MAJOR = int(torch.__version__.split(".")[0])
+    TORCH_MINOR = int(torch.__version__.split(".")[1])
+    if TORCH_MAJOR == 2 and TORCH_MINOR >= 4:
```
ksivaman, line 257: Same comment as above

transformer_engine/pytorch/utils.py
yaox12: I'm good with it.
Furthermore, since we now have `packaging` as a dependency, how about using `packaging.version` for a more robust version comparison?

transformer_engine/pytorch/jit.py
yaox12: Done.

transformer_engine/pytorch/transformer.py
```diff
@@ -430,9 +432,7 @@ def __init__(
         self.drop_path = DropPath(drop_path_rate) if drop_path_rate > 0.0 else None
 
         # Set bias+dropout+add fusion grad_enable execution handler.
-        TORCH_MAJOR = int(torch.__version__.split(".")[0])
-        TORCH_MINOR = int(torch.__version__.split(".")[1])
-        use_nvfuser = TORCH_MAJOR > 1 or (TORCH_MAJOR == 1 and TORCH_MINOR >= 10)
+        use_nvfuser = is_torch_min_version("1.10.0a0") and not is_torch_min_version("2.2.0a0")
```
yaox12, line 435: We stop using nvfuser starting from torch 2.2, so I think this update is needed. Tell me if you have other concerns.

yaox12:
@ksivaman Can you review again?

yaox12:
Does this look good to you? @ksivaman

transformer_engine/pytorch/utils.py
```diff
@@ -414,3 +424,16 @@ def canonicalize_process_group(
     if group is None:
         return torch.distributed.distributed_c10d._get_default_group()
     return group
+
+
+def torch_get_autocast_gpu_dtype() -> torch.dtype:
+    """Get PyTorch autocast GPU dtype."""
+    if is_torch_min_version("2.4.0a0"):
```
ksivaman, line 431: Generally LGTM, via the `a0` suffix everywhere?

yaox12: The PyTorch version in NGC PyTorch container is tagged in a way like `2.7.0a0+7c8ec84dab`. `packaging.version.Version` is smart enough to have 
```python
In [1]: from packaging.version import Version

In [2]: Version("2.7.0a0+7c8ec84dab") >= Version("2.7a0")
Out[2]: True

In [3]: Version("2.7.0") >= Version("2.7a0")
Out[3]: True
```

timmoon10 (dismissed):
Overall LGTM

transformer_engine/pytorch/utils.py
yaox12: Just realize that there's already a cached PyTorch version check in TE https://github.com/NVIDIA/TransformerEngine/blob/4ceb3d4c51d09cc66e080e62c3e45f9113d75683/transformer_engine/pytorch/__init__.py#L25-L28, so I don't intend to introduce another one like `is_torch_min_version()`.

I replaced all `is_torch_min_version()` with `torch_version() >= (x, y, x)` now. And I find that this `PkgVersion(str(torch.__version__)).release` will interpret  `2.7.0a0+7c8ec84dab` as `(2, 7, 0)`, so `a0` is not needed anymore.

ksivaman (approved):
LGTM

transformer_engine/pytorch/utils.py
timmoon10: Nice, I prefer using tuples for version comparisons. See https://github.com/NVIDIA/TransformerEngine/pull/1722#discussion_r2065049457.

PR #1307, opened 2024-11-02
[JAX] Collective GEMM custom op with `nvte_cublas_gemm` (no comm. overlap)

denera:
# Description

Implements both old-style and new FFI-based XLA custom calls in C++, and the corresponding JAX primitive including custom partitioning rules.

Custom partitioning rules for a `LHS:([B,] M, K) x RHS:([B,] K, N) = OUT:([B,] M, N)` batched mat-mul operation where `[B]` is the batch dimension:
- Preserve the partitioning of the `[B]` dimension for all operands.
- Always all-gather LHS along the `M` dimension.
- Error out if RHS is partitioned in both `K` and `N` dimensions.
- Force the `K` dimension of LHS to match the partitioning of the `K` dimension of RHS.
- If `K` dimension is partitioned but `M` dimension is not, `jax.lax.psum` (all-reduce) the output over the TP mesh resource.
- If both the `M` and `K` dimensions are partitioned, `jax.lax.psum_scatter` (reduce-scatter) the output over the TP mesh resource.

In practice, the RHS matrix (typically the weight tensor) should be allocated with transposed contracting dimensions `([B,] N, K)` for optimal GEMM heuristics in cuBlasLt. This layout is also mandatory for FP8 inputs.

This PR does **NOT** update fused ops or Flax/Praxis modules to use the new GEMM custom op over the existing XLA pattern matching approach.

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- [x] Added XLA custom calls for `nvte_cublas_gemm`.
- [x] Added JAX primitive for the new XLA custom call.
- [x] Added new serial unit test.
- [ ] Add distributed unit test.

# Checklist:

### DIFF
--- a/tests/jax/test_custom_call_compute.py
+++ b/tests/jax/test_custom_call_compute.py
@@ -25,7 +25,6 @@ from transformer_engine.jax.cpp_extensions.transpose import (
     _jax_dbias_cast_transpose,
 )
 from transformer_engine.jax.cpp_extensions.quantization import _jax_cast_fp8
-from transformer_engine.jax.gemm import fp8_gemm, gemm
 from transformer_engine.jax import cpp_extensions as tex
 
 
@@ -414,55 +413,6 @@ class TestFP8Dot:
                 jnp.asarray(ref_b1_grad, np.float32),
                 dtype=FP8Helper.BWD_DTYPE,
             )
-
-
-class TestGemm:
-
-    @staticmethod
-    def _generate_inputs(b, m, n, k, dtype):
-        key = jax.random.PRNGKey(0)
-        subkeys = jax.random.split(key, 3)
-        a = jax.random.normal(subkeys[0], (b, m, k), dtype)
-        b = jax.random.normal(subkeys[1], (n, k), dtype)
-        bias_dtype = dtype if dtype not in [jnp.float8_e4m3fn, jnp.float8_e5m2] else jnp.bfloat16
-        bias = jax.random.normal(subkeys[2], (n,), bias_dtype)
-        return a, b, bias
-
-    @staticmethod
-    def _generate_fp8_inputs(b, m, n, k, fp8_dtype):
-        a, b, bias = TestGemm._generate_inputs(b, m, n, k, jnp.bfloat16)
-        a_scale, b_scale = map(lambda x: (jnp.max(jnp.abs(x)) / 127.0).astype(jnp.float32), [a, b])
-        a_q, b_q = map(
-            lambda x, x_scale: jnp.round(x / x_scale).astype(fp8_dtype),
-            [(a, a_scale), (b, b_scale)],
-        )
-        return a, a_q, jnp.reciprocal(a_scale), b, b_q, jnp.reciprocal(b_scale), bias
-
-    @pytest.mark.parametrize("m,n,k", GEMM_CASES)
-    @pytest.mark.parametrize("use_bias", (False, True))
-    @pytest.mark.parametrize("do_gelu", (False, True))
-    def test_gemm(self, b, m, n, k, use_bias, do_gelu):
-        a, b, bias = self._generate_inputs(b, m, n, k, jnp.bfloat16)
-
-        primitive_out = gemm(a, b, bias=bias if use_bias else None, layout="NT", do_gelu=do_gelu)
-        ref_out = jnp.dot(a, b)
-        if use_bias:
-            ref_out += bias
-        if do_gelu:
-            ref_out = jax.nn.gelu(ref_out)
-
-        assert_allclose(primitive_out, ref_out, dtype=jnp.bfloat16)
-
-    @pytest.mark.skipif(not is_fp8_supported, reason=reason)
-    @pytest.mark.parametrize("m,n,k", GEMM_CASES)
-    @pytest.mark.parametrize("fp8_dtype", FP8_COMPUTE_TYPE)
-    def test_fp8_gemm(self, m, n, k, fp8_dtype):
-        a, a_q, a_scale_inv, b, b_q, b_scale_inv, _ = self._generate_fp8_inputs(m, n, k, fp8_dtype)
-
-        primitive_out = fp8_gemm(a_q, a_scale_inv, b_q, b_scale_inv, out_dtype=jnp.bfloat16)
-        ref_out = jnp.dot(a, b)
-
-        assert_allclose(primitive_out, ref_out, dtype=fp8_dtype)
 
 
 @pytest.fixture(name="random_inputs")
--- a/tests/jax/test_custom_call_compute.py
+++ b/tests/jax/test_custom_call_compute.py
@@ -425,19 +425,16 @@ class TestGemm:
         a = jax.random.normal(subkeys[0], (b, m, k), dtype)
         b = jax.random.normal(subkeys[1], (n, k), dtype)
         bias_dtype = dtype if dtype not in [jnp.float8_e4m3fn, jnp.float8_e5m2] else jnp.bfloat16
-        bias = jax.random.normal(subkeys[2], (n, ), bias_dtype)
+        bias = jax.random.normal(subkeys[2], (n,), bias_dtype)
         return a, b, bias
 
     @staticmethod
     def _generate_fp8_inputs(b, m, n, k, fp8_dtype):
         a, b, bias = TestGemm._generate_inputs(b, m, n, k, jnp.bfloat16)
-        a_scale, b_scale = map(
-            lambda x: (jnp.max(jnp.abs(x)) / 127.).astype(jnp.float32),
-            [a, b]
-        )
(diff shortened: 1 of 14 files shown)


nouiz:
```Error out if RHS is partitioned in both K and N dimensions.
```
Why? Normal JAX behavior is to do some gathering.

huanghua1994:
It seems that currently the batch size is not handled in the C++ code. Since JAX is using row-major storage for tensor by default, probably the batch dimension should be combined with the `m` dimension for LHS or the `n` dimension for RHS?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -0,0 +1,691 @@
...
+            is_fp8 = True
+        if jax_dtype_is_fp8(rhs_dtype):
+            assert (
+                rhs_scale_inv_aval.size == 1
+                and dtypes.canonicalize_dtype(rhs_scale_inv_aval.dtype) == jnp.float32
+            ), "Missing RHS operand scale inverse in FP8 GEMM."
+
+        # Validate operand layouts
+        lhs_inner_dim, rhs_inner_dim = map(
+            lambda inner_dim, ndims: (ndims - inner_dim) if inner_dim < 0 else inner_dim,
```
abhinavgoel95, line 102: @denera  should be `ndims + inner_dim` when inner_dim is negative, right?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -0,0 +1,691 @@
...
+        check_valid_batch_dims(batch_dims)
+        lhs_bdims, *_, bias_bdims, gelu_input_bdims, out_amax_bdims, out_scale_bdims = batch_dims
+
+        # FP8 GEMM only supports non-transposed LHS and transposed RHS
+        lhs, _, rhs, *_ = batched_args
+        lhs_trans = contracting_dims[0] != lhs.ndim - 1
+        rhs_trans = contracting_dims[1] == rhs.ndim - 1
+        lhs = jnp.matrix_transpose(lhs) if lhs_trans and jax_dtype_is_fp8(lhs.dtype) else lhs
+        rhs = jnp.matrix_transpose(rhs) if not rhs_trans and jax_dtype_is_fp8(rhs.dtype) else rhs
+        contracting_dims = (1, 1)
```
abhinavgoel95, line 385: @denera is there a need to hard-code this?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -0,0 +1,691 @@
...
+            rhs,
+            *batched_args[3:],
+            out_dtype=out_dtype,
+            contracting_dims=contracting_dims,
+            fuse_gelu=fuse_gelu,
+            fuse_bias=fuse_bias,
+            grad=grad,
+            accumulate=accumulate,
+            use_split_accumulator=use_split_accumulator,
+        )(lhs_bdims, out_amax_bdims, out_scale_bdims, gelu_input_bdims, bias_bdims)
```
abhinavgoel95, line 399: This gives me an error.
Line: https://github.com/NVIDIA/TransformerEngine/pull/1307/files#diff-f5b74ca3c5a70acb3d764e9b8adea40b8bab554fe4d2362f3052b7b932c0464dR187-R194 returns a tuple.

```
TypeError: 'list' object is not callable
```

cc @denera

abhinavgoel95 (commented):
@denera I have some questions about the PR.

transformer_engine/jax/cpp_extensions/gemm.py
denera: My understanding is that non-transposed LHS and transposed RHS is a cuBlasLt requirement for FP8 GEMM, so this is ensuring that the custom op can do the GEMM even if it receives incorrect layouts. I was planning to add a warning here as well to make it clear that there will be a perf hit.

transformer_engine/jax/fp8.py
```diff
@@ -110,6 +113,8 @@ def __init__(
         self._scale_list[FP8MetaPackage.WEIGHT_IDX] = weight_scale
         self._amax_list[FP8MetaPackage.GRAD_IDX] = grad_amax
         self._scale_list[FP8MetaPackage.GRAD_IDX] = grad_scale
+        self._amax_list[FP8MetaPackage.OUTPUT_IDX] = output_amax
+        self._scale_list[FP8MetaPackage.OUTPUT_IDX] = output_scale
```
phu0ngng, line 117: Hi, 
For the delayed scaling FP8 recipe, the output `amax` and `scale` are not used anywhere afterwards, so we don't need to output and store them.

denera: The FP8 GEMM+RS overlap needs output amax/scale when the communication buffer type is FP8 -- i.e. the overlap algorithms/kernels communicate FP8 GEMM output and fuse BF16 upcast into the sum-reduce.

This PR does not implement TP overlap, but PR #1337 extends the same operations to support TP overlap, so I'm including the output amax/scale infrastructure here.

tests/jax/test_custom_call_compute.py
```diff
@@ -415,6 +416,55 @@ def ref_func(x, ln_s, y, z, w, v, amax_list_1, amax_list_2, scale_list_1, scale_
...
+    def _generate_fp8_inputs(b, m, n, k, fp8_dtype):
+        a, b, bias = TestGemm._generate_inputs(b, m, n, k, jnp.bfloat16)
+        a_scale, b_scale = map(lambda x: (jnp.max(jnp.abs(x)) / 127.0).astype(jnp.float32), [a, b])
+        a_q, b_q = map(
+            lambda x, x_scale: jnp.round(x / x_scale).astype(fp8_dtype),
+            [(a, a_scale), (b, b_scale)],
+        )
+        return a, a_q, jnp.reciprocal(a_scale), b, b_q, jnp.reciprocal(b_scale), bias
+
+    @pytest.mark.parametrize("m,n,k", GEMM_CASES)
```
huanghua1994, line 441: Need to provide a list for test parameter `b` (batch size)

tests/jax/test_custom_call_compute.py
```diff
@@ -415,6 +416,55 @@ def ref_func(x, ln_s, y, z, w, v, amax_list_1, amax_list_2, scale_list_1, scale_
...
+    @pytest.mark.parametrize("use_bias", (False, True))
+    @pytest.mark.parametrize("do_gelu", (False, True))
+    def test_gemm(self, b, m, n, k, use_bias, do_gelu):
+        a, b, bias = self._generate_inputs(b, m, n, k, jnp.bfloat16)
+
+        primitive_out = gemm(a, b, bias=bias if use_bias else None, layout="NT", do_gelu=do_gelu)
```
huanghua1994, line 447: Do we really need to provide or use `layout` parameter here? On one hand, user or other functions in TE is unlikely to use this argument (I think C/C++ code would need it but not python code), on the other hand does it make dist-mem sharding complicated?

tests/jax/test_custom_call_compute.py
denera: Actually this PR is not supposed to modify `test_custom_call_compute.py`. These changes are erroneous and need to be removed. Thank you for catching it!

tests/jax/test_custom_call_compute.py
denera: Changes to this file are erroneous and I just pushed up a commit to restore the original.

All testing for the new collective GEMM custom op are written in `test_distributed_gemm.py` instead.

tests/jax/test_distributed_gemm.py
```diff
@@ -0,0 +1,302 @@
...
+    fsdp = False
+    mesh_shape = dict(tp=NUM_DEVICES)
+    resources = dict(cp_resource="tp", tp_resource="tp")
+    if parallel_dist in ["DP_TP", "FSDP_TP"]:
+        batched = True
+        mesh_shape.update(dict(tp=NUM_DEVICES // 2, dp=NUM_DEVICES // 2))
+        resources.update(dict(dp_resource="dp"))
+        if parallel_dist == "FSDP_TP":
+            fsdp = True
+            mesh_shape.update(dict(tp=NUM_DEVICES // 2, dp=1, zp=NUM_DEVICES // 2))
```
huanghua1994, line 58: This mesh shape calculation is incorrect. Suggested revision:
```python
    if parallel_dist in ["DP_TP", "FSDP_TP"]:
        batched = True
        tp = NUM_DEVICES // 2
        dp = NUM_DEVICES // tp
        mesh_shape.update(dict(tp=tp, dp=dp))
        resources.update(dict(dp_resource="dp"))
        if parallel_dist == "FSDP_TP":
            fsdp = True
            dp = 1
            zp = NUM_DEVICES // tp
            mesh_shape.update(dict(tp=tp, dp=1, zp=zp))
            resources.update(dict(fsdp_resource="zp"))
```

denera:
Closing in favor of #1846

PR #1324, opened 2024-11-08
Build with uv instead of just pip

### LINKED ISSUE #1323: TransformerEngine doesn't work with uv
`setup.py` calls [uninstall_te_wheel_packages](https://github.com/NVIDIA/TransformerEngine/blob/main/build_tools/utils.py#L304) which fails because `pip` is not a module. This is expected because I use uv instead of pip. I think TransformerEngine should not assume pip.

Full error:
```
error: Failed to run uv compile   × Failed to download and build `transformer-engine @
  │ git+https://github.com/NVIDIA/TransformerEngine.git`
  ╰─▶ Build backend failed to determine requirements with `build_wheel()`
      (exit status: 1)

      [stderr]
      /home/jennifer/.cache/uv/builds-v0/.tmp22OFxW/bin/python: No module
      named pip
      Traceback (most recent call last):
        File "<string>", line 14, in <module>
        File
      "/home/jennifer/.cache/uv/builds-v0/.tmp22OFxW/lib/python3.12/site-packages/setuptools/build_meta.py",
      line 333, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File
      "/home/jennifer/.cache/uv/builds-v0/.tmp22OFxW/lib/python3.12/site-packages/setuptools/build_meta.py",
      line 303, in _get_build_requires
          self.run_setup()
        File
      "/home/jennifer/.cache/uv/builds-v0/.tmp22OFxW/lib/python3.12/site-packages/setuptools/build_meta.py",
      line 521, in run_setup
          super().run_setup(setup_script=setup_script)
        File
      "/home/jennifer/.cache/uv/builds-v0/.tmp22OFxW/lib/python3.12/site-packages/setuptools/build_meta.py",
      line 319, in run_setup
          exec(code, locals())
        File "<string>", line 148, in <module>
        File
      "/home/jennifer/.cache/uv/git-v0/checkouts/f679917ede501d2d/e5ffaa7/build_tools/utils.py",
      line 305, in uninstall_te_wheel_packages
          subprocess.check_call(
        File
      "/home/jennifer/.rye/py/cpython@3.12.7/lib/python3.12/subprocess.py",
      line 413, in check_call
          raise CalledProcessError(retcode, cmd)
      subprocess.CalledProcessError: Command
      '['/home/jennifer/.cache/uv/builds-v0/.tmp22OFxW/bin/python',
      '-m', 'pip', 'uninstall', '-y', 'transformer_engine_cu12',
      'transformer_engine_torch', 'transformer_engine_paddle',
      'transformer_engine_jax']' returned non-zero exit status 1.

. uv exited with status: exit status: 1
```

jennifgcrl:
# Description

Fixes #1323 

## Type of change

- [x] Infra/Build change
- [ ] Code refractor

## Changes

- build_tools picks uv or python -m pip depending on if pip exists or not
- additional `install_and_import`s for uv's clean build environments

# Checklist:

### DIFF
--- a/build_tools/build_ext.py
+++ b/build_tools/build_ext.py
@@ -25,7 +25,10 @@ from .utils import (
     get_frameworks,
     cuda_path,
     get_max_jobs_for_parallel_build,
+    install_and_import,
 )
+
+install_and_import("pybind11[global]")
 
 
 class CMakeExtension(setuptools.Extension):
--- a/build_tools/build_ext.py
+++ b/build_tools/build_ext.py
@@ -25,10 +25,7 @@ from .utils import (
     get_frameworks,
     cuda_path,
     get_max_jobs_for_parallel_build,
-    install_and_import,
 )
-
-install_and_import("pybind11[global]")
 
 
 class CMakeExtension(setuptools.Extension):
--- a/build_tools/utils.py
+++ b/build_tools/utils.py
@@ -255,9 +255,7 @@ def get_frameworks() -> List[str]:
     _frameworks = [framework.lower() for framework in _frameworks]
     for framework in _frameworks:
         if framework not in supported_frameworks:
-            raise ValueError(
-                f"Transformer Engine does not support framework={framework}"
-            )
+            raise ValueError(f"Transformer Engine does not support framework={framework}")
 
     return _frameworks
 
@@ -297,17 +295,10 @@ def copy_common_headers(
         shutil.copy(path, new_path)
 
 
-def pip_or_uv() -> List[str]:
-    if find_spec("pip") is not None:
-        return [sys.executable, "-m", "pip"]
-    else:
-        return ["/usr/bin/env", "uv", "pip"]
-
-
 def install_and_import(package):
     """Install a package via pip (if not already installed) and import into globals."""
     main_package = package.split("[")[0]
-    subprocess.check_call([*pip_or_uv(), "install", package])
+    subprocess.check_call([sys.executable, "-m", "pip", "install", package])
     globals()[main_package] = importlib.import_module(main_package)
 
 
--- a/build_tools/utils.py
+++ b/build_tools/utils.py
@@ -15,6 +15,7 @@ import sys
 from pathlib import Path
 from subprocess import CalledProcessError
 from typing import List, Optional, Tuple, Union
+from importlib.util import find_spec
 
 
 @functools.lru_cache(maxsize=None)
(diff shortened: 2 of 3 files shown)


build_tools/utils.py
```diff
@@ -294,24 +295,32 @@ def copy_common_headers(
...
+                "-m",
+                "pip",
+                "uninstall",
+                "-y",
+                "transformer_engine_cu12",
+                "transformer_engine_torch",
+                "transformer_engine_paddle",
+                "transformer_engine_jax",
+            ]
+        )
```
timmoon10, line 326: This is a reasonable change that I would accept on its own. It doesn't fully solve the root issue (the package structure is different when installing from PyPI or from source), so I could imagine users running into problems if they install TE multiple times. But it seems fair to expect users to handle this on their own if they are using a non-standard build environment.

timmoon10 (changes requested):
`uv` looks neat, but I'm uncomfortable hard-coding like this until it has wider adoption.

I've written up my thoughts on the root problems at https://github.com/NVIDIA/TransformerEngine/issues/1323#issuecomment-2465575031. I think the best approach for now is to adopt your changes to `uninstall_te_wheel_packages` and to build TE with build isolation disabled. `uv` should support this (https://github.com/astral-sh/uv/issues/1715). As an aside, I agree with @Taytay's assessment that build isolation is not a good fit for the ML ecosystem since it depends on custom, vendor-provided packages that are tricky to install. It's annoying the lengths that the Python developers went to to coerce package maintainers to use build isolation.

build_tools/build_ext.py
```diff
@@ -25,8 +25,11 @@
     get_frameworks,
     cuda_path,
     get_max_jobs_for_parallel_build,
+    install_and_import,
 )
 
+install_and_import("pybind11[global]")
```
ksivaman, line 31: This installation should be guarded depending on which framework the extensions are being built for

jennifgcrl:
@timmoon10 Thanks for looking into this! I've updated the PR to be just the `uninstall_te_wheel_packages` change.

I think it's easier for me to just maintain a fork than `--no-build-isolation` since (as far as I can tell) there's no way to tell rye to pass `--no-build-isolation` to uv.

Do you think we can (for the time being) support both build isolation and no build isolation? One way this might work is: if NVTE_FRAMEWORK contains frameworks that aren't in the current python environment, try installing them; otherwise use the existing package.

PR #1142, opened 2024-08-27
[PyTorch] Userbuffers support in operation-based API

timmoon10:
# Description

This PR adds basic support in the linear operation for using Userbuffers to overlap tensor-parallel communication with GEMMs. This is implemented as fused operations:
```python
model = te.ops.Sequential(
    te.ops.BasicLinear(...),
    te.ops.Bias(...),
    te.ops.ReduceScatter(...),
)  # Fused into UserbuffersForwardLinear
```
I've tried to avoid touching the core UB infrastructure in [`transformer_engine/pytorch/module/base.py`](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/module/base.py), so I've kept the messy API and hackily worked around some bugs. This feature should be considered experimental and unstable.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Add fused operation for linear forward with Userbuffers
- Add fused operation for linear backward with Userbuffers

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #825, opened 2024-04-30
Avoid amax roll for non-run modules

ksivaman:
# Description

The amax history should be rolled only for modules which ran and thus produced a non-zero amax. This bug was introduced in #575.

Fixes #814 

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

PR #820, opened 2024-04-27
[PyTorch] Refactor FP8 workspaces in linear modules

timmoon10:
This PR refactors the logic for FP8 weight workspaces in `te.Linear`, `te.LayerNormLinear`, and `te.LayerNormMLP`. The existing logic is somewhat convoluted since it was designed to pass around raw UINT8 buffers and `Float8Tensor` support was kludged in https://github.com/NVIDIA/TransformerEngine/pull/452. For example, when `te.Linear` has FP32 params, it maintains two workspace `Float8Tensor`s (FP8 weight and transpose) and in each forward pass it will extract out the buffers to create another temporary `Float8Tensor`. This PR streamlines the process so it will just maintain a single workspace `Float8Tensor`.

Motivations:
- This fixes an FP8 recipe bug introduced in https://github.com/NVIDIA/TransformerEngine/pull/575 (see https://github.com/NVIDIA/TransformerEngine/pull/786#pullrequestreview-2024087226). The FP8 scale update kernel updates the scales for weights in every forward pass, even ones where the FP8 weights are not updated, so we can run into situations where the FP8 scales don't match the FP8 weights. This PR fixes this by taking advantage of the fact that `Float8Tensor` has a private copy of the FP8 scale-inverse won't be affected by scale updates until its values are updated.
- FP8 compute can sometimes result in performance degradation due to CPU overheads (see https://github.com/NVIDIA/TransformerEngine/issues/761). The `Float8Tensor` constructor requires a CUDA kernel launch to initialize the FP8 scale-inverse, so creating unnecessary `Float8Tensor`s adds non-trivial CPU overhead. Benchmarking the forward pass of small `te.Linear`s, this PR gives a 1.12x speedup. 
- I find the logic in this PR a bit easier to reason about, although I'd appreciate feedback. It feels nicer to let `Float8Tensor` internally handle things like FP8 casting and interacting with `fp8_meta`.

PR #575, opened 2023-12-22
[PyTorch] cuda graph support

ksivaman:
This PR adds the following features (high-level):

 - `make_graphed_callables` API similar to the [PyTorch API](https://pytorch.org/docs/stable/generated/torch.cuda.make_graphed_callables.html) with some additional arguments for FP8 usage. Support for fp8 weight caching via existing `is_first_microbatch`argument is also retained.
 - Restructuring and amax reduction logic with a simpler design and handling of various parallelisms with minimal book-keeping compared to the previous approach.
 - Forward and backward amaxes are reduced within the scope of current iteration, solving numerous bugs w.r.t. checkpointing and removing the need to save global buffers.
 - Support for nested/multiple FP8 autocast contexts with different recipes and distributed groups.
 - Amax reductions are module independent and happen at at autocast level. This also resolves numerous bugs and allows for support for MoE/LoRA like models.
 - Redesign of transposes for `Float8Tensor` that makes the transposes persistent for graph capture. Also fixes use cases for the vanilla optimizers (non fp8-distopt).
 - The scaling inverses for weight tensors are no longer frozen when caching weights across microbatches.

PR #1326, opened 2024-11-09
[PyTorch] Remove special handling for FP8 params in FP8 recipe infrastructure

timmoon10:
# Description

https://github.com/NVIDIA/TransformerEngine/pull/1142 exposed a very subtle bug that caused non-deterministic test failures in [`test_fusible_ops_with_userbuffers.py`](https://github.com/NVIDIA/TransformerEngine/blob/main/tests/pytorch/distributed/test_fusible_ops_with_userbuffers.py).

<details> <summary> Bug description </summary>

`test_fusible_ops_with_userbuffers.py` runs multiple test cases at a time because launching a parallel job is expensive, so it constructs and destroys multiple TE models with FP8 parameters. [Python IDs may be reused after an object is deallocated](https://docs.python.org/3/library/functions.html#id), so the Python ID for FP8 tensors is sometimes reused. However, `Float8Tensor.post_optimizer_step_fwd_amax_reduction` uses Python IDs to check whether to perform amax reductions and FP8 scale updates. I observed that this was causing FP8 scale updates at weird times, which corrupted UB buffers, which caused hangs.

🫠 

</details>

In short, the problem is from this weird callback in `Float8Tensor`:
https://github.com/NVIDIA/TransformerEngine/blob/2643ba1df43397cc84c9da5fe719a66d87ad9a0a/transformer_engine/pytorch/tensor/float8_tensor.py#L77

This hack was added in https://github.com/NVIDIA/TransformerEngine/pull/575 so that we would properly update FP8 scales for FP8 params after the optimizer step. However, we've made improvements since then:
- https://github.com/NVIDIA/TransformerEngine/pull/825: There is no harm if we perform FP8 scale updates more often than necessary.
- https://github.com/NVIDIA/TransformerEngine/pull/820: FP8 weights keep a private copy of its FP8 scale-inv, so we can perform FP8 scale updates whenever we want.

Thus, there's no need to do an FP8 scale update for the weights immediately after the optimizer step. We just need to do it sometime before the next optimizer step and there should be no change in numerics. In fact, these FP8 scales are already participating in the forward pass amax reduction and scale update, so avoiding those operations reduces runtime overheads. Also, this just makes `Float8Tensor` more sane and less tightly coupled with the FP8 recipe infrastructure.

## Type of change

- [ ] Infra/Build change
- [x] Code refractor

## Changes

- Remove special handling for FP8 params from FP8 recipe infrastructure

# Checklist:


timmoon10:
The convergence tests in pipeline 20334396 timed out, but all the tests that did run passed.

ksivaman (approved):

PR #1341, opened 2024-11-18
[PyTorch] Bugfix for wgrad bulk overlap conflict when dgrad overlap is reduce-scatter

denera:
# Description

When Userbuffers config dictionary sets overlap method to `ring-exchange` or `pipeline` for any `*_dgrad` layer, that layer's `*_wgrad` overlap needs to be disabled in order for `ub_overlap_rs_dgrad=True` option for related TE modules to function correctly.

This PR fixes a bug where the "*_wgrad" overlap was persisting in the Userbuffer configuration and the corresponding UB object was being initialized even when it was not needed.

+ @fanshiqing for viz.

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- `*_wgrad` overlap is now removed from `methods["bulk"]` list when the same layer's `*_dgrad` overlap has its method set to either `ring-exchange` or `pipeline`.
- `add_ub(name, **ub_cfg)` is now only called if `name` is in the original user-provided `ub_cfg`. This avoids creating UB objects with default configs that may conflict with the user's intended TP overlap use.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/base.py
+++ b/transformer_engine/pytorch/module/base.py
@@ -391,7 +391,7 @@ def initialize_ub(
 
     # Loop over user configs and disable dgrad and wgrad bulk overlaps for every layer that has a
     # reduce-scatter dgrad overlap.
-    ub_cfg = dict() if ub_cfg is None else ub_cfg
+    ub_cfg = {} if ub_cfg is None else ub_cfg
     for name in dgrad_reduce_scatter_overlap:
         if name in ub_cfgs:
             final_cfg = get_default_config(name)
--- a/transformer_engine/pytorch/module/base.py
+++ b/transformer_engine/pytorch/module/base.py
@@ -418,9 +418,8 @@ def initialize_ub(
         if name in ub_cfgs:
             final_cfg = get_default_config(name)
             final_cfg.update(ub_cfgs[name])
-            final_cfg["fp8_buf"] = (
-                (name in layers_all_gather_overlap)
-                or ub_cfgs[name].get("fp8_buf", False)
+            final_cfg["fp8_buf"] = (name in layers_all_gather_overlap) or ub_cfgs[name].get(
+                "fp8_buf", False
             )
             add_ub(name, **final_cfg)
 
--- a/transformer_engine/pytorch/module/base.py
+++ b/transformer_engine/pytorch/module/base.py
@@ -404,6 +404,7 @@ def initialize_ub(
                 layers_reduce_scatter_overlap.remove(wgrad_name)
                 layers_all_gather_overlap.remove(name)
                 layers_reduce_scatter_overlap.append(name)
+                final_cfg["is_reduce_scatter"] = True
 
                 methods["bulk"].remove(name)
                 methods["bulk"].remove(wgrad_name)
--- a/transformer_engine/pytorch/module/base.py
+++ b/transformer_engine/pytorch/module/base.py
@@ -391,7 +391,7 @@ def initialize_ub(
 
     # Loop over user configs and disable dgrad and wgrad bulk overlaps for every layer that has a
     # reduce-scatter dgrad overlap.
-    ub_cfg = {} if ub_cfg is None else ub_cfg
+    ub_cfgs = {} if ub_cfgs is None else ub_cfgs
     for name in dgrad_reduce_scatter_overlap:
         if name in ub_cfgs:
             final_cfg = get_default_config(name)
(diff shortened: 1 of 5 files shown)


timmoon10 (dismissed):
LGTM, pending CI

transformer_engine/pytorch/module/base.py
```diff
@@ -389,27 +389,38 @@ def add_ub(
             )
         _ub_communicators[name] = ub_obj
 
-    if ub_cfgs is not None:
-        for name in dgrad_reduce_scatter_overlap:
-            if name in ub_cfgs and "method" in ub_cfgs[name] and ub_cfgs[name]["method"] != "bulk":
+    # Loop over user configs and disable dgrad and wgrad bulk overlaps for every layer that has a
+    # reduce-scatter dgrad overlap.
+    ub_cfg = {} if ub_cfg is None else ub_cfg
```
fanshiqing, line 394: local variable `ub_cfg` referenced before assignment.

`ub_cfg` --> `ub_cfgs`?

transformer_engine/pytorch/module/base.py
```diff
@@ -389,27 +389,38 @@ def add_ub(
...
                 layers_all_gather_overlap.remove(name)
                 layers_reduce_scatter_overlap.append(name)
+
                 methods["bulk"].remove(name)
-                new_method = ub_cfgs[name]["method"]
+                methods["bulk"].remove(wgrad_name)
+                new_method = final_cfg["method"]
                 methods[new_method].append(name)
 
+            ub_cfg[name] = final_cfg
```
fanshiqing, line 413: `ub_cfg` --> `ub_cfgs`?

transformer_engine/pytorch/module/base.py
```diff
@@ -389,27 +389,38 @@ def add_ub(
...
+
+    # Now initialize the UB objects for each layer
     for name in methods["ring_exchange"] + methods["pipeline"] + methods["bulk"]:
-        ub_cfg = get_default_config(name)
         if ub_cfgs is not None and name in ub_cfgs:
```
fanshiqing, line 417: `if ub_cfgs is not None and name in ub_cfgs` ==> `if name in ub_cfgs`

transformer_engine/pytorch/module/base.py
```diff
@@ -389,27 +389,38 @@ def add_ub(
...
-            fp8_buf = (name in layers_all_gather_overlap) or (
+            final_cfg = get_default_config(name)
+            final_cfg.update(ub_cfgs[name])
+            final_cfg["fp8_buf"] = (name in layers_all_gather_overlap) or (
```
fanshiqing, line 420: if I am using 'bf16' dtype for training, and `name in layers_all_gather_overlap` is true for some pattern (e.g., `fc2_dgrad`), then `final_cfg["fp8_buf"]` will be set to True, which is unexpected.

fanshiqing: Okay [line 361](https://github.com/NVIDIA/TransformerEngine/blob/2ca29de02a3dd31c4caadc76be376f6a33312190/transformer_engine/pytorch/module/base.py#L361) has already using `use_fp8` to guard the effectiveness of `fp8_buf`, but it would be better if we can advance this guard when we set the default value of `cfg['fp8_buf']` to avoid confusion.

denera: We can't advance this guard because it's not only for all-gather overlaps. I'll run through the logic to explain.

`use_fp8` tells the initialization whether the user intends to invoke comm+GEMM overlap under the `te.fp8_autocast()` context (i.e. with FP8 GEMM inputs).

`fp8_buf` tells the initialization whether the communication buffer should be allocated as FP8.

For all-gather overlaps, the buffer has to match the GEMM input type, so it will always be allocated in FP8 when `use_fp8 == True` (i.e. GEMM inputs are FP8) regardless of what `fp8_buf` is set to in the user's layer configuration. In other words, the user does not get a choice here.

For reduce-scatter overlaps, the GEMM output has to match the buffer type, which can be either FP8 or BF16 when GEMM inputs are FP8. In this scenario, setting `fp8_buf = True` means that we communicate FP8 data between devices, and then fuse the BF16 upcast into the sum-reduce.

Advancing this guard to the default config options means that the user is denied the option to set `fp8_buf` for reduce-scatter overlaps, and that RS overlaps always communicate BF16 data, which is not always the optimal choice.

PR #1345, opened 2024-11-21
Fix cuda graph capture for grouped gemm

xrennvidia:
# Description

Cuda graph does not work with Grouped GEMM.
The saved forward activations are corrupted before bwd_graph is replayed. Explicitly setting `retain_graph=True` can hold the activations and fix the issue. 

## Type of change

- [ ] Infra/Build change
- [ ] Code refractor

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/graph.py
+++ b/transformer_engine/pytorch/graph.py
@@ -214,6 +214,7 @@ def _make_graphed_callables(
                     grad_outputs=tuple(torch.empty_like(o) for o in outputs if o.requires_grad),
                     only_inputs=True,
                     allow_unused=allow_unused_input,
+                    retain_graph=True,
                 )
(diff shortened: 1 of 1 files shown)


timmoon10 (commented):
Wouldn't we expect this to increase memory usage?

I see that `torch.cuda.make_graphed_callables` doesn't set `retain_graph=True`:
https://github.com/pytorch/pytorch/blob/c25b201583fc28243b87c460a2f18e2531a676e7/torch/cuda/graphs.py#L326-L336
We want to match plain PyTorch as much as possible unless there is a good reason to introduce divergence. If this is MoE-specific, perhaps we could add a kwarg like `retain_graph_in_backward` that is `False` by default.

timmoon10 (approved):
LGTM

PR #1389, opened 2025-01-02
Better cuBLAS handle management

### LINKED ISSUE #1372: Should cublasLtHandle_t be Destroyed?
In cublas_gemm, it always create a new cublasLtHandle_t, but no corresponding handle destroy. It seems may cause memory or resource leak, any reason?

ptrendx:
# Description

Create only 1 (per thread) cublasLtHandle in order to not leak memory. We do not want to actually destroy the handle, since it would incur an implicit cudaDeviceSynchronize call (see [cuBLAS docs](https://docs.nvidia.com/cuda/cublas/#cublasltdestroy)).

Fixes #1372

## Type of change

- [ ] Infra/Build change
- [ ] Code refactor

## Changes

- Introduced a singleton manager for cuBLAS handles similar to the one used for cuDNN handles.

# Checklist:


ksivaman (approved):
LGTM

shenzhenghai:
May I have a question about the commit? If there is multiple devices invoke cublas_gemm(), it seems always use the same cublasLt handle? 
The document mentioned "one cuBLASLt handle should be created for each device".

shenzhenghai:
> May I have a question about the commit? If there is multiple devices invoke cublas_gemm(), it seems always use the same cublasLt handle? The document mentioned "one cuBLASLt handle should be created for each device".

@ptrendx @ksivaman any comment on my concern? Thanks.

ptrendx:
@shenzhenghai Yeah, if your usage is to have multiple devices per thread then you are correct - it would use the same handle, which is wrong. I don't necessarily like the fact that fixing this will require calling cudaGetDevice before every cuBLAS call - this adds overhead and the most typical usage by the DL Frameworks nowadays is 1 GPU per process. But I agree that the functional correctness is the most important - will submit a fix (also will fix the cuDNN handle management, since it has similar problem).

timmoon10 (approved):
LGTM

PR #1401, opened 2025-01-11
Use log1p(x) instead of log(1+x)

kit1980:
# Description

This function is more accurate than torch.log() for small values of input - https://pytorch.org/docs/stable/generated/torch.log1p.html

Found with TorchFix https://github.com/pytorch-labs/torchfix/

## Type of change

- [ ] Infra/Build change
- [X] Code refactor


xrennvidia:
Hi @kit1980 

Thanks for catching this. `log1p` is more accurate for small values, and it should be same as `log` for big values? I just wonder if this is only a fix for small values, and how this will impact big values? Thanks.

kit1980:
@xrennvidia I think it's the same for big values, only for small values there are numerical instabilities that `log1p` designed to address.

xrennvidia:
> @xrennvidia I think it's the same for big values, only for small values there are numerical instabilities that `log1p` designed to address.

Awesome, thanks a lot. I will approve the PR.

xrennvidia (approved):
LGTM.

PR #1408, opened 2025-01-13
Support `store_param_remainders` feature from Apex in TE Fused Adam

sanandaraj5597:
# Description

When the master parameter is in FP32 and the model parameters are in BF16, we can store the trailing 16 remainder bits and reconstruct the master FP32 param from (BF16 model param + the remainder). 

This helps us half the master parameter memory usage.

### DIFF
--- a/transformer_engine/pytorch/csrc/extensions/multi_tensor/multi_tensor_adam.cu
+++ b/transformer_engine/pytorch/csrc/extensions/multi_tensor/multi_tensor_adam.cu
@@ -181,14 +181,12 @@ struct AdamFunctorMaster {
 
 template <typename PARAM_T, typename GRAD_T, typename FULL_T, typename index_t>
 struct AdamFunctorMasterParamRemainder {
-
   __device__ __forceinline__ void operator()(index_t chunk_size, volatile int *noop_gmem,
                                              TensorListMetadata<5> &tl,  // NOLINT(*)
                                              const float beta1, const float beta2,
                                              const float beta1_correction,
                                              const float beta2_correction, const float epsilon,
                                              const float lr, adamMode_t mode, const float decay) {
-
     index_t tensor_loc = tl.block_to_tensor[blockIdx.x];
 
     index_t chunk_idx = tl.block_to_chunk[blockIdx.x];
@@ -218,8 +216,8 @@ struct AdamFunctorMasterParamRemainder {
         int16_t int16[2];
       };
       fp32_or_int162 local_master_param[ILP];
-      int16_t local_p[ILP]; 
-      int16_t local_p_rem[ILP]; 
+      int16_t local_p[ILP];
+      int16_t local_p_rem[ILP];
       MATH_T r_g[ILP];
       MATH_T r_m[ILP];
       MATH_T r_v[ILP];
@@ -250,7 +248,7 @@ struct AdamFunctorMasterParamRemainder {
         local_master_param[ii].int16[0] = local_p_rem[ii];
       }
 
-      MATH_T* r_p = reinterpret_cast<MATH_T*>(local_master_param);
+      MATH_T *r_p = reinterpret_cast<MATH_T *>(local_master_param);
 
 #pragma unroll
       for (int ii = 0; ii < ILP; ii++) {
@@ -296,7 +294,6 @@ struct AdamFunctorMasterParamRemainder {
     }
   }
 };
-
 
 template <typename PARAM_T, typename GRAD_T, typename FULL_T, typename index_t>
 struct AdamFunctor {
@@ -668,9 +665,9 @@ void multi_tensor_adam_cuda(int chunk_size, at::Tensor noop_flag,
 }
 
 void multi_tensor_adam_param_remainder_cuda(int chunk_size, at::Tensor noop_flag,
-                                            std::vector<std::vector<at::Tensor>> tensor_lists, 
-                                            const float lr, const float beta1, const float beta2, 
-                                            const float epsilon, const int step, const int mode, 
+                                            std::vector<std::vector<at::Tensor>> tensor_lists,
+                                            const float lr, const float beta1, const float beta2,
+                                            const float epsilon, const int step, const int mode,
                                             const int bias_correction, const float weight_decay) {
   using namespace at;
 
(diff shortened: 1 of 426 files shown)


transformer_engine/pytorch/optimizers/fused_adam.py
```diff
@@ -243,13 +256,14 @@ def _apply_scale(self, state_name, unscaled_state, scaled_state, scale):
             unscaled_state.mul_(rscale)
             scaled_state.copy_(unscaled_state)
 
-    def get_unscaled_state(self, param, state_name):
+    def get_unscaled_state(self, param, state_name, store_param_remainders=False):
```
MaciejBalaNV, line 259: The default value of `store_param_remainders` is `False` here, but it's `True` by default in the constructor. I think it's misleading, why not just set it to `True` here as well?

MaciejBalaNV:
I'm getting NaNs when using this feature. You can reproduce it by running `test_fused_optimizer` tests, after setting `store_param_remainders=True` in `_initialize_state` method (otherwise it fails earlier) and by commenting out `torch.testing.assert_close(ref_params, master_params)` check (this is expected to fail, since we now keep `master_params` as `int16`). 

Still, with all these changes, the tests fail at `torch.testing.assert_close(ref_params, model_params_to_fp32, rtol=1e-2, atol=1e-2, equal_nan=True)` with an error message that weights are NaN.

transformer_engine/pytorch/optimizers/fused_adam.py
sanandaraj5597: I don't want to store param remainders for state_name other than master_params, that's why it's defaulted to false.

timmoon10 (changes requested):

transformer_engine/pytorch/csrc/extensions/multi_tensor/multi_tensor_adam.cu
```diff
@@ -548,6 +664,69 @@ void multi_tensor_adam_cuda(int chunk_size, at::Tensor noop_flag,
...
+  const auto p_in_type = tensor_lists[1][0].scalar_type();
+  auto tl_size = tensor_lists.size();
+
+  // case 5:  g, p, m, v, p_master
+  TORCH_CHECK(tl_size == 5, "tensor list must contain 5");
+
+  if (requires_64bit_indexing) {
+    // g, p, m, v, p_master
+    DISPATCH_DOUBLE_FLOAT_HALF_AND_BFLOAT(
+        p_in_type, 0, "adam",
```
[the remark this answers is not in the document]
sanandaraj5597: Fixed.

transformer_engine/pytorch/csrc/extensions/multi_tensor/multi_tensor_adam.cu
```diff
@@ -179,6 +179,122 @@ struct AdamFunctorMaster {
   }
 };
 
+template <typename PARAM_T, typename GRAD_T, typename FULL_T, typename index_t>
```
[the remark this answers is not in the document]
sanandaraj5597: Fixed

transformer_engine/pytorch/optimizers/fused_adam.py
```diff
@@ -118,6 +125,7 @@ def __init__(
         exp_avg_dtype=torch.float32,
         exp_avg_sq_dtype=torch.float32,
         use_decoupled_grad=False,
+        store_param_remainders=True,
```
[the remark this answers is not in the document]
sanandaraj5597: Fixed.

transformer_engine/pytorch/optimizers/fused_adam.py
sanandaraj5597: Worked around it. Resolving conversation.

transformer_engine/pytorch/optimizers/fused_adam.py
```diff
@@ -192,6 +201,10 @@ def __init__(
         }
         self._scales = {}
         self.use_decoupled_grad = use_decoupled_grad
+        # Works only when master params is in FP32
+        self.store_param_remainders = (
+            store_param_remainders and master_weights and master_weight_dtype == torch.float32
+        )
```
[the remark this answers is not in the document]
sanandaraj5597: There are few reasons I don't want to do this:
1) If there are FP32 params in one `param_group` and BF16 params in another `param_group`, I think it is more robust to have the BF16 param group to use this feature. What you suggested would make it weak, I think, impose a constraint that this optimization can be used only if all `param_groups` have BF16 params.
2) If we want to support both FP16 and BF16 in the same param group, I check for FP16 parameter presence inside the step function and ensure this memory saving technique is not used. Check line 497 in `fused_adam.py`
3) I've run tests with FP16 params and this feature enabled and I don't see any failures. 
4) You don't need to have exp_avg* to be in FP32 to use this opt. I'm trying to support this feature even if they are something else. Again, I tested this in the unit tests and it worked fine numerically.

The code is smart enough to pick this optimization only if param type is BF16 and master param type is FP32. Any combination should work for my best knowledge.

sanandaraj5597:
@MaciejBalaNV I added a failure guard with capturable mode. We don't have a plan to use CUDA graphs with optimizers, so not worth having this support.

transformer_engine/pytorch/csrc/extensions/multi_tensor/multi_tensor_adam.cu
```diff
@@ -681,51 +678,25 @@ void multi_tensor_adam_param_remainder_cuda(int chunk_size, at::Tensor noop_flag
...
   auto tl_size = tensor_lists.size();
 
   // case 5:  g, p, m, v, p_master
   TORCH_CHECK(tl_size == 5, "tensor list must contain 5");
+  TORCH_CHECK(p_in_type == at::ScalarType::BFloat16,
+              "Adam with BF16 param remainders requires BF16 params");
+
+  // g, p, m, v, p_master
+  DISPATCH_DOUBLE_FLOAT_HALF_AND_BFLOAT(
+      p_in_type, 0, "adam",
```
timmoon10, line 692: We've already confirmed that `p_in_type` is BF16, so dispatching for FP16 and FP32 is unnecessary. See https://github.com/NVIDIA/TransformerEngine/pull/1408#discussion_r1924430661.
```suggestion
```

transformer_engine/pytorch/optimizers/fused_adam.py
timmoon10: It's better, although we still have the problem that state scaling and BF16 remainders are both using this function in different ways. It's troubling that `get_unscaled_state` might not get the unscaled state.

timmoon10 (approved):
Making this feature opt-in makes me feel a lot less worried. Users should know what they are doing before enabling this optimization.

I'll change this PR to merge into `release_v2.0` and start a CI pipeline.

transformer_engine/pytorch/csrc/extensions/multi_tensor/multi_tensor_adam.cu
sanandaraj5597: Agreed, but it shouldn't break the code I guess.

transformer_engine/pytorch/optimizers/fused_adam.py
sanandaraj5597: I agree your point. But using this function both with/without feature makes it look very efficient. Writing a separate function needs new function usage step function, checkpointing, etc.

I've also tried to add assert checks inside the function to tighten the understanding/correctness. Hope you are fine with it.

transformer_engine/pytorch/optimizers/fused_adam.py
```diff
@@ -511,6 +569,12 @@ def step(self, closure=None, grad_scaler=None):
                         "FusedAdam does not support a mix of float16 and bfloat16 model weights."
                     )
 
+                    if self.store_param_remainders:
+                        raise RuntimeError(
+                            "FusedAdam doesn't support a mix of FP16/BF16 weights + Store param"
+                            " remainder."
+                        )
+
```
timmoon10, line 577: Pylint is complaining that this isn't reachable:
```suggestion
```

timmoon10:
https://github.com/NVIDIA/TransformerEngine/pull/1443 is identical to this PR, but rebased on [the `release_v2.0` branch](https://github.com/NVIDIA/TransformerEngine/tree/release_v2.0).

PR #1435, opened 2025-01-29
[PyTorch] Reduce tensor dimensions in MXFP8 tests

timmoon10:
# Description

We increased the tensor dimensions in some unit tests due to an MXFP8 requirement that tensor dims are multiples of 128, but this also increased numerical errors and caused test failures. https://github.com/NVIDIA/TransformerEngine/pull/1431 and https://github.com/NVIDIA/TransformerEngine/pull/1437 relaxed that constraint to being a multiple of 32, so this PR reduces the unit test tensor dimensions.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring
- [x] Testing

## Changes

-  Revert tensor dimensions in MXFP8 tests

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #1437, opened 2025-01-29
 [common] Generalized MXFP8 fused kernels w.r.t. input tensor dimensions

Oleg-Goncharov:
# Description

Lifted restrictions on input tensor dimensions in MXFP8 fused quantize/dequantize kernels, allowing arbitrary number of rows (or the product of all tensor dimensions except the last of high-dimensional tensors).
The number of columns (or the dimensionality of the last dimension of high-dimensional tensors) must satisfy the memory alignment requirement for Tensor Memory Accelerator (TMA) to be a multiple of 16 bytes.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Removed the `isFullTile` requirement in the `mxfp8_quantize` and `mxfp8_dequantize` functions
- Added 16-byte alignment check when creating a TMA descriptor 
- MXFP8 Test Suite (quantize and dequantize): Added the alignment/padding requirement for dimensions of the tensors with scaling factors to be multiples of [128,4] for row-wise and [4,128] for column-wise scaling
- Modified `exp2f_rcp` function so it returns `1.0` if the `biased_exp==0` to avoid `NaN` in the `amax=0` scenario

# Checklist:


tests/cpp/operator/test_cast_mxfp8.cu
```diff
@@ -397,10 +430,16 @@ void performTest_x2(const ProcessingMethod processing_method,
 }
 
 std::vector<std::pair<size_t, size_t>> matrix_sizes = {
+    {1, 16},
+    {3, 32},
+    {16, 48},
+    {21, 64},
+    {37, 96},
```
timmoon10, line 437: Do we need all these cases? It seems excessive to dedicate 60% of the tests to non-performant edge cases. I think it would be fine to go with `{1, 16}`, `{16, 48}`, `{993, 512}`.

transformer_engine/common/common.cu
```diff
@@ -97,7 +97,15 @@ void create_2D_tensor_map(CUtensorMap &tensorMap, const SimpleTensor &tensor,
...
-  NVTE_CHECK(isPointerAligned(dataPtr, 16), "Tensor data must be 16B aligned");
+
+  constexpr int TMA_gmem_alignment = 16;  // Alignment of the global memory address
+  NVTE_CHECK(isPointerAligned(dataPtr, TMA_gmem_alignment),
+             "Tensor data pointer must be 16B aligned");
+
+  const int TMA_needed_size = TMA_gmem_alignment / type_size;
+  NVTE_CHECK(globalX % TMA_needed_size == 0, "Shape not supported. Expected multiple of " +
+                                                 std::to_string(TMA_needed_size) + ", got " +
+                                                 std::to_string(globalX));
```
timmoon10, line 108: `NVTE_CHECK` already converts to string internally. It can even print out `std::vector<T>`.
```suggestion
  const int TMA_needed_size = TMA_gmem_alignment / type_size;
  NVTE_CHECK(globalX % TMA_needed_size == 0, "Shape not supported. Expected multiple of " +
                                                 TMA_needed_size + ", got " +
                                                 globalX);
```

timmoon10 (approved):
Mostly LGTM. It looks like this is needed before we can merge https://github.com/NVIDIA/TransformerEngine/pull/1435.

tests/cpp/operator/test_cast_mxfp8.cu
Oleg-Goncharov: Agree, we can remove them. I just tested different combinations of edge cases which overlap a bit

timmoon10: I think we went too far in the other direction. The existing test cases (`{128, 128}`, `{256, 256}`, `{768, 1024}`) were important to test since they're closer to the standard use-cases.

transformer_engine/common/common.cu
timmoon10: Ah, it should be something like:
```suggestion
  const int TMA_needed_size = TMA_gmem_alignment / type_size;
  NVTE_CHECK(globalX % TMA_needed_size == 0, "Shape not supported. Expected multiple of ",
                                             TMA_needed_size, ", got ", globalX);
```
This is not high priority, but taking advantage of these convenience features makes it easier to write detailed error messages.

ptrendx (approved):

timmoon10:
Pipeline 23264045

timmoon10:
I see test failures when combined with https://github.com/NVIDIA/TransformerEngine/pull/1435: https://github.com/NVIDIA/TransformerEngine/pull/1435#issuecomment-2626453407
It seems to pass tests when run in isolation: pipeline 23274414

timmoon10 (approved):
LGTM. I still see errors when this is combined with #1435 (pipeline 23302072), but that seems related to MXFP8 gated activations, LayerNorm, RMSNorm. The MXFP8 quantize and dequantize kernels seem to be doing the right thing.

PR #1468, opened 2025-02-09
[MoE][PyTorch] Add prob permutation to mask-based MoE permutation; Fix FP8 related codes

hxbai:
# Description

Add probs permutation codes to the mask-based permutation. With this, we can apply the probs to the MoE expert MLP rather than to the unpermutation to avoid saving huge input tensor of unpermute function.

Fix FP8 Tensor usages in the permutation codes since TE 2.0 has some breaking changes on FP8 Tensor interfaces.

Depiction for probs application:
image

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add probs permutation to the functions of `moe_permute` and `moe_sort_chunks_by_index`
- Fix FP8 related codes

# Checklist:


tests/pytorch/test_permutation.py
```diff
@@ -873,6 +894,209 @@ def backward_wrapper(
         print(f"chunk sort\t\tbwd: pytorch: {t1:.3f} ms,  TE: {t2:.3f} ms")
 
 
+def _test_permutation_mask_map_alongside_probs(
```
phu0ngng, line 897: Hi, 

I found this test almost identical to the `_test_permutation_mask_map` except that we have TP thus involves calling `te_sort_chunks_by_index`.

Is it correct that the  `_test_permutation_mask_map` is the special case of `_test_permutation_mask_map_alongside_probs`, in which the `tp_size=1`?

If yes, I suggest we combine these two test and eliminate code duplications.

transformer_engine/pytorch/triton/permutation.py
```diff
@@ -235,18 +263,35 @@ def _unpermute_kernel(
...
+                    inp = inp.to(data_type, bitcast=True)
+                inp = inp.to(compute_type)
+                if WITH_MERGING_PROBS:
+                    merging_prob_off = (
+                        pid * stride_merging_probs_token + expert_idx * stride_merging_probs_expert
+                    )
+                    merging_prob = tl.load(merging_probs_ptr + merging_prob_off).to(compute_type)
+                    inp *= merging_prob
                 accumulator += inp
+            if PERMUTE_PROBS:
```
phu0ngng, line 275: Hi,

I wonder in which case we can have `PERMUTE_PROBS != WITH_MERGING_PROBS`.
I interpret the code as when we have `WITH_MERGING_PROBS=True`, we do the accumulation then reset the prob to zeros in the `if PERMUTE_PROBS` block. Then in which case we don't need to reset them?

transformer_engine/pytorch/permutation.py
```diff
@@ -468,6 +495,7 @@ def moe_permute(
     num_out_tokens: int = -1,
     max_token_num: int = -1,
     map_type: str = "mask",
+    probs: torch.Tensor = None,
 ) -> Tuple[torch.Tensor, torch.Tensor]:
```
timmoon10, line 499: We should make sure the type hints are accurate:
```suggestion
) -> Union[Tuple[torch.Tensor, torch.Tensor],Tuple[torch.Tensor, torch.Tensor, torch.Tensor]]:
```
Messy type hints are a clue that the design is messy. I prefer to avoid variadic returns like this since users might need to have extra logic to unpack the outputs. Better to always return a tuple with 3 elements, where `permuted_probs` is `None` if not needed:
```python
) -> Tuple[torch.Tensor, torch.Tensor, Optional[torch.Tensor]]:
```

transformer_engine/pytorch/permutation.py
```diff
@@ -494,18 +526,29 @@ def moe_permute(
...
+        if probs is None:
+            return output, row_id_map
+        return output, row_id_map, permuted_probs
     raise ValueError("map_type should be one of 'mask' or 'index'")
 
 
 def moe_unpermute(
     inp: torch.Tensor,
     row_id_map: torch.Tensor,
-    probs: torch.Tensor = None,
```
timmoon10, line 508: Are we worried about breaking backward compatibility? If we wanted to be fully careful, we could keep the `probs` kwarg and add some extra logic:
```python
if probs is not None:
    if merging_probs is not None:
        raise ValueError("Both merging_probs and probs kwarg are provided. probs is deprecated.")
    warnings.warn("probs kwarg is deprecated. Use merging_probs kwarg instead.")
    merging_probs = probs
```
If we are sure that this code is only used in Mcore and we accept that we will not support MoE on older Mcore versions, we can skip this.

transformer_engine/pytorch/permutation.py
```diff
@@ -563,57 +609,76 @@ def forward(
...
+            probs_grad = None
+        return act_grad, None, None, probs_grad
 
 
 def moe_sort_chunks_by_index(
     inp: torch.Tensor,
     split_sizes: torch.Tensor,
     sorted_index: torch.Tensor,
+    probs: torch.Tensor = None,
 ) -> Tuple[torch.Tensor, torch.Tensor]:
```
timmoon10, line 682: This type hint is also inaccurate.

tests/pytorch/test_permutation.py
```diff
@@ -509,19 +509,28 @@ def _test_permutation_mask_map(
...
+        permute_fwd_input = _permute_fwd_input_quantizer(permute_fwd_input)
+        permute_bwd_input = _permute_bwd_input_quantizer(permute_bwd_input)
+        unpermute_bwd_input = _unpermute_bwd_input_quantizer(unpermute_bwd_input)
 
-        pytorch_permute_fwd_input = permute_fwd_input.from_float8(torch.float16)
-        pytorch_permute_bwd_input = permute_bwd_input.from_float8(torch.float16)
-        pytorch_unpermute_bwd_input = unpermute_bwd_input.from_float8(torch.float16)
+        pytorch_permute_fwd_input = permute_fwd_input.dequantize().to(torch.float16)
+        pytorch_permute_bwd_input = permute_bwd_input.dequantize().to(torch.float16)
+        pytorch_unpermute_bwd_input = unpermute_bwd_input.dequantize().to(torch.float16)
```
timmoon10, line 533: We can cast directly into FP16:
```suggestion
        pytorch_permute_fwd_input = permute_fwd_input.dequantize(dtype=torch.float16)
        pytorch_permute_bwd_input = permute_bwd_input.dequantize(dtype=torch.float16)
        pytorch_unpermute_bwd_input = unpermute_bwd_input.dequantize(dtype=torch.float16)
```
We can apply this change in several other places in this file.

timmoon10 (commented):
Mostly looks reasonable, but we should make sure the user-facing APIs are not too messy.

transformer_engine/pytorch/triton/permutation.py
hxbai: Hi, Phuong,

Yes, it is easily confused by this part. I just added a graph to this PR description to depict the usage.

The previous codes are for the right workflow and this PR is to support the left workflow. For the right workflow, we don't permute the probs tensor and pass it directly to the unpermute operation. For the left workflow, we permute the probs tensor and apply it on the GroupedGEMM; for this case, no probs is passed to the unpermute operation.

- Left: moe_permute(probs=probs), moe_unpermute(merging_probs=None)
- Right: moe_permute(probs=None), moe_unpermute(merging_probs=probs)

For this kernel, it is used by both permute_bwd and unpermute_fwd:

- permute_bwd in the left workflow: `PERMUTE_PROBS=True` and `WITH_MERGING_PROBS=False`
- permute_bwd in the right workflow: `PERMUTE_PROBS=False` and `WITH_MERGING_PROBS=False`
- unpermute_fwd in the left workflow: `PERMUTE_PROBS=False` and `WITH_MERGING_PROBS=False`
- unpermute_fwd in the left workflow: `PERMUTE_PROBS=False` and `WITH_MERGING_PROBS=True`

So, these two args would not be True at the same time.

transformer_engine/pytorch/permutation.py
hxbai: Yes, I think we do not need to change this argument name for compatibility. I just want to distinguish the different meanings of the probs tensor in permute and unpermute function. I'll change it back.

tests/pytorch/test_permutation.py
hxbai: Sure.

transformer_engine/pytorch/permutation.py
hxbai: Hi, Tim,

I'm also not sure how to handle this part. If we modify the return numbers, we may break some compatibility. The current MCore assumes that this function returns two tensors.

What about adding two new functions `moe_permute_with_probs` and `moe_sort_chunks_by_index_with_probs` and leaving the previous function unchanged? Do you think it is a good idea?

tests/pytorch/test_permutation.py
hxbai: Hi, it is not exactly same.

For `_test_permutation_mask_map`, the probs are applied to the unpermutation and the results of permute and unpermute are verified separately.

For `_test_permutation_mask_map_alongside_probs`, it is roughly an end-to-end test. The probs are applied to the unpermutation for the PyTorch version and applied to the permute output for the TE version. We want to make sure that the two methods can have same final output value.

phu0ngng: I see. Thank you!

transformer_engine/pytorch/permutation.py
timmoon10: If the name is better, it would be better to change it while this feature is new and Mcore is the only major user. Adding the deprecation logic is not too burdensome.

transformer_engine/pytorch/permutation.py
hxbai: I added new functions `moe_permute_with_probs` and `moe_sort_chunks_by_index_with_probs` and kept the original ones unchanged.

transformer_engine/pytorch/permutation.py
hxbai: OK. I modified the name and added the deprecation warning.

transformer_engine/pytorch/permutation.py
hxbai: Added new function `moe_sort_chunks_by_index_with_probs`.

timmoon10 (approved):
LGTM. Test failures in pipeline 24020411 are unrelated.

PR #901, opened 2024-06-10
[C/PyTorch] Removed MPI dependence in Userbuffers

denera:
# Description

This PR splits off the userbuffers MPI-dependence removal from PR #760.

With these changes, userbuffers is now bootstrapped via callbacks to `torch.distributed` collectives. In the absence of the MPI dependence, userbuffers is always compiled as part of the PyTorch extension and no longer requires the `NVTE_WITH_USERBUFFERS=1` flag.

The old MPI-based bootstrapping can be re-activated via `UB_MPI_BOOTSTRAP=1` at compile time.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Changes

- TE build no longer supports the `NVTE_WITH_USERBUFFERS=1` option and Userbuffers is now always compiled into the PyTorch extensions module.
- MPI collectives in Userbuffers bootstrapping are replaced with callbacks to `torch.distributed` collectives.
- New `UB_MPI_BOOTSTRAP=1` option in the TE build activates the old Userbuffers bootstrapping via MPI collectives.
- `transformer_engine.pytorch.module.base.initialize_ub(...)` is now more conveniently accessible via `transformer_engine.pytorch.initialize_ub(...)`.
- Userbuffer communicators can now be cleaned up via `transformer_engine.pytorch.destroy_ub(...)`.
- `transformer_engine.pytorch.initialize_ub(...)` now requires the tensor-parallel process group instead of just the tensor-parallel size.
- Added comm+GEMM overlap example with `te.LayerNormMLP`.

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1470, opened 2025-02-10
Add support for UB MNNVL

nvcastet:
# Description

- Register UB buffers with Multi-Node NVLink when possible.

- Remove logic for the inter-node communicator setup which was never used in TE, therefore being dead code.
This helps with removing the assumption that a NVLINK domain is a single node and that NVLINK domain sizes need to be symmetric in a job. AKA with DP4-TP16, the cluster orchestrator can allocate 4x 16-GPU NVLINK domains but also 1x 32-GPU NVLINK domain + 2x 160GPU NVLINK domains for example.
This removes the hostname exchange which is not valid for kubernetes env running one container per process/gpu.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Support for TP across multi-node NVLINK.

### DIFF
--- a/transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers-host.cpp
+++ b/transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers-host.cpp
@@ -9,6 +9,7 @@
 #include <cuda_runtime_api.h>
 #include <inttypes.h>
 #include <math.h>
+#include <nvml.h>
 #include <sched.h>
 #include <stdio.h>
 #include <string.h>
@@ -29,7 +30,6 @@
 #ifdef NVTE_UB_WITH_MPI
 static MPI_Comm EXT_COMM_WORLD = MPI_COMM_WORLD;
 static MPI_Comm EXT_COMM_INTRA;
-static MPI_Comm EXT_COMM_INTER;
 
 #define UB_MPI_CHECK(expr)                                                                   \
   do {                                                                                       \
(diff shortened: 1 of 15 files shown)


nvcastet:
CC @ptrendx

transformer_engine/common/comm_gemm_overlap/userbuffers/userbuffers-host.cpp
```diff
@@ -182,29 +215,14 @@ int create_communicator_grouped2(communicator **comm, int myrank, int numranks,
...
 
-  (*comm)->num2_nodes = tensornodes;
-  (*comm)->my2_node = (mynode / datanodes) % tensornodes;
-  (*comm)->first2_node = mynode - (*comm)->my2_node * datanodes;
-
-  (*comm)->fifo = reinterpret_cast<ub_request *>(malloc(sizeof(ub_request) * NVTE_MAX_REQUESTS));
-  (*comm)->nblocks = 8;
-  (*comm)->alignblock = 1024 * 512;
-  (*comm)->minblock = 1024 * 2 * 1024;
-  (*comm)->asyncblocks = 16;
```
timmoon10, line 203: If we are not setting these class members, we should remove them from the header file as well.

timmoon10 (commented):
Overall looks reasonable, although a more detailed PR description would be helpful for understanding your intentions and for future `git blame`s. If I understand correctly, this PR accomplishes two things:
1. Register UB buffers with Multi-Node NVLink when possible. This seems straightforward, although we should also handle the case where the TP group is larger than a node and MNNVL is not supported (e.g. by throwing an exception).
2. Remove logic for the inter-node communicator. It seems you just delete code without any logic changes, so was the inter-node communicator just dead code? If so, my guess is that it was made redundant in https://github.com/NVIDIA/TransformerEngine/pull/901.

timmoon10 (approved):
LGTM, pending CI

PR #829, opened 2024-05-01
Generation tutorial for Gemma model

pggPL:
# Description

I added the tutorials with finetuning and with generation for the Gemma model. Moreover I added few features that 
were neccessary to make my tutorials work.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Changes

- Two new notebooks in the docs: one with finetuning for Gemma - analogous to the tutorial with Llama, one with generation for the Gemma,
- Generalized the kernel for rotary positional encoding to allow the sequences to start with different encoding positions,
- Added the kernel to effectively save key and value to the kv_cache,
- Expanded the class `InferenceParams` - which is responsible for caching k and v,
- Changed DotProductAttention to run THD attention when there are ragged tensors in kv_cache.

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

# Future work:

`TransformerLayer` does not support thd and it is a problem. The solutions right now works that way:
- one need to call `setup_before_new_input` before forward to indicate the sequence lengths,
- then one passes forward with TransformerLayer with `self_attn_format='thd'` and padded sequences with shape `bshd`,
- all layers get input is [b,s,*] format, not in [t, *] (including attention)
- `InferenceParams.retrieve_from_kv_cache` retrieves key_layer in bshd or ths format depending of `inference_params.qkv_format`,

As can be seen, it is quite messy workaround. How I think it should be done in the future:
- TransformerLayer supports thd and we do not need  `setup_before_new_input` at all,
- `InferenceParams` store lengths of cached sequences for each layer,
- for each TransformerLayer invocation, provided sequences are copied to the cache and lengths of cached sequences for this layer are updated,

To do this one will need to remove `save_to_kv_cache()` kernel and write `save_to_kv_cache_sbhd()` and  `save_to_kv_cache_thd()` (no bshd, because cache has shape sbhd both for bshd abd sbhd).
[...]

PR #1478, opened 2025-02-11
RoPE functionality enhancements

sudhakarsingh27:
# Description
TLDR;
Enable application of staggered rope embeddings to different sequences within the same batch.

During generation tasks, different sequences in a batch might have different start positions (technically different end positions as well but that's bounded by the max sequence length in the batch so something we can afford to ignore for now). This change simply modifies the rope kernel to apply the rope embeddings in a staggered manner to different sequences in the batch using an argument `start_positions`. 

(The `start_positions` and related changes are directly adapted from https://github.com/NVIDIA/TransformerEngine/pull/829 which was authored by @pggPL)

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Non breaking changes to `apply_rotary_pos_emb` function but this is non breaking since `start_positions` is a default kwarg here.
- Breaking changes to `FusedRoPEFunc` and all the extensions/kernels that are called internally by this function.


tests/pytorch/test_fused_rope.py
```diff
@@ -41,20 +42,118 @@ def apply_rotary_pos_emb_thd(
...
+    if cp_size > 1:
+        return torch.cat(
+            [
+                apply_rotary_pos_emb(
+                    x.unsqueeze(1), _get_thd_freqs_on_this_cp_rank(cp_rank, cp_size, x, freqs)
+                )
+                for x in torch.split(t, seqlens)
+            ]
+        ).squeeze(1)
+    else:
```
yaox12, line 62: I feel it's better to change the if branches to
```python
if start_positions is None:
    assert cp_size == 1
    ...
else:
    # keep the original code
```
BTW, does `start_positions` works with CP? If so, we should enable it for the CP path as well. With #1626, the fused and unfused versions have the same support matrix for qkv formats/cp, etc.

tests/pytorch/test_fused_rope.py
```diff
@@ -41,20 +42,118 @@ def apply_rotary_pos_emb_thd(
...
+                [
+                    apply_rotary_pos_emb(
+                        x.unsqueeze(1), freqs[start_positions[i] : (x.size(0) + start_positions[i])]
+                    )
+                    for i, x in enumerate(torch.split(t, seqlens))
+                ]
+            ).squeeze(1)
+
+
+def apply_rotary_pos_emb_with_start_positions(
```
yaox12, line 81: It's better to merged this with `apply_rotary_pos_emb` in `rope.py`.

yaox12 (commented):
Generally LGTM. With https://github.com/NVIDIA/TransformerEngine/pull/1626, the fused and unfused versions have the same support matrix for RoPE options, so it's better to merge the unfused implementation to `apply_rotary_pos_emb` in `rope.py`.

transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -87,38 +92,45 @@ __device__ void fused_rope_block_backward(const scalar_t *src, const float *freq
...
-                                          const int stride_h, const int stride_d,
-                                          const int o_stride_s, const int o_stride_b,
-                                          const int o_stride_h, const int o_stride_d) {
+__global__ void fused_rope_forward_kernel(
+    const scalar_t *src, const float *freqs, const int *start_positions, scalar_t *dst, const int h,
+    const int d, const int d2, const int stride_s, const int stride_b, const int stride_h,
+    const int stride_d, const int freqs_stride_s, const int freqs_stride_b, const int o_stride_s,
+    const int o_stride_b, const int o_stride_h, const int o_stride_d) {
   int s_id = blockIdx.x, b_id = blockIdx.y;
+  int begin_offset = (start_positions == 0) ? 0 : start_positions[b_id];
```
yaox12, line 101: Better to use `nullptr` than `0`? Same for some other places.

cyanguwa:
I agree with @yaox12's comments. I think we need to add some documentation about our support matrix for `starting_positions`, how to use it (`None` or `[s,b,1,d] freqs`), and we should support all three `qkv_format` and CP/non-CP cases.

tests/pytorch/test_fused_rope.py
sudhakarsingh27: `start_positions` argument is used for inference use-cases where the input tokens after the "prefill" phase is over, have positions starting at an offset from left. And inference as far as we know doesn't use CP. But the `start_positions` argument could theoretically be used to offset the RoPE embeddings for all the CP ranks and it should work fine.

I'll add some doc/note in the code that makes this explicit

tests/pytorch/test_fused_rope.py
```diff
@@ -202,3 +321,104 @@ def test_fused_rope_thd(
...
+
+@pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16, torch.float16])
+@pytest.mark.parametrize("seq_length", [2048, 4096])
+@pytest.mark.parametrize("hidden_size", [128, 256])
+@pytest.mark.parametrize("rotary_percent", [0.5, 1.0])
+@pytest.mark.parametrize("margin", [10])
+@pytest.mark.parametrize("start_positions", [True, False])
+@pytest.mark.parametrize("tensor_format", ["bshd", "sbhd"])
+@pytest.mark.parametrize("loss_func", [_overlapping_grad, _non_overlapping_grad])
+def test_fused_rope_staggered_inputs(
```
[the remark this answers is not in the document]
sudhakarsingh27: `start_positions` is like `cu_seqlens` in shape: (b,) and contains rope offset for each sequence. 

(I ended up removing the support for `sb1d` rope tensor because it seems like it caused some confusion and can be achieved by using `start_positions`)

tests/pytorch/test_fused_rope.py
sudhakarsingh27: Done

tests/pytorch/test_fused_rope.py
```diff
@@ -182,3 +182,174 @@ def test_fused_rope_thd(
...
+
+@pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16, torch.float16])
+@pytest.mark.parametrize("seq_length", [2048, 4096])
+@pytest.mark.parametrize("hidden_size", [128, 256])
+@pytest.mark.parametrize("rotary_percent", [0.5, 1.0])
+@pytest.mark.parametrize("margin", [10])
+@pytest.mark.parametrize("start_positions", [True, False])
+@pytest.mark.parametrize("tensor_format", ["bshd", "sbhd"])
+@pytest.mark.parametrize("loss_func", [_overlapping_grad, _non_overlapping_grad])
+def test_fused_rope_non_thd_staggered_inputs(
```
cyanguwa, line 195: Could this test be merged with `test_fused_rope`? Similarly, for `test_fused_rope_thd_staggered_inputs` with `test_fused_rope_thd`.

tests/pytorch/test_fused_rope.py
```diff
@@ -182,3 +182,174 @@ def test_fused_rope_thd(
...
+    dtype: torch.dtype,
+    tensor_format: str,
+    loss_func: Callable,
+    start_positions: bool,
+    seq_length: int,
+    hidden_size: int,
+    rotary_percent: float,
+    margin: int,
+) -> None:
+    if margin == 0 and start_positions == True:
```
cyanguwa, line 205: `margin` seems to be 10 always in `parameterize`, so skip seems to be moot. Could we remove it or add `margin=0` test?

tests/pytorch/test_fused_rope.py
```diff
@@ -182,3 +182,174 @@ def test_fused_rope_thd(
...
+        # If sequence to encode has the same lesngth as length of encoding
```
cyanguwa, line 206: "lesngth" -> "length"

tests/pytorch/test_fused_rope.py
```diff
@@ -182,3 +182,174 @@ def test_fused_rope_thd(
...
+
+    # fused
+    output_fused = apply_rotary_pos_emb(
+        t,
+        emb,
+        start_positions=start_positions,
+        fused=True,
+        tensor_format="thd",
+        cu_seqlens=cu_seqlens,
+        cp_size=cp_size,  # cp_size is restricted to `1` deliberately
```
cyanguwa, line 346: Maybe we don't need this in three places? 
"# cp_size is restricted to `1` deliberately"

transformer_engine/common/include/transformer_engine/fused_rope.h
```diff
@@ -37,19 +38,20 @@ extern "C" {
...
+                             const int stride_s_or_t, const int stride_b, const int stride_h,
+                             const int stride_d, cudaStream_t stream);
 
 /*! \brief Compute the backward of the fused rope.
  *
  *  \param[in]     output_grads    Incoming gradient tensor for backward.
  *  \param[in]     cu_seqlens      The cumulative sum of sequence lengths tensor.
  *                                 (Required for the thd format, empty tensor for other formats)
  *  \param[in]     freqs           The freqs tensor.
+ *  \param[in]     start_positions The tensor with positions of first tokens in sequences.
```
cyanguwa, line 54: Could we make the docstring consistent between fwd and bwd? Also, I thought we didn't support training, so no bwd implementation was needed?

transformer_engine/pytorch/dot_product_attention/rope.py
```diff
@@ -199,6 +216,9 @@ def _apply_rotary_pos_emb_base(
     freqs: torch.Tensor
         Rotary positional embedding tensor of shape `[s2, 1, 1, d2]` and dtype 'float',
         with `s2 >= s` and `d2 <= d`.
+    start_positions: torch.Tensor, default = None.
+        Tokens in a sequence `i` should be applied with position encoding offset by
+        `start_positions[i]`. If `start_positions=None`, there's no offset.
     tensor_format: {'sbhd', 'bshd'}, default = 'sbhd'
         Should be `bshd` if `t` is of shape `[bs, seq, ...]`, or `sbhd` if `t` is of shape
         `[seq, bs, ...]`.
```
cyanguwa, line 224: Does `_apply_rotary_pos_emb_base` support `thd`? We seem to be using it with `thd` in `apply_rotary_pos_emb`.

transformer_engine/pytorch/dot_product_attention/rope.py
```diff
@@ -208,14 +228,33 @@ def _apply_rotary_pos_emb_base(
     max_seq_len = freqs.shape[0]
     cur_seq_len = t.shape[1] if tensor_format == "bshd" else t.shape[0]
 
+    # In case `start_positions` are provided, create a staggered `freqs` tensor
+    # offset by the values in `start_positions`.
+    # Note: `start_positions` usage with cp_size > 1 hasn't been tested
+    # because context parallelism isn't used during inference.
```
cyanguwa, line 234: Maybe just say we don't support `start_positions` with CP? Or add it to the support matrix?

transformer_engine/pytorch/dot_product_attention/rope.py
```diff
@@ -252,13 +291,14 @@ def _get_freqs_on_this_cp_rank(
         )
 
     # cp_size == 1
-    return freqs[:seqlen]
+    return freqs
```
cyanguwa, line 294: What if `seqlen` is smaller than `s2`?

transformer_engine/pytorch/dot_product_attention/rope.py
```diff
@@ -252,13 +291,14 @@ def _get_freqs_on_this_cp_rank(
...
 
 
 def apply_rotary_pos_emb(
```
cyanguwa, line 297: Could we add a support matrix for `apply_rotary_pos_emb` detailing the status of our support for `qkv_format`, fused/unfused, training/inference, CP/non-CP, `start_positions`, please?

tests/pytorch/test_fused_rope.py
sudhakarsingh27: I was thinking about it and ended up going this way. But is a single overloaded test a good idea?

tests/pytorch/test_fused_rope.py
sudhakarsingh27: Good catch,

transformer_engine/pytorch/dot_product_attention/rope.py
sudhakarsingh27: That comparison is already done in `apply_rotary_pos_emb_base` and the slicing logic too

tests/pytorch/test_fused_rope.py
sudhakarsingh27: ok

transformer_engine/common/include/transformer_engine/fused_rope.h
sudhakarsingh27: You're right I didn't think about it and the changes seemed symmetrical so I changed both fwd/bwd functions. Since we don't test the backward pass, it might be a good idea to not have it in the backward pass kernels.

transformer_engine/pytorch/dot_product_attention/rope.py
sudhakarsingh27: `apply_rotary_pos_emb` still calls `_apply_rotary_pos_emb_base` for `thd` (individually for each sequence)

transformer_engine/pytorch/dot_product_attention/rope.py
sudhakarsingh27: ok

transformer_engine/pytorch/dot_product_attention/rope.py
sudhakarsingh27: Added now, how does it look?

transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -223,12 +225,14 @@ void fused_rope_forward_launcher(const scalar_t *input, const int *cu_seqlens, c
...
-      input, cu_seqlens, freqs, output, interleaved, cp_size, cp_rank, s, h, d, d2, stride_s_or_t,
-      stride_b, stride_h, stride_d, o_stride_s_or_t, o_stride_b, o_stride_h, o_stride_d);
+      input, cu_seqlens, freqs, start_positions, output, interleaved, cp_size, cp_rank, s, h, d, d2,
+      stride_s_or_t, stride_b, stride_h, stride_d, o_stride_s_or_t, o_stride_b, o_stride_h,
+      o_stride_d);
   NVTE_CHECK_CUDA(cudaGetLastError());
 }
 
 template <typename scalar_t>
+
```
yaox12, line 235:
```suggestion
```

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -278,13 +278,13 @@ void fp8_block_scaling_partial_cast(const at::Tensor &inp, at::Tensor out, const
  **************************************************************************************************/
 
 at::Tensor fused_rope_forward(const at::Tensor &input, const at::Tensor &freqs,
-                              const NVTE_QKV_Format qkv_format, const bool interleaved,
-                              const std::optional<at::Tensor> cu_seqlens, const int cp_size,
-                              const int cp_rank);
+                              const at::Tensor &start_positions, const NVTE_QKV_Format qkv_format,
+                              const bool interleaved, const c10::optional<at::Tensor> cu_seqlens,
```
yaox12, line 282: Please use `std::optional`. PyTorch upstream is deprecating c10::optional. See https://github.com/pytorch/pytorch/pull/150464. And in https://github.com/NVIDIA/TransformerEngine/pull/1683, I replaced all `c10::optional` with `std::optional`.

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -278,13 +278,13 @@ void fp8_block_scaling_partial_cast(const at::Tensor &inp, at::Tensor out, const
...
+                              const int cp_size, const int cp_rank);
 
 at::Tensor fused_rope_backward(const at::Tensor &output_grads, const at::Tensor &freqs,
                                const NVTE_QKV_Format qkv_format, const bool interleaved,
-                               const std::optional<at::Tensor> cu_seqlens, const int cp_size,
+                               const c10::optional<at::Tensor> cu_seqlens, const int cp_size,
```
yaox12, line 287: Ditto.

transformer_engine/pytorch/csrc/extensions/apply_rope.cpp
```diff
@@ -7,9 +7,9 @@
 #include "extensions.h"
 
 at::Tensor fused_rope_forward(const at::Tensor &input, const at::Tensor &freqs,
-                              const NVTE_QKV_Format qkv_format, const bool interleaved,
-                              const std::optional<at::Tensor> cu_seqlens, const int cp_size,
-                              const int cp_rank) {
+                              const at::Tensor &start_positions, const NVTE_QKV_Format qkv_format,
+                              const bool interleaved, const c10::optional<at::Tensor> cu_seqlens,
```
yaox12, line 11: Ditto.

transformer_engine/pytorch/csrc/extensions/apply_rope.cpp
```diff
@@ -87,16 +88,17 @@ at::Tensor fused_rope_forward(const at::Tensor &input, const at::Tensor &freqs,
...
+                          cp_size, cp_rank, s, b, h, d, d2, stride_s, stride_b, stride_h, stride_d,
+                          at::cuda::getCurrentCUDAStream());
 
   return output;
 }
 
 at::Tensor fused_rope_backward(const at::Tensor &output_grads, const at::Tensor &freqs,
                                const NVTE_QKV_Format qkv_format, const bool interleaved,
-                               const std::optional<at::Tensor> cu_seqlens, const int cp_size,
+                               const c10::optional<at::Tensor> cu_seqlens, const int cp_size,
```
yaox12, line 101: Ditto.

transformer_engine/pytorch/dot_product_attention/rope.py
```diff
@@ -119,13 +119,21 @@ def forward(
...
+        start_positions: Union[torch.Tensor, None] = None,
         tensor_format: str = "sbhd",
         interleaved: bool = False,
         cu_seqlens: Union[torch.Tensor, None] = None,
         cp_size: int = 1,
         cp_rank: int = 0,
     ) -> torch.Tensor:
         """Fused RoPE forward."""
+
+        if start_positions is None:
```
yaox12, line 131: How about wrap `start_positions` with `std::optional` in the extension interface, so we don't need to create an empty tensor every time? We did observe some CPU overheads of it before.

transformer_engine/pytorch/dot_product_attention/rope.py
```diff
@@ -268,6 +305,17 @@ def apply_rotary_pos_emb(
...
+    Support matrix:
+    Fused/Unfused:
+        Training:
+            qkv_formats:            "thd", "bshd"/"sbhd"
+            context parallel:       yes
+            start_positions:        no
+        Inference:
+            qkv_formats:            "thd", "bshd"/"sbhd"
+            context parallelism:    no
+            start_positions:        yes
```
cyanguwa, line 317: Can we add `interleaving` in the matrix as well? Thanks!

tests/pytorch/test_fused_rope.py
sudhakarsingh27: Done

transformer_engine/common/include/transformer_engine/fused_rope.h
sudhakarsingh27: backward removed

yaox12 (approved):
LGTM.

tests/pytorch/test_fused_rope.py
```diff
@@ -43,14 +44,33 @@ def test_fused_rope(
...
+    start_positions: bool,
 ) -> None:
+    if margin == 0 and start_positions == True:
+        # This makes sure that the `start_positions` offsets being applied
+        # are with the maximum length of the rope embeddings.
+        pytest.skip("Skipping test with margin=0 and start_positions=True")
+
+    if start_positions == True and cp_size > 1:
+        # Note: `start_positions` is used only during inference and cp_size > 1
+        # isn't used during inference, so skipping this path
```
cyanguwa, line 56: Could we change the comment here to something like "start_positions is only supported for CP=1 and inference only" please? part 1.

tests/pytorch/test_fused_rope.py
```diff
@@ -114,10 +146,26 @@ def test_fused_rope_thd(
     loss_func: Callable,
     cp_size: int,
     interleaved: bool,
+    start_positions: bool,
+    margin: int,
 ) -> None:
+
+    if start_positions == True and cp_size > 1:
+        # Note: `start_positions` is used only during inference and cp_size > 1
+        # isn't used during inference, so skipping this path
```
cyanguwa, line 155: Could we change the comment here to something like "start_positions is only supported for CP=1 and inference only" please? part 2.

transformer_engine/pytorch/dot_product_attention/rope.py
```diff
@@ -292,27 +341,44 @@ def apply_rotary_pos_emb(
     cp_rank: int, default = 0.
         Context parallel rank. Only valid when `tensor_format` is 'thd' and `fused` is True.
     """
+
+    # Note: `start_positions` is used only during inference and so not supported
+    # with context parallelism.
```
cyanguwa, line 346: Could we change the comment here to something like "start_positions is only supported for CP=1 and inference only" please? part 4.

cyanguwa (approved):

PR #1486, opened 2025-02-15
Explicitly use `python3` and `pip3` executables

timmoon10:
# Description

Not all systems configure `python` to point to the Python 3 executable. For example, try installing Python in an Ubuntu Docker container and calling `python`:
```bash
docker run ubuntu:24.04 bash -c "apt-get update && apt-get install -y python3 && python --version"
```
This will fail with:
```
bash: line 1: python: command not found
```
This PR explicitly specifies `python3` and `pip3` so it can work on such systems. This will be especially helpful for testing.

I've also replaced some Python executables (Pylint, cpplint, pre-commit) with `python -m`. I don't like running containers as root, so Pip installs locally and the executables are not in my `PATH`.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Explicitly use `python3` and `pip3` instead of `python` and `pip`
- Run Pylint, cpplint, and pre-commit with `python -m` 

# Checklist:


ptrendx (approved):

qa/L0_jax_unittest/test.sh
```diff
@@ -4,20 +4,20 @@
...
 set -xe
 
-pip install "nltk>=3.8.2"
-pip install pytest==8.2.1
+pip3 install "nltk>=3.8.2"
+pip3 install pytest==8.2.1
 : ${TE_PATH:=/opt/transformerengine}
 
-pytest -c $TE_PATH/tests/jax/pytest.ini -v $TE_PATH/tests/jax -k 'not distributed' --ignore=$TE_PATH/tests/jax/test_praxis_layers.py
+python3 -m pytest -c $TE_PATH/tests/jax/pytest.ini -v $TE_PATH/tests/jax -k 'not distributed' --ignore=$TE_PATH/tests/jax/test_praxis_layers.py
```
KshitijLakhani, line 11: @timmoon10 This should also be helpful if there's multiple pytest versions right (by using the one in the current environment) ?

KshitijLakhani (approved):
@timmoon10 Is there a need to make this change in any of the below or is the scope limited to testing code only ?
1. `build_wheels.sh` 
2. nit: assert String in `jax/__init__.py` and `pytorch/__init__.py` 
3. nit: FA string in `attention.py` 

Do we also plan to go explicit with the `pip3` usage in the README docs or is that out of scope ?

Static code review. LGTM !

PR #1474, opened 2025-02-11
Fix issues for MCore DDP.

Victarry:
# Description

Please include a brief summary of the changes, relevant motivation and context.

Fixes # (issue)
```
4: [rank4]: Traceback (most recent call last):
4: [rank4]:   File "/workspace/megatron-lm/pretrain_gpt.py", line 245, in <module>
4: [rank4]:     pretrain(
4: [rank4]:   File "/workspace/megatron-lm/megatron/training/training.py", line 313, in pretrain
4: [rank4]:     iteration, num_floating_point_operations_so_far = train(
4: [rank4]:                                                       ^^^^^^
4: [rank4]:   File "/workspace/megatron-lm/megatron/training/training.py", line 1157, in train
4: [rank4]:     train_step(forward_step_func,
4: [rank4]:   File "/workspace/megatron-lm/megatron/training/training.py", line 631, in train_step
4: [rank4]:     losses_reduced = forward_backward_func(
4: [rank4]:                      ^^^^^^^^^^^^^^^^^^^^^^
4: [rank4]:   File "/workspace/megatron-lm/megatron/core/pipeline_parallel/schedules.py", line 456, in forward_backward_no_pipelining
4: [rank4]:     backward_step(input_tensor, output_tensor, output_tensor_grad, model_type, config)
4: [rank4]:   File "/workspace/megatron-lm/megatron/core/pipeline_parallel/schedules.py", line 356, in backward_step
4: [rank4]:     custom_backward(output_tensor[0], output_tensor_grad[0])
4: [rank4]:   File "/workspace/megatron-lm/megatron/core/pipeline_parallel/schedules.py", line 155, in custom_backward
4: [rank4]:     Variable._execution_engine.run_backward(
4: [rank4]:   File "/workspace/megatron-lm/megatron/core/distributed/distributed_data_parallel.py", line 223, in param_hook
4: [rank4]:     param.grad is not None
4: [rank4]: AssertionError: param.grad being None is not safe when overlap_grad_reduce is True
```
## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Please list the changes introduced in this PR:

- Change A
- Change B

# Checklist:

[...]

PR #1488, opened 2025-02-17
[PyTorch] Fix fuse_wgrad_accumulation for GroupedLinear

yaox12:
# Description

Due to the wrong indent, the wgrad computation is not called when `ctx.fuse_wgrad_accumulation == True`.

Also update the test to cover this case.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/test_numerics.py
+++ b/tests/pytorch/test_numerics.py
@@ -1548,6 +1548,7 @@ def test_grouped_linear_accuracy_parallel_mode(parallel_mode, recipe):
         recipe=recipe,
         fp8_model_params=True,
         parallel_mode=parallel_mode,
+        fuse_wgrad_accumulation=True,
     )
 
 
@@ -1562,6 +1563,7 @@ def test_grouped_linear_accuracy_single_gemm(recipe):
         fp8=True,
         recipe=recipe,
         fp8_model_params=True,
+        fuse_wgrad_accumulation=True,
     )
 
 
--- a/tests/pytorch/test_numerics.py
+++ b/tests/pytorch/test_numerics.py
@@ -1400,7 +1400,9 @@ def test_layernorm_mlp_accuracy(dtype, bs, model, activation, normalization):
             assert_allclose(te_output, torch_output, atol[dtype], rtol[dtype])
 
 
-def _test_grouped_linear_accuracy(block, num_gemms, bs, dtype, config, recipe, fp8, fuse_wgrad_accumulation):
+def _test_grouped_linear_accuracy(
+    block, num_gemms, bs, dtype, config, recipe, fp8, fuse_wgrad_accumulation
+):
     reset_rng_states()
     if fp8:
         FP8GlobalStateManager.reset()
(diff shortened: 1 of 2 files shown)


transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -178,17 +178,18 @@ def forward(
...
-            saved_inputs, saved_weights = [], []
             ctx.weights_shape_1 = weights[0].shape[1]
 
             tensors_to_save, tensor_objects = prepare_for_saving(*inputmats, *weights_fp8, *biases)
             ctx.save_for_backward(*tensors_to_save)
             ctx.tensor_objects = tensor_objects
 
             ctx.weights_requires_grad = weights[0].requires_grad
+            if fuse_wgrad_accumulation and ctx.weights_requires_grad:
+                ctx.main_grads = [weights[i].main_grad for i in range(num_gemms)]
```
timmoon10, line 189: It's recommended to use [`ctx.save_for_backward`](https://pytorch.org/docs/stable/generated/torch.autograd.function.FunctionCtx.save_for_backward.html#torch.autograd.function.FunctionCtx.save_for_backward) instead of storing tensors directly in `ctx`. They warn about messing up the grad graph and memory leaks, although I'm not sure what cases they are specifically worried about.

tests/pytorch/test_numerics.py
```diff
@@ -1447,7 +1449,11 @@ def _test_grouped_linear_accuracy(block, num_gemms, bs, dtype, config, recipe, f
     outputs = [out, inp_hidden_states.grad]
     for p in block.parameters():
         if p.requires_grad:
-            outputs.append(p.grad)
+            if getattr(p, "main_grad", None) is not None:
+                outputs.append(p.main_grad)
+                assert p.grad is None  # grad should be None if fuse_wgrad_accumulation is True
```
timmoon10, line 1454: It turns out Mcore expects `p.grad` to not be `None`: https://github.com/NVIDIA/TransformerEngine/pull/1474#issuecomment-2652816973
https://github.com/NVIDIA/TransformerEngine/pull/1474 sets `grad` to an uninitialized tensor and assumes Mcore will ignore it.

transformer_engine/pytorch/module/grouped_linear.py
yaox12: I agree. We were saving `main_grad` tensors in `ctx.save_for_backward` in TE 1.x. But I'm see there is comment here. https://github.com/NVIDIA/TransformerEngine/blob/6673f1658c605f7999e04a43bde1493c4b9ad741/transformer_engine/pytorch/module/linear.py#L356
I'm wondering if we have seeing issues with `ctx.save_for_backward`?

timmoon10: Interesting, we should follow the example of `Linear` then.

@ksivaman This change is from commit `7e58678` in the internal repo. Do you remember why we can't store `main_grad` in `saved_tensors`?

timmoon10 (approved):

transformer_engine/pytorch/module/grouped_linear.py
yaox12: Oh I know, previous `prepare_for_saving` saves `tensor.data` instead of the tensor itself. So for `main_grad` that need to be modified inplace, this could be an issue.

Now #1474 changed `prepare_for_saving` to save the tensor itself, this is no longer a problem.

PR #1498, opened 2025-02-20
[Pytorch] Decoupling framework extensions from common module

KshitijLakhani:
# Description

**Motivation:** Currently, framework extensions has a dependency on `transformer_engine/common` (`libtransformer_engine.so`) which is met via the symbols exposed by `libtransformer_engine.version`.  However, this could result in two problems:
1. Over exposure of internal symbols from transformer_engine to the user
2. Possible compatibility issues if a different compiler / compiler version is used for building the .so and building the framework extensions code

This PR is the first in a series of decoupling PRs to follow

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

Files in the framework extensions identified to have this dependency in the first pass are - 

In `pytorch/csrc/extensions`:
`attention.cu`, `gemm.cpp`, `multi_tensor_adam.cu`

In `jax/csrc/extensions`:
`attention.cpp`, `utils.cu`

This PR only decouples `pytorch/csrc/extensions/attention.cu` by :
1. Using `nvte_*` calls to replace the usage of `transformer_engine::Tensor`
2. Moving the structs in `t_e/common/fused_attn/thd_util.h` to a new file `t_e/pytorch/csrc/thd_util.cuh`
4. Templatizing `thd_partition_indices_kernel` and `thd_read_half_tensor_kernel` so that they are "re-compiled" on instantiation in the framework extensions

# Checklist:


transformer_engine/common/fused_attn/thd_utils.h
```diff
@@ -7,6 +7,7 @@
 #ifndef TRANSFORMER_ENGINE_FUSED_ATTN_THD_UTILS_H_
 #define TRANSFORMER_ENGINE_FUSED_ATTN_THD_UTILS_H_
 
+#include <assert.h>
```
KshitijLakhani, line 10: In place of `#include "../cudnn_utils.h"` that was otherwise used in `thd_utils.cu`.
There did not seem to be a need for `cudnn_utils.h` for the two kernels `thd_partition_indices_kernel` and `thd_read_half_tensor_kernel`. `assert.h` suffices.

transformer_engine/pytorch/csrc/extensions/attention.cu
```diff
@@ -4,11 +4,9 @@
  * See LICENSE for license information.
  ************************************************************************/
 
-#include "common/common.h"
 #include "common/fused_attn/thd_utils.h"
 #include "extensions.h"
-
-using namespace transformer_engine::fused_attn;
+#include "thd_util.cuh"
```
KshitijLakhani, line 9: This is the new header file created  in the framework extensions for the structs that were being used in `attention.cu` from `transformer_engine/common/fused_attn/thd_utils.h`.

transformer_engine/pytorch/csrc/thd_util.cuh
```diff
@@ -0,0 +1,59 @@
+/*************************************************************************
```
KshitijLakhani, line 1: Is `transformer_engine/common/include/transformer_engine` a better place for this ?

transformer_engine/common/fused_attn/thd_utils.h
```diff
@@ -33,39 +34,80 @@ __forceinline__ __device__ int binary_search(int target, int *array, int len) {
...
+    index += cu_seqlens_s[seq_id] * world_size + seq_len / 2 * offset;
+    output[token_id] = index;
+  }
+}
 
 /***************************************************************************************************
  * Support THD format for Context Parallel: Read the half of a THD tensor
  **************************************************************************************************/
-
+// Templatizing this kernel ONLY so that when it is used in framework
```
ptrendx, line 68: Maybe a better solution actually is to just create an unnamed namespace around those kernels? Especially since the naming of those templates are confusing, especially this one: I would interpret `half_dtype` as the FP16 you support in this kernel (like FP16 or BF16 for example).

transformer_engine/common/fused_attn/thd_utils.h
```diff
@@ -7,6 +7,7 @@
 #ifndef TRANSFORMER_ENGINE_FUSED_ATTN_THD_UTILS_H_
```
ptrendx, line 7: I don't know how I feel about this file staying in common. Why not moving it to pytorch/csrc instead?

cyanguwa: I wonder if we could do something like [pytorch/csrc/multi_tensor_apply.cuh](https://github.com/NVIDIA/TransformerEngine/tree/main/transformer_engine/pytorch/csrc) for `thd_utils.h` and `thd_utils.cuh`?

transformer_engine/pytorch/csrc/thd_util.cuh
```diff
@@ -0,0 +1,59 @@
...
+ * Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+#pragma once
```
cyanguwa, line 6: I wonder if we should do the usual include guards instead of this pragma, which is not standard.
```
#ifndef TRANSFORMER_ENGINE_FUSED_ATTN_THD_UTILS_H_
#define TRANSFORMER_ENGINE_FUSED_ATTN_THD_UTILS_H_
...
#endif
```

transformer_engine/common/fused_attn/thd_utils.h
KshitijLakhani: @ptrendx Thank you for bringing this up. I forgot to add this as part of my comments in the PR. I agree and was also of the opinion that this should not be in common because there's no use for it there (as of now) so I wanted to inquire if there's a known use for it in the future.

@cyanguwa I agree with you. The new file that I added, `t_e/pytorch/csrc/thd_util.cuh`, which selectively contains structs from `t_e/common/fused_attn/thd_utils.h` can just be repurposed for this use. I'll try it and post an update.

KshitijLakhani: Pushed a commit for this

transformer_engine/common/fused_attn/thd_utils.h
KshitijLakhani: Moving this function out of common and into framework extensions removes the need for it to be a template anymore.
RE: https://github.com/NVIDIA/TransformerEngine/pull/1498#discussion_r1964743044

transformer_engine/pytorch/csrc/thd_utils.cuh
```diff
@@ -33,39 +86,78 @@ __forceinline__ __device__ int binary_search(int target, int *array, int len) {
 /***************************************************************************************************
  * Support THD format for Context Parallel: Generate partitioned indices for input tokens
  **************************************************************************************************/
-
+// Templatizing this kernel ONLY so that when it is used in framework
+// extensions, it is dynamically compiled
```
cyanguwa, line 90: We could probably remove these comments now.

transformer_engine/pytorch/csrc/thd_utils.cuh
```diff
@@ -3,12 +3,65 @@
...
 
-#ifndef TRANSFORMER_ENGINE_FUSED_ATTN_THD_UTILS_H_
-#define TRANSFORMER_ENGINE_FUSED_ATTN_THD_UTILS_H_
-
+#include <assert.h>
 #include <cuda.h>
 #include <cuda_bf16.h>
+/***************************************************************************************************
+ * Support THD format for Context Parallel: softmax_lse related operations
+ **************************************************************************************************/
```
cyanguwa, line 14: Maybe have a cleanup of these comments. They seem unrelated to the functions below. Same for the Gradients ones.

ptrendx (approved):

PR #1500, opened 2025-02-21
Delete extra tensor objects after restoring float8 tensors

sudhakarsingh27:
# Description

After restoring the `float8` tensors in the backward passes of `LayernormMLP`, `LayernormLinear` and `Linear`, the tensor objects are not needed but `ctx.tensor_objects` still holds the reference and hence it results in extra memory usage. This fixes it.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Delete extra reference to `tensor_objects` once they're used in the backwards of `LayernormMLP`, `LayernormLinear` and `Linear` modules. 

# Checklist:


transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
```diff
@@ -105,8 +105,8 @@ def prepare_for_saving(self) -> Tuple[list[Optional[torch.Tensor]], Float8Tensor
 
         """
         tensors = [self._data, self._transpose]
-        # self._data = None
-        # self._transpose = None
+        self._data = None
+        self._transpose = None
```
ksivaman, line 109: @pggPL IIRC you removed these during a numerics debugging effort, do you remember why?

pggPL: If weight is in fp8 I want to have it in save_for_backward() - for offloading. If there is forward, but backward is not invoked, it will result in removing the weight. I discussed it with @ptrendx and he proposed some solution with flag `internal` in `prepare_for_saving` - to set it True if tensor is not owned and remove tensors iff they are internal. It seems that we forgot about this.

ptrendx: Ok, so that would be solved by overriding this function in Float8Tensor and MXFP8Tensor to just return self and None instead.

Also, in https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/tensor/quantized_tensor.py#L30 why do we check for exactly Tensor or Param and not just isinstance(torch.Tensor)? This should solve this as well, right?

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -448,6 +448,9 @@ def backward(
                 mu,
                 rsigma,
             ) = restore_from_saved(ctx.tensor_objects, saved_tensors)
+            # Delete the references to tensor objects once they've been consumed
+            # by the `restor_from_saved` method to construct back the actual tensors.
```
ptrendx, line 452: nit: restor**e**_from_saved, same for the subsequent comments.

transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
ptrendx: Well, QuantizedTensor is in a way a standard tensor - at least it can be passed whole through save_for_backward, so there is nothing to restore afterwards.

transformer_engine/pytorch/module/layernorm_linear.py
sudhakarsingh27: ys, I should also fix the  ky in my kyboard

ptrendx (approved):

PR #1502, opened 2025-02-22
Fix a crash in NeMo 2.0 during module._apply(lambda t: t.cpu())

guyueh1:
# Description

In Nemo 2.0 during job exit, lightning calls a `module._apply(lambda t: t.cpu())` on the GPT model which triggers an illegal memory access error in the TE dequantize kernel. This PR fixes the issue. 

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


guyueh1:
@timmoon10 
Re: We are passing a CPU tensor into the quantize kernel, and I don't think we ever move it to GPU. 
I checked the device of 'tensor' in debug mode, after `torch.cuda.synchronize()` it is on cpu, however, the `self._quantizer.quantize(tensor)` worked fine. So there should be some code to handle cpu or move it to gpu somewhere.

I also don't fully understand the race condition, it just happened to work. I will dig into it, and reply here my findings.

guyueh1:
Findings: 
1. Adding `torch.cuda.synchronize()` at beginning of `_set_data` works (the original proposal of this PR)
2. @timmoon10 's suggestion (`self.data = self._quantizer.quantize(tensor.to(device=self.device))`) also works  
Still digging why IMA happens

guyueh1:
I decide to revert the `torch.cuda.synchronize()` change because I can't understand why it would work; I apply a new way to fix it by making sure the `tensor` is moved to `self.device` if it were on CPU. I confirmed it fixed the IMA.
@timmoon10 what you think about the current version

timmoon10 (approved):
LGTM

Issue #1283, opened 2024-10-22
ImportError in cuda 12


wplf:
Hi, ImportError happened when I import TE.
```
 File "/usr/local/lib/python3.10/dist-packages/transformer_engine/pytorch/__init__.py", line 29, in _load_library
    solib = importlib.util.module_from_spec(spec)
ImportError: /usr/local/lib/python3.10/dist-packages/transformer_engine/transformer_engine_torch.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZN5torch7LibraryC1ENS0_4KindESsSt8optionalIN3c1011DispatchKeyEEPKcj
```

The environ is  
- torch `2.3.0a0+40ec155e58.nv24.03`
- cuda 12.4
- cudnn 90000

I have found this symbol in https://github.com/vllm-project/vllm/issues/5501. 
This symbol is defined in  libtorch_cpu.so https://github.com/vllm-project/vllm/issues/5501#issuecomment-2167338928

Could you help me with this? Thanks.

wplf:
Thank you very much. I've got the source of bug. 

To clarify, the test machine is `cuda 12.4, torch=2.3.0`. And in the beginning, torch is from nvcr:23.10, which is 2.3.0, build TE is built on another machine `cuda 12.1, torch=2.2.2`, and got this bug.

I've tried to build TE from source, but got another bug `cicc is not found`. 
I've checkout that `cicc` is in the folder of `nvvm` in cuda and tried to add the folder of `cicc` into `$PATH`, but still got `cicc` is not found. Do you have any suggestion for this?

timmoon10:
Strange that you need to go into the CUDA internals. Can you try setting `CUDA_PATH` in your environment? Here's a list of common problems when building TE: https://github.com/NVIDIA/TransformerEngine/issues/355#issuecomment-2394353816.

wplf:
CUDA and cicc is both in my path, I guess this is the machine's problem. 
Thank you very much

wplf:
Hi, thank you very much.
I build TE from source in the exactly same environment. This problem disappears. 

Thanks you, again.

Issue #756, opened 2024-04-06
_ZN18transformer_engine6getenvIiEET_RKSsRKS1_ on the latest main branch


leiwen83:
Hi,

When I try the example located in docs/examples/te_llama with latest code, met below error:

/usr/local/lib/python3.10/dist-packages/transformer_engine_extensions.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZN18transformer_engine6getenvIiEET_RKSsRKS1_

lucifer1004:
Same here.

danthe3rd:
Getting the same as well on `release_v1.5`. `release_v1.4` works fine tho

ptrendx:
@pggPL Could you take a look at this issue?

jinzex:
The issue happens because we recently included `common/util/system.h` into PyTorch source CPP files at https://github.com/NVIDIA/TransformerEngine/pull/713/files to utilize the `transformer_engine::getenv` method. As a result, the PyTorch `transformer_engine_extensions` will now search for symbols within the `libtransformer_engine.so` library.

However, if the user builds TransformerEngine in a local environment (conda or PyPI), because the pre-built PyTorch, either from conda or PyPI, has the CXX11_ABI set as False, whereas the common library is using the new CXX11_ABI. As a result, there will be an undefined symbol problem when importing `transformer_engine_extensions`.

The following could be a possible fix: compile system.cpp directly to avoid linking directly with libtransformer_engine.so. Please let me know if this makes sense, and I could open a PR for it.

https://github.com/NVIDIA/TransformerEngine/blob/bfe21c3d68b0a9951e5716fb520045db53419c5e/setup.py#L443-L449

possible fix:
```python
    sources = [
        src_dir / "common.cu",
        src_dir / "ts_fp8_op.cpp",
        # We need to compile system.cpp because the extension uses transformer_engine::getenv.
        # This is a workaround to avoid linking directly with libtransformer_engine.so.
        root_path / "transformer_engine" / "common" / "util" / "system.cpp",
    ] + \
    _all_files_in_dir(extensions_dir)
```

timmoon10:
Ah, ABI issues make things challenging. If ABI incompatibility breaks `transformer_engine::getenv`, there's no reason to believe we can rely on any other C++ interfaces between `libtransformer_engine.so` and the PyTorch extension. This makes me think the right solution is to make the following refactors:
[...]

PR #1511, opened 2025-02-26
Export only necessary symbols from libtransformer_engine.so

KshitijLakhani:
# Description

Make the `libtransformer_engine.version` export map leaner. Additionally, perform any other decouplings between framework extensions and common.

NOTE: Two good references for export map symbols - 
1. https://accu.org/journals/overload/15/79/wakeling_1372/
2. How To Write Shared Libraries : Ulrich Drepper, 2005

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

1. Limit the symbols exposed during the linking of `libtransformer_engine.so` so the symbol interface is lean ([Related PR 1498](https://github.com/NVIDIA/TransformerEngine/pull/1498))
2. Combine `transformer_engine/common/util/system.cpp` and `transformer_engine/common/util/system.h` into a single header only file so that `getenv()` does not need to be specifically exposed 
(RE: https://github.com/NVIDIA/TransformerEngine/pull/1511#discussion_r1972176727)
3. Start decoupling `transformer_engine/jax/csrc/extensions/attention.cpp`  from `transformer_engine/common` by creating a `product()` function in jax extensions misc.

# Checklist:


transformer_engine/common/libtransformer_engine.version
```diff
@@ -1,4 +1,18 @@
...
-	global: *nvte*; *transformer_engine*;
+	global:
+		extern "C++" {
+			nvte_*;
+			transformer_engine::cuda::sm_count*;
+			transformer_engine::cuda::sm_arch*;
+			transformer_engine::cuda::supports_multicast*;
+			transformer_engine::cuda::stream_priority_range*;
+			transformer_engine::ubuf_built_with_mpi*;
+			*transformer_engine::getenv*;
```
timmoon10, line 10: We've experienced ABI problems with `getenv` before (see https://github.com/NVIDIA/TransformerEngine/issues/1283#issuecomment-2430589220 and https://github.com/NVIDIA/TransformerEngine/issues/756#issuecomment-2044108227).
```suggestion
```

transformer_engine/common/libtransformer_engine.version
```diff
@@ -1,4 +1,20 @@
 {
-	global: *nvte*; *transformer_engine*;
+	global:
+		extern "C++" {
+			nvte_*;
```
cyanguwa, line 4: These functions are pure C so probably don't need to be in extern C++.

transformer_engine/common/libtransformer_engine.version
```diff
@@ -1,4 +1,20 @@
...
+			transformer_engine::cuda::supports_multicast*;
+			transformer_engine::cuda::stream_priority_range*;
+			transformer_engine::cuda::current_device*;
+			transformer_engine::cuda_driver::get_symbol*;
+			transformer_engine::ubuf_built_with_mpi*;
+			*transformer_engine::rtc*;
+			transformer_engine::nvte_cudnn_handle_init*;
+			transformer_engine::typeToSize*;
+			*transformer_engine::CommOverlapBase*;
+			*transformer_engine::CommOverlapP2PBase*;
```
KshitijLakhani, line 16: A `*` prior to eh symbol name is used when `typeinfo` symbols (such as dynamic casting, exception handling, etc.) need to be exported.

If only the functions within a class or only a function or data type is to be exposed just the wildcard `*` after the symbol is enough

KshitijLakhani:
I ran the failing CI tests `L0_unittest_jax`, `L0_unittest_pytorch`, `L0_jax_wheel` and `L0_pytorch_wheel` on `L40` 1 GPU node locally and I m unable to replicate the errors reported in the CI so I believe it's a known node issue that causes the CI to fail for these tests.
This should not block merging.

transformer_engine/common/libtransformer_engine.version
KshitijLakhani: @cyanguwa It seems like there are non-c style functions with `nvte_` which get mangled in the symbol table (e.g. `nvte_unpermuted()`). Therefore, if I simply move out `nvte_*` out of the "extern c++" then I get a linker error for `nvte_unpermute()` for example.

However, if we do wish to move `nvte_` out of the `extern C++` it is more correct to use `*nvte_*` as that takes care of the symbols correctly without causing linker errors, but it may also end up exposing more symbols as part of the wildcard than the other option.

So, either we let it be as is or move it out and an additional wildcard `*` prior to `nvte_*` . I'd prefer to keep it as is.

cyanguwa: Sounds good.

transformer_engine/common/util/system.h
```diff
@@ -7,25 +7,86 @@
...
 
+#include <cstdint>
+#include <cstdlib>
+#include <filesystem>
+#include <fstream>
+#include <sstream>
 #include <string>
 
-namespace transformer_engine {
+//#include "../common.h"
```
cyanguwa, line 17: Remove this commented line please.

cyanguwa (approved):

transformer_engine/common/util/system.h
KshitijLakhani: Good catch. Thank you. Addressed in f920e8c97c841113274b7b87cf2e88ba1d8372d2

transformer_engine/common/util/system.h
```diff
@@ -7,25 +7,85 @@
...
+NVTE_INSTANTIATE_GETENV(int8_t, 0);
+NVTE_INSTANTIATE_GETENV(int16_t, 0);
+NVTE_INSTANTIATE_GETENV(int32_t, 0);
+NVTE_INSTANTIATE_GETENV(int64_t, 0);
+NVTE_INSTANTIATE_GETENV(uint8_t, 0);
+NVTE_INSTANTIATE_GETENV(uint16_t, 0);
+NVTE_INSTANTIATE_GETENV(uint32_t, 0);
+NVTE_INSTANTIATE_GETENV(uint64_t, 0);
+NVTE_INSTANTIATE_GETENV(std::string, std::string());
+NVTE_INSTANTIATE_GETENV(std::filesystem::path, std::filesystem::path());
```
timmoon10, line 84: Explicit template instantiation like this is not great in header files: anything that includes this header will have to compile everything. We can avoid that with another templated helper function.
```c++
namespace detail {

template <typename T>
inline typename std::enable_if<std::is_arithmetic<T>::value, T>::type getenv_helper(
    const char *variable, const T &default_value) {
  // Implementation for numeric types
  const char *env = std::getenv(variable);
  if (env == nullptr || env[0] == '\0') {
    return default_value;
  }
  T value;
  std::istringstream iss(env);
  iss >> value;
  NVTE_CHECK(iss, "Invalid environment variable value");
  return value;
}

template <typename T>
inline typename std::enable_if<!std::is_arithmetic<T>::value, T>::type getenv_helper(
    const char *variable, const T &default_value) {
  // Implementation for string-like types
  const char *env = std::getenv(variable);
  if (env == nullptr || env[0] == '\0') {
    return default_value;
  } else {
    return env;
  }
}

template <typename T>
inline T getenv_default_value() {
  return 0;
}

template <>
inline std::string getenv_default_value<std::string>() {
  return {};
}

template <>
inline std::filesystem::path getenv_default_value<std::filesystem::path>() {
  return {};
}

}  // namespace detail

/*! \brief Get environment variable and convert to type */
template <typename T = std::string>
inline T getenv(const char *variable, const T &default_value) {
  return detail::getenv_helper<T>(variable, default_value);
}

/*! \brief Get environment variable and convert to type
 *
 * If the environment variable is unset or empty, a falsy value is
 * returned.
 */
template <typename T = std::string>
inline T getenv(const char *variable) {
  return getenv<T>(variable, detail::getenv_default_value<T>());
}
```

KshitijLakhani:
Launched L0 pipeline for JAX and PyTorch on GL: https://gitlab-master.nvidia.com/dl/transformerengine/transformerengine/-/pipelines/24798479

CI failures:
- L0 cppunittest for B200 fails (same as main CI)
- L0 jax unittest for A100 fails (I ran the test successfully locally on A100x2 on cluster)
- L0 pytorch T4 1 GPU (job is stalled)

Seems okay to merge. Will update this comment if there's any changes

transformer_engine/common/util/system.h
```diff
@@ -7,25 +7,96 @@
...
+inline std::filesystem::path getenv_default_value<std::filesystem::path>() {
+  return std::filesystem::path();
+}
+
+}  // namespace detail
+
 /*! \brief Get environment variable and convert to type
  *
- * If the environment variable is unset or empty, a falsy value is
+ * If the environment variable is unset or empty, a false value is
```
timmoon10, line 83: Nit: "falsy" is more accurate since we're not actually returning `false` (https://developer.mozilla.org/en-US/docs/Glossary/Falsy). That said, it seems "falsy" and "truthy" are more used in the JavaScript world.
```suggestion
 * If the environment variable is unset or empty, a falsy value is
```

timmoon10 (approved):
LGTM. We can merge once we fix the DCO failure.

transformer_engine/common/util/system.h
KshitijLakhani: Thanks for the link to falsy. I wasn't aware of falsy and assumed it was a typo :P

KshitijLakhani:
@timmoon10 Addressed the DCO check (missing sign off in commit SHA 28fd27d7131cf7144d65f989b1a0ddf2f1731e63) and also github-accepted your suggestion for reverting falsy.No additional code changes. 
Going ahead with merge. Thanks !

transformer_engine/common/util/system.h
timmoon10: Seems like it's slang. When looking for the formal definition, I was surprised that it isn't really used anywhere official except JavaScript.

PR #1515, opened 2025-02-27
Fix shape of new quantized tensor in `make_like`

ksivaman:
# Description

In functions such as `split`, `chunk` etc. in which the shape of the input and output differs, the returned tensor is correct but with the incorrect shape which leads to bugs, e.g. in FSDP2 or checkpoint loading. A small repro:

```
import torch
from transformer_engine.pytorch.tensor.float8_tensor import Float8Quantizer

t = torch.randn(4, 4)
quantizer = Float8Quantizer(
    scale=torch.full([1], 1.0, dtype=torch.float32, device="cuda"),
    amax=torch.empty([1], dtype=torch.float32, device="cuda"),
    fp8_dtype=tex.DType.kFloat8E4M3,
)
x = quantizer(t.cuda())
a, b = x.chunk(2, dim=0)
print(x.shape, a.shape, b.shape)
print(x, a, b)
``` 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- If data

# Checklist:

### DIFF
--- a/tests/pytorch/test_float8tensor.py
+++ b/tests/pytorch/test_float8tensor.py
@@ -161,6 +161,36 @@ class TestFloat8Tensor:
         with pytest.raises(AssertionError):
             torch.testing.assert_close(x_fp8 + y_fp8, x_ref - y_fp8, **tols)
 
+    @pytest.mark.parametrize("dims", [2, [4, 4], [8, 5, 3, 3]])
+    def test_chunk_op(
+        self,
+        dims: DimsType,
+        fp8_dtype: tex.DType = tex.DType.kFloat8E4M3,
+        scale: float = 3.5,
+        dtype: torch.dtype = torch.float32,
+    ) -> None:
+        """Test for ops for which shape of inputs and outputs differ."""
+
+        # Initialize random data
+        dims = _to_list(dims)
+        x_ref = torch.randn(dims, dtype=dtype, device="cpu")
+        x_fp8 = to_float8(x_ref, fp8_dtype=fp8_dtype, scale=1.0)
+
+        # Get chunks.
+        chunk1, chunk2 = x_fp8.chunk(2, dim=0)
+
+        # Test chunks.
+        torch.testing.assert_close(x_fp8[0 : dims[0] // 2,], chunk1, atol=0, rtol=0)
+        torch.testing.assert_close(x_fp8[dims[0] // 2 :,], chunk2, atol=0, rtol=0)
+
+        # Check shapes.
+        assert (
+            chunk1.shape == torch.Size([x_fp8.shape[0] // 2]) + x_fp8.shape[1:]
+        ), "Wrong shape for chunk1"
+        assert (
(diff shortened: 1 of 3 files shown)


timmoon10 (approved):
This is fine as a hack,.

Rambling digression: I don't like the `data` kwarg. It's not generic and going forward there's no reason to expect this logic to be valid in the future (FP6 will probably require a randomly sized blob of bytes). Also, if you're providing `data` you should also probably provide `scale_inv` (and maybe column-wise data as well). I think it's fine if it's an optional kwarg in `Float8Tensor`, but it's a bad design to expose in `QuantizedTensor`.

ptrendx:
I agree with Tim, why not just add the shape argument to the usage of the make_like in float8tensor:
```diff
diff --git a/transformer_engine/pytorch/tensor/float8_tensor.py b/transformer_engine/pytorch/tensor/float8_tensor.py
index 49bf4facf..0063b286a 100644
--- a/transformer_engine/pytorch/tensor/float8_tensor.py
+++ b/transformer_engine/pytorch/tensor/float8_tensor.py
@@ -402,7 +402,7 @@ class Float8Tensor(Float8TensorBase, QuantizedTensor):
                 [data] + list(args[1:]),
                 kwargs,
             )
-            return [Float8Tensor.make_like(tensor, data=split_tensor) for split_tensor in func_out]
+            return [Float8Tensor.make_like(tensor, data=split_tensor, shape=split_tensor.shape) for split_tensor in func_out]
         if func == aten.new_zeros.default:
             tensor = args[0]
             data = tensor._data
@@ -412,7 +412,7 @@ class Float8Tensor(Float8TensorBase, QuantizedTensor):
                 [data] + list(args[1:]),
                 kwargs,
             )
-            return Float8Tensor.make_like(tensor, data=func_out)
+            return Float8Tensor.make_like(tensor, data=func_out, shape=data.func_out.shape)
         if func == torch.ops.aten.as_strided.default:
             tensor = args[0]
             data = tensor._data
@@ -422,7 +422,7 @@ class Float8Tensor(Float8TensorBase, QuantizedTensor):
                 [data] + list(args[1:]),
                 kwargs,
             )
-            return Float8Tensor.make_like(tensor, data=func_out)
+            return Float8Tensor.make_like(tensor, data=func_out, shape=data.func_out.shape)
         if func == torch.ops.aten.detach.default:
             return cls.detach(args[0])
         if func == torch.ops.aten.clone.default:
```
I confirmed that it also resolves the given repro.
Also, could you add that repro to the sanity tests?

ksivaman:
Added the shapes to `make_like` calls in `float8_tensor`, but I'm also keeping the original change to `quantized_tensor` since that is logically the correct shape to use in case the `tensor`'s and `data`'s shape do differ. We can remove the `data` kwarg completely from `quantized_tensor` which will require implementation of `make_like` in the `Float8Tensor` and `MXFP8Tensor`, so leaving that for a different PR.

PR #1521, opened 2025-02-28
[PyTorch] Remove Megatron-LM convergence test

timmoon10:
# Description

The L3 convergence test is more of a run script than an actual test and it just outputs logs that need to be manually inspected for problems. It also uses an idiosyncratic dataset. In the future we should think of a better approach for validating convergence, perhaps using an MLPerf dataset and actually failing if the learning quality or performance fall outside of some hard-coded bounds.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring
- [x] Testing

## Changes

- Remove Megatron-LM convergence test

# Checklist:

- [ x The functionality is complete


timmoon10:
Pipeline 24689192

ksivaman:
I think it's good to keep these tests/files until we have a concrete replacement ready. I agree that we should run these more often and have an automatic convergence check for the metrics, maybe val loss/ppl?

qa/L3_pytorch_convergence_test/test.sh
```diff
@@ -1,14 +0,0 @@
...
-set -e
-
-: ${TE_PATH:=/opt/transformerengine}
-
-pip install prettytable
-git clone https://github.com/NVIDIA/Megatron-LM.git
-cd Megatron-LM
-git checkout b3375a0e38c10e2300ef4be031f7dcabab52b448
-pytest -v -s $TE_PATH/tests/pytorch/distributed/test_convergence.py
-python $TE_PATH/tests/pytorch/distributed/print_logs.py
```
ksivaman, line 14: We could modify this to not just print but have a threshold for validation PPL. Currently the test is only functional aside from the manual checking of logs.

PR #1523, opened 2025-02-28
CP implementation refinement for BSHD/SBHD format

xrennvidia:
# Description

- fix recompilation of torch.compile, create separate functions for different input arguments.
- cache the computed `cu_seqlens`, so tiny `floor_divide` kernels are removed.
- do not explicitly initialize/allocate fwd out buffer, instead repurpose the returned tensor of the first out correction to be the out tensor.
- add `pad_between_seqs` to the API of transformer layer, MHA, and DPA. If provided, TE uses its value directly, otherwise TE infers its value as before.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:


transformer_engine/pytorch/attention.py
```diff
@@ -7183,6 +7169,7 @@ def forward(
         alibi_slopes: Optional[torch.Tensor] = None,
         fast_zero_fill: bool = True,
         inference_params: Optional[InferenceParams] = None,
+        pad_between_seqs: Optional[bool] = None,
```
cyanguwa, line 7172: Could you also add this argument to MultiheadAttention and TransformerLayer please? So that users can pass it in when they are using TransformerLayers. Thanks.

xrennvidia: Added.

cyanguwa (approved):
LGTM

This PR targets a few of the small kernels in CP, either fuses them via torch.compile, or caches their generation, to reduce CPU overhead for CP, especially the P2P path.

xrennvidia:
All CP tests passed, and the failed tests are not due to CP.

PR #1525, opened 2025-02-28
[PyTorch][MoE] Enable New Recipes for Grouped Linear

yaox12:
# Description

- [x] Enable MXFP8 for `GroupedLinear` by adding scaling factor swizzling.
- [x] Enable per-tensor current scaling for `GroupedLinear`.
- [x] Enable FP8 blockwise scaling for `GroupedLinear`.
- [x] Let per-tensor cast (without transpose) also go through the fused code path. 
- [x] Make FP8 padding/unpadding support customized padding dim since MXFP8 requires aligning to 32.

Rely on #1559 and #1638.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -203,6 +203,7 @@ def general_grouped_gemm(
             for o in out
         ]  # this should differ with respect to single output
 
+    original_scale_inverses_list = [swizzle_inputs(A[i], B[i], layout) for i in range(num_gemms)]
```
yaox12, line 206: A weird thing here is that when I attempted to swizzle the scales in C++, i.e., inserting two lines
```cpp
    swizzle_scaling_factors(te_A, transa);
    swizzle_scaling_factors(te_B, !transb);
```
after
https://github.com/NVIDIA/TransformerEngine/blob/97344d669670234de6517053b3eebefb05e3a40b/transformer_engine/pytorch/csrc/extensions/gemm.cpp#L325-L327
it will give incorrect results. I'm quite confusing.

ksivaman: @ptrendx This looks like the same issue I encountered as well

yaox12:
@ksivaman Can I have your review or approval? CI passed except for some irrelevant failures.

transformer_engine/common/include/transformer_engine/transpose.h
```diff
@@ -90,8 +90,8 @@ void nvte_fp8_transpose_dbias(const NVTETensor input, NVTETensor transposed_outp
  *                                     transposed.
  *  \param[in]     stream              CUDA stream used for the operation.
  */
-void nvte_multi_cast_transpose(size_t num_tensors, const NVTETensor* input_list,
-                               NVTETensor* output_list, cudaStream_t stream);
+void nvte_multi_quantize(size_t num_tensors, const NVTETensor* input_list, NVTETensor* output_list,
```
ptrendx, line 93: Can we still keep both functions and deprecate the old interface?

transformer_engine/pytorch/cpp_extensions/gemm.py
ptrendx: From a quick look into it - that's because swizzle_scaling_factors is not doing the right thing - it creates a pyTorch tensor for the new scaling factor but then lets it go out of scope releasing the memory.
```python
  auto swizzled_scale_inv = at::empty(scale_inv_shape_int, options);
```
This tensor should be returned by that swizzle function to make sure the memory is still there when the caller uses it.

transformer_engine/pytorch/csrc/extensions/quantizer.cpp
```diff
@@ -109,6 +109,7 @@ std::pair<TensorWrapper, py::object> Float8Quantizer::create_tensor(
   }
   const py::object py_columnwise_data = create_transpose ? py::cast(columnwise_data) : py::none();
   opts = opts.dtype(torch::kFloat32);
+  // TODO: Replace with an empty tensor.
```
ptrendx, line 112: Why can't we do that now?

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -264,6 +263,13 @@ def backward(ctx, grad_output: torch.Tensor) -> Tuple[Union[torch.Tensor, None],
                 accumulate_wgrad_into_param_main_grad = ctx.fuse_wgrad_accumulation
 
             if ctx.requires_dgrad:
+                dgrad_gemm_use_split_accumulator = _2X_ACC_DGRAD
+                if ctx.fp8:
+                    recipe = ctx.fp8_recipe
+                    if hasattr(recipe, "fp8_gemm_dgrad"):
+                        dgrad_gemm_use_split_accumulator = (
+                            recipe.fp8_gemm_dgrad.use_split_accumulator
+                        )
```
ptrendx, line 272: As noted in the current scaling recipe PR I don't like this, but leaving that for now for the subsequent refactor.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -416,6 +429,8 @@ class GroupedLinear(TransformerEngineBaseModule):
                   the model is trained with lower precision and the original FP32 parameters
                   would not fit in GPU memory.
 
+    Note: GroupedLinear doesn't really handle the TP communications inside. The `tp_size` and
+          `parallel_mode` are used to determine the shapes of weights and biases.
```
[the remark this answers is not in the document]
yaox12: Done.

transformer_engine/pytorch/csrc/extensions/quantizer.cpp
yaox12: I tried to do it but it seems the userbuffer and fp8 dpa rely on it. We probably need a dedicated PR to remove it.

transformer_engine/pytorch/cpp_extensions/gemm.py
yaox12: Thanks. That makes sense. I think we can put the fix in another PR.

transformer_engine/common/include/transformer_engine/transpose.h
yaox12: I'm not sure if I understand your question. Do you mean keeping `nvte_multi_cast_transpose` (with the equivalent behavior of `nvte_multi_quantize`) but marking it as deprecated?

I changed the name because the transpose is now optional, so naming it `*_cast_transpose` might be a bit confusing.

BestJuly:
Hi @yaox12, can we add some logic in `grouped_linear.py` so that if the input tensors are already quantized, we can avoid unnecessary processes. Otherwise, when `fp8=True` and tensors are already in `fp8`, then it will go through the following processes     1) cast fp8 to bf16, 2) cast bf16 back to fp8
```
        # Cast input to expected dtype
        inputmats_no_fp8 = [cast_if_needed(mat, activation_dtype) for mat in inputmats]
        inputmats = []
        ...
        if fp8:
            inputmats = tex.fused_multi_quantize(
                inputmats_no_fp8, None, input_quantizers, TE_DType[activation_dtype]
            )
            ...
        else:
            inputmats = inputmats_no_fp8
            ...
```
I am not sure whether such changes would be fine to be added in this PR or using an additional PR.

yaox12:
> Hi @yaox12, can we add some logic in `grouped_linear.py` so that if the input tensors are already quantized, we can avoid unnecessary processes. Otherwise, when `fp8=True` and tensors are already in `fp8`, then it will go through the following processes 1) cast fp8 to bf16, 2) cast bf16 back to fp8
> 
> ```
>         # Cast input to expected dtype
>         inputmats_no_fp8 = [cast_if_needed(mat, activation_dtype) for mat in inputmats]
>         inputmats = []
>         ...
>         if fp8:
>             inputmats = tex.fused_multi_quantize(
>                 inputmats_no_fp8, None, input_quantizers, TE_DType[activation_dtype]
>             )
>             ...
>         else:
>             inputmats = inputmats_no_fp8
>             ...
> ```
> 
> I am not sure whether such changes would be fine to be added in this PR or using an additional PR.

For now, I think `GroupedLinear` doesn't accept quantized tensor as the input. And it's not trivial to do it because we would need a standalone transpose kernel for the backward, which is missing for some recipes. I prefer to put it in another PR.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -663,17 +712,37 @@ def forward(
...
+        """Customize quantizers based on current scaling recipe + linear."""
+        assert (
+            recipe.float8_current_scaling()
+        ), "current scaling recipe quantizer customization here"
+        if fwd:
+            for i in range(self.num_gemms):
+                # set configs about amax epsilon and power_2_scale
+                self.quantizers["scaling_fwd"][
+                    self._offsets["input"] + i * self._num_fp8_tensors_per_gemm["fwd"]
+                ].force_pow_2_scales = recipe.fp8_quant_fwd_inp.power_2_scale
```
zhongbozhu, line 729: It seems that this pow2 scale config is not getting passed to the multi-quantize function? I think it's mostly because the `nvte_quantize` API lacks an input parameter for advanced knobs of numerics, which I added this commit: https://github.com/kwyss-nvidia/TransformerEngine/commit/dc86fd55cb5373316a38bbee714ad8af11dcfcb8 

Will push to get this new `nvte_quantize_v2` API merged in PR [1559](https://github.com/NVIDIA/TransformerEngine/pull/1559).

transformer_engine/common/transpose/multi_cast_transpose.cu
```diff
@@ -195,15 +197,25 @@ __global__ void __launch_bounds__(threads_per_block)
 
 }  // namespace
 
-void multi_cast_transpose(const std::vector<Tensor*> input_list, std::vector<Tensor*> output_list,
-                          cudaStream_t stream) {
+void multi_fp8_quantize(const std::vector<Tensor*> input_list, std::vector<Tensor*> output_list,
```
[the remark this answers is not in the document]
zhongbozhu: I am guessing this line takes care of the bf16 case, when scale_ptr=nullptr, scale=1: https://github.com/NVIDIA/TransformerEngine/blob/37a87b162b9f700cb13cf65aec9c3dc96d5f883e/transformer_engine/common/transpose/multi_cast_transpose.cu#L85

Then I suggest we don't make this naming for multi-quantize also evil like `nvte_quantize`.

transformer_engine/pytorch/csrc/extensions/transpose.cpp
```diff
@@ -6,27 +6,38 @@
...
 
   // create TE tensors from input
   for (size_t i = 0; i < input_list.size(); i++) {
-    auto input_tensor = makeTransformerEngineTensor(input_list[i], none);
+    auto input_tensor = makeTransformerEngineTensor(input_list[i]);
     const NVTEShape input_shape = input_tensor.shape();
 
     transformer_engine::TensorWrapper output_tensor;
 
+    if (!detail::IsFloat8Quantizers(quantizer_list[i].ptr())) {
```
zhongbozhu, line 38: Does it mean that we only have fused quantize kernel for delayed scaling?

mxfp8, fp8 current scaling and subchannel recipe will be redirected to the quantize api that is supposed to work just like calling tex.quantize in python layer?

transformer_engine/pytorch/csrc/extensions/transpose.cpp
```diff
@@ -48,37 +59,21 @@ std::vector<py::object> fused_multi_quantize(std::vector<py::handle> input_list,
...
-    nvte_multi_cast_transpose(nvte_tensor_input_list.size(), nvte_tensor_input_list.data(),
-                              nvte_tensor_output_list.data(), at::cuda::getCurrentCUDAStream());
+    nvte_multi_quantize(nvte_tensor_input_list.size(), nvte_tensor_input_list.data(),
+                        nvte_tensor_output_list.data(), at::cuda::getCurrentCUDAStream());
   } else {
-    for (size_t i = 0; i < nvte_tensor_output_list.size(); i++) {
-      nvte_quantize(nvte_tensor_input_list[i], nvte_tensor_output_list[i],
-                    at::cuda::getCurrentCUDAStream());
+    for (size_t i = 0; i < py_output_objects_list.size(); i++) {
+      quantize(input_list[i], quantizer_list[i], py_output_objects_list[i], std::nullopt);
```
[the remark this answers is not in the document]
yaox12: Yes.

transformer_engine/pytorch/module/grouped_linear.py
yaox12: For now, in `fused_multi_quantize`, only delayed scaling goes to the fused quantize kernel while other recipes will go to `tex.quantize`, where these pow2/epsilon configs are handled.

transformer_engine/common/transpose/multi_cast_transpose.cu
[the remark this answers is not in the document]
yaox12: I think bf16 are gated at the python level, for example, https://github.com/NVIDIA/TransformerEngine/blob/37a87b162b9f700cb13cf65aec9c3dc96d5f883e/transformer_engine/pytorch/module/grouped_linear.py#L142-L145
`*_quantize` are only called when fp8 is enabled.

transformer_engine/pytorch/csrc/extensions/transpose.cpp
yaox12: Yes, it is for now.

transformer_engine/common/transpose/multi_cast_transpose.cu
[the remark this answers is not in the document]
zhongbozhu: I think what you said makes sense, even though there is nothing preventing people from use a fused quantize kernel for bf16, we are not really exposing this possibility to the python layer. So basically, for users still interested in using bf16 for moe, the speedup of using group linear only comes from the grouped gemm that is now totally based on multi-stream implementation, right?

Just curious, what is your estimate for performance comparison between a sequential quantize kernel (no matter it's a cast to bf16 or mxfp8, subchannel) and a fused quantize kernel. Should we invest resource in the future to develop a fused_comp_amax or fused_subchannel_quantize?

zhongbozhu:
What is the mechanism to enable using group linear from TE instead of calling sequential linear in MegatronLM/Nemo, do we need to ping them for a small PR to toggle it open?

transformer_engine/common/transpose/multi_cast_transpose.cu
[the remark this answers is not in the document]
yaox12: We did observe some speedup from the fused quantize kernel for delayed scaling. The main speedup comes from reducing launch overheads of a few quantization kernels to one. There could be no speedup when the quantize kernels take a long time and the launch overheads are not exposed, and the performance of single tensor quantization could be better because it's JITed while the fused kernel is precompiled.

> Should we invest resource in the future to develop a fused_comp_amax or fused_subchannel_quantize?
I think we're facing similar situations as in delayed scaling. If launch overheads are not exposed, or the fused kernels are not as optimized as the single-tensor quantization kernels, we would not expect speedup. And there're so many variants of quantization kernels in TE now, turning them to fused versions could be a huge effort.

zhongbozhu:
Is this MoE feature going to automatically support FP8 primary weight when this https://github.com/NVIDIA/TransformerEngine/pull/1641 get merged?

tests/pytorch/test_numerics.py
```diff
@@ -1582,31 +1576,29 @@ def test_grouped_linear_accuracy(
 
 
 @pytest.mark.parametrize("parallel_mode", ["column", "row"])
-@pytest.mark.parametrize("recipe", fp8_recipes)
+@pytest.mark.parametrize("recipe", fp8_recipes + [None])
 def test_grouped_linear_accuracy_parallel_mode(parallel_mode, recipe):
```
zhongbozhu, line 1580: Why are we testing the parallel version of group linear when TP cannot really be larger than 1?

```
        if self.tp_size > 1 and bias:
            raise ValueError(
                "GroupedLinear doesn't support bias when TP > 1. "
                "Because the TP communication is handled outside of this module."
            )

        self.parallel_mode = parallel_mode
        assert (
            self.parallel_mode in GemmParallelModes
        ), f"parallel_mode {parallel_mode} not supported"

        if self.parallel_mode == "column":
            self.out_features = divide(self.out_features, self.tp_size)
        elif self.parallel_mode == "row":
            self.in_features = divide(self.in_features, self.tp_size)

        self.sequence_parallel = (self.tp_size > 1) and sequence_parallel
```

transformer_engine/pytorch/module/base.py
```diff
@@ -428,6 +428,12 @@ def __init__(self) -> None:
         self.fsdp_group = None
         self._fp8_workspaces: Dict[str, QuantizedTensor] = {}
         self.activation_dtype: Optional[torch.dtype] = None
+        # Max. number of fp8 tensors per GEMM = 3 (input, weight, output) for fwd and
+        # 2 (grad_output and grad_input) for bwd
+        self._num_fp8_tensors_per_gemm = {
```
zhongbozhu, line 433: This is a good idea, can we know fix this "magical number" code in `make_quantizers` of block scaling recipe here: https://github.com/NVIDIA/TransformerEngine/blob/a8f0fe03b4ac801ee50b35ff24ec2998eaa301ac/transformer_engine/pytorch/fp8.py#L997 ?

CC @timmoon10 @kwyss-nvidia

tests/pytorch/test_numerics.py
yaox12: It was to test we were handling the bias correctly. Yes, we can remove it now.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -617,20 +658,26 @@ def forward(
             grad_output_quantizers, _ = [None] * self.num_gemms, [None] * self.num_gemms
             if self.fp8:
                 input_quantizers = [
-                    self.quantizers["scaling_fwd"][self._offsets["input"] + i]
+                    self.quantizers["scaling_fwd"][
```
yaox12, line 661: This is a breaking change for delayed scaling. The checkpoint saved with older TE versions loaded with new TE versions could be wrong.

But it's necessary for blockwise scaling.

Should we add a switch for delayed scaling and others? cc @timmoon10

transformer_engine/common/transpose/multi_cast_transpose.cu
```diff
@@ -149,37 +148,40 @@ __global__ void __launch_bounds__(threads_per_block)
     }
   }
 
-  // Copy transposed output from registers to global memory
-  __shared__ OVecT shared_output_t[THREADS_PER_WARP][THREADS_PER_WARP + 1];
+  if (args.output_t_list[tensor_id] != nullptr) {
```
zhongbozhu, line 151: just to check my understanding of this code change: is it true that the only change to this MR is to conditionally skip the transpose op based on whether the transposed tensor list is given?

zhongbozhu:
LGTM!

transformer_engine/common/transpose/multi_cast_transpose.cu
yaox12: Yes. This is a change for delayed scaling on Blackwell, where transpose is not need.

kwyss-nvidia:
Thanks for putting this MR together. Looks solid. I have a few minor comments and questions.

transformer_engine/pytorch/module/base.py
timmoon10: We don't want this to become our convention. Different layers will have different numbers of quantizers (e.g. `Quantize` op forward has 1 quantizer, attention forward has 3 quantizers but no weights, `Linear` forward should have 2 quantizers). I prefer treating this as an implementation detail within the modules than to move it to the base class.

transformer_engine/pytorch/module/grouped_linear.py
timmoon10: I think we broke checkpoints in the 2.0 release, so it's a somewhat narrow window where this change would affect customers. This is fine with me.

timmoon10 (commented):
Overall LGTM. My biggest suggestions would be to restore `nvte_multi_cast_transpose` for backward compatibility and remove `_num_fp8_tensors_per_gemm` from the base module.

transformer_engine/common/include/transformer_engine/transpose.h
yaox12: This change was to let delayed scaling on Blackwell also go to this fused `multi_cast_transpose` kernel, but as you mentioned, the performance may not be good. So I reverted it.

transformer_engine/pytorch/module/base.py
yaox12: Moved it to `GroupedLinear`.

transformer_engine/common/transpose/multi_cast_transpose.cu
yaox12: Reverted it.

PR #1526, opened 2025-02-28
Fix installation from PyPI wheels after a source install

ksivaman:
# Description

When TE is installed from [PyPI](https://pypi.org/project/transformer-engine) in an environment where TE is already installed from source, the wheel installation is incorrect resulting in an application crash during runtime. In this case, the following workflow currently occurs on executing `pip install transformer-engine[pytorch]`:

1. pip collects packages `transformer-engine` (metapackage), `transformer-engine-cu12` (core package), `transformer-engine-torch` (framework extensions).
2. `transformer-engine-torch` and `transformer-engine-cu12` are correctly installed first as they are the dependencies. This places `libtransformer_engine.so` and `transformer_engine_torch*.so` under the `transformer_engine` directory which already exists in the described failing case (source installation).
3. Before installing `transformer-engine` (metapackage), it uninstalls the `transformer-engine` (source install) and wipes out the shared object files just placed in step 2.

## Type of change

- [x] Infra/Build change
- [x] Code refactoring

## Changes

The root cause for this bug is that the `transformer-engine` package has a different structure when installed from source or from the PyPI distribution. In this PR:

- Instead of `transformer_engine/`, the shared objects are stored in a different subdirectory depending on whether installing from source or PyPI (`transformer_engine/src_lib` and `transformer_engine/wheel_lib` respectively) so that they don't interfere with each other.
- All 3 `__init__.py` files (core, PyTorch, Jax) check both locations when loading shared object files. There isn't a conflict because pip uninstalls the source version when installing the metapackage wheel, and we manually remove TE's PyPI packages when installing from source.
- The above changes break Jax imports. To fix this, `tranformer_engine_jax` is loaded similar to `transformer_engine_torch` which is a cleaner and doesn't depend on the SO living inside the `transformer_engine` directory.

I've manually tested some combinations of installing from source (+inplace) and wheel in various orders for PyTorch and Jax which seem to work:

- source &rarr; PyPI.
- PyPI &rarr; source.
- editable build &rarr; PyPI.
- PyPI &rarr; editable build.

# Checklist:


transformer_engine/common/__init__.py
```diff
@@ -82,7 +82,19 @@ def _load_cudnn():
...
     """Load shared library with Transformer Engine C extensions"""
 
-    so_path = get_te_path() / "transformer_engine" / f"libtransformer_engine.{_get_sys_extension()}"
+    so_path = (
+        get_te_path()
+        / "transformer_engine"
+        / "wheel_lib"
+        / f"libtransformer_engine.{_get_sys_extension()}"
+    )
+    if not so_path.exists():
```
ptrendx, line 91: I understand that we are counting on both of these never being valid at the same time, but I would still like these lines to be swapped here (so source has precedence if it exists).

transformer_engine/jax/__init__.py
```diff
@@ -44,16 +47,23 @@ def _load_library():
...
     extension = _get_sys_extension()
     try:
-        so_dir = get_te_path() / "transformer_engine"
+        so_dir = get_te_path() / "transformer_engine" / "wheel_lib"
         so_path = next(so_dir.glob(f"{module_name}.*.{extension}"))
     except StopIteration:
-        so_dir = get_te_path()
-        so_path = next(so_dir.glob(f"{module_name}.*.{extension}"))
+        try:
+            so_dir = get_te_path() / "transformer_engine" / "src_lib"
```
ptrendx, line 54: same

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -17,17 +17,18 @@
 from jax.sharding import PartitionSpec, NamedSharding
 from jax import ffi
 
-from transformer_engine.jax.attention import CPStrategy, SequenceDescriptor
-
-from transformer_engine import transformer_engine_jax
-from transformer_engine.transformer_engine_jax import (
+import transformer_engine_jax
+from transformer_engine_jax import (
```
ptrendx, line 21: This is kind of unrelated to this PR but why do you need both of these lines?

ksivaman: There are other functions from `transformer_engine_jax` which are used directly instead of importing it separately in the next line, such as `transformer_engine_jax.get_fused_attn_backend(...)`.  Not sure the rationale behind doing both styles but we could unify it in a separate (pure refactor) PR.

transformer_engine/common/__init__.py
ksivaman: Done

ksivaman:
Pipeline 24862635

ptrendx (approved):

PR #1529, opened 2025-02-28
Blackwell devel commoverlap mlperftests

vasunvidia:
# Description

This PR extends current test for TP overlap layer to add more configurable parameters.

Fixes # (issue)

## Type of change

- [ x ] New feature (non-breaking change which adds functionality)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/distributed/run_layer_with_overlap.py
+++ b/tests/pytorch/distributed/run_layer_with_overlap.py
@@ -304,12 +304,12 @@ def _train(opts):
         opts.tp = WORLD_SIZE
 
     # Tensor dim overrides for tensors that do not require TP communication
-    if ops.in_features is not None:
+    if opts.in_features is not None:
         assert opts.layer_type is te.Linear and opts.linear_parallel_mode == "row", (
             "--in-features is only used to configure row-tensor-parallel Linear layers. Use"
             " --num-heads or --head-dim for other cases."
         )
-    if ops.out_features is not None:
+    if opts.out_features is not None:
         assert opts.layer_type is te.LayerNormLinear and opts.linear_parallel_mode == "column", (
             "--out-features is only used to configure column-tensor-parallel LayerNormLinear"
             " layers. Use --num-heads or --head-dim for other cases."
--- a/tests/pytorch/distributed/run_layer_with_overlap.py
+++ b/tests/pytorch/distributed/run_layer_with_overlap.py
@@ -267,11 +267,6 @@ def _train(opts):
         opts.tcp_init = True
         opts.bind_to_device = True
         opts.bootstrap_backend = "mpi"
-    elif "SLURM_PROCID" in os.environ:
-        WORLD_RANK = (int)(os.environ["SLURM_PROCID"])
-        LOCAL_RANK = (int)(os.environ["SLURM_LOCALID"])
-        WORLD_SIZE = (int)(os.environ["SLURM_NTASKS"])
-        LOCAL_SIZE = WORLD_SIZE
     else:
         WORLD_RANK = int(os.getenv("RANK", "0"))
         WORLD_SIZE = int(os.getenv("WORLD_SIZE", "1"))
--- a/tests/pytorch/distributed/run_layer_with_overlap.py
+++ b/tests/pytorch/distributed/run_layer_with_overlap.py
@@ -11,6 +11,7 @@ import subprocess
 import argparse
 import warnings
 import pprint
+import yaml
 
 import torch
 import torch.distributed as dist
@@ -46,7 +47,7 @@ def _get_layer_args(config, tp_group, tp_size, reference=False):
     input_shape = [config.seq_length, config.batch_size, hidden_size]
     args = [hidden_size]
     kwargs = {
-        "params_dtype": torch.float32,
+        "params_dtype": torch.float32 if not config.use_bf16_params else torch.bfloat16,
         "device": "cuda",
         "tp_group": tp_group,
         "tp_size": tp_size,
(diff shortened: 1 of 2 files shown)


ptrendx (approved):

tests/pytorch/distributed/run_layer_with_overlap.py
```diff
@@ -87,6 +94,13 @@ def _get_layer_args(config, tp_group, tp_size, reference=False):
         kwargs["ub_bulk_dgrad"] = not config.overlap_rs_dgrad and not reference
         kwargs["ub_bulk_wgrad"] = not config.overlap_rs_dgrad and not reference
 
+    if config.ub_cfg is not None and isinstance(config.ub_cfg, str):
+        try:
+            import yaml
+        except ImportError:
+            raise RuntimeError("'yaml' package not found!")
```
timmoon10, line 101: This is hacky.
```suggestion
```
Much better to add PyYAML as a test-time dependency: https://github.com/NVIDIA/TransformerEngine/blob/f3a009da7b577655fe3564982d3990b06aca53b1/setup.py#L109

vasunvidia: Addressed comment.

timmoon10 (approved):
LGTM, pending CI and fixing the Git history.

tests/pytorch/distributed/run_layer_with_overlap.py
```diff
@@ -210,6 +267,12 @@ def _train(opts):
         opts.tcp_init = True
         opts.bind_to_device = True
         opts.bootstrap_backend = "mpi"
+    elif "SLURM_PROCID" in os.environ:
+        WORLD_RANK = (int)(os.environ["SLURM_PROCID"])
+        LOCAL_RANK = (int)(os.environ["SLURM_LOCALID"])
+        WORLD_SIZE = (int)(os.environ["SLURM_NTASKS"])
+        assert opts.tp is not None, "SLURM runs should provide --tp argument."
```
timmoon10: If we make `--tp` a required arg, we need to set it in `test_comm_gemm_overlap.py`.

tests/pytorch/distributed/run_layer_with_overlap.py
```diff
@@ -223,9 +286,27 @@ def _train(opts):
...
+        tp_rank_list = list(range(tp_base_rank, tp_base_rank + LOCAL_SIZE))
+        new_group_kwargs = {
+            "backend": "nccl",
+            "ranks": tp_rank_list,
+        }
+    else:
+        opts.tp = WORLD_SIZE
+
+    assert opts.in_features is None or opts.layer_type is te.Linear
+    assert opts.out_features is None or opts.layer_type is te.LayerNormLinear
```
timmoon10, line 307: I don't really like how strange and unintuitive the requirements are for `--in-features`/`--out-features`. I see we do it this way because we initialize UB based on `hidden_size`, but we should at least add comments and error messages to make it less mysterious. If we keep this API, we should also enforce its requirements for the parallel mode.
```suggestion
    # Tensor dim overrides for tensors that do not require TP communication
    if ops.in_features is not None:
        assert opts.layer_type is te.Linear and opts.linear_parallel_mode == "row, "--in-features is only used to configure row-tensor-parallel Linear layers. Use --num-heads or --head-dim for other cases." 
    if ops.out_features is not None:
        assert opts.layer_type is te.LayerNormLinear and opts.linear_parallel_mode == "column, "--in-features is only used to configure column-tensor-parallel LayerNormLinear layers. Use --num-heads or --head-dim for other cases."
```

timmoon10 (changes requested):

tests/pytorch/distributed/run_layer_with_overlap.py
```diff
@@ -303,8 +303,11 @@ def _train(opts):
     else:
         opts.tp = WORLD_SIZE
 
-    assert opts.in_features is None or opts.layer_type is te.Linear
-    assert opts.out_features is None or opts.layer_type is te.LayerNormLinear
+    # Tensor dim overrides for tensors that do not require TP communication
+    if ops.in_features is not None:
+        assert opts.layer_type is te.Linear and opts.linear_parallel_mode == "row, "--in-features is only used to configure row-tensor-parallel Linear layers. Use --num-heads or --head-dim for other cases."
```
vasunvidia: "row, seems to be typo

timmoon10: 🤦

tests/pytorch/distributed/run_layer_with_overlap.py
```diff
@@ -210,6 +267,11 @@ def _train(opts):
         opts.tcp_init = True
         opts.bind_to_device = True
         opts.bootstrap_backend = "mpi"
+    elif "SLURM_PROCID" in os.environ:
+        WORLD_RANK = (int)(os.environ["SLURM_PROCID"])
+        LOCAL_RANK = (int)(os.environ["SLURM_LOCALID"])
+        WORLD_SIZE = (int)(os.environ["SLURM_NTASKS"])
+        LOCAL_SIZE = WORLD_SIZE
```
timmoon10, line 274: This breaks our CI pipelines. On SLURM systems we allocate a single node with `srun -N 1` (so `SLURM_NTASKS=1`) and then start distributed sessions with `torchrun`. When I look at the logs, it seems like we are just running these tests without any parallelism.
```suggestion
```
If we want to keep support for SLURM envvars, we should put them after the `torchrun` envvars.

timmoon10 (approved):

PR #1083, opened 2024-08-07
Update FP8 scale-inverse in kernels with FP8 output

timmoon10:
# Description

We currently treat the FP8 scale-inverse (the dequantization scaling factor) as part of the FP8 recipe, along with the FP8 scale (the quantization scaling factor) and the absmax history. However, this is uncomfortable because any change to the FP8 recipe will invalidate the corresponding FP8 data. We work around this by creating copies of the scale-invs whenever there might be a recipe update, e.g. in between the forward and backward passes of the linear layer:
https://github.com/NVIDIA/TransformerEngine/blob/6717554f11f9b8bd79f917560e525d538c95b3bc/transformer_engine/pytorch/module/linear.py#L318

This adds non-trivial CPU overhead (I estimate ~20% for the PyTorch linear layer forward pass on an L40).

A better approach is to treat the scale-inv as part of the FP8 data, something that should be output along with the FP8 bits and should never change independently of the FP8 bits. The FP8 recipe tells us how we want to cast into FP8, while the scale-inv tells us how to convert back to higher precision. Note that this generalizes nicely to block-scaling schemes, where the scale-inv tensor may be large and must be packaged with the data during communication.

This PR makes initial work toward this scheme by including scale-inv updates in most of the kernels with FP8 output: casting, activations, LayerNorm, RMSNorm. It doesn't seem that cuBLAS supports this, so I've added a small kernel that is launched after FP8 GEMMs. I have not attempted to propagate this change into Userbuffers or attention. I've also updated the PyTorch `Linear` and `LayerNormLinear` modules to avoid maintaining extra copies of the scale-inv and I see a 1.12x speedup in the `Linear` forward pass.

I'm a little apprehensive since this is technically a breaking change. Every time we generate FP8 values we will overwrite the FP8 recipe scale-inv. I have a hard time imagining why we would ever use a stale FP8 scale-inv though if the FP8 data has already been overwritten.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [x] Code refractor

## Changes

- Update FP8 scale-inverse in cast-transpose kernels
- Update FP8 scale-inverse in cast and activation kernels
- Update FP8 scale-inverse in LayerNorm and RMSNorm kernels
[...]

PR #1544, opened 2025-03-06
[PyTorch] Enable fp8_primary_weights for current scaling

kunlunl:
# Description

Enable fp8_primary_weights for current scaling

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

The core code of this PR is adding a `cast_master_weights_to_fp8` function in `transformer_engine/pytorch/tensor/utils.py`. It is used to update fp8 model weights using master weights, especially when the optimizer uses zero-1 algorithm. The other codes are basically exist to make this function work.

The reason I added this function is: Each recipe has its own logic to cast master weights to model weights. For example, in delay scaling, the amax reduction (between data parallel group) needs to be done after quantization; but in current scaling, the amax reduction needs to be done before quantization. And these will be more complicated if we add more recipes in the future. So I don't want Megatron or other users to handle these different logics for each recipe, instead, I want to provide a unified API to Megatron or other users, make them don't need to care about which recipe of these fp8 weights is used. They only need to pass in the model weights, master weights and other necessary parameters (explained below), and we can handle the recipes inside this function.

This `cast_master_weighs_to_fp8` function has 4 parameters: a list of model weights, a list of master weights, a list of start offsets, and a distributed group (typically the data parallel group). The 1st, 2nd and 4th parameters are simple, the 3rd parameter is: when zero-1 is enabled, the master weight of a model weight may be distributed across multiple ranks, so from the perspective of each rank itself, it has to know the starting offset of a master weight in its corresponding model weight, so that it can cast the master weight to appropriate slice of the corresponding model weight.

# Checklist:


transformer_engine/pytorch/module/linear.py
```diff
@@ -167,31 +167,29 @@ def forward(
         nvtx_range_pop(f"{nvtx_label}.input_cast_comm")
 
         # Cast weight to expected dtype
-        weightmat = weight
         if not fp8:
             weightmat = cast_if_needed(weightmat, activation_dtype)
         else:
-            if not isinstance(weight, QuantizedTensor):
```
ptrendx, line 174: Why removing this condition?

kunlunl: If weight is `QuantizedTensor`( i.e., fp8_primary_weights is enabled), I need to create a transpose cache during the first micro batch. This can only be done in forward pass because users (like megatron) many want to overlap weight all-gather and the forward pass. We don't know when the weight all-gather is completed, but we know it must has completed when the module forward() starts, then it's safe to create transpose cache.

In order to create transpose cache, I want to call update_usage() when the weight is `QuantizedTensor` and `is_first_microbatch` is `True`. I need add this update_usage() to all modules (linear, layernorm_linear, layernorm_mlp, etc.), so for convenience, I put this logic into `get_weight_workspace`. I think conceptually speaking it also makes sense because the function of `get_weight_workspace` is to transform weight to a fp8 workspace which is used in fp8 gemm, it should also work when weight is fp8 weight, in this case, `get_weight_workspace` return the weight itself directly (if it's the first micro batch, also calss update_usage() too (see https://github.com/NVIDIA/TransformerEngine/pull/1544/files#diff-2edfb64def69fb8a9d0aec02651a90981f759236844def110a8e2bb36289e9cdR995-R1001).

transformer_engine/pytorch/tensor/utils.py
```diff
@@ -0,0 +1,187 @@
...
+        multi_tensor_applier(
+            multi_tensor_scale, dummy_overflow_buf, [scales, packed_scale_views], 1.0
+        )
+        torch.reciprocal(packed_scales, out=packed_scales)
+        multi_tensor_applier(
+            multi_tensor_scale, dummy_overflow_buf, [packed_scale_views, scale_invs], 1.0
+        )
+
+
+def cast_master_weights_to_fp8_current_scaling(params, group):
```
zhongbozhu, line 132: What can we do in the current scaling code to help your design achieve SOL? I think you once mentioned a skip_compute_amax flag in the current scaling quantizers, right?

transformer_engine/pytorch/tensor/utils.py
```diff
@@ -0,0 +1,187 @@
...
+import torch.distributed as dist
+
+from .float8_tensor import Float8Quantizer, Float8CurrentScalingQuantizer
+from .mxfp8_tensor import MXFP8Quantizer
+
+from transformer_engine_torch import multi_tensor_scale
+from ..optimizers.multi_tensor_apply import multi_tensor_applier
+
+
+def cast_master_weights_to_fp8(model_weights, master_weights, start_offsets, group):
```
zhongbozhu, line 15: This function will be called from MegatronLM right? Is there a way to also pass in the current recipe so that we can know how to apply numerical options for compute_scale here. Eg. pow2scale for weight quantization (even though it's by default turned off, but it will be better if we still make it configurable when using fp8_model_init).

transformer_engine/pytorch/tensor/utils.py
kunlunl: The limitation of the current scaling quantizer is, it does compute_amax, amax_allreduce, compute_scale all in cpp code (https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/csrc/extensions/cast.cpp#L50-L69), so I cannot merge the allreduce of multiple fp8 weights amaxes into a single all-reduce operation. Theoretically, we should compute amaxes for all weights at first, put these amaxes into a contiguous buffer, and perform all-reduce on this contiguous buffer at once.

To achieve this, I think we need to:
1. Expose the "nvte_compute_amax" and "nvte_compute_scale_from_amax" to python level.
2. Inside python (the above `cast_master_weights_to_fp8_current_scaling` function), call nvte_compute_amax to compute amaxes for all weights and put them into a contiguous buffer.
3. Still inside the `cast_master_weights_to_fp8_current_scaling`, perform all-reduce on the contiguous amax buffer.
4. Compute scale_inv for every weights, using the reduced amaxes. (I saw that nvte_compute_scale_from_amax only compute scale, not scale_inv, so I also need to add scale_inv to the compute_scale_from_amax_kernel).
5. Add a "skip_compute_amax" flag to the current scaling quantizer, avoid overwriting the amaxes in quantize().

Another optional optimization is, we can fuse the multiple compute_amax kernel together, using similar method to multi_tensor_apply. This also applies to compute_scale_from_amax and quantize kernel. This can reduce the cpu overhead because we have many weights.

transformer_engine/pytorch/tensor/utils.py
zhongbozhu: I agree. It will be a good idea to pass in numerical options from recipe to quantizers during init.

transformer_engine/pytorch/tensor/utils.py
zhongbozhu: I think we better call this `nvte_compute_amax` function `nvte_multi_compute_amax` right? I think your plan is great, because by doing so, we can naturally have the support for group_linear used in MoE?

kunlunl: Yeah, if we do the optional optimization, it should be called as `nvte_multi_compute_amax`. But in this PR I didn't implement it. I implemented: 
1. Expose the `nvte_compute_amax` to python
2. `A multi_tensor_compute_scale_and_scale_inv` to merge all the scale and scale_inv calculations into one (or several, based on multi-tensor-applier) kernel.
3. Merge all amax all-reduce into one NCCL kernel
4. Add `with_computing_amax` and `with_computing_scale` flag to skip/or not skip the amax calculation and scale calculation inside `CurrentScalingQuantizer`

kunlunl:
@zhongbozhu @timmoon10 @ptrendx Could you help review this PR when you have time? Thanks in advance!! :)

transformer_engine/pytorch/tensor/utils.py
```diff
@@ -0,0 +1,266 @@
...
+        # master_weight may be smaller than model_weight because it could be distributed across
+        # multiple ranks. So we need to create a dummy weight using the raw data from model_weight.
+        shard_model_weight_raw = model_weight._data.view(-1)[start_offset:end_offset]
+        shard_model_weight_fp8 = quantizer.create_tensor_from_data(
+            shard_model_weight_raw.view(1, -1),
+            model_weight.dtype,
+        )
+
+        # Cast master weight to fp8.
+        quantizer.update_quantized(master_weight.view(1, -1), shard_model_weight_fp8)
```
zhongbozhu, line 261: just out of curiosity, we call `update_quantized` here with output buffer equal to `shard_model_weight_fp8` but we are not returning this `shard_model_weight_fp8` to the distributed optimizer, how does this newly quantized weight eventually arrive at each layer's weight?

tests/pytorch/distributed/run_cast_master_weights_to_fp8.py
```diff
@@ -0,0 +1,388 @@
...
+    """Get the underlying data of a quantized tensor, used in zero-1 optimizer"""
+    if isinstance(quantized_tensor, Float8Tensor):
+        assert hasattr(quantized_tensor, "_data"), "Float8Tensor does not have _data attribute"
+        assert quantized_tensor._data.dtype == torch.uint8, "Float8Tensor _data must be uint8"
+        return quantized_tensor._data
+    else:
+        raise ValueError(f"Unsupported quantized tensor type: {type(quantized_tensor)}")
+
+
+class MiniZero_1:
```
zhongbozhu, line 37: This is amazing!

tests/pytorch/distributed/run_cast_master_weights_to_fp8.py
```diff
@@ -0,0 +1,388 @@
...
+        # -----------------------------------------------------------------------------------------
+        # Step 5: Copy the updated weights (not all weights) to the weight buffer
+        # -----------------------------------------------------------------------------------------
+        for i in range(len(self.weights)):
+            master_weight = self.master_weights[i]
+            if master_weight is None:
+                continue
+            start_offset = self.start_offsets[i]
+            if isinstance(self.weights[i], QuantizedTensor):
+                weight = _get_raw_data(self.weights[i])
```
zhongbozhu, line 175: Ah I see, this explains my previous question.

transformer_engine/pytorch/tensor/utils.py
kunlunl: the `shard_model_weight_fp8` doesn't need to be returned. The input model weight is actually the weight of each layer, and because `shard_model_weight_fp8` is a slice of the model weight, so it has the same underlying gpu memory with model weight. So if we update `shard_model_weight_fp8`, the model weight will also be updated.

kunlunl: Here are what MCore currently do:
* https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/optimizer/distrib_optimizer.py#L2004-L2006
* https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/fp8_utils.py#L42-L60 (for TE2.0)
* https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/fp8_utils.py#L71-L80 (for TE1.x)
You can see it also update a slice of the original model weight. I just collect all the model weights and send them to TE, so we can handle different fp8 recipe here. Basically I'm moving the `quantize_param_fragment` of MCore to TE, so if we add more recipes in the future, we don't need to modify MCore anymore. I'm thinking the `quantize_param_fragment` maybe a better name.

tests/pytorch/distributed/run_cast_master_weights_to_fp8.py
kunlunl: Yes, see here: https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/fp8_utils.py#L47-L60

transformer_engine/pytorch/csrc/common.h
```diff
@@ -145,7 +145,9 @@ class Float8CurrentScalingQuantizer : public Quantizer {
   at::Tensor scale_inv;
   at::Tensor amax;
   DType dtype;
+  bool with_computing_amax;
   bool with_amax_reduction;
+  bool with_computing_scale;
```
timmoon10, line 150: If you need these options, it's a sign that you're doing something more advanced and CS-specific than the quantizer was designed for.
```suggestion
  bool with_amax_reduction;
```

transformer_engine/common/recipe/current_scaling.cu
```diff
@@ -135,7 +137,7 @@ void nvte_compute_amax(const NVTETensor input_, const NVTETensor output_, cudaSt
              "Output tensor for amax computation has invalid amax tensor  "
              "(expected FP32, got dtype=",
              to_string(output.amax.dtype), ")");
-  CheckOutputTensor(output, "output_compute_amax");
+  CheckOutputTensor(output, "output_compute_amax", allow_empty_output);
```
timmoon10, line 140: No need for `allow_empty_output`. We never use the data in `output` anyways.
```suggestion
  CheckOutputTensor(output, "output_compute_amax", true);
```

transformer_engine/pytorch/csrc/extensions/multi_tensor/multi_tensor_compute_scale.cu
```diff
@@ -0,0 +1,66 @@
...
+    for (int i_start = threadIdx.x; i_start < n && i_start < chunk_size; i_start += blockDim.x) {
+      float scale_val = transformer_engine::compute_scale_from_amax(amax[i_start], max_fp8,
+                                                                    force_pow_2_scales, epsilon);
+      scale[i_start] = scale_val;
+      transformer_engine::reciprocal(scale_inv + i_start, scale_val);
+    }
+  }
+};
+
+void multi_tensor_compute_scale_and_scale_inv_cuda(
```
timmoon10, line 58: This is a little problematic since we consider `scale_inv` as part of the FP8 data, so conceptually it should be updated in the FP8 cast kernel rather than with `scale` (see https://github.com/NVIDIA/TransformerEngine/pull/1083).

That said, the priority is avoiding a bunch of extra kernel launches in `cast_master_weights_to_fp8`. We could rework it to separate the `scale` and `scale_inv` updates, but I'm fine leaving this as an expedient hack.

timmoon10 (changes requested):
Moving the FP8 logic in Mcore distopt to TE does make a lot of sense. My biggest complaint is that we're pushing the FP8 CS quantizer beyond what makes sense. If we're manually computing the amax and doing the amax reduction, then it's easier to manually do the FP8 cast than to make the CS quantizer excessively configurable.

transformer_engine/pytorch/csrc/common.h
zhongbozhu: I think Kunlun needs these flags to skip amax calculation in the quantize function itself. The idea is that he could call the compute amax directly in his `cast_master_weights_to_fp8` for a list of master weights, then aggregate the amax reduction of them into one NCCL call. I cannot think of a better way to achieve it.

transformer_engine/pytorch/tensor/utils.py
```diff
@@ -0,0 +1,266 @@
...
+
+from .float8_tensor import Float8Quantizer, Float8CurrentScalingQuantizer
+from .mxfp8_tensor import MXFP8Quantizer
+
+import transformer_engine_torch as tex
+from transformer_engine_torch import multi_tensor_scale, multi_tensor_compute_scale_and_scale_inv
+from ..optimizers.multi_tensor_apply import multi_tensor_applier
+
+
+def cast_master_weights_to_fp8(model_weights, master_weights, start_offsets, group):
```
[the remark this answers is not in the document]
zhongbozhu: I think function can also extend to mxfp8 and other subchannel methods:
```
    if len(delayed_scaling_params) > 0:
        cast_master_weights_to_fp8_delayed_scaling(delayed_scaling_params, group)
    if len(current_scaling_params) > 0:
        cast_master_weights_to_fp8_current_scaling(current_scaling_params, group)
    # TODO: Add NV sub-channel support here.
```

transformer_engine/pytorch/tensor/utils.py
```diff
@@ -0,0 +1,266 @@
...
+            model_weight.dtype,
+        )
+
+        # Cast master weight to fp8.
+        quantizer.update_quantized(master_weight.view(1, -1), shard_model_weight_fp8)
+
+    if len(amaxes) > 0:
+        dummy_overflow_buf = torch.tensor([0], dtype=torch.int, device=amaxes[0].device)
+
+        # Reduce amaxes.
```
zhongbozhu, line 124: Based on offline discussion with @timmoon10 , this amax reduction is likely to duplicate the amax reduction happening inside `fp8_autocast_exit` function [here](https://github.com/NVIDIA/TransformerEngine/blob/96f9c6dedc72be13ab4374c4ab453690f4d7d072/transformer_engine/pytorch/fp8.py#L358).

transformer_engine/pytorch/tensor/utils.py
kunlunl: I pushed a commit to remove the `with_computing_amax` and `with_computing_scale` from CS quantizer. And also create a temporary Float8Quantizer to quantize master weights to shard model weights.

> Currently we're stretching the quantizer beyond what makes sense.

I agree. But as long as we want to aggregate the all-reduce of amax, I have to separate the calculation of amax and the quantization part in the CS quantizer. I don't have a good idea that can perfectly express this task yet. Do you think the above method (using Float8Quantizer instead of CS quantizer) is ok or I need to expose a function to directly cast master weights to model weights?

transformer_engine/pytorch/tensor/utils.py
kunlunl: Yes, so if I understand correctly, I don't need to modify this part.

Resolve this thread, please reopen it if I'm wrong.

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -301,6 +301,19 @@ def update_usage(
             f"{self.__class__.__name__} class does not implement update_usage function"
         )
 
+    def replace_raw_data(self, new_raw_data: torch.Tensor):
```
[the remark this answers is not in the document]
kunlunl: we indeed need this function, because mcore needs to put all weights in a contiguous buffer. I want to create an unified interface, so that when I add NV sub-channel support in the future, I don't need to touch MCore again. 

But I totally agree `replace_raw_data` is a very unclear function, especially for mxfp8. I have two ideas: 
* first, add this function only to Float8Tensor, not the base class. but I also need to modify the NV sub-channel tensor in the future if I choose this way, this makes me feel like I'm randomly adding methods to different classes without a unifying principle.
* second, write a `replace_raw_data` function in the `transformer_engine/pytorch/tensor/utils.py`, and handle all the case in this function. This is also a bit ugly, since this function will be coupled to the implementation details of each Tensor class. 

I choose the second way, because at least I can centralize the replace logic, until I find a better solution. I pushed a commit, so you can see the implementation of the new `replace_raw_data` function in `transformer_engine/pytorch/tensor/utils.py`.

Do you think this is acceptable?

transformer_engine/pytorch/tensor/utils.py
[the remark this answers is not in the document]
kunlunl: The task assigned to me also contains add NV sub-channel support. I plan to implement it inside this function, after the NV sub-channel recipe is merged.

Regarding to the name, I think both are fine.
But considering the zero-1 algorithm may not only needed by mcore, but also needed by other users, who directly use TE to build their framework, this function may be also very useful for them. Without this function, they have to handle different recipe by themselves. So I think its functionality is also general in some sense. Would you give it a chance without megatron in its name? maybe some other names like "quantize_fragment"?

transformer_engine/common/recipe/recipe_common.cuh
```diff
@@ -0,0 +1,56 @@
+/*************************************************************************
+ * Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+
+#ifndef RECIPE_COMMON_CUH_
+#define RECIPE_COMMON_CUH_
```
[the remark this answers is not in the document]
kunlunl: Done.

transformer_engine/common/recipe/current_scaling.cu
kunlunl: Done.

transformer_engine/pytorch/csrc/extensions/multi_tensor/multi_tensor_compute_scale.cu
kunlunl: Thanks!
So I resolved this thread. Please reopen it if necessary.

transformer_engine/pytorch/tensor/utils.py
kunlunl: This is not duplicated. The reduction group of fp8_autocast_exit is tp_group. We also need to do amax reduction among the dp group when we use zero-1.

zhongbozhu:
LGTM!

qa/L0_pytorch_unittest/test.sh
```diff
@@ -39,6 +39,9 @@ python3 -m pytest -v -s $TE_PATH/tests/pytorch/test_parallel_cross_entropy.py ||
 python3 -m pytest -v -s $TE_PATH/tests/pytorch/test_cpu_offloading.py || test_fail "test_cpu_offloading.py"
 NVTE_DEBUG=1 NVTE_DEBUG_LEVEL=1 python3 -m pytest -o log_cli=true --log-cli-level=INFO -v -s $TE_PATH/tests/pytorch/fused_attn/test_fused_attn.py || test_fail "test_fused_attn.py"
 NVTE_DEBUG=1 NVTE_DEBUG_LEVEL=1 python3 -m pytest -o log_cli=true --log-cli-level=INFO -v -s $TE_PATH/tests/pytorch/fused_attn/test_paged_attn.py || test_fail "test_paged_attn.py"
+python3 -m pytest -v -s $TE_PATH/tests/pytorch/test_fp8_model_init.py || test_fail "test_fp8_model_init.py"
+python3 -m pytest -v -s $TE_PATH/tests/pytorch/test_replace_raw_data.py || test_fail "test_replace_raw_data.py"
```
ksivaman, line 43: As these tests are purely functional, they don't warrant their own test files in my opinion. Let's add `Float8CurrentScaling` to the recipes list in the sanity tests, which already test FP8 model init and thus would achieve the same and more with much fewer changes.

transformer_engine/pytorch/tensor/utils.py
```diff
@@ -0,0 +1,254 @@
...
+from .quantized_tensor import QuantizedTensor
+from .float8_tensor import Float8Tensor, Float8Quantizer, Float8CurrentScalingQuantizer
+from .mxfp8_tensor import MXFP8Tensor, MXFP8Quantizer
+
+import transformer_engine_torch as tex
+from transformer_engine_torch import multi_tensor_scale, multi_tensor_compute_scale_and_scale_inv
+from ..optimizers.multi_tensor_apply import multi_tensor_applier
+
+
+def replace_raw_data(tensor: QuantizedTensor, new_raw_data: torch.Tensor):
```
ksivaman, line 16: Why can't we add this to `Float8Tensor`? This way, `tensor.replace_raw_data(data)` can be used and we can implement this method for whichever classes support this and add a generic `NotImplementedError` version in `QuantizedTensor`?

transformer_engine/pytorch/tensor/__init__.py
```diff
@@ -7,10 +7,13 @@
 import torch
 
 from .quantized_tensor import QuantizedTensor, Quantizer
+from .utils import cast_master_weights_to_fp8, replace_raw_data
 
 __all__ = [
     "QuantizedTensor",
     "Quantizer",
+    "cast_master_weights_to_fp8",
+    "replace_raw_data",
```
ksivaman, line 16: Don't like this name. It's too generic. Also not sure if both of these belong to the `__all__` here. Whichever files need this, why can't they be imported directly from `utils`?

ksivaman (changes requested):

transformer_engine/pytorch/fp8.py
```diff
@@ -526,18 +543,29 @@ def fp8_model_init(enabled: bool = True, recipe: Optional[Recipe] = None) -> Non
...
     recipe: transformer_engine.common.recipe.Recipe, default = `None`
             Recipe used to create the parameters. If left to None, it uses the default FP8 recipe.
+    preserve_high_precision_init_val: bool, default = `False`
+             when enabled, store the high precision tensor used to initialize FP8 parameters
+             in CPU memory, and add two function attributes named `get_high_precision_init_val()`
+             and `clear_high_precision_init_val()` to FP8 parameters to get/clear this high
+             precision tensor. The purpose is that users can use this high-precision copy
+             to initialize master weights, avoiding the loss of precision that can occur when
+             using FP8 parameters directly. Note that after the master weights are initialized,
+             users should call `clear_high_precision_init_val()` to release this CPU memory.
```
ksivaman, line 553: Question: Would it be better to combine the functionality of `get_high_precision_init_val` and `clear_high_precision_init_val` just for better management of memory? Once `get_high_precision_init_val` is called on a parameter (which I assume is only 1 time), it could return the high precision tensor on GPU and clear the CPU memory? This way the user is not responsible for clearing memory separately. Is there a reason this functionality is separate?

transformer_engine/pytorch/tensor/utils.py
kunlunl: I initially implemented it exactly the same as you said. I added a method function `replace_raw_data` to `QuantizedTensor`, and added the real implementation to Float8Tensor, also added `NotImplementedError` to MXFP8Tensor.

But Tim said this should not belong to Quantized tensor. The comment is here: https://github.com/NVIDIA/TransformerEngine/pull/1544#discussion_r2004195462

transformer_engine/pytorch/fp8.py
kunlunl: I separate it as there maybe scenarios that the high_precision_init_val needs to be used more than once. I have one case that in delay scaling (before TE2.0), not only the master weights need to be initialized with this high precision tensor, the first value of amax_history also need to be initialized with it. I found only in this way the loss can be bitwise matched with not using fp8_primary_weights.

Theoretically I can get it once in MCore, and when needed, pass it to the right place. But passing this value in MCore is painful, so I choose to make a `get` function and a `clear` function.

transformer_engine/pytorch/tensor/__init__.py
kunlunl: Do you think "quantize_to_fragment" is a proper name?

Yes, I can remove it from `__all__`

transformer_engine/pytorch/tensor/utils.py
timmoon10: https://github.com/NVIDIA/TransformerEngine/pull/1544#discussion_r2004195462

transformer_engine/pytorch/tensor/__init__.py
zhongbozhu: I think having "master weights" in the function name is actually very important. It will tell users that they should call it in the distributed optimizer.

qa/L0_pytorch_unittest/test.sh
kunlunl: About the fp8_model_init test, actually I'm testing the preserve_high_precision_init_val. So I guess it's not enough to just add current scaling to the recipes list?
I moved these two tests to test_sanity.py and rename the test_fp8_model_init to `test_fp8_model_init_high_precision_init_val`.

transformer_engine/pytorch/tensor/__init__.py
kunlunl: Pushed a commit to remove it from __all__

timmoon10 (approved):

PR #1548, opened 2025-03-07
Don't set data to null

ksivaman:
# Description

For pipeline parallelism, where multiple forwards may be run before a backward (for the same module), we cannot set the data to `None` as it won't be available for the next forward step.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- When preparing to save base tensors, don't set the data to `None`.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
+++ b/transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
@@ -105,8 +105,6 @@ class Float8TensorBase:
 
         """
         tensors = [self._data, self._transpose]
-        self._data = None
-        self._transpose = None
         return tensors, self
 
     def restore_from_saved(
--- a/transformer_engine/pytorch/tensor/_internal/mxfp8_tensor_base.py
+++ b/transformer_engine/pytorch/tensor/_internal/mxfp8_tensor_base.py
@@ -100,8 +100,6 @@ class MXFP8TensorBase:
 
         """
         tensors = [self._rowwise_data, self._columnwise_data]
-        self._rowwise_data = None
-        self._columnwise_data = None
         return tensors, self
 
     def restore_from_saved(


timmoon10 (commented):
We should think carefully since we keep going back and forth on this (see https://github.com/NVIDIA/TransformerEngine/pull/1500). Also, if we go for this change, we should update the function docstrings for `prepare_for_saving` since they're no longer correct.

Can you explain the bug a bit more? 
- If the input tensor is BF16, then it doesn't matter if we do multiple forwards since the FP8 data is internal state.
- If the input tensor is FP8, then I can see how we are messing with read-only data. We do have some logic to keep track of whether the FP8 data is internal state:
https://github.com/NVIDIA/TransformerEngine/blob/de06a34cebde85736ddb5cf9056c0db4c3db3465/transformer_engine/pytorch/module/linear.py#L127
- If the problem is coming from the weights, then there's some messiness. If the weights are in FP8 or we're using the FP8 weight cache, then we want the FP8 weights to be persistent over multiple forwards and we don't want to clear data. If the weights are in BF16 and we're casting to FP8 lazily, then it would be a memory optimization to clear data eagerly. Hm...

I propose that we make a separate function `clear_data` for clearing tensor data in `Float8TensorBase`/`MXFP8TensorBase`. We can then call this in `_Linear.forward` when permitted:
```python
tensors_to_save, tensor_objects = prepare_for_saving(
    saved_inputmat,
    weightmat,
    weight,
    bias,
)
if own_quantized_input:
    saved_inputmat.clear_data()
```

timmoon10 (approved):
Actually, on second thought I think this PR is fine. I think the actual bugfix in https://github.com/NVIDIA/TransformerEngine/pull/1500 comes from resetting `ctx.tensor_objects`. Clearing the tensor data after the forward pass is actually unnecessary since the buffers are already being stored by autograd.

ptrendx:
Wait wait wait, I don't think we understand the problem at all here. How would the data from the other microbatch affect this - it is a different tensor.
I believe the problem is instead in the weight workspace when the fp8_model_init is not used - this will create the weight workspace tensors as Base objects and so be subject to this data removal. Instead of not removing the data we should just make sure that the internal flag is used only for temporary tensors (and so usage of weight cache should disable it)

Issue #1491, opened 2025-02-18
`Float8Quantizer::create_tensor` calculates `scale_inv` instead of creating an empty buffer


yaox12:
https://github.com/NVIDIA/TransformerEngine/blob/b39397c541292f336c5964dd1661d80c08dc4c78/transformer_engine/pytorch/csrc/extensions/quantizer.cpp#L112

This brings some overheads. For example, in `fused_multi_quantize`, the `reciprocal` kernels take most of the overall time.

<img width="724" alt="Image" src="https://github.com/user-attachments/assets/e28d35dc-bafd-4c93-bcfc-917895be5732" />

There was an optimization that updates FP8 scale-inverse in kernels with FP8 output https://github.com/NVIDIA/TransformerEngine/pull/1083, why did we change it?

timmoon10:
Logically, `scale` is part of the recipe and `scale_inv` is part of the data:
- You can change `scale` at any time, for whatever reason, to any value. It is only relevant at the exact moment you are doing an FP8 cast.
- Changing `scale_inv` destroys FP8 data, just as if you overwrote the FP8 data itself. Also, consider that `scale_inv` generalizes to a tensor in MXFP8 and other block-scaling formats.

Before https://github.com/NVIDIA/TransformerEngine/pull/1083, we needed messy logic for the scale update after the forward pass: update `scale_inv` together with `scale` but also make a copy so that we don't destroy cached tensors for the backward pass. This added the CPU overhead of an extra kernel launch (`cudaMemcpy`) and allocating temporary PyTorch tensors.

After https://github.com/NVIDIA/TransformerEngine/pull/1083, we can decouple `scale` and `scale_inv`: populate `scale_inv` when performing the FP8 cast and update `scale` after the forward and backward passes. In most cases, this adds very little overhead to existing kernels (existing FP8 kernels output to `amax`, so just output to `scale_inv` as well). There are a few kernels that don't output `scale_inv` (e.g. cuBLAS) so we need a separate scale-inverse kernel, but that is no worse than the extra `cudaMemcpy` in the old implementation.

[...]

PR #1471, opened 2025-02-11
[PyTorch] Enabling Per-Tensor Current Scaling Recipe 

zhongbozhu:
# Description

[WIP] Enable per-tensor current scaling recipe, as an alternative to delayed scaling. 

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Commit history preserved (before squashing the commits): https://github.com/zhongbozhu/TransformerEngine/commits/zhongboz/backup_adding_cs_to_main_pr/

Please list the changes introduced in this PR:

- [x] Check in kernels and C++ wrappers needed for FP8 current scaling recipe
- [x] Wire the C++ interface to python level quantizer API
- [x] Current scaling recipe setup
- [x] Expose detailed numerical configs in quantizer API
- [x] Layer-level test finished, both single GPU and multi-GPU
- [x] Run E2E training 

Optional:
- [x] Zero tolerance with respect to golden values
- [x] Support current scaling also in layernorm_mlp & transformer module

#### C++ Unit Tests

```
# build C++ tests
TE_PATH=<your_path> ./qa/L0_cppunittest/test.sh
# rebuild C++ test cases
cmake --build tests/cpp/build -j
# list all test cases
ctest --test-dir tests/cpp/build --show-only
# test cast
ctest --test-dir tests/cpp/build -R OperatorTest/CastCSTestSuite.TestCastCS --output-on-failure
# test cast transpose
ctest --test-dir tests/cpp/build -R OperatorTest/CTCSTestSuite.TestCastTransposeCS --output-on-failure
```

#### Python Unit Tests
```
# test the quantizer
pytest tests/pytorch/test_float8tensor.py::TestCurrentScalingFloat8Tensor::test_quantize -s -v

# test the layer forward backward
pytest tests/pytorch/test_float8_current_scaling_exact.py::TestFP8CurrentScalingRecipeLinear::test_fp8_current_scaling_with_linear_module -s -v

pytest tests/pytorch/test_float8_current_scaling_exact.py::TestFP8CurrentScalingRecipeLayerNormLinear::test_fp8_current_scaling_with_layernorm_linear_module -s -v

# single GPU full functional test
[...]

PR #1554, opened 2025-03-10
[PyTorch] Support TP Overlap in Per-Tensor Current Scaling Recipe

BestJuly:
# Description

Enable TP/Comm Overlap for Per-Tensor Current Scaling Recipe.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- [x] Unit test related code changes
- [x] Module related code changes (i.e., linear, layernorm_linear, layernorm_mlp)
- [x] Quantization related code changes (help to use the correct `scale_inv` value)

## Unit Tests
Python Unit Tests
```
# Test Current Scaling only
pytest -s tests/pytorch/distributed/test_comm_gemm_overlap.py -k "FP8 and CURRENT" -v

# Test all overlapping cases
pytest -s tests/pytorch/distributed/test_comm_gemm_overlap.py -v

# Particular pattern test, take `LayernormLinear` module as an example
# Set UB_SKIPMC=1 if necessary
torchrun --nproc_per_node 2 tests/pytorch/distributed/run_layer_with_overlap.py --seq-length=1024 --batch-size=1 --num-heads=8 --head-dim=64 --layer-type=layernormlinear --fp8 --quantization=fp8_current_scaling
```

# Checklist:

### DIFF
--- a/tests/pytorch/distributed/run_layer_with_overlap.py
+++ b/tests/pytorch/distributed/run_layer_with_overlap.py
@@ -1,5 +1,4 @@
 #!/usr/bin/python3
-# UB_SKIPMC=1 torchrun --nproc_per_node 2 run_layer_with_overlap.py --seq-length=1024 --batch-size=1 --num-heads=8 --head-dim=64 --layer-type=linear --fp8 --fp8-recipe=tensorwise
 
 # Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
--- a/tests/pytorch/distributed/run_layer_with_overlap.py
+++ b/tests/pytorch/distributed/run_layer_with_overlap.py
@@ -365,7 +365,9 @@ def _train(opts):
     dist_print(f"Initialized default NCCL process group with {WORLD_SIZE} GPUs")
 
     # Initialize the Transformer Engine layer with overlap
-    args, kwargs, input_shape = _get_layer_args(opts, nccl_world, opts.tp, num_layers=opts.num_layers)
+    args, kwargs, input_shape = _get_layer_args(
+        opts, nccl_world, opts.tp, num_layers=opts.num_layers
+    )
     # Intialize userbuffers
     ub_cfgs = None
     if opts.overlap_rs_dgrad:
@@ -391,7 +393,9 @@ def _train(opts):
     dist.barrier()
 
     # Initialize the reference model and copy all parameters
-    ref_args, ref_kwargs, _ = _get_layer_args(opts, nccl_world, opts.tp, num_layers=opts.num_layers, reference=True)
+    ref_args, ref_kwargs, _ = _get_layer_args(
+        opts, nccl_world, opts.tp, num_layers=opts.num_layers, reference=True
+    )
     with te.fp8_model_init(enabled=opts.fp8_init):
         ref_model = multi_module_model(opts.layer_type, opts.num_layers, *ref_args, **ref_kwargs)
     dist_print("Initialized reference model...", debug=True)
--- a/tests/pytorch/distributed/run_layer_with_overlap.py
+++ b/tests/pytorch/distributed/run_layer_with_overlap.py
@@ -24,6 +24,22 @@ warnings.filterwarnings("ignore", category=FutureWarning)
 warnings.filterwarnings("ignore", category=UserWarning)
 
 
+class multi_module_model(torch.nn.Module):
+    def __init__(self, module, num_layers, *args, **kwargs):
+        super().__init__()
+        self.num_layers = num_layers
+        self.layers = torch.nn.ModuleList([module(*args, **kwargs) for _ in range(num_layers)])
+
+    def forward(self, x):
+        consistent_shape = x.shape
+        for layer in self.layers:
+            x = layer(x)
+            # For stacked layers, we need to reshape the output to the consistent shape
+            if self.num_layers > 1:
+                x = x.reshape(consistent_shape)
+        return x
+
+
 def _te_layer_argtype(name):
     te_layers = [
         te.Linear,
@@ -40,10 +56,14 @@ def _te_layer_argtype(name):
     return layer_map[name.lower()]
 
 
-def _get_layer_args(config, tp_group, tp_size, reference=False):
+def _get_layer_args(config, tp_group, tp_size, num_layers, reference=False):
     hidden_size = config.num_heads * config.head_dim
     ffn_hidden_size = 4 * hidden_size
     qkv_size = 3 * hidden_size
+    # Shape is fine if using TransformerLayer for stacked layers
+    # For other layer modules, we need to reshape the input to the consistent shape
+    if num_layers > 1 and config.layer_type != te.TransformerLayer:
+        qkv_size = ffn_hidden_size = hidden_size
     input_shape = [config.seq_length, config.batch_size, hidden_size]
     args = [hidden_size]
     kwargs = {
@@ -106,6 +126,9 @@ def _parse_args(argv=None, namespace=None):
         description="Test a Transformer Engine layer with GEMM+comm overlap via Userbuffers."
     )
     parser.add_argument("-l", "--layer-type", type=_te_layer_argtype, default=te.LayerNormMLP)
+    parser.add_argument(
+        "--num-layers", type=int, default=1, help="Number of identical layers to stack."
+    )
     parser.add_argument("-b", "--batch-size", type=int, default=2, help="Input batch size.")
     parser.add_argument("-s", "--seq-length", type=int, default=1024, help="Input sequence length.")
     parser.add_argument(
@@ -348,7 +371,7 @@ def _train(opts):
     dist_print(f"Initialized default NCCL process group with {WORLD_SIZE} GPUs")
 
     # Initialize the Transformer Engine layer with overlap
-    args, kwargs, input_shape = _get_layer_args(opts, nccl_world, opts.tp)
+    args, kwargs, input_shape = _get_layer_args(opts, nccl_world, opts.tp, num_layers=opts.num_layers)
     # Intialize userbuffers
     ub_cfgs = None
     if opts.overlap_rs_dgrad:
@@ -366,7 +389,7 @@ def _train(opts):
     )
 
     with te.fp8_model_init(enabled=opts.fp8_init):
-        test_model = opts.layer_type(*args, **kwargs)
+        test_model = multi_module_model(opts.layer_type, opts.num_layers, *args, **kwargs)
     dist_print("Initialized test model...", debug=True)
     if WORLD_RANK == 0:
         pprint.pprint(kwargs)
@@ -374,9 +397,9 @@ def _train(opts):
     dist.barrier()
 
     # Initialize the reference model and copy all parameters
-    ref_args, ref_kwargs, _ = _get_layer_args(opts, nccl_world, opts.tp, reference=True)
+    ref_args, ref_kwargs, _ = _get_layer_args(opts, nccl_world, opts.tp, num_layers=opts.num_layers, reference=True)
     with te.fp8_model_init(enabled=opts.fp8_init):
-        ref_model = opts.layer_type(*ref_args, **ref_kwargs)
+        ref_model = multi_module_model(opts.layer_type, opts.num_layers, *args, **kwargs)
     dist_print("Initialized reference model...", debug=True)
     for test_param, ref_param in zip(test_model.parameters(), ref_model.parameters()):
         with torch.no_grad():
--- a/tests/pytorch/distributed/run_layer_with_overlap.py
+++ b/tests/pytorch/distributed/run_layer_with_overlap.py
@@ -144,12 +144,12 @@ def _parse_args(argv=None, namespace=None):
         "--fp8", action="store_true", default=False, help="Enables the te.fp8_autocast() context."
     )
     parser.add_argument(
-        "--fp8-recipe", 
-        type=str.lower, 
-        default="delayed", 
-        choices=["delayed", "tensorwise"], 
-        help='Which fp8 recipe to use for FP8 tensors in the forward and backward pass',
-        dest='fp8_recipe'
+        "--fp8-recipe",
+        type=str.lower,
+        default="delayed",
+        choices=["delayed", "tensorwise"],
+        help="Which fp8 recipe to use for FP8 tensors in the forward and backward pass",
+        dest="fp8_recipe",
     )
     parser.add_argument(
         "--fp8-init", action="store_true", default=False, help="Initialize primary weights in FP8."
@@ -389,7 +389,9 @@ def _train(opts):
     # Fp8 recipe setup
     fp8_format = Format.HYBRID
     if opts.fp8_recipe == "delayed":
-        fp8_recipe = DelayedScaling(fp8_format=fp8_format, amax_history_len=32, amax_compute_algo="max")
+        fp8_recipe = DelayedScaling(
+            fp8_format=fp8_format, amax_history_len=32, amax_compute_algo="max"
+        )
     elif opts.fp8_recipe == "tensorwise":
         fp8_recipe = Float8CurrentScaling(fp8_format=fp8_format)
 
--- a/tests/pytorch/distributed/run_layer_with_overlap.py
+++ b/tests/pytorch/distributed/run_layer_with_overlap.py
@@ -143,12 +143,11 @@ def _parse_args(argv=None, namespace=None):
         "--fp8", action="store_true", default=False, help="Enables the te.fp8_autocast() context."
     )
     parser.add_argument(
-        "--fp8-recipe",
+        "--quantization",
         type=str.lower,
-        default="delayed",
-        choices=["delayed", "tensorwise"],
-        help="Which fp8 recipe to use for FP8 tensors in the forward and backward pass",
-        dest="fp8_recipe",
+        default="none",
+        choices=["none", "fp8_delayed_scaling", "fp8_current_scaling"],
+        help="Quantization recipe",
     )
     parser.add_argument(
         "--fp8-init", action="store_true", default=False, help="Initialize primary weights in FP8."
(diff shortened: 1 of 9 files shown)


transformer_engine/pytorch/csrc/extensions/quantizer.cpp
```diff
@@ -223,9 +223,8 @@ std::pair<TensorWrapper, py::object> Float8CurrentScalingQuantizer::create_tenso
   }
   const py::object py_columnwise_data = create_transpose ? py::cast(columnwise_data) : py::none();
 
-  //unlike delayed scaling, in current scaling, scale is not known, so scale_inv should be empty buffer
-  opts = opts.dtype(torch::kFloat32).device(torch::kCUDA);
-  at::Tensor scale_inv = at::empty(scale_inv_torch_shape, opts);
+  // In current scaling, scale is not known but we initialize it with 1 to avoid division by zero. If scale is already calculated, it can be correctly set.
+  at::Tensor scale_inv = at::reciprocal(scale);
```
timmoon10, line 227: In principle we don't know if the value of `scale` is valid at this point. It is a workspace buffer for an intermediate value during quantization, and we only expose it outside quantization to make debugging more convenient: https://github.com/NVIDIA/TransformerEngine/pull/1471#discussion_r1964410144

The actual bug is that the UB comm doesn't handle scales properly in `copy_into_buffer` and `get_buffer`. This is an indirect fix that assumes that UB is constructing the gathered tensor after it has called the quantizer. This might not always be true: pipeline parallelism involves multiple forward passes and then multiple backward passes, so `input_quantizer` might have been called multiple times in between a forward and backward pass.

I'm working on a more general bugfix, so I'm fine accepting this as a temporary fix.

tests/pytorch/distributed/run_layer_with_overlap.py
```diff
@@ -107,6 +107,14 @@ def _parse_args(argv=None, namespace=None):
...
         "--fp8", action="store_true", default=False, help="Enables the te.fp8_autocast() context."
     )
+    parser.add_argument(
+        "--fp8-recipe",
+        type=str.lower,
+        default="delayed",
+        choices=["delayed", "tensorwise"],
+        help="Which fp8 recipe to use for FP8 tensors in the forward and backward pass",
+        dest="fp8_recipe",
+    )
```
timmoon10, line 117: We should be able to generalize to MXFP8:
```suggestion
    parser.add_argument(
        "--quantization",
        type=str.lower,
        default="none",
        choices=["none", "fp8_delayed_scaling", "fp8_current_scaling"],
        help="Quantization recipe",
    )
```
See the test I've been using in my MXFP8 branch: https://github.com/timmoon10/TransformerEngine/blob/b8b325bb35a3ac233097a58a8657922f9c9611b5/tests/pytorch/distributed/test_fusible_ops_with_userbuffers.py#L485

transformer_engine/pytorch/csrc/extensions/quantizer.cpp
BestJuly: UB is initialized without quantization-related info. I think that is the reason why `copy_into_buffer` does not handle the scale part. Then `get_buffer` will use the scale information from quantizer, which will duplicate the calculation of `scale_inv`. Agreed that the current change here is indeed a temporary fix.

tests/pytorch/distributed/run_layer_with_overlap.py
BestJuly: Have made corresponding changes according to the suggestions.

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -207,6 +212,17 @@ def forward(
...
         nvtx_range_pop(f"{nvtx_label}.norm")
 
+        # For Float8CurrentScalingQuantizer, layernorm/rmsnorm has not been fused with quantizer.
+        # So the output of normalization is in high precision, and we need to quantize it to FP8 and put in the buffer.
+        if ub_overlap_ag_fprop and isinstance(input_quantizer, Float8CurrentScalingQuantizer):
+            if ub_bulk_dgrad:
+                # reduce duplicated transpose in `_fix_gathered_fp8_transpose`
+                input_quantizer.set_usage(rowwise=True, columnwise=False)
+            ub_obj_fprop = get_ub(ub_name + "_fprop")
+            ln_out_tmp = ln_out
```
zhongbozhu, line 222: maybe we should give it a better name

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -171,14 +172,18 @@ def forward(
                 if isinstance(input_quantizer, MXFP8Quantizer):
                     with_quantized_norm = False
             else:
-                input_quantizer.set_usage(
-                    rowwise=True,
-                    columnwise=backward_needs_input,
-                )
+                if fp8 and ub_bulk_dgrad:
+                    # reduce duplicated transpose in `_fix_gathered_fp8_transpose`
+                    input_quantizer.set_usage(rowwise=True, columnwise=False)
```
zhongbozhu, line 177: This optimization can also apply to function grad_output_preprocess in base.py, in which we did grad_output = quantizer(grad_output).

Since if we want to do allgather of grad_output under sequence + row parallel, we expect UB to do the allgather of grad_output (or dY), then we should also have an set_usage(rowwise=True, columnwise=False) before it. Otherwise we will also be wasting one at::empty() for malloc and one cast_transpose which could have been just cast.

The actual transpose of grad_output happens in linear/layernormlinear after we call the ub get_buffer.

                if isinstance(grad_output, QuantizedTensor):
                    # This is a no-op if platform supports non-TN FP8 GEMM or the transpose
                    # already exists.
                    grad_output.update_usage(rowwise_usage=True, columnwise_usage=True)

zhongbozhu: There is one thing that I discussed offline with @timmoon10  about this issue when we one day might add TPGEMM overlap to MXFP8.

For MXFP8, the scaling granularity is 1x32, therefore the "transpose" isn't just the data transpose, but also the scaling granularity change to 32x1. So, we might need to do two allgather ops in UB to overlap with GEMM.

In this case, we still need to set_usage(rowwise=True, columnwise=True) to make sure we have two copies of data to communicate.

So instead of doing if ub_bulk_dgrad:, maybe we should do

if ub_bulk_dgrad and is_per_tensor_recipe():
The is_per_tensor_scaling covers both delayed scaling and per-tensor current scaling. Both recipe should benefit from this optimization.

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -207,6 +212,17 @@ def forward(
         ln_out_return = ln_out if return_layernorm_output else None
         nvtx_range_pop(f"{nvtx_label}.norm")
 
+        # For Float8CurrentScalingQuantizer, layernorm/rmsnorm has not been fused with quantizer.
+        # So the output of normalization is in high precision, and we need to quantize it to FP8 and put in the buffer.
+        if ub_overlap_ag_fprop and isinstance(input_quantizer, Float8CurrentScalingQuantizer):
+            if ub_bulk_dgrad:
+                # reduce duplicated transpose in `_fix_gathered_fp8_transpose`
+                input_quantizer.set_usage(rowwise=True, columnwise=False)
```
zhongbozhu, line 220: why should we call `set_usage(rowwise=True, columnwise=False)` twice?

transformer_engine/pytorch/module/layernorm_linear.py
BestJuly: Use `ln_out_local` instead, which should be more accurate.

transformer_engine/pytorch/module/layernorm_linear.py
BestJuly: The `set_usage` will be indeed called several times. Currently, I disable duplicated transpose processing by another `set_usage` call. Otherwise, we need to add such expressions in a lot of conditions for each possible case. Not sure the overhead for multiple calling `set_usage`.

transformer_engine/common/recipe/__init__.py
```diff
@@ -77,6 +77,10 @@ def float8_current_scaling(self):
         """Whether the given recipe is (per-tensor) current scaling."""
         return isinstance(self, Float8CurrentScaling)
 
+    def per_tensor_scaling(self):
```
timmoon10, line 80: We are using this to check FP8 in addition to per-tensor scaling, so the name should reflect this:
```suggestion
    def float8_per_tensor_scaling(self):
```

timmoon10 (approved):
Overall LGTM

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -176,9 +176,13 @@ def forward(
...
                 )
 
+        # Reduce duplicated transpose in `_fix_gathered_fp8_transpose`
+        if fp8 and FP8GlobalStateManager.get_fp8_recipe().per_tensor_scaling() and ub_bulk_dgrad:
+            input_quantizer.set_usage(rowwise=True, columnwise=False)
+
         ub_obj_fprop = None
         ln_out = None
-        if ub_overlap_ag_fprop:
+        if ub_overlap_ag_fprop and not isinstance(input_quantizer, Float8CurrentScalingQuantizer):
```
zhongbozhu, line 185: We should leave a comment here to explain why we skip current scaling quantizer.

The reason I guess should be: for current scaling, we don't want layernorm to output fp8 but rather bf16, right?

transformer_engine/pytorch/module/linear.py
```diff
@@ -149,10 +149,14 @@ def forward(
                     quantizer=input_quantizer,
                 )
             else:
-                input_quantizer.set_usage(
-                    rowwise=True,
-                    columnwise=backward_needs_input,
-                )
+                if ub_bulk_dgrad:
+                    # reduce duplicated transpose in `_fix_gathered_fp8_transpose`
+                    input_quantizer.set_usage(rowwise=True, columnwise=False)
```
zhongbozhu, line 154: we should also add a per_tensor_scaling check here in the if condition?

transformer_engine/pytorch/module/linear.py
```diff
@@ -424,6 +429,7 @@ def backward(ctx, grad_output: torch.Tensor) -> Tuple[Union[torch.Tensor, None],
                     ub_obj_dgrad = ctx.ub_obj_gradout
                     ub_type_dgrad = tex.CommOverlapType.AG
                     ub_obj_dgrad.copy_into_buffer(inputmat, ctx.input_quantizer, local_chunk=True)
+                    inputmat = ub_obj_dgrad.get_buffer(ctx.input_quantizer)
```
zhongbozhu, line 432: why do we need this line?

We are using inputmat_total for wgrad calculation, and user buffer will handle the allgather of inputmat to get input_total, so we already have a line like below that fetches the gathered inputmat.
```
            wgrad = None
            if ctx.requires_wgrad:
                if ctx.ub_bulk_dgrad:
                    inputmat_total = ub_obj_dgrad.get_buffer(ctx.input_quantizer)
```

transformer_engine/common/recipe/__init__.py
BestJuly: Thanks for pointing out, have made the corresponding changes.

transformer_engine/pytorch/module/layernorm_linear.py
BestJuly: Have added comments to address this.

transformer_engine/pytorch/module/linear.py
BestJuly: Done.

transformer_engine/pytorch/module/linear.py
BestJuly: Good catch. Removing this line does not affect any results. Have deleted this line.

zhongbozhu:
Done some tests myself, LGTM

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -207,6 +217,14 @@ def forward(
         ln_out_return = ln_out if return_layernorm_output else None
         nvtx_range_pop(f"{nvtx_label}.norm")
 
+        # For Float8CurrentScalingQuantizer, layernorm/rmsnorm has not been fused with quantizer.
+        # So the output of normalization is in high precision, and we need to quantize it to FP8 and put in the buffer.
+        if ub_overlap_ag_fprop and isinstance(input_quantizer, Float8CurrentScalingQuantizer):
+            ub_obj_fprop = get_ub(ub_name + "_fprop")
+            ln_out_local = ln_out
+            ln_out = ub_obj_fprop.get_buffer(input_quantizer, local_chunk=True)
+            input_quantizer.quantize(ln_out_local, out=ln_out)
```
zhongbozhu, line 226: Hmmm, I think I still have questions about this line, and we need to figure this out before we merge. 

This line (plus another one in layernorm_mlp) are the only places that you call `get_buffer` first and then use `quantizer.quantize(inp, out)` to do the quantize. This pattern will imply that the output tensor's dataptr is managed by user buffer system. 

It's nice that this might save us some time copies data around, but I am worried that in actual E2E training, is it okay to save such tensor for backward if the dataptr belongs to ubuf? Shouldn't the data be compromised if we move to the next layer. 

Unfortunately, our layer level unit test cannot catch it so I am not sure how can I prove my point. @timmoon10 what do you think?

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -263,6 +273,14 @@ def forward(
 
         ln_out_return = ln_out if return_layernorm_output else None
 
+        # For Float8CurrentScalingQuantizer, layernorm/rmsnorm has not been fused with quantizer.
+        # So the output of normalization is in high precision, and we need to quantize it to FP8 and put in the buffer.
+        if ub_overlap_ag and isinstance(fc1_input_quantizer, Float8CurrentScalingQuantizer):
+            ub_obj_lnout = get_ub("fc1_fprop")
+            ln_out_local = ln_out
+            ln_out = ub_obj_lnout.get_buffer(fc1_input_quantizer, local_chunk=True)
+            fc1_input_quantizer.quantize(ln_out_local, out=ln_out)
```
zhongbozhu, line 282: And here is the other one.

transformer_engine/pytorch/module/layernorm_linear.py
timmoon10: Good catch, I think this is a bug from the 2.0 release. Before that, the GEMM function had an option to copy values from the UB buffer into a plain PyTorch tensor: https://github.com/NVIDIA/TransformerEngine/blob/c9ea6be92948e1ec553037f1a04900617b9f7f6b/transformer_engine/pytorch/module/layernorm_linear.py#L258
We didn't handle this when updating to 2.0, and I agree that it seems likely that `ln_out` is corrupted in between the forward and backward pass.

zhongbozhu: yes

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -176,11 +176,23 @@ def forward(
...
+            input_quantizer.set_usage(rowwise=True, columnwise=False)
+
         ub_obj_fprop = None
         ln_out = None
-        if ub_overlap_ag_fprop:
+        # For DelayScaling, output of normalization will be in fp8.
+        # For Float8CurrentScaling, we want the output of normalization in high precision, then quantize to fp8.
+        if ub_overlap_ag_fprop and not fp8:
             ub_obj_fprop = get_ub(ub_name + "_fprop")
             ln_out = ub_obj_fprop.get_buffer(input_quantizer, local_chunk=True)
```
zhongbozhu, line 193: Why are we making a difference between bf16 case and fp8 case?

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -176,11 +176,23 @@ def forward(
...
+        elif ub_overlap_ag_fprop and not isinstance(input_quantizer, Float8CurrentScalingQuantizer):
+            ln_out = input_quantizer.make_empty(inputmat.shape, dtype=inputmat.dtype, device="cuda")
```
zhongbozhu, line 195: This elif used to belong to fp8 delayed scaling right? For delayed scaling, we still want layernorm to directly output to an fp8 buffer, only difference is whether this fp8 buffer is allocated by pytorch or ubuf.

Here if we just allocate with inputmat.dtype, are we wasting some memory? Since we were only supposed to allocate the size enough for a fp8 tensor.

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -357,7 +385,7 @@ def forward(
                 weight,
                 bias,
                 ln_weight,
-                ln_out,
+                ln_out.clone() if ub_overlap_ag_fprop else ln_out,  # avoid saving a UB buffer
```
zhongbozhu, line 388: I think this fix alone is enough to fix it?

I mean the only issue we had was about saving ub tensor for backward, so why don't we just clone it in the end so we didn't have to allocate first and then copy into buffer. 

I think either fix is okay but I don't think we should do both. @timmoon10  what's your opinion?

transformer_engine/pytorch/module/layernorm_linear.py
BestJuly: Because for bf16 case, ub can be still used so I do not change the logic here. 
The main fix is about that the return value is changed to `ln_out.clone() if ub_overlap_ag_fprop else ln_out`.

zhongbozhu:
Cannot find any more issues, guess we are good to go? @timmoon10

timmoon10:
This fixes some `TransformerLayer` UB test failures (by capping to TP=4), but it introduces some other `TransformerLayer`  UB test failures (numerical error slightly beyond bounds with FP8 current scaling).

This PR has a critical bugfix (see https://github.com/NVIDIA/TransformerEngine/pull/1554#discussion_r1992417405), so I think it's better to merge quickly and fiddle with the test tolerances later.

transformer_engine/pytorch/csrc/extensions/quantizer.cpp
yaox12: > I'm working on a more general bugfix

Hi @timmoon10, are you still working on this? I recently find that these `at::reciprocal` calls significantly harm the performance, especially in the `GroupedLinear` and MoE models, just like this [issue](https://github.com/NVIDIA/TransformerEngine/issues/1491).

PR #1565, opened 2025-03-12
[JAX] Unbalanced Context Parallelism with THD format

zlsh80826:
# Description

https://github.com/NVIDIA/TransformerEngine/pull/1454 adds the THD + context parallelism with the load balancing which should have the best perf. This PR is to add the non-balancing implementaiton, which doesn't need to reorder the tokens before and after the attention.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Support unbalanced context parallelism with THD format
- Add related unit tests

# Checklist:


transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -2214,21 +2215,31 @@ def scan_kv_block(idx, carry):
...
+                        kv_segment_ids,
+                        q_segment_pos,
+                        kv_segment_pos,
+                        config,
+                    )
+
+                causal_padding_config = subblock_config
+                padding_config = replace(subblock_config, attn_mask_type=AttnMaskType.PADDING_MASK)
+
+                output_per_step, softmax_aux_per_step, _ = lax.cond(
```
mgoldfarb-nvidia, line 2239: If we support regular padding mask w/ non load balanced then I think this condition needs to change slightly since in that case we'd apply padding mask for all steps not just the diagonals.

PR #1593, opened 2025-03-19
Fix mxfp8 columnwise data missing

guyueh1:
…raining

# Description

When we use mxfp8 to first run a few validation steps and then run a training step, the weight columnwise data in the row-wise linear layer is missing. This PR fixes it.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/base.py
+++ b/transformer_engine/pytorch/module/base.py
@@ -957,12 +957,9 @@ class TransformerEngineBaseModule(torch.nn.Module, ABC):
         out = None
         if cache_name is not None:
             out = self._fp8_workspaces.get(cache_name, None)
-            if (out is not None and 
-                (
-                    (out._rowwise_data is None and quantizer.rowwise_usage) \
-                    or \
-                    (out._columnwise_data is None and quantizer.columnwise_usage)
-                )
+            if out is not None and (
+                (out._rowwise_data is None and quantizer.rowwise_usage)
+                or (out._columnwise_data is None and quantizer.columnwise_usage)
             ):
                 out = None
                 del self._fp8_workspaces[cache_name]
--- a/transformer_engine/pytorch/module/base.py
+++ b/transformer_engine/pytorch/module/base.py
@@ -957,6 +957,15 @@ class TransformerEngineBaseModule(torch.nn.Module, ABC):
         out = None
         if cache_name is not None:
             out = self._fp8_workspaces.get(cache_name, None)
+            if (out is not None and 
+                (
+                    (out._rowwise_data is None and quantizer.rowwise_usage) \
+                    or \
+                    (out._columnwise_data is None and quantizer.columnwise_usage)
+                )
+            ):
+                out = None
+                del self._fp8_workspaces[cache_name]
 
         # Gather cached Fp8 workspace if it's distributed
         # NOTE: FSDP sharding is supported only for Fp8 buffers and will not work
--- a/transformer_engine/pytorch/module/base.py
+++ b/transformer_engine/pytorch/module/base.py
@@ -957,7 +957,7 @@ class TransformerEngineBaseModule(torch.nn.Module, ABC):
         out = None
         if cache_name is not None:
             out = self._fp8_workspaces.get(cache_name, None)
-            if out is not None and (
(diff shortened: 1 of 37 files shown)


ksivaman (approved):
LGTM

guyueh1:
The previous change didn't cover all buggy cases. I had to change it like now.
Basically, the fp8_workspaces will be updated if it found that for the same name, the tensor usage is different from what the quantizer specifies.
I think this needs more careful review, cc @ksivaman @ptrendx

pggPL:
The problem is when we change quantizer use/recipe during training. For example in your case - validation does not need gradient, but training needs. We plan to handle switching the recipe, but it is not done yet.

But I'm quite confused how you get this error. In first training step you should run with `is_first_microbatch` set to `True` or `None` and it should result with updating the cache with correct tensor.

guyueh1:
> The problem is when we change quantizer use/recipe during training. For example in your case - validation does not need gradient, but training needs. We plan to handle switching the recipe, but it is not done yet.
> 
> But I'm quite confused how you get this error. In first training step you should run with `is_first_microbatch` set to `True` or `None` and it should result with updating the cache with correct tensor.

To give you an example that illustrate the bug:

We first run 1 microbatches of validation, then we run 1 microbatches of training,
1. 1st validation iteration: is_first_microbatch = True, update_workspace = True, cache_name = "weight", quantizer.rowwise_usage=True, quantizer.columnwise_usage=False
2. 1st training iteration: is_first_microbatch = True, update_workspace = True, cache_name = "weight", quantizer.rowwise_usage=True, quantizer.columnwise_usage=True
After 1, `self._fp8_workspace['weight']` will save a quantized mxfp8 tensor with no columnwise data. Then after 2, theoretically we should update the workspace object in `self_fp8_workspace['weight']` in-place in [line 1000](https://github.com/NVIDIA/TransformerEngine/blob/3cbb861bfe0468a5d5f82bd0d0f512f61caa4b41/transformer_engine/pytorch/module/base.py#L1000), here we should update `out` to have both rowwise and columnwise data, but in practice this didn't happen, after `tex.quantize(...)` the out still don't have columnwise data, thus raised the error.

ptrendx:
Right, the problem in that line 1000 is that it gives the out parameter to the tex.quantize function, which then just assumes that it is the right output and does not check what the quantizer said. There are 2 possibilities here, the tex.quantize should realize that the provided output is not actually correct and:
 - either add the missing fields
 - or error out
The second option would require the caller to provide the right output (probably by creating a new tensor), which would incur some slight overhead. The first option would require us to actually go over the functions (since tex.quantize is not the only one that outputs quantized tensors) and modify them all to be able to properly handle this. I'm inclined to go with the first option for the perf reasons (and will implement it), but wanted to hear your opinions as well @guyueh1 @ksivaman @timmoon10 @pggPL.

ksivaman:
I agree that functions like `tex.quantize` that produce fp8 output should give precedence to the quantizer and not what exists in the provided output tensor, so that's better for the longer term. For now will merge is CI is clean

transformer_engine/pytorch/module/base.py
```diff
@@ -999,6 +999,16 @@ def get_weight_workspace(
...
+            if (
+                quantizer is not None
+                and isinstance(out, MXFP8Tensor)
+                and (
+                    (out._rowwise_data is None and quantizer.rowwise_usage)
+                    or (out._columnwise_data is None and quantizer.columnwise_usage)
+                )
+            ):
+                out = None
+                del self._fp8_workspaces[cache_name]
```
timmoon10, line 1011: Linter warnings:
```suggestion
            if quantizer is not None and isinstance(out, MXFP8TensorBase):
                if quantizer.rowwise_usage and out._rowwise_data is None:
                    out = None
                    del self._fp8_workspaces[cache_name]
                elif quantizer.columnwise_usage and out._columnwise_data is None:
                    out = None
                    del self._fp8_workspaces[cache_name]
```

guyueh1:
L0 test seems to be failing for irrelevant things (paged attention), how should we fix it? @timmoon10

PR #1574, opened 2025-03-13
Fix issues in fused_attn_bwd

xrennvidia:
# Description

- Fix dtypes in `fused_attn_bwd` for CP, and make them consistent with the similar non-CP fix in #1566 . This helps solve some convergence issues with CP+P2P and CP+A2A, with FP8 attention.
- Clone fwd `scale` for the recomputation in `fused_attn_bwd`: for non-CP, clone `S`'s scale only; for CP, clone `QKV`, `O` and `S`'s scales
- Fix the amax update in CP+P2P

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1597, opened 2025-03-20
[PyTorch] Preserve FP8 scaling factors for attention backward

cyanguwa:
# Description

Since TE 2.0, the reduction of `q`, `k`, `v` scaling factors happens automatically with `QKV_quantizer`. If the forward and backward calls of `FusedAttnFunc` take place in two different `fp8_autocast` contexts, these scaling factors get updated in between. We need the unchanged scaling factors for the backward. This PR does that.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Save a clone of FP8 `q`, `k`, `v` tensors instead of letting their scaling factors be updated.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/attention.py
+++ b/transformer_engine/pytorch/attention.py
@@ -4676,7 +4676,7 @@ class FusedAttnFunc(torch.autograd.Function):
                 if is_output_fp8:
                     out_save = out_fp8.dequantize()
 
-            fp8_tensors = (q_fp8, k_fp8, v_fp8, out_fp8)
+            fp8_tensors = (q_fp8.clone(), k_fp8.clone(), v_fp8.clone(), out_fp8.clone())
         else:
             # q, k, v, out_ret: torch.float16 or torch.bfloat16
             out_ret, aux_ctx_tensors = fused_attn_fwd(


timmoon10 (commented):
Why do we need to preserve the scaling factors?
- If we are dequantizing these tensors from FP8, we just need the scale-invs. If something is changing the scale-invs, that's a bug just like if it overwrote the FP8 data (see https://github.com/NVIDIA/TransformerEngine/pull/1083).
- If we are quantizing new tensors into FP8, why does it matter if the scales are different?

xrennvidia:
> Why do we need to preserve the scaling factors?
> 
> * If we are dequantizing these tensors from FP8, we just need the scale-invs. If something is changing the scale-invs, that's a bug just like if it overwrote the FP8 data (see [Update FP8 scale-inverse in kernels with FP8 output #1083](https://github.com/NVIDIA/TransformerEngine/pull/1083)).
> * If we are quantizing new tensors into FP8, why does it matter if the scales are different?

Hi @timmoon10 
You are right, scale_inv is correct, the problem is scale. Flash Attn bprop does recompute which needs forward scaling factor of S [(refer here)](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/common/fused_attn/fused_attn_fp8.cu#L2882).

I am making the fix in another PR, we will not need this PR any more. We need to do something like
`ctx.S_quantizer = S_quantizer.copy()`
`ctx.S_quantizer.scale = S_quantizer.scale.clone()`

xrennvidia:
[PR 1574](https://github.com/NVIDIA/TransformerEngine/pull/1574) have fixed the issue, so close the PR.

PR #1609, opened 2025-03-25
[PyTorch] Fix issues for MCore DDP in grouped GEMM.

BestJuly:
# Description

[PR1474](https://github.com/NVIDIA/TransformerEngine/pull/1474) has fixed some issues in some modules. However, the logic in `grouped_linear.py` does not change. This PR is to add the fix to grouped GEMM, otherwise, then training MoE models with grouped GEMM, the following error will occur:
```
...
[rank0]: param.grad is not None 
[rank0]: AssertionError: param.grad being None is not safe when overlap_grad_reduce is True
```
The root cause is that the saved weight for backward usages does not have attribute `grad_added_to_main_grad`.

I also change the name from `w` to `weight` so that it can be consistent with the one in layernorm_linear.py.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Save `origin_weight` in forward pass and use it in the corresponding part since it has params attributes `grad_added_to_main_grad`

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/grouped_linear.py
+++ b/transformer_engine/pytorch/module/grouped_linear.py
@@ -78,8 +78,8 @@ class _GroupedLinear(torch.autograd.Function):
         skip_fp8_weight_update,
         *weights_and_biases,
     ) -> torch.Tensor:
-
         # pylint: disable=missing-function-docstring
+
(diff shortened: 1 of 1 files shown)


ksivaman (approved):
LGTM

PR #1614, opened 2025-03-25
[Pytorch] NVIDIA-DL-Framework-Inspect support – part 1 – core

pggPL:
# Description

Core part of NVIDIA-DL-Framework-Inspect support - which affects the core TE files.

# Checklist:


transformer_engine/pytorch/module/linear.py
```diff
@@ -1106,13 +1126,29 @@ def forward(
             else:
                 bias_tensor = None
 
+            quantizers = (
+                self._get_quantizers(fp8_output, fp8_grad)
+                if not TEDebugState.debug_enabled
+                else self._get_debug_quantizers(fp8_output, fp8_grad)
+            )
+            debug = TEDebugState.debug_enabled
```
ptrendx, line 1134:
```suggestion
            debug = TEDebugState.debug_enabled
            quantizers = (
                self._get_quantizers(fp8_output, fp8_grad)
                if not debug
                else self._get_debug_quantizers(fp8_output, fp8_grad)
            )
```

transformer_engine/pytorch/module/linear.py
```diff
@@ -1106,13 +1126,29 @@ def forward(
...
+            if debug:
+                if not any_feature_enabled(quantizers):
+                    # If no feature is used, then run faster implementation with debug = False.
+                    quantizers = self._get_quantizers(fp8_output, fp8_grad)
+                    debug = False
+
+            if debug and isinstance(weight_tensor, QuantizedTensor):
+                raise RuntimeError("FP8 weights are not supported in debug mode.")
```
ptrendx, line 1142:
```suggestion

                if isinstance(weight_tensor, QuantizedTensor):
                    raise RuntimeError("FP8 weights are not supported in debug mode.")
```

transformer_engine/debug/pytorch/utils.py
```diff
@@ -0,0 +1,13 @@
...
+
+"""Utils functions for the debug module."""
+
+
+def any_feature_enabled(quantizers):
+    """Returns True if at least one API call is made from DebugQuantizer."""
+    for q in quantizers:
+        if q.any_feature_enabled():
+            return True
+    return False
```
ptrendx, line 13:
```suggestion
    return any([q.any_feature_enabled() for q in quantizers])
```

transformer_engine/pytorch/module/base.py
```diff
@@ -1064,3 +1106,47 @@ def _load_from_state_dict(
...
+        Validate name passed to the module.
+        This is invoked in the forward() method as module names are assigned after Model is initialized in Megatron-LM.
+        If no name is assigned, it creates a default name with layer count as the variable.
+        """
+        assert TEDebugState.debug_enabled
+        import nvdlfw_inspect.api as debug_api
+
+        if self.name is None:
+            debug_api.log_message(
+                "[DEBUG-WARNING] Names are not provided to debug modules. ",
```
ptrendx, line 1121: The tag should probably be a constant defined somewhere in debug utils rather than hardcoded here.

transformer_engine/pytorch/module/base.py
```diff
@@ -1007,6 +1039,11 @@ def get_weight_workspace(
                     out = None
                     del self._fp8_workspaces[cache_name]
 
+            is_debug = isinstance(quantizer, DebugQuantizer)
+            is_out_debug_tensor = out is not None and isinstance(out, DebugQuantizedTensor)
+            if is_debug != is_out_debug_tensor:
+                out = None
```
ptrendx, line 1045: Could we pass additional `debug` parameter (or just check the state in the global object) before doing those cases that could only happen during debug? I'd like to minimize the number of Python calls that we have to do.

transformer_engine/pytorch/module/base.py
```diff
@@ -412,6 +418,9 @@ def __init__(self) -> None:
         self._fp8_workspaces: Dict[str, QuantizedTensor] = {}
         self.activation_dtype: Optional[torch.dtype] = None
 
+        if not TEDebugState.debug_enabled:
+            TEDebugState.initialize()
```
ptrendx, line 422: I thought that `TEDebugState.debug_enabled` would be `False` when we do not want to do debug. So would we try to initialize the state every time then?

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -1292,13 +1311,29 @@ def forward(
             else:
                 bias_tensor = getattr(self, self.bias_names[0])  # Unused
 
+            debug = TEDebugState.debug_enabled
```
ptrendx, line 1314: Could we move this to the beginning of the function? You use the state there as well.

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -1415,17 +1473,36 @@ def forward(
             is_first_microbatch = False
 
         with self.prepare_forward(inp, num_gemms=2) as inp:
+
+            quantizers = (
+                self._get_quantizers()
+                if not TEDebugState.debug_enabled
+                else self._get_debug_quantizers()
+            )
+            debug = TEDebugState.debug_enabled
```
ptrendx, line 1482: Move to the beginning of the function.

transformer_engine/debug/pytorch/debug_state.py
```diff
@@ -0,0 +1,69 @@
...
+class TEDebugState:
+    """
+    A class to manage the state of debug layers.
+    """
+
+    layer_count = 1
+    layers_initialized = {}
+    weight_tensor_tp_group_reduce = True
+    debug_enabled = None
+    initialized = False
```
ptrendx, line 21: You have the initialized variable but you don't use it - shouldn't `initialize` function check if the state was already initialized before doing anything?

transformer_engine/debug/pytorch/debug_quantization.py
```diff
@@ -0,0 +1,524 @@
...
+            args["tensor"] = columnwise_gemm_tensor
+            args["rowwise"] = False
+            debug_api.transformer_engine.inspect_tensor_postquantize(**args)
+
+    def quantize(
+        self,
+        tensor: torch.Tensor,
+        *,
+        out: Optional[Union[torch.Tensor, DebugQuantizedTensor]] = None,
+        dtype: torch.dtype = None,
```
ptrendx, line 254: What is the use case for this?

transformer_engine/pytorch/module/base.py
pggPL: I removed this prefix for now, since we do not use it in the other places.

transformer_engine/pytorch/module/base.py
pggPL: We will need to have logic of changing quantizers there in the near future - when quantizer does not match the cache tensor. I can leave this for now and this will need to be changed in the PR which would enable recipe change during training.

transformer_engine/debug/pytorch/debug_quantization.py
pggPL: Weight caching for fake quantization for example. We want to save them in correct precision.

transformer_engine/pytorch/module/base.py
pggPL: to raise an error if one module is initialized before usage of nvinspect api, and another is initialized after

transformer_engine/debug/pytorch/debug_state.py
pggPL: Oh, I haven't removed it and lint does not error it for some reason. Now we have 3 states of debug enabled true/false/none indicating that. Maybe it would result is some small speedup during initialization, but I don't think it is worth it.

transformer_engine/pytorch/utils.py
```diff
@@ -329,6 +330,16 @@ def round_up_to_nearest_multiple(value, multiple):
...
 
 
+def needs_quantized_gemm(obj, rowwise=True):
+    """Used to check if obj will need quantized gemm or normal gemm."""
+    if isinstance(obj, DebugQuantizedTensor):
+        return (
+            type(obj.get_tensor(not rowwise))  # pylint: disable=unidiomatic-typecheck
+            is not torch.Tensor
+        )
+    return type(obj) is not torch.Tensor  # pylint: disable=unidiomatic-typecheck
```
ksivaman, line 340:
```suggestion
        return not isinstance(obj.get_tensor(not rowwise), torch.Tensor)
    return not isinstance(obj, torch.Tensor)
```

transformer_engine/pytorch/module/base.py
```diff
@@ -1042,6 +1079,11 @@ def get_weight_workspace(
             else:
                 tex.quantize(tensor, quantizer, out, skip_update_flag)
 
+        if not needs_quantized_gemm(type(out)):  # only holds for debug quantizer
```
ksivaman, line 1082:
```suggestion
        if not needs_quantized_gemm(out):  # only holds for debug quantizer
```

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -109,6 +110,21 @@ def general_gemm(
...
+        A = A.get_tensor(not transa)
+        B = B.get_tensor(transb)
+
+        assert (type(A) in [torch.Tensor, torch.nn.parameter.Parameter]) == (
+            type(B) in [torch.Tensor, torch.nn.parameter.Parameter]
+        ), (
+            "[Debug tools] Processed tensors should both be FP8 tensors or both be torch tensors  "
+            f"            but type(A) = {type(A)},             "
+            f"  type(B) = {type(B)}"
+        )
```
ksivaman, line 126: This feels like an awkward check, could you explain it, i.e. what are we trying to verify here? Can `get_tensor` even return non `torch.Tensor` types? Note that `torch.nn.parameter.Parameter` is `torch.Tensor` as well, so if we are checking this, it's better to move to the suggested change.

ksivaman, line 126:
```suggestion
        assert isinstance(A, torch.Tensor) == isinstance(B, torch.Tensor), (
            "[Debug tools] Processed tensors should both be FP8 tensors or both be torch tensors  "
            f"            but type(A) = {type(A)},             "
            f"  type(B) = {type(B)}"
        )
```

transformer_engine/pytorch/utils.py
ptrendx: @ksivaman I don't believe this suggestion is correct. Basically you want to differentiate here between regular torch.Tensor and quantized tensor (which could be a subclass of torch.Tensor). In my internal quantizer PR #1638 I introduce a function to check if it is quantized tensor, maybe that would be better instead.

transformer_engine/pytorch/cpp_extensions/gemm.py
ptrendx: QuantizedTensor is a subclass of torch.Tensor so it would pass yield `True` in `isinstance(A, torch.Tensor)`. See my previous comment.

pggPL: It was checking if both tensors are torch.Tensor ot Float8Tensors. Since more tensor class were added in the meantime and this check is not necessary, I removed it.

transformer_engine/debug/pytorch/debug_quantization.py
pggPL: Suppose weight is stored in FP32, but activation is in BF16. Then we want weight after fake-quantization to be in BF16 to prevent cast each forward.

transformer_engine/debug/pytorch/debug_state.py
```diff
@@ -0,0 +1,68 @@
...
+            cls.debug_enabled = debug_api.DEBUG_MANAGER is not None
+
+    @classmethod
+    def reset(cls):
+        """Resets layer count and stats buffers."""
+        from ..features.utils.stats_buffer import STATS_BUFFERS
+
+        STATS_BUFFERS.reset()
+        cls.debug_enabled = None
+        cls.layers_initialized.clear()
```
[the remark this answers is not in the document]
pggPL: this call is used mostly for testing - your scenario would work

transformer_engine/pytorch/module/base.py
```diff
@@ -1042,6 +1079,11 @@ def get_weight_workspace(
             else:
                 tex.quantize(tensor, quantizer, out, skip_update_flag)
 
+        if not needs_quantized_gemm(out):  # only holds for debug quantizer
+            assert (
+                out.dtype == workspace_dtype
+            ), "Activation dtype cannot be changed with nvidia-dlframework-inspect."
```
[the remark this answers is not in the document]
pggPL: I moved this check to the debug quantizer function, when this is trivial what does error mean.

transformer_engine/debug/pytorch/debug_state.py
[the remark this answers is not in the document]
ptrendx: Then maybe let's change it to `_reset`?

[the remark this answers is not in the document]
pggPL: done

transformer_engine/pytorch/module/linear.py
```diff
@@ -1193,8 +1230,9 @@ def forward(
 
     def _get_quantizers(self, fp8_output, fp8_grad):
         if not self.fp8:
-            return [None] * 5
+            return [None] * 6
         grad_input_quantizer = None
+        grad_weight_quantizer = None
```
timmoon10, line 1235: I strongly oppose adding a quantizer for the grad weight. It doesn't make sense logically since we never plan to output a quantized grad weight, barring some major development in optimizers. I get that the asymmetry is weird, but the real problem is that `output_quantizer` and `grad_input_quantizer` are also illogical. We should be using the `input_quantizer`/`grad_output_quantizer` from the layer that consumes them, like what te.Sequential does, but that's very tricky to implement.

timmoon10 (changes requested):

transformer_engine/pytorch/module/linear.py
pggPL: Ok, but maybe one wants to do some experiments with weight quantization and debug tools helps with that. Moreover this quantizer enables us to log the statistics.

ptrendx: @timmoon10 Actually, I would argue that `grad_weight_quantizer` makes more sense compared to the other quantizers you listed since it is actually "local" vs the other ones logically belonging to the other operations. Even in te.Sequential the grad_weight quantizer would belong to the op that holds those weights.
I also don't agree that `we never plan to output a quantized grad weight` - while the convergence story of this would need to be understood, projects like MS-AMP tried to provide FP8 gradient functionality since it saves a lot of memory during training. So it is not completely out of the question that we will want to provide that as well.

transformer_engine/pytorch/module/linear.py
```diff
@@ -840,6 +851,7 @@ def __init__(
         ub_bulk_dgrad: bool = False,
         ub_bulk_wgrad: bool = False,
         ub_name: Optional[str] = None,
+        name: Optional[str] = None,
```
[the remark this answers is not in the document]
pggPL: I added some simple docs and I will use it in the NVTX markers and add it in error messages in the another PR.

transformer_engine/debug/pytorch/debug_quantization.py
pggPL: we do it here https://github.com/NVIDIA/TransformerEngine/blob/66d6afbf6e4bccd3d40e85a3d5ed307229137304/transformer_engine/pytorch/module/linear.py#L200

ptrendx (approved):
From my side LGTM.

transformer_engine/pytorch/module/linear.py
timmoon10: The MS-AMP case is a good example. I suppose the right answer would be to gracefully support multiple quantizer types. In most cases we would have `grad_weight_quantizer=None`. `grad_weight_quantizer` would only be non-trivial if we want quantized grads or we're doing something weird like debug mode.

Unresolving for future visibility.

timmoon10 (approved):
I haven't fully gone through the code, but things look reasonable to me. I would be fine merging.

PR #1581, opened 2025-03-15
[PyTorch] Optimize MXFP8 all-gathers

timmoon10:
# Description

This is a follow-up to https://github.com/NVIDIA/TransformerEngine/pull/1505, which added support for MXFP8 tensors with only column-wise data. This PR makes two changes to MXFP8 all-gathers
- Avoid communicating row-wise MXFP8 data if not needed
- Coalesce MXFP8 NCCL collectives into a single kernel. We temporarily removed it due to a performance issue, which has recently been fixed.

When looking at the backward pass of a Transformer layer, I now see good bulk overlapping of input all-gathers and dgrad GEMMs.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Avoid communicating row-wise MXFP8 data if not needed
- Coalesce MXFP8 NCCL collectives into a single kernel

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #1620, opened 2025-03-27
[PyTorch] Debug NCCL communication overlapping in linear backward with FP8 data

timmoon10:
# Description

When running tensor-parallel linear modules without Userbuffers, the backward pass will attempt to overlap the input all-gather with the dgrad GEMM. This works well for MXFP8 (see https://github.com/NVIDIA/TransformerEngine/pull/1581), but we are not seeing good overlapping with FP8 because we compute FP8 transposes immediately after the all-gather.

This PR restores overlapping in the FP8 case by separating the FP8 transpose from the FP8 all-gather. In principle we could have done this by putting the all-gather and transpose on a separate CUDA stream, but we would not work if running with `CUDA_DEVICE_MAX_CONNECTIONS=1`.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add FP8-specific logic in linear backward passes to overlap input all-gather with dgrad GEMM

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/distributed.py
+++ b/transformer_engine/pytorch/distributed.py
@@ -19,7 +19,7 @@ from torch.distributed.fsdp import FullyShardedDataParallel as FSDP
 from torch.distributed.fsdp._common_utils import _get_module_fsdp_state
 from torch.distributed.fsdp._traversal_utils import _get_fsdp_states_with_modules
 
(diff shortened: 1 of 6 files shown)


timmoon10:
Force-merging so this can land in 2.2.

PR #1621, opened 2025-03-27
[PyTorch] Support default process group with FP8 current scaling

timmoon10:
# Description

We currently experience segfaults when FP8 current scaling is configured with the default PyTorch process group (`None`). This is because we access the process group from C++ and don't handle the case where it is `None`. This PR adds logic to canonicalize the process group before using it.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Canonicalize amax reduction group before FP8 current scaling quantizer uses it
- Remove logic for amax reduction group size from FP8 current scaling quantizer

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/tensor/float8_tensor.py
+++ b/transformer_engine/pytorch/tensor/float8_tensor.py
@@ -222,7 +222,9 @@ class Float8CurrentScalingQuantizer(Quantizer):
 
         # Deprecated option
         if amax_reduction_size is not None:
-            warnings.warn("amax_reduction_size kwarg is deprecated", DeprecationWarning, stacklevel=2)
+            warnings.warn(
+                "amax_reduction_size kwarg is deprecated", DeprecationWarning, stacklevel=2
+            )
 
     def update_quantized(
         self,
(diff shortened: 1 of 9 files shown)


transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -208,20 +207,25 @@ def __init__(
...
         self.with_amax_reduction = with_amax_reduction
         self.amax_reduction_group = amax_reduction_group
-        self.amax_reduction_size = amax_reduction_size
         self.force_pow_2_scales = force_pow_2_scales
         self.amax_epsilon = amax_epsilon
 
+        # Deprecated option
+        if amax_reduction_size is not None:
+            warnings.warn(
+                "amax_reduction_size kwarg is deprecated", DeprecationWarning, stacklevel=2
```
ptrendx, line 226: If we are going to merge this in 2.2 then we don't need to deprecate it since it never existed in a stable release - we can just remove it.

ptrendx (approved):

PR #1628, opened 2025-03-31
[Common] Improved performance of mxfp8 cast kernels

Oleg-Goncharov:
# Description

Modified and tuned fused cast mxfp8 kernels for better performance. This CUDA kernel template implements optimized scaling along columnwise, rowwise, or both dimensions. It uses compile-time specialization and SIMD instructions to maximize compute performance. Data is transferred asynchronously via TMA into shared memory, processed in stages with prefetching overlap, and results are cached as needed. 

## Kernel Logic Overview
The kernel is a function template designed to handle multiple scaling scenarios, leveraging compile-time logic to generate specialized, optimized code paths. It consists of a common processing part shared across all scenarios, along with three specialized branches:

**1. Columnwise scaling**
**2. Rowwise scaling**
**3. Both columnwise and rowwise scaling**

Each branch introduces specific optimizations based on the template parameters provided at instantiation.

Since the kernel is compute-bound, performance is improved by favoring SIMD (vectorized) instructions over scalar ones. For example, operations like amax computation and element rescaling (multiplication) are performed on two elements simultaneously using a single instruction.

## Common Part
The common part of the kernel performs the following tasks:

- Defines block and thread offsets.
- Manages memory barriers.
- Handles data transfers and updates.

Data transfer between global memory and shared memory (SHMEM) is performed using the TMA engine. SHMEM serves as the main workspace where input, output, and intermediate cached data (e.g., activations) reside. Due to the limited size of shared memory, the assigned data block (chunk) is partitioned into smaller tiles processed iteratively in stages.

Due to the asynchronous nature of the TMA engine, the kernel overlaps computation and memory transfer. At each iteration (except the last), data for the next stage is prefetched while the current stage's data is being processed.

#### Thread-wise Reduction
The dBias value only needs to be computed once per input and is performed during the scaling phase:
- Columnwise scaling: dBias is computed in the columnwise phase.
- Rowwise scaling: dBias is computed in the rowwise phase.
- Both scalings: dBias is computed in the columnwise phase.

#### Block-wise Reduction
Depending on where thread-wise reduction occurred:
- When performed during the columnwise phase, accumulated values are stored directly into the workspace.
- When performed during the rowwise phase, elements must be un-swizzled to ensure proper indexing. Additionally, partial dBias results across waves are further reduced using an intermediate buffer that shares memory with the input buffer

### 1. Columnwise Scaling
Thread-to-data mapping: One thread processes one column (threads per block = number of columns).
Processing: Each thread moves vertically down its assigned column, reading one element at a time into registers (REG) from SHMEM.
#### Computation:
1. Compute the amax while reading.
2. After reading 32 elements, calculate the scaling factor and rescale all cached elements.
3. Store the rescaled elements back into SHMEM.

### 2. Rowwise Scaling
Thread-to-data mapping: One thread processes 32 consecutive (rowwise) elements (one MXFP8 block per thread, no inter-thread communication).
E.g., for a chunk size of 128, four threads per row are needed.
Processing: Each thread traverses its assigned block horizontally, reading and writing four elements at a time (PACK_SIZE = 4). The layout is treated as a closed ring: when a thread reaches the rightmost element of the block, it wraps around to the leftmost element to continue processing. This ring-based traversal pattern, along with swizzling, helps minimize shared memory bank conflicts.

NOTE: `PACK_SIZE` parameter can be fine tuned to slightly improve the performance in some scenarios, although it may have a negative impact on performance in other scenarios. E.g., with the `--fast_math` flag active, setting the `PACK_SIZE=8` has the following performance impact vs. `PACK_SIZE=4`:

#### Computation:
1. Compute amax during reads.
2. Rescale cached elements (stored in registers).
3. Store the rescaled elements back into SHMEM, applying an un-swizzling pattern to restore original indices.

**A few examples of threads-to-data mapping between shared memory and registers (lower bar) while iterating through waves:**
`tid = 0 (tid_y = 0; tid_x = 0)`
`belongs to Group 0`
`initial offset = 0 elements`
Rowwise threads mapping 0

`tid = 11 (tid_y = 2; tid_x = 3)`
`belongs to Group 2`
`initial offset = 8 elements`
Rowwise threads mapping 1

`tid = 26 (tid_y = 6; tid_x = 2)`
`belongs to Group 6`
`initial offset = 24 elements`
Rowwise threads mapping 2

**Threads-to-data mapping for a single warp during a single wave:**
This shows how a warp spans across the 32 banks of shared memory, avoiding bank conflicts when writing FP8 data during a single wave. Since four FP8 elements fit into a single 4-byte bank, vectorized writes (4×FP8) align naturally with the memory banks.

Rowwise_threads mapping

### 3. Both Columnwise and Rowwise Scaling (2x)
Scaling across both dimensions is performed sequentially in two phases:

#### 1st Phase (Columnwise Scaling)
Same as the columnwise-only case, with one difference: if **IS_DACT** or **IS_ACT** template parameters are set, activations are computed and cached in a dedicated buffer within shared memory.

#### 2nd Phase (Rowwise Scaling)
Instead of recomputing activations, the kernel reuses cached activations from the first phase.
Only the amax computation and rescaling are performed; no dBias computation is required, as it was already completed in the first phase.
**The activations cache buffer shares memory with the input buffer of the corresponding stage.**

## Notes:
- Vectorization is key to performance, particularly for amax computation and rescaling.
- Swizzling and un-swizzling patterns are carefully used to minimize shared memory bank conflicts.
- Shared memory reuse (between activations cache, intermediate buffers, and input/output buffers) maximizes efficient memory utilization.

```
Benchmark scenario:
Input tensor 4096x13312, BF16
Output tensor 4096x13312, MXFP8 E4M3
Activation type (if used) GeLU/dGeLU
Kernel 1 - Fused Cast MXFP8
Kernel 2 - Reduce dBias
```
Performance was measured with NVIDIA Nsight Systems version 2025.2.1.130-252135690618v0 using the corresponding test suites. Runtime is measured in microseconds. 

Performance Comparison All

Performance Comparison 2 All

NOTE: `CAST+DBIAS+DACT` with `--fast_math` has relatively high speedup factors, because originally there was no such an option to compile `cast.cu` with this flag enabled, so this metric does not provide a direct performance comparison.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Increased block size from `64x64` to `128x128` (except for CAST_DBIAS)
- Modified the scheme of how tensor elements are processed. ROWWISE: 32 elements per thread (i.e. single scaling factor)
- Added micro-optimizations (e.g. using `MUL2` instructions; fusing MUL and CVT instructions directly in PTX)

# Checklist:


transformer_engine/common/util/ptx.cuh
```diff
@@ -164,6 +164,159 @@ __device__ __forceinline__ void fence_proxy_async_shared_cta() {
   asm volatile("fence.proxy.async.shared::cta;");
 }
 
+// SIMD like "Fused" cast + multiplication (x2)
+template <typename T1, typename T2>
+__device__ __forceinline__ void mul_cvt_2x(T1 &out_2x, const T2 &in_2x, const float2 &scale_2x);
```
ptrendx, line 169:
```suggestion
__device__ __forceinline__ void mul_cvt_2x(T1 &out_2x[2], const T2 &in_2x[2], const float2 &scale_2x);
```

transformer_engine/common/util/ptx.cuh
```diff
@@ -164,6 +164,159 @@ __device__ __forceinline__ void fence_proxy_async_shared_cta() {
...
+      "mov.b64 {val2,val1}, val_pair_after; \n\t"
+      "cvt.rn.satfinite.e5m2x2.f32 %0, val1, val2; \n\t"
+      "}"
+      : "=h"(reinterpret_cast<uint16_t &>(out_2x))
+      : "r"(reinterpret_cast<const uint32_t &>(in_2x)),
+        "l"(reinterpret_cast<const uint64_t &>(scale_2x)));
+}
+
+template <typename T>
+__device__ __forceinline__ void abs_max_2x(T &dest, const T &p1, const T &p2);
```
ptrendx, line 302: Same issue here.

transformer_engine/common/util/ptx.cuh
```diff
@@ -164,6 +164,159 @@ __device__ __forceinline__ void fence_proxy_async_shared_cta() {
...
+
+template <>
+__device__ __forceinline__ void abs_max_2x<bf16>(bf16 &dst, const bf16 &p1, const bf16 &p2) {
+  asm volatile(
+      "max.xorsign.abs.bf16x2 %0, %1, %2;"
```
timmoon10, line 307: Is there any significance to using xorsign, or is this just the most convenient instruction for absmax?

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -28,36 +28,29 @@
...
-constexpr size_t THREADS_PER_CHUNK_X_ROWWISE =
-    MXFP8_CHUNK_DIM_X / ELEMS_PER_THREAD;  //   4 = 64 / 16
-constexpr size_t THREADS_PER_CHUNK_Y_ROWWISE =
-    MXFP8_THREADS_PER_CHUNK / THREADS_PER_CHUNK_X_ROWWISE;         //  16 = 64 / 4
-constexpr size_t THREADS_PER_CHUNK_X_COLWISE = MXFP8_CHUNK_DIM_X;  //  64
-constexpr size_t MXFP8_BUFF_STAGES_NUM =
-    MXFP8_BUFFER_DIM_Y / THREADS_PER_CHUNK_Y_ROWWISE;                        //   2 = 32 / 16
-constexpr size_t MXFP8_ITERATIONS = MXFP8_CHUNK_DIM_Y / MXFP8_BUFFER_DIM_Y;  //   2 = 64 / 32
-static_assert(MXFP8_ITERATIONS >= MXFP8_PREFETCH_BUFFERS_NUM);
+constexpr size_t ALIGNMENT = 128;
```
timmoon10, line 31: Nit: This is MXFP8-specific, so better to put it within the `mxfp8_kernel` namespace.

transformer_engine/common/util/ptx.cuh
Oleg-Goncharov: It is a bit inconvenient, but in the `max` instruction, the `.abs` modifier requires the `.xorsign`, so they go together

transformer_engine/common/util/cast_kernels.cuh
Oleg-Goncharov: `ALIGNMENT` is required by the TMA, which we also use in regular FP8 cast. I think it should rather be moved to `utils.cuh` and named as, e.g., the `TMA_SHMEM_ALIGNMENT`, as the gated cast kernels also use it (and the value is the same, i.e. 128, but there it is currently named as `ALIGNMENT_SIZE`)

tests/cpp/operator/CMakeLists.txt
```diff
@@ -3,28 +3,28 @@
...
+            #    test_cast_transpose_dbias.cu
+            #    test_cast_transpose_dbias_dgelu.cu
+            #    test_cast_transpose_dgeglu.cu
+            #    test_act.cu
+            #    test_normalization.cu
+            #    test_normalization_mxfp8.cu
+            #    test_multi_cast_transpose.cu
+            #    test_multi_padding.cu
+            #    test_causal_softmax.cu
+            #    test_swizzle.cu
```
ptrendx, line 27:
```suggestion
               test_cast.cu
               test_cast_current_scaling.cu
               test_cast_dbias.cu
               test_cast_dbias_dgelu.cu
               test_cast_gated_swiglu.cu
               test_cast_mxfp8_gated_swiglu.cu
               test_qdq.cu
               test_cast_mxfp8.cu
               test_dequantize_mxfp8.cu
               test_transpose.cu
               test_cast_transpose.cu
               test_cast_transpose_current_scaling.cu
               test_cast_transpose_dbias.cu
               test_cast_transpose_dbias_dgelu.cu
               test_cast_transpose_dgeglu.cu
               test_act.cu
               test_normalization.cu
               test_normalization_mxfp8.cu
               test_multi_cast_transpose.cu
               test_multi_padding.cu
               test_causal_softmax.cu
               test_swizzle.cu
```

tests/cpp/operator/test_cast_mxfp8.cu
```diff
@@ -483,6 +483,7 @@ std::vector<std::vector<size_t>> matrix_sizes = {
     {1024},
     {8, 32, 1024},
     {16, 8, 4, 512},
+    // {4096, 13312},
```
ptrendx, line 486: Do you want this shape to be included or just needed it for the benchmarking?

transformer_engine/common/common.h
```diff
@@ -212,6 +212,11 @@ constexpr T DIVUP(const T &x, const T &y) {
   return (((x) + ((y)-1)) / (y));
 }
 
+template <typename T1, typename T2>
+constexpr __device__ __host__ __forceinline__ uint64_t DIVUP_TO_MULTIPLE(const T1 &N, const T2 &M) {
+  return DIVUP(static_cast<uint64_t>(N), static_cast<uint64_t>(M)) * M;
```
ptrendx, line 217: Maybe add `static_assert` that T1 and T2 are integral types?

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -28,36 +28,33 @@
...
+
+constexpr size_t SCALE_DIM_Y = 32;
+constexpr size_t SCALE_DIM_X = 32;
+
+constexpr size_t BUFFS_NUM = 2;
+constexpr size_t PACK_SIZE = 4;
+constexpr size_t WAVES = SCALE_DIM_X / PACK_SIZE;
+
+// Number of 1-byte elements that span 32 banks (4-byte each) of shared memory
+constexpr size_t BANK_WIDTH = (32 * 4) / 1;  // 128
```
ptrendx, line 41: This name is confusing - I would think that it would mean the width of a single bank. Maybe `TOTAL_BANKS_WIDTH'?

tests/cpp/operator/test_cast_mxfp8.cu
Oleg-Goncharov: No, we can remove it, I used it just for benchmarking

transformer_engine/common/util/ptx.cuh
```diff
@@ -169,6 +169,180 @@ __device__ __forceinline__ void fence_proxy_async_shared_cta() {
...
+
+using fp8e4m3x2 = fp8x2_t<fp8e4m3>;
+using fp8e5m2x2 = fp8x2_t<fp8e5m2>;
+
+static_assert(sizeof(fp8e4m3x2) == 2);
+static_assert(sizeof(fp8e5m2x2) == 2);
+
+// SIMD like "Fused" cast + multiplication (x2)
+template <typename T1, typename T2>
+__device__ __forceinline__ void mul_cvt_2x(T1 &out, const T2 &in, const float2 &scale);
```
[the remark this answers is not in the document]
Oleg-Goncharov: Yes, indeed, simple function overloading works. I also slightly simplified the code by using new FPx2 struct

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -28,36 +28,33 @@
...
+// Number of threads (rowwise scaling) that span 32 banks (4-byte banks) of shared memory
+constexpr size_t THREADS_PER_BANK = TOTAL_BANKS_WIDTH / SCALE_DIM_X;  // 4 = 128 / 32
+
+template <typename ParamOP, float (*OP)(float, const ParamOP &)>
+constexpr __device__ __forceinline__ bool is_cached_act_op() {
+  return (OP == gelu<float, float>) || (OP == dgelu<float, float>) ||
+         (OP == sigmoid<float, float>) || (OP == dsigmoid<float, float>) ||
+         (OP == qgelu<float, float>) || (OP == dqgelu<float, float>) ||
+         (OP == silu<float, float>) || (OP == dsilu<float, float>);
+}
```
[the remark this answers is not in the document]
Oleg-Goncharov: When I initially (a bit before when all the optimizations were introduced) measured the performance difference between the cached/non-cached, recomputing the activations was faster in case of simple types like relu and srelu. But measuring it now for relu and srelu, I see the difference is around ±1%, so I completely removed this condition to make the code uniform. Now all activations are cached regardless of their type.

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -67,398 +64,475 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
...
-  const int dbias_rowwise_offset_Y = blockIdx.y * MXFP8_CHUNKS_PER_BLOCK_Y + tid_rowwise_Y;
-  const int dbias_rowwise_block_offset_X =
-      blockIdx.x * MXFP8_CHUNKS_PER_BLOCK_X * MXFP8_CHUNK_DIM_X + thread_offset_X_rowwise;
-  const int dbias_colwise_offset_Y = blockIdx.y;
-  const int dbias_colwise_block_offset_X =
-      blockIdx.x * MXFP8_CHUNKS_PER_BLOCK_X * MXFP8_CHUNK_DIM_X + tid_colwise_X;
-  const int dbias_stride = cols;
+  using IType2 =
+      std::conditional_t<std::is_same_v<IType, float>, float2,
+                         std::conditional_t<std::is_same_v<IType, bf16>, ptx::bf16x2, ptx::fp16x2>>;
```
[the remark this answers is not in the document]
Oleg-Goncharov: Thank you for the suggestion! It also a bit simplified the code in few other places.

ptrendx (approved):

tests/jax/test_custom_call_compute.py
```diff
@@ -78,8 +78,14 @@ def is_shape_supported_by_mxfp8(input_shape):
         return False
 
 
-def assert_bitwise_scaled_tensors(a: ScaledTensor, b: ScaledTensor):
+def assert_bitwise_scaled_tensors(
+    a: ScaledTensor, b: ScaledTensor, precise_comparison: bool = True
+):
```
timmoon10, line 83: If I call `assert_bitwise_scale_tensors`, I would expect it to assert that scaled tensors are bit-wise exact. To handle non-exact comparisons, better to rename the function:
```suggestion
def assert_allclose_scaled_tensors(
    a: ScaledTensor, b: ScaledTensor, bitwise: bool = True
):
```

tests/jax/test_custom_call_compute.py
```diff
@@ -768,12 +761,24 @@ def _test_quantize_dact_dbias(
...
 
         if is_dbias:
-            assert_allclose(te_dbias, jax_dbias)
+            # TE kernels cast the intermediate results to the input dtype which reduces precision compared to the JAX implementation, for dbias this typically only affects bfloat16.
+            precise_comparison = not (
+                in_dtype == jnp.bfloat16 and scaling_mode.is_1d_block_scaling()
+            )
+            assert_allclose(
+                te_dbias, jax_dbias, dtype=in_dtype if precise_comparison else out_dtype
+            )
```
timmoon10, line 781: If FP16 is also doing this cast, how is it maintaining bit-wise accuracy?

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -67,398 +56,478 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
...
+          if constexpr (!std::is_same_v<IType, float>) {
+            elt = static_cast<float>(static_cast<IType>(elt));
+          }
+          if constexpr (IS_DBIAS) {
+            partial_dbias_colwise += elt;
+          }
+          // Cache computed activations to avoid computing them again in the 2nd pass along another dimension
+          if constexpr (IS_CACHED_ACT_OP) {
+            cached_act_sh[shmem_offset_colwise] = static_cast<IType>(elt);
+          }
```
timmoon10, line 246: I see the performance benefit of caching the output value in `IType`, but have we confirmed that it doesn't affect convergence? We could avoid this by caching in FP32, but I suspect we may not have enough shared memory.

transformer_engine/common/util/cast_kernels.cuh
```diff
@@ -67,398 +56,478 @@ __global__ void __launch_bounds__(MXFP8_THREADS_PER_CHUNK)
...
+            float act_in_elt = static_cast<float>(act_in_sh[shmem_offset_colwise]);
+            elt *= OP(act_in_elt, {});
+          }
+          // Numerical truncation: Downcast to IType (BF16/FP16), then upcast it back to FP32
+          if constexpr (!std::is_same_v<IType, float>) {
+            elt = static_cast<float>(static_cast<IType>(elt));
+          }
+          if constexpr (IS_DBIAS) {
+            partial_dbias_colwise += elt;
+          }
```
timmoon10: Why do we need to downcast before computing dbias? It hurts the numerical accuracy and I don't see a perf benefit since we're computing in FP32.

Oleg-Goncharov: Oh, you are right, down-casting floated above the dbias computation. I'll fix it

transformer_engine/common/util/cast_kernels.cuh
Oleg-Goncharov: MLPerf team hasn't reported any convergence issues using this branch. And we are tight in shared memory, so have to reuse the input allocation for data caching

phu0ngng:
Hi,

Regarding the recent changes to the JAX unit test, I think it's fine to proceed with the current approach for now so that we can upstream this PR soon.

Looking ahead, I would prefer that we align the JAX-based implementation with the TE kernel for the following reasons:
- We want to benchmark end-to-end workloads both with and without TE custom calls. If the TE kernel stores intermediate results in dtype while the JAX-based implementation uses FP32, the comparison will no longer be apples-to-apples.
- Since we have already agreed to store intermediate results in dtype going forward—and other JAX-based frameworks (such as MaxText) are also adopting this approach—I don’t see a compelling reason to do otherwise for our JAX-based implementation.

What do you think? @jberchtold-nvidia @ptrendx

jberchtold-nvidia:
> * Since we have already agreed to store intermediate results in dtype going forward—and other JAX-based frameworks (such as MaxText) are also adopting this approach—I don’t see a compelling reason to do otherwise for our JAX-based implementation.

Sounds good, I agree

tests/jax/test_custom_call_compute.py
Oleg-Goncharov: CI had only BF16 failing cases, so the tolerances were slightly increased only here

PR #1632, opened 2025-04-01
Symmetric memory all reduce

wdykas:
# Description

This add the ability to use the new symmetric PyTorch api for all reduces in TE. 2999 is the internal Megatron MR for use.

This functionality is only available with the torch in container 25.03 or torch 2.7 and beyond

Note:
Currently this introduces an entirely new all reduce function called symmetric_all_reduce. Technically this could all be merged into a single all reduce function and maintain backwards compatibility but for now I have it separated since this is new functionality that most users probably will not be setting.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- add new all reduce functions using the torch symmetric api
- Add argument symmetric_ar_type to control underlying algorithm being used

# Checklist:


transformer_engine/pytorch/distributed.py
```diff
@@ -1153,6 +1160,106 @@ def gather_along_first_dim(
...
+    msg.copy_(inp.reshape(-1))
+
+    all_reduce_impl(
+        msg,
+        "sum",
+        group_name,
+    )
+
+    # Copy the result back to the input tensor
+    inp.copy_(msg.reshape(tensor_shape))
```
denera, line 1258:
```suggestion
    if inp not in symmetric_mem_cache.values():
        # Re-use an existing symmetric memory tensor if possible, allocate one otherwise
        msg = get_symmetric_memory_tensor(inp.shape, inp.numel(), inp.dtype, inp.device, group_name)
        
        msg.copy_(inp.reshape(-1))
        
    else:
        msg = inp

    all_reduce_impl(
        msg,
        "sum",
        group_name,
    )
    
    if inp not in symmetric_mem_cache.values():
        # Copy the result back to the input tensor
        inp.copy_(msg.reshape(tensor_shape))
```

I would recommend refactoring this to allow for the input itself to be a cached symmetric memory tensor, and expose new standalone API to extract a cached symmetric memory tensor like:
```python
def get_symmetric_memory_tensor(tensor_numel, tensor_dtype, tensor_device, group_name, tag=None):
    # Create a cache key based on tensor properties and group
    cache_key = (tensor_numel, tensor_dtype, tensor_device, group_name, tag)

    # Check if we already have a symmetric memory tensor for this configuration
    if cache_key not in symmetric_mem_cache:
        # Create a new symmetric memory tensor if not in cache
        msg = symm_mem.empty(
            tensor_numel,
            dtype=tensor_dtype,
            device=tensor_device,
        )
        # Perform the rendezvous once for this tensor
        symm_mem.rendezvous(msg, group=tp_group)
        # Store in cache
        symmetric_mem_cache[cache_key] = msg
    else:
        # Reuse the existing symmetric memory tensor
        msg = symmetric_mem_cache[cache_key]

    return msg
```

This enables an optimization where a symmetric memory tensor can directly be used as the output buffer of a preceding operation (e.g. GEMM), thus avoiding two copy operations in the symmetric all-reduce.

denera (changes requested):
Unless there's a reason why we shouldn't use this in `LayerNormMLP`, I would recommend extending the implementation to cover that module too just for the sake of feature parity even if there are no active use-cases.

transformer_engine/pytorch/distributed.py
wdykas: oh this is a good idea. I just have to figure out how to recognize that the tensor is symmetric already. The problem is that the symm_mem.empty type is just torch.Tensor. An alternative could be to put a flag to use that tensor if we can't automatically check that its a symmetric tensor

denera: For our purposes, I think it's sufficient to check if the tensor is in `symmetric_mem_cache`. That cache should only ever contain symmetric memory tensors, identified by tensor metadata, process group, and a custom tag (in case some module wants to generate a symmetric memory tensor that won't be used by other all-reduces of the same size). We can be explicit in the documentation and specify that externally created symmetric memory tensors that are not in the TE cache will not be able to avoid the copy operation.

wdykas: ok cool I made the changes, let me know what you think. I think if pytorch makes an api or attribute available to recognize symmetric tensors we can support bypassing the cache all together just seems like it isn't exposed in any way right now. I made it a bit different than  your proposal. I had to use:
```
input_id = id(inp)
is_cached = any(id(cached_tensor) == input_id for cached_tensor in symmetric_mem_cache.values())
```
Because torch doesn't like the `if inp in symmetric_mem_cache.values()`

example error:
```
RuntimeError: The size of tensor a (536870912) must match the size of tensor b (8192) at non-singleton dimension 2
```

transformer_engine/pytorch/distributed.py
```diff
@@ -1234,6 +1241,150 @@ def gather_along_first_dim(
...
+
+    input_id = id(inp)
+    is_cached = any(id(cached_tensor) == input_id for cached_tensor in symmetric_mem_cache.values())
+    # Check if the input tensor is already in the symmetric memory cache. If it is we can avoid copy overheads.
+    if is_cached:
+        all_reduce_impl(
+            inp,
+            "sum",
+            group_name,
+        )
```
denera, line 1369: Does this work correctly if `inp` is not 1D?

In a real use case, we will likely call `get_symmetric_memory_tensor()` to get a buffer of the appropriate size, reshape it into what we need for the output buffer of a specific operation, and then do the all-reduce after the operation. That means the input tensor will frequently be a *view* of a symmetric memory tensor.

Can `all_reduce_impl()` handle that correctly, or does it always need a 1D input? We should test this and, if necessary, do the all-reduce operation over `inp.reshape(-1)` instead of `inp` directly (just like how it would be when we copy the input into a new symmetric memory tensor).

denera (approved):
LGTM, pending verification of correctness when input is a non-1D view of a symmetric memory tensor.

transformer_engine/pytorch/distributed.py
wdykas: I just tried and it seems to work from my test when sending in views and comparing the outputs. Should we store shaped symmetric tensors instead then? In the case right now we get buffer reuse if they are of different shapes but it requires that reshape.

denera: I think it's definitely preferable to store 1D symmetric tensors and re-use them for different shapes. I just wanted to make sure that the all-reduce implementation was not dependent on receiving 1D inputs, and it sounds like it's not, so there are no issues leaving it as is. We just need to make sure `get_symmetric_memory_tensor()` is documented to indicate that it always returns 1D tensors.

wdykas: ok cool I added a note to the function that it always returns a 1D tensor

wdykas: I tried doing the suggestions in the dco for signing but it messed up the entire MR. Is there an easier way to do this signing thing its asking for?

wdykas: sorry for the extra problems. I remade the merge request here https://github.com/NVIDIA/TransformerEngine/pull/1682/files If this one cannot be fixed

PR #1640, opened 2025-04-03
Add adam bf16 state with original fp32 kernel

BestJuly:
# Description

This PR is based on previous [PR#1465](https://github.com/NVIDIA/TransformerEngine/pull/1465).

The original description is 
> This PR supports Adam bf16 state to reduce memory usage, we have tested some smaller models or largest models (LLM models), and we get a similar convergence even with this bf16 lower precision state. For the deepseek-v3 report, they also used the bf16 state to reduce training memory usage.

The difference is that in this PR, we do not change the kernel, but we still use bf16 optimizer states to save memory.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/pytorch/optimizers/fused_adam.py
```diff
@@ -313,8 +317,11 @@ def get_unscaled_state(self, param, state_name):
             else:
                 assert state[state_name].dtype == torch.float32
             unscaled = state[state_name]
+        elif dtype == torch.bfloat16:
+            assert state[state_name].dtype == torch.bfloat16
+            unscaled = state[state_name].float()
```
yaox12, line 322: Since we're converting states to fp32 here, does it still save memory?

BestJuly: During the optimizer updating process, the backward pass has been completed and many tensors do not need to be stored any more. That is to say, it should not increase the peak memory usage in some conditions (unless it leads to another peak) so it should help reduce memory usage.

BestJuly: Add more memory consumption log here addressing this comment. I do observed larger memory consumption with 1 GPU or 2. But for DeepSeek-v3 FP8 cases (I use a proxy model similar to that), with this PR (together require another in Megatron), we do observe less memory usage.
```
# no bf16 optim
[Rank 0] (after 10 iterations) memory (MB) | allocated: 49160.2412109375 | max allocated: 64577.431640625 | reserved: 68684.0 | max reserved: 68684.0
 [2025-04-07 01:11:31] iteration       10/ 2288818 | consumed samples:         2560 | elapsed time per iteration (ms): 12436.4 | throughput per GPU (TFLOP/s/GPU): 412.7 | learning rate: 1.484024E-07 | global batch size:   256 | lm loss: 1.321985E+01 | load_balancing_loss: 9.884739E-01 | loss scale: 1.0 | grad norm: 5.136 | num zeros: 897426880.0 | params norm: 1860.172 | number of skipped iterations:   0 | number of nan iterations:   0 |
 
# bf16 optim
[Rank 0] (after 10 iterations) memory (MB) | allocated: 40797.1552734375 | max allocated: 56226.98583984375 | reserved: 61072.0 | max reserved: 61072.0
 [2025-04-07 01:14:41] iteration       10/ 2288818 | consumed samples:         2560 | elapsed time per iteration (ms): 11146.6 | throughput per GPU (TFLOP/s/GPU): 460.4 | learning rate: 1.484024E-07 | global batch size:   256 | lm loss: 1.322008E+01 | load_balancing_loss: 9.884660E-01 | loss scale: 1.0 | grad norm: 5.136 | num zeros: 897426944.0 | params norm: 1532.583 | number of skipped iterations:   0 | number of nan iterations:   0 |
```

yaox12: It's similar to this issue https://github.com/NVIDIA/Megatron-LM/issues/1477. When the number of GPUs is small, the peak memory usage could occur in optimizer, then casting to FP32 could lead to high memory consumption. If the number of GPUs is large, the peak memory usage usually occurs in loss calculation so casting to FP32 doesn't matter.

BestJuly: Yes, so this should not be a bug but makes sense in theory.    
One way to further reduce the the memory is to cast bf16 to fp32 inside the kernel instead of using `.float()`. @kunlunl mentioned that mcore directly uses [get_unscaled_state](https://github.com/NVIDIA/Megatron-LM/blob/819dd004ad521d4b7195a4cab9f16e8f8dc4f099/megatron/core/optimizer/distrib_optimizer.py#L825) function and the in checkpoint storage it should be float32 (refer to [comment](https://github.com/NVIDIA/TransformerEngine/pull/1465#discussion_r2000321892) in [PR#1465](https://github.com/NVIDIA/TransformerEngine/pull/1465), so I kept the current solution in this PR.

transformer_engine/pytorch/optimizers/fused_adam.py
```diff
@@ -259,6 +259,10 @@ def _apply_scale(self, state_name, unscaled_state, scaled_state, scale):
             scale (torch.Tensor): A FP32 tensor representing the scaling factor.
         """
         assert unscaled_state.dtype == torch.float32
+        if scaled_state.dtype == torch.bfloat16:
+            scaled_state.copy_(unscaled_state.bfloat16())
+            return
```
timmoon10, line 264: It's uncomfortable that FP16 involves scaling while BF16 does not (see similar complaint at https://github.com/NVIDIA/TransformerEngine/pull/1465). Even if this does the optimal behavior, it's not obvious to the user that changing the dtype would also change the compute. Perhaps in a future PR we should add an option for state scaling so that we disentangle the dtype from the algorithm logic.

timmoon10 (approved):
Overall LGTM. I don't like how state scaling and state dtype are entangled, but I suppose it's a preexisting problem and fixing it is beyond the scope of this PR.

PR #1642, opened 2025-04-03
Add experimental Shardy support.

jreiffers:
This is experimental and not intended for production use. If you try this, please file issues for problems you encounter.

# Description

Shardy is a new partitioning system in JAX. It currently replaces the sharding propagation passes in GSPMD.

This PR adds Shardy support for TE's attention, normalization, quantization and softmax primitives.

To use this, simply enable Shardy:

```
jax.config.update("jax_use_shardy_partitioner", True)
```

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Adds functions to provide Shardy sharding rules to primitives.

### DIFF
--- a/examples/jax/encoder/test_model_parallel_encoder.py
+++ b/examples/jax/encoder/test_model_parallel_encoder.py
@@ -258,6 +258,8 @@ def get_state_sharding(state, params_sharding):
 def train_and_evaluate(args):
     """Execute model training and evaluation loop."""
     print(args)
+    jax.config.update("jax_use_shardy_partitioner", args.enable_shardy)
+
     train_ds, test_ds, num_embed = get_datasets(args.max_seq_len)
 
     num_gpu = jax.local_device_count()
@@ -441,6 +443,9 @@ def encoder_parser(args):
     parser.add_argument(
         "--enable-sp", action="store_true", default=False, help="Enable sequence parallelism."
     )
+    parser.add_argument(
+        "--enable-shardy", action="store_true", default=False, help="Enable Shardy (experimental)."
+    )
 
     return parser.parse_args(args)
 
@@ -451,10 +456,9 @@ class TestEncoder(unittest.TestCase):
     is_fp8_supported, fp8_reason = is_fp8_available(ScalingMode.DELAYED_TENSOR_SCALING)
     is_mxfp8_supported, mxfp8_reason = is_fp8_available(ScalingMode.MXFP8_1D_SCALING)
 
-    @classmethod
-    def setUpClass(cls):
+    def setUp(self):
         """Run 3 epochs for testing"""
-        cls.args = encoder_parser(["--epochs", "3"])
+        self.args = encoder_parser(["--epochs", "3"])
 
     @unittest.skipIf(not is_bf16_supported(), "Device compute capability 8.0+ is required for BF16")
     def test_te_bf16(self):
@@ -502,6 +506,13 @@ class TestEncoder(unittest.TestCase):
         self.args.fp8_recipe = "MXFP8BlockScaling"
         actual = train_and_evaluate(self.args)
         assert actual[0] < 0.455 and actual[1] > 0.785
+
+    @unittest.skipIf(not is_bf16_supported(), "Device compute capability 8.0+ is required for BF16")
+    def test_te_bf16_shardy(self):
+        """Test Transformer Engine with BF16"""
+        self.args.enable_shardy = True
+        actual = train_and_evaluate(self.args)
+        assert actual[0] < 0.45 and actual[1] > 0.79
 
 
 if __name__ == "__main__":
--- a/examples/jax/encoder/test_model_parallel_encoder.py
+++ b/examples/jax/encoder/test_model_parallel_encoder.py
@@ -512,7 +512,28 @@ class TestEncoder(unittest.TestCase):
         """Test Transformer Engine with BF16"""
         self.args.enable_shardy = True
         actual = train_and_evaluate(self.args)
-        assert actual[0] < 0.45 and actual[1] > 0.79
+        assert actual[0] < 0.455 and actual[1] > 0.785
+
+    @unittest.skipIf(not is_fp8_supported, fp8_reason)
+    def test_te_delayed_scaling_fp8_shardy(self):
+        """Test Transformer Engine with DelayedScaling FP8"""
+        self.args.enable_shardy = True
+        self.args.use_fp8 = True
+        self.args.fp8_recipe = "DelayedScaling"
+        actual = train_and_evaluate(self.args)
+        assert actual[0] < 0.455 and actual[1] > 0.785
+
+    @unittest.skipIf(not is_fp8_supported, fp8_reason)
+    def test_te_delayed_scaling_fp8_with_sp_shardy(self):
+        """Test Transformer Engine with DelayedScaling FP8 + SP"""
+        self.args.enable_shardy = True
+        self.args.enable_sp = True
+        self.args.use_fp8 = True
+        self.args.fp8_recipe = "DelayedScaling"
+        actual = train_and_evaluate(self.args)
+        assert actual[0] < 0.455 and actual[1] > 0.785
+
+    # TODO(jreiffers): Add mxfp8 Shardy tests once supported in JAX.
 
 
 if __name__ == "__main__":
--- a/tests/jax/test_distributed_layernorm_mlp.py
+++ b/tests/jax/test_distributed_layernorm_mlp.py
@@ -399,4 +399,5 @@ class TestDistributedLayernormMLP:
             dtype,
             use_fp8=True,
             fp8_recipe=fp8_recipe,
+            use_shardy=False,
         )
--- a/tests/jax/test_distributed_layernorm_mlp.py
+++ b/tests/jax/test_distributed_layernorm_mlp.py
@@ -334,9 +334,18 @@ class TestDistributedLayernormMLP:
     @pytest_parametrize_wrapper("dtype", DTYPES)
     @pytest_parametrize_wrapper("use_bias", [True, False])
     @pytest_parametrize_wrapper("use_shardy", [False, True])
-    def test_layernorm_mlp_layer(self, mesh_config, activation_type, use_bias, input_shape, dtype, use_shardy):
+    def test_layernorm_mlp_layer(
+        self, mesh_config, activation_type, use_bias, input_shape, dtype, use_shardy
+    ):
         self._test_layernorm_mlp(
-            mesh_config, activation_type, use_bias, input_shape, dtype, use_fp8=False, fp8_recipe=None, use_shardy=use_shardy
+            mesh_config,
+            activation_type,
+            use_bias,
+            input_shape,
+            dtype,
+            use_fp8=False,
+            fp8_recipe=None,
+            use_shardy=use_shardy,
         )
 
     @pytest.mark.skipif(not is_fp8_supported, reason=reason)
--- a/tests/jax/test_distributed_layernorm_mlp.py
+++ b/tests/jax/test_distributed_layernorm_mlp.py
@@ -144,15 +144,7 @@ class TestDistributedLayernormMLP:
             )
         )
 
-    @pytest.mark.skipif(not is_fp8_supported, reason=reason)
-    @pytest_parametrize_wrapper("mesh_config", generate_fsdp_and_tp_configs())
-    @pytest_parametrize_wrapper("input_shape", INPUT_SHAPE)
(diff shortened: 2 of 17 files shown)


transformer_engine/jax/quantize/scaling_modes.py
```diff
@@ -165,6 +210,26 @@ def get_scale_shape(
...
+
+        Args:
+            input_rank: The rank of the input tensor (for which we produce the scale tensor)
+            unique_var: An otherwise unused Shardy variable name prefix
+
+        Returns:
+            The Shardy rules for the scaling mode
+        """
+        input_spec = [f"x{i}" for i in range(input_rank)]
+        input_spec[-2] = CompoundFactor(unique_var, "block_size")
```
phu0ngng, line 224: Hi, could you help me to understand why the `-2` axis of `input_spec` here is taken to embed with block scaling information?

transformer_engine/jax/quantize/scaling_modes.py
```diff
@@ -165,6 +210,26 @@ def get_scale_shape(
...
+
+        rowwise = input_spec[:-1] + [f"{unique_var}_"]
+        colwise = input_spec[:-2] + [unique_var, f"{unique_var}__"]
```
phu0ngng, line 227: Why `rowwise` and `colwise` have different specs?

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -534,6 +535,60 @@ def sharded_impl(x, scale, gamma, beta):
...
+            result_types,
+        )
+
+        scale_rules = ScalingMode(scaling_mode).get_shardy_sharding_rules(
+            len(value_types[0].shape), unique_var="i"
+        )
+        x_axes = scale_rules.input_spec
+
+        out = x_axes[:-1] + ("k",)
+        if is_2x and norm_type != NVTE_Norm_Type.RMSNorm:
```
phu0ngng, line 570: Why do we exclude `RMSNorm` here?

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -534,6 +535,60 @@ def sharded_impl(x, scale, gamma, beta):
...
+            colwise_out = out
+        else:
+            colwise_out = ("…4",)
```
phu0ngng, line 573: Does it mean `colwise_out` is the same as the sharding rule of `amax`? If yes, can we explicitly specify None sharding instead?

tests/jax/test_distributed_layernorm_mlp.py
```diff
@@ -142,9 +142,11 @@ def layernorm_fp8_mlp_prim_func(
     @pytest_parametrize_wrapper("dtype", DTYPES)
     @pytest_parametrize_wrapper("use_bias", [True, False])
     @pytest_parametrize_wrapper("fp8_recipe", SUPPORTED_RECIPES)
+    @pytest_parametrize_wrapper("use_shardy", [True, False])
```
phu0ngng, line 145: In general, we are trying to reduce the L0 unit test time as much as possible, and we wouldn't want the testing time to increase linearly with the number of added features.

Could we split all the shardy tests to separate tests, and instead of parameterizing all input combinations, we only test sharding with a subset of input parameters?

 Here, since different input shapes, data types, and activation types are already tested without shardy, we can test shardy with a single choice of input shape, data types, and gated activation.
 
 The same logic applies to other unit tests.

transformer_engine/jax/cpp_extensions/normalization.py
jreiffers: This is the equivalent of `None`, because `…4` is not used otherwise.

transformer_engine/jax/cpp_extensions/normalization.py
jreiffers: Good catch. I don't recall, might just be a bad rebase or other mess-up. I'll check.

transformer_engine/jax/quantize/scaling_modes.py
jreiffers: They have different shapes, so they require different specs in Shardy.

tests/jax/test_distributed_layernorm_mlp.py
jreiffers: Does the approach I took in `test_distributed_fused_attn` look OK to you? I didn't touch the other ones, because the number of tests didn't seem significant enough, but adding that is easy enough.

jreiffers:
PTAL. If this looks OK in principle, I'll do the rebase.

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -406,6 +407,58 @@ def sharded_impl(x, scale):
...
+        del out_dtype, act_enum, act_len, scale_dtype, scale_shapes, is_outer, mesh, result_types
+
+        x_rank = len(value_types[0].shape)
+        scale_rules = ScalingMode(scaling_mode).get_shardy_sharding_rules(
+            x_rank - 1, unique_var="i"
+        )
+        x_axes = scale_rules.input_spec + (f"x{x_rank-1}",)
+        out = (*x_axes[:-2], x_axes[-1])
+        if scaling_mode == ScalingMode.NVTE_MXFP8_1D_SCALING.value:
+            scale_inv = scale_rules.rowwise_rule[:-1] + (scale_rules.rowwise_rule[-1],)
```
phu0ngng, line 433: Hi, 
Isn't this equivalent to `scale_inv = scale_rules.rowwise_rule`?

Anyway, for ActLuPrimitive, we hardcoded the `flatten_axis=-2`, meaning that given an input shape `[M, 2, K]`, the rowwise_scale_inv.shape will be `[M, 2, K/32]` while the colwise_scale_inv.shape will be `[M/32, 2, K]`. I guess we need to update this accordingly somehow.

transformer_engine/jax/quantize/scaling_modes.py
```diff
@@ -217,6 +262,41 @@ def get_scale_shape(
...
+        """
+        bx = unique_var
+        by = f"{unique_var}_"
+
+        # We have to use two different factors in the two CompoundFactors because of Shardy
+        # verifier requirements, even though they are the same.
+        input_spec = [f"x{i}" for i in range(input_rank - 2)] + [
+            CompoundFactor(bx, "block_size_x"),
+            CompoundFactor(by, "block_size_y"),
+        ]
```
phu0ngng, line 283: We should take into account the `flatten_axis` here.

We quantize the [-1] axis for rowwise and [flatten - 1] axis for colwise.
For example, given a input_shape `dim0, dim1, dim2, dim3`, then:
```
flatten_axis    rowwise_scale_shape.                     colwise_scale_shape
-1                   [dim0, dim1, dim2, dim3/32]           [dim0, dim1, dim2/32, dim3] 
-2                   [dim0, dim1, dim2, dim3/32]           [dim0, dim1/32, dim2, dim3]
```

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -819,6 +872,44 @@ def sharded_impl(dz, x, scale):
...
+                colwise_out = tuple(x_axes)
+        else:
+            colwise_out = ("j",)
+
+        dbias = x_axes[-2:] if is_dbias else ("k",)
+        amax = ("…4",)
+
+        return SdyShardingRule(
+            (("…0",), tuple(x_axes), ("…2",)),
+            (out, colwise_out, scale_rules.rowwise_rule, scale_rules.colwise_rule, amax, dbias),
```
phu0ngng, line 909: Same for `DActLuDBiasQuantizePrimitive`, we hardcoded the flatten_axis=-2 and the last dim spec may not be None. So the `colwise_scale` and `rowwise_scale` may need to be reworked here.

transformer_engine/jax/cpp_extensions/activation.py
jreiffers: Yeah, this was a bad rebase. Thanks for noticing.

> Anyway, for ActLuPrimitive, we hardcoded the flatten_axis=-2, meaning that given an input shape [M, 2, K], the rowwise_scale_inv.shape will be [M, 2, K/32] while the colwise_scale_inv.shape will be [M/32, 2, K]. I guess we need to update this accordingly somehow.

Can you elaborate?

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -406,6 +407,58 @@ def sharded_impl(x, scale):
...
+            scale_inv = scale_rules.rowwise_rule[:-1] + (scale_rules.rowwise_rule[-1],)
+            colwise_scale_inv = scale_rules.colwise_rule[:-1] + (scale_rules.colwise_rule[-1],)
+        else:
+            scale_inv = scale_rules.rowwise_rule
+            colwise_scale_inv = scale_rules.colwise_rule
+
+        if is_2x:
+            if scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING.value:
+                colwise_out = tuple(
+                    multidim_transpose(x_axes, static_axis_boundary=-1, transpose_axis=-1)
```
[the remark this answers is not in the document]
jreiffers: Done.

transformer_engine/jax/quantize/scaling_modes.py
jreiffers: Done, thanks for the example!

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -406,6 +407,54 @@ def sharded_impl(x, scale):
...
+        is_outer,
+        mesh,
+        value_types,
+        result_types,
+    ):
+        del out_dtype, act_enum, act_len, scale_dtype, scale_shapes, is_outer, mesh, result_types
+
+        x_rank = len(value_types[0].shape)
+        scale_rules = ScalingMode(scaling_mode).get_shardy_sharding_rules(
+            x_rank - 1, unique_var="i", flatten_axis=-1
```
phu0ngng, line 428:
```suggestion
            x_rank - 1, unique_var="i", flatten_axis=-2
```

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -519,6 +520,57 @@ def sharded_impl(x, scale, gamma, beta):
...
+            out_dtype,
+            scale_dtype,
+            scale_shapes,
+            is_outer,
+            mesh,
+            result_types,
+        )
+
+        scale_rules = ScalingMode(scaling_mode).get_shardy_sharding_rules(
+            len(value_types[0].shape), unique_var="i"
```
phu0ngng, line 550:
```suggestion
            len(value_types[0].shape), unique_var="i", flatten_axis=-1
```

phu0ngng (changes requested):
The rest looks good to me!

tests/jax/test_distributed_layernorm_mlp.py
```diff
@@ -257,9 +251,59 @@ def test_layernorm_mlp_grad(
...
+        mesh_config,
+        activation_type,
+        use_bias,
+        input_shape,
+        dtype,
+        use_fp8,
+        fp8_recipe,
+        use_shardy,
     ):
+        jax.config.update("jax_use_shardy_partitioner", use_shardy)
```
phu0ngng, line 306: Hi, 

Here, I think we need to set `jax.config.update("jax_use_shardy_partitioner", False)` after the test.
Otherwise, later tests use Shardy, as it was turned on at one point but never turned off.

examples/jax/encoder/test_model_parallel_encoder.py
```diff
@@ -503,6 +507,13 @@ def test_te_mxfp8_with_sp(self):
         actual = train_and_evaluate(self.args)
         assert actual[0] < 0.455 and actual[1] > 0.785
 
+    @unittest.skipIf(not is_bf16_supported(), "Device compute capability 8.0+ is required for BF16")
+    def test_te_bf16_shardy(self):
+        """Test Transformer Engine with BF16"""
+        self.args.enable_shardy = True
+        actual = train_and_evaluate(self.args)
+        assert actual[0] < 0.45 and actual[1] > 0.79
```
phu0ngng, line 515:
```suggestion
        assert actual[0] < 0.45 and actual[1] > 0.79
```
```suggestion
        assert actual[0] < 0.455 and actual[1] > 0.785
```
So that it has the same tols as `test_te_bf16`. The test passes after relaxing the tols.

phu0ngng (approved):

phu0ngng:
Pipeline #26929185.

PR #1647, opened 2025-04-05
[DRAFT] JAX Current Scaling

jberchtold-nvidia:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/examples/jax/encoder/test_model_parallel_encoder.py
+++ b/examples/jax/encoder/test_model_parallel_encoder.py
@@ -447,8 +447,8 @@ def encoder_parser(args):
 class TestEncoder(unittest.TestCase):
     """Encoder unittests"""
 
-    is_fp8_supported, fp8_reason = is_fp8_available(ScalingMode.NVTE_DELAYED_TENSOR_SCALING)
-    is_mxfp8_supported, mxfp8_reason = is_fp8_available(ScalingMode.NVTE_MXFP8_1D_SCALING)
+    is_fp8_supported, fp8_reason = is_fp8_available(ScalingMode.DELAYED_TENSOR_SCALING)
+    is_mxfp8_supported, mxfp8_reason = is_fp8_available(ScalingMode.MXFP8_1D_SCALING)
 
     @classmethod
     def setUpClass(cls):
--- a/tests/jax/test_custom_call_compute.py
+++ b/tests/jax/test_custom_call_compute.py
@@ -48,21 +48,21 @@ FP8_COMPUTE_TYPE = [jnp.float8_e4m3fn, jnp.float8_e5m2]
 LN_CASES = [(256, 128), (128, 256)]
 DTYPES = [jnp.bfloat16, jnp.float32]
 is_fp8_supported, reason = helper.is_fp8_available()
-is_mxfp8_supported, reason = helper.is_fp8_available(ScalingMode.NVTE_MXFP8_1D_SCALING)
+is_mxfp8_supported, reason = helper.is_fp8_available(ScalingMode.MXFP8_1D_SCALING)
 
 supported_scaling_modes = []
 """ Find supported scaling modes"""
 if is_fp8_supported:
-    supported_scaling_modes.append(ScalingMode.NVTE_DELAYED_TENSOR_SCALING)
+    supported_scaling_modes.append(ScalingMode.DELAYED_TENSOR_SCALING)
 if is_mxfp8_supported:
-    supported_scaling_modes.append(ScalingMode.NVTE_MXFP8_1D_SCALING)
+    supported_scaling_modes.append(ScalingMode.MXFP8_1D_SCALING)
 
 
 def is_shape_supported_by_mxfp8(input_shape):
     try:
         if isinstance(input_shape, type(pytest.param(0))):
             input_shape = input_shape.values[0]
-        ScalingMode.NVTE_MXFP8_1D_SCALING.get_scale_shape_2x(input_shape)
+        ScalingMode.MXFP8_1D_SCALING.get_scale_shape_2x(input_shape)
         return True
     except:
         # get_scale_shapes will raise an exception if the shape is not supported
@@ -167,7 +167,7 @@ class TestActivation:
         )
 
         quantizer = QuantizerFactory.create(
-            scaling_mode=ScalingMode.NVTE_DELAYED_TENSOR_SCALING,
+            scaling_mode=ScalingMode.DELAYED_TENSOR_SCALING,
             q_dtype=output_type,
             q_axis=QuantizeAxis.ROWWISE,
         )
@@ -192,7 +192,7 @@ class TestActivation:
 
         te_quantizer, jax_quantizer = QuantizerFactory.create(
             n_quantizers=2,
-            scaling_mode=ScalingMode.NVTE_DELAYED_TENSOR_SCALING,
+            scaling_mode=ScalingMode.DELAYED_TENSOR_SCALING,
             q_dtype=output_type,
             q_axis=q_axis,
         )
@@ -215,7 +215,7 @@ class TestActivation:
         self.activation_type = activation_type
 
         quantizer = QuantizerFactory.create(
-            scaling_mode=ScalingMode.NVTE_MXFP8_1D_SCALING, q_dtype=output_type, q_axis=q_axis
+            scaling_mode=ScalingMode.MXFP8_1D_SCALING, q_dtype=output_type, q_axis=q_axis
         )
 
         output = tex.act_lu(x, activation_type, quantizer)
@@ -335,7 +335,7 @@ class TestNorm:
             pytest.skip("RMSNorm and zero_centered_gamma is not supported!")
 
         quantizer = QuantizerFactory.create(
-            scaling_mode=ScalingMode.NVTE_DELAYED_TENSOR_SCALING, q_dtype=out_dtype, q_axis=q_axis
+            scaling_mode=ScalingMode.DELAYED_TENSOR_SCALING, q_dtype=out_dtype, q_axis=q_axis
         )
         self._test_norm_grad(
             n, hidden, norm_type, zero_centered_gamma, epsilon, inp_dtype, quantizer
@@ -406,7 +406,7 @@ class TestNorm:
             epsilon=epsilon,
             inp_dtype=inp_dtype,
             out_dtype=out_dtype,
-            scaling_mode=ScalingMode.NVTE_DELAYED_TENSOR_SCALING,
+            scaling_mode=ScalingMode.DELAYED_TENSOR_SCALING,
             q_axis=q_axis,
         )
 
@@ -423,7 +423,7 @@ class TestNorm:
             epsilon=epsilon,
             inp_dtype=inp_dtype,
             out_dtype=out_dtype,
-            scaling_mode=ScalingMode.NVTE_MXFP8_1D_SCALING,
+            scaling_mode=ScalingMode.MXFP8_1D_SCALING,
             q_axis=QuantizeAxis.ROWWISE_COLWISE,
         )
 
@@ -475,7 +475,7 @@ class TestQuantize:
             q_axis=q_axis,
         )
 
-        n_iterations = 3 if scaling_mode == ScalingMode.NVTE_DELAYED_TENSOR_SCALING else 1
+        n_iterations = 3 if scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING else 1
         for _ in range(n_iterations):
             x = jax.random.uniform(key, input_shape, in_dtype)
 
@@ -483,7 +483,7 @@ class TestQuantize:
             assert_dequantized_scaled_tensor(scaled_tensor, x)
 
     def test_quantize_bitwise(self, in_dtype, input_shape, q_dtype, scaling_mode, q_axis):
-        if scaling_mode == ScalingMode.NVTE_MXFP8_1D_SCALING and not is_shape_supported_by_mxfp8(
+        if scaling_mode == ScalingMode.MXFP8_1D_SCALING and not is_shape_supported_by_mxfp8(
             input_shape
         ):
             pytest.skip(f"Input shape {input_shape} is not supported by MXFP8")
@@ -511,7 +511,7 @@ class TestFusedQuantize:
     @pytest_parametrize_wrapper("q_axis", [QuantizeAxis.ROWWISE, QuantizeAxis.ROWWISE_COLWISE])
     def test_quantize_dbias(self, in_dtype, input_shape, out_dtype, scaling_mode, q_axis):
         transpose_axis = -1
-        if scaling_mode == ScalingMode.NVTE_MXFP8_1D_SCALING and not is_shape_supported_by_mxfp8(
+        if scaling_mode == ScalingMode.MXFP8_1D_SCALING and not is_shape_supported_by_mxfp8(
             input_shape
         ):
             pytest.skip(f"Input shape {input_shape} is not supported by MXFP8")
@@ -594,7 +594,7 @@ class TestFusedQuantize:
             in_dtype=in_dtype,
             input_shape=input_shape,
             out_dtype=in_dtype,
-            scaling_mode=ScalingMode.NVTE_NO_SCALING,
+            scaling_mode=ScalingMode.NO_SCALING,
             activation_type=activation_type,
             is_dbias=is_dbias,
             q_axis=QuantizeAxis.ROWWISE,
@@ -613,7 +613,7 @@ class TestFusedQuantize:
             in_dtype=in_dtype,
             input_shape=input_shape,
             out_dtype=out_dtype,
-            scaling_mode=ScalingMode.NVTE_DELAYED_TENSOR_SCALING,
+            scaling_mode=ScalingMode.DELAYED_TENSOR_SCALING,
             activation_type=activation_type,
             is_dbias=is_dbias,
             q_axis=q_axis,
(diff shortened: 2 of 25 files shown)


examples/jax/encoder/test_model_parallel_encoder.py
```diff
@@ -447,8 +447,8 @@ def encoder_parser(args):
 class TestEncoder(unittest.TestCase):
     """Encoder unittests"""
 
-    is_fp8_supported, fp8_reason = is_fp8_available(ScalingMode.NVTE_DELAYED_TENSOR_SCALING)
-    is_mxfp8_supported, mxfp8_reason = is_fp8_available(ScalingMode.NVTE_MXFP8_1D_SCALING)
+    is_fp8_supported, fp8_reason = is_fp8_available(ScalingMode.DELAYED_TENSOR_SCALING)
```
phu0ngng, line 450: Do you have any reason for this renaming?

transformer_engine/jax/quantize/quantizer.py
```diff
@@ -342,6 +342,124 @@ def update(self, new_amax: jnp.ndarray):
         self.scale = self._compute_scale(amax_history, self.scale, self.q_dtype)
         self.amax_history = self._roll_and_reset_amax_history(amax_history)
 
+@register_pytree_node_class
+@dataclass
+class CurrentScaleQuantizer(Quantizer):
```
phu0ngng, line 347: Here, I think we can make `DelayedScaleQuantizer` inherit from `CurrentScaleQuantizer` as `DelayedScaleQuantizer` only has `amax_history` in addition to a different `update()` method.

jberchtold-nvidia: Good idea, I've updated to this

transformer_engine/jax/quantize/scaling_modes.py
```diff
@@ -83,6 +85,37 @@ def get_scale_shape(
         """
         del data_shape, is_colwise
         return (1,)
+    
+
+class CurrentScalingModeMetadataImpl(ScalingModeMetadataImpl):
```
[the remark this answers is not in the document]
jberchtold-nvidia: Good idea, I've instead made DelayedScalingMetadataImpl inherit from CurrentScalingMetadataImpl to match the pattern used above for DelayedScaleQuantizer and CurrentScaleQuantizer, but let me know if you prefer different here. Thanks!

transformer_engine/jax/quantize/quantizer.py
```diff
@@ -159,34 +160,26 @@ def get_scale_dtype(self):
...
     def tree_flatten(self):
         """Flatten the quantizer for JAX tree operations.
 
         Returns:
             Tuple of (children, aux_data) for tree operations
         """
-        children = (self.scale, self.amax_history)
+        children = ()
         aux_data = (self.q_dtype, self.scaling_mode, self.q_layout)
         return (children, aux_data)
```
phu0ngng, line 184: This function is the same as the one in the base class, thus, it is not needed.

transformer_engine/jax/quantize/helper.py
```diff
@@ -94,7 +94,7 @@ def _check_fp8_support(scaling_mode, gpu_id) -> Tuple[bool, str]:
         A tuple of (bool, str) indicating support and any error message
     """
     gpu_arch = get_device_compute_capability(gpu_id)
-    if scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING:
+    if scaling_mode in (ScalingMode.DELAYED_TENSOR_SCALING, ScalingMode.CURRENT_TENSOR_SCALING):
```
phu0ngng, line 97: Use `scaling_mode.value.is_tensor_scaling()` here.

transformer_engine/jax/csrc/extensions/activation.cpp
```diff
@@ -44,6 +44,10 @@ Error_Type ActLuFFI(cudaStream_t stream, Buffer_Type input_buf, Buffer_Type scal
   auto output_tensor = TensorWrapper(get_nvte_scaling_mode(scaling_mode));
   output_tensor.set_rowwise_data(output, static_cast<DType>(out_dtype), output_shape);
 
+  NVTE_CHECK(scaling_mode != JAXX_Scaling_Mode::CURRENT_TENSOR_SCALING,
+             "Current tensor scaling does not support fused operations. Please call this primitive "
```
phu0ngng, line 48:
```suggestion
             "Current tensor scaling does not support fused operations yet. Please call this primitive "
```

transformer_engine/jax/csrc/extensions/quantization.cpp
```diff
@@ -159,6 +161,21 @@ Error_Type DBiasQuantizeFFI(cudaStream_t stream, Buffer_Type input_buf, Buffer_T
     }
   }
 
+  if (scaling_mode == JAXX_Scaling_Mode::CURRENT_TENSOR_SCALING) {
+    nvte_compute_amax(input_tensor.data(),   // input data
+                      output_tensor.data(),  // output data (for amax)
+                      stream);
```
phu0ngng, line 167: If we have TP, then this amax is the amax of the sharded tensor. So, the loss curve with TP and without TP may be slightly different at first. Will this be ok @ptrendx?

Or should one do the amax reduction in JAX instead?

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -1056,6 +1088,20 @@ def quantize_dact_dbias(
             dbias = _jax_dbias(output, dtype=x.dtype, flatten_axis=-2)
         return output.astype(x.dtype), dbias
 
+    if quantizer.scaling_mode == ScalingMode.CURRENT_TENSOR_SCALING:
+        # Current scaling does not support fused operations. Perform dact in higher precision then quantize after.
+        out, _ = quantize_dact_dbias(
```
phu0ngng, line 1093: Use `dact_lu` here.

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -827,9 +836,26 @@ def layernorm_fwd(
...
+        out, mu, rsigma = layernorm_fwd(
+            x=x.astype(jnp.float32),
+            gamma=gamma.astype(jnp.float32),
+            beta=beta.astype(jnp.float32),
+            zero_centered_gamma=zero_centered_gamma,
+            epsilon=epsilon,
+            quantizer=None,
+        )
+        out = quantizer.quantize(out, dq_dtype=x.dtype)
+        # out,_ = _quantize_dbias_impl(out, is_dbias=False, quantizer=quantizer, dq_dtype=x.dtype)
```
phu0ngng, line 850:
```suggestion
```

tests/jax/test_custom_call_compute.py
```diff
@@ -448,8 +481,8 @@ def test_norm_forward_with_block_scaling_fp8(
 }
 
 ALL_QUANTIZE_TEST_SHAPES = [
-    (32, 64),
-    (2, 64, 32),
+    (32, 256, 128),
+    (64, 32, 32, 256),
```
phu0ngng, line 485: Any reason to change the size here?

I wanted to have small sizes to check if the MXFP8 code works correctly with scale padding.
In addition, I wanted to make sure it works with input.ndim == 2.

jberchtold-nvidia: This is unrelated to the current scaling. Quantize tests were failing when `flatten_axis=-2` because the we can't flatten on axis 0 with none to the left. This set of shapes is only run when test level is L2 so was probably missing from our L0 runs, I think this was an issue before this PR.

Ack that we want to test small shapes and ndim=2 for MXFP8, so I'll revert for now, which means L2 will still be broken but we can fix it in a follow-up, possibly skipping the flatten_axis=-2 tests when ndim=2

transformer_engine/jax/csrc/extensions/quantization.cpp
jberchtold-nvidia: Right now no parallelism is supported in this PR, it will come in a follow-up, and I added asserts in our primitives partitioning methods to say current scaling partitioning isn't supported instead of trying to partition and giving an incorrect result.

I also discussed with Przemek yesterday and he mentioned that we are supposed to compute per-device amax and scale, we aren't supposed to do a global reduced amax. However, this means we also have a per-device scale_inv for the tensor corresponding to the device's shard. This is what PyTorch is doing currently. But we're unclear how we can get a per-device scale_inv for each shard of the tensor in JAX that persists outside of our primitive and into the gemm.

jberchtold-nvidia: > Wait, if we perform this amax reduction in sharded tensor in CPP, then we will have the issue that we don't know how to reduce the scale_inv in the sharded_impl.
Yes, this is the core problem here. In PyTorch, they persist these device-local scale_invs through to the gemm as they apply the same sharding and keep the same shards and scale_invs on each device. They don't do any reduction of scale_inv. We're unsure if this is possible in JAX's sharding framework

transformer_engine/jax/csrc/extensions/activation.cpp
jberchtold-nvidia: Updated all occurrences of this check

transformer_engine/jax/quantize/helper.py
jberchtold-nvidia: Updated all checks of `in (ScalingMode.DELAYED_TENSOR_SCALING, ScalingMode.CURRENT_TENSOR_SCALING)`

transformer_engine/jax/quantize/quantizer.py
jberchtold-nvidia: Good point, removed

transformer_engine/jax/quantize/scaling_modes.py
```diff
@@ -226,12 +233,14 @@ class ScalingMode(Enum):
...
     - DELAYED_TENSOR_SCALING: Uses delayed scaling with FP8 data type and float32 scales
     - MXFP8_1D_SCALING: Uses block-based scaling with FP8 data type and E8M0 scales
+    - CURRENT_TENSOR_SCALING: Uses current scaling with FP8 data type and float32 scales
     - NO_SCALING: No scaling applied
     """
 
     NO_SCALING = JAXX_Scaling_Mode.NO_SCALING
     DELAYED_TENSOR_SCALING = JAXX_Scaling_Mode.DELAYED_TENSOR_SCALING
     MXFP8_1D_SCALING = JAXX_Scaling_Mode.MXFP8_1D_SCALING
+    CURRENT_TENSOR_SCALING = JAXX_Scaling_Mode.CURRENT_TENSOR_SCALING
```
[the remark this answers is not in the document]
jberchtold-nvidia: Done

transformer_engine/common/normalization/layernorm/ln_api.cpp
```diff
@@ -31,19 +31,23 @@ void layernorm_fwd(const Tensor& x,      // BxSxhidden_size
...
   }
 
-  NVTE_CHECK(x.data.shape.size() == 2);
-  NVTE_CHECK(gamma.data.shape == beta.data.shape);
-  NVTE_CHECK(x.data.shape[1] == gamma.data.shape[0]);
+  NVTE_CHECK(x.data.shape.size() == 2, "x must be 2D tensor.");
+  NVTE_CHECK(gamma.data.shape == beta.data.shape, "Gamma and Beta must have the same shape.");
+  NVTE_CHECK(x.data.shape[1] == gamma.data.shape[0], "Gamma must have the same hidden size.");
+
+  NVTE_CHECK(epsilon >= 0.f, "Epsilon must be non-negative.");
```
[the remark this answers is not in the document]
jberchtold-nvidia: Good idea, updated

transformer_engine/jax/cpp_extensions/activation.py
jberchtold-nvidia: Done

transformer_engine/jax/cpp_extensions/normalization.py
jberchtold-nvidia: Done

tests/jax/test_custom_call_compute.py
jberchtold-nvidia: Reverted for now

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -1022,10 +1065,10 @@ def rmsnorm_fwd(
         _,
         rsigma,
     ) = NormFwdPrimitive.outer_primitive.bind(
-        x,
+        x.astype(jnp.float32),
```
jberchtold-nvidia, line 1068: After casting gamma and beta to fp32 in `_jax_*norm`, it fixes issues with current scaling where a small number of values would be mismatches due to zero-centered gamma addition being done in fp16.

However, this caused the norm forward block scaling tests to fail with a similar mismatch, only when zero-centered gamma is enabled. I suspect this may be a difference in background between delayed/current scaling and MXFP8, or potentially TE vs. CuDNN as mxfp8 forces cudnn impl regarding whether it decides to shift gamma by 1.0 in the incoming gamma dtype or if it's fused in the norm operation as float32.

Making a note of this here to discuss, as I don't think casting to float32 is ideal here because it can't be fused before going into our custom primitive.

phu0ngng: Hi @jberchtold-nvidia, 

I think PR#1700 should resolve this issue so that we no longer need to explicitly cast the NormFwdPrimitive inputs to float32. 
Please rebase and confirm.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -378,7 +382,10 @@ def grouped_gemm(
             rhs_shape = rhs.data.shape
             out_dtype = lhs.dq_dtype
             # For ScaledTensors and DELAYED_TENSOR_SCALING, need to handle internal data_layout
-            if lhs.scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING:
+            if lhs.scaling_mode in (
+                ScalingMode.DELAYED_TENSOR_SCALING,
+                ScalingMode.CURRENT_TENSOR_SCALING,
+            ):
```
phu0ngng, line 388:
```suggestion
            if lhs.scaling_mode.is_tensor_scaling():
```

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -406,7 +413,10 @@ def grouped_gemm(
         if scaling_mode == ScalingMode.NO_SCALING:
             lhs_3d = _shape_normalization(lhs, lhs_dn)
             rhs_3d = _shape_normalization(rhs, rhs_dn)
-        elif scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING:
+        elif scaling_mode in (
+            ScalingMode.DELAYED_TENSOR_SCALING,
+            ScalingMode.CURRENT_TENSOR_SCALING,
+        ):
```
phu0ngng, line 419:
```suggestion
        elif scaling_mode.is_tensor_scaling():
```

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -913,9 +923,25 @@ def layernorm_fwd(
...
+            epsilon=epsilon,
+            quantizer=None,
+        )
+        out, _ = _quantize_dbias_impl(out, is_dbias=False, quantizer=quantizer, dq_dtype=x.dtype)
+        return out, mu, rsigma
+
     is_2x2x = quantizer.is_2x2x()
     # TE/common normalization doesn't support 2x delayed scaling
-    if quantizer.is_2x2x() and quantizer.scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING:
+    if quantizer.is_2x2x() and quantizer.scaling_mode in (
```
phu0ngng, line 941: `is_tensor_scaling()` here and the rest.

transformer_engine/jax/quantize/quantizer.py
```diff
@@ -217,24 +200,26 @@ def _quantize_func(
...
             dq_dtype: Data type for dequantized values
-            flatten_axis: The quantization axis for the tensor
+
         Returns:
             A ScaledTensor1x containing the quantized data
         """
         dq_dtype = dq_dtype if dq_dtype is not None else x.dtype
 
-        compute_dtype = self.scale.dtype
+        compute_dtype = self.scaling_mode.get_scale_dtype()
```
phu0ngng, line 209: Just notice that we take `compute_dtype = scale.dtype` here.
I think we should always have `compute_dtype = jnp.float32`.

jberchtold-nvidia: Sounds good, updated for both delayed scaling and current scaling

transformer_engine/jax/cpp_extensions/normalization.py
jberchtold-nvidia: 👍

transformer_engine/jax/cpp_extensions/gemm.py
jberchtold-nvidia: Done

transformer_engine/jax/cpp_extensions/gemm.py
jberchtold-nvidia: Done

transformer_engine/jax/cpp_extensions/normalization.py
jberchtold-nvidia: Rebased and confirmed it works, just has slightly larger tolerances due to the intermediate storage in in_dtype (e.g. bf16) for the unfused norm + quantize. This is the same precision as the unfused norm + quantize we used when CuDNN is below a certain version. 

I've updated the norm tests for current scaling to reflect this and added a comment documenting the reason

PR #1653, opened 2025-04-08
Split wgrad&dgrad from backward() to support a2a overlap

lhb8125:
# Description

Add a flag split_bw to control if we should separate wgrad from backward() and schedule it in another function to better hide the a2a communication when training moe models.
This MR is to support the 1f1b with a2a overlap in MCore, similar with the idea in DualPipe.
This feature has an assertion:
```
ub_bulk_wgrad == False
```
because the knob will bind the output of wgrad with dgrad , which complicates the computing context of wgrad.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add class WeightGradStore to put and pop the context of wgrad computation;
- Wrap&store the wgrad computation of class Linear/LayernormLinear/GroupedLinear and pop it in wgrad_comp();
- Add some unit tests in test_numerics.py;

# Checklist:


transformer_engine/pytorch/module/linear.py
```diff
@@ -1262,3 +1274,24 @@ def _customize_quantizers_float8_current_scaling(self, fwd: bool, recipe: Recipe
                 self.quantizers["scaling_bwd"][
                     tex.FP8BwdTensors.GRAD_OUTPUT1
                 ].amax_reduction_group = self.tp_group
+
+    def wgrad_comp(self):
```
denera, line 1278: This is identical boilerplate code between `Linear`, `LayerNormLinear` and `GroupedLinear`. I think we should stick it into [`TransformerEngineBaseModule`](https://github.com/NVIDIA/TransformerEngine/blob/a8f0fe03b4ac801ee50b35ff24ec2998eaa301ac/transformer_engine/pytorch/module/base.py#L410) and let derived modules specialize it further if necessary.

On that note, `LayerNormMLP` is missing the same split BW implementation. Unless there's a reason why it should be avoided, we should add it for the sake of feature parity across modules even if there is no current/active use-case for it.

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -1452,3 +1468,24 @@ def _customize_quantizers_float8_current_scaling(self, fwd: bool, recipe: Recipe
             self.quantizers["scaling_bwd"][
                 tex.FP8BwdTensors.GRAD_OUTPUT1
             ].amax_epsilon = recipe.fp8_quant_bwd_grad.amax_epsilon
+
+    def wgrad_comp(self):
```
denera, line 1472: See comment about this method in `linear.py`.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -677,3 +692,25 @@ def forward(
         if self.return_bias:
             return out, [cast_if_needed(b, self.activation_dtype) for b in bias_tensors]
         return out
+
+    def wgrad_comp(self):
```
denera, line 696: See comment about this method in `linear.py`.

transformer_engine/pytorch/module/linear.py
```diff
@@ -1262,3 +1274,24 @@ def _customize_quantizers_float8_current_scaling(self, fwd: bool, recipe: Recipe
...
+        """
+        Execute the delayed weight gradient computation.
+        This method is called after the main backward pass to compute weight gradients.
+        """
+        if not self.wgrad_store.split_bw():
+            return
+        with torch.cuda.nvtx.range("_Linear_wgrad"):
+            (wgrad, grad_bias_, _, _), _ = self.wgrad_store.pop()
+            if not self.fuse_wgrad_accumulation:
```
denera, line 1287: Unless I'm misreading the code, fused wgrad accumulation is incompatible with split BW, so there should never be a situation where `self.wgrad_store.split_bw() == self.fuse_wgrad_accumulation == True`. However, I don't see an assert for this in the module initialization. We should add one with a useful error message.

denera (changes requested):

transformer_engine/pytorch/module/linear.py
lhb8125: I think they are compatible. When fused wgrad accumulation is true, the main grad of parameter will be updated in wgrad_comp() function automatically, which indeed is incompatible with overlap_grad_reducescatter=true, and we add an assertion in mcore code.

transformer_engine/pytorch/module/linear.py
lhb8125: Make sense, I apply the changes in the latest commit.

transformer_engine/pytorch/module/layernorm_linear.py
lhb8125: Done.

transformer_engine/pytorch/module/grouped_linear.py
lhb8125: Done.

transformer_engine/pytorch/module/linear.py
lhb8125: BTW, where can I check the CI status? I noticed that I was granted the Developer permission of TE gitlab repo.

transformer_engine/pytorch/module/linear.py
denera: Just went back over this, it's my bad, I mistakenly thought there was an exclusion in the fwd/bwd implementation that wasn't being captured in the split wgrad.

transformer_engine/pytorch/module/base.py
```diff
@@ -1104,3 +1104,21 @@ def _load_from_state_dict(
...
+            unfused_weights = [getattr(self, name) for name in self.weight_names]
+            weight_tensor = noop_cat(unfused_weights)
+            if weight_tensor.grad is None:
+                weight_tensor.grad = wgrad.to(weight_tensor.dtype)
+        if self.use_bias:
+            bias_tensor = noop_cat([getattr(self, name) for name in self.bias_names])
+            if bias_tensor.grad is None:
+                bias_tensor.grad = grad_bias_.to(bias_tensor.dtype)
+        del grad_bias_
+        del wgrad
```
denera, line 1124:
```suggestion
        if self.wgrad_store is None or not self.wgrad_store.split_bw():
            return
        with torch.cuda.nvtx.range(f"_{self.__name__}_wgrad"):
            (wgrad, grad_bias_, _, _), _ = self.wgrad_store.pop()
            if not self.fuse_wgrad_accumulation:
                unfused_weights = [getattr(self, name) for name in self.weight_names]
                weight_tensor = noop_cat(unfused_weights)
                if weight_tensor.grad is None:
                    weight_tensor.grad = wgrad.to(weight_tensor.dtype)
            if self.use_bias:
                bias_tensor = noop_cat([getattr(self, name) for name in self.bias_names])
                if bias_tensor.grad is None:
                    bias_tensor.grad = grad_bias_.to(bias_tensor.dtype)
            del grad_bias_
            del wgrad
```

Optionally, this should be able to eliminate `def backward_dw()` from both `Linear` and `LayerNormLinear`.

denera (approved):
LGTM!

transformer_engine/pytorch/module/base.py
lhb8125: Good point!

lhb8125:
@denera I added unit tests in PyTorch distributed tests, please review them.
BTW, I notice some unittests failed like cpp_unittest and fp8-cs/fp8-blockscaling in distributed/test_numerics.py, while this PR does not introduce those because I also observe same fails on main branch. Is that a known issue?

ptrendx:
What about LayerNormMLP? Since the functionality is advertised as general in all TE modules LNMLP should also be changed.

transformer_engine/pytorch/module/linear.py
```diff
@@ -881,6 +889,7 @@ def __init__(
         ub_bulk_dgrad: bool = False,
         ub_bulk_wgrad: bool = False,
         ub_name: Optional[str] = None,
+        split_bw: bool = False,
```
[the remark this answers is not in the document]
ksivaman: Same comment for all modules

lhb8125:
> What about LayerNormMLP? Since the functionality is advertised as general in all TE modules LNMLP should also be changed.

@ptrendx LNMLP has already been changed to delay the wgrad computation.

transformer_engine/pytorch/module/base.py
```diff
@@ -1140,6 +1140,27 @@ def _load_from_state_dict(
             state_dict, prefix, local_metadata, strict, missing_keys, unexpected_keys, error_msgs
         )
 
+    def backward_dw(self):
+        """
+        Execute the delayed weight gradient computation.
+        This method is called after the main backward pass to compute weight gradients.
```
denera, line 1146: _**Note:** This is an optional recommendation we can address in a future PR if the functionality is time sensitive._

The usage intent of `.backward_dw()` being called manually after `loss.backward()` requires users to search through their model for TE sub-modules like:
```python
y = model(x)
loss = cost_fn(y)
loss.backward()
for module in model.modules():
    if hasattr(module, "backward_dw"):
        module.backward_dw()
optimizer.step()
```

We can provide a cleaner and more intuitive user experience via a context function like:
```python
y = model(x)
loss = cost_fn(y)
with te.delayed_wgrad(enabled=True, exclude=None):
    loss.backward()
optimizer.step()
```

In this scenario, we no longer need TE modules to be initialized with a `split_bw` or `delayed_wgrad_compute` flag. Instead, TE modules or module types within the context that are not in the exclude list can add themselves to a global list of delayed wgrad computations when PyTorch autograd executes through our custom backward pass. We can then trigger `.backward_dw()` on members of this list when exiting the context after the native PyTorch `.backward()` is complete.

It would be great if @ksivaman and/or @ptrendx can chime in on this with their thoughts as well.

ksivaman: I like the idea of having an easier user experience, but a couple of thoughts:
1. How is the `exclude` list being created here? If the user has to filter through the model and find relevant TE modules to exclude then it defeats the purpose of this change. This is fine if mostly we want `exclude=None` but that removes the per module control that the user has with the current functionality.
2. With this change the user no longer has the granular control of when to call wgrad for different modules and instead it falls upon TE to do it all at once back-to-back during the context exit. I'm not sure if this is something that is desired? @lhb8125

denera: > 1. How is the `exclude` list being created here? If the user has to filter through the model and find relevant TE modules to exclude then it defeats the purpose of this change. This is fine if mostly we want `exclude=None` but that removes the per module control that the user has with the current functionality.

I would propose supporting three different forms of excludes here:
1. Exclude by type, e.g. `te.delayed_grad(enabled=True, exclude=[te.LayerNormLinear, te.LayerNormMLP])` where TE modules check for `any(isinstance(self, excluded_type) for excluded_type in delayed_wgrad_excludes)`.
2. Exclude by specific module, e.g. `te.delayed_grad(enabled=True, exclude=[layer1, layer2, layer3])` where TE modules check for `self in delayed_wgrad_excludes`.
3. Exclude by user function, e.g. `te.delayed_grad(enabled=True, exclude=disable_delayed_wgrad)` where TE modules check for `disable_delayed_wgrad(self)`

You are correct that constructing the module list in option 2 and the exclude function in option 3 involves some kind of filtering through the model, but I think even then there are clear advantages to the context:
- Filtering is no longer mandatory.
- Boilerplate code for basic filtering is absorbed into TE.
- Avoids recursive `model.module()` calls for users who either don't need filtering or want to filter by module type.
- Makes functional filtering a more user friendly.

> 2. With this change the user no longer has the granular control of when to call wgrad for different modules and instead it falls upon TE to do it all at once back-to-back during the context exit. I'm not sure if this is something that is desired? @lhb8125

This proposed change does not eliminate `.backward_dw()` from the modules, so it can still be called directly.

We can make this use-case easier too by implementing the context as a `contextlib.AbstractContextManager` object instead of a `@contextlib.contextmanager` decorated function. This would allow advanced users to invoke the context entry but skip the context exit in favor of manual control over when to call `.backward_dw()` for TE modules.

ksivaman: As discussed with @lhb8125 offline, there is indeed a need for call `.backward_dw` manually in a fine-grained manner depending on the pipeline schedule, so we'll defer this for future work.

transformer_engine/pytorch/module/linear.py
```diff
@@ -861,6 +869,8 @@ class Linear(TransformerEngineBaseModule):
                   it controls the type used to allocate the initial parameters. Useful when
                   the model is trained with lower precision and the original FP32 parameters
                   would not fit in GPU memory.
+    delay_wgrad_compute : bool, default = `False`
+                         Whether to delay weight gradient computation
```
ksivaman, line 873:
```suggestion
    delay_wgrad_compute : bool, default = `False`
                         Whether or not to delay weight gradient computation. If set to `True`,
                         it's the user's responsibility to call `module.backward_dw` to compute
                         weight gradients.
```

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -992,6 +1004,8 @@ class LayerNormLinear(TransformerEngineBaseModule):
                   it controls the type used to allocate the initial parameters. Useful when
                   the model is trained with lower precision and the original FP32 parameters
                   would not fit in GPU memory.
+    delay_wgrad_compute : bool, default = `False`
+                         Whether to delay weight gradient computation
```
ksivaman, line 1008:
```suggestion
    delay_wgrad_compute : bool, default = `False`
                         Whether or not to delay weight gradient computation. If set to `True`,
                         it's the user's responsibility to call `module.backward_dw` to compute
                         weight gradients.
```

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -1296,6 +1327,8 @@ class LayerNormMLP(TransformerEngineBaseModule):
                      batch size per training step. Needed for JIT Warmup, a technique where jit
                      fused functions are warmed up before training to ensure same kernels are
                      used for forward propogation and activation recompute phase.
+    delay_wgrad_compute : bool, default = `False`
+                         Whether to delay weight gradient computation
```
ksivaman, line 1331:
```suggestion
    delay_wgrad_compute : bool, default = `False`
                         Whether or not to delay weight gradient computation. If set to `True`,
                         it's the user's responsibility to call `module.backward_dw` to compute
                         weight gradients.
```

ksivaman (approved):

ksivaman (approved):

PR #1655, opened 2025-04-08
[JAX] Scaling Enum Abstracting

phu0ngng:
# Description

This PR discoupled the Scaling Enum value in JAX extension from the Scaling Enum value from TE/Common.

In addition, I also updated part of the NormPrimitive so that it is consistent with others. 

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/activation.py
+++ b/transformer_engine/jax/cpp_extensions/activation.py
@@ -162,7 +162,7 @@ class ActLuPrimitive(BasePrimitive):
         assert scale_aval is None or scale_aval.dtype == jnp.float32
 
         out = ffi.ffi_lowering(ActLuPrimitive.name)(
-            ctx, x, scale, act_enum=act_enum, scaling_mode=scaling_mode, is_2x=is_2x
+            ctx, x, scale, act_enum=act_enum, scaling_mode=scaling_mode.value, is_2x=is_2x
         )
         return out
 
@@ -545,7 +545,7 @@ class DActLuDBiasQuantizePrimitive(BasePrimitive):
             dz,
             x,
             scale,
-            scaling_mode=scaling_mode,
+            scaling_mode=scaling_mode.value,
             is_2x=is_2x,
             is_dbias=is_dbias,
             act_enum=int(act_enum),
@@ -928,7 +928,7 @@ def act_lu(
             out_dtype=x.dtype,
             act_enum=act_type_id,
             act_len=act_len,
-            scaling_mode=ScalingMode.NVTE_DELAYED_TENSOR_SCALING.value,
+            scaling_mode=ScalingMode.NVTE_NO_SCALING.value,
             is_2x=False,
             scale_dtype=jnp.float32,
             scale_shapes=((), ()),
@@ -1042,7 +1042,7 @@ def quantize_dact_dbias(
             # outputs float32 for dbias accumulation
             out_dtype=(jnp.float32 if is_dbias else x.dtype),
             # default value for no scaling, TE/common ignore this value when scale is unset
-            scaling_mode=ScalingMode.NVTE_DELAYED_TENSOR_SCALING.value,
+            scaling_mode=ScalingMode.NVTE_NO_SCALING.value,
             is_2x=False,  # unused
             scale_dtype=jnp.float32,  # unused
             scale_shapes=((), ()),  # unused
--- a/transformer_engine/jax/cpp_extensions/activation.py
+++ b/transformer_engine/jax/cpp_extensions/activation.py
@@ -282,7 +282,7 @@ class ActLuPrimitive(BasePrimitive):
         out_sharding = NamedSharding(mesh, PartitionSpec(*out_spec), desc="ActLuPrimitive.out")
 
         if is_2x:
-            if scaling_mode == ScalingMode.NVTE_DELAYED_TENSOR_SCALING.value:
+            if scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING.value:
                 colwise_out_spec = multidim_transpose(out_spec, transpose_axis=-1)
             else:
                 colwise_out_spec = out_spec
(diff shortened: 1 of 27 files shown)


transformer_engine/jax/csrc/extensions/misc.h
```diff
@@ -40,5 +40,28 @@ enum class QuantizeLayout {
   ROWWISE_COLWISE,
 };
 
+enum class JAXX_Scaling_Mode : int64_t {
```
ksivaman, line 43: Is this a typo propagated everywhere or is it meant to be `JAXX`?

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -1042,7 +1042,7 @@ def quantize_dact_dbias(
             # outputs float32 for dbias accumulation
             out_dtype=(jnp.float32 if is_dbias else x.dtype),
             # default value for no scaling, TE/common ignore this value when scale is unset
-            scaling_mode=ScalingMode.NVTE_DELAYED_TENSOR_SCALING.value,
+            scaling_mode=ScalingMode.NVTE_NO_SCALING.value,
```
jberchtold-nvidia, line 1045: Nice! 👍

transformer_engine/jax/csrc/extensions/misc.h
jberchtold-nvidia, line 43: Why `int64_t` vs. `int32_t`? Not an issue just curious because I don't think we'll ever hit 2 billion scaling modes

transformer_engine/jax/csrc/extensions/quantization.cpp
```diff
@@ -27,10 +29,37 @@ pybind11::tuple GetDBiasQuantizeWorkspaceSizes(size_t batch_size, size_t hidden_
   int temp = 0;
 
   auto input_tensor = TensorWrapper(reinterpret_cast<void *>(&temp), input_shape, in_dtype);
-  auto output_tensor = TensorWrapper(reinterpret_cast<void *>(&temp), output_shape, out_dtype);
-  output_tensor.set_columnwise_data(reinterpret_cast<void *>(&temp), out_dtype, output_trans_shape);
   auto dbias_tensor = TensorWrapper(reinterpret_cast<void *>(&temp), dbias_shape, in_dtype);
 
+  auto output_tensor = TensorWrapper(get_nvte_scaling_mode(scaling_mode));
+  // Only the pointers will be checked for scale_inv, thus the shapes do not matter
```
jberchtold-nvidia, line 35: 👍

transformer_engine/jax/quantize/quantizer.py
```diff
@@ -172,7 +172,7 @@ class DelayedScaleQuantizer(Quantizer):
         amax_history: History of maximum absolute values
     """
 
-    scaling_mode: ScalingMode = ScalingMode.NVTE_DELAYED_TENSOR_SCALING
+    scaling_mode: ScalingMode = ScalingMode.NVTE_DELAYED_TENSOR_SCALING.value
```
jberchtold-nvidia, line 175: Why is the `.value` added here? 

My understanding is this would make something like the following incorrect, but let me know if I'm mistaken. Thanks!

in diff of `rmsnorm_fwd` above
```scaling_mode=quantizer.scaling_mode.value,```

The types would be
```
quantizer.scaling_mode: JAXX_Scaling_Mode
quantizer.scaling_mode.value: int
```

Then inside norm lowering we have
```
        return ffi.ffi_lowering(NormFwdPrimitive.name)(
            ctx,
            x,
            scale,
            gamma,
            beta,
            norm_type=norm_type.value,
            zero_centered_gamma=zero_centered_gamma,
            epsilon=epsilon,
            sm_margin=sm_margin,
            scaling_mode=scaling_mode.value, <------ this would be (some int).value which would error, right?
            is_2x=is_2x,
        )
```

And any comparisons inside the norm primitive of `scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING.value` would be `False` because `scaling_mode` is of type `int` and `ScalingMode.DELAYED_TENSOR_SCALING.value` is of type `JAXX_Scaling_Mode` Python enum

File: transformer_engine/jax/quantize/scaling_modes.py [hunk 5]
```diff
@@ -304,7 +304,7 @@ def __eq__(self, other):
         """
         if not isinstance(other, ScalingMode):
             return False
-        return self.value == other.value
+        return self.value == other.value or self == other.value or self.value == other
```
jberchtold-nvidia, line 307: I don't know if this will ever be `True` because we're checking above that `isinstance(other, ScalingMode)`. 

So if we had the following check:
```
scaling_mode = ScalingMode.DELAYED_TENSOR_SCALING
...
if scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING.value:
   ...
```

Then `other` is `ScalingMode.DELAYED_TENSOR_SCALING.value` which is of type `JAXX_Scaling_Mode` which would mean we would have already early exited above because `other` isn't of type `ScalingMode`, right?

jberchtold-nvidia: Though I do like the idea of this check which would make us less prone to bugs where we either added and extra `.value` or missed a `.value` and the condition was never hit. I've had that happen a couple times when working on a feature

transformer_engine/jax/csrc/extensions/misc.h
phu0ngng: This enum is used by the XLA auto-decoding.
So similar to other `int` attributes that we try to lower from JAX, it needs to be in `int64_t`

transformer_engine/jax/quantize/scaling_modes.py
phu0ngng: Same here. Let me revert it. Thanks.

PR #1658, opened 2025-04-08
[Draft][JAX] grouped_gemm() uses variadic arguments

huanghua1994:
# Description

This PR optimizes the `grouped_gemm()` implementation in JAX. The original implementation manually flattens all input matrices before lowering to C++ function and manually split the output into a list of tensors. Using variadic arguments allows the code to avoid extra copying of inputs and outputs.

This PR is marked as a draft since PR #1545 breaks both the original implementation and this implementation for MXFP8 on Blackwell.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

Rewrite `grouped_gemm()` and `GroupedGemmPrimitive` using variadic arguments.

# Checklist:


transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -79,10 +56,40 @@ Error_Type GroupedGemmImpl(uint8_t *lhs_ptr, const DType &lhs_dtype, uint8_t *lh
   std::vector<NVTETensor> out_list;
   std::vector<NVTETensor> workspace_list;
 
+  int lhs_list_offset = 0;
+  int rhs_list_offset = num_gemms;
+  int lhs_sinv_list_offset = 2 * num_gemms;
+  int rhs_sinv_list_offset = 3 * num_gemms;
+  int bias_list_offset = 4 * num_gemms;
+  int out_list_offset = 0;
```
phu0ngng, line 64: We should check if we receive enough number of buffers for the given `num_gemms` before calculating the offset.

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -107,37 +114,34 @@ Error_Type GroupedGemmImpl(uint8_t *lhs_ptr, const DType &lhs_dtype, uint8_t *lh
       rhs_sinv_shape[1] = sinv_k;
```
phu0ngng, line 115: The scaling_shape calculation here is incorrect when it has padding.
I think it's better to take the dimension information from `lhs_sinv_i->dimension()`.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -41,93 +41,53 @@ class GroupedGemmPrimitive(BasePrimitive):
...
-        del bias_contig_aval, dim_list_aval
-        del num_gemms, scaling_mode
-        out_flat_aval = jax.core.ShapedArray(shape=(out_flat_size,), dtype=out_dtype)
-        wkspace_size = get_cublas_workspace_size_bytes() * num_cublas_streams
-        wkspace_aval = jax.core.ShapedArray(shape=(wkspace_size,), dtype=jnp.uint8)
-        return (out_flat_aval, wkspace_aval)
+    def abstract(*args, num_gemms, scaling_mode, out_dtype, has_bias):
+        del scaling_mode, has_bias
+        A_list = args[0:num_gemms]
+        B_list = args[num_gemms : 2 * num_gemms]
```
phu0ngng, line 52: Here, I think we should also check if the `len(args)` matches the expected input length for the `num_gemms`.
Besides, let's add a comment here on what we expect in `args` and its layout.

transformer_engine/jax/csrc/extensions/gemm.cpp
huanghua1994: Fixed

transformer_engine/jax/cpp_extensions/gemm.py
huanghua1994: Fixed

transformer_engine/jax/csrc/extensions/gemm.cpp
huanghua1994: Fixed

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -79,63 +67,94 @@ Error_Type GroupedGemmImpl(uint8_t *lhs_ptr, const DType &lhs_dtype, uint8_t *lh
...
-      lhs_i.set_rowwise_scale_inv(static_cast<void *>(lhs_sinv_ptr), DType::kFloat8E8M0,
-                                  lhs_sinv_shape);
-      rhs_i.set_rowwise_scale_inv(static_cast<void *>(rhs_sinv_ptr), DType::kFloat8E8M0,
-                                  rhs_sinv_shape);
+      auto lhs_sinv_shape_ = lhs_sinv_i.dimensions();
+      auto rhs_sinv_shape_ = rhs_sinv_i.dimensions();
+      for (int i = 0; i < 2; i++) {
+        lhs_sinv_shape[i] = lhs_sinv_shape_[i];
+        rhs_sinv_shape[i] = rhs_sinv_shape_[i];
+      }
```
phu0ngng, line 128: Hi, 
I think the `scale_inv`s are 1D since we already squeezed them on the JAX side. Why do we loop through 2 iterations here?

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -163,50 +185,15 @@ Error_Type GroupedGemmImpl(uint8_t *lhs_ptr, const DType &lhs_dtype, uint8_t *lh
...
-                                  .Arg<Buffer_Type>()      // dim_list
-                                  .Ret<Buffer_Type>()      // out_flatten
-                                  .Ret<Buffer_Type>()      // workspace_flatten
+                                  .RemainingArgs()         // input list
+                                  .RemainingRets()         // output list
                                   .Attr<int64_t>("num_gemms")
-                                  .Attr<JAXX_Scaling_Mode>("scaling_mode"),
+                                  .Attr<JAXX_Scaling_Mode>("scaling_mode")
+                                  .Attr<int64_t>("has_bias")
+                                  .Attr<int64_t>("workspace_size"),
```
phu0ngng, line 196: Why do we need to bind `workspace_size` when we can infer it from the Buffer?

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -79,63 +67,94 @@ Error_Type GroupedGemmImpl(uint8_t *lhs_ptr, const DType &lhs_dtype, uint8_t *lh
...
-                                  rhs_sinv_shape);
+      auto lhs_sinv_shape_ = lhs_sinv_i.dimensions();
+      auto rhs_sinv_shape_ = rhs_sinv_i.dimensions();
+      for (int i = 0; i < 2; i++) {
+        lhs_sinv_shape[i] = lhs_sinv_shape_[i];
+        rhs_sinv_shape[i] = rhs_sinv_shape_[i];
+      }
+
+      TensorWrapper lhs_i_(NVTE_MXFP8_1D_SCALING);
+      TensorWrapper rhs_i_(NVTE_MXFP8_1D_SCALING);
```
phu0ngng, line 131: I would prefer we use `JAXX_Scaling_Mode` in JAX csrc so that we can abstract away from NVTEScalingMode.

transformer_engine/jax/csrc/extensions/gemm.cpp
huanghua1994: Fixed

transformer_engine/jax/csrc/extensions/gemm.cpp
huanghua1994: Fixed

transformer_engine/jax/csrc/extensions/gemm.cpp
huanghua1994: scale_inv is reshaped to a 2D matrix in JAX

Issue #1654, opened 2025-04-08
[PyTorch] Unable to run FP8 example on 5090


TidalPaladin:
**Describe the bug**

Running the PyTorch example code from the Transformer Engine documentation fails on 5090

```
Traceback (most recent call last):
  File "/workspace/fp8.py", line 22, in <module>
    loss.backward()
  File "/usr/local/lib/python3.12/dist-packages/torch/_tensor.py", line 648, in backward
    torch.autograd.backward(
  File "/usr/local/lib/python3.12/dist-packages/torch/autograd/__init__.py", line 353, in backward
    _engine_run_backward(
  File "/usr/local/lib/python3.12/dist-packages/torch/autograd/graph.py", line 824, in _engine_run_backward
    return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/autograd/function.py", line 307, in apply
    return user_fn(self, *args)
           ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/transformer_engine/pytorch/module/linear.py", line 552, in backward
    wgrad, grad_bias_, _, rs_out = general_gemm(
                                   ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/transformer_engine/pytorch/cpp_extensions/gemm.py", line 141, in general_gemm
    out, bias_grad, gelu_input, extra_output = tex.generic_gemm(*args, **kwargs)
                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: /workspace/transformerengine/transformer_engine/common/gemm/cublaslt_gemm.cu:395 in function cublas_gemm: Assertion failed: status != CUBLAS_STATUS_NOT_SUPPORTED. Unable to find suitable cuBLAS GEMM algorithm
```

**Steps/Code to reproduce bug**

Run the following example code in the PyTorch container

```python
import torch
import transformer_engine.pytorch as te
from transformer_engine.common import recipe

# Set dimensions.
in_features = 768
out_features = 3072
hidden_size = 2048

# Initialize model and inputs.
[...]

Issue #1585, opened 2025-03-18
cuBLAS Error


wangli68:
When my pytorch is 2.5 and my transformer engine is 2.1, I run the following original network structure:
```python
class SelfAttention(nn.Module):
    def __init__(self, dim: int, num_heads: int, eps: float = 1e-6):
        super().__init__()
        self.dim = dim
        self.num_heads = num_heads
        self.head_dim = dim // num_heads

        self.q = nn.Linear(dim, dim)
        self.k = nn.Linear(dim, dim)
        self.v = nn.Linear(dim, dim)
        self.o = nn.Linear(dim, dim)
        self.norm_q = RMSNorm(dim, eps=eps)
        self.norm_k = RMSNorm(dim, eps=eps)

    def forward(self, x, freqs):
        q = self.norm_q(self.q(x))
        k = self.norm_k(self.k(x))
        v = self.v(x)
        x = flash_attention(
            q=rope_apply(q, freqs, self.num_heads),
            k=rope_apply(k, freqs, self.num_heads),
            v=v,
            num_heads=self.num_heads
        )
        return self.o(x)
``` 
Note that at this point, I have changed the lightning network structure to the corresponding te.xx, but it reports the following error:

rank1]:   File "/output/DiffSynth-Studio/diffsynth/models/wan_video_dit.py", line 128, in forward
[rank1]:     q = self.norm_q(self.q(x))
[rank1]:   File "/openbayes/home/Test/env/ao25/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1736, in _wrapped_call_impl
[rank1]:     return self._call_impl(*args, **kwargs)
[rank1]:   File "/openbayes/home/Test/env/ao25/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1747, in _call_impl
[rank1]:     return forward_call(*args, **kwargs)
[rank1]:   File "/openbayes/home/Test/env/ao25/lib/python3.10/site-packages/torch/_dynamo/eval_frame.py", line 632, in _fn
[rank1]:     return fn(*args, **kwargs)
[rank1]:   File "/openbayes/home/Test/env/ao25/lib/python3.10/site-packages/transformer_engine/pytorch/module/linear.py", line 1085, in forward
[rank1]:     out = linear_fn(*args)
[rank1]:   File "/openbayes/home/Test/env/ao25/lib/python3.10/site-packages/torch/autograd/function.py", line 575, in apply
[...]

PR #1659, opened 2025-04-09
rtx5090 arch fix support

sudhakarsingh27:
# Description

Extend arch support logic for RTX5090 and related archs.

Fixes # (issue)
https://github.com/NVIDIA/TransformerEngine/issues/1654

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Extend arch support logic for RTX5090 and related archs.

# Checklist:


transformer_engine/pytorch/csrc/extensions/quantizer.cpp
```diff
@@ -7,9 +7,9 @@
 #include <pybind.h>
 
 #include "common.h"
+#include "common/common.h"
```
sudhakarsingh27, line 10: @ptrendx what's the right way to use the function defined in `common` in framework specific dirs like this one?

transformer_engine/pytorch/csrc/extensions/quantizer.cpp
```diff
@@ -103,7 +103,7 @@ std::pair<TensorWrapper, py::object> Float8Quantizer::create_tensor(
   }
   const py::object py_data = rowwise_usage ? py::cast(data) : py::none();
   at::Tensor columnwise_data;
-  bool create_transpose = columnwise_usage && !non_tn_fp8_gemm_supported();
+  bool create_transpose = columnwise_usage && !is_supported_nontn_fp8_gemm();
```
sudhakarsingh27, line 106: @ptrendx I realized that we could also remove the usage of this function for example with the `is_supported_nontn_fp8_gemm` function from common.

transformer_engine/common/common.cu
```diff
@@ -135,3 +135,19 @@ bool is_supported_by_CC_100() {
 }
 
 }  // namespace transformer_engine
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+bool nvte_is_supported_nontn_fp8_gemm() {
```
timmoon10, line 142: This name feels awkward to me. How about:
```suggestion
bool nvte_is_non_tn_fp8_gemm_supported() {
```
This matches the PyTorch function: https://github.com/NVIDIA/TransformerEngine/blob/04642bf8910da5b800b10d98e32ad1a310e106cd/transformer_engine/pytorch/utils.py#L253
Alternatively, we could name it `nvte_is_only_tn_fp8_gemm_supported`, so we don't not avoid multiple negatives like `!nvte_is_non_tn_fp8_gemm_supported()`.

timmoon10 (commented):
Overall LGTM, although I feel the function name could be tweaked.

transformer_engine/common/common.cu
sudhakarsingh27: I changed the name to suggested. Also, I feel like a better version of this function should be `get_supported_gemm_types` and then in the code we could do `"<gemm_type>" in get_supported_gemm_types()`. Since this is a very specific use-case for now, maybe we could let this as is?

transformer_engine/common/common.h
```diff
@@ -599,4 +599,14 @@ bool is_supported_by_CC_100();
...
 
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+bool nvte_is_non_tn_fp8_gemm_supported();
+
+#ifdef __cplusplus
+}  // extern "C"
+#endif
```
timmoon10, line 610: Exposing this as a C function is strange to me. We do that for the publicly exposed functions so that we support C users, but this is a utility function that we only expect to use within C++.

I think the right approach is to put this within the `transformer_engine` namespace. I am thinking about adding some cuBLAS utility functions to help with https://github.com/NVIDIA/TransformerEngine/issues/1585, so I may move this there in the future.

sudhakarsingh27: If we add it to the `transformer_engine` namespace, we'll end up exposing this function in `lib_transformerengine.version` as 
```
              transformer_engine::nvte_cudnn_handle_init*;
```

and I believe we don't want to do that

sudhakarsingh27: Although I do agree that putting a single C function there does seem odd - as you suggested maybe a utility functions file would be better but even then it'd be a separate file with a single utility function

timmoon10: We're already exposing the function, so we might as well make it a C++ function. `transformer_engine::is_non_tn_fp8_gemm_supported` is a better name than `nvte_is_non_tn_fp8_gemm_supported`.

timmoon10: Hm, I see that @ptrendx prefers if we move away from C++ utility functions, and do everything via publicly exposed C APIs. Then the right approach is to define `nvte_is_non_tn_fp8_gemm_supported` within `transformer_engine/common/include/transformer_engine.h`, not in `common.h`. This also means that we can revert `libtransformer_engine.version`, since that exposes all the `nvte_` functions.

sudhakarsingh27: Do you mean, if we define `nvte_is_non_tn_fp8_gemm_supported` in `transformer_engine/common/include/transformer_engine.h`, then we don't need to explicitly state that it's a C function in `lib*.version`?

sudhakarsingh27: Also the function is defined in `common.cu` and declared in `common.h`. Is it fine to define it in `transformer_engine.h`?

timmoon10: `transformer_engine.h` is reasonable, although it would be even better to have a `utils.h` header within `transformer_engine/common/include/transformer_engine`. We can move things around later though.

timmoon10 (approved):

transformer_engine/pytorch/utils.py
```diff
@@ -251,11 +251,12 @@ def is_bf16_compatible() -> None:
     return torch.cuda.get_device_capability()[0] >= 8
 
 
-def non_tn_fp8_gemm_supported() -> bool:
+def is_non_tn_fp8_gemm_supported() -> bool:
```
ptrendx, line 254: We could also call into that nvte function here, but I'm fine with leaving this in this PR.

ptrendx (approved):
LGTM.

ptrendx:
I retried the jax unittest - seems like a temporary issue with the connection to HuggingFace.

sudhakarsingh27:
Looks like @yaox12 has added a function in `pytorch/csrc/util.h` but I had deleted that file. Should I add that file back? @ptrendx

ptrendx (approved):

PR #1662, opened 2025-04-09
add view/reshape to blockwise tensor

Autumn1998:
# Description

Based on the MXFP8 implementation, this PR implemented view and reshape methods for the 1x128 scaling tensor in the sub-channel recipe, and added corresponding unit tests

Merged In https://github.com/NVIDIA/TransformerEngine/pull/1641

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/test_float8blockwisetensor.py
+++ b/tests/pytorch/test_float8blockwisetensor.py
@@ -373,17 +373,22 @@ class TestFloat8BlockwiseTensor:
 
     @pytest.mark.parametrize("fp8_dtype", [tex.DType.kFloat8E4M3, tex.DType.kFloat8E5M2], ids=str)
     @pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float32], ids=str)
-    @pytest.mark.parametrize("dims", [[16, 16, 512], [16, 16, 512, 16], [12,7,11],[13,14,16],[2,3,5]])
-    def test_view_and_reshape_1D(self, fp8_dtype: tex.DType, dtype: torch.dtype, dims: List[int]) -> None:
+    @pytest.mark.parametrize(
+        "dims", [[16, 16, 512], [16, 16, 512, 16], [12, 7, 11], [13, 14, 16], [2, 3, 5]]
+    )
+    def test_view_and_reshape_1D(
+        self, fp8_dtype: tex.DType, dtype: torch.dtype, dims: List[int]
+    ) -> None:
         """Test view operations that preserve tensor shape"""
         device = "cuda"
+
         def is_bitwise_equal(a, b):
             if a.numel() != b.numel():
                 return False
             a_flat = a.reshape(-1).view(torch.uint8)
             b_flat = b.reshape(-1).view(torch.uint8)
             return torch.all((a_flat ^ b_flat) == 0)
-        
+
         x_hp = torch.rand(dims, dtype=dtype, device=device)
         quantizer = Float8BlockQuantizer(
             fp8_dtype=fp8_dtype,
@@ -419,20 +424,22 @@ class TestFloat8BlockwiseTensor:
         assert is_bitwise_equal(x_fp8_reshape._rowwise_data, x_fp8._rowwise_data)
         assert is_bitwise_equal(x_fp8_reshape._rowwise_scale_inv, x_fp8._rowwise_scale_inv)
 
-
     @pytest.mark.parametrize("fp8_dtype", [tex.DType.kFloat8E4M3, tex.DType.kFloat8E5M2], ids=str)
     @pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float32], ids=str)
-    @pytest.mark.parametrize("dims", [[16, 16, 512, 16], [2,512,512,128], [3,13,14,16]])
-    def test_view_and_reshape_2D(self, fp8_dtype: tex.DType, dtype: torch.dtype, dims: List[int]) -> None:
+    @pytest.mark.parametrize("dims", [[16, 16, 512, 16], [2, 512, 512, 128], [3, 13, 14, 16]])
+    def test_view_and_reshape_2D(
+        self, fp8_dtype: tex.DType, dtype: torch.dtype, dims: List[int]
+    ) -> None:
         """Test view operations that preserve tensor shape"""
         device = "cuda"
+
         def is_bitwise_equal(a, b):
             if a.numel() != b.numel():
                 return False
             a_flat = a.reshape(-1).view(torch.uint8)
             b_flat = b.reshape(-1).view(torch.uint8)
             return torch.all((a_flat ^ b_flat) == 0)
-        
+
         x_hp = torch.rand(dims, dtype=dtype, device=device)
         quantizer = Float8BlockQuantizer(
             fp8_dtype=fp8_dtype,
@@ -467,8 +474,6 @@ class TestFloat8BlockwiseTensor:
         # Check the bitwise equality of the inner data
         assert is_bitwise_equal(x_fp8_reshape._rowwise_data, x_fp8._rowwise_data)
         assert is_bitwise_equal(x_fp8_reshape._rowwise_scale_inv, x_fp8._rowwise_scale_inv)
-
-
 
     @pytest.mark.parametrize("fp8_dtype", [tex.DType.kFloat8E4M3], ids=str)
     @pytest.mark.parametrize("dtype", [torch.bfloat16], ids=str)
--- a/tests/pytorch/test_float8blockwisetensor.py
+++ b/tests/pytorch/test_float8blockwisetensor.py
@@ -371,6 +371,105 @@ class TestFloat8BlockwiseTensor:
         with pytest.raises(AssertionError):
             torch.testing.assert_close(x_view.dequantize(), -x_hp, **_tols[fp8_dtype])
 
+    @pytest.mark.parametrize("fp8_dtype", [tex.DType.kFloat8E4M3, tex.DType.kFloat8E5M2], ids=str)
+    @pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float32], ids=str)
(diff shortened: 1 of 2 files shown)


transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -543,19 +543,79 @@ def forward(
...
+        elif len(shape) == 1 and isinstance(shape[0], Iterable):
+            shape = shape[0]
+        if -1 in shape:
+            shape = list(shape)
+            d_inferred = -math.prod(ctx.shape) // math.prod(shape)
+            for i, d in enumerate(shape):
+                if d == -1:
+                    shape[i] = d_inferred
+                    break
+        if shape[-1] != ctx.shape[-1]:
```
timmoon10, line 558: We assume here that `ctx.shape` is available, but earlier we have logic to handle `ctx=None`:
https://github.com/NVIDIA/TransformerEngine/blob/402b270bf8f6b7a6f445ab67e43055aca2f31f19/transformer_engine/pytorch/tensor/float8_blockwise_tensor.py#L541

I think we should just assume that `ctx` is a valid context object and stop supporting the `ctx=None` case. It's nonstandard, we never take advantage of that case (perhaps the original use-case was removed?), and `MXFP8Tensor` and `Float8Tensor` don't support it. We should also make a similar change to `_ReshapeFunc`.

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -543,19 +543,79 @@ def forward(
...
     ) -> Tuple[Optional[torch.Tensor], ...]:
         # pylint: disable=missing-function-docstring
-
         if isinstance(grad, Float8BlockwiseQTensor):
-            raise NotImplementedError("View bwd not implemented")
+            new_data = (
+                grad._rowwise_data.view(*ctx.shape) if grad._rowwise_data is not None else None
+            )
+            if grad._columnwise_data is not None:
+                new_columnwise_data = grad._columnwise_data.view(ctx.shape[-1], -1)
```
timmoon10, line 603: Why are we flattening the column-wise data to 2D?
```suggestion
                columnwise_shape = [ctx.shape[-1]] + list(ctx.shape[:-1])
                new_columnwise_data = grad._columnwise_data.view(columnwise_shape)
```

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -543,19 +543,79 @@ def forward(
...
+                "Float8BlockwiseQTensor does not support reshaping inner dimension "
+                f"(attempted to reshape dims={tuple(tensor.shape)} to {tuple(shape)})"
+            )
+
+        if list(shape) == list(tensor.shape):
+            return tensor
+
+        if tensor._is_2D_scaled:
+            raise NotImplementedError("View of 2D scaled tensor is not implemented")
+
```
timmoon10, line 569: As long as we only reshape the outer dims, shouldn't viewing/reshaping work for both 1D and 2D scaled tensors?
```suggestion
```
Similar question for `_ReshapeFunc`.

tests/pytorch/test_float8blockwisetensor.py
```diff
@@ -371,6 +371,37 @@ def test_view_same_shape(
...
+        )
+        x_fp8 = quantizer.make_empty(x_hp.shape, dtype=dtype, device=device)
+        quantizer.update_quantized(x_hp.clone(), x_fp8)
+
+        # Test view
+        x_hp_view = x_hp.view(-1, dims[-1]).contiguous()
+        x_fp8_view = x_fp8.view(-1, dims[-1])
+        torch.testing.assert_close(
+            x_fp8_view.dequantize().contiguous(), x_hp_view, **_tols[fp8_dtype]
+        )
```
timmoon10, line 396: Instead of comparing `x_fp8_view` and `x_hp_view` with FP8 tolerances, it makes more sense to me if we compare `x_fp8_view` and `x_fp8` with exact tolerances. Also, we should check that the pointers are identical.

tests/pytorch/test_float8blockwisetensor.py
```diff
@@ -371,6 +371,37 @@ def test_view_same_shape(
...
+
+        # Test reshape
+        x_hp_reshape = x_hp.reshape(-1, dims[-1]).contiguous()
+        x_fp8_reshape = x_fp8.reshape(-1, dims[-1])
+        torch.testing.assert_close(
+            x_fp8_reshape.dequantize().contiguous(), x_hp_reshape, **_tols[fp8_dtype]
+        )
```
timmoon10, line 403: It would be better to compare `x_fp8_reshape` and `x_fp8` with exact tolerances.

timmoon10 (commented):
This would be a nice change. If I remember correctly, the lack of a view function was causing problems when we passed a block-scaled FP8 tensor into an autograd function?

It would be worth double checking that this is doing the right thing. The view impl in `__torch_dispatch__` is still incorrect since it only handles row-wise data (`MXFP8Tensor` is also incorrect). That said, we can merge this quickly if this helps fix the immediate bug.

tests/pytorch/test_float8blockwisetensor.py
Autumn1998: Yes, I just added the bitwise-check for the `x_fp8_view` and `x_fp8`, and the assert for identical pointers

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
Autumn1998: have added the support for the 2D-scaled case.

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
Autumn1998: I initially copied the code from the `MXFP8` tensor implementation. After reviewing the logic here, I believe the following approach is better. Could there be a bug in the `MXFP8` tensor?

```
if grad._is_2D_scaled:
    columnwise_shape = list(ctx.shape[-2:]) + list(ctx.shape[:-2])
else:
    columnwise_shape = [ctx.shape[-1]] + list(ctx.shape[:-1])
```

Additionally, I think the reshape/view operation for columnwise_data might be incorrect. Unlike the rowwise data, we cannot keep the inner dimension unchanged. We can only choose one  whose inner dim stay unchanged between rowwise data and columnwise tensor.  Perhaps it would be helpful to add an assertion or warning for tensors with columnwise tensor?

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -538,24 +538,100 @@ def forward(
...
+        # Construct new tensor if shape is provided
+        new_rowwise_data = None
+        new_columnwise_data = None
+        if tensor._rowwise_data is not None:
+            new_rowwise_data = tensor._rowwise_data.view(*shape)
+        if tensor._columnwise_data is not None:
+            # Maybe changed in the future
+            if tensor._is_2D_scaled:
+                columnwise_shape = list(shape[-2:]) + list(shape[:-2])
```
kunlunl, line 586: Here should be `columnwise_shape = [shape[-1], shape[-2]] + list(shape[:-2])`, not `[shape[-2], shape[-1] + list(shape[:-2])`?

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -538,24 +538,100 @@ def forward(
...
         # pylint: disable=missing-function-docstring
-
         if isinstance(grad, Float8BlockwiseQTensor):
-            raise NotImplementedError("View bwd not implemented")
+            new_data = (
+                grad._rowwise_data.view(*ctx.shape) if grad._rowwise_data is not None else None
+            )
+            if grad._columnwise_data is not None:
+                if grad._is_2D_scaled:
+                    columnwise_shape = list(ctx.shape[-2:]) + list(ctx.shape[:-2])
```
kunlunl, line 616: Save as above, should be `[ctx.shape[-1], ctx.shape[-2] + ...`, not `[ctx.shape[-2], ctx.shape[-1] + ...`?

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -592,9 +667,51 @@ def forward(
...
+
+        # Construct new tensor if shape is provided
+        new_rowwise_data = None
+        new_columnwise_data = None
+        if tensor._rowwise_data is not None:
+            new_rowwise_data = tensor._rowwise_data.reshape(*shape)
+        if tensor._columnwise_data is not None:
+            # Maybe changed in the future
+            if tensor._is_2D_scaled:
+                columnwise_shape = list(shape[-2:]) + list(shape[:-2])
```
kunlunl, line 698: Same as above, should be [ctx.shape[-1], ctx.shape[-2] + ..., not [ctx.shape[-2], ctx.shape[-1] + ...?

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -604,5 +721,27 @@ def backward(
...
 
         if isinstance(grad, Float8BlockwiseQTensor):
-            raise NotImplementedError("Reshape bwd not implemented yet.")
+            new_rowwise_data = None
+            new_columnwise_data = None
+            if grad._rowwise_data is not None:
+                new_rowwise_data = grad._rowwise_data.view(*ctx.shape)
+            if grad._columnwise_data is not None:
+                if grad._is_2D_scaled:
+                    columnwise_shape = list(ctx.shape[-2:]) + list(ctx.shape[:-2])
```
kunlunl, line 730: Same as above, should be [ctx.shape[-1], ctx.shape[-2] + ..., not [ctx.shape[-2], ctx.shape[-1] + ...?

Autumn1998:
The content of the PR has been merged in https://github.com/NVIDIA/TransformerEngine/pull/1641

PR #1668, opened 2025-04-10
[PyTorch] More precise test for the CPU offloading.

pggPL:
# Description

Tests for cpu offloading do not include Transformer Layer and are not precise - the check only if memory consumption with the offloading was smaller than without it.

This PR makes this tests precise - they check if the memory usage reduction is **exactly** as big as it should be.

I identified some room for improvement of the offloading for the attention/transformer layer:

1. Flash Attention seems to save some tensor for backward internally, so we cannot offload it.
2. TransformerLayer uses dropout 2 times and we cannot access dropout mask to mark it with `.activation_offload=True`.

There is solution to problem 2: add new bias dropout fusion function with custom dropout and activate it when CPU offload is enabled. This change is not included in this PR, I want to know what do you think about that.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/test_cpu_offloading.py
+++ b/tests/pytorch/test_cpu_offloading.py
@@ -10,7 +10,7 @@ import torch
 import transformer_engine.pytorch as te
 from transformer_engine.pytorch.fp8 import FP8GlobalStateManager
 
-# Check if FP8 supported
+# Check if FP8 is supported
 fp8_available, reason_for_no_fp8 = FP8GlobalStateManager.is_fp8_available()
 
 SIZE = 512
@@ -18,14 +18,14 @@ NUM_HEADS = 8
 NUM_LAYERS = 5
 EPSILON = 0.1
 
-# Flash attention saves some internal tensor for the backward,
+# Flash attention saves some internal tensor for the backward pass
 # that cannot be offloaded to CPU.
 assert os.getenv("NVTE_FLASH_ATTN") == "0"
 
 # Offloading is supported for attention only for fused and flash attention backends,
-# so use of bfloat16 is required.
+# so the use of bfloat16 is required.
 #
-# For transformer layer activation offloading with dropout is not supported,
+# For the TransformerLayer, activation offloading with dropout is not supported,
 # so we set hidden_dropout to 0.0.
 model_types = {
     "linear":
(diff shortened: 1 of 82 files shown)


tests/pytorch/test_cpu_offloading.py
```diff
@@ -2,84 +2,143 @@
...
 
     assert with_offloading < without_offloading
+
+    # The only difference between the memory consumption of with_offloading
+    # and without_offloading_one_layer should be the size of the FP8 weights cache,
+    # which is not offloaded to the CPU.
+    memory_consumption_diff = abs(with_offloading - without_offloading_one_layer)
+    assert (
+        memory_consumption_diff < _get_fp8_weight_cache_size(models_list[1:], fp8_recipe) + EPSILON
+    )
```
timmoon10, line 144: Being pedantic, we're not actually checking that the memory savings are exactly as expected. We're just checking that the memory usage does not increase as we add more layers (except for the weight cache). I could imagine cases where CPU offloading adds a constant memory overhead. That said, I'd expect this check to catch most typical bugs.

tests/pytorch/test_cpu_offloading.py
```diff
@@ -2,84 +2,143 @@
...
+    None,  # non-fp8
+    # recipe.MXFP8BlockScaling(), - scale inverse tensors offloading doest not work yet
+    recipe.Float8CurrentScaling(),
+    recipe.DelayedScaling(),
+]
 
 SIZE = 512
+NUM_HEADS = 8
+NUM_LAYERS = 5
+EPSILON = 0.1
```
timmoon10, line 28: For an FP8 linear layer, the cached input is 32 MiB and the weight cache is 0.5 MiB. A 0.1 MiB tolerance seems sensible, although we may want to revisit if we ever reduce the data dimensions. A relative tolerance would be more generalizable, but I don't think it's necessary.

timmoon10 (approved):
LGTM

tests/pytorch/test_cpu_offloading.py
pggPL: There is the check above this which checks that.

PR #1671, opened 2025-04-11
Added attention activation offloading support for TE v2.0

sanandaraj5597:
# Description

Added support for offloading with LoRa, had to add activation offloading support for FP8 attention.


pggPL (approved):
LGTM

PR #1672, opened 2025-04-11
fix(grouped_gemm): fix error when at::from_blob pass zero shape

cos120:
# Description

In grouped gemm, when shape of D is zero, will raise `The specified pointer resides on host memory and is not registered with any CUDA device.`

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Add zero shape check.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/csrc/extensions/gemm.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/gemm.cpp
@@ -337,7 +337,9 @@ std::optional<std::vector<at::Tensor>> te_general_grouped_gemm(
         pytorch::detail::getGemmOutputShape(te_A.shape(), transa, te_B.shape(), transb);
     bool D_numel_is_zero = false;
     std::vector<int64_t> D_shape;
+    bool D_shape_zero = false;
     for (size_t t : size_t_shape) {
+      D_shape_zero = t == 0 || D_shape_zero;
       D_shape.push_back(t);
       if (t == 0) {
         D_numel_is_zero = true;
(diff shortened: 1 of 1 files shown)


pggPL:
Thank you for the contribution! Can you sign your commit please - instruction in the [link](https://github.com/NVIDIA/TransformerEngine/blob/2856c3e09e935baf2d3fbfc583739873a2dfbfeb/CONTRIBUTING.rst#L69)

There is logic with `D_numel_is_zero` which is intended to prevent this error. We have some tests for this case which seem to pass. Can you provide some code which fails or tell which line throws such error? I see that now `out_tensor` is not initialized in case of zer-size output, which can result in some errors, but I do not see why how the error you mention can happen.

It would be nice to remove `D_numel_is_zero` if indeed current logic is flawed.

cos120:
> Thank you for the contribution! Can you sign your commit please - instruction in the [link](https://github.com/NVIDIA/TransformerEngine/blob/2856c3e09e935baf2d3fbfc583739873a2dfbfeb/CONTRIBUTING.rst#L69)
> 
> There is logic with `D_numel_is_zero` which is intended to prevent this error. We have some tests for this case which seem to pass. Can you provide some code which fails or tell which line throws such error? I see that now `out_tensor` is not initialized in case of zer-size output, which can result in some errors, but I do not see why how the error you mention can happen.
> 
> It would be nice to remove `D_numel_is_zero` if indeed current logic is flawed.

Oh, I am using TE2.1, and I cherry-pick my fix from 2.1 to main, I do not notice that this has been fixed on main branch. I will close this pr.

PR #1673, opened 2025-04-11
[JAX] Improving the test_multiprocessing_encoder.py run script

phu0ngng:
# Description

Improving the test_multiprocessing_encoder.py run script

## Type of change

- [] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:

### DIFF
--- a/examples/jax/encoder/run_test_multiprocessing_encoder.sh
+++ b/examples/jax/encoder/run_test_multiprocessing_encoder.sh
@@ -13,8 +13,6 @@ TEST_CASES=(
 "test_te_delayed_scaling_fp8_shardy"
 )
 
-HAS_FAILURE=0  # Global failure flag
-
 echo
 echo "*** Executing tests in examples/jax/encoder/test_multiprocessing_encoder.py ***"
 
@@ -39,7 +37,6 @@ for TEST_CASE in "${TEST_CASES[@]}"; do
 
   # Check and print the log content accordingly
   if grep -q "FAILED" "${TEST_CASE}_gpu_0.log"; then
-    HAS_FAILURE=1
     echo "... $TEST_CASE FAILED"
     tail -n +7 "${TEST_CASE}_gpu_0.log"
   elif grep -q "SKIPPED" "${TEST_CASE}_gpu_0.log"; then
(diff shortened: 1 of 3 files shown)


examples/jax/encoder/run_test_multiprocessing_encoder.sh
```diff
@@ -4,20 +4,50 @@
...
+    echo "... $TEST_CASE FAILED"
+    tail -n +7 "${TEST_CASE}_gpu_0.log"
+  elif grep -q "SKIPPED" "${TEST_CASE}_gpu_0.log"; then
+    echo "... $TEST_CASE SKIPPED"
+  else
+    echo "... $TEST_CASE PASSED"
+  fi
+
+  # Remove the log file after processing it
+  rm ${TEST_CASE}_gpu_*.log
```
jberchtold-nvidia, line 50: Can we move this removing of the logs to only happen for SKIPPED or PASSED tests? If there's a failure it might be nice to automatically keep logs so we don't need to run this script again with this line commented out. But I agree removing logs in cases where tests aren't failing is a nice improvement. Wdyt?

jberchtold-nvidia (approved):

examples/jax/encoder/run_test_multiprocessing_encoder.sh
phu0ngng: If the test fails, we print out the log via `tail` at L42. So I don't think we need to keep the log.

examples/jax/encoder/test_multiprocessing_encoder.py
```diff
@@ -604,7 +604,7 @@ def test_te_bf16(self):
     def test_te_delayed_scaling_fp8(self):
         """Test Transformer Engine with DelayedScaling FP8"""
         result = self.exec(True, "DelayedScaling")
-        assert result[0] < 0.505 and result[1] > 0.755
+        assert result[0] < 0.505 and result[1] > 0.754
```
jberchtold-nvidia, line 607: I think this is okay because our multi-GPU but single process test in `test_multigpu_encoder.py` has relaxed these numbers further than this small 0.001 diff we're seeing
```
    @unittest.skipIf(not is_fp8_supported, fp8_reason)
    def test_te_delayed_scaling_fp8(self):
        """Test Transformer Engine with DelayedScaling FP8"""
        self.args.use_fp8 = True
        self.args.fp8_recipe = "DelayedScaling"
        assert actual[0] < 0.535 and actual[1] > 0.73
```

phu0ngng:
pipelines #26957438

PR #1675, opened 2025-04-11
[JAX] Add collective GEMM without compute/communication overlap

philipphack:
# Description

Rebase of #1307:

Implements XLA custom calls in C++, and the corresponding JAX primitive including custom partitioning rules.

Custom partitioning rules for a LHS:([B,] M, K) x RHS:([B,] K, N) = OUT:([B,] M, N) batched mat-mul operation where [B] is the batch dimension:

Preserve the partitioning of the [B] dimension for all operands.
Always all-gather LHS along the M dimension.
Error out if RHS is partitioned in both K and N dimensions.
Force the K dimension of LHS to match the partitioning of the K dimension of RHS.
If K dimension is partitioned but M dimension is not, jax.lax.psum (all-reduce) the output over the TP mesh resource.
If both the M and K dimensions are partitioned, jax.lax.psum_scatter (reduce-scatter) the output over the TP mesh resource.
In practice, the RHS matrix (typically the weight tensor) should be allocated with transposed contracting dimensions ([B,] N, K) for optimal GEMM heuristics in cuBlasLt. This layout is also mandatory for FP8 inputs.

This PR does NOT update fused ops or Flax/Praxis modules to use the new GEMM custom op over the existing XLA pattern matching approach.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

 Added JAX primitive for the XLA custom call.
 Added serial unit test.
 Added distributed unit test.


transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -514,3 +544,678 @@ def grouped_gemm(
...
+        batched_output,
+        contracting_dims,
+        fuse_gelu,
+        fuse_bias,
+        grad,
+        accumulate,
+        use_split_accumulator,
+    ):
+        """
+        Fused attention fwd lowering rules
```
denera, line 752: Looks like leftover incorrect docstring from the copied primitive template.

denera (approved):
LGTM, pending very minor docstring fix.

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -34,6 +34,14 @@
...
 __all__ = ["quantize", "quantize_dbias"]
 
+def _jax_cast_fp8(inputs, scale, amax, out_dtype):
+    """
+    JAX native fp8 casting implementation
+    """
+    casted_output = _jax_quantize(inputs, scale, dq_dtype=out_dtype)
+    updated_amax = jax.lax.max(amax, jnp.max(jnp.abs(inputs)).astype(amax.dtype))
+    return casted_output, updated_amax
+
```
phu0ngng, line 44:
```suggestion
```
Please use `_jax_quantize()` instead.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -514,3 +544,678 @@ def grouped_gemm(
...
+
+    if scaling_mode == ScalingMode.NVTE_NO_SCALING:
+        lhs_scale_inv = jnp.ones(1, dtype=jnp.float32)
+        rhs_scale_inv = jnp.ones(1, dtype=jnp.float32)
+    if scaling_mode == ScalingMode.NVTE_DELAYED_TENSOR_SCALING:
+        lhs_scale_inv = lhs.scale_inv.reshape(-1)
+        rhs_scale_inv = rhs.scale_inv.reshape(-1)
+    if scaling_mode == ScalingMode.NVTE_MXFP8_1D_SCALING:
+        lhs_scale_inv = lhs_scale_inv.reshape(-1)
+        rhs_scale_inv = rhs_scale_inv.reshape(-1)
```
phu0ngng, line 1201: Hi, 
1. Why do we need to reshape the scale_inv for DelayedScaling?
2. For MXFP8, don't we need to call `swizzle_scale`?

denera:
Closing in favor of #1846

PR #1681, opened 2025-04-14
Fix #1524 and other softmax mask functionality

### LINKED ISSUE #1524: Causal mask ignored in DotProductAttention
for transformer_engine version 1.14.0+87fbe812f
in transformer_engine.jax.flax.module
in Softmax class

in __call__ line 191-198:

```
# For the case that self.softmax == SoftmaxType.SCALED_UPPER_TRIANG_MASKED
# and kernel is unavailable, then try on pure scaled softmax custom calls.
if is_softmax_kernel_available(
    SoftmaxType.SCALED, batch, heads, q_seqlen, k_seqlen, dtype
):
    outputs = softmax(logits, None, self.scale_factor, SoftmaxType.SCALED)
else:
    outputs = jax_nn.softmax(logits * self.scale_factor)  # <- self.softmax_type ignored
```

after else, self.softmax_type is ignored and no causal attention is performed when e.g. self.softmax_type = SoftmaxType.SCALED_UPPER_TRIANG_MASKED

KshitijLakhani:
# Description

Fixes #1524 
Simplifies and corrects logic around mapping attn mask to softmax mask, while adding additional testing capabilities

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

1. Add test cases for better coverage in `test_layer.py`

- causal and window size None
- causal and window size default (-1,1)
- no_mask and window size default (-1,1)
- no_mask and window size default (2,2)
- padding and window size None
- padding_causal and window_size (2,2)

2. Correctly map `padding_causal` to `scaled_masked` instead of `scaled_upper_triangle`
3. Add fallback jax implementation for `scaled_upper_triangle` (#1524 )
4. Add tests for `Softmax` wrapper/module in `jax.flax.module.Softmax`
5. Simplify logic for jax based softmax
6. Add support table in comments for attn mask, SWA and softmax type

## TODO:
Add support for `causal_bottom_right` and `padding_causal_bottom_right` as well. 

# Checklist:

### DIFF
--- a/tests/jax/test_layer.py
+++ b/tests/jax/test_layer.py
@@ -218,7 +218,48 @@ ATTRS = [
     {
         _KEY_OF_TRANSPOSE_BS: False,
         _KEY_OF_RELATIVE_EMBEDDING: False,
+        _KEY_OF_SELF_ATTN_MASK_TYPE: "causal",
+        _KEY_OF_WINDOW_SIZE: None,
+        _KEY_OF_FLOAT32_ATTENTION_LOGITS: True,
+    },
+    # attrs23
+    {
+        _KEY_OF_TRANSPOSE_BS: False,
+        _KEY_OF_RELATIVE_EMBEDDING: False,
+        _KEY_OF_SELF_ATTN_MASK_TYPE: "causal",
+        _KEY_OF_FLOAT32_ATTENTION_LOGITS: True,
+    },
+    # attrs24
+    {
+        _KEY_OF_TRANSPOSE_BS: False,
+        _KEY_OF_RELATIVE_EMBEDDING: False,
+        _KEY_OF_SELF_ATTN_MASK_TYPE: "no_mask",
+    },
+    # attrs25
+    {
+        _KEY_OF_TRANSPOSE_BS: False,
+        _KEY_OF_RELATIVE_EMBEDDING: False,
+        _KEY_OF_SELF_ATTN_MASK_TYPE: "no_mask",
+        _KEY_OF_WINDOW_SIZE: (2, 2),
+    },
+    # attrs26
+    {
+        _KEY_OF_TRANSPOSE_BS: False,
+        _KEY_OF_RELATIVE_EMBEDDING: False,
         _KEY_OF_SELF_ATTN_MASK_TYPE: "padding",
+        _KEY_OF_WINDOW_SIZE: (2, 2),
+    },
+    # attrs27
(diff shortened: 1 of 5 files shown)


tests/jax/test_layer.py
```diff
@@ -215,12 +215,53 @@ def enable_fused_attn():
...
+        _KEY_OF_RELATIVE_EMBEDDING: False,
+        _KEY_OF_SELF_ATTN_MASK_TYPE: "padding",
+        _KEY_OF_WINDOW_SIZE: None,
+    },
+    # attrs28
+    {
+        _KEY_OF_TRANSPOSE_BS: False,
+        _KEY_OF_RELATIVE_EMBEDDING: False,
+        _KEY_OF_WINDOW_SIZE: (2, 2),
+    },
```
cyanguwa, line 264: 1. are attrs22 and attrs23 the same? I assume if window size is not set, then it is None?
2. similarly, are attrs25 and attrs28 equivalent? 
3. could we add tests for `padding_causal`?
4. in the future, could we add `causal_bottom_right` and `padding_causal_bottom_right` as well? They should be supported at the moment, but it seems that they are missing from the docstring/support matrix and the unit tests.

transformer_engine/jax/flax/transformer.py
```diff
@@ -447,6 +447,14 @@ class DotProductAttention(nn.Module):  # pylint: disable=too-few-public-methods
 
         .. note:: THD format only supports 'padding' or 'causal_padding' mask type.
 
+       attn_mask_type       mask/sequence_descriptor       SW         softmax type
```
cyanguwa, line 450: I guess you meant "SWA" here instead of "SW".

cyanguwa (approved):

transformer_engine/jax/flax/transformer.py
KshitijLakhani: @cyanguwa I'm using the convention used in AND, OR logic etc for [Don't Care](https://en.wikipedia.org/wiki/Don%27t-care_term). It suggests that does not matter whether SWA is Yes, No, None the Softmax type will be the same.
Do you suggest we be explicit and I replace X with Yes/No/None instead ?

KshitijLakhani: @cyanguwa Resolved in SHA 47e151a8f40cc02ec136d4a72e4c9c090205bb30

PR #1683, opened 2025-04-15
[PyTorch] Move swizzle scaling factor to cpp

yaox12:
# Description

1. Move swizzle scaling factor to cpp, so we don't need to reset it after the gemm.
2. Replace `c10::optional` with `std::optional` since we're using C++17. PyTorch upstream is deprecating `c10::optional`. See https://github.com/pytorch/pytorch/pull/150464.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/common/transformer_engine.cpp
+++ b/transformer_engine/common/transformer_engine.cpp
@@ -258,11 +258,7 @@ size_t nvte_tensor_numel(const NVTETensor tensor) {
 size_t nvte_tensor_element_size(const NVTETensor tensor) {
   if (tensor == nullptr) return sizeof(float);
   const auto &t = *reinterpret_cast<const transformer_engine::Tensor *>(tensor);
-  if (t.data.dptr != nullptr) {
-    return transformer_engine::typeToSize(t.data.dtype);
-  } else {
-    return transformer_engine::typeToSize(t.columnwise_data.dtype);
-  }
+  return transformer_engine::typeToSize(t.dtype());
 }
 
 void *nvte_tensor_data(const NVTETensor tensor) {
--- a/transformer_engine/common/transformer_engine.cpp
+++ b/transformer_engine/common/transformer_engine.cpp
@@ -258,7 +258,11 @@ size_t nvte_tensor_numel(const NVTETensor tensor) {
 size_t nvte_tensor_element_size(const NVTETensor tensor) {
   if (tensor == nullptr) return sizeof(float);
   const auto &t = *reinterpret_cast<const transformer_engine::Tensor *>(tensor);
-  return transformer_engine::typeToSize(t.data.dtype);
+  if (t.data.dptr != nullptr) {
+    return transformer_engine::typeToSize(t.data.dtype);
+  } else {
+    return transformer_engine::typeToSize(t.columnwise_data.dtype);
+  }
 }
 
 void *nvte_tensor_data(const NVTETensor tensor) {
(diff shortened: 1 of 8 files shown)


transformer_engine/pytorch/csrc/extensions/gemm.cpp
```diff
@@ -175,8 +175,14 @@ std::vector<py::object> gemm(py::handle A, bool transa, py::handle B, bool trans
   const int sm_count = transformer_engine::cuda::sm_count(device_id);
   int num_math_sms = sm_count - transformer_engine::getenv<int>("NVTE_EXT_MARGIN_SM", sm_count);
 
+  std::vector<std::optional<at::Tensor>> original_scale_inverses_list;
```
timmoon10, line 178: We're actually holding onto the swizzled scales, not the original:
```suggestion
  std::vector<std::optional<at::Tensor>> swizzled_scale_inverses_list;
```
The purpose of this list is also a little unintuitive since we never use it. Perhaps we should add a comment that it is keeping PyTorch tensors alive.

We should make similar changes in `te_general_grouped_gemm`.

timmoon10 (approved):
This is a nice change. The previous swizzling logic was pretty weird and this makes the Python-C++ boundary much more reasonable for MXFP8 tensors.

transformer_engine/common/transformer_engine.cpp
```diff
@@ -258,7 +258,11 @@ size_t nvte_tensor_numel(const NVTETensor tensor) {
 size_t nvte_tensor_element_size(const NVTETensor tensor) {
   if (tensor == nullptr) return sizeof(float);
   const auto &t = *reinterpret_cast<const transformer_engine::Tensor *>(tensor);
-  return transformer_engine::typeToSize(t.data.dtype);
+  if (t.data.dptr != nullptr) {
+    return transformer_engine::typeToSize(t.data.dtype);
+  } else {
+    return transformer_engine::typeToSize(t.columnwise_data.dtype);
+  }
```
ptrendx, line 265:
```suggestion
    return transformer_engine::typeToSize(t.dtype());
```

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -396,7 +396,8 @@ void nvshmem_finalize();
  * swizzle
  **************************************************************************************************/
 
-void swizzle_scaling_factors(transformer_engine::TensorWrapper &input, bool trans);
+std::optional<at::Tensor> swizzle_scaling_factors(transformer_engine::TensorWrapper &input,
```
ptrendx, line 399: I don't like this function being in this header, since it is no longer an extension but rather internal utility function.

ptrendx: Also I wonder if we shouldn't define a separate type for this return type. For example
```cpp
using BackingStorage = std::optional<std::vector<at::Tensor>>;
```
to make it extra clear what the purpose of this is.

transformer_engine/common/transformer_engine.cpp
yaox12: Updated。

transformer_engine/pytorch/csrc/extensions/gemm.cpp
yaox12: Updated.

transformer_engine/pytorch/csrc/extensions.h
yaox12: Moved it to ``transformer_engine/pytorch/csrc/util.h`.

As for the second comment, I didn't define a seprate type but add comments to this swizzling function and where we hold the tensors.

timmoon10 (approved):

PR #1686, opened 2025-04-15
[PyTorch] Bunch of memory management fixes

pggPL:
# Description

This PR changes:
- adds tensor removing after prepare for saving for the internal tensors,
- cached weight is changed not to be internal tensor - otherwise it would be removed between the microbatches,
- test that catches the error above,
- scaling factors are saved for backward - it enables offloading them,
- enables weight offloading test for MXFP8 - it was disabled because of the problems with the scaling factors offloading.
- fixes .clear() - removes also scale_inv tensors and works correctly for .*TensorBase

**This PR breaks compatibility with the Thunder**

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
```diff
@@ -88,6 +88,13 @@ def __new__(
 
         return instance
 
+    def clear(self):
+        """Deallocate this tensor's memory. Typically not needed and must be used carefully."""
+        self._data = torch.Tensor() if self._data is not None else None
+        self._transpose = torch.Tensor() if self._transpose is not None else None
```
timmoon10, line 94: It's valid for `_data` and `_transpose` to be `None`, so we could simplify:
```suggestion
        self._data = None
        self._transpose = None
```

transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
```diff
@@ -100,7 +107,10 @@ def get_metadata(self) -> Dict[str, Any]:
 
     def prepare_for_saving(self) -> Tuple[list[Optional[torch.Tensor]], Float8TensorBase]:
         """Prepare the tensor base for saving for backward"""
-        tensors = [self._data, self._transpose]
+        tensors = [self._data, self._transpose, self._scale_inv]
+        self._data = None
+        self._transpose = None
+        self._scale_inv = None
```
timmoon10, line 113: We make assumptions elsewhere that `_scale_inv` is a tensor with a single entry, so we could encounter weird bugs if we ever touch the tensor in between saving and restoring. Perhaps this tensor is small enough that it's fine to remain on GPU?

timmoon10, line 113: We've gone back and forth a few times about whether to clear the data when saving (added in https://github.com/NVIDIA/TransformerEngine/pull/1500, removed in https://github.com/NVIDIA/TransformerEngine/pull/1548). Based on the last discussion (https://github.com/NVIDIA/TransformerEngine/pull/1548#pullrequestreview-2668006078, https://github.com/NVIDIA/TransformerEngine/pull/1548#issuecomment-2707658650), making sure that the weight cache produces PyTorch tensors should fix the bugs we were seeing before. Pinging @ksivaman and @ptrendx for a sanity check.

transformer_engine/pytorch/tensor/_internal/mxfp8_tensor_base.py
```diff
@@ -81,6 +81,15 @@ def __new__(
...
         return instance
 
+    def clear(self):
+        """Deallocate this tensor's memory. Typically not needed and must be used carefully."""
+        self._rowwise_data = torch.Tensor() if self._rowwise_data is not None else None
+        self._columnwise_data = torch.Tensor() if self._columnwise_data is not None else None
+        self._rowwise_scale_inv = torch.Tensor() if self._rowwise_scale_inv is not None else None
+        self._columnwise_scale_inv = (
+            torch.Tensor() if self._columnwise_scale_inv is not None else None
+        )
```
timmoon10, line 91:
```suggestion
        self._rowwise_data = None
        self._columnwise_data = None
        self._rowwise_scale_inv = None
        self._columnwise_scale_inv = None
```

transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py
```diff
@@ -56,6 +56,15 @@ def __new__(
...
         return instance
 
+    def clear(self):
+        """Deallocate this tensor's memory. Typically not needed and must be used carefully."""
+        self._rowwise_data = torch.Tensor() if self._rowwise_data is not None else None
+        self._columnwise_data = torch.Tensor() if self._columnwise_data is not None else None
+        self._rowwise_scale_inv = torch.Tensor() if self._rowwise_scale_inv is not None else None
+        self._columnwise_scale_inv = (
+            torch.Tensor() if self._columnwise_scale_inv is not None else None
+        )
```
timmoon10, line 66:
```suggestion
        self._rowwise_data = None
        self._columnwise_data = None
        self._rowwise_scale_inv = None
        self._columnwise_scale_inv = None
```

tests/pytorch/test_sanity.py
```diff
@@ -382,7 +384,11 @@ def _test_sanity_common(block, dtype, config, fp8_recipe, skip_wgrad, skip_dgrad
 
     use_fp8 = fp8_recipe is not None
     with fp8_autocast(enabled=use_fp8, fp8_recipe=fp8_recipe):
-        te_out = block(te_inp)
+        if not microbatching:
+            te_out = block(te_inp)
+        else:
+            _ = block(te_inp, is_first_microbatch=True)
+            te_out = block(te_inp, is_first_microbatch=False)
```
timmoon10, line 391: This is nice, we've been needing this test for a while.

timmoon10 (dismissed):
Overall LGTM, with some minor suggestions.

transformer_engine/pytorch/module/base.py
```diff
@@ -1064,7 +1064,14 @@ def get_weight_workspace(
                 raise ValueError(
                     "tensor and quantizer kwargs must be provided to construct FP8 workspace"
                 )
+
+            # Ensure the tensor in the cache is an instance of torch.Tensor,
+            # as it persists beyond a single forward pass.
+            # Setting internal=True would cause the data to be removed in prepare_for_saving(...).
+            quantizer_internal = quantizer.internal
+            quantizer.internal = False
```
ptrendx, line 1072: Shouldn't this only happen when `cache_name is not None` (so only when we actually use the cache)?

transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py
ptrendx: @timmoon10 You actually need to set them to `torch.Tensor()` here AFAIK - otherwise the deallocation happens only after the autograd function where the `clear` method was called, rather than right away.

transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
ptrendx: Same comment as for blockwise tensor.

transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
ptrendx: I am aware of this - this PR came from my discussion with Pawel on Monday :-). But I agree - we need to check this carefully with multiple scenarios. Maybe @guyueh1 could help run the NeMo test suite with this change?

transformer_engine/pytorch/tensor/_internal/mxfp8_tensor_base.py
ptrendx: Same comment as with blockwise tensor.

ptrendx (commented):
We need to tell Thuder people once this gets merged - this will change number of tensors being saved between forward and backward.

transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py
timmoon10: Won't they be handled by Python GC in the same way? If we want to eagerly deallocate, we need to delve into the PyTorch tensor:
```python
        for tensor in (self._rowwise_data, self._columnwise_data, self._rowwise_scale_inv, self._columnwise_scale_inv):
            if tensor is not None:
                tensor.data = torch.Tensor()
        self._rowwise_data = None
        self._columnwise_data = None
        self._rowwise_scale_inv = None
        self._columnwise_scale_inv = None
```

transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
pggPL: We use `prepare_for_saving` only for TensorBase classes, which are internal. QunatizedTensors are saved inside ctx and this function is not invoked.

transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py
pggPL: I just copied the behavior from the Float8Tensor class from the code, but you are right @ptrendx - it is wrong. I changed that.

transformer_engine/pytorch/module/base.py
pggPL: You are right.

transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
```diff
@@ -90,6 +90,13 @@ def __new__(
 
         return instance
 
+    def clear(self):
+        """Deallocate this tensor's memory. Typically not needed and must be used carefully."""
+        self._data.data = torch.Tensor() if self._data is not None else None
```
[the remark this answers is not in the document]
pggPL: You are absolutely right!

ptrendx:
@guyueh1 offline confirmed that the NeMo/MCore sanity tests pass with this change, including parallelisms.

ptrendx (dismissed):

ptrendx (dismissed):

transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
```diff
@@ -102,7 +109,10 @@ def get_metadata(self) -> Dict[str, Any]:
 
     def prepare_for_saving(self) -> Tuple[list[Optional[torch.Tensor]], Float8TensorBase]:
         """Prepare the tensor base for saving for backward"""
-        tensors = [self._data, self._transpose]
+        tensors = [self._data, self._transpose, self._scale_inv]
```
ksivaman, line 112:
```suggestion
        tensors = [self._data, self._transpose]
```

ksivaman: Note: For `DelayedScaling`, removing the `scale_inv` is causing the fused attention tests to fail @pggPL

transformer_engine/pytorch/utils.py
```diff
@@ -33,7 +31,7 @@ def clear_tensor_data(*tensors: Tuple[Optional[torch.Tensor], ...]) -> None:
     """
     for t in tensors:
         if t is not None:
-            if isinstance(t, QuantizedTensor):
+            if hasattr(t, "clear"):
```
ksivaman, line 34: Nit: I understand why this change was made but it would be robust would be to check for each base type separately and not for this attribute which PyTorch could add in the future to their tensors. Maybe we can come up with some common dummy base class for all the _internal BaseTensors for these types of situations? @ptrendx

transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
pggPL: It was not that - we incorrectly compute owns_input in Linear and cleared tensor we should not clear. I fixed this issue in new commit.

timmoon10 (dismissed):
LGTM

ksivaman (approved):

timmoon10:
The Thunder integration bug is fixed with https://github.com/Lightning-AI/lightning-thunder/pull/1826.

PR #1707, opened 2025-04-21
[PyTorch] FP8 Subchannel Recipe With FP8 Gather And Configurable Scaling Factor Tensor Swizzling

zhongbozhu:
# Description

Holding on for the merge of https://github.com/NVIDIA/TransformerEngine/pull/1711 which has major changes towards linear modules. 

Two Goals:

1. Support Doing FP8 gather for the FP8 subchannel recipe, get rid of the force high precision gather branch, the benefit is that we don't have to save full precision tensor for backward, lower memory usage
2. Bookkeeping of whether a FP8 subtensor scaling has a compact or non-compact scaling factor tensor. CuBLAS requires certain swizzling of the scaling factors layouts (transpose, padding), in the case where no gather will be performed, quantization kernel should directly output the desired shape. Otherwise, it should output a compact format. 

What needs to be done:
1. Need a 1x128 quantize kernel that supports three modes: quantize only, quantize transpose and quantize non-transpose, here the quantize non-transpose means that using one kernel to do 1x128 and 128x1 quantization, but for the 128x1 quantize, we skip the transpose for it, for the purpose of doing gather of it in backward 
2. Enable coalesced gather of FP8 quantized tensor and its scaling factors
3. Need to correctly handle whether the scaling factor tensor has been transposed or padded, as opposed to being plain compact tensor. 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- CUDA Kernel Ready
- Scaling Factor Swizzling Control
- FP8 Gather
- Integrate dequantize function from https://github.com/kwyss-nvidia/TransformerEngine/commit/00c433232c1348fdf30d6831f9a50fb38e535cb0

## Unit Tests

```
# quantizer test
pytest tests/pytorch/test_float8_blockwise_scaling_exact.py::test_quantization_1D_block_tiling_with_compact_data_and_scales -s -v

# layer test
pytest tests/pytorch/test_float8_blockwise_scaling_exact.py::TestFP8BlockScalingRecipeLayerNormLinear::test_fp8_current_scaling_with_layernorm_linear_module -s -v

# distributed test with fp8 gather 
pytest tests/pytorch/distributed/test_numerics.py -s -v
```

# Checklist:


transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -385,6 +410,103 @@ __global__ void __launch_bounds__(kThreadsPerBlock) block_scaled_1d_cast_transpo
...
+    const int r_s = thr_idx_in_warp * thr_tile_row;                 // Row in shared memory
+    int c_s = warp_idx * num_smem_reads;                            // Column in shared memory
+    size_t r_g = static_cast<size_t>(blockIdx.y) * kTileDim + r_s;  // Row in global memory
+    const size_t c_g =
+        static_cast<size_t>(blockIdx.x) * kTileDim + c_s * kNVecSMem;  // Column in global memory
+    const size_t num_ele = c_g < row_length
+                               ? min(static_cast<size_t>(thr_tile_col), row_length - c_g)
+                               : 0;  // For not aligned case
+#pragma unroll
+    for (int iter = 0; iter < num_iterations; ++iter) {
```
kwyss-nvidia, line 439: Nit: This loop can be removed since num_iterations must be 1.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -385,6 +410,103 @@ __global__ void __launch_bounds__(kThreadsPerBlock) block_scaled_1d_cast_transpo
...
+        CType scale;
+        scale = compute_scale_from_types<IType, OType>(amax, epsilon, pow_2_scaling);
+        thr_scale.data.elt[reg_idx] = scale;
+        // Step 3.5: Write scale_inv_t
+        bool write_scale_inv = is_src_lane;
+        if constexpr (!kAligned) {
+          write_scale_inv &= (c_g + reg_idx < row_length);
+        }
+        if (write_scale_inv) {
+          CType scale_inv = 1.0 / scale;
```
kwyss-nvidia, line 481: Nit: `__frcp_rn` could be more unambiguous. `__fdiv_rn` is used in compute_scale because it is consistent independent of the fast_math compiler flags, while there are potential gotchas with div operator.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -385,6 +410,103 @@ __global__ void __launch_bounds__(kThreadsPerBlock) block_scaled_1d_cast_transpo
...
+          size_t row_idx = static_cast<size_t>(blockIdx.y);
+          size_t col_idx = static_cast<size_t>(blockIdx.x) * kTileDim + c_s * kNVecSMem + reg_idx;
+          tile_scales_inv_t[row_idx * scale_t_stride_y + col_idx * scale_t_stride_x] = scale_inv;
```
kwyss-nvidia, line 484: Assuming that scales are compact, and scale_t_stride_x reflects that, there's an opportunity for VecStore here.

In my opinion, the kernel API is more flexible than necessary. It's OK to enforce dense compact scales for this setting of non-transposed columnwise data.

transformer_engine/pytorch/csrc/common.h
```diff
@@ -172,9 +172,12 @@ class Float8BlockQuantizer : public Quantizer {
   bool force_pow_2_scales = false;
   // Amax within quantization tile has a floor of epsilon.
   float amax_epsilon = 0.0;
+  // If true, the columnwise scaling will be transposed.
+  bool columnwise_transpose = true;
 
  private:
   int block_scaling_dim = 2;
+  bool compact_scales = false;
```
kwyss-nvidia, line 180: My preference is for enumerating the type of column formats we require. Certain combinations of columnwise_transpose and compact_scales are not useful, so I prefer if they're not configurable and we won't waste time supporting them.

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -136,10 +142,10 @@ def get_scale_shape(self, shape: Iterable[int], columnwise: bool) -> Tuple[int,
         assert self.block_scaling_dim == 1, "Only 1D or 2D blocks supported"
         if columnwise:
             outer = math.ceil(M / self.block_len)
-            inner = round_up_to_nearest_multiple(K, 4)
-            return (outer, inner)
+            inner = round_up_to_nearest_multiple(K, 4) if not self.compact_scales else K
+            return (outer, inner) if self.columnwise_transpose else (inner, outer)
         outer = math.ceil(K / self.block_len)
-        inner = round_up_to_nearest_multiple(M, 4)
+        inner = round_up_to_nearest_multiple(M, 4) if not self.compact_scales else M
```
kwyss-nvidia, line 148: Don't we want compact scales to also imply the scale dimensions aren't transposed?

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
zhongbozhu: I was leaving some possibility for some parameter tuning in which `num_iterations` might be larger than 1.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
zhongbozhu: The tricky thing is, currently the `scale_t_stride_y` is given a value not as we expected: it was given a 1. It gives room for changing the input args to determine whether this scale_inv tensor is a transposed tensor. This pattern is the same across all the three modes, and personally I think it's quite nice if we can just use input args to repurpose the kernel.

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
zhongbozhu: yeah, when I was working on the kernel, it forgot to consider the transpose requirement of scales. So I hacked the test case to pass numerical tests of the kernel. Will change it after we came up with a better API.

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
zhongbozhu: thanks for the suggestion

transformer_engine/common/transpose/cast_transpose.h
```diff
@@ -31,26 +31,27 @@ void quantize_transpose_square_blockwise(const SimpleTensor &input, SimpleTensor
...
   NONE,
   // Columnwise data transposed from original shape.
   // Scales in GEMM format corresponding to GEMM ingesting transposed column data.
-  COLUMNWISE_TRANSPOSE,
-  // TODO: FP8 all gather requires some changes.
-  // 1. The transpose gets in the way of the all gather.
-  // 2. Compact scales are better for gathering than the GEMM format.
-  COLUMNWISE
+  // On Hopper sm90, GEMM_READY means that columnwise quantization also fuses transpose op
+  // On higher sm versions with TN,NT,NN fp8 gemm, GEMM_READY doesn't fuse transpose
```
kwyss-nvidia, line 50: We don't have a GEMM yet integrated for sm versions other than hopper. The less said for now the better.

kwyss-nvidia:
The dequantize code for the quantized tensor assumes GEMM_READY format. It should at least assert that assumption.

kwyss-nvidia:
I like the new API with the GEMM_READY and COMPACT formats. Whether to override `set_usage` to also include the new settings I'm less sure about.

kwyss-nvidia:
There are fields on the Quantizer that capture the RowFmt and ColFmt. I think it will also be valuable to have fields on the QuantizedTensor that indicate how the data was quantized. I know the tensor has a handle to the quantizer, but the quantizer's usage can be modified, so having static descriptions of the data would be more reliable.

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -42,13 +45,45 @@ def __init__(
         amax_epsilon: float = 0.0,
         force_pow_2_scales: bool = True,
         block_scaling_dim: int = 2,
+        need_gather: bool = False,
```
yaox12, line 48: I suggest using `compact` as the argument as this feature is needed in other scenarios as well, for example, FP8 alltoall and token permutation (memory reordering in the `m` dimension).

zhongbozhu: using `need_compact` right now to avoid duplicated entries in codebase search

transformer_engine/pytorch/distributed.py
```diff
@@ -953,6 +955,41 @@ def _all_gather_fp8(
     return out, handle
 
 
+def _post_process_fp8_blockwise_gather(
```
timmoon10, line 958: I really don't like that this is reintroducing recipe-specific logic. How about we modify `_all_gather_fp8_blockwise` so it returns a custom async handle:
```python
@dataclass
class _FP8BlockwiseAllGatherAsyncHandle:

    tensor: Float8BlockwiseQTensorBase
    quantizer: Float8BlockQuantizer
    async_handle: torch.distributed.Work
    _synchronized: bool = False

    def wait(self) -> None:
        if self._synchronized:
            return
        self.async_handle.wait()
        _post_process_fp8_blockwise_gather(self.tensor, self.quantizer)
        self._synchronized = True

def _all_gather_fp8_blockwise(...):

    ...

    # Post-process after communication
    if async_op:
        handle = _FP8BlockwiseAllGatherAsyncHandle(out, quantizer, handle)
    else:
        _post_process_fp8_blockwise_gather(out, quantizer)

    return out, handle
```

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -27,6 +27,26 @@ def _empty_tensor() -> torch.Tensor:
     return torch.Tensor().cuda()
 
 
+# for fp8 blockwise recipe, we supproted options to output compact format versus GEMM_READY format for quantization
+# but for FP8 GEMMs, only TN layout is supported in hopper, so there transpose needs to be done before calling the GEMM
+# sanity check here to make sure all the data & scale formats have been set to GEMM_READY, not compact
+def _sanity_check_data_format(A, B, layout):
```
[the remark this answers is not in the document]
zhongbozhu: The issue is actually about nvte_shape, when you have both rowwise data and columnwise data, the shape function will always check the shape of its rowwise data, so if I "forget" about doing the transpose of colunwise data, its shape is not really checked. 

We also cannot just check shape for because we are only supposed to make a data GEMM_READY after comm right before the GEMM, so it's not a good idea that a simple `shape` function should know who is calling it. 

I have also thought about adding compact/gemm_ready into Tensor struct, but it's also too recipe specific, which is finally the reason why I did it in python here.

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -42,13 +45,45 @@ def __init__(
...
+            else tex.ColwiseFmt.GEMM_READY_DATA_AND_SCALES
+        )
+
+    # override base class set_usage method
+    def set_usage(
+        self,
+        *,
+        rowwise: Optional[bool] = None,
+        columnwise: Optional[bool] = None,
+        need_compact: Optional[bool] = None,
```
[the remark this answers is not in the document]
zhongbozhu: I also like the idea of putting it into the base class, but compact versus gemm_ready doesn't make sense for float8 & mxfp8. 

Fp8 is neither compact or GEMM_READY => no swizzle of scale, but its columnwise data is always transposed, and since it's just a data transpose, it saves us a lot of trouble

MXFP8 data is always compact, but its scales have padding and needs to be swizzled before GEMM.

transformer_engine/pytorch/distributed.py
zhongbozhu: Good idea

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -689,6 +693,16 @@ class TensorWrapper {
...
 
+/*! \enum RowwiseFmt
+ *  \brief Rowwise format for Fp8 Sub-tensor scaling where scale_inv is not scalar tensor.
+ */
+enum class RowwiseFmt { GEMM_READY_DATA_AND_SCALES = 0, COMPACT_DATA_AND_SCALES = 1 };
+
+/*! \enum ColwiseFmt
+ *  \brief Columnwise format for Fp8 Sub-tensor scaling where scale_inv is not scalar tensor.
+ */
+enum class ColwiseFmt { GEMM_READY_DATA_AND_SCALES = 0, COMPACT_DATA_AND_SCALES = 1 };
```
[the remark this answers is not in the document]
zhongbozhu: But here the compact also means that the columnwise data is not transposed, so in case TP=1, we are paying one extra transpose op. With Blackwell supporting more than just TN gemm has saved too much software headache, but we cannot enjoy that for subchannel.

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -302,6 +302,13 @@ enum NVTEQuantizationConfigAttribute {
    conditional early even when captured in a static CUDA graph.
   */
   kNVTEQuantizationConfigNoopTensor = 2,
+  /*! Data format for an FP8 block-scaled tensor
+   *
+   *  This is not the right design since the tensor format is a
+   *  property of the tensor, not the quantization. This enum will
+   *  likely be refactored away in the future.
+   */
+  kNVTEQuantizationConfigFloat8BlockScaleTensorFormat = 3,
```
timmoon10, line 311: This is not the right design since the data format is not specific to quantization. This should be part of the tensor class since the data format is needed to properly use the tensor, e.g. when dequantizing or performing a GEMM. However, the current implementation of the tensor class makes it complicated and intrusive to add the data format. We can accept this as a temporary hack.

Related: https://github.com/NVIDIA/TransformerEngine/pull/1707#discussion_r2072256608

timmoon10 (dismissed):
LGTM, pending CI

timmoon10 (dismissed):

timmoon10 (approved):

timmoon10 (approved):

PR #1712, opened 2025-04-23
Warn when using fp8 weights + non-fp8 computation

kunlunl:
# Description

Currently, TE's `Linear`, `LayerNormLinear` and `GroupedLinear` de-quantized the fp8 weights to bf16 when self.fp8 is False. This MR changes the de-quantization to raising an error, to prevent fp8 weights + non-fp8 computation.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/grouped_linear.py
+++ b/transformer_engine/pytorch/module/grouped_linear.py
@@ -677,8 +677,10 @@ class GroupedLinear(TransformerEngineBaseModule):
             weight_tensors = [getattr(self, f"weight{i}") for i in range(self.num_gemms)]
             bias_tensors = [getattr(self, f"bias{i}") for i in range(self.num_gemms)]
             if not self.fp8:
-                warnings.warn("You are using quantized weights without quantized compute. "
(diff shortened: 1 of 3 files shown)


timmoon10 (commented):
This is technically fine, but I don't really like it philosophically. I prefer designing our behavior to be as logical and consistent as possible, and then giving users the freedom to do whatever they want. Some users might do something stupid, but others might be doing something novel and clever. I don't like that this check entangles weight initialization and compute, and I could imagine some narrow use-cases where users might want FP8 weights and BF16 compute (aggressive memory optimization, initializing amax histories).

I definitely see the argument for prioritizing user-friendliness though and I wouldn't fight this change if @ksivaman or @ptrendx have other opinions. Even if we keep the existing behavior we should add a warning so that users know that they're doing something weird.

kunlunl:
OK, I changed it to using warnings instead of raising an error.

timmoon10 (approved):

PR #1717, opened 2025-04-24
Support `nvidia-cu*` wheels for core lib compilation; miscellaneous build improvements

### LINKED ISSUE #1400: Import fails when working from a TE directory
When working from a TE directory in the case where TE is installed at a different location, import fails as the shared object cannot be found. This is because `import transformer_engine` call in `transformer_engine/common/__init__.py` prioritizes importing from current working directory as opposed to the actual install location, and then [get_te_path](https://github.com/NVIDIA/TransformerEngine/blob/7b861e75d7590f98a3450b85fe90a35343947b75/transformer_engine/common/__init__.py/#L29) returns the incorrect result.

ksivaman:
# Description

Various improvements to build system and wheels, detailed in the _**changes**_ section.

Fixes #1400 

## Type of change

- [x] Infra/Build change
- [x] Code refactoring

## Changes

- Use wheels for cudart, cublas, and cudnn if cuda toolkit is not available during compile time.
- Remove `nvcc` dependency completely from JAX framework extensions. This includes testing that the JAX sdist is installable in a no-nvcc env.
- No need to uninstall wheel packages during source installation. This is also needed for future support with UV, and resolves some esoteric bugs when compiling in virtual envs.
- List the frameworks and necessary nvidia libraries as setup requirements. In a future refactor to support `pyproject.toml`, these will move to the build requirements.
- Properly check for TE install path in `get_te_path()`.

## Next work

- Make PyTorch extensions `nvcc` independent and change CUDA extension to CPP extension.
- If/when `nvcc` wheels are made available the cuda toolkit dependency should be fully erased, even from core lib compilation.

# Checklist:


timmoon10 (commented):
Does the CUDA Python package include nvcc?
- If yes: How are we going to build the core lib bdist? If we are committing to build with the CUDA Python packages, then I think this PR is reasonable. If we build with the CUDA Toolkit, we need runtime logic in `transformer_engine/common/__init__.py` to load CUDA `.so`s if the CUDA Toolkit is not detected.
- If no: The changes to the core lib build system are not needed and we need runtime logic to load CUDA `.so`s if needed.

build_tools/utils.py
```diff
@@ -185,7 +208,34 @@ def cuda_path() -> Tuple[str, str]:
     if not nvcc_bin.is_file():
         raise FileNotFoundError(f"Could not find NVCC at {nvcc_bin}")
 
-    return cuda_home, nvcc_bin
+    return nvcc_bin
+
+
+@functools.lru_cache(maxsize=None)
+def get_cuda_include_dirs() -> Tuple[str, str]:
+    """Returns the NVCC include directory found via pip package."""
```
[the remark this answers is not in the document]
ksivaman: Done

build_tools/utils.py
```diff
@@ -199,18 +249,37 @@ def cuda_archs() -> str:
...
+        return tuple(int(v) for v in version)
+
+    try:
+        version_str = version("nvidia-cuda-runtime-cu12")
+        version_tuple = tuple(int(part) for part in version_str.split(".") if part.isdigit())
+        return version_tuple
+    except importlib.metadata.PackageNotFoundError:
+        raise RuntimeError(
+            "Cannot return cuda version as both NVCC executable and cudaRT pip wheel were not"
+            " found."
```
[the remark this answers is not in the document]
ksivaman: Done

transformer_engine/common/CMakeLists.txt
```diff
@@ -23,13 +23,49 @@ endif()
...
+
+  if (NOT EXISTS "${cuda_wheel_path}")
+    message(FATAL_ERROR "CUDA not found in pip wheels either. Please install CUDAToolkit or nvidia-cuda-runtime-cu12 (pip).")
+  endif()
+  if (NOT EXISTS "${cublas_wheel_path}/lib")
+    message(FATAL_ERROR "CUBLAS not found in pip wheels either. Please install CUDAToolkit or nvidia-cublas-cu12 (pip).")
+  endif()
+  if (NOT EXISTS "${cudnn_wheel_path}/lib")
+    message(FATAL_ERROR "CUDNN not found in pip wheels either. Please install CUDAToolkit or nvidia-cudnn-cu12 (pip).")
+  endif()
```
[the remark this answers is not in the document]
ksivaman: Copy n paste error for the message. Answering the next part in a separate comment.

transformer_engine/common/CMakeLists.txt
```diff
@@ -23,13 +23,49 @@ endif()
...
+# If CUDA toolkit not found, try pip-installed Python packages
+if (NOT CUDAToolkit_FOUND)
+  message(WARNING "CUDAToolkit not found via find_package. Trying Python wheel fallback...")
+
+  set(cuda_wheel_path "${Python_SITELIB}/nvidia/cuda_runtime")
+  set(cublas_wheel_path "${Python_SITELIB}/nvidia/cublas")
+  set(cudnn_wheel_path "${Python_SITELIB}/nvidia/cudnn")
+
+  if (NOT EXISTS "${cuda_wheel_path}")
+    message(FATAL_ERROR "CUDA not found in pip wheels either. Please install CUDAToolkit or nvidia-cuda-runtime-cu12 (pip).")
```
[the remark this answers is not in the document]
ksivaman: Done

setup.py
```diff
@@ -90,7 +88,15 @@ def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
     """
 
     # Common requirements
-    setup_reqs: List[str] = []
+    setup_reqs: List[str] = [
```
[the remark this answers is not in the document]
ksivaman: We could technically remove these reqs from this PR and only add them when eventually we have a path where toolkit isn't needed at all for core build and pytorch. Just added these here to be consistency across core/jax/pytorch build requirements.

transformer_engine/pytorch/setup.py
```diff
@@ -55,7 +55,17 @@
         description="Transformer acceleration library - Torch Lib",
         ext_modules=ext_modules,
         cmdclass={"build_ext": CMakeBuildExtension},
-        install_requires=["torch"],
+        setup_requires=[
```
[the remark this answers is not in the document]
ksivaman: `torch.utils.cpp_extension.CUDAExtension` does not automatically detect the pip packages for `nvidia*` libs. The eventual goal for the pytorch extensions is to get converted to `CPPExtension` at which point we'll have to modify our `include_dirs` similar to how this PR does for the JAX setup.

transformer_engine/common/CMakeLists.txt
```diff
@@ -23,13 +23,49 @@ endif()
...
 
-# CUDA Toolkit
-find_package(CUDAToolkit REQUIRED)
-if (CUDAToolkit_VERSION VERSION_LESS 12.0)
-  message(FATAL_ERROR "CUDA 12.0+ is required, but found CUDA ${CUDAToolkit_VERSION}")
+# Attempt to find CUDA Toolkit normally
+find_package(CUDAToolkit)
+
+# If CUDA toolkit not found, try pip-installed Python packages
+if (NOT CUDAToolkit_FOUND)
```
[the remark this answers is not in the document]
timmoon10: I think it would be better to wait until that nvcc package is published, since this is currently dead code. For the time being, I think it's reasonable to require the CUDA Toolkit to build the core lib. Since we distribute it as a bdist, it shouldn't affect users who don't have the CUDA Toolkit.

transformer_engine/common/__init__.py
```diff
@@ -123,7 +151,25 @@ def _load_nvrtc():
...
 
 
+def _load_cublas():
+    """Load CUBLAS shared library."""
+
+    # Attempt to locate CUBLAS in Python dist-packages
+    lib_path = glob.glob(
+        os.path.join(
+            sysconfig.get_path("purelib"),
+            f"nvidia/cuda_nvrtc/lib/libnvrtc.{_get_sys_extension()}.*[0-9]",
```
timmoon10, line 161: Should this be something like:
```suggestion
            f"nvidia/cublas/lib/libcublas.{_get_sys_extension()}.*[0-9]",
```

transformer_engine/common/__init__.py
```diff
@@ -123,7 +151,25 @@ def _load_nvrtc():
     return ctypes.CDLL(f"libnvrtc.{_get_sys_extension()}", mode=ctypes.RTLD_GLOBAL)
 
 
+def _load_cublas():
```
timmoon10, line 154: Is this redundant with `_load_nvidia_cuda_library`?

transformer_engine/common/__init__.py
```diff
@@ -45,20 +51,32 @@ def _get_sys_extension():
...
-            len(lib_path) == 1
-        ), f"Found {len(lib_path)} libcudnn.{_get_sys_extension()}.x in nvidia-cudnn-cuXX."
-        return ctypes.CDLL(lib_path[0], mode=ctypes.RTLD_GLOBAL)
+
+    path_found = len(so_paths) > 0
+    ctypes_handles = []
+
+    if path_found:
+        for so_path in so_paths:
+            ctypes_handles.append(ctypes.CDLL(so_path, mode=ctypes.RTLD_GLOBAL))
```
timmoon10, line 73: If we find multiple `.so`s, does it make sense to load all of them? Also, we should probably throw a warning if we don't find anything.

transformer_engine/common/__init__.py
```diff
@@ -26,8 +26,14 @@ def is_package_installed(package):
...
 
-def get_te_path():
+def get_te_path() -> Path:
     """Find Transformer Engine install path using pip"""
+
+    # If TE is installed in the main python package directory, prioritize it
+    # for cases where user is running from a TE directory where an editable
+    # build is not installed.
+    if (Path(sysconfig.get_path("purelib")) / "transformer_engine").exists():
+        return Path(sysconfig.get_path("purelib"))
```
timmoon10, line 36: This is helpful for developers making pure Python changes, but it's inconvenient for my C++ workflow since you can't just change the local `.so`s. I think there are ways we could make this nicer (check install dir by default, add envvar to allow manual overload), but it may be better for a separate PR.
```suggestion
```

transformer_engine/common/__init__.py
ksivaman: Oops, from an earlier iteration, forgot to delete.

transformer_engine/common/__init__.py
ksivaman: If we don't use cuda toolkit for compilation (for JAX extensions in this PR but for everything in future PRs) then we need to load all SOs, for e.g. `libcudnn`, `libcudnn_graph`, etc.. Since this strictly checks the nvidia packages only, the SOs are never duplicate here.

I avoided throwing any errors or warnings here is because this isn't a necessary path (e.g. if `cudatoolkit` was used for compilation) so it would be disruptive and would show up on every import of TE.

transformer_engine/common/__init__.py
ksivaman: Could you describe the workflow you refer to? If you have a local/editable build, then this change should be a noop as the TE directory would be absent from site-packages.

timmoon10 (approved):
Seems reasonable, once we get past the linter errors and pass CI.

This should definitely help users downloading the TE sdist from PyPI. Once we remove the nvcc dependency from the PyTorch extensions, the only "difficult" dependency will be the C++ compiler.

transformer_engine/common/__init__.py
timmoon10: https://xkcd.com/1172/

timmoon10 (approved):

PR #1722, opened 2025-04-26
Remove CPU overheads of torch.cuda.get_device_properties() by caching it

xrennvidia:
# Description

- Removed the CPU overheads of multiple calls to `torch.cuda.get_device_properties()` by caching it with `lru_cache` annotation.
- Fixed `batch_p2p_comm` check by making it aware of device compute capability and cp size.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:


timmoon10 (commented):
Couldn't we accomplish something similar by wrapping `get_device_compute_capability` with [`functools.lru_cache`](https://docs.python.org/3/library/functools.html#functools.lru_cache)? If we want it to respect `torch.cuda.current_device`, we could make a helper function:
```python
@functools.lru_cache
def _get_device_compute_capability(device: torch.device) -> Tuple[int, int]:
    props = torch.cuda.get_device_properties(device)
    return (props.major, props.minor)

def get_device_compute_capability() -> Tuple[int, int]:
    """CUDA compute capability of current GPU"""
```

This wouldn't require changes outside of `transformer_engine/pytorch.utils.py`.

transformer_engine/pytorch/utils.py
```diff
@@ -41,10 +41,10 @@ def clear_tensor_data(*tensors: Tuple[Optional[torch.Tensor], ...]) -> None:
             del t
 
 
-def get_device_compute_capability() -> Tuple[int, int]:
+def get_device_compute_capability() -> int:
```
timmoon10, line 44: We should think before we change the version format. As I see it, there are three reasonable ways to handle version comparisons in Python:
- [`packaging.version.Version`](https://packaging.pypa.io/en/latest/version.html#packaging.version.Version)
  - Pros: smart, general (even with non-int versions)
  - Cons: in the past `packaging` has not always been available, cumbersome to construct 
- `int`
  - Pros: matches CUDA convention for SM arch, very easy to construct
  - Cons: does not generalize (e.g. if you have major.minor format, you can't add patch), only supports int versions
- `tuple[int, ...]`
  - Pros: general (although comparisons are only valid for int versions), easy to construct
  - Cons: unintuitive

xrennvidia: I have no strong preference, which one do you want me to change to?

timmoon10: Let's revert the API to use tuples.

Interally, I can see the argument for using `transformer_engine::cuda::sm_arch`. That's what's used in the C++ core lib.

xrennvidia: BTW, TE-JAX uses `int`, reverting this to `tuples` will be inconsistent to TE-JAX.

cyanguwa: Thanks for the offline discussion. I think I'm leaning towards `Tuple` vs `int`, because it's more Pythonic and easier to construct/do comparison with. I wondered about the performance of `lru_cache` vs `sm_arch`, but I think they might just be the same (`lru_cache` hashes the numbers before saving them whereas `sm_arch` saves the numbers in some static vector<int>). The minor performance difference may not be important. What's important is that we are caching, as opposed to not caching before.

xrennvidia: OK, I can revert to `tuple` by adding `lru_cache` annotation. Actually, that will make the PR much simpler :)

xrennvidia: Reverted

vasunvidia (approved):
Looks good to me.

PR #1724, opened 2025-04-28
Enhance recipe compatibility

negvet:
# Description

Enables recipe update on the fly. Added a user warning when this happens.

Enables TE1.x checkpoint loading.

Implemented check that is supposed to catch user-defined bug, when there is a mismatch between the recipes in `fp8_model_init` and `fp8_autocast`.
Example case to check: recipe is `DelayedScaling` (`DelayedScaling` is set in `fp8_autocast()`), but the weight tensor is `MXFP8Tensor` (`MXFP8BlockScaling` is set in `fp8_model_init()`).

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Added check in the base module class
- Added test

# Checklist:


transformer_engine/pytorch/module/base.py
```diff
@@ -1179,6 +1182,48 @@ def _validate_name(self):
...
+        but the weight tensor is MXFP8Tensor (MXFP8BlockScaling is set in fp8_model_init()).
+        """
+        if not self.fp8 and not self.fp8_calibration:
+            return
+        if not hasattr(self, "weight_names") or not self.weight_names:
+            return
+
+        recipe = self.fp8_meta["recipe"]
+        expected_tensor_class = None
+        if recipe.delayed() or recipe.float8_current_scaling():
```
ptrendx, line 1206: Maybe add a field in the recipe that has this exepected tensor class? Also, I guess for completeness we should not only deal with QuantizedTensor, but also  with the *Base (e.g. Float8TensorBase) classes (I know that they are not used for weights in fp8_model_init).

transformer_engine/pytorch/module/base.py
```diff
@@ -1179,6 +1182,48 @@ def _validate_name(self):
...
+            expected_tensor_class = Float8Tensor
+        elif recipe.mxfp8():
+            expected_tensor_class = MXFP8Tensor
+        elif recipe.float8_block_scaling():
+            expected_tensor_class = Float8BlockwiseQTensor
+        else:
+            raise RuntimeError(f"Unsupported recipe type: {recipe.__class__.__name__}")
+
+        weight_tensors = [getattr(self, name) for name in self.weight_names]
```
ptrendx, line 1215: @timmoon10 I vaguealy recall that getattr from nn.Module was slow and that you created a faster function for it at some point, do you remember the details?

transformer_engine/pytorch/module/base.py
negvet: > Maybe add a field in the recipe that has this exepected tensor class?

That is what I also like more. I did not do so due to circular import.
Now, I fixed it with local import, if this is acceptable.
But probably, we need some redesign to better fix this circular import issue (in another PR).

negvet: > Also, I guess for completeness we should not only deal with QuantizedTensor, but also with the *Base (e.g. Float8TensorBase) classes (I know that they are not used for weights in fp8_model_init).
Fixed

negvet:
TE1.x did not save the recipe in the state.
Based on that, the assumption in https://github.com/NVIDIA/TransformerEngine/pull/1724/commits/5023a32a0f704f63785fdfe6dff7966a2cfe7c9d is the following: if no recipe in state -> it is TE1.x checkpoint with DelayedScaling.
Does it sound like a safe enough assumption?

With this fix, TE1.x checkpoint can be loaded in ToT.

transformer_engine/pytorch/module/base.py
```diff
@@ -57,6 +57,12 @@
 _NUM_MAX_UB_STREAMS = 3
 _MIN_STREAM_PRIORITY, _MAX_STREAM_PRIORITY = None, None
 layers_atomic_ring_exchange = []
+_QUANTIZED_WEIGHT_TENSOR_TYPES = (
```
ptrendx, line 60: We recently merged the change to introduce `QuantizedTensorBase` class so you should be able to remove this.

transformer_engine/common/recipe/__init__.py
```diff
@@ -7,9 +7,14 @@
...
 import os
 from enum import Enum
-from typing import Literal, Optional, Union, Callable, NamedTuple
+from typing import Literal, Optional, Type, Union, Callable, NamedTuple
 from pydantic.dataclasses import dataclass
 
+from transformer_engine.pytorch.tensor.float8_blockwise_tensor import Float8BlockwiseQTensor
+from transformer_engine.pytorch.tensor.float8_tensor import Float8Tensor
+from transformer_engine.pytorch.tensor.mxfp8_tensor import MXFP8Tensor
+from transformer_engine.pytorch.tensor.quantized_tensor import QuantizedTensor
```
ptrendx, line 16: Actually you can't do it that way - this recipe class is in common, so is used in the JAX version as well. You will need to do the import and setting in the `__post_init__` method as well (with try/except block).

negvet: Right, fixed in https://github.com/NVIDIA/TransformerEngine/pull/1724/commits/d6ea98167c39da0185aa89d2a98b080a1a26dfdb

transformer_engine/common/recipe/__init__.py
```diff
@@ -174,17 +174,29 @@ def scaling_factor_compute(amax: Tensor,
     reduce_amax: bool = True
     fp8_dpa: bool = False
     fp8_mha: bool = False
+    expected_tensor_class: Optional[Type] = None
 
     def __post_init__(self) -> None:
         assert self.fp8_format != Format.E5M2, "Pure E5M2 training is not supported."
+        try:
+            from transformer_engine.pytorch.tensor.float8_tensor import Float8Tensor
```
ksivaman, line 182: Not a fan of adding framework specific logic in the core lib at all, even if it is guarded with try-except. So far we haven't had to do this and probably shouldn't introduce it here unless a strong reason exists to do so.

How about we reverse the logic and instead of including tensor info in the recipe, we include some recipe info in the tensor, and that should achieve the same goal of being able to equate the match the two together, right?

transformer_engine/pytorch/module/base.py
```diff
@@ -1341,6 +1365,38 @@ def _validate_name(self):
...
+            return
+        if not hasattr(self, "weight_names") or not self.weight_names:
+            return
+
+        recipe = self.fp8_meta["recipe"]
+        weight_tensors = [getattr(self, name) for name in self.weight_names]
+        for i, tensor in enumerate(weight_tensors):
+            if isinstance(tensor, QuantizedTensorBase) and not isinstance(
+                tensor, recipe.expected_tensor_class
+            ):
```
ksivaman, line 1392:
```suggestion
            if isinstance(tensor, QuantizedTensorBase) and not tensor.recipe() == recipe.__class__.__name__:
```

ksivaman: Something like this to address my above comment maybe? This would require some hardcoding, i.e. each of the tensor types would be associated with a fixed "string" as recipe class name but I think that should be ok? In my example `MXFP8Tensor.recipe_name()` would return `MXFP8BlockScaling`. We could think of better ideas.

@negvet @ptrendx

ksivaman (changes requested):
https://github.com/NVIDIA/TransformerEngine/pull/1724/files#r2090393437

transformer_engine/pytorch/module/base.py
negvet: This is a good approach, although string-based comparison is not that robust (what if name is changed, inheritance is not naturally supported etc.). 

What about the following update of your proposal:
```
# In each tensor class (e.g., Float8Tensor)
@classmethod
def get_compatible_recipe_types(cls):
    return [DelayedScaling, Float8CurrentScaling]
```
and then we just check `isinstance(recipe, tuple(tensor.__class__.get_compatible_recipe_types()))`

negvet: But this is still a workaround, fitted for this particular case.
Another one could be a static `recipe_to_tensor_map` defined in the pytorch module.
Although if there is a stronger demand for a framework-specific logic in the recipe, we might want to consider to define abstract base recipe classes in the common module and move the framework-specific implementations to their respective modules (`class DelayedScaling(BaseDelayedScaling)`). But not sure if we need that.

ksivaman: The framework specific logic lives in the quantizers, the recipe itself makes sense in common.

I like the `get_compatible_recipe_types` idea. Some early nitpicks to help review later:

```
@classmethod
def get_compatible_recipes(cls):
    return (DelayedScaling, Float8CurrentScaling)
```

and we can check `isinstance(recipe, tensor.get_compatible_recipes())`

negvet:
As discussed with @ptrendx, allowed quantizer to know about the recipe, rather than a tensor.
See https://github.com/NVIDIA/TransformerEngine/pull/1724/commits/a2910d7a34a668b9fd42ceb5bbac90080a635083

ksivaman (approved):
LGTM!

PR #1728, opened 2025-04-29
[JAX] Decouple Recipe and ScalingMode

jberchtold-nvidia:
# Description

Currently the recipe and scaling mode are coupled in the TE/JAX extension. This is okay for the current recipes, such as delayed scaling, current scaling, and MXFP8 block scaling, as there is only a single scaling mode used per recipe. However, for the DeepSeek recipe this assumption no longer holds. For the DeepSeek recipe we will need 1x128 1D block scaling for inputs and 128x128 2D block scaling for weights. As a result, we need to decouple the two concepts of recipe and scaling mode.

This PR only decouples the recipe and scaling mode, it does not implement the DeepSeek recipe.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Refactor `QuantizeConfig` in `helper.py` to be have a single static object instead of a class with independent static fields. This allows for proper inheritance and abstract methods on the base class to allow for custom logic per recipe.
- Decouple ScalingMode and recipe by enabling the QuantizeConfig to define a function `(UsageType) -> ScalingMode` where usage type is an enum, `X`, `KERNEL`, `DGRAD`. This allows for recipes to use different scaling modes depending on the tensor.
- Update quantizer factory to utilize this decoupling

# Checklist:


transformer_engine/jax/quantize/helper.py
```diff
@@ -154,6 +160,142 @@ def _format2dtypes(format_: recipe.Format):
...
+    X = 0
+    KERNEL = 1
+    DGRAD = 2
+
+
+@dataclass
+class UsageContext:
+    """Context of where a particular quantizer will be used which is needed by some recipes."""
+
+    usage_type: UsageType
```
phu0ngng, line 175: Hi, why do we need a new class just to wrap around an enum?

transformer_engine/jax/quantize/helper.py
```diff
@@ -154,6 +160,142 @@ def _format2dtypes(format_: recipe.Format):
...
+
+
+@dataclass
+class QuantizerParams:
```
phu0ngng, line 179: I have a look at this `QuantizerParams` and on how it is used, and I would prefer not to have this class for the following reasons:
1. Whenever we need to query the `scaling_mode` or `q_dtype` or `q_layout` info, instead of 
```
QuantizeConfig.RECIPE_MANAGER.get_quantizer_params(UsageContext(UsageType.X)).scaling_mode
```
We could have simply done 
```
QuantizeConfig.RECIPE_MANAGER.get_scaling_mode(UsageType.X)
```
It is way simpler for other people to follow and make a contribution later.
2. With this `QuantizerParam`, we add one more level of object inside the Quantizer, which does not give any benefits. For the Quantizer create, we could do 
```
q_x = QuantizerFactory.create(RECIPE_MANAGER.get_scaling_mode(UsageType.X),
                                                    RECIPE_MANAGER.get_quantize_dtype(UsageType.X),
                                                    RECIPE_MANAGER.get_quantize_layout(UsageType.X),
                                                    **args_x)
 ```

File: transformer_engine/jax/quantize/helper.py [hunk 3]
```diff
@@ -241,11 +362,15 @@ def initialize(cls, fp8_recipe: recipe.Recipe) -> None:
...
+        recipe_manager = RecipeManagerFactory.create_recipe_manager(fp8_recipe)
+        fp8_available, reason_for_no_fp8 = recipe_manager.is_supported()
+        assert fp8_available, reason_for_no_fp8
+
         cls.INITIALIZED = True
         cls.MARGIN = fp8_recipe.margin if "margin" in dir(fp8_recipe) else 0.0
         cls.FP8_FORMAT = fp8_recipe.fp8_format
         cls.FWD_DTYPE, cls.BWD_DTYPE = _format2dtypes(cls.FP8_FORMAT)
-        cls.SCALING_MODE = _get_scaling_mode(fp8_recipe)
+        cls.RECIPE_MANAGER = recipe_manager
```
phu0ngng, line 373: Why don't we merge `QuantizeConfig` and `RecipeManager` into a single class?
I don't see a clear need for them to exist separately.

transformer_engine/jax/flax/module.py
```diff
@@ -363,14 +370,14 @@ def generate_quantize_meta(quantizer_name: str):
...
+                (get_quantize_config().AMAX_HISTORY_LEN,),
                 jnp.float32,
             ).value
             return QuantizeMeta(scale=scale, amax_history=amax_history)
 
-        if QuantizeConfig.SCALING_MODE == ScalingMode.DELAYED_TENSOR_SCALING or isinstance(
-            fp8_recipe, recipe.DelayedScaling
-        ):
+        if get_quantize_config().get_scaling_mode(
+            UsageType.X
```
jberchtold-nvidia: This is fine for now as delayedscaling is the only recipe that uses any Flax variable state and the recipe has the same scaling mode for x, kernel, and grad.

However, in future if we add more recipes with state we can refactor this to do both:
a) Generate quantizer meta variables on a per-tensor basis, so each can decide to store metadata independently
b) Make a more generic way of specifying a scaling mode needs metadata, our current system is hardcoded to delayed scaling in a few places

transformer_engine/jax/quantize/quantizer.py
```diff
@@ -912,27 +922,44 @@ def create_set(
...
-        bwd_dtype = bwd_dtype or QuantizeConfig.BWD_DTYPE
+            x_scaling_mode = get_quantize_config().get_scaling_mode(UsageType.X)
+            kernel_scaling_mode = get_quantize_config().get_scaling_mode(UsageType.KERNEL)
+            grad_scaling_mode = get_quantize_config().get_scaling_mode(UsageType.DGRAD)
+
+        fwd_dtype = fwd_dtype or get_quantize_config().FWD_DTYPE
+        bwd_dtype = bwd_dtype or get_quantize_config().BWD_DTYPE
         if is_2x2x is None:
-            if scaling_mode.is_1d_block_scaling():
+            # TODO(Jeremy): check x, kernel, grad separately for 2x
```
jberchtold-nvidia, line 941: This assumes 2x is the same for x, kernel, grad. To avoid more changes in this PR, I've made this assumption. But I can make this more generic to handle 2x for each tensor type x, kernel, or grad, independently.

transformer_engine/jax/quantize/helper.py
jberchtold-nvidia: Good point, done in latest version

transformer_engine/jax/quantize/helper.py
jberchtold-nvidia: Good point, I've remove it from this PR and just used the scaling mode. My original approach was too generalized without it being necessary.

transformer_engine/jax/quantize/helper.py
jberchtold-nvidia: Removed in latest version. This was because I had thought we may add more context in the future, but I agree it's better to avoid this wrapper until it is actually needed

transformer_engine/jax/quantize/quantizer.py
phu0ngng: Sounds good.

phu0ngng (dismissed):
LGTM!

PR #1735, opened 2025-04-29
[JAX] Fix failing L2 JAX unit tests

jberchtold-nvidia:
# Description

Fixes two issues with JAX L2 unit tests: one with quantization flattening axes and one with test_helper incorrectly asserting MXFP8 context when fp8_autocast enabled=False.
Both of these were issues with the L2 tests themselves not being updated to match the latest code so only the tests needed to be updated.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Fused quantize tests were failing due to the quantization flattening axis being -2, but one of the L2 shapes was left at 2 dimensions. This test is invalid as there are no axes to the left of -2 to flatten. The test is now skipped.
- Scales for tensor scaling were being compared in fp32. For certain unfused operations, we are not expecting scales to be bitwise identical between TE/JAX and pure-JAX because the unfused TE operations, e.g. for current scaling, store the intermediate result in the input dtype, eg. bf16. This caused some small mismatches in the L2 test set. I've updated this to assert the scales in the tolerance of the input dtype instead.
- For the MXFP8 context in `test_helper.py`, the test asserted that the MXFP8 context was set when `with fp8_autocast(enabled=False, ...)`. This test has been updated to assert the quantize config is unchanged.

# Checklist:


tests/jax/test_helper.py
```diff
@@ -73,8 +73,7 @@ def test_fp8_autocast_delayed_scaling(self):
         self._check_defult_state()
 
         with fp8_autocast(enabled=False, fp8_recipe=DelayedScaling()):
-            self.assertFalse(QuantizeConfig.is_fp8_enabled())
-            self._compare_delay_scaling(get_delayed_scaling(), DelayedScaling())
+            self._check_defult_state()
```
phu0ngng, line 76: Typo

tests/jax/test_custom_call_compute.py
```diff
@@ -615,6 +620,12 @@ def test_quantize_dbias(
         ):
             pytest.skip(f"Input shape {input_shape} is not supported by MXFP8")
 
+        if (flatten_axis < 0 and flatten_axis + len(input_shape) <= 0) or flatten_axis <= 0:
+            pytest.skip(
+                f"Flatten axis {flatten_axis} is not supported for input shape {input_shape}. There"
+                " must be at least one axis on either side of the flatten_axis split."
+            )
+
```
phu0ngng, line 628: We should not skip this test, but rather ensure that all the test cases are working.
How about adding this fix?
```
        if flatten_axis == -2:
            input_shape = input_shape[:-1] + (2,) + input_shape[-1:]
```
Besides, all the shapes we test should work for MXFP8 already. Could you help to remove the skipif not good shape for MXFP8 above as well?

jberchtold-nvidia: > We should not skip this test, but rather ensure that all the test cases are working.

This test case can't work cause the shape is invalid for flatten_axis -2. I think it's a bit confusing to change the shape inside the test cause the test name will include the original shape, so we'll have a test named `..._shape-32-64_flatten-axis--2...`. But it's really not going to be testing (32, 64) with flatten_axis -2, it'll be testing (32, 2, 64).

What if we stored pairs of shape + flatten_axis and used those instead and listed valid combinations? So we still cover the (32, 2, 64) case but it's more explicit in the test names and listed shapes about what is running.

Like 
```
QUANTIZE_TEST_SHAPES_AND_FLATTEN_AXES = {
  "L2": [
       ((32, 64), -1),
       ((32, 2, 64), -2),
    ....
```

phu0ngng: Great idea! I think we could update other quantize tests with this new parameterized way as well.

jberchtold-nvidia: Updated

tests/jax/test_helper.py
jberchtold-nvidia: Fixed

phu0ngng (approved):
LGTM

PR #1742, opened 2025-05-01
[PyTorch] Debug UB conversion function from NVTEShape to std::vector

timmoon10:
# Description

https://github.com/NVIDIA/TransformerEngine/pull/1708 introduced a bug in a Userbuffers convenience macro that converts an `NVTEShape` to an `std::vector`. That PR changed `NVTEShape` to hold data internally, so each instance will have a different data pointer. If a function invocation is passed into the `AS_VECTOR` macro, the function will be called multiple times and each returned value will have different data pointers.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Debug UB conversion function from NVTEShape to std::vector

# Checklist:

### DIFF
--- a/transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
+++ b/transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
@@ -21,11 +21,17 @@
 #define HALF_BYTES 2
 #define UB_MAX_SM 32
 
-#define AS_VECTOR(shape) std::vector<size_t>(shape.data, shape.data + shape.ndim)
-
 using namespace std::placeholders;
 
 namespace transformer_engine {
+
+namespace {
+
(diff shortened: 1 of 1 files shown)


denera (approved):
LGTM

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -21,12 +21,18 @@
 #define HALF_BYTES 2
 #define UB_MAX_SM 32
 
-#define AS_VECTOR(shape) std::vector<size_t>(shape.data, shape.data + shape.ndim)
```
timmoon10, line 24: If we pass a function invocation here, this macro resolves to:
```
std::vector<size_t>(shape().data, shape().data + shape().ndim);
```
Each of those function returns will have a different location in the stack, so the constructor will do something weird. I was seeing errors from 18-D tensors.

PR #1744, opened 2025-05-02
[C][PyTorch] Move multi tensors kernels from PyTorch extensions to core

ksivaman:
# Description

Follow up PR in an effort to make framework extensions pure CPP.

## Type of change

- [x] Infra/Build change
- [x] Code refactoring

## Changes

- Move fp8 block scaling related kernels to core lib.
- Remove support for double precision inputs.

# Checklist:


transformer_engine/pytorch/csrc/common.cpp
```diff
@@ -97,6 +97,45 @@ transformer_engine::TensorWrapper makeTransformerEngineTensor(at::Tensor tensor)
...
+  for (const auto& at_list : at_tensor_lists) {
+    NVTE_CHECK(at_list.size() == num_tensors, "Wrong number of tensors");
+    std::vector<NVTETensor> te_list;
+    te_list.reserve(num_tensors);
+
+    for (const auto& at_tensor : at_list) {
+      transformer_engine::TensorWrapper tensorWrapper = makeTransformerEngineTensor(at_tensor);
+      NVTETensor te_tensor = tensorWrapper.data();
+      tensorWrappers.push_back(std::move(tensorWrapper));
+      te_list.push_back(std::move(te_tensor));
```
timmoon10, line 125: Move operators don't seem right here since they can leave the original variable in an undefined state. If I were reading this for the first time, I would be uncertain if `te_tensor` is invalid after `tensorWrapper` has been moved. In practice it's fine since `NVTETensor` is a POD, so it will actually copy instead of move, but it's not clean.
```suggestion
      tensorWrappers.push_back(makeTransformerEngineTensor(at_tensor));
      te_list.push_back(tensorWrappers.back().data());
```

transformer_engine/common/multi_tensor/multi_tensor_apply.cuh
```diff
@@ -46,57 +91,32 @@ __global__ void multi_tensor_apply_kernel(int64_t chunk_size, volatile int *noop
...
 
 template <int depth, bool USE_FP8 = false, typename T, typename... ArgTypes>
-void multi_tensor_apply(int64_t block_size, int64_t chunk_size, const at::Tensor &noop_flag,
-                        const std::vector<std::vector<at::Tensor>> &tensor_lists, T callable,
-                        ArgTypes... args) {
+void multi_tensor_apply(int64_t block_size, int64_t chunk_size,
+                        const transformer_engine::Tensor &noop_flag,
+                        transformer_engine::Tensor **tensor_lists, const size_t num_tensor_lists,
+                        const size_t num_tensors_per_list, T callable, const int device_id,
+                        cudaStream_t stream, ArgTypes... args) {
```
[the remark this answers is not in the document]
timmoon10: We still have `device_id` in the C API.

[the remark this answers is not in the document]
timmoon10: The current plan is to mark these APIs as experimental and `device_id` in particular as deprecated. Later, we will have a standalone PR that removes this feature, as well as corresponding tests in `test_fused_optimizer.py`.

timmoon10 (commented):
Mostly LGTM, although I see some C++ test failures.

transformer_engine/common/multi_tensor/multi_tensor_apply.cuh
```diff
@@ -46,61 +91,39 @@ __global__ void multi_tensor_apply_kernel(int64_t chunk_size, volatile int *noop
...
-  TORCH_CHECK(ref_device.type() == at::kCUDA, "expected input to be on cuda");
-  for (int l = 0; l < depth; l++) {  // No range-based for because I need indices
-    TORCH_CHECK(tensor_lists[l].size() == len0, "Size mismatch among tensor lists");
-    for (int t = 0; t < tensor_lists[l].size(); t++) {
-      // TODO:  Print which tensor fails.
-      bool contiguous_memory = tensor_lists[l][t].is_contiguous();
-      contiguous_memory =
-          (contiguous_memory || tensor_lists[l][t].is_contiguous(at::MemoryFormat::ChannelsLast) ||
-           tensor_lists[l][t].is_contiguous(at::MemoryFormat::ChannelsLast3d));
-      TORCH_CHECK(contiguous_memory, "A tensor was not contiguous.");
```
MaxwellDing, line 71: does This means that the kernel now supports computation on non-contiguous tensors?

ksivaman: The tensors must still be contiguous. We have moved multi tensor kernels from PyTorch extensions to our core lib and we have C APIs for these as well, which only work with contiguous tensors. However, the raw tensor data that we use in our core library does not allow us to check for contiguity, so we check for contiguity on the frameworks side instead. If the user chooses to use C API directly they must ensure contiguous tensors are passed!

MaxwellDing: > The tensors must still be contiguous. We have moved multi tensor kernels from PyTorch extensions to our core lib and we have C APIs for these as well, which only work with contiguous tensors. However, the raw tensor data that we use in our core library does not allow us to check for contiguity, so we check for contiguity on the frameworks side instead. If the user chooses to use C API directly they must ensure contiguous tensors are passed!

Appreciate your reply! however, I have not found contiguity check on the pytorch frameworks side. Regarding multitensor, the only code line I could find that involves continuity is commented out. Will this cause an issue?
https://github.com/NVIDIA/TransformerEngine/blob/097afc00d72800ca7328ae1ff8a0d84399b51880/transformer_engine/pytorch/csrc/common.cpp#L70

PR #1759, opened 2025-05-08
Cache torch.Tensor() to reduce CPU overhead

BestJuly:
# Description
Cache torch.Tensor() to reduce CPU overhead in
- `clear_tensor_data()`
- `clear()`

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py
- transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
- transformer_engine/pytorch/tensor/_internal/mxfp8_tensor_base.py
- transformer_engine/pytorch/tensor/utils.py
- transformer_engine/pytorch/utils.py

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/tensor/utils.py
+++ b/transformer_engine/pytorch/tensor/utils.py
@@ -16,12 +16,6 @@ from .float8_blockwise_tensor import Float8BlockwiseQTensor, Float8BlockQuantize
 from ..optimizers.multi_tensor_apply import multi_tensor_applier
 
 
-@functools.lru_cache(maxsize=None)
-def _empty_tensor() -> torch.Tensor:
-    """Get tensor with no entries and no data"""
-    return torch.Tensor()
-
-
 def replace_raw_data(tensor: QuantizedTensor, new_raw_data: torch.Tensor):
     r"""Change a quantized tensor's data buffer while preserving values
 
--- a/transformer_engine/pytorch/tensor/utils.py
+++ b/transformer_engine/pytorch/tensor/utils.py
@@ -5,7 +5,7 @@
 """Helper functions for using fp8 tensors as weights"""
 
 import torch
-
+import functools
 import transformer_engine_torch as tex
 from transformer_engine_torch import multi_tensor_scale, multi_tensor_compute_scale_and_scale_inv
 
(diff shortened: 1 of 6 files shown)


transformer_engine/pytorch/tensor/utils.py
```diff
@@ -16,6 +16,12 @@
 from ..optimizers.multi_tensor_apply import multi_tensor_applier
 
 
+@functools.lru_cache(maxsize=None)
+def _empty_tensor() -> torch.Tensor:
```
ksivaman, line 20: Why the redefinition? Let's reuse the one from `transformer_engine/pytorch/utils.py`

ksivaman:
Makes sense! Could you also modify `transformer_engine/pytorch/cpp_extensions/gemm.py` to also use the `_empty_tensor` from `utils.py`?

BestJuly:
> Makes sense! Could you also modify `transformer_engine/pytorch/cpp_extensions/gemm.py` to also use the `_empty_tensor` from `utils.py`?
I also noticed this, but in the `gemm.py`, the function will return `torch.Tensor().cuda()` while in `clear()` part it uses `torch.Tensor()`. Should we all use `torch.Tensor().cuda()`, or use an argument to decide which should be returned. I am a little bit confused about the different usages here so I kept it in the initial commit.

ksivaman:
It should be safe to just return the `.cuda()` version of tensor for everything.

BestJuly:
In grouped linear, in CPU side it will call `num_expert * (aten::empty, aten::_has_compatible_shallow_copy_type)` and this will help to reduce the `aten::empty` process. The `aten::_has_compatible_shallow_copy_type` is introduced in `t.data = _empty_tensor()` and will still exist.
image

transformer_engine/pytorch/tensor/utils.py
BestJuly: Have removed duplicated `_empty_tensor()` in both `tensor/utils.py` and `gemm.py`.

transformer_engine/pytorch/tensor/utils.py
```diff
@@ -5,7 +5,7 @@
 """Helper functions for using fp8 tensors as weights"""
 
 import torch
-
+import functools
```
ksivaman, line 8:
```suggestion

```

ksivaman: Unused import

ksivaman (approved):
LGTM

PR #1762, opened 2025-05-08
[PyTorch] Activation offloading refactor

pggPL:
# Description

Activation offloading refactor. 

Changes:

1. Enables offloading all activations for all layers - not only these supported by the Transformer Engine.
2. Enables offloading for MXFP8 and Blockwise scaling.
3. Exposes flexible offloading scheduling - please see example in tutorial - https://github.com/NVIDIA/TransformerEngine/blob/713cec8b22134fbe74c0b9cdac89df7a50640d62/examples/pytorch/cpu_offloading/example.ipynb
4. Adds tutorial mentioned above.
5. Fixes bunch of bugs like duplication of offloaded tensors - it should be more effective.
6. Provides new default offloading synchronization policy, that it should be easier to use and more effective.
7. Adds much more tests, which will make this feature more stable.
8. Offloading starts as fast as possible - not after end of the layer.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:


transformer_engine/pytorch/attention/multi_head_attention.py
```diff
@@ -854,6 +854,11 @@ def forward(
         # Core attention computation
         # ===========================
 
+        from transformer_engine.pytorch.cpu_offload import start_offload_if_offload_enabled
+
+        """start_offload_if_offload_enabled(
+            query_layer, key_layer, value_layer, offload_base_tensor=True
+        )"""
```
ptrendx: Commented out code and unused import.

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -446,15 +443,6 @@ def forward(
...
             ctx.quantized_weight = quantized_weight
-            if fuse_wgrad_accumulation and weight.requires_grad:
-                # This check is needed to ensure that main_grad is not created
-                # during the forward pass when using MCore FSDP as it creates
-                # the main_grad buffer lazily before backprop
-                if hasattr(weight, "__fsdp_param__"):
-                    # MCore FSDP creates main_grad lazily before backward
-                    ctx.main_grad_func = weight.get_main_grad
-                else:
-                    ctx.main_grad_func = lambda: weight.main_grad
```
ptrendx, line 470: This does not seem to be CPU offloading related, why are we removing it?

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -529,17 +517,11 @@ def backward(
...
             # Delete the references to tensor objects once they've been consumed
             # by the `restore_from_saved` method to construct back the actual tensors.
             ctx.tensor_objects = None
 
-            # Since main_grad can be modified inplace, it should not be a part of saved_tensors
-            main_grad = (
-                ctx.main_grad_func()
-                if weight is not None and ctx.fuse_wgrad_accumulation and ctx.requires_wgrad
-                else None
-            )
```
ptrendx, line 554: Same.

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -551,22 +560,6 @@ def forward(
...
-                    # MCore FSDP creates main_grad lazily before backward
-                    ctx.fc1_main_grad_func = (
-                        fc1_weight.get_main_grad if fc1_weight.requires_grad else lambda: None
-                    )
-                    ctx.fc2_main_grad_func = (
-                        fc2_weight.get_main_grad if fc2_weight.requires_grad else lambda: None
-                    )
-                else:
-                    ctx.fc1_main_grad_func = lambda: fc1_weight.main_grad
-                    ctx.fc2_main_grad_func = lambda: fc2_weight.main_grad
```
ptrendx, line 611: Why are we removing this?

transformer_engine/pytorch/module/linear.py
```diff
@@ -396,18 +385,8 @@ def forward(
...
             ctx.fuse_wgrad_accumulation = fuse_wgrad_accumulation
-            if fuse_wgrad_accumulation and weight.requires_grad:
-                # This check is needed to ensure that main_grad is not created
-                # during the forward pass when using MCore FSDP as it creates
-                # the main_grad buffer lazily before backprop
-                if hasattr(weight, "__fsdp_param__"):
-                    # MCore FSDP creates main_grad lazily before backward
-                    ctx.main_grad_func = weight.get_main_grad
-                else:
-                    ctx.main_grad_func = lambda: weight.main_grad
```
ptrendx, line 446: Same, why are we removing this?

transformer_engine/pytorch/module/linear.py
```diff
@@ -458,19 +437,6 @@ def backward(ctx, grad_output: torch.Tensor) -> Tuple[Union[torch.Tensor, None],
             # by the `restore_from_saved` method to construct back the actual tensors.
             ctx.tensor_objects = None
 
-            # Since main_grad can be modified inplace, it should not be a part of saved_tensors
-            main_grad = (
-                ctx.main_grad_func()
-                if weight is not None and ctx.fuse_wgrad_accumulation and ctx.requires_wgrad
-                else None
-            )
```
ptrendx, line 506: This probably should be left?

transformer_engine/pytorch/optimizers/fused_adam.py
```diff
@@ -374,7 +374,7 @@ def _initialize_state(
         if store_param_remainders:
             data = torch.zeros_like(param, dtype=torch.int16)
         else:
-            data = torch.empty_like(param, dtype=dtype)
+            data = torch.empty_like(param.dequantize(), dtype=dtype)
```
ptrendx: This seems wrong. `param` could be a regular Tensor I believe, also `empty_like` should work as is, without explicit call to dequantize.

transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py
```diff
@@ -124,9 +124,17 @@ def restore_from_saved(
         self._columnwise_scale_inv = tensors[3]
         return tensors[4:]
 
-    def get_data_tensors(self):
+    def get_data_tensors(self, scales: bool = False):
         """Get this Tensor's data."""
-        return self._rowwise_data, self._columnwise_data
+        if scales:
```
ptrendx: Why do you need this special case? Is it because you want to offload only the data and not the scale tensors?

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -397,6 +397,42 @@ def clone(self) -> Float8BlockwiseQTensor:
...
+            dtype=self.dtype,
+            fp8_dtype=self._fp8_dtype,
+            rowwise_data=new_rowwise_data,
+            rowwise_scale_inv=new_rowwise_scale_inv,
+            columnwise_data=new_columnwise_data,
+            columnwise_scale_inv=new_columnwise_scale_inv,
+            quantizer=self._quantizer,
+            is_2D_scaled=self._is_2D_scaled,
+            requires_grad=self.requires_grad,
+        )
```
ptrendx: Shouldn't this already be possible with `self._quantizer.make_empty(self.shape, dtype=self.dtype)`?

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -262,6 +262,39 @@ def clone(self) -> MXFP8Tensor:
...
+            if self._rowwise_data is not None
+            else None
+        )
+        new_columnwise_data = (
+            torch.empty_like(self._columnwise_data, *args, **kwargs)
+            if self._columnwise_data is not None
+            else None
+        )
+        new_rowwise_scale_inv = (
+            torch.empty_like(self._rowwise_scale_inv, *args, **kwargs)
```
ptrendx: Same as in the blockwise, but here especially so because in MXFP8 case the scale inverses have to be zero-initialized.

transformer_engine/pytorch/module/layernorm_linear.py
pggPL: At some point offloading logic supported weight offloading(it is deprecated because it didn't work properly and it is not used anywhere). The problem is that when tensor is offloaded, the .empty_like() don't copy .main_grad. Thus it is saved for backward and connected with weight tensors in backward. This created more issues, the code above aims to solve.

Since we do not try to offload weights anymore, we can remove all these workarounds.

transformer_engine/pytorch/module/layernorm_linear.py
pggPL: answered above

transformer_engine/pytorch/module/layernorm_mlp.py
pggPL: answered above

transformer_engine/pytorch/module/linear.py
pggPL: answered above

transformer_engine/pytorch/module/linear.py
pggPL: answered above

transformer_engine/pytorch/optimizers/fused_adam.py
pggPL: For regular tensor dequantize changes dtype (and memory format but I guess this is not issue here), which is changed anyways. I needed to change that in case param was QuantizedTensor.

transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py
pggPL: No, I want to prevent offloading the scales of weights. In .prepare_for_saving() scales are returned and we save them.  I want to mark them with .is_weight to prevent offload, so I need to have some possibility of getting them.

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
pggPL: It is - it's not that simple unfortunately - the quantizer may have different usages than tensor. But with some small changes it seems that everything works correctly.

niyunsheng:
Excellent work. Could you also add support for non-reentrant mode?

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1131,14 +1130,6 @@ def forward(
                     inference_params=inference_params,
                 )
 
-            from transformer_engine.pytorch.cpu_offload import CPUOffloadEnabled
```
pggPL, line 1497: I removed this warning, because it is not True and we do not want to expose old code path to new users and I think all current users make sure they do not use unfused attention.

tests/pytorch/test_cpu_offloading_old_code_path.py
timmoon10, line 1: We should update this test with the changes in https://github.com/NVIDIA/TransformerEngine/pull/2158. The old test didn't make the input with `requires_grad=True`, so the memory estimates were incorrect. It turns out memory usage is not the same between the 1 layer case and the CPU offloading case because the input tensor caching is different between those two cases.

timmoon10: Also, `_old_code_path` is a pretty bad name. Maybe `test_cpu_offloading_v1.py` and `test_cpu_offloading_v2.py`?

transformer_engine/pytorch/optimizers/fused_adam.py
timmoon10: Biases and norm params are in `torch.Tensor`, which don't have a `dequantize` function.

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -59,6 +59,12 @@ def update_usage(
             f"{self.__class__.__name__} class does not implement update_usage function"
         )
 
+    def get_usage(self) -> Tuple[bool, bool]:
+        """Get the usage of the tensor"""
+        raise NotImplementedError(
+            f"{self.__class__.__name__} class does not implement get_usage function"
+        )
+
```
timmoon10: We shouldn't lock ourselves into only supporting row-wise and column-wise usages. We may need to support usages for communication, attention, convolution, and who knows what else:
```suggestion
```

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -423,6 +446,18 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
         if func == torch.ops.aten.copy_.default:
             dst = args[0]
             src = args[1]
+            if isinstance(dst, QuantizedTensor) and isinstance(src, QuantizedTensor):
+                assert dst._quantizer == src._quantizer, \
+                    "Quantizers must be the same for copy_ of QuantizedTensor"
```
timmoon10: It's a perfectly valid use-case to copy between quantized tensors with different quantizers, or even with completely different data types. This case should be treated as an optimization, with a fallback to dequantize-quantize
```suggestion
            if (
                isinstance(dst, QuantizedTensor)
                and isinstance(src, QuantizedTensor)
                and dst._quantizer == src._quantizer
            ):
```

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -435,6 +470,32 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
         if func == torch.ops.aten.view.default:
             raise NotImplementedError("{cls.__name__} class does not support tensor views")
 
+        # Empty like op
+        if func == torch.ops.aten.empty_like.default:
+            tensor = args[0]
+            device = kwargs.get("device", tensor.device)
+            requires_grad = kwargs.get("requires_grad", tensor.requires_grad)
+            pin_memory = kwargs.get("pin_memory", False)
```
timmoon10, line 477: Does `__torch_dispatch__` put these options in `args` or `kwargs`?

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -435,6 +470,32 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+            tensor._quantizer.set_usage(rowwise=quantizer_rowwise_usage, columnwise=quantizer_columnwise_usage)
+            return out
+        
+        if func in (torch.ops.aten.numel.default, torch.ops.aten.is_pinned.default):
+            data_tensors = tensor.get_data_tensors()
+            for t in data_tensors:
+                if t is not None:
+                    return func(t)
+            raise ValueError("No data to get, both rowwise_data and columnwise_data are False")
+
```
timmoon10: There are a some correctness issues. For example, FP4 data tensors don't match the tensor shape. Also, if we are handling the numel op, do we also want to handle the tensor shape?
```suggestion
        if func == torch.ops.aten.numel.default:
            tensor = args[0]
            return math.prod(tensor.size())
            
        if func == torch.ops.aten.is_pinned.default:
            tensor = args[0]
            for t in tensor.get_data_tensors():
                if t is not None:
                    return func(t)
            return False  # Or error out?

```

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -341,7 +348,14 @@ class QuantizedTensor(torch.Tensor):
...
     """
 
-    def __new__(cls, shape: Iterable[int], dtype: torch.dtype, *, requires_grad: bool = False):
+    def __new__(
+        cls,
+        shape: Iterable[int],
+        dtype: torch.dtype,
+        *,
+        requires_grad: bool = False,
+        device: Optional[torch.device] = None,
```
timmoon10, line 354: If we are adding CPU support to quantized tensors, we should add standalone CPU tests, perhaps in https://github.com/NVIDIA/TransformerEngine/blob/main/tests/pytorch/test_float8tensor.py. In order for the fallback `__torch_dispatch__` to do the right thing, we should also consider adding logic in the dequantize function to check that the data is on GPU.

transformer_engine/pytorch/cpu_offload.py
```diff
@@ -3,685 +3,744 @@
...
+        old_code_path.mark_activation_offload(*tensors)
 
-class CpuOffloadSavedTensorHook:
-    """Contex-manager that executes a pair of pack/unpack hooks for saved tensors.
 
-    In this context, the ``on_save_for_backward`` method will be called every time
-    a tensor is saved for backward (this includes intermediary results saved using
-    :func:`~torch.autograd.function._ContextMethodMixin.save_for_backward` but
-    also those recorded by a PyTorch-defined operation).
+def mark_not_offload(*tensors: torch.Tensor):
```
timmoon10, line 45: From what I understand, the API for tensor offloading is weirdly between opt-in and opt-out:
- We automatically offload all tensors in the autograd `ctx`, so we call `mark_not_offload` to avoid offloading params.
- We want to overlap CPU offloading with compute, so we call `start_offload` to explicitly offload activations.

Would it be better to commit to an opt-in API, and no longer automatically offload tensors in the `ctx`? That would be better for avoiding some bugs (I see several cases where we unnecessarily offload biases). On the other hand, there might be some convenience in new operations automatically offloading everything, and skipping params might be considered an optimization. What do you think?

transformer_engine/pytorch/cpu_offload_old_path.py
timmoon10, line 1: `_old_path` is not a good name. Maybe `_v1` and `_v2` would be better.

timmoon10 (dismissed):

transformer_engine/pytorch/cpu_offload.py
pggPL: I consciously decided to use opt-out approach. It enables using offloading for every layer, not only from TE. We have some internal people using this MR in such use case. It's also makes using offloading much easier to use if one works one of abstraction layer above us and only composes something from TE and non TE modules.

transformer_engine/pytorch/optimizers/fused_adam.py
pggPL: All torch.Tensor instances have this function, nevertheless, I removed this change, because the tests seem to pass without it.

tests/pytorch/test_cpu_offloading_old_code_path.py
pggPL: We plan to remove old code path at the end of the year - this is needed as a fallback option for DL training folks, if something will break with new code path. Since mcore imports from transformer_engine.pytorch.cpu_offloading, I think that we should use this name for new path.

transformer_engine/pytorch/quantized_tensor.py
pggPL: In this case yes. I know that sometimes things like this are put in args, not kwargs (for copy_) for example, but it seems to be correct.

transformer_engine/pytorch/tensor/quantized_tensor.py
pggPL: I changed that into dictionary, what do you think?

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -59,7 +65,7 @@ def update_usage(
             f"{self.__class__.__name__} class does not implement update_usage function"
         )
 
-    def get_usage(self) -> Tuple[bool, bool]:
+    def get_usage(self) -> Dict[str, bool]:
```
timmoon10: Nit: A tensor may have multiple usages.
```suggestion
    def get_usages(self) -> Dict[str, bool]:
```

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -450,17 +457,21 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
             dst = args[0]
             src = args[1]
-            if isinstance(dst, QuantizedTensor) and isinstance(src, QuantizedTensor):
-                assert dst._quantizer == src._quantizer, \
-                    "Quantizers must be the same for copy_ of QuantizedTensor"
+            if (
+                isinstance(dst, QuantizedTensor)
+                and isinstance(src, QuantizedTensor)
+                and type(dst._quantizer) is type(src._quantizer)
+                and all(d == s for d, s in zip(dst.get_usage(), src.get_usage()))
```
timmoon10, line 440: Iterating over a dict just iterates over the keys, so this will pass trivially. Also, it would be a pain to guarantee that all tensors produce dicts with the same ordering. Maybe:
```suggestion
                and set(src.get_usage().keys()) == set(dst.get_usage().keys())
                and all(src.get_usage()[usage] == dst.get_usage()[usage] for usage in src.get_usage().keys())
```

transformer_engine/pytorch/tensor/storage/float8_tensor_storage.py
```diff
@@ -227,3 +227,10 @@ def update_usage(
         if not needs_data_transpose:
             self._transpose = None
             self._transpose_invalid = True
+
+    def get_usage(self) -> Dict[str, bool]:
+        """Get the usage of the tensor"""
+        return {
+            "rowwise": self._data is not None,
+            "columnwise": self._transpose is not None and not self._transpose_invalid,
+        }
```
timmoon10: We should think more carefully about this. On Hopper, row-wise usage corresponds to non-transposed data and column-wise usage corresponds to transposed. On Blackwell, both row-wise and column-wise usage use the same non-transposed data. Logically, this should be something like:
```suggestion
    def get_usage(self) -> Dict[str, bool]:
        """Get the usage of the tensor"""
        usages = {"rowwise": self._data is not None}
        if is_non_tn_fp8_gemm_supported():
            usages["columnwise"] = self._data is not None
        else:
            usages["columnwise"] = self._transpose is not None and not self._transpose_invalid
        return usages
```

tests/pytorch/test_cpu_offloading_old_code_path.py
timmoon10: Temporary code has a tendency to stick around, and I'd rather eliminate the possibility of names like `_old_old_code_path`. We can keep `test_cpu_offloading.py`, but let's change old test to `test_cpu_offloading_v1.py`. Also, as long as we are keeping the old test we should keep the changes from https://github.com/NVIDIA/TransformerEngine/pull/2158.

tests/pytorch/test_cpu_offloading.py
```diff
@@ -33,181 +47,642 @@
...
+            "linear",
+            "layernorm_linear",
+            "layernorm_mlp",
+            "grouped_linear",
+            "multihead_attention",
+            "transformer_layer",
+        ]
+
+    @staticmethod
+    def create_layer(layer_type: str):
```
timmoon10, line 112: We should include `te.ops.Linear` in this function.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -718,12 +725,12 @@ def forward(
                     use_flash_attn_3=use_flash_attn_3,
                 )
         else:
+
```
ptrendx:
```suggestion
```

transformer_engine/pytorch/attention/multi_head_attention.py
```diff
@@ -951,6 +951,10 @@ def forward(
         # Core attention computation
         # ===========================
 
+        from transformer_engine.pytorch.cpu_offload import start_offload, is_cpu_offload_enabled
```
ptrendx: Why do we import it here? This will result in small CPU overhead during every forward. We should be able to import it only once.

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -469,6 +483,7 @@ def forward(
                     ctx.main_grad_func = weight.get_main_grad
                 else:
                     ctx.main_grad_func = lambda: weight.main_grad
+            ctx.quantized_weight = quantized_weight
```
ptrendx: This is already done above the if statement.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -189,6 +189,9 @@ def forward(
                 for i in range(num_gemms):
                     weight_quantizers[i].calibrate(weights[i])
 
+        if cpu_offloading:
+            mark_not_offload(*weights_fp8, *weights)
```
ptrendx, line 203: Don't we also need the start_offloading piece?

transformer_engine/pytorch/cpu_offload.py
```diff
@@ -3,685 +3,746 @@
...
-    def __init__(self) -> None:
-        self.inside_context = False
+    for tensor in tensors:
+        if tensor is not None:
+            setattr(tensor, "_TE_do_not_offload", True)
 
-    def __enter__(self):
-        global CPUOffloadEnabled
-        CPUOffloadEnabled = True
+    restore_from_saved(tensor_obj, tensors)
```
ptrendx, line 56: This will be pretty expensive, since we are doing this every single forward pass. Can we instead do something like "add_property" to the tensor classes?

ptrendx: Or actually since you are doing a similar pattern in other places - maybe something like a map over the internal elements? As in, a function that accepts a lambda and then invokes it over all tensors inside the quantized tensor?

transformer_engine/pytorch/cpu_offload.py
```diff
@@ -3,685 +3,746 @@
...
-    def tensor_pop(self, tensor_tag: Any, **kwargs):
-        """Tensor pop."""
-        raise NotImplementedError(
-            "`tensor_pop is not implented in OffloadHandler class. "
-            "Inherit this class and implement your custom tensor_pop."
-        )
+    @staticmethod
+    def tensor_group_process_before_offload(tensor_group: TensorGroup) -> tuple[TensorGroup, Any]:
+        """
+        Public API call - for a tensor group, just before offloading logic happens.
```
ptrendx: Do we need to say that it is a public API call? That should be visible from the fact that it does not start from `_`.

transformer_engine/pytorch/cpu_offload.py
```diff
@@ -3,685 +3,746 @@
...
+
+        aux["views"] = []
+        for tensor_id in range(  # pylint: disable=consider-using-enumerate
+            len(tensor_group.tensor_list)
+        ):
+            tensor = tensor_group.tensor_list[tensor_id]
+            if _check_if_offload_base_tensor(tensor):
+                aux["views"].append((tensor.shape, tensor.stride(), tensor.storage_offset()))
+                tensor = tensor._base
+                assert tensor is not None
```
ptrendx: Some error message?

tests/pytorch/test_cpu_offloading.py
pggPL: done

transformer_engine/pytorch/tensor/quantized_tensor.py
pggPL: fixed

transformer_engine/pytorch/cpu_offload.py
pggPL: done

transformer_engine/pytorch/module/grouped_linear.py
pggPL: done

transformer_engine/pytorch/module/layernorm_linear.py
pggPL: fixed

transformer_engine/pytorch/cpu_offload.py
pggPL: We need to mark every data/scale tensor, because sometimes tensors are saved after the application of prepare_for_saving - then inside cpu offloading code I see only these internal tensors, not the quantized tensor object.

I think the only overhead is that we set attributes of quantized tensor to null and set them again in restore_from_saved. Is it that substantial?

transformer_engine/pytorch/quantized_tensor.py
pggPL: done

transformer_engine/pytorch/tensor/storage/float8_tensor_storage.py
pggPL: done

transformer_engine/pytorch/cpu_offload_old_path.py
pggPL: done

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1247,12 +1254,9 @@ def forward(
         # used when some tensors are base tensors and loose the "dtype" attribute
         ctx.nominal_dtype = out_nominal_dtype
 
-        from transformer_engine.pytorch.cpu_offload import (
-            CPUOffloadEnabled,
-            mark_activation_offload,
-        )
+        from transformer_engine.pytorch.cpu_offload import NVTE_CPU_OFFLOAD_V1
```
ptrendx: Do we have to do this import every time? Why can't we do it once at the beginning of the file?

pggPL: fixed

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1115,6 +1120,9 @@ def forward(
         nvtx_label = "transformer_engine.FusedAttnFunc.forward"
         nvtx_range_push(f"{nvtx_label}")
 
+        if is_cpu_offload_enabled():
+            start_offload(q, k, v, offload_base_tensor=True)
```
sanandaraj5597, line 1125: What does offload base tensor do?

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -407,6 +437,26 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
         if func == torch.ops.aten.copy_.default:
             dst = args[0]
             src = args[1]
+            if (
```
sanandaraj5597, line 440: Are you trying to create the entire Quantized tensor class on CPU so you don't need to rip off tensors from quant tensor objects?

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -288,11 +289,18 @@ def make_empty(
         scale_inv = None
         amax_rowwise = None
         if self.rowwise_usage:
-            data = torch.empty(self.convert_shape_for_fp4(shape), dtype=torch.uint8, device=device)
+            data = torch.empty(
```
sanandaraj5597, line 292: Do we support NVFP4 offloading with the new path?

transformer_engine/pytorch/cpu_offload.py
```diff
@@ -3,698 +3,748 @@
...
+        """
+        Changes tensors to base tensors and saves view options in aux.
+
+        It we save multiple tensors which in fact are views of the same base tensor,
+        this will save unnecessary calls to .contiguous(). It is used for example in
+        MultiHeadAttention for interleavedq, k, v tensors.
+        """
+
+        def _check_if_offload_base_tensor(tensor: torch.Tensor) -> bool:
+            if getattr(tensor, "offload_base_tensor", False):
```
sanandaraj5597, line 142: If a tensor has multiple views, the first offloaded tensor is made contiguous? Does this mean there's a performance penalty of doing a contiguous operation everytime we offload the first tensor view?

transformer_engine/pytorch/cpu_offload.py
```diff
@@ -3,698 +3,748 @@
...
-    The device-to-host and host-to-device copying happen in the same stream
-    as the computation kernels, thus the copying will block computation.
+    @staticmethod
+    def _switch_to_views(tensor_group: TensorGroup) -> TensorGroup:
+        """
+        Switch to views - reverse of _switch_to_base_tensors.
+        """
+        for tensor_id, tensor in enumerate(tensor_group.tensor_list):
+            if tensor_group.aux["views"][tensor_id] is not None:
+                tensor_group.tensor_list[tensor_id] = tensor.as_strided(
```
sanandaraj5597, line 217: Same question as above, does this create reshape operations on GPU?

transformer_engine/pytorch/cpu_offload.py
```diff
@@ -3,698 +3,748 @@
...
-        return tensor_tag
+            with torch.cuda.stream(self.offload_stream):
+                if allocate_cpu_buffers:
+                    # empty_like is defined also for QuantizedTensors
+                    offloaded_tensor = torch.empty_like(
+                        tensor, device=torch.device("cpu"), pin_memory=True
+                    )
+                    self.cpu_tensor_group.tensor_list.append(offloaded_tensor)
+                else:
+                    assert self.cpu_tensor_group.tensor_list[tensor_id].shape == tensor.shape, (
```
sanandaraj5597, line 284: Does this mean this retaining pinned CPU memory will work only if you have symmetric/same layers across the model?

transformer_engine/pytorch/cpu_offload.py
```diff
@@ -3,698 +3,748 @@
...
+                        "CPU buffer shape does not match the offloaded tensor shape:"
+                        f" {self.cpu_tensor_group.tensor_list[tensor_id].shape} != {tensor.shape}  "
+                        "                       Make sure that tensor shaped do not change between"
+                        " iterations if retain_pinned_cpu_buffers is True."
+                    )
+                    offloaded_tensor = self.cpu_tensor_group.tensor_list[tensor_id]
+                offloaded_tensor.copy_(tensor, non_blocking=True)
```
sanandaraj5597, line 291: This will call the copy of the quantized tensor overloaded copy you developed right? Does it incur any CPU overhead?

transformer_engine/pytorch/cpu_offload.py
```diff
@@ -3,698 +3,748 @@
...
+        for tensor in self.cpu_tensor_group.tensor_list:
+
+            # Notice that reloaded tensor is allocated on main stream,
+            # not offloaded stream. It is because PyTorch memory allocator
+            # cannot move tensors from pool of one stream to another without
+            # calling cudaFree and cudaMalloc again.
+
+            # empty_like is defined also for QuantizedTensors.
+            reloaded_tensor = torch.empty_like(tensor, device=torch.device("cuda"))
+            self.offload_stream.wait_stream(torch.cuda.current_stream())
```
sanandaraj5597, line 335: So the stream is shared for both offloading and reloading? Is it easy to have different streams? If at all there is a HW with bi-dir BW, it might be useful.

transformer_engine/pytorch/cpu_offload.py
```diff
@@ -3,698 +3,748 @@
...
+
+    def _init_offload_synchronization_dicts(self, num_offloaded_layers: int):
+        """
+        If synchronization dictionary is not provided, the number of offloaded layers is used to initialize
+        offload_layer_map, finish_offload_map and start_reload_map.
+
+        The aim is to minimize memory usage by the end of the forward pass.
+
+        The optimal strategy for that is to offload layers 0, ..., num_offloaded_layers - 1.
+        For layer i offload needs to finish before num_layers - num_offloaded_layers + i.
```
sanandaraj5597, line 554: The optimal strategy would be divide the forward execution time of num layers into num offload layer pieces and split it equally IMO. 
Maybe that's what you are doing?

sanandaraj5597:
Does this need any code in MCore?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
pggPL: If multiple tensors are views on one big tensor, this will offload this big tensor once.

transformer_engine/pytorch/tensor/nvfp4_tensor.py
pggPL: yes

transformer_engine/pytorch/cpu_offload.py
pggPL: No, .contiguous() is not called at all in offloading, comment above is inaccurate.

transformer_engine/pytorch/cpu_offload.py
pggPL: No, .as_strided( just sets the pointers.

transformer_engine/pytorch/cpu_offload.py
pggPL: No, it can be different, only tensors for particular layer needs to have constant shape. So having different layers in model is ok, but for example changing batch size or sequence dimension is not - the cpu buffers will not match tensors of different sizes.

transformer_engine/pytorch/cpu_offload.py
pggPL: Good point. If I understand correctly we would benefit from that only if we offload and reload at the same time. This is not the case for old code path and default offloading in new code path. I'm not sure if this will be used in PP aware offloading. If yes I may add this here, otherwise maybe in the other PR?

pggPL:
No, Mcore works with this code.

transformer_engine/pytorch/cpu_offload.py
ptrendx: We should then make the docstring more accurate, right?

transformer_engine/pytorch/cpu_offload.py
ptrendx: I think if we want to add it then it should land in another PR, this one is already very big and took a long time to merge :-).

transformer_engine/pytorch/cpu_offload.py
ptrendx: No. With the proposed approach we start the offloading as soon as possible and give it the most possible time. With the splitting equally offloading is delayed until a particular piece is executed, which could lead to performance issues. For example consider jitter effect -> let's say that in the typical case the offloading takes a little less than the time needed to run the forward pass. That additional time of forward time is "wasted" (no offloading happens at that time). Now let's say there is an iteration when the offloading takes longer (for whatever reason) - this delay could make the offloading exposed since we do not have a buffer. In the proposed scheme we would start the offloading much earlier and therefore have a buffer to cover that additional cost.

transformer_engine/pytorch/cpu_offload.py
sanandaraj5597: I agree.

transformer_engine/pytorch/cpu_offload.py
sanandaraj5597: Ah, I got it. Seems better, lets do this.

transformer_engine/pytorch/cpu_offload.py
pggPL: Yes, it will. The copy is fast - just runs copy for each of subtensors.

transformer_engine/pytorch/quantized_tensor.py
pggPL: Basically yes. I considered many scenarios and this design seems to work in most of them - ripping off destroys the object which can sometimes result in errors, although it has bigger cpu overhead.

sanandaraj5597:
LGTM. Can merge.

ptrendx (approved):
LGTM

PR #1768, opened 2025-05-11
Fix split_overlap_ag `aggregate=True` chunk offset calculation

guyueh1:
# Description

For `split_overlap_ag` when UB `aggregate=True`, fix the chunk offset calculation for GEMM.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


denera (dismissed):
LGTM, but I wonder how this has been passing our numerical correctness checks. If we don't include the `aggregate=True` option in the tests, we should definitely add it.

guyueh1:
> LGTM, but I wonder how this has been passing our numerical correctness checks. If we don't include the `aggregate=True` option in the tests, we should definitely add it.

Agreed. The aggregate=True is supported by the `tests/pytorch/distributed/test_comm_gemm_overlap.py`, but not on by default. I think the parameter combination sweep in CI did not include it?

denera:
Yes, it looks like it's not included in the CI sweep.

Could you add a new `test_aggregated_all_gather` in there to test `aggregate=True`? I don't believe this needs to be done for all precisions. Just BF16 should be fine.

guyueh1:
> Yes, it looks like it's not included in the CI sweep.
> 
> Could you add a new `test_aggregated_all_gather` case in there to test `aggregate=True`? I don't believe this needs to be done for all precisions. Just BF16 should be fine.

where is the CI suite defined?

guyueh1:
@denera or @timmoon10 
follow-up the last comment, where is the CI tests defined? I can add a test for aggregate=True

guyueh1:
Done adding unit tests for `aggregate=True`, @denera please review and start pipeline for this PR

guyueh1:
Fixed one error in `test_comm_gemm_overlap.py`; The `test_numerics.py` failure and the two L1 thunder failures seem irrelevant, are they safe to ignore? 
need to approve CI again @denera

ksivaman (approved):
Merging with approval from @denera

PR #1771, opened 2025-05-12
Set sequence_parallel before super().__init__() in norm modules

negvet:
# Description

The current implementation has a problematic ordering:

- In `__init__`, `super().__init__()` is called first
- Then `self.sequence_parallel` is set
- If `reset_parameters()` gets called during the parent's initialization, it will try to use `self.sequence_parallel` before it's been set.

This could lead to the `sequence_parallel` not being properly applied to the parameters during initialization.

Currently, this is fixed with (in `__init__`):
https://github.com/NVIDIA/TransformerEngine/blob/51cd441501e8e6dee18c00056f008e1b53b89ebd/transformer_engine/pytorch/module/layernorm.py#L107-L109
but this does not seem right

## Type of change

- [ ] Infra/Build change
- [X] Code refactoring

## Changes

Set sequence_parallel before super().__init__() in norm modules

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/layernorm.py
+++ b/transformer_engine/pytorch/module/layernorm.py
@@ -133,7 +133,7 @@ class LayerNorm(_LayerNormOp):
         super().reset_parameters()
 
         # Set flag for sequence parallelism (custom Megatron-LM integration)
-        if getattr(self, "sequence_parallel", None) is not None:
+        if self.sequence_parallel is not None:
             self.weight.sequence_parallel = self.sequence_parallel
             self.bias.sequence_parallel = self.sequence_parallel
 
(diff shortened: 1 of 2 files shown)


ptrendx:
If you always set sequence_parallel before `super.__init__` you could also change the logic in the module's `reset_parameters` to just use sequence_parallel rather than doing `getattr(self, "sequence_parallel", None)`.

negvet:
> If you always set sequence_parallel before `super.__init__` you could also change the logic in the module's `reset_parameters` to just use sequence_parallel rather than doing `getattr(self, "sequence_parallel", None)`.

good point, I missed that, updated https://github.com/NVIDIA/TransformerEngine/pull/1771/commits/0288e53a6c6e39839e7019fa9e625a79c9c03af5

ptrendx (approved):

PR #1772, opened 2025-05-12
[JAX] Support logical partitioning axes in TE Flax modules

jberchtold-nvidia:
# Description

TE/JAX Flax modules are using weight initialization that only supports physical mesh partitioning axes, not logical axes used in some frameworks, such as MaxText. To support partitioning in these frameworks weight initialization was updated to a newer way that supports logical axes.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Update Flax modules in `module.py` and `transformer.py` to replace usages of `nn_partitioning.param_with_axes` with `self.param(..., nn.with_logical_partitioning(...), ...)`
- Fix bug where ffn1 intermediate result in the LayerNormMLP layer incorrectly added a sharding constraint to replicate the tensor and following gemm, leading to OOM on some configs that normal MaxText did not have OOM on.

# Checklist:

### DIFF
--- a/transformer_engine/jax/sharding.py
+++ b/transformer_engine/jax/sharding.py
@@ -334,7 +334,7 @@ def get_non_contracting_logical_axes(ndim, logical_axes, contracting_dims):
     """
     if not logical_axes:
         return None
-    elif len(logical_axes) < ndim:
+    if len(logical_axes) < ndim:
         logical_axes = logical_axes + (None,) * (ndim - len(logical_axes))
     assert len(logical_axes) == ndim
 
--- a/transformer_engine/jax/sharding.py
+++ b/transformer_engine/jax/sharding.py
@@ -333,7 +333,7 @@ def get_non_contracting_logical_axes(ndim, logical_axes, contracting_dims):
         Tuple of logical axes for non-contracting dimensions.
     """
     if not logical_axes:
-        logical_axes = (None,) * ndim
+        return None
     elif len(logical_axes) < ndim:
         logical_axes = logical_axes + (None,) * (ndim - len(logical_axes))
     assert len(logical_axes) == ndim
--- a/transformer_engine/jax/sharding.py
+++ b/transformer_engine/jax/sharding.py
@@ -10,6 +10,7 @@ parallelism (FSDP). It includes functions for sharding constraints, mesh managem
 and collective operations.
 """
 import os
+from typing import Optional
 from contextlib import contextmanager
 from dataclasses import dataclass
 from enum import Enum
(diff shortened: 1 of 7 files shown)


transformer_engine/jax/sharding.py
```diff
@@ -333,8 +333,8 @@ def get_non_contracting_logical_axes(ndim, logical_axes, contracting_dims):
         Tuple of logical axes for non-contracting dimensions.
```
sudhakarsingh27, line 333: This should be updated I guess then.

transformer_engine/jax/sharding.py
```diff
@@ -333,8 +333,8 @@ def get_non_contracting_logical_axes(ndim, logical_axes, contracting_dims):
...
     """
     if not logical_axes:
-        logical_axes = (None,) * ndim
-    elif len(logical_axes) < ndim:
```
sudhakarsingh27, line 337: Could you also explain the rationale here?

jberchtold-nvidia: We have two concepts here, and both use `None` so it's not the clearest. 

1. `logical_axes = None`, this is the default value for many of our Flax modules. For example, if `input_axes = None` as the default arg value in the constructor of the module, it means to not apply a sharding constraint to the input. At least that is my best guess on it's intention as I don't think it'd make sense to default to forcing tensors to be replicated. This is also supported by the fact that `with_sharding_constraint_by_logical_axes` is implemented as a no-op if the axes are `None`, meaning the existing sharding will apply and no additional constraints are applied.
2. `logical_axes = (None, None, ..., None)`, adding a sharding constraint to the tensor to make it replicated instead of partitioned. A tuple of `None`s per axis in the tensor's shape. This is distinct and not a default value anywhere and it means "enforce the tensor is replicated".

The function that I'm updating here, `get_non_contracting_logical_axes`, had originally converted (1) to (2). This means that if this function were called and the logical axes inputs were the default value, `None`, we would get out `(None, None, ..., None)` from this function. This means we are changing the behavior from (1) passing through a tensor without applying sharding constraints, the default, to (2) enforcing the sharding is replicated. This is not the intended behavior of this function.

This led to `ffn1`, the intermediate value inside the LayerNormMLP layer that is sharded according to outputs of `get_non_contracting_logical_axes`, to be replicated and led to us getting OOM on model configs that didn't with TE disabled.

sudhakarsingh27: That's sneaky! Good catch

sudhakarsingh27: Also, should we include this description in the code?

transformer_engine/jax/sharding.py
jberchtold-nvidia: Switched this code to use an assert instead as none which better fit the use cases. So the doc comment is now correct

transformer_engine/jax/sharding.py
jberchtold-nvidia: Updated `with_sharding_constraint_by_logical_axes` documentation to clarify this distinction

sudhakarsingh27 (approved):
LGTM!

(Major change is replacing `param_with_axes` with `with_logical_partitioning` which enables MaxText runs with partitioning - especially TP)

PR #1779, opened 2025-05-13
[common] Added support of FP4 data type

Oleg-Goncharov:
# Description

Added support of FP4 data type.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Changed the type of the element size field from **size_t** to **double** to support fractional byte sizes (half of a byte for FP4)

# Checklist:


transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -557,7 +557,7 @@ CommOverlapP2PBase::CommOverlapP2PBase(const std::vector<size_t> &buffer_shape,
 
   // Create workspace tensor with userbuffer
   NVTE_CHECK(buffer_shape.size() == 2, "Userbuffer shape must be 2-dimensional!");
-  size_t buffer_bytes = buffer_shape[0] * buffer_shape[1] * typeToSize(buffer_dtype);
+  size_t buffer_bytes = static_cast<size_t>(buffer_shape[0] * buffer_shape[1] * typeToSize(buffer_dtype));
```
ptrendx, line 560: I would say that expressions like this should be a separate function, checking whether the size is divisible by 2 when type is fp4.

transformer_engine/common/common.h
```diff
@@ -552,7 +586,7 @@ inline bool is_aligned_tensor_data(const Tensor &t, size_t alignment) {
   return is_aligned_ptr(static_cast<const void *>(t.data.dptr), alignment);
 }
 
-size_t typeToSize(const DType type);
+double typeToSize(const DType type);
```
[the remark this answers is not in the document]
ptrendx: Nice suggestion.

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -393,6 +397,13 @@ enum class DType {
  */
 bool is_fp8_dtype(const DType t);
 
+/*! \brief Check if TE datatype is FP4
+ *
+ * Return true if TE datatype is FP4
+ *  \param[in] DType      TE Datatype of interest
+ */
+bool is_fp4_dtype(const DType t);
```
ptrendx, line 405: Both this and the fp8 function should be inline - this is header only part.

transformer_engine/common/utils.cuh
```diff
@@ -10,6 +10,7 @@
 #include <cuda_bf16.h>
 #include <cuda_fp16.h>
 #include <cuda_fp8.h>
+#include <cuda_fp4.h>
```
ptrendx, line 13: Need guard.

transformer_engine/common/normalization/common.cpp
```diff
@@ -212,8 +212,9 @@ CudnnNormalizationPlan::CudnnNormalizationPlan(NVTE_Norm_Type NormType, NVTE_Nor
   }
 
   const auto gamma_dtype = use_zero_centered_gamma_in_weight_dtype() ? wtype : ctype;
+  NVTE_CHECK(gamma_dtype != DType::kFloat4E2M1, "Gamma of type FP4 is not supported");
```
timmoon10: Nit: enumerating supported cases will generalize better than enumerating the unsupported cases.
```suggestion
  NVTE_CHECK(gamma_dtype == DType::kFloat32
             || gamma_dtype == DType::kFloat16
             || gamma_dtype == DType::kBFloat16,
             "Unsupported dtype for gamma parameter");
```

transformer_engine/common/common.h
```diff
@@ -412,8 +468,9 @@ struct TypeInfo {
       using type = byte;                                     \
       { __VA_ARGS__ }                                        \
     } break;                                                 \
+      SWITCH_FP4_TYPE_HANDLE(type, __VA_ARGS__)              \
     default:                                                 \
-      NVTE_ERROR("Invalid type.");                           \
+      NVTE_ERROR("Invalid type MARKED.");                    \
```
timmoon10: Debugging code? If these error messages are not descriptive enough, I wonder how we should improve them. They should already include the function names, but perhaps that is too cumbersome.
```suggestion
      NVTE_ERROR("Invalid type.");                    \
```
I see `MARKED` in several other functions as well.

transformer_engine/common/common.h
```diff
@@ -636,7 +706,7 @@ CUtensorMapDataType get_CUtensorMapDataType(DType dtype);
 void create_2D_tensor_map(CUtensorMap &tensorMap, const SimpleTensor &tensor,
                           const uint64_t globalY, const uint64_t globalX, const uint32_t shmemY,
                           const uint32_t shmemX, const uint32_t stride_elems,
-                          const uint32_t offset_elems, const size_t type_size);
+                          const uint32_t offset_elems, const double type_size);
```
timmoon10: We should use the size in bits instead of `double`:
```suggestion
                          const uint32_t offset_elems, const size_t type_num_bits);
```

tests/cpp/test_common.h
```diff
@@ -515,8 +544,16 @@ constexpr int32_t blackwellComputeCapability = 100;
...
+            { \
+                using type = fp8e8m0; \
+                {__VA_ARGS__} \
+            } \
+        break; \
+        SWITCH_FP4_TYPE_HANDLE(type, __VA_ARGS__) \
         default: \
-            NVTE_ERROR("Invalid type."); \
+            printf("dtype: %d\n", static_cast<int>(dtype)); \
+            NVTE_ERROR("Invalid type MARKED TEST."); \
```
timmoon10, line 556: Debug code here and elsewhere in the file.
```suggestion
            NVTE_ERROR("Invalid type."); \
```

transformer_engine/common/common.h
Oleg-Goncharov: Ah, yes, forgot to remove it after debugging

timmoon10 (approved):
LGTM

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -87,6 +88,10 @@ enum NVTEScalingMode {
    */
   NVTE_BLOCK_SCALING_1D = 2,
   NVTE_BLOCK_SCALING_2D = 3,
+  /*! Single NVFP4 scale per block of 16 contiguous elements in forward pass (FWD),
+    and single MXFP8 scale per block of 32 contiguous elements in backward pass (BWD).
+  */
+  NVTE_FWD_NVFP4_BWD_MXFP8_SCALING = 4,
```
timmoon10, line 94: I'm not a fan of this name since it's constraining our usage. Better to describe what it is, and let users do with it what they will. There's a high probability that we'll eventually do some NVFP4 operation in the backward pass, so this name will become inaccurate.
```suggestion
  /*! Single NVFP4 scale per block of 16 contiguous elements for row-wise usage,
    and single MXFP8 scale per block of 32 contiguous elements for column-wise usage.
  */
  NVTE_HYBRID_NVFP4_MXFP8_SCALING = 4,
```

There's something to be said for just having a pure NVFP4 tensor, and letting the framework extensions deal with the logic of mixing NVFP4 and MXFP8 usages based on the recipe. That would simplify the core lib, but it somewhat awkward for the frameworks.

PR #1749, opened 2025-05-05
Don't serialize a None tensor if not using fp8

pstjohn:
# Description

Currently, TE will serialize a `None` tensor in a layer's `_extra_state` key in the state dict if there's no fp8 metadata. This extra key has been causing some issues in downstream libraries that are not set up to handle the extra_state variable info; so where it's possible to omit it it will likely improve compatibility.

Fixes # (issue)

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Returns early from `get_extra_state` with a None if no fp8_checkpoint is being saved.

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #1786, opened 2025-05-14
Revert "Don't serialize a None tensor if not using fp8"

pstjohn:
Reverts NVIDIA/TransformerEngine#1749

This actually doesn't do what I expected it to; which is to avoid creating the `_extra_state` key in the torch state_dict. A closer reading of pytorch's state_dict methods indicates that it checks for the presence of this function rather than examining it's output when determining whether to serialize this extra parameter.

https://github.com/pytorch/pytorch/blob/dda2c7c8fc0bf87a5996c27df89a514d9d2b0a5e/torch/nn/modules/module.py#L2148-L2152

We'll need to take a different approach to ensure that we're compatible with `save_pretrained` / `from_pretrained`.


timmoon10 (commented):
Is there any harm in keeping https://github.com/NVIDIA/TransformerEngine/pull/1749? It doesn't help with the `_extra_state` issue, but it does remove pickling when FP8 is not enabled.

timmoon10 (approved):

PR #1789, opened 2025-05-15
Added token ignoring for CE loss

sanandaraj5597:
# Description

Added support for ignoring tokens while computing CE loss and gradients for loss.


ksivaman (commented):
Seems reasonable. Let's expand the current test to add this as well?

sanandaraj5597:
@ksivaman Added tests, I think we are good to go.

ksivaman (approved):

PR #1793, opened 2025-05-16
[PyTorch][MoE] Reduce CPU Overhead By Fuse Torch Empty Calls

zhongbozhu:
# Description

Bulk allocates quantization buffer for FP8 GroupLinear to reduce CPU overhead. 

Related PR about code refactoring: https://github.com/NVIDIA/TransformerEngine/pull/1836

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- tex.fused_multi_quantize now with split & bulk alloc feature fused

# Checklist:


transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -200,8 +201,8 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
         py::arg("otype"), py::arg("sm_margin"), py::arg("zero_centered_gamma"));
   m.def("rmsnorm_bwd", &transformer_engine::pytorch::rmsnorm_bwd, "Backward of RMSNorm");
   m.def("fused_multi_quantize", &transformer_engine::pytorch::fused_multi_quantize,
-        "Fused Multi-tensor Cast + Transpose", py::arg("input_list"), py::arg("output_list"),
-        py::arg("quantizer_list"), py::arg("otype"));
+        "Fused Multi-tensor Cast + Transpose", py::arg("input_list"), py::arg("input_view"),
+        py::arg("m_splits"), py::arg("quantizer_list"), py::arg("otype"));
```
timmoon10, line 205: I think the function name should be a bit more descriptive since it's now more MoE-specific than before:
```suggestion
  m.def("fused_multi_quantize_from_view", &transformer_engine::pytorch::fused_multi_quantize,
        "Fused split + multi-tensor quantize", py::arg("input_list"), py::arg("input_view"),
        py::arg("m_splits"), py::arg("quantizer_list"), py::arg("otype"));
```

zhongbozhu: Yeah, I also plan to remove having to pass in the input_list because torch.split is another big source of cpu overhead. 

However, there are two cases to consider for removing torch split: 
- FP8: split happens in multi-quantize
- BF16: split happens right before gemm

It would make the code cleaner if in the bf16 path, there will be a line that explicitly says it splits the views, but since we are in the MoE path, passing N py objects into C++ and fetch it back is too slow. Still I am not sure how to avoid that torch.split though.

timmoon10: If we go that way, the FP8 operation should have a name like `fused_split_quantize`. Then the following code seems logical:
```python
if with_quantized_compute:
    inputmats = tex.fused_split_quantize(inp_view, m_splits, input_quantizers)
else:
    inputmats = torch.split(inp_view, m_splits)

general_grouped_gemm(weights_fp8, inputmats, ...)
```
If we really wanted to optimize the BF16 path, we could modify the grouped GEMM function to take in the full tensor and split it internally, but I don't like how that would result in different APIs for BF16 and FP8.

transformer_engine/pytorch/csrc/extensions/transpose.cpp
```diff
@@ -4,54 +4,305 @@
...
+
+    py::object ret =
+        Float8BlockwiseQTensorClass(rowwise_data, rowwise_scale, columnwise_data, columnwise_scale,
+                                    fp8_dtype, quantizer_list[i], is_2D_scaled);
+
+    output_list_py.emplace_back(std::move(ret));
+
+    // as for tensor wrappers, these tensor wrappers are going to be quantized, so no need to insert empty tensors here
+    if (m_splits[i] > 0) {
+      TensorWrapper te_tensor = makeTransformerEngineTensor(
```
HuangHunag-MT, line 222: Hi

In the original logic:

1. Even when input_tensor.numel() == 0, we would still call either:
- makeTransformerEngineTensor() (in create_tensor when output_list == std::nullopt), or
- makeTransformerEngineTensor((*output_list)[i], quantizer_list[i])
This PR appears to change that behavior. Could you share the rationale behind this modification?

2. Additionally, while working on a per-tensor delayed-FP8 implementation inspired by this PR, I noticed that when m_splits contains 0 values and we skip makeTransformerEngineTensor, it would encounter a error in the following GEMM:
`Unable to cast Python instance of type <class 'NoneType'> to C++ type 'at::Tensor'`

zhongbozhu: Thanks for the comment. 

1. yes, the part you are talking about here is one the hacky part in the code. In the original for loop of creating tensors, input tensor wrappers & output tensor wrappers have the same list size (ie. smaller than m_split by not appending the empty tensors). Unfortunately, when working on the fused allocation function, I returned a list of output tensor wrappers of the original size containing empty tensors. In order to make this PR compatible with all the fp8 recipes (ie. use fused allocation for FP8 subchannel & fall back to a for loop for others), I had to skip empty tensors for both input and output tensor wrappers and add this hacky non_empty_tensor_idx to the loop of quantizing. 
```
      quantize_cpp(tensor_wrappers_input[non_empty_tensor_idx], quantizer_list[i], quantizers[i],
                   tensor_wrappers_output[non_empty_tensor_idx], std::nullopt);
```

2. now for fp8 delayed scaling, things become a little different. Since fp8 delayed scaling had this fused quantization implemented, so we don't have to use a for loop and call quantize. `nvte_multi_cast_transpose` accepts tensor lists without empty tensors (haven't tried if it still works if empty tensors are passed in), so for both its input and output list, it needs to be same length and have empty tensors skipped.

timmoon10 (approved):
LGTM

yaox12 (approved):

yaox12:
CI passed except for irrelevant failures. Moving to merge.

PR #1788, opened 2025-05-15
Fix miscellaneous bugs during library loading

ksivaman:
# Description

A cleanup and minor refactor of the runtime library loading infrastructure during TE import, along with miscellaneous bug fixes.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Infra/Build change
- [x] Code refactoring

## Changes

- Avoid use of `subprocess` to run python/pip, this is leading to esoteric failures during library loading in various environments.
- Consolidate logic for finding TE's shared object files in a single function based on various cases.
- Cleanup the framework's `__init__.py` files.
- Fix various cases of library loading after editable/source/wheel install and catch errors in case of any discrepancies.
- Remove a few unused lines in main `setup.py`.
- Fix duplication of shared object file during editable installation.

## Testing checklist (local)

- [x] Verified correct working of all 4 "cases" as mentioned in the comments for the `_get_shared_object_file` function in `transformer_engine/common/__init__.py` for both frameworks.
- [x] Tested wheel for core and sdist for both frameworks, including catching all errors during in the `load_framework_extension` function in `transformer_engine/common/__init__.py`.

## Future work

Create robust build testing suite to check for above cases in the CI.

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #1801, opened 2025-05-20
Avoid searching unnecessary dirs for shared libs

timmoon10:
# Description

https://github.com/NVIDIA/TransformerEngine/pull/1788 changed the `.so` import logic to search multiple paths since the different build paths (editable builds, source builds, Pip wheels) might put the `.so`s in different places. However, the current implementation is slow since it traverses the entire directory (including `.git`). This PR fixes the search logic to only look at required paths.

I've also modified the search logic to prioritize the imported TE module. If I manually set `PYTHONPATH` to point to my local install and it has everything it needs, why should we care at all about whatever is installed in the site-packages dir? I don't think this should affect correctness with the different build paths, but pinging @ksivaman.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Only look in search dirs when importing `.so`s.
- Only look for `.so`s in site-packages dir if not found in imported module.

# Checklist:

### DIFF
--- a/transformer_engine/common/__init__.py
+++ b/transformer_engine/common/__init__.py
@@ -4,25 +4,26 @@
 
 """FW agnostic user-end APIs"""
 
-import sys
-import glob
-import sysconfig
-import subprocess
 import ctypes
+import functools
+import glob
+import importlib
+from importlib.metadata import version, metadata, PackageNotFoundError
 import logging
 import os
-import platform
-import importlib
-import functools
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


ksivaman (approved):
With this, we are giving top priority to local shared objects even if they are stale and if a more recent SO is available in an installation elsewhere. It's not my preferred choice but I don't feel strongly enough about it to block, especially as I can't think of a cleaner way to fix user installations yet (`--user`).

**_Note_**: The onus is on the developer to ensure a clean working dir for those specific failing cases.

PR #1818, opened 2025-05-23
[PyTorch+JAX] Enable loading only one framework extension when both frameworks are importable

denera:
# Description

Top-level Transformer Engine import always attempts to load the compiled extensions for all importable frameworks, which inevitably errors out when both frameworks are importable, but TE was installed with only one of the compiled extensions. This defeats the point of allowing users to control which extension to install with `NVTE_FRAMEWORK=<framework>` or  `pip install transformer_engine[<framework>]`. 

This PR safeguards the `load_framework_extension()` calls in `transformer_engine/<framework>/__init__.py` to detect when the framework is present but the matching extension is not, issue a useful `RuntimeWarning`, and then replace the `RuntimeError` with an `ImportError` that would be caught by the top-level `transformer_engine/__init__.py` same as when the framework itself fails to import.

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

# Checklist:


timmoon10 (approved):
LGTM. This should fix the same bug as https://github.com/NVIDIA/TransformerEngine/pull/1825. We could follow the suggestion at https://github.com/NVIDIA/TransformerEngine/pull/1825#discussion_r2110033863 and change some of the `RuntimeError`s to `FileNotFoundError`s.

I think I slightly prefer https://github.com/NVIDIA/TransformerEngine/pull/1825, since that improves the error messages for some other use cases (e.g. throw `FileNotFoundError` for failure when importing the core lib). This PR is somewhat orthogonal though, and I like how it converts the `.so` load error to `ImportError`.

denera:
@timmoon10 I'm not sure if this is orthogonal anymore to #1825. @ksivaman incorporated the same core idea into his PR by making `load_framework_extension()` throw a `FileNotFoundError` instead of `RuntimeError`, and then catching that in the top-level `transformer_engine/__init__.py` alongside `ImportError`s. So I think we can just close this PR now.

ksivaman:
Closing in favor of #1834

PR #1823, opened 2025-05-27
FP8 Param support for offloading

sanandaraj5597:
This PR adds support for CPU offloading when FP8 params feature is enabled.


pggPL:
This fixes some very specific use case scenario - the problem of double offloading happens more often. But I think it is ok to merge it now and work on new offloading logic in parallel.

pggPL (approved):

PR #1824, opened 2025-05-27
Add cuBLASMp-backed GEMM-like API to TE common

mk-61:
# Description

This change adds to TE common an API for distributed GEMMs, implemented by cuBLASMp.
Functionality should be similar to the existing comm_gemm_overlap implementation, which is bases on UserBuffers.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- A new API added 

# Checklist:


transformer_engine/common/include/transformer_engine/comm_gemm.h
```diff
@@ -0,0 +1,45 @@
...
+
+void nvte_gemm_reduce_scatter(CommGemmCtx* ctx, int64_t m, int64_t n, int64_t k, const NVTETensor a,
+                              const NVTETensor b, const NVTETensor d, const NVTETensor bias,
+                              const NVTETensor pre_act_out, bool transa, bool transb, bool grad,
+                              bool accumulate, int comm_sm_count, cudaStream_t main_stream);
+
+void nvte_gemm_all_reduce(CommGemmCtx* ctx, int64_t m, int64_t n, int64_t k, const NVTETensor a,
+                          const NVTETensor b, const NVTETensor d, const NVTETensor bias,
+                          const NVTETensor pre_act_out, bool transa, bool transb, bool grad,
+                          bool accumulate, int comm_sm_count, cudaStream_t main_stream);
```
denera: @mk-61 Could you please add docstrings to these?

transformer_engine/common/comm_gemm/comm_gemm.cpp
```diff
@@ -0,0 +1,504 @@
...
+                static_cast<const Tensor*>(pre_act_out), transa, transb, grad, accumulate,
+                comm_sm_count, main_stream);
+}
+
+void nvte_gemm_reduce_scatter(CommGemmCtx* ctx, int64_t m, int64_t n, int64_t k, const NVTETensor a,
+                              const NVTETensor b, const NVTETensor d, const NVTETensor bias,
+                              const NVTETensor pre_act_out, bool transa, bool transb, bool grad,
+                              bool accumulate, int comm_sm_count, cudaStream_t main_stream) {
+  NVTE_API_CALL(nvte_gemm_reduce_scatter);
+  cublasmp_gemm(GemmRsInitMatrices, ctx, CUBLASMP_MATMUL_ALGO_TYPE_DEFAULT, m, n, k,
```
denera: We need to expose the algorithm choice in the top-level API instead of always choosing the default here. Our downstream dependencies don't always use the same method/algo for every GEMM+RS overlap.

transformer_engine/common/include/transformer_engine/comm_gemm.h
```diff
@@ -0,0 +1,45 @@
...
+#include <nccl.h>
+#include <stdint.h>
+
+#include "transformer_engine.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct CommGemmCtx CommGemmCtx;
```
timmoon10: Publicly visible classes should be namespaced with `NVTE`:
```suggestion
typedef struct NVTECommGemmCtx NVTECommGemmCtx;
```

almogsegal:
@mk-61 cuBLASMp v0.5.0 is out, you can use the wheels for PyPI: https://pypi.org/project/nvidia-cublasmp-cu12/0.5.0.898/

transformer_engine/common/util/logging.h
```diff
@@ -87,4 +92,27 @@
...
+
+#define NVTE_CHECK_CUBLASMP(expr)                             \
+  do {                                                        \
+    const cublasMpStatus_t status = (expr);                   \
+    if (status != CUBLASMP_STATUS_SUCCESS) {                  \
+      NVTE_ERROR("cuBLASMp Error: ", std::to_string(status)); \
+    }                                                         \
+  } while (false)
+
+#define NVTE_CHECK_MPI(expr)                         \
```
denera: The cuBlasMp bindings don't depend on MPI and the unit tests have their own `CHECK_MPI` macro, so this `NVTE_CHECK_MPI` seems to be unused. Can we remove it?

mk-61: Good catch (remnants of now removed libcal dependencies), fixed.

setup.py
```diff
@@ -66,6 +67,18 @@ def setup_common_extension() -> CMakeExtension:
     if bool(int(os.getenv("NVTE_BUILD_ACTIVATION_WITH_FAST_MATH", "0"))):
         cmake_flags.append("-DNVTE_BUILD_ACTIVATION_WITH_FAST_MATH=ON")
 
+    if bool(int(os.getenv("NVTE_WITH_CUBLASMP", "1"))):
```
timmoon10: This is still experimental and not really exposed to users, so how about we disable it by default and set `NVTE_WITH_CUBLASMP=1` in our internal build?
```suggestion
    if bool(int(os.getenv("NVTE_WITH_CUBLASMP", "0"))):
```
If we expose this by default, we need to add it as a Python dependency so users installing with Pip will have access to cuBLASMp and NVSHMEM.

tests/cpp/CMakeLists.txt
```diff
@@ -37,10 +37,12 @@ find_library(TE_LIB NAMES transformer_engine PATHS "${TE_LIB_PATH}/.." ${TE_LIB_
 message(STATUS "Found transformer_engine library: ${TE_LIB}")
 include_directories(../../transformer_engine/common/include)
 include_directories(../../transformer_engine/common)
+include_directories(../../transformer_engine)
 include_directories(${CMAKE_SOURCE_DIR})
 
 find_package(CUDAToolkit REQUIRED)
 include(${CMAKE_SOURCE_DIR}/../../3rdparty/cudnn-frontend/cmake/cuDNN.cmake)
 
+add_subdirectory(comm_gemm)
```
timmoon10, line 46: This is fine as a quick hack, but if we add more L1 tests we should have separate build processes for L0 and L1 tests. This way the L0 tests don't need to build the L1 tests or require MPI. It also removes the need for the hacky regex exclusions in the QA script.

transformer_engine/common/CMakeLists.txt
```diff
@@ -141,6 +149,25 @@ if (NVTE_ENABLE_NVSHMEM)
...
+                 NAMES cublasmp libcublasmp
+                 PATHS ${CUBLASMP_DIR}
+                 PATH_SUFFIXES lib
+                 REQUIRED)
+    find_library(NVSHMEM_HOST_LIB
+                 NAMES nvshmem_host libnvshmem_host.so.3
+                 PATHS ${NVSHMEM_DIR}
+                 PATH_SUFFIXES lib
+                 REQUIRED)
+  target_link_libraries(transformer_engine PUBLIC ${CUBLASMP_LIB} ${NVSHMEM_HOST_LIB})
```
timmoon10, line 166: Do we see linker errors when importing `libtransformer_engine.so`? If we do, we might need to manually import cuBLASMp and NVSHMEM like we do with other NVIDIA libs:
https://github.com/NVIDIA/TransformerEngine/blob/92f431bfee3bebbf49d1cc6f6bc37796bffd8bb7/transformer_engine/common/__init__.py#L304-L308
This is because the NVIDIA libs might be installed in places where the linker can't automatically find them, especially if users install them via Pip.

tests/cpp/comm_gemm/test_comm_gemm.cu
```diff
@@ -0,0 +1,440 @@
...
+                               cudaMemcpyDefault));
+
+    auto out_golden = CopyMatrix(out_golden_global, dims.d_rows_start, dims.d_cols_start,
+                                 dims.d_rows_num, dims.d_cols_num, m);
+    NVTE_CHECK(out.size() == out_golden.size());
+    for (size_t i = 0; i < out.size(); ++i) {
+      if (rand() % 100 < 3) {
+        // std::cerr << "== at " << rank_ << ": " << i << ": " << static_cast<float>(out[i]) << " "
+        //           << static_cast<float>(out_golden[i]) << std::endl;
+      }
```
ptrendx: Leftover code.

transformer_engine/common/comm_gemm/comm_gemm.cpp
```diff
@@ -0,0 +1,502 @@
...
+                        bool transb) {
+  const auto a0 = a->flat_first_dim();
+  const auto a1 = a->flat_last_dim();
+  const auto b0 = b->flat_first_dim();
+  const auto b1 = b->flat_last_dim();
+  const auto d0 = d->flat_first_dim();
+  const auto d1 = d->flat_last_dim();
+
+  if (transa) {
+    NVTE_CHECK(a1 == k);
```
ptrendx: This and a lot of the other subsequent NVTE_CHECKs: we should have descriptive error messages there to give users some feedback on what went wrong.

transformer_engine/common/comm_gemm/comm_gemm.cpp
```diff
@@ -0,0 +1,502 @@
...
+  const auto d0 = d->flat_first_dim();
+  const auto d1 = d->flat_last_dim();
+
+  if (transa) {
+    NVTE_CHECK(a0 == m);
+    NVTE_CHECK_CUBLASMP(cublasMpMatrixDescriptorInit(k, m, block_size(ctx, k), m, 0, 0,
+                                                     block_size(ctx, k), get_cuda_dtype(a->dtype()),
+                                                     ctx->grid_col_major.get(), ctx->a_desc.get()));
+  } else {
+    NVTE_CHECK(false);
```
ptrendx: NVTE_ERROR? Also, error message should be provided.

transformer_engine/common/comm_gemm/comm_gemm.cpp
```diff
@@ -0,0 +1,502 @@
...
+using InitMatricesFn = void (*)(CommGemmCtx*, int64_t*, int64_t, int64_t, int64_t, const Tensor*,
+                                const Tensor*, const Tensor*, bool, bool);
+
+void cublasmp_gemm(InitMatricesFn init_matrices_fn, CommGemmCtx* ctx, cublasMpMatmulAlgoType_t algo,
+                   int64_t m, int64_t n, int64_t k, const Tensor* a, const Tensor* b,
+                   const Tensor* d, const Tensor* bias, const Tensor* pre_act_out, bool transa,
+                   bool transb, bool grad, bool accumulate, int comm_sm_count,
+                   cudaStream_t main_stream) {
+  for (auto t : {a, b, d}) {
+    NVTE_CHECK(is_tensor_scaling(t->scaling_mode), "Unsupported scaling mode");
```
ptrendx: Add the scaling mode that triggered the problem in the error message.

transformer_engine/common/include/transformer_engine/comm_gemm.h
```diff
@@ -0,0 +1,75 @@
...
+/*! \brief AllGather-GEMM
+ *
+ * m, n, k - Global GEMM dimensions.
+ * Tensors and boolean flags have the same meaning as in nvte_cublas_gemm.
+ */
+void nvte_all_gather_gemm(CommGemmCtx* ctx, int64_t m, int64_t n, int64_t k, const NVTETensor a,
+                          const NVTETensor b, const NVTETensor d, const NVTETensor bias,
+                          const NVTETensor pre_act_out, bool transa, bool transb, bool grad,
+                          bool accumulate, int comm_sm_count, cudaStream_t main_stream,
+                          cublasMpMatmulAlgoType_t algo = CUBLASMP_MATMUL_ALGO_TYPE_DEFAULT);
```
ptrendx: Do we want to expose the algo from cuBLASMp? Maybe better would be to have our own enum (just in case we would want/need to add something else through that API without going to cuBLASMP necessarily in the future?

transformer_engine/common/include/transformer_engine/comm_gemm.h
```diff
@@ -0,0 +1,75 @@
...
+ */
+
+/*! \brief AllGather-GEMM
+ *
+ * m, n, k - Global GEMM dimensions.
+ * Tensors and boolean flags have the same meaning as in nvte_cublas_gemm.
+ */
+void nvte_all_gather_gemm(CommGemmCtx* ctx, int64_t m, int64_t n, int64_t k, const NVTETensor a,
+                          const NVTETensor b, const NVTETensor d, const NVTETensor bias,
+                          const NVTETensor pre_act_out, bool transa, bool transb, bool grad,
```
ptrendx, line 84: Bool needs stdbool.h (or we can use ints there).

transformer_engine/common/common.h
```diff
@@ -382,9 +384,19 @@ struct BitsNumber {
 template <typename T>
 struct TypeInfo {
 #if FP4_TYPE_SUPPORTED
-  using types = std::tuple<byte, int16, int32, int64, fp32, fp16, bf16, fp8e4m3, fp8e5m2, fp4e2m1>;
+  using types = std::tuple<byte, int16, int32, int64, fp32, fp16, bf16, fp8e4m3, fp8e5m2, fp4e2m1
+#if CUDA_VERSION >= 12080
+                           ,
+                           fp8e8m0
+#endif
```
ptrendx, line 391: Nice, but then we should also put the fp4e2m1 type here as well.

transformer_engine/common/include/transformer_engine/comm_gemm.h
```diff
@@ -0,0 +1,75 @@
...
+ * Tensors and boolean flags have the same meaning as in nvte_cublas_gemm.
+ */
+void nvte_gemm_all_reduce(CommGemmCtx* ctx, int64_t m, int64_t n, int64_t k, const NVTETensor a,
+                          const NVTETensor b, const NVTETensor d, const NVTETensor bias,
+                          const NVTETensor pre_act_out, bool transa, bool transb, bool grad,
+                          bool accumulate, int comm_sm_count, cudaStream_t main_stream,
+                          cublasMpMatmulAlgoType_t algo = CUBLASMP_MATMUL_ALGO_TYPE_DEFAULT);
+
+/*! \brief Get local number of rows or columns.
+ */
```
ptrendx: This does require more explanation.

transformer_engine/common/include/transformer_engine/comm_gemm.h
```diff
@@ -0,0 +1,75 @@
...
+void nvte_gemm_reduce_scatter(CommGemmCtx* ctx, int64_t m, int64_t n, int64_t k, const NVTETensor a,
+                              const NVTETensor b, const NVTETensor d, const NVTETensor bias,
+                              const NVTETensor pre_act_out, bool transa, bool transb, bool grad,
+                              bool accumulate, int comm_sm_count, cudaStream_t main_stream,
+                              cublasMpMatmulAlgoType_t algo = CUBLASMP_MATMUL_ALGO_TYPE_DEFAULT);
+
+/*! \brief GEMM-AllReduce.
+ *
+ * m, n, k - Global GEMM dimensions.
+ * Tensors and boolean flags have the same meaning as in nvte_cublas_gemm.
```
ptrendx: I would say we should either repeat it here or, at the very least, tell more about the parameters that do not exist in the regular gemm.

transformer_engine/common/include/transformer_engine/comm_gemm.h
```diff
@@ -0,0 +1,75 @@
...
+ */
+CommGemmCtx* nvte_comm_gemm_ctx_create(ncclComm_t comm, int nranks, int rank, int local_device);
+
+/*! \brief Destroy a comm-gemm context.
+ */
+void nvte_comm_gemm_ctx_destroy(CommGemmCtx* ctx);
+
+/*
+ * Refer here: https://docs.nvidia.com/cuda/cublasmp/usage/tp.html for additional details.
+ */
```
ptrendx: What additional details? If on context useful when using those functions then it should be part of the docstring of the functions affected.

transformer_engine/common/include/transformer_engine/comm_gemm.h
```diff
@@ -0,0 +1,75 @@
...
+#include "transformer_engine.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct CommGemmCtx CommGemmCtx;
+
+/*! \brief Create a comm-gemm context.
+ */
```
ptrendx: Should have parameters etc. listed and described. Also, I don't quite like the "comm-gemm" name - if you do not know what this is about it would not tell you anything on its own. Maybe collective gemm would be better?

transformer_engine/common/include/transformer_engine/comm_gemm.h
```diff
@@ -0,0 +1,75 @@
...
+
+/*! \brief Destroy a comm-gemm context.
+ */
+void nvte_comm_gemm_ctx_destroy(CommGemmCtx* ctx);
+
+/*
+ * Refer here: https://docs.nvidia.com/cuda/cublasmp/usage/tp.html for additional details.
+ */
+
+/*! \brief AllGather-GEMM
```
ptrendx: This and the subsequent functions should be better documented about what they actually do. Here is a snippet generated by Claude, maybe we could use some of that to make it clearer?

### Context Management
- **`nvte_comm_gemm_ctx_create()`** - Creates a communication context with NCCL communicator, rank information, and device details
- **`nvte_comm_gemm_ctx_destroy()`** - Cleans up the context

### Communication-GEMM Patterns
These functions implement common distributed computing patterns that fuse communication with matrix multiplication:

- **`nvte_all_gather_gemm()`** - Performs AllGather communication followed by GEMM
  - Gathers distributed data from all ranks, then computes matrix multiplication
  - Useful when input data is distributed but full matrix is needed for computation

- **`nvte_gemm_reduce_scatter()`** - Performs GEMM followed by ReduceScatter communication  
  - Computes matrix multiplication, then distributes results across ranks with reduction
  - Common in distributed neural network training for gradient aggregation

- **`nvte_gemm_all_reduce()`** - Performs GEMM followed by AllReduce communication
  - Computes matrix multiplication, then reduces results across all ranks
  - Used when all ranks need the complete reduced result

### Utility
- **`nvte_comm_gemm_numroc()`** - Calculates local matrix dimensions from global size
  - Determines how many rows/columns each rank should handle

transformer_engine/common/include/transformer_engine/comm_gemm.h
mk-61: Done

setup.py
mk-61: Done

transformer_engine/common/CMakeLists.txt
mk-61: Not sure how the magic you quoted works, I rely on ldconfig, in general.

tests/cpp/comm_gemm/test_comm_gemm.cu
mk-61: Done

transformer_engine/common/comm_gemm/comm_gemm.cpp
mk-61: Done.

transformer_engine/common/comm_gemm/comm_gemm.cpp
mk-61: Done

transformer_engine/common/comm_gemm/comm_gemm.cpp
mk-61: Done

transformer_engine/common/include/transformer_engine/comm_gemm.h
mk-61: Well, I know a lot of people would prefer this (and most AI bots advise it too).

But my original idea about this API was _not_ to create an extra layer of abstraction. More like a thin - as thin as possible - TE-native (as in tensors, dtypes) binding to cuBLASMp.

transformer_engine/common/include/transformer_engine/comm_gemm.h
mk-61: Added, although the name itself originated somewhere in dark ages of PBLAS / ScaLAPACK libraries, so not entirely made up.

transformer_engine/common/include/transformer_engine/comm_gemm.h
mk-61: Done

transformer_engine/common/include/transformer_engine/comm_gemm.h
mk-61: Moved to the top, to module-level comments.
cuBLASMp really has a nice documentation there.

transformer_engine/common/include/transformer_engine/comm_gemm.h
mk-61: Borrowed & incorporated most of these.

transformer_engine/common/include/transformer_engine/comm_gemm.h
mk-61: Included, although it is only needed for C compiler, and I don't think our headers would compile anyways (e.g. I've seen enum class used somewhere).

timmoon10 (commented):
Build system logic looks good to me. I'm happy merging once tests pass and @ptrendx is fine with the impl.

transformer_engine/common/include/transformer_engine/comm_gemm.h
```diff
@@ -0,0 +1,151 @@
...
+#endif
+
+typedef struct NVTECommGemmCtx NVTECommGemmCtx;
+
+/*! \brief Create a comm-gemm context.
+ *
+ *  \param[in]  comm          NCCL communicator.
+ *  \param[in]  nranks        Number of ranks.
+ *  \param[in]  rank          Local rank.
+ *  \param[in]  local_device  CUDA device, used by this rank.
```
ptrendx: Why do we need to ask for the device? If we already have the NCCL communicator then we should have a context there as well, so we can just query it with cudaGetDevice, no?

transformer_engine/common/include/transformer_engine/comm_gemm.h
ptrendx: I know that we are not super great about this C compatibility right now (although most places are ok), but we should strive for it to avoid problems in the future that would be very hard to debug. One additional concern is the usage of the default arguments which is also not legal under C.

transformer_engine/common/include/transformer_engine/comm_gemm.h
ptrendx: I understand that, but that is not really shown in the API - `nvte_all_gather_gemm` does not sound like a cublasmp wrapper, but rather like a general API. Because of that I would prefer that indirection there.

transformer_engine/common/comm_gemm/comm_gemm.cpp
```diff
@@ -0,0 +1,505 @@
...
+                        bool transb) {
+  const auto a0 = a->flat_first_dim();
+  const auto a1 = a->flat_last_dim();
+  const auto b0 = b->flat_first_dim();
+  const auto b1 = b->flat_last_dim();
+  const auto d0 = d->flat_first_dim();
+  const auto d1 = d->flat_last_dim();
+
+  if (transa) {
+    NVTE_CHECK(a1 == k, "Unsupported tensor dimensions");
```
ptrendx:
```suggestion
    NVTE_CHECK(a1 == k, "Unsupported tensor dimensions. Expected shape compatible with [*, ", k,
               "] and got ", a->shape());
```

transformer_engine/common/comm_gemm/comm_gemm.cpp
```diff
@@ -0,0 +1,505 @@
...
+
+  // Pre-creating matrix descriptors here, will be initialized with the actual params later.
+  auto a_desc = CublasMpMatrixDescCreate(1, 1, 1, 1, 0, 0, 1, CUDA_R_16F, row_major.get());
+  auto b_desc = CublasMpMatrixDescCreate(1, 1, 1, 1, 0, 0, 1, CUDA_R_16F, row_major.get());
+  auto d_desc = CublasMpMatrixDescCreate(1, 1, 1, 1, 0, 0, 1, CUDA_R_16F, row_major.get());
+
+  auto matmul_desc = CublasMpMatmulDescCreate(CUBLAS_COMPUTE_32F);
+
+  return new NVTECommGemmCtx{
+      .nranks = nranks,
```
ptrendx, line 459: This type of initalization is technically C++20 feature and we are using C++17 during compilation. Also, while it should behave properly, it would probably be nice to explicitly set workspace field to nullptr.

transformer_engine/common/include/transformer_engine/comm_gemm.h
mk-61: Good catch, fixed - in fact, the variable wasn't used at all.

transformer_engine/common/comm_gemm/comm_gemm.cpp
mk-61: Fixed - added new enum and made API C-compatible.

transformer_engine/common/comm_gemm/comm_gemm.cpp
mk-61: Checked on both GCC and Clang, named initializers work starting with C++14, even though technically it's in C++20 standard.
And I would really hate to loose readability by reverting to positional initialization!

transformer_engine/common/include/transformer_engine/comm_gemm.h
```diff
@@ -0,0 +1,156 @@
+/*************************************************************************
+ * Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+
+/*! \file gemm.h
```
ptrendx:
```suggestion
/*! \file comm_gemm.h
```

ptrendx (approved):

PR #1827, opened 2025-05-27
[PyTorch] Recipe heuristics for initializing quantized weights

timmoon10:
# Description

When initializing a model with quantized weights, the required data is different for training and inference (training requires row-wise data for forward GEMM and column-wise data for dgrad GEMM, inference only requires column-wise). This PR adds a `heuristic` option to the quantization recipes, with support for `"performance"` and `"inference"`. In the future, we may want to consider a `"memory"` heuristic that prioritizes memory usage over performance.

This is similar to the heuristic API in https://github.com/NVIDIA/TransformerEngine/pull/1300.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add recipe heuristics for initializing quantized weights

# Checklist:

### DIFF
--- a/transformer_engine/common/recipe/__init__.py
+++ b/transformer_engine/common/recipe/__init__.py
@@ -61,10 +61,12 @@ class QParams:
     amax_epsilon: float = 0.0
 
 
+@dataclass
 class Recipe:
-    """
-    Base recipe class.
-    """
+    """Configuration for quantization scheme."""
+
+    # Recipe-specific heuristics (options: "performance", "inference")
+    heuristic: str = "performance"
 
     def mxfp8(self):
         """Whether the given recipe is MXFP8 block scaling."""
@@ -185,7 +187,8 @@ class DelayedScaling(Recipe):
             f"format={str(self.fp8_format).split('.')[1]}, "
             f"amax_history_len={self.amax_history_len}, "
             f"fp8_dpa={self.fp8_dpa}, "
-            f"fp8_mha={self.fp8_mha}"
+            f"fp8_mha={self.fp8_mha}, "
+            f"heuristic={self.heuristic}"
         )
 
 
@@ -228,7 +231,8 @@ class Float8CurrentScaling(Recipe):
             f"fp8_gemm_dgrad={self.fp8_gemm_dgrad}, "
             f"fp8_gemm_wgrad={self.fp8_gemm_wgrad}, "
             f"fp8_dpa={self.fp8_dpa}, "
-            f"fp8_mha={self.fp8_mha}"
+            f"fp8_mha={self.fp8_mha}, "
+            f"heuristic={self.heuristic}"
         )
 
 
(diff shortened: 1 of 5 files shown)


transformer_engine/pytorch/module/base.py
```diff
@@ -1238,15 +1252,19 @@ def get_weight_workspace(
...
-                tensor.update_usage(
-                    rowwise_usage=quantizer.rowwise_usage,
-                    columnwise_usage=quantizer.columnwise_usage,
-                )
+            update_rowwise_usage = True if quantizer.rowwise_usage else None
+            update_columnwise_usage = True if quantizer.columnwise_usage else None
+            tensor.update_usage(
+                rowwise_usage=update_rowwise_usage,
+                columnwise_usage=update_columnwise_usage,
+            )
```
timmoon10, line 1267: Destroying unnecessary usages causes problems when alternating between training steps (column-wise data needed) and validation steps (column-wise data not needed).  See https://github.com/NVIDIA/TransformerEngine/pull/1832#discussion_r2113016499.

transformer_engine/common/recipe/__init__.py
```diff
@@ -61,10 +61,12 @@ class QParams:
...
 
+@dataclass
 class Recipe:
-    """
-    Base recipe class.
-    """
+    """Configuration for quantization scheme."""
+
+    # Recipe-specific heuristics (options: "performance", "inference")
+    heuristic: str = "performance"
```
ptrendx, line 69: The name is not the best - wouldn't you want performance during inference?

transformer_engine/pytorch/module/base.py
ptrendx: TBH this issue is just because of optimizer not doing the right job with quantizing. If we made it so it uses the quantizer then we would not need this part at all.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -384,6 +384,7 @@ def _set_data(self, tensor: torch.Tensor) -> None:
 
         # Quantize to FP8
         assert self._quantizer is not None, "Can't quantize without a quantizer"
+        self._quantizer.internal = False
```
ptrendx, line 387: This makes me think that internal should maybe be an option to tex.quantize rather than the member of quantizer.

transformer_engine/pytorch/module/base.py
timmoon10: The problem is fundamental to validation steps. The validation steps don't require column-wise data, so we were aggressively deallocating to save memory. Once we started more training steps, we failed in bprop because we were missing column-wise data.

transformer_engine/common/recipe/__init__.py
timmoon10: Not necessarily if you're memory constrained.

Perhaps a naming scheme like "training_performance", "inference_performance", "training_memory", "inference_memory" would be better?

timmoon10:
https://github.com/NVIDIA/TransformerEngine/pull/1847 handles the main issue this PR was addressing: unnecessary memory usage when initializing quantized weights for use in inference. This API is more general though and we may revisit in the future for specialized use-cases, e.g. when it is worth sacrificing performance for reduced memory usage.

PR #1832, opened 2025-05-28
Fix MXFP8-training related issue

WanZzzzzz:
# Description

Fix 2 bugs in mxfp8 training.

(1) The following error happened after validation since colwise data is cleared by update_usage() during validation.

```
0: [rank0]:   File "/usr/local/lib/python3.12/dist-packages/transformer_engine/pytorch/tensor/_internal/mxfp8_tensor_base.py", line 192, in update_usage
0: [rank0]:     raise RuntimeError(
0: [rank0]: RuntimeError: Requested column-wise usage, but MXFP8Tensor is missing column-scaled FP8 data
```

(2) This happened after training completes and copying tensor to CPU:

```
1: [rank1]:   File "/usr/local/lib/python3.12/dist-packages/transformer_engine/pytorch/tensor/mxfp8_tensor.py", line 394, in _set_data                                                                           1: [rank1]:     new_device = tensor.device if tensor.is_cuda else self.device
1: [rank1]:                                   ^^^^^^^^^^^^^^
1: [rank1]: AttributeError: 'MXFP8TensorBase' object has no attribute 'is_cuda'
```

## Type of change

[ ] Bug fix (non-breaking change which fixes an issue)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/base.py
+++ b/transformer_engine/pytorch/module/base.py
@@ -1242,7 +1242,7 @@ class TransformerEngineBaseModule(torch.nn.Module, ABC):
 
         # FP8 primary weights
         if isinstance(tensor, QuantizedTensor):
-            if update_workspace and quantizer is not None:
+            if update_workspace and quantizer is not None and quantizer.columnwise_usage:
                 tensor.update_usage(
                     rowwise_usage=quantizer.rowwise_usage,
                     columnwise_usage=quantizer.columnwise_usage,
--- a/transformer_engine/pytorch/tensor/mxfp8_tensor.py
+++ b/transformer_engine/pytorch/tensor/mxfp8_tensor.py
@@ -346,7 +346,8 @@ class MXFP8Tensor(MXFP8TensorBase, QuantizedTensor):
         casts to FP8.
 
         """
-
+        if isinstance(tensor, MXFP8TensorBase):
+            return
         # Tensor device
         new_device = tensor.device if tensor.is_cuda else self.device
         if not devices_match(new_device, tensor.device):
--- a/transformer_engine/pytorch/module/linear.py
+++ b/transformer_engine/pytorch/module/linear.py
@@ -117,6 +117,7 @@ class _Linear(torch.autograd.Function):
         skip_fp8_weight_update: bool,
         symmetric_ar_type: str,
         debug: Optional[bool] = False,
+        needs_columnwise_after_validation: bool = True,
     ) -> torch.Tensor:
         # pylint: disable=missing-function-docstring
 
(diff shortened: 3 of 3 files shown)


transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -346,7 +346,8 @@ def _set_data(self, tensor: torch.Tensor) -> None:
         casts to FP8.
 
         """
-
+        if isinstance(tensor, MXFP8TensorBase):
+            return
```
timmoon10, line 350: This will result in situations where calling `mxfp8_tensor.data = bf16_tensor` will silently do nothing. Better to prevent this function from constructing a `MXFP8TensorBase`. We can do this by changing https://github.com/NVIDIA/TransformerEngine/blob/0d45556ccdb8e418f6eeb1a1e7440a1893c15320/transformer_engine/pytorch/tensor/mxfp8_tensor.py#L388
```python
self._quantizer.internal = False
self.data = self._quantizer.quantize(tensor)
```

transformer_engine/pytorch/module/base.py
```diff
@@ -1242,7 +1242,7 @@ def get_weight_workspace(
 
         # FP8 primary weights
         if isinstance(tensor, QuantizedTensor):
-            if update_workspace and quantizer is not None:
+            if update_workspace and quantizer is not None and quantizer.columnwise_usage:
                 tensor.update_usage(
                     rowwise_usage=quantizer.rowwise_usage,
                     columnwise_usage=quantizer.columnwise_usage,
```
timmoon10, line 1248: This change makes the `update_usage` logic redundant, so we might as well remove it:
```suggestion
```

That leads us to a more fundamental challenge: inference and validation have different data requirements, and we don't really expose a good way to distinguish between them. If you're doing MXFP8 inference, you don't want MXFP8 column-wise data since that's wasted memory. If you're loading a BF16 checkpoint, doing MXFP8 validation, and then doing MXFP8 training, you need to cast to MXFP8 row-wise and column-wise data when loading the checkpoint. Removing this `update_usage` logic has the effect of "however you've initialized the weights, you're stuck with those buffers forever". This is unfortunate for inference because the default behavior is to initialize weights for training, i.e. with both row-wise and column-wise data (https://github.com/NVIDIA/TransformerEngine/pull/1827 does add an option to initialize for inference).

I think removing this logic does make sense eventually, especially when Mcore adds support for https://github.com/NVIDIA/TransformerEngine/pull/1827, but for now I'm wary of breaking users doing inference. For now, is it possible to fix the bug by configuring distopt to avoid overlapping the param AG with validation?

WanZzzzzz: Thanks! Why do you think this bug has to do with overlapping the param AG with validation? I thought as long as we do training after validation, this error will occur.

timmoon10: Ah, you're right. I'll incorporate this logic in https://github.com/NVIDIA/TransformerEngine/pull/1827.

timmoon10:
I've included these fixes in https://github.com/NVIDIA/TransformerEngine/pull/1827.

PR #1835, opened 2025-05-30
Use versioned flavor of get driver entrypoint function

ptrendx:
# Description

Fixes the issue with cuStreamGetCtx pointing to cuStreamCtx_v2 in the CUDA 13 drivers.

Summary (mostly) by copilot:
This pull request updates the `transformer_engine/common/util/cuda_driver.cpp` and `transformer_engine/common/util/cuda_driver.h` files to enhance compatibility with different CUDA versions. The changes introduce a mechanism to query driver entry points based on the CUDA version, improving flexibility in handling CUDA driver symbols.

Enhancements for CUDA version compatibility:

* **Updated `get_symbol` function in `transformer_engine/common/util/cuda_driver.cpp`:** Refactored the function to support querying driver entry points using either a versioned or non-versioned mechanism. The function now accepts a `cuda_version` parameter and dynamically resolves the appropriate entry point function (`cudaGetDriverEntryPoint` or `cudaGetDriverEntryPointByVersion`).
* **Modified `get_symbol` function declaration in `transformer_engine/common/util/cuda_driver.h`:** Added an optional `cuda_version` parameter with a default value of `12010` (our oldest supported version) to allow backward compatibility while enabling version-specific queries.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/common/util/cuda_driver.cpp
+++ b/transformer_engine/common/util/cuda_driver.cpp
@@ -14,9 +14,9 @@ namespace transformer_engine {
 namespace cuda_driver {
 
 typedef cudaError_t (*VersionedGetEntryPoint)(const char *, void **, unsigned int,
-                                              unsigned long long,
+                                              unsigned long long,  // NOLINT(*)
                                               cudaDriverEntryPointQueryResult *);
-typedef cudaError_t (*GetEntryPoint)(const char *, void **, unsigned long long,
+typedef cudaError_t (*GetEntryPoint)(const char *, void **, unsigned long long,  // NOLINT(*)
                                      cudaDriverEntryPointQueryResult *);
 
 void *get_symbol(const char *symbol, int cuda_version) {


timmoon10 (approved):
LGTM

flx42:
I verified that it works!

ptrendx:
@timmoon10 sorry for the churn, I was looking at how others dealt with this issue and found this issue from cutlass: https://github.com/NVIDIA/cutlass/issues/2079 - since we, just like them, link against libcudart.so.12, the check for CUDA 12.5 during the compilation is not enough and we actually need to dynamically load the symbols. Fortunately, since we already link against libcudart, we don't need to try to find the lib by name (so at least there is that). @flx42 Could you verify this new version?

flx42:
> @timmoon10 sorry for the churn, I was looking at how others dealt with this issue and found this issue from cutlass: [NVIDIA/cutlass#2079](https://github.com/NVIDIA/cutlass/issues/2079) - since we, just like them, link against libcudart.so.12, the check for CUDA 12.5 during the compilation is not enough and we actually need to dynamically load the symbols. Fortunately, since we already link against libcudart, we don't need to try to find the lib by name (so at least there is that). @flx42 Could you verify this new version?

Still works fine!

timmoon10 (approved):

PR #1838, opened 2025-06-02
[JAX] Fix OTYPE for FP8 GEMM

phu0ngng:
# Description

Verified that the `ComputeType = FP32` regardless of the `element_type` thus set the `element_type = dq_dtype`.

TODO: Verifying if the dequantize happens in the epilogue of the Matmul, thus requiring no additional scaling ops, via nsys.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -43,7 +43,7 @@ jobs:
         run: |
           apt-get update
           apt-get install -y git python3.9 pip ninja-build cudnn9-cuda-12
-          pip install cmake torch pydantic importlib-metadata>=1.0 packaging pybind11
+          pip install cmake torch pydantic importlib-metadata>=1.0 packaging pybind11 numpy einops
       - name: 'Checkout'
         uses: actions/checkout@v3
         with:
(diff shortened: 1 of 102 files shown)


jberchtold-nvidia:
LGTM pending CI failures. The L0_jax-unittest seems like an infra failure as it mentions "No visible GPUs", the L1_jax_distributed_unittest is the issue on the nightlies that I'm investigating. `L0_jax_distributed_unittest` does seem relevant to this PR, slightly higher loss on an encoder test which I think would make sense with less precision in the GEMMs, so might just need a threshold update. Thanks for the PR!

phu0ngng:
Actually, it should not affect the encoder test loss as the GEMM is still performed in FP32. 
Turned to Draft for further investigation.

jberchtold-nvidia:
> Actually, it should not affect the encoder test loss as the GEMM is still performed in FP32. Turned to Draft for further investigation.

The internal compute dtype of the gemm kernel is FP32, but in a distributed gemm with TP, will the sums of partial results be in BF16 now instead of FP32?

jberchtold-nvidia (dismissed):

PR #1844, opened 2025-06-03
[JAX] TensorUsage + FP8 GEMM with all layouts handling on BW

phu0ngng:
# Description

TensorUsage + FP8 GEMM with all layouts handling on BW.

Verified that no NT enforcements by JAX, i.e. no additional transpose.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/jax/quantize/quantizer.py
+++ b/transformer_engine/jax/quantize/quantizer.py
@@ -907,7 +907,7 @@ class QuantizerFactory:
                 is_2x2x = True
             elif scaling_mode.is_tensor_scaling():
                 is_2x2x = not is_fp8_gemm_with_all_layouts_supported()
-            else:   # NO_SCALING ignores is_2x2x for now
+            else:  # NO_SCALING ignores is_2x2x for now
                 is_2x2x = False
         is_inference_mode = QuantizeConfig.INFERENCE_MODE
         assert not is_inference_mode, "Inference mode is not supported yet!"


transformer_engine/jax/quantize/device_utils.py
```diff
@@ -0,0 +1,34 @@
...
+    Get the compute capability of the device.
+    """
+    return transformer_engine_jax.get_device_compute_capability(gpu_id)
+
+
+@functools.lru_cache(maxsize=None)
+def is_gemm_with_all_layouts_supported() -> bool:
+    """Return True if using Blackwell architecture, False otherwise."""
+    compute_capability = get_device_compute_capability()
+    return 100 <= compute_capability <= 120
```
denera: I believe this should be `(100 <= compute capability < 120) and (compute capability >= 130)` instead. Blackwell GeForce series cards (12.x) are still restricted to TN layout (or NT, depending on the column vs. row-major perspective) just like Hopper.

phu0ngng: You are right. Will update the < 120. 
I did not add the 130 requirements as I'm not sure if we are supposed to talk about 130 yet.

denera: The TE/PyTorch equivalent of this check includes 130, so I assumed it's okay to add it. @ptrendx can you clarify?

transformer_engine/jax/layernorm_dense.py
```diff
@@ -209,8 +210,8 @@ def _layernorm_dense_fwd_rule(
         output += jnp.reshape(bias, bias_new_shape)
 
     ctx = (
-        casted_ln_out.get_colwise_tensor() if quantizer_set.x.is_2x2x() else None,
-        casted_kernel.get_rowwise_tensor() if quantizer_set.kernel.is_2x2x() else None,
+        casted_ln_out.get_tensor(TensorUsage.LHS_TRANS),
```
jberchtold-nvidia, line 213: Should we keep the check for whether we're training or not? So it'd be something like this
```
casted_ln_out.get_tensor(TensorUsage.LHS_TRANS) if quantizer_set.x.is_2x2x() else None
```
because on Hopper this layout wouldn't exist when doing 1x for inference, right?

transformer_engine/jax/quantize/scaling_modes.py
```diff
@@ -13,17 +13,44 @@
...
+    "QuantizeShardyRules",
+    "ScalingMode",
+    "TensorUsage",
+]
+
+
+class TensorUsage(Enum):
+    """Enum for tensor usage.
+
+    This enum for the usage of a tensor.
```
jberchtold-nvidia: Can this docstring be elaborated slightly to indicate that "usage" here means usage in GEMMs (LHS/RHS)?

jberchtold-nvidia (commented):
Overall looks good, some small comments. I like the TensorUsage concept

transformer_engine/jax/layernorm_dense.py
phu0ngng: No, we don't need to do that here.
Whether it's training or not should be handled in the `get_tensor()` method, and currently, we don't support it yet.

jberchtold-nvidia: Ah I see, that will make it easier if it's automatic. So the idea is based on whether it's x or kernel, we know which usage (LHS/RHS) will be used in the forward and the backward, so we can automatically know to remove it when doing inference (not currently, but we have all the required info to support it in future)?

jberchtold-nvidia (dismissed):

phu0ngng:
Adding @huanghua1994 to review the changes in `grouped_gemm`.

jberchtold-nvidia (dismissed):
LGTM

denera (approved):

PR #1847, opened 2025-06-04
[PyTorch] Inference mode disables initializing quantized weights with column-wise usage

timmoon10:
# Description

When initializing a model with quantized weights, the required data is different for training and inference (training requires row-wise data for forward GEMM and column-wise data for dgrad GEMM, inference only requires column-wise). This PR adds logic so the model will only initialize quantized weights with the data required for inference when initialized within [no-grad mode](https://docs.pytorch.org/docs/stable/notes/autograd.html#no-grad-mode) or [inference mode](https://docs.pytorch.org/docs/stable/notes/autograd.html#inference-mode). It is also less aggressive about deallocating weight data in order to handle cases where we were alternating between training and validation modes.

This is an alternative to https://github.com/NVIDIA/TransformerEngine/pull/1827. The heuristic API in that PR are somewhat redundant with these plain PyTorch APIs, but it is also more general.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Avoid initializing quantized weights with column-wise usage if grads are not enabled
- Do not deallocate unnecessary usages in weight tensors during forward pass

# Checklist:

### DIFF
--- a/tests/pytorch/test_sanity.py
+++ b/tests/pytorch/test_sanity.py
@@ -1379,7 +1379,7 @@ def test_inference_mode(
 
     # Construct module
     module = None
-    with torch.inference_mode():
+    with torch.no_grad():
         with fp8_model_init(enabled=with_quantization, recipe=quantization_recipe):
             if module_name == "Linear":
                 module = Linear(hidden_size, hidden_size)
--- a/tests/pytorch/test_sanity.py
+++ b/tests/pytorch/test_sanity.py
@@ -1415,5 +1415,6 @@ def test_inference_mode(
         kwargs = {}
         if module_name == "GroupedLinear":
             kwargs["m_splits"] = [sequence_length]
-        y = module(x, **kwargs)
+        with fp8_autocast(enabled=with_quantization, fp8_recipe=quantization_recipe):
+            y = module(x, **kwargs)
     check_weights()
--- a/tests/pytorch/test_sanity.py
+++ b/tests/pytorch/test_sanity.py
@@ -10,6 +10,7 @@ import torch
 import pytest
 import os
 
+import transformer_engine.pytorch
 from transformer_engine.pytorch.fp8 import (
     fp8_autocast,
     FP8GlobalStateManager,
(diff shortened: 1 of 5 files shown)


tests/pytorch/test_sanity.py
```diff
@@ -10,6 +10,7 @@
 import pytest
 import os
 
+import transformer_engine.pytorch
```
ksivaman, line 13: Why is this needed?

tests/pytorch/test_sanity.py
```diff
@@ -1338,3 +1341,80 @@ def backward(ctx, grad_output):
...
+    if quantization == "fp8_delayed_scaling":
+        quantization_recipe = recipe.DelayedScaling()
+    elif quantization == "fp8_current_scaling":
+        quantization_recipe = recipe.Float8CurrentScaling()
+    elif quantization == "mxfp8":
+        quantization_recipe = recipe.MXFP8BlockScaling()
+
+    # Construct module
+    module = None
+    with torch.inference_mode():
```
ksivaman: This is non-standard usage for `torch.inference_mode`, and in this PR we don't seem to be doing anything inference specific? The `torch.no_grad` context feels like the more appropriate choice here.

ksivaman:
```suggestion
    with torch.no_grad():
```

tests/pytorch/test_sanity.py
timmoon10: I find it convenient to be able to access a class without explicitly doing `from ... import ...`:
https://github.com/NVIDIA/TransformerEngine/blob/649b04c8a99c9fc2c432af3313030700fe0333f8/tests/pytorch/test_sanity.py#L1393

It's just a matter of style though. Within the package we explicitly list the imports to order to guarantee only relative imports, but this isn't relevant for tests since we always do absolute imports. Also, [Google's style guide recommends against it](https://google.github.io/styleguide/pyguide.html#22-imports).

tests/pytorch/test_sanity.py
timmoon10: Fair enough, `inference_mode` primarily affects the C++ autograd infrastructure and it doesn't expose much externally.

ksivaman (approved):
LGTM

PR #1849, opened 2025-06-04
Add support for overlapping wgrad NCCL AG with dgrad GEMM

djns99:
# Description

Adds support for overlapping the wgrad All Gather operation with dgrad GEMM

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add new API to UB object for obtaining reference to communication stream
- Update *Linear layers to start wgrad allgather after dgrad communication completes

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/linear.py
+++ b/transformer_engine/pytorch/module/linear.py
@@ -692,8 +692,11 @@ class _Linear(torch.autograd.Function):
                     # for the dgrad GEMM. We work around by explicitly
                     # overlapping the NCCL operation with the dgrad GEMM.
                     ctx.grad_output_quantizer.set_usage(rowwise=False, columnwise=True)
-                    s = ub_obj_dgrad.get_communication_stream()
-                    with torch.cuda.stream(s):
+                    # Get the communication stream from the dgrad GEMM and set it as the current torch stream
+                    dgrad_comm_stream = ub_obj_dgrad.get_communication_stream()
+                    with torch.cuda.stream(dgrad_comm_stream):
+                        # Syncs with the current stream (dgrad_comm_stream) before starting the all-gather
+                        # This ensures that we don't start until all communication for the dgrad GEMM is complete
                         grad_output, grad_output_work = gather_along_first_dim(
                             grad_output_arg,
                             ctx.tp_group,
--- a/transformer_engine/pytorch/module/layernorm_linear.py
+++ b/transformer_engine/pytorch/module/layernorm_linear.py
@@ -754,8 +754,12 @@ class _LayerNormLinear(torch.autograd.Function):
                     # for the dgrad GEMM. We work around by explicitly
                     # overlapping the NCCL operation with the dgrad GEMM.
                     ctx.grad_output_quantizer.set_usage(rowwise=False, columnwise=True)
-                    s = ub_obj_dgrad.get_communication_stream()
-                    with torch.cuda.stream(s):
+
+                    # Get the communication stream from the dgrad GEMM and set it as the current torch stream
+                    dgrad_comm_stream = ub_obj_dgrad.get_communication_stream()
+                    with torch.cuda.stream(dgrad_comm_stream):
+                        # Syncs with the current stream (dgrad_comm_stream) before starting the all-gather
+                        # This ensures that we don't start until all communication for the dgrad GEMM is complete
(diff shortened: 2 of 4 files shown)


transformer_engine/pytorch/module/linear.py
```diff
@@ -695,14 +695,20 @@ def backward(ctx, grad_output: torch.Tensor) -> Tuple[Union[torch.Tensor, None],
...
-                        quantizer=ctx.grad_output_quantizer,
-                    )
+                    s = ub_obj_dgrad.get_communication_stream()
+                    with torch.cuda.stream(s):
+                        grad_output, grad_output_work = gather_along_first_dim(
+                            grad_output_arg,
+                            ctx.tp_group,
+                            async_op=True,
+                            quantizer=ctx.grad_output_quantizer,
+                        )
```
erhoo82, line 705: Since `async_op` uses a pre-allocated torch stream assigned for the TP communication group, grepping the UB communication stream and using it as a context is pretty much a no-op.

Also, the async communication in `gather_along_first_dim` would do synchronization with torh main stream, right? Then, we cannot pull the communication start timeline before the last GEMM chunk.

djns99: > grepping the UB communication stream and using it as a context is pretty much a no-op.

Ok maybe I am misunderstanding, but my understanding is the point of this is that the async_op creates an cudaEvent dependency between the current stream and the preallocated stream. By using `with` block it changes which stream it syncs with. As you say without this it would sync with the main stream instead and not overlap at all.

timmoon10: The stream synchronization logic seems correct to me. We could accomplish the same thing with:
```python
with torch.cuda.stream(s):
    grad_output, _ = gather_along_first_dim(grad_output_arg, ctx.tp_group, quantizer=ctx.grad_output_quantizer)
torch.cuda.current_stream().wait_stream(s)
```

We are effectively serializing the UB and NCCL communication, which shouldn't have an effect if we are fully utilizing bandwidth. This also avoids some correctness issues I saw in https://github.com/NVIDIA/TransformerEngine/pull/1711 when I had simultaneous UB and NCCL communication (see https://github.com/NVIDIA/TransformerEngine/pull/1711/commits/8e63f8cc21e8daf937ad84b80388dcea7f136650).

setup.py
```diff
@@ -72,6 +72,11 @@ def setup_common_extension() -> CMakeExtension:
     if bool(int(os.getenv("NVTE_BUILD_ACTIVATION_WITH_FAST_MATH", "0"))):
         cmake_flags.append("-DNVTE_BUILD_ACTIVATION_WITH_FAST_MATH=ON")
 
+    # Add custom CMake arguments from environment variable
+    nvte_cmake_extra_args = os.getenv("NVTE_CMAKE_EXTRA_ARGS")
+    if nvte_cmake_extra_args:
+        cmake_flags.extend(nvte_cmake_extra_args.split())
+
```
timmoon10, line 73: What CMake args did you need for your build?
```suggestion
```

timmoon10 (approved):
LGTM

setup.py
djns99: I like adding CCACHE as it speeds up recompiling if I have to switch between branches. Any reason we can't keep this? I guess it doesnt belong in this PR but

transformer_engine/pytorch/module/linear.py
djns99: > We could accomplish the same thing with:
> ```with torch.cuda.stream(s):
>    grad_output, _ = gather_along_first_dim(grad_output_arg, ctx.tp_group, quantizer=ctx.grad_output_quantizer)
> torch.cuda.current_stream().wait_stream(s)```

@timmoon10 @erhoo82 do you have a preference. Technically, the current version has one less cuda event synchronize but that should not impact perf so we can pick whichever you think is more readable.
The only real advantage of the current one is we can easily get more overlap if we add any kernels between the AG and the GEMM  (by moving the `.wait()` later)

timmoon10: I think the current impl is fine. I was originally suspicious from seeing a `wait` immediately after the AG, but looking again it makes sense to me.

setup.py
timmoon10: I think it's fine as an advanced option. It's pretty self-contained.

timmoon10 (approved):

PR #1852, opened 2025-06-05
pyproject.toml

ksivaman:
# Description

Move to a `pyproject.toml` setup for build ([pep517](https://peps.python.org/pep-0517/)).

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Use `pypyproject.toml` setup for wheel and source builds.
- Remove `nvidia-*` libs as dependencies. These are currently not being used and will be added to build dependencies in a future PR supporting full e2e compilation and usage without CUDA toolkit.

# Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@
 # See LICENSE for license information.
 
 [build-system]
-requires = ["setuptools>=61.0", "cmake>=3.21", "wheel", "pybind11[global]", "ninja", "pip"]
+requires = ["setuptools>=61.0", "cmake>=3.21", "wheel", "pybind11[global]", "ninja", "pip", "torch>=2.1", "jax[cuda12]", "flax>=0.7.1"]
 
 # Use legacy backend to import local packages in setup.py
 build-backend = "setuptools.build_meta:__legacy__"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,10 @@
+# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+[build-system]
+requires = ["setuptools>=61.0", "cmake>=3.21", "wheel", "pybind11[global]", "ninja", "pip"]
+
+# Use legacy backend to import local packages in setup.py
+build-backend = "setuptools.build_meta:__legacy__"
+
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,6 @@ from build_tools.utils import (
     found_ninja,
     found_pybind11,
     get_frameworks,
-    install_and_import,
     remove_dups,
     cuda_toolkit_include_path,
 )
@@ -36,7 +35,6 @@ os.environ["NVTE_PROJECT_BUILDING"] = "1"
 if "pytorch" in frameworks:
     from torch.utils.cpp_extension import BuildExtension
 elif "jax" in frameworks:
-    install_and_import("pybind11[global]")
     from pybind11.setup_helpers import build_ext as BuildExtension
 
 
--- a/setup.py
+++ b/setup.py
@@ -80,26 +80,13 @@ def setup_common_extension() -> CMakeExtension:
     )
 
 
-def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
+def setup_requirements() -> Tuple[List[str], List[str]]:
     """Setup Python dependencies
 
     Returns dependencies for build, runtime, and testing.
     """
 
     # Common requirements
-    setup_reqs: List[str] = []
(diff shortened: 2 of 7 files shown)


pyproject.toml
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+[build-system]
+requires = ["setuptools>=61.0", "cmake>=3.21", "wheel", "pybind11[global]", "ninja", "pip", "torch>=2.1", "jax[cuda12]", "flax>=0.7.1"]
```
timmoon10, line 6: Does this mean PyTorch is a dependency for JAX-only builds, and vice versa? That doesn't seem right.

One problem with the pyproject.toml approach is that build-time dependencies are static, but we would really like to have dynamic dependencies. Should we migrate away from supporting a monolithic TE package and commit to using framework-specific subpackages? The TE core package only has runtime dependencies on the framework subpackages, and runtime dependencies can be dynamic.

setup.py
```diff
@@ -82,45 +76,23 @@ def setup_common_extension() -> CMakeExtension:
     )
 
 
-def setup_requirements() -> Tuple[List[str], List[str], List[str]]:
+def setup_requirements() -> Tuple[List[str], List[str]]:
     """Setup Python dependencies
 
     Returns dependencies for build, runtime, and testing.
```
timmoon10:
```suggestion
    Returns dependencies for runtime and testing.
```

.github/workflows/build.yml
```diff
@@ -19,7 +19,7 @@ jobs:
         run: |
           apt-get update
           apt-get install -y git python3.9 pip ninja-build cudnn9-cuda-12
-          pip install cmake==3.21.0
+          pip install cmake==3.21.0 pybind11[global]
```
timmoon10: Now that Ninja is a Python dependency, we should install with Pip instead of apt.
```suggestion
          apt-get install -y git python3.9 pip cudnn9-cuda-12
          pip install cmake==3.21.0 ninja-build pybind11[global]
```
We may also want to consider using the cuDNN Python package as well.

pyproject.toml
ksivaman: A couple of important points to note:

1. These static build dependencies are only relevant when using build isolation, which is to be viewed purely as an extra feature on top of the current recommendation of using `--no-build-isolation`.
2. This is also only the case for source installations. As installations from PyPI circumvents this by having a wheel for the core build and specific dependencies from the framework packages.

Feature 1 will help users who want to be able to do a simple build without installing any dependencies beforehand and in that case they will get both frameworks. The **_main_** issue that I see with having the framework packages completely separate even for a source install is that we won't be able to support the workflow of doing a simple `pip install` from the top level directory.

.github/workflows/build.yml
ksivaman: cuDNN python package support will be a part of a separate effort to remove dependency on toolkit.

pyproject.toml
timmoon10: I think requiring `--no-build-isolation` is fine for now (I wonder if there's a way to error out if it's not set), but doesn't this mean you also need `--no-deps` for the source build? Or else you might install PyTorch/JAX unnecessarily. Maybe better to assume that the user has PyTorch/JAX in their environment.

ksivaman: > doesn't this mean you also need --no-deps for the source build?

No, because of the distinction between build vs runtime dependencies. `--no-build-isolation` is strictly for build time dependencies, whereas `--no-deps` is for runtime dependencies. With `--no-build-isolation` (the current recommended install method), the user must preinstall all needed build deps for the necessary framework extension they want to compile (include the fw itself). Few cases:

1. If they only have `torch` installed then the Jax extensions won't be built and the install deps for TE/Jax will not be installed, and vice-versa.
2. If the user has both frameworks but doesn't set `NVTE_FRAMEWORK` then everything will be built and all runtime deps will be installed.
3. If the user has both frameworks and sets `NVTE_FRAMEWORK` then only that framework's extension will be built and thus only it's runtime deps will be installed.

> I wonder if there's a way to error out if it's not set

I think this is not needed because:

1. We updated all installation instructions and docs in `v2.3` to ensure that `--no-build-isolation` is the default everywhere.
2. This PR does add support for build isolation, purely as an new feature or fallback. Worst case the user env has a few extra libs/dependencies installed that they might not need, but this is a good trade-off for the basic user that doesn't want to manage deps and worry about the env. This will be powerful especially once we remove the toolkit restriction because then we will not longer have any system package dependency and user can simply `pip install .`.

timmoon10 (approved):
LGTM

PR #1858, opened 2025-06-06
[PyTorch] Add support for FP8 current scaling in operation-based API

timmoon10:
# Description

This PR makes some minor changes needed for te.Sequential to support FP8 current scaling. It also modifies the te.Sequential unit tests to more gracefully handle quantization schemes other than FP8 delayed scaling.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring
- [x] Testing

## Changes

- Support FP8 current scaling in operation-based API
- Generalize support for quantizers in tests for operation-based API

# Checklist:

### DIFF
--- a/tests/pytorch/distributed/test_fusible_ops.py
+++ b/tests/pytorch/distributed/test_fusible_ops.py
@@ -22,11 +22,16 @@ import transformer_engine.common.recipe
 import transformer_engine.pytorch as te
 from transformer_engine.pytorch.fp8 import FP8GlobalStateManager
 from transformer_engine.pytorch.tensor import QuantizedTensor
-from transformer_engine.pytorch.tensor.float8_tensor import Float8Quantizer
+from transformer_engine.pytorch.tensor.float8_tensor import Float8Quantizer, Float8CurrentScalingQuantizer
 import transformer_engine.pytorch.ops as te_ops
 from transformer_engine.pytorch.ops._common import is_float8_tensor
 from transformer_engine.pytorch.utils import is_bf16_compatible
 import transformer_engine_torch as tex
+
+# Import utility functions
+_current_file = pathlib.Path(__file__).resolve()
+sys.path.append(str(_current_file.parent.parent))
+from utils import dtype_tols, make_recipe
 
 
 # Check what quantization schemes are supported
@@ -34,7 +39,7 @@ fp8_available, reason_for_no_fp8 = FP8GlobalStateManager.is_fp8_available()
 mxfp8_available, reason_for_no_mxfp8 = FP8GlobalStateManager.is_mxfp8_available()
 quantization_list: list[Optional[str]] = [None]
 if fp8_available:
-    quantization_list.append("fp8")
+    quantization_list.extend(("fp8_delayed_scaling", "fp8_current_scaling"))
 if mxfp8_available:
     quantization_list.append("mxfp8")
 
@@ -63,11 +68,12 @@ def reset_rng(seed: int = 1234) -> None:
 @torch.no_grad()
 def make_reference_and_test_tensors(
     shape: int | Iterable[int],
+    quantization: Optional[str] = None,
     ref_dtype: torch.dtype = torch.float64,
     ref_device: torch.device = "cpu",
     test_dtype: torch.dtype = torch.float32,
     test_device: torch.device = "cuda",
-    test_is_fp8: bool = False,
+    test_is_quantized: bool = False,
     requires_grad: bool = True,
 ) -> tuple[torch.Tensor, torch.Tensor]:
     """Construct tensors with the same values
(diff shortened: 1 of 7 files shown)


transformer_engine/pytorch/ops/fused/userbuffers_forward_linear.py
```diff
@@ -208,7 +208,9 @@ def _functional_forward(
             if input_quantizer is not None:
                 if not isinstance(x_local, QuantizedTensorBase):
                     input_quantizer.set_usage(rowwise=True, columnwise=weight_requires_grad)
-                    if isinstance(input_quantizer, Float8Quantizer):
+                    if isinstance(
+                        input_quantizer, (Float8Quantizer, Float8CurrentScalingQuantizer)
+                    ):
```
ksivaman, line 213:
```suggestion
                    if input_quantizer._get_compatible_recipe().float8_per_tensor_scaling():
```

transformer_engine/pytorch/ops/basic/basic_linear.py
```diff
@@ -324,12 +324,38 @@ def pre_forward(self, *args, **kwargs) -> None:
...
+
             # Make sure weight tensor has correct quantizer
             # Note: Quantizer might have changed if quantization
             # recipe changed
-            if isinstance(weight_quantizer, Float8Quantizer) and isinstance(
-                weight, Float8TensorBase
-            ):
+            if isinstance(
+                weight_quantizer, (Float8Quantizer, Float8CurrentScalingQuantizer)
+            ) and isinstance(weight, Float8TensorBase):
```
ksivaman, line 358:
```suggestion
            if weight_quantizer._get_compatible_recipe().float8_per_tensor_scaling() and
               isinstance(weight, Float8TensorBase):
```

transformer_engine/pytorch/ops/basic/basic_linear.py
```diff
@@ -324,12 +324,38 @@ def pre_forward(self, *args, **kwargs) -> None:
             weight_quantizer.set_usage(rowwise=True, columnwise=is_grad_enabled)
             grad_output_quantizer.set_usage(rowwise=True, columnwise=weight_requires_grad)
 
+            # Recipe-specific configuration
+            recipe = FP8GlobalStateManager.get_fp8_recipe()
+            if recipe.float8_current_scaling():
```
ksivaman, line 329: **Note**: This being required here is unfortunate. Most of what I would suggest here are probably out of scope for this PR.

tests/pytorch/distributed/test_fusible_ops.py
```diff
@@ -76,78 +85,55 @@ def make_reference_and_test_tensors(
...
             amax=torch.zeros(1, dtype=torch.float32, device=test_device),
             fp8_dtype=tex.DType.kFloat8E4M3,
         )
         test = quantizer(test)
-    elif test.data_ptr() == ref.data_ptr():
-        test = test.clone()
+    elif quantization == "fp8_current_scaling":
+        quantizer = Float8CurrentScalingQuantizer(
+            fp8_dtype=tex.DType.kFloat8E4M3,
+            device=test_device,
```
ksivaman, line 114: Orthogonal but why does this device arg exist here?

tests/pytorch/test_fusible_ops.py
```diff
@@ -24,10 +26,20 @@
...
+    Float8CurrentScalingQuantizer,
+    Float8Quantizer,
+)
+from transformer_engine.pytorch.tensor.mxfp8_tensor import MXFP8Tensor
 from transformer_engine.pytorch.utils import is_bf16_compatible
 import transformer_engine_torch as tex
 
+# Import utility functions
+_current_file = pathlib.Path(__file__).resolve()
+sys.path.append(str(_current_file.parent))
```
ksivaman, line 40: Not a fan of this. But I see we currently do this for other tests as well so for this PR it's ok. In a separate refactor we should convert the `tests` dir to also be a module.

transformer_engine/pytorch/ops/basic/basic_linear.py
timmoon10: I don't think there's a good way around this. The amax reduction logic is specific to linear layers with FP8 current scaling, so there's no other logical place to put it.

tests/pytorch/distributed/test_fusible_ops.py
timmoon10: Mostly for completeness. We have a `ref_device` option since I prefer computing the reference impl on CPU, which helps catch CUDA-related bugs.

tests/pytorch/test_fusible_ops.py
timmoon10: I agree. Ideally we would use relative imports like `from .utils import dtype_tols`, but Python doesn't allow that if you are running the script directly, which is how Pytest runs the tests.

transformer_engine/pytorch/ops/fused/userbuffers_forward_linear.py
timmoon10: I think this fix adds overhead and makes the code less clear. `_get_compatible_recipe` (added in https://github.com/NVIDIA/TransformerEngine/pull/1724) returns a type since it's used for type comparisons. If we want to use `float8_per_tensor_scaling`, then we need to instantiate a new `DelayedScaling` or `Float8CurrentScaling` that doesn't match the current recipe. Alternatively, we can compare classes with `is` or `isinstance`, which is basically what we're already doing.

transformer_engine/pytorch/ops/basic/basic_linear.py
timmoon10: I don't think `_get_compatible_recipe` benefits us here.

ksivaman (approved):
LGTM

PR #1864, opened 2025-06-10
[PyTorch] Support L2Norm basic op and module -> use for qk_norm

negvet:
# Description

Support `L2Norm` basic op and module.
Enable it for qk normalization via `use_qk_norm`.
Exposure to MHA and transformer block modules.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Support `L2Norm` basic op and module
- Support qk_norm

# Checklist:


transformer_engine/pytorch/module/l2norm.py
timmoon10, line 1: Do we need to have a separate implementation here? We have separate LayerNorm and RMSNorm modules for backward compatibility, so no design reason to copy that here.

transformer_engine/pytorch/module/l2norm.py
```diff
@@ -0,0 +1,71 @@
...
+        # JIT warmup for L2Norm fused operations
+        if seq_length and micro_batch_size:
+            device = getattr(self, "device", torch.device("cuda"))
+            if hasattr(device, "type") and device.type == "cuda":
+                set_jit_fusion_options()
+                # For L2Norm, we don't know the hidden size until forward pass,
+                # but we can warm up with common sizes used in attention mechanisms
+                common_hidden_sizes = [768, 1024, 2048, 4096, 8192]
+                for hidden_size in common_hidden_sizes:
+                    warmup_jit_l2norm_all_dtypes(hidden_size, seq_length, micro_batch_size)
```
timmoon10: Why not JIT lazily during the first forward?

transformer_engine/pytorch/ops/basic/l2norm.py
```diff
@@ -0,0 +1,147 @@
...
+    canonicalize_device,
+    canonicalize_dtype,
+    clear_tensor_data,
+)
+from ..op import BasicOperation, OperationContext
+from .._common import maybe_autocast_dtype, reshape
+from ...jit import l2norm_fused, l2norm_fwd_fused, l2norm_backward_fused
+
+
+class L2Norm(BasicOperation):
```
timmoon10: This name makes me think you are actually computing the L2 norm, like [`torch.linalg.norm`](https://docs.pytorch.org/docs/stable/generated/torch.linalg.norm.html#torch.linalg.norm). What do other implementations call it? If there are no prominent implementations, we could consider calling it `L2Normalization` to avoid being ambiguous.

transformer_engine/pytorch/ops/basic/l2norm.py
```diff
@@ -0,0 +1,147 @@
...
+        dtype: Optional[torch.dtype] = None,
+    ) -> None:
+        super().__init__()
+        self.eps: float = eps
+        self.device = canonicalize_device(device)
+        self.dtype = canonicalize_dtype(dtype)
+
+    def reset_parameters(self) -> None:
+        """L2Norm has no parameters to reset"""
+
```
timmoon10: Not needed:
```suggestion
```

transformer_engine/pytorch/ops/basic/l2norm.py
```diff
@@ -0,0 +1,147 @@
...
+        self,
+        *,
+        eps: float = 1e-6,
+        device: Optional[torch.device | str] = None,
+        dtype: Optional[torch.dtype] = None,
+    ) -> None:
+        super().__init__()
+        self.eps: float = eps
+        self.device = canonicalize_device(device)
+        self.dtype = canonicalize_dtype(dtype)
```
timmoon10: Do we need this? For LayerNorm/RMSNorm, we need to coerce the data so they are compatible with cuDNN or our custom kernels. However, `torch.compile` should be more general and should accommodate CPU data or mixed dtypes.

transformer_engine/pytorch/ops/basic/l2norm.py
```diff
@@ -0,0 +1,147 @@
...
+        device = self.device
+        if device.type != "cuda":
+            device = canonicalize_device(None)
+        dtype = maybe_autocast_dtype(default_dtype=self.dtype or input_.dtype)
+
+        # Reshape input for computation
+        input_dims = tuple(input_.size())
+        inner_dim = input_dims[-1]
+        x = reshape(input_, (-1, inner_dim), device=device, dtype=dtype)
+
```
timmoon10: For LayerNorm/RMSNorm, we reshape to 2D due to limitations in our custom kernels. However, the `torch.compile` implementation can gracefully handle multi-dimensional tensors.
```suggestion
        x = input_
```

transformer_engine/pytorch/ops/basic/l2norm.py
```diff
@@ -0,0 +1,147 @@
...
+        # Check if output is quantized
+        output_quantizer = None
+        if (
+            FP8GlobalStateManager.is_fp8_enabled()
+            and next_op is not None
+            and next_op.num_quantizers("forward") > 0
+        ):
+            output_quantizer = next_op.get_quantizer("forward", 0)
+
```
timmoon10: There's no benefit from quantizing early since we don't have a fused kernel.
```suggestion
```

tests/pytorch/test_fusible_ops.py
```diff
@@ -1273,6 +1273,76 @@ def test_rmsnorm(
         torch.testing.assert_close(dx_test, x_ref.grad, **tols)
         torch.testing.assert_close(dw_test, w_ref.grad, **tols)
 
+    @pytest.mark.parametrize("in_shape", ((32,), (6, 16, 64), (32, 64)))
+    @pytest.mark.parametrize("dtype", _dtypes)
+    @pytest.mark.parametrize("quantization", (None, "fp8", "mxfp8"))
```
timmoon10: We don't have a fused kernel, so there's not much value in testing quantization.
```suggestion
```

tests/pytorch/test_fusible_ops.py
```diff
@@ -1273,6 +1273,76 @@ def test_rmsnorm(
...
+        # Expected numerical error
+        tols = dtype_tols(dtype)
+        if quantized_compute:
+            tols = dtype_tols(tex.DType.kFloat8E4M3)
+        else:
+            # L2Norm requires slightly looser tolerances
+            if dtype == torch.float16:
+                tols = dict(rtol=5e-3, atol=5e-4)
+            elif dtype == torch.bfloat16:
+                tols = dict(rtol=2e-2, atol=5e-3)
```
timmoon10: Why? Perhaps sqrt would hurt accuracy, but why don't we see problems with LayerNorm or RMSNorm?

transformer_engine/pytorch/module/l2norm.py
negvet: This is to achieve consistency of activation recomputation. If only compile during the first forward pass, the JIT compiler might generate different kernels when the same operation is called later during activation recomputation. 

@ksivaman could you please take a look and confirm that this issue will be solved with the current approach?

transformer_engine/pytorch/module/l2norm.py
negvet: Got it, removed this module implementation in https://github.com/NVIDIA/TransformerEngine/pull/1864/commits/aabeea28dd3338aff22b329c1d20fea85db6d90d

transformer_engine/pytorch/ops/basic/l2norm.py
negvet: Good point. My naming followed the one from llama4 in transformers - `Llama4TextL2Norm`.
But I agree that it sounds like a norm computation. In addition, we have `transformer_engine/common/multi_tensor/l2norm.cu`, which is a L2 norm computation. 
Renamed to `L2Normalization`.

transformer_engine/pytorch/ops/basic/l2norm.py
negvet: I see, removed it

transformer_engine/pytorch/ops/basic/l2norm.py
negvet: Fixed

transformer_engine/pytorch/ops/basic/l2norm.py
negvet: Fixed

tests/pytorch/test_fusible_ops.py
negvet: Fixed!

tests/pytorch/test_fusible_ops.py
negvet: Thanks for noticing.
I have narrowed it down, and it turns out that only for bf16 backprop I need to increase atol from 1.5e-3 to 2e-3.
It is hard to say for sure. sqrt should probably not be the reason for that. I would say that implementation quality might be the issue for jit-compiled generic operations. Probably intermediate bf16 precision can lead to that as well.

PR #1865, opened 2025-06-11
[PyTorch] Add save_original_input in Linear/GroupedLinear to save memory

hxbai:
# Description

Add save_original_input in Linear/GroupedLinear. If it is on, the modules save the original inputs and only do rowwise quant in forward and colwise quant in backward (for blockwise and mxfp8 recipe). With this, in some scenarios, we can save memory usage at the expense of separated quantizations. 

In the current implementation, the Linear/GroupedLinear always saves the casted/quantized input tensors for backward.
But this may increase the memory usage if this input tensor is used by multiple modules, such as:
- The output tensor of fused attn is saved by its self and used by the following proj_linear. We want the proj_linear to save the original input to save memory usage.
- For multi-latent attention in DeepSeek models, the output of input_layernorm is used by q_down_linear and kv_down_linear. We want the two linears to save the original input.
- In Megatron-LM, we added an output-discarding recomputation that trickily releases some saved tensors and restores them by recomputation. This relies on that the saved tensor is the original output tensor.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- transformer_engine/pytorch/module/linear.py
- transformer_engine/pytorch/module/grouped_linear.py

# Checklist:

### DIFF
--- a/tests/pytorch/test_numerics.py
+++ b/tests/pytorch/test_numerics.py
@@ -1055,8 +1055,11 @@ def test_mha_accuracy(dtype, bs, model, mask_type):
             assert_allclose(te_output, torch_output, atol[dtype], rtol[dtype])
 
 
-def _test_granular_accuracy(block, bs, dtype, config, delay_wgrad_compute=False):
+def _test_granular_accuracy(block, bs, dtype, config, delay_wgrad_compute=False, recipe=None):
     reset_rng_states()
+    fp8 = recipe is not None
+    if fp8:
+        FP8GlobalStateManager.reset()
 
     inp_hidden_states = torch.randn(
         (config.seq_len, bs, config.hidden_size),
@@ -1066,9 +1069,10 @@ def _test_granular_accuracy(block, bs, dtype, config, delay_wgrad_compute=False)
     )
     inp_hidden_states.retain_grad()
 
-    out = block(inp_hidden_states)
-    if isinstance(out, (List, Tuple)):
-        out = out[0]
+    with fp8_autocast(enabled=fp8, fp8_recipe=recipe):
+        out = block(inp_hidden_states)
+        if isinstance(out, (List, Tuple)):
+            out = out[0]
     loss = out.sum()
     loss.backward()
     if delay_wgrad_compute:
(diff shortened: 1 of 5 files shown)


hxbai:
Hi @timmoon10 could you help take a look at this PR? Thanks!

transformer_engine/pytorch/module/linear.py
```diff
@@ -556,6 +573,20 @@ def backward(ctx, grad_output: torch.Tensor) -> Tuple[Union[torch.Tensor, None],
...
+                        if isinstance(quantizer, (Float8Quantizer, Float8CurrentScalingQuantizer)):
+                            quantizer.set_usage(
+                                rowwise=True,
+                                columnwise=not ctx.backward_input_needs_gather,
+                            )
+                        else:
+                            quantizer.set_usage(rowwise=False, columnwise=True)
+                        inputmat = quantizer(inputmat)
+                else:
+                    inputmat = cast_if_needed(inputmat, ctx.activation_dtype)
```
timmoon10: We can figure out the right behavior by checking `inputmat`:
```suggestion
            if ctx.requires_wgrad:
                input_is_quantized = isinstance(inputmat, QuantizedTensorBase)
                if ctx.fp8:
                    if not input_is_quantized:
                        quantizer = ctx.input_quantizer
                        if (
                            isinstance(quantizer, (Float8Quantizer, Float8CurrentScalingQuantizer)
                            and ctx.backward_input_needs_gather
                        ):
                            quantizer.set_usage(rowwise=True, columnwise=False)
                        else:
                            quantizer.set_usage(rowwise=False, columnwise=True)
                        inputmat = quantizer(inputmat)
                else:
                    if input_is_quantized:
                        inputmat = inputmat.dequantize(dtype=ctx.activation_dtype)
                    else:
                        inputmat = cast_if_needed(inputmat, ctx.activation_dtype)
```
This can adapt to whatever weird thing the forward pass does (e.g. if it caches the input in FP8 but does an BF16 GEMM).

tests/pytorch/test_numerics.py
```diff
@@ -1264,6 +1268,68 @@ def test_linear_accuracy_delay_wgrad_compute(dtype, bs, model, bias, fuse_wgrad_
...
 
+@pytest.mark.parametrize("dtype", param_types)
+@pytest.mark.parametrize("bs", batch_sizes)
+@pytest.mark.parametrize("model", ["small"])
+@pytest.mark.parametrize("fuse_wgrad_accumulation", all_boolean)
+@pytest.mark.parametrize("recipe", fp8_recipes + [None])
+@pytest.mark.parametrize("fp8_model_params", all_boolean)
+def test_linear_accuracy_save_original_input(
+    dtype, bs, model, fuse_wgrad_accumulation, recipe, fp8_model_params
+):
```
timmoon10: How long do these tests take? Most of these other configs are orthogonal to `save_original_input` and are already tested elsewhere, so no need to have a combinatorial number of test cases.
```suggestion
@pytest.mark.parametrize("dtype", param_types)
@pytest.mark.parametrize("recipe", fp8_recipes + [None])
def test_linear_accuracy_save_original_input(
    *,
    dtype: torch.dtype,
    bs, : int = 1,
    model: str = "small",
    recipe: Optional[recipe.Recipe],
) -> None:
```

tests/pytorch/test_numerics.py
```diff
@@ -1727,6 +1793,7 @@ def _test_grouped_linear_accuracy(
 @pytest.mark.parametrize("fuse_wgrad_accumulation", all_boolean)
 @pytest.mark.parametrize("bias", all_boolean)
 @pytest.mark.parametrize("delay_wgrad_compute", all_boolean)
+@pytest.mark.parametrize("save_original_input", all_boolean)
```
timmoon10: It's better practice to have a separate test for `save_original_input`. Testing all features together involves O(2^n) cases, while testing independently requires O(2*n).

transformer_engine/pytorch/module/linear.py
```diff
@@ -433,6 +447,9 @@ def forward(
                     FP8GlobalStateManager.IS_FIRST_FP8_MODULE = _first_fp8_module
             ctx.wgrad_store = wgrad_store
 
+            ctx.save_original_input = save_original_input and backward_needs_input
+            ctx.own_quantized_input = own_quantized_input
+
```
timmoon10: These are redundant since we can check `inputmat` directly:
```suggestion
```

hxbai: Removed.

tests/pytorch/test_numerics.py
hxbai: Added separate cases for `test_grouped_linear_accuracy_save_original_input` and ` test_padding_grouped_linear_accuracy_save_original_input`.

tests/pytorch/test_numerics.py
hxbai: Removed some unnecessary combinations.

transformer_engine/pytorch/module/linear.py
hxbai: Modified accordingly. For `(Float8Quantizer, Float8CurrentScalingQuantizer)`, the kernels do not support only quantizing the columnwise data, so I changed here a little bit.

hxbai: For GroupedLinear, I leave it unmodified. Because the GroupedLinear needs extra `split` operations, and it is hard to decouple `save_original_input` with the normal path.

PR #1871, opened 2025-06-12
[JAX] Grouped GEMM & Dense support MXFP8 and handle empty matrices

huanghua1994:
# Description

This PR adds two features for grouped GEMM and grouped dense:

1. Support MXFP8 on Blackwell
2. Handle empty matrices 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -247,10 +313,41 @@ Error_Type GroupedGemmFFI(cudaStream_t stream, Buffer_Type lhs_data, Buffer_Type
...
+    for (int i = 0; i < num_non_empty_gemms; i++) {
+      // The i-th GEMM will use the (i % num_streams)-th stream to compute,
+      // use the same stream to swizzle the scaling factors to make sure that
+      // the swizzling is done before the GEMM computation starts.
+      int stream_id = i % num_streams;
+      cudaStream_t stream_i = nvte_get_compute_stream(stream_id);
+      nvte_swizzle_scaling_factors(lhs_swizzle_list[i], lhs_list[i], stream_i);
+      nvte_swizzle_scaling_factors(rhs_swizzle_list[i], rhs_list[i], stream_i);
+    }
+  }
```
phu0ngng, line 345: Hi, I suggest we move this implementation to TE/Common so that TE/PyT may be able to use it in the future.

huanghua1994: TE/PyT handles `scale_inv` matrices one by one ([here](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/csrc/extensions/gemm.cpp#L400) and [here](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/csrc/util.cpp)) using the same stream. I will ask them if they want to follow our approach, and probably they will just change the interface of `swizzle_scaling_factors` instead of reusing code.

timmoon10: Perhaps we could implement something like `nvte_multi_tensor_swizzle_scaling_factors` that takes a list of tensors. Internally it would use this multi-stream implementation, but we could easily switch out with a single-tensor implementation in the future.

phu0ngng: We want to have this PR to be included in TEv2.5, as the grouped_gemm should support zero input.
Thus, splitting the swizzling to `nvte_multi_tensor_swizzle_scaling_factors` will be addressed in a separate PR.

phu0ngng:
Pipeline #30061653 passed, thus the PR is good to go.

PR #1879, opened 2025-06-13
[Pytorch] Bugfix in te fusion ce implementation

BestJuly:
# Description

Fix the issue in [Megatron-LM#1627](https://github.com/NVIDIA/Megatron-LM/issues/1627).    
Add related unit test that capture this.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


tests/pytorch/test_parallel_cross_entropy.py
```diff
@@ -61,19 +62,33 @@ def one_iteration_test(
...
             ref_loss.backward()
 
-        test_loss = torch.flatten(test_loss) if not reduce_loss else test_loss
+        # Compare losses
+        if use_sum_backward:
+            # When using sum().backward(), compare the summed losses
+            torch.testing.assert_close(test_loss.sum(), ref_loss.sum(), check_dtype=False)
+        else:
+            test_loss = torch.flatten(test_loss) if not reduce_loss else test_loss
+            torch.testing.assert_close(test_loss, ref_loss, check_dtype=False)
```
timmoon10: We still want to check that each entry of the loss output is correct, even if we're going to sum afterward:
```suggestion
        test_loss = torch.flatten(test_loss) if not reduce_loss else test_loss
```

tests/pytorch/test_parallel_cross_entropy.py
```diff
@@ -51,6 +51,7 @@ def one_iteration_test(
         label_smoothing: float,
         reduce_loss: bool,
         ignore_idx: bool = False,
+        use_sum_backward: bool = False,
```
timmoon10: Why not check  `not reduce_loss`? We don't check the backward pass when `reduce_loss=False` and `use_sum_backward=False`, which seems like a hole in our testing.
```suggestion
```
This also means we don't need to add a new test case.

timmoon10 (commented):
Overall makes sense, but we should change the tests so that every test case checks the backward pass. We missed this bug because we only tested the backward pass with `reduce_loss=True`.

tests/pytorch/test_parallel_cross_entropy.py
BestJuly: That is a good suggestion. I have modified the test code accordingly.

tests/pytorch/test_parallel_cross_entropy.py
```diff
@@ -134,3 +138,19 @@ def test_ignore_idx(self):
...
+    #     """
+    #     # Set up infrastructure for non-reduced loss (reduction='none')
+    #     self.generate_infra(False, 0)  # reduce_loss=False, label_smoothing=0
+    #     self.one_iteration_test(
+    #         dtype=torch.float32,
+    #         swap_dim=False,
+    #         label_smoothing=0,
+    #         reduce_loss=False,
+    #         ignore_idx=False,
+    #     )
```
timmoon10:
```suggestion
```

timmoon10 (approved):
LGTM, pending CI.

ksivaman (approved):

PR #1883, opened 2025-06-16
[PyTorch][MoE] Kernels fusions for the MoE router

Autumn1998:
# Description

Provide function used in the router fusion and corresponding unit-test.

1. fuse the topk+softmax/sigmoid
2. fuse the score function used in the aux loss
3. fuse the aux loss computation

All 3 parts include the forward and backward.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


tests/pytorch/test_fused_router.py
timmoon10, line 1: We should also add this test to this QA script: https://github.com/NVIDIA/TransformerEngine/blob/main/qa/L0_pytorch_unittest/test.sh

tests/pytorch/test_fused_router.py
```diff
@@ -0,0 +1,410 @@
...
+        topk=topk,
+        use_pre_softmax=use_pre_softmax,
+        num_groups=num_groups,
+        group_topk=group_topk,
+        scaling_factor=scaling_factor,
+        score_function=score_function,
+        expert_bias=expert_bias_clone,
+    )
+
+    assert torch.allclose(probs, probs_fused, atol=atol, rtol=rtol)
```
timmoon10: It would be nicer to replace `torch.allclose` with [`torch.testing.assert_close`](https://docs.pytorch.org/docs/stable/testing.html#torch.testing.assert_close). It has a helpful error message and it automatically chooses the tols based on the dtype.
```suggestion
    torch.testing.assert_close(probs, probs_fused)
```

tests/pytorch/test_fused_router.py
```diff
@@ -0,0 +1,410 @@
...
+
+
+@pytest.mark.parametrize("dtype", [torch.float32])
+@pytest.mark.parametrize("num_tokens", [2048, 7168, 32111])
+@pytest.mark.parametrize("num_experts", [128, 32])
+@pytest.mark.parametrize("topk", [4, 8])
+@pytest.mark.parametrize("group_topk", [None, 4])
+@pytest.mark.parametrize("scaling_factor", [None, 1.2])
+@pytest.mark.parametrize("enable_bias", [True, False])
+def test_topk_sigmoid(
```
timmoon10, line 219: How long does this test suite take to run? The number of test cases grows very quickly if you have one test with many parameters (O(2^n) cases), so it may be better to split it up into multiple tests with only a few parameters (O(2*n) cases).

transformer_engine/common/include/transformer_engine/fused_router.h
```diff
@@ -0,0 +1,129 @@
...
+ *  \param[in]     num_tokens      Number of tokens.
+ *  \param[in]     num_experts     Number of experts.
+ *  \param[in]     topk            Topk value.
+ *  \param[in]     score_function  Score function, 0: sigmoid, 1: softmax.
+ *  \param[out]    scores          Output tensor for scores.
+ *  \param[in]     routing_map     Routing map.
+ *  \param[in]     intermediate_output  Intermediate output from the forward pass. (Softmax/sigmoid output)
+ *  \param[in]     stream          CUDA stream used for the operation.
+ */
+void nvte_fused_scores_for_aux_loss_forward(const NVTETensor logits, int num_tokens,
```
timmoon10: What is the naming convention for this loss function in other MoE implementations? It feels like `aux_loss` is too general and it might get confusing if some other multi-objective training method becomes popular in the future. Maybe something like `moe_aux_loss` would be more specific.

tests/pytorch/test_fused_router.py
Autumn1998: @timmoon10 updated

tests/pytorch/test_fused_router.py
Autumn1998: The kernels are small and will not consume long time . All 237 test cases only need several seconds(<5s).

transformer_engine/common/include/transformer_engine/fused_router.h
Autumn1998: I have update all aux_loss to moe_aux_loss.

tests/pytorch/test_fused_router.py
Autumn1998: @timmoon10

tests/pytorch/test_fused_router.py
timmoon10: Seems like the test was removed from the QA script.

transformer_engine/common/include/transformer_engine/fused_router.h
```diff
@@ -0,0 +1,131 @@
...
+ *  \param[in]     scaling_factor  Scaling factor.
+ *  \param[in]     score_function  Score function, 0: sigmoid, 1: softmax.
+ *  \param[in]     expert_bias     Expert bias. (Only used at the sigmoid case)
+ *  \param[out]    probs           Output tensor for probabilities.
+ *  \param[out]    routing_map     Output tensor for routing map.
+ *  \param[out]    intermediate_output  Output tensor for intermediate output. (Softmax/sigmoid output)
+ *  \param[in]     stream          CUDA stream used for the operation.
+ */
+void nvte_fused_topk_with_score_function_forward(
+    const NVTETensor logits, int num_tokens, int num_experts, int topk, bool use_pre_softmax,
```
timmoon10: Nit: The C API should not use `bool` since it is a C++ type:
```suggestion
    const NVTETensor logits, int num_tokens, int num_experts, int topk, int use_pre_softmax,
```

transformer_engine/common/fused_router/fused_moe_aux_loss.cu
```diff
@@ -0,0 +1,288 @@
...
+}
+
+template <typename DataType, typename IndexType>
+void fused_moe_aux_loss_forward_kernel_launcher(const DataType* probs,
+                                                const IndexType* tokens_per_expert,
+                                                int total_num_tokens, int num_tokens,
+                                                int num_experts, int topk, float coeff,
+                                                DataType* aux_loss, float* Const_buf,
+                                                cudaStream_t stream) {
+  if (cuda::sm_arch(cuda::current_device()) >= 900) {
```
timmoon10, line 171: Do we handle the CUDA stream properly in this path?

PR #1890, opened 2025-06-18
Added MCore FSDP support for TE

sanandaraj5597:
# Description

Added support for gradient accumulation fusion when using MCore FSDP with TE.

### DIFF
--- a/transformer_engine/pytorch/module/linear.py
+++ b/transformer_engine/pytorch/module/linear.py
@@ -463,14 +463,13 @@ class _Linear(torch.autograd.Function):
                     else None
                 )
             else:
-                main_grad = weight.main_grad
+                main_grad = weight.get_main_grad()
 
             if ctx.cpu_offloading:
                 if ctx.grad_added_to_main_grad:
                     weight = ctx.weight_object
                 if ctx.requires_wgrad and ctx.fuse_wgrad_accumulation:
-                    if not hasattr(param, '__fsdp_param__'):
-                        weight.main_grad = main_grad
+                    weight.main_grad = main_grad
 
             # Gather intermediate/activation tensors if needed
             # NOTE: weight_fp8 = weight when ctx.fp8 == False and torch.disttributed.FSDP already
--- a/transformer_engine/pytorch/module/linear.py
+++ b/transformer_engine/pytorch/module/linear.py
@@ -397,7 +397,11 @@ class _Linear(torch.autograd.Function):
             ctx.grad_output_quantizer = grad_output_quantizer
             ctx.fuse_wgrad_accumulation = fuse_wgrad_accumulation
             if fuse_wgrad_accumulation and weight.requires_grad:
-                ctx.main_grad = weight.main_grad
+                # This check is needed to ensure that main_grad is not created 
+                # during the forward pass when using MCore FSDP as it creates
+                # the main_grad buffer lazily before backprop
+                if not hasattr(param, '__fsdp_param__'):
+                    ctx.main_grad = weight.main_grad
 
             ctx.debug = debug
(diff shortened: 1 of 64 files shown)


shjwudp:
Thank you! This is very beneficial for improving FSDP's performance. I think it would be best to make the memory allocation for main_grad an explicit function. What do you think?

sanandaraj5597:
@shjwudp I think I updated the code based on the recent code factor. Please check once.

sanandaraj5597:
@timmoon10 that's a good point. When we use TE with MCore FSDP, the main grads buffers are lazily allocated during backward. So, if we are using the fuser with FSDP then we need to allocate the buffer and then use it.

transformer_engine/pytorch/module/linear.py
```diff
@@ -397,7 +397,11 @@ def forward(
             ctx.grad_output_quantizer = grad_output_quantizer
             ctx.fuse_wgrad_accumulation = fuse_wgrad_accumulation
             if fuse_wgrad_accumulation and weight.requires_grad:
-                ctx.main_grad = weight.main_grad
+                # This check is needed to ensure that main_grad is not created
+                # during the forward pass when using MCore FSDP as it creates
+                # the main_grad buffer lazily before backprop
+                if not hasattr(param, "__fsdp_param__"):
+                    ctx.main_grad = weight.main_grad
```
timmoon10: Instead of spreading out the conditional logic in multiple places, how about something like:
```suggestion
                if hasattr(weight, "__fsdp_param__"):
                    # MCore FSDP creates main_grad lazily before backward
                    ctx.main_grad_func = weight.get_main_grad
                else:
                    ctx.main_grad_func = lambda: weight.main_grad
```
Then we can just call the function directly in backward:
```python
main_grad = None
if ctx.fuse_wgrad_accumulation and ctx.requires_wgrad:
    main_grad = ctx.main_grad_func()
```

timmoon10, line 404: Instead of spreading out the conditional logic in multiple places, how about something like:
```suggestion
                if hasattr(weight, "__fsdp_param__"):
                    # MCore FSDP creates main_grad lazily before backward
                    ctx.main_grad_func = weight.get_main_grad
                else:
                    ctx.main_grad_func = lambda: weight.main_grad
```
Then we can just call the function directly in backward:
```python
main_grad = None
if ctx.fuse_wgrad_accumulation and ctx.requires_wgrad:
    main_grad = ctx.main_grad_func()
```

shjwudp (approved):
LGTM! Thanks!

transformer_engine/pytorch/ops/basic/basic_linear.py
```diff
@@ -978,6 +978,9 @@ def op_backward(
         accumulate_into_main_grad = self._accumulate_into_main_grad
         grad_weight = None
         if ctx.weight_requires_grad and accumulate_into_main_grad:
+            if hasattr(self.weight, "__fsdp_param__"):
+                self.weight.main_grad = self.weight.get_main_grad()
```
timmoon10, line 970: Does Mcore FSDP properly clean up `main_grad` if we set it externally like this?

timmoon10 (approved):
LGTM, pending CI

transformer_engine/pytorch/ops/basic/basic_linear.py
sanandaraj5597: Main gradients are something that aren't cleared up across the training runs. Its held by the optimizer forever.

sanandaraj5597: But if you are asking about FSDP main grad clearance, yes MCore FSDP does it properly.

PR #1897, opened 2025-06-26
[PyTorch Debug] More advanced stats for Quantized Tensors

pggPL:
# Description

This PR adds more statistics for Quantized Tensors that can be used to debug FP8 convergence issues.

It also adds inspect_tensor_all api call, which was necessary to support this features.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring


transformer_engine/debug/features/api.py
```diff
@@ -231,6 +233,9 @@ def inspect_tensor(
         tp_group: torch.distributed.ProcessGroup,
     ) -> None:
         """
+        This is legacy call, we advise to use *inspect_tensor_all* and *inspect_tensor_all_enabled* instead.
```
timmoon10: If we don't want users to call `inspect_tensor` and other legacy functions, we should raise a deprecation warning.

Also, it seems like a shame to so quickly abandon these function names and replace them with clunky `_all` variants. Is there a way to maintain backward compatibility? Maybe have kwargs like `high_precision_tensors` and `quantized_tensors`, and have `quantized_tensors=False` by default.

transformer_engine/debug/features/api.py
```diff
@@ -292,6 +298,62 @@ def inspect_tensor_postquantize(
...
+        quantized_tensor_rowwise: Optional[Union[torch.Tensor, QuantizedTensor]] = None,
+        quantized_tensor_columnwise: Optional[Union[torch.Tensor, QuantizedTensor]] = None,
+        quantizer: Optional[Quantizer] = None,
+    ) -> None:
+        """
+        Similar to *inspect_tensor* and *inspect_tensor_postquantize*. It is run always once per tensor.
+        It allows to inspect both quantized and high precision tensors.
+        The feature LogFp8TensorStats uses this call to collect FP8 statistics after the quantization.
+
+        If tensor and the transpose are quantized differently
```
timmoon10: Accidentally deleted some of the docstring?

docs/debug/3_api_te_calls.rst
```diff
@@ -12,14 +12,7 @@ Let's look deeper into how Nvidia-DL-Framework-Inspect with Transformer Engine w
...
    Fig 1: Example of Nvidia-DL-Framework-Inspect affecting training script with 1 Linear Layer. For tensors mentioned in ``config.yaml``, behavior of ``modify_tensor_enabled()`` and ``modify_tensor()`` calls are substituted with definitions from the feature class. Other calls return default values - in fact they do nothing.
 
-In this page, all calls from TransformerEngine to the Nvidia-DL-Framework-Inspect for each GEMM are listed. The order of these calls is illustrated in the image below.
-
-.. figure:: ./img/api_calls2.svg
-   :align: center
-
-   Fig 2: The calls to Nvidia-DL-Framework-Inspect done for Transformer Engine. There are 2 types of calls: GEMM calls and routing calls.
-
-
```
ptrendx, line 22: Why removing this figure? Especially since the text afterwards not changing suggests that this should still be relevant.

transformer_engine/debug/features/api.py
```diff
@@ -227,6 +227,9 @@ def inspect_tensor(
         layer_name: str,
         tensor_name: str,
         tensor: torch.Tensor,
+        rowwise_quantized_tensor: Optional[torch.Tensor],
+        columnwise_quantized_tensor: Optional[torch.Tensor],
```
ptrendx, line 249: Why do we need to have both of them? Also, in the `log_fp8_tensor_stats.py` you assume that they are actually identical and just a `QuantizedTensor` - shouldn't we then just request a single `QuantizedTensor` instead?

transformer_engine/debug/features/api.py
```diff
@@ -329,6 +336,8 @@ def inspect_tensor_postquantize_enabled(
         iteration: int,
     ) -> bool:
         """
+        This is legacy call, we advise to use *inspect_tensor_all* and *inspect_tensor_all_enabled* instead.
```
ptrendx: If it is deprecated then we should say it is deprecated rather than "legacy". Also, we should issue a deprecation warning the first time somebody uses it.

transformer_engine/debug/features/utils/stats_computation.py
```diff
@@ -87,62 +113,198 @@ def _get(buffers, stat_name):
...
+        * 100,
+        lambda buffers: 100 * sum(_get(buffers, "overflows_num")) / sum(_get(buffers, "numel")),
     ),
 }
+
+
+def add_underflows_stats(recipe_name: str, columnwise: bool = False):
+    """Register *both* underflow stats (num and %) for the given recipe."""
+    columnwise_suffix = "_columnwise" if columnwise else ""
+    data_tensor_idx = 1 if columnwise else 0
```
ptrendx: This seems pretty brittle.

transformer_engine/debug/features/api.py
pggPL: Sometimes we may use fakequant for one gemm and fp8 quantize for other, so these will not be the same object.

transformer_engine/debug/features/api.py
pggPL: done

transformer_engine/debug/features/utils/stats_computation.py
pggPL: moved data_tensor_idx logic inside the get_data_tensors() function

docs/debug/3_api_te_calls.rst
pggPL: I needed to change it and it seems that I lost pptx file somewhere.

transformer_engine/debug/features/utils/stats_computation.py
```diff
@@ -87,62 +113,207 @@ def _get(buffers, stat_name):
...
+    This replaces the earlier separate helpers and avoids duplicated boilerplate.
+    """
+    # Determine which attribute holds the scale-inverse tensor.
+
+    def get_scale_inv(quantized_tensor, columnwise):
+        if hasattr(quantized_tensor, "_scale_inv"):
+            return getattr(quantized_tensor, "_scale_inv")
+        if columnwise:
+            return getattr(quantized_tensor, "_columnwise_scale_inv")
+        return getattr(quantized_tensor, "_rowwise_scale_inv")
```
timmoon10, line 254: This will work for now, but we don't make any commitments to keep this scale-inv naming scheme. We are doing some experiments with a format that doesn't store the scale-inv at all, only the amax. It would be more robust to explicitly enumerate the supported types:
```suggestion
    def get_scale_inv(quantized_tensor, columnwise):
        if isinstance(quantized_tensor, Float8TensorBase):
            return quantized_tensor._scale_inv
        if isinstance(quantized_tensor, (MXFP8TensorBase, Float8BlockwiseQTensorBase)):
            if quantized_tensor._rowwise_scale_inv is not None:
                return quantized_tensor._rowwise_scale_inv
            return quantized_tensor._columnwise_scale_inv
        raise ValueError(f"Unsupported quantized tensor format ({quantized_tensor.__class__.__name__})")
```

timmoon10 (approved):
Overall LGTM

PR #1899, opened 2025-06-26
[PyTorch] Tests for loading previously-generated checkpoints

timmoon10:
# Description

Some downstream users experienced pain when we merged https://github.com/NVIDIA/TransformerEngine/pull/1033 because it affected the checkpoint format, namely by adding `_extra_state` to the state dict for the `LayerNorm` and `RMSNorm` modules. This PR adds tests for loading previously-generated checkpoints with [`strict=True`](https://docs.pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.load_state_dict). Even if we decide to change the checkpoint format in the future, these tests will make sure that it is a deliberate decision.

Note that these tests don't check correctness since that's tested in a few other places:
https://github.com/NVIDIA/TransformerEngine/blob/cc0cb35d4cb516669e1f166add37ccdc30f71488/tests/pytorch/test_numerics.py#L846
https://github.com/NVIDIA/TransformerEngine/blob/cc0cb35d4cb516669e1f166add37ccdc30f71488/tests/pytorch/test_fusible_ops.py#L1971

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring
- [x] Testing

## Changes

- Add tests for loading previously-generated checkpoints

# Checklist:

### DIFF
--- a/qa/L0_pytorch_unittest/test.sh
+++ b/qa/L0_pytorch_unittest/test.sh
@@ -45,7 +45,7 @@ NVTE_FLASH_ATTN=0 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_cp
 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_fused_attn.xml $TE_PATH/tests/pytorch/fused_attn/test_fused_attn.py || test_fail "test_fused_attn.py"
 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_kv_cache.xml $TE_PATH/tests/pytorch/fused_attn/test_kv_cache.py || test_fail "test_kv_cache.py"
 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_hf_integration.xml $TE_PATH/tests/pytorch/test_hf_integration.py || test_fail "test_hf_integration.py"
-TE_TEST_CHECKPOINT_ARTIFACT_PATH=$TE_PATH/artifacts/tests/pytorch/test_checkpoint python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_checkpoint.xml $TE_PATH/tests/pytorch/test_checkpoint.py || test_fail "test_checkpoint.py"
+NVTE_TEST_CHECKPOINT_ARTIFACT_PATH=$TE_PATH/artifacts/tests/pytorch/test_checkpoint python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_checkpoint.xml $TE_PATH/tests/pytorch/test_checkpoint.py || test_fail "test_checkpoint.py"
 
 if [ "$RET" -ne 0 ]; then
     echo "Error in the following test cases:$FAILED_CASES"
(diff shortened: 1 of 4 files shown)


qa/L0_pytorch_unittest/test.sh
```diff
@@ -45,6 +45,7 @@ NVTE_FLASH_ATTN=0 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_cp
 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_fused_attn.xml $TE_PATH/tests/pytorch/fused_attn/test_fused_attn.py || test_fail "test_fused_attn.py"
 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_kv_cache.xml $TE_PATH/tests/pytorch/fused_attn/test_kv_cache.py || test_fail "test_kv_cache.py"
 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_hf_integration.xml $TE_PATH/tests/pytorch/test_hf_integration.py || test_fail "test_hf_integration.py"
+TE_TEST_CHECKPOINT_ARTIFACT_PATH=$TE_PATH/artifacts/tests/pytorch/test_checkpoint python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_checkpoint.xml $TE_PATH/tests/pytorch/test_checkpoint.py || test_fail "test_checkpoint.py"
```
ksivaman: For the CI purposes, this doesn't actually test older checkpoints, right?

ksivaman:
```suggestion
NVTE_TEST_CHECKPOINT_ARTIFACT_PATH=$TE_PATH/artifacts/tests/pytorch/test_checkpoint python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_checkpoint.xml $TE_PATH/tests/pytorch/test_checkpoint.py || test_fail "test_checkpoint.py"
```

tests/pytorch/test_checkpoint.py
```diff
@@ -0,0 +1,175 @@
...
+                return te.ops.Linear(32, 32)
+        raise ValueError(f"Unrecognized module name ({name})")
+
+    @staticmethod
+    @functools.lru_cache(maxsize=None)
+    def _checkpoint_dir() -> pathlib.Path:
+        """Path to directory with checkpoint files"""
+
+        # Check environment variable
+        path = os.getenv("TE_TEST_CHECKPOINT_ARTIFACT_PATH")
```
ksivaman:
```suggestion
        path = os.getenv("NVTE_TEST_CHECKPOINT_ARTIFACT_PATH")
```

qa/L0_pytorch_unittest/test.sh
timmoon10: I'm planning on keeping the files in our internal repos and modifying the CI pipeline to install them at this path. Unfortunately, this means non-NVIDIA users won't be able to run the tests themselves.

My original approach was to include the checkpoint files with Git LFS, but that could add some burden for users who don't have Git configured properly. We've had users struggle even with Git submodules, so I don't think it would be worth it to introduce LFS for such a basic test.

yaox12:
@timmoon10 @ksivaman This PR caused a CI failure. Can you take a look?

timmoon10:
@yaox12 https://github.com/NVIDIA/TransformerEngine/pull/1934 changed the pickling logic for MXFP8 tensors, which broke checkpoint backward compatibility. https://github.com/NVIDIA/TransformerEngine/pull/1943 should fix it.

PR #1909, opened 2025-06-30
[PyTorch debug] Improve precision debug tools performance

pggPL:
# Description

This PR aims to speed up layers which are not affected by any feature in particular iteration. They should be *exactly* as fast as layers without initializing debug tools.

I needed to fix 3 things:

- There was a lot of CPU overhead when we tried to decide if layer uses any feature in current iteration. We have called `inspect_tensor_enabled` and few similar calls for each layer, iteration and tensor. I changed calls like `inspect_tensor_enabled`- they may return tuple (bool, int), where int indicated number of iteration the feature will be enabled next time. If each tensor for one layer returns (bool, n) we run non-debug layer for next n iterations,
- `debug_api.step()` is called after every iteration. Inside it, we call `STATS_BUFFER.log()` which performs synchonization and some cpu ops, even if no stats is logged. I disable this logic if no stat was logged.
- `COMM/GEMM` overlap was disabled for the whole time, now it is disabled when layer is affected by at least one feature.

If we want to only log some stats every n iterations, then this PR should make it work as fast as non-debug workflow when n -> infinity.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:


pggPL:
PR ready for review, waiting for https://github.com/NVIDIA/nvidia-dlfw-inspect/pull/7 to be merged to update version and run tests (edit: merged by now).

pggPL:
Tested on E2E, perf of layers without features matches the perf of layers without debug initialized.

transformer_engine/debug/pytorch/utils.py
```diff
@@ -5,6 +5,19 @@
...
+    """Returns next iteration at which the debug should be run."""
+    return min(q.get_next_debug_iter() for q in quantizers)
+
+
 def any_feature_enabled(quantizers):
     """Returns True if at least one API call is made from DebugQuantizer."""
     return any(q.any_feature_enabled() for q in quantizers)
+
+
+def _as_pair(x):
```
timmoon10: This is a code smell. We use it to coerce some function returns into a consistent format, which is a hint that those functions should not have variable return formats.

transformer_engine/pytorch/module/base.py
```diff
@@ -1402,12 +1404,55 @@ def backward_dw(self):
...
+
+        # If layer is run first time in new iteration,
+        # we need to check if the debug should be enabled for this layer -
+        # maybe in previous iterations debug features returned information
+        # that no feature will be active for this layer for multiple next iterations.
+        started_new_iteration = TEDebugState.get_iteration() != getattr(
+            self, "debug_last_iteration_name", None
+        )
+        if self.next_iter_when_debug_should_be_run is not None and started_new_iteration:
+            if self.next_iter_when_debug_should_be_run is False:
```
timmoon10: I'm getting code smells. `next_iter_when_debug_should_be_run` seems like it should be an `int`, so it's very weird to me that it can be `False`.

transformer_engine/debug/pytorch/debug_state.py
```diff
@@ -62,6 +62,13 @@ def set_weight_tensor_tp_group_reduce(cls, enabled):
         """Sets weight tensor reduction mode."""
         cls.weight_tensor_tp_group_reduce = enabled
 
+    @classmethod
+    def get_iteration(cls):
```
timmoon10, line 66: I'm seeing some weird usage of this function in `TransformerEngineBaseModule.validate_debug`. The function name makes me expect this behavior:
```suggestion
    def get_iteration(cls) -> Optional[int]:
```
If there is anything unintuitive going on, we should document it.

transformer_engine/pytorch/module/base.py
```diff
@@ -1402,12 +1404,55 @@ def backward_dw(self):
                 if bias_tensor.grad is None:
                     bias_tensor.grad = bgrad.to(bias_tensor.dtype)
 
+    def validate_debug(self):
```
timmoon10: Nit: It's not obvious to me that `validate_debug` returns a `bool`.
```suggestion
    def is_debug_iter(self) -> bool:
```

transformer_engine/debug/features/log_tensor_stats.py
```diff
@@ -97,7 +98,14 @@ def inspect_tensor_enabled(
...
         """API call used to determine whether to run look_at_tensor_before_process() in the forward."""
-        return self._check_params(config, layer_name, iteration=iteration)
+        run_current, next_iter = next_enabled_iter(
+            config.get("start_step", None),
+            config.get("end_step", None),
+            config.get("start_end_list", None),
+            config.get("freq", 1),
+            iteration,
+        )
+        return run_current, next_iter
```
timmoon10, line 109: Nit: `inspect_tensor_enabled` makes me expect a `bool` to be returned. Maybe better to split this up into two functions: `inspect_tensor_enabled` and `inspect_tensor_next_enabled_iter`.

transformer_engine/debug/pytorch/debug_quantization.py
```diff
@@ -70,6 +71,12 @@ def __init__(
 
         self.rowwise_gemm_name, self.columnwise_gemm_name = _tensor_to_gemm_names_map[tensor_name]
 
+        # next iteration when this quantizer will call any API
+        # it is inf at the init and it is computed after_enabled api calls.
+        # inf at the beginning means that if nothing will be done,
+        # this quantizer will never call any API.
+        self.next_debug_iter = float("inf")
```
timmoon10: `None` or `-1` are better for indicating an invalid iter. If `next_debug_iter` can be `float`, then that makes me think we support fractional iters.
```suggestion
        self.next_debug_iter = None
```

transformer_engine/debug/pytorch/debug_quantization.py
```diff
@@ -102,15 +109,23 @@ def get_plans_for_output_tensors(self) -> Tuple[bool, str]:
         """
         import nvdlfw_inspect.api as debug_api
 
-        inspect_tensor_enabled = debug_api.transformer_engine.inspect_tensor_enabled(
-            layer_name=self.layer_name, tensor_name=self.tensor_name, iteration=self.iteration
+        inspect_tensor_enabled, next_iter = _as_pair(
```
timmoon10: I'm suspicious how we put these `_as_pair`s everywhere. Some of the `*_enabled` functions return iter info and some do not, and we're trying to force these dissimilar things to have the same API.

transformer_engine/debug/pytorch/debug_quantization.py
```diff
@@ -446,29 +502,28 @@ def update_quantized(
...
-            or self.columnwise_tensor_plan == API_CALL_MODIFY
-        ):
-            return True
-        if self.parent_quantizer is not None:
-            if self.rowwise_tensor_plan != STANDARD_FP8_QUANTIZE:
-                return True
-            if self.columnwise_tensor_plan != STANDARD_FP8_QUANTIZE:
-                return True
-        return False
+    def get_next_debug_iter(self) -> int | float:
```
timmoon10: Returning a `float` makes me think we're supporting fractional iters. More straightforward to use `None` or `-1` instead of `float("inf")`.
```suggestion
    def get_next_debug_iter(self) -> Optional[int]:
```

transformer_engine/debug/features/api.py
```diff
@@ -133,9 +145,14 @@ def modify_tensor_enabled(
         gemm: str,
         tensor_name: str,
         iteration: int,
-    ) -> bool:
+    ) -> bool | Tuple[bool, int | float]:
```
timmoon10: Why imply inconsistent output types?
```suggestion
    ) -> Tuple[bool, Optional[int]]:
```
This just makes them much more awkward to use, i.e. we need to put `_as_pair` everywhere. I can see two cases:
- We want to enforce the same API for functions that return iter info and functions that don't. But what benefit does that bring? Better to have different APIs and only do iter-related logic when it's actually needed.
- We want to ensure backward compatibility, e.g. with subclasses implemented outside of TE. Perhaps it would be better to create a new API for functions that return iter info?

tests/pytorch/debug/test_perf.py
```diff
@@ -0,0 +1,76 @@
...
+            model = torch.nn.Sequential(
+                te.Linear(1, 1, name="linear1"), te.Linear(1, 1, name="linear2")
+            ).cuda()
+            NUM_ITERS = 18000
+        elif layer == "transformer":
+            model = torch.nn.Sequential(
+                te.TransformerLayer(1, 1, 1, name="transformer1"),
+                te.TransformerLayer(1, 1, 1, name="transformer2"),
+            ).cuda()
+            NUM_ITERS = 2000
```
timmoon10, line 37: This seems like an expensive test. How long does it take to run?

qa/L0_pytorch_debug_unittest/test.sh
timmoon10, line 1: These tests aren't being run in our CI pipelines. Also, we should keep in mind that our testing budget is limited, so I'd prefer moving as much as possible from L0 to L1. However, that's beyond the scope of this PR and we can make that change in a follow-up PR.

timmoon10 (dismissed):
This PR seems functionally correct, so I won't block it. However, there are a lot of questionable stylistic choices that make it harder to understand and awkward to use.

tests/pytorch/debug/test_perf.py
pggPL: Around 20 seconds

qa/L0_pytorch_debug_unittest/test.sh
pggPL: I added them, that was mistake that they were not run - I was sure they are. But yes, we can think about that outside this PR.

tests/pytorch/debug/test_api_features.py
```diff
@@ -26,20 +26,20 @@ def test_transformer_engine_no_config(feature_dirs):
             "decoder.1.attn.qkv", gemm="fprop", iteration=0
         )
 
-        # modify_tensor_enabled - False by default
+        # modify_tensor_enabled - (False, float("inf")) by default
```
ptrendx: This comment should now say `(False, None)`, right?

tests/pytorch/debug/test_api_features.py
```diff
@@ -265,21 +265,20 @@ def assert_empty():
...
         assert not debug_api.transformer_engine.inspect_tensor_enabled(
             "decoder.2.mlp.fc1", tensor_name="activation", iteration=200
-        )
+        )[0]
         assert not debug_api.transformer_engine.inspect_tensor_enabled(
             "decoder.1.mlp.fc1", tensor_name="gradient", iteration=200
-        )
+        )[0]
 
         expected_underflows = (tensor_fp8._data == 0).sum() * 100 / (100 * 100 * 5)
```
ptrendx, line 276: The overflows got removed because they are incorrect?

tests/pytorch/debug/test_configs/log_config.yaml
```diff
@@ -0,0 +1,19 @@
...
+        - tensor: activation
+          stats: [cur_amax, dynamic_range, mean, std, l1_norm]
+          start_step: 1
+          freq: 3
+    LogFp8TensorStats:
+      enabled: True
+      tensors: weight
+      stats: [underflows%]
+      start_step: 1
+      freq: 3
```
ptrendx: Maybe it would be good to have a test that has features with 2 different frequencies?

tests/pytorch/debug/test_perf.py
```diff
@@ -0,0 +1,76 @@
...
+    # then it changed into non-debug mode and should not have any non-negligible cpu overhead
+    # compared to layer without debug tools initialized.
+
+    with_debug_tools = _run_cpu_overhead(True, layer, configs_dir, feature_dirs)
+    without_debug_tools = _run_cpu_overhead(False, layer, configs_dir, feature_dirs)
+
+    print(f"with_debug_tools: {with_debug_tools} s")
+    print(f"without_debug_tools: {without_debug_tools} s")
+
+    assert with_debug_tools < without_debug_tools * 1.1  # 10% overhead margin
```
ptrendx: What is the typical overhead in practice? Is 10% high enough margin that we wouldn't need to revisit it in the future?

transformer_engine/debug/features/api.py
```diff
@@ -101,13 +101,23 @@ def _process_transformer_engine_config(self, config, **kwargs):
...
+        gemm: str,
+        iteration: int,
+    ) -> bool | Tuple[bool, Optional[int]]:
         """
         If the tensor is not processed using *modify_tensor* and the fp8 recipe is enabled,
         then the decision whether to cast it to fp8 is based on the value returned by the call *fp8_gemm_enabled*.
         If the tensor is processed using *modify_tensor* or fp8 autocast is not enabled,
         the result of this call does not matter.
 
+        This method may return a tuple (bool, Optional[int]), where the int indicates the next iteration when the feature will be enabled.
```
ptrendx, line 117: Is this file not checked by linter? This line is too long.

transformer_engine/debug/features/api.py
```diff
@@ -124,7 +134,7 @@ def fp8_gemm_enabled(self, config: Dict, layer_name: str, gemm: str, iteration:
 
         bool - default is True
```
ptrendx: This should change to reflect the API change as well.

transformer_engine/debug/features/api.py
```diff
@@ -432,6 +456,37 @@ def output_assertions_hook(self, api_name, ret, **kwargs):
...
+                isinstance(feature_output, tuple)
+                for feature_output in multi_feature_outputs.values()
+            )
+            if all_ret_tuple:
+                run_current = any(
+                    feature_output[0] for feature_output in multi_feature_outputs.values()
+                )
+                next_iter = min(
+                    feature_output[1] for feature_output in multi_feature_outputs.values()
+                )
```
ptrendx: Does this work with the None values? A quick test doing min([None, 1]) gives me TypeError...

transformer_engine/debug/features/log_tensor_stats.py
```diff
@@ -97,7 +98,15 @@ def inspect_tensor_enabled(
         self, config: Dict, layer_name: str, tensor_name: str, iteration: int
     ):  # pylint: disable=unused-argument
         """API call used to determine whether to run look_at_tensor_before_process() in the forward."""
-        return self._check_params(config, layer_name, iteration=iteration)
+        run_current, next_iter = next_enabled_iter(
```
ptrendx, line 101: This function call replaces `_check_params` but inside the `next_enabled_iter` it does not call `_check_params`. Does this mean that `_check_params` is no longer used/needed?

transformer_engine/debug/features/utils/__init__.py
```diff
@@ -5,3 +5,38 @@
...
+        if iteration % freq == 0 and s <= iteration <= e:
+            run_current = True
+
+        first = max(iteration + 1, s)
+        offset = first % freq
+        candidate = first if offset == 0 else first + (freq - offset)
+        if candidate <= e:
+            return run_current, candidate
+
+    raise RuntimeError("No next iteration found")
```
ptrendx: Shouldn't it return `(False, None)` then rather than raising the error?

transformer_engine/debug/pytorch/utils.py
ptrendx: You are correct. The problem is that those functions were part of the public API so we need to keep the backwards compatibility with the previous format. We should probably mark the single boolean return as deprecated though, right @pggPL?

transformer_engine/debug/pytorch/debug_quantization.py
ptrendx: Another thought - seems that the `_as_pair` is always paired with the `update_next_iter`. Maybe we could instead have a single function that will figure out internally whether it is a tuple or not and update the state? Having 2 is more prone to bugs.

transformer_engine/pytorch/distributed.py
```diff
@@ -1388,24 +1397,18 @@ def gather_along_first_dim(
         final_quantizer = (
             None if not needs_quantized_gemm(inp, rowwise=True) else quantizer.parent_quantizer
         )
-        # Temporary fix for TP communication of Float8BlockwiseQTensorBase
-        if isinstance(rowwise, Float8BlockwiseQTensorBase):
-            rowwise = inp._original_tensor
```
ptrendx, line 1393: Why are those removed?

transformer_engine/pytorch/distributed.py
```diff
@@ -1388,24 +1397,18 @@ def gather_along_first_dim(
...
-            # Temporary fix for TP communication of Float8BlockwiseQTensorBase
-            if isinstance(columnwise, Float8BlockwiseQTensorBase):
-                columnwise = inp._original_tensor
             columnwise_total, _ = gather_along_first_dim(
                 columnwise, process_group, False, final_quantizer_columnwise
             )
             out_obj.columnwise_gemm_tensor = columnwise_total
         else:
-            out_obj.rowwise_gemm_tensor = out_obj.rowwise_gemm_tensor
+            out_obj.columnwise_gemm_tensor = out_obj.rowwise_gemm_tensor
```
ptrendx: To be honest, I don't understand the original logic (Which seemed to not do anything) neither the new logic, could you give some explanation here?

transformer_engine/pytorch/module/base.py
ptrendx: Seems that the `next_iter_when_debug_should_be_run` function can output `float('inf')` but not a bool, so yeah, there seems to be some inconsistency here.

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -162,6 +161,13 @@ def forward(
         with_input_all_gather = parallel_mode == "column" and sequence_parallel
 
         # Configure Userbuffers communication (comm+GEMM overlap)
+        if debug:  # turn off userbuffers in debug mode
+            ub_overlap_ag_fprop = False
+            ub_overlap_rs_fprop = False
+            ub_overlap_ag_dgrad = False
+            ub_overlap_rs_dgrad = False
+            ub_bulk_wgrad = False
+            ub_bulk_dgrad = False
```
ptrendx, line 169: Is this just because our naming in the autograd functions of different modules is inconsistent?

tests/pytorch/debug/test_api_features.py
pggPL: This line was not used in the tests.

We do not support overflows, they will be introduced in the PR with stats for quantized tensors.

transformer_engine/debug/features/log_tensor_stats.py
pggPL: Check params is function from nvinspect package - the class inherits from`BaseLogTensorStats` from nvinspect. It is not longer needed here.

transformer_engine/debug/features/utils/__init__.py
pggPL: Something was not pushed correctly, yes - I fixed this.

transformer_engine/pytorch/distributed.py
pggPL: When I merged debug tools blockwise gather had some workaround with all-gather - it was done in high precision. It is no longer needed, since all-gather in fp8 is supported for blockwise.

transformer_engine/pytorch/module/layernorm_linear.py
pggPL: I don't understand the command. Do you mean why I turn off the userbuffers?

transformer_engine/pytorch/distributed.py
pggPL: done

tests/pytorch/debug/test_perf.py
pggPL: I need this margin to make this test reliable pass, overhead is much smaller.

tests/pytorch/debug/test_api_features.py
pggPL: fixed

transformer_engine/debug/pytorch/debug_quantization.py
pggPL: I changed it into one function

transformer_engine/debug/features/api.py
pggPL: Liine too long check was removed from linter, probably because of autoformatting https://github.com/NVIDIA/TransformerEngine/blame/main/pylintrc

transformer_engine/debug/features/api.py
pggPL: fixed

transformer_engine/debug/features/api.py
pggPL: fixed

tests/pytorch/debug/test_configs/log_config.yaml
pggPL: changed this tests

transformer_engine/pytorch/module/layernorm_linear.py
ptrendx: No, the question is why you changed the previous code which had a function call here to turn off unsupported features during debug into doing it manually in every module.

pggPL: Before I disabled UB permanently if debug was enabled. I want to disable them only for debug iterations now and I think this is the easiest way to do so (with least additional lines)

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -260,6 +260,10 @@ def onnx_dequantize(self, tensor) -> torch.Tensor:
     def _get_compatible_recipe(self) -> Union[type[Recipe], None]:
         """Returns recipe class that is compatible with this quantizer"""
 
+    def supports_only_rowwise_all_gather(self) -> bool:
+        """Returns True if the quantizer supports only rowwise all-gather"""
+        return False
```
timmoon10, line 265: Future optimization: We might be able to remove this by incorporating the [swap-first-dims kernel](https://github.com/NVIDIA/TransformerEngine/pull/1998) in `gather_along_first_dim`.

timmoon10 (approved):
Style is much improved. It'll be nice to finally get these perf improvements merged in.

PR #1912, opened 2025-06-30
[PyTorch] Optimize `create_tensor` in quantizer

yaox12:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/csrc/quantizer.cpp
+++ b/transformer_engine/pytorch/csrc/quantizer.cpp
@@ -495,7 +495,7 @@ std::pair<TensorWrapper, py::object> MXFP8Quantizer::create_tensor(
     }
     auto sinv0 = roundup(numel / last_dim, 128);
     auto sinv1 = roundup(last_dim / MXFP8_BLOCK_SIZE, 4);
-    rowwise_scale_inv = at::zeros({sinv0, sinv1}, opts);
+    rowwise_scale_inv = at::empty({sinv0, sinv1}, opts);
     tensor.set_rowwise_data(data.data_ptr(), this->dtype, shape);
     tensor.set_rowwise_scale_inv(
         rowwise_scale_inv.data_ptr(), DType::kFloat8E8M0,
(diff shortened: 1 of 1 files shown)


ptrendx (changes requested):
That change unfortunately cannot be done and at::zeros is there on purpose - cuBLAS requires the scale inv to be zero initialized, otherwise it will result in silent wrong answer errors.

yaox12:
> That change unfortunately cannot be done and at::zeros is there on purpose - cuBLAS requires the scale inv to be zero padded, otherwise it will result in silent wrong answer errors.

Yes, I realized that. I plan to write out zeros in the quantization kernel. We're benchmarking MoE model training in MXFP8 and seeing a lot of these `aten::zeros` and `swizzle_scaling_factor` kernels, which makes the speedup of MXFP8 very limited. @xrennvidia 
For the swizzling kernels, I plan to implement a `multi_tensor_swizzle` kernel to fuse them.

Screenshot 2025-06-30 at 2 04 14 PM

zhongbozhu:
Maybe we should do the following two things:
- I am gonna write a fused torch.empty feature for MXFP8 like https://github.com/NVIDIA/TransformerEngine/pull/1793 just merged
- while the allocation is torch.empty instead of torch.zeros, the multi_tensor_swizzle will do the zero padding on the tensors

yaox12:
> Maybe we should do the following two things:
> 
> * I am gonna write a fused torch.empty feature for MXFP8 like [[PyTorch][MoE] Reduce CPU Overhead By Fuse Torch Empty Calls #1793](https://github.com/NVIDIA/TransformerEngine/pull/1793) just merged
> * while the allocation is torch.empty instead of torch.zeros, the multi_tensor_swizzle will do the zero padding on the tensors

Sounds great!

This is a better solution, since both zero padding and swizzling are required by cuBLAS, it's better to do both of them in one place.
I just realized filling zeros in the quantization kernel was not able to cover all cases, since there're other producers of MXFP8 tensors, such as cuDNN normalizations.

PR #1917, opened 2025-07-01
Call `pre_(first_)forward` only when global state changes

janekb04:
# Description

Change `FusibleOperation.pre_forward` to `FusibleOperation.pre_first_forward` and call it only before the first forward iteration using a given recipe type in `OperationFuser.__call__`. Additionally, reset fusions on recipe type change to enable recipe-dependent fusions and retrigger a call to `pre_first_forward`. See removed calls in the visual comparison of forward pass below.

Resultant performance gain in benchmark scenario: +4% for FP8 Sequential.

## Visual comparison

image

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Rename `FusibleOperation.pre_forward` to `FusibleOperation.pre_first_forward`
- Call `pre_first_forward` only before the first forward in `OperationFuser` instead of calling it every iteration
- Pass `recipe: Optional[Recipe]` as argument to `pre_first_forward` instead of access via `FP8GlobalStateManager`
- Remove redundant `set_usage` calls in `BasicLinear.pre_first_forward`
- Cache `is_fp8_enabled` and `type(fp8_recipe)` in `Sequential` and reset `_module_groups` on change

## Checklist:

## Performance data

Below is reported the running time of a GPT encoder transformer layer (averaged over 10k runs) using [my benchmark script](https://github.com/janekb04/transformer-engine-benchmark-sequential). Here, Fused refers to an implementation using the fused modules `LayerNormLinear` and `LayerNormMLP`, Sequential refers to an implementation implementing those two modules with `te.ops.Sequential`, and Builtin refers to using `te.TransformerLayer`.

### B100 results

|Commit|Fused BF16|Fused FP8|Sequential BF16|Sequential FP8|Builtin BF16|Builtin FP8|
|-|-|-|-|-|-|-|
|6f4310d700f7445fd12d524c645b1e72fb8886f7|1.57ms|1.89ms|1.56ms|2.07ms|2.25ms|2.64ms|
|8bcdb796f2624aed331b167d53fb56e85d1f3478|1.60ms|1.92ms|1.53ms|1.99ms|2.24ms|2.64ms|

### H100 results

|Commit|Fused BF16|Fused FP8|Sequential BF16|Sequential FP8|Builtin BF16|Builtin FP8|
|-|-|-|-|-|-|-|
|6f4310d700f7445fd12d524c645b1e72fb8886f7|1.16ms|1.54ms|1.08ms|1.70ms|1.85ms|2.37ms|
|8bcdb796f2624aed331b167d53fb56e85d1f3478|1.15ms|1.53ms|1.06ms|1.61ms|1.84ms|2.35ms|

### DIFF
--- a/transformer_engine/pytorch/ops/fuser.py
+++ b/transformer_engine/pytorch/ops/fuser.py
@@ -344,6 +344,9 @@ class OperationFuser:
         self._forward_ops = [(op, (idx,)) for idx, op in enumerate(self._basic_ops)]
         self._backward_ops = list(reversed(self._forward_ops))
 
+        # Flag for checking if this is the first iteration
+        self._is_first_forward = True
+
         # Fuse ops if needed
         if fuse_ops:
             self.fuse_ops()
@@ -391,8 +394,11 @@ class OperationFuser:
             )
 
         # Initialization before forward pass
-        for op in self._basic_ops:
-            op.pre_forward()
+        if self._is_first_forward:
+            recipe = FP8GlobalStateManager.get_fp8_recipe()
+            for op in self._basic_ops:
+                op.pre_first_forward(recipe=recipe)
+            self._is_first_forward = False
 
         # Canonicalize op kwargs
         if basic_op_kwargs is None:
--- a/transformer_engine/pytorch/ops/fuser.py
+++ b/transformer_engine/pytorch/ops/fuser.py
@@ -395,7 +395,8 @@ class OperationFuser:
 
         # Initialization before forward pass
         if self._is_first_forward:
-            recipe = FP8GlobalStateManager.get_fp8_recipe()
+            with_quantized_compute = FP8GlobalStateManager.is_fp8_enabled()
+            recipe = FP8GlobalStateManager.get_fp8_recipe() if with_quantized_compute else None
             for op in self._basic_ops:
                 op.pre_first_forward(recipe=recipe)
             self._is_first_forward = False
--- a/transformer_engine/pytorch/ops/op.py
+++ b/transformer_engine/pytorch/ops/op.py
@@ -65,7 +65,11 @@ class FusibleOperation(torch.nn.Module, metaclass=abc.ABCMeta):
     def is_fused_op(self) -> bool:
         """Whether this op is the fusion of one or more basic ops"""
 
-    def pre_forward(self) -> None:
+    def pre_first_forward(
+        self,
+        *,
+        recipe: Optional[Recipe],
+    ) -> None:
         """Preprocessing before forward pass"""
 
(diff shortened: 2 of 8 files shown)


transformer_engine/pytorch/ops/fuser.py
```diff
@@ -67,6 +67,8 @@ def forward(
         input_: torch.Tensor,
         fuser: OperationFuser,
         basic_op_kwargs: list[dict[str, Any]],
+        fp8_enabled: bool,
+        fp8_recipe: Recipe,
```
timmoon10: "Quantized" is better than "fp8" since this logic applies to non-FP8 formats:
```suggestion
        with_quantized_compute: bool,
        quantization_recipe: Recipe,
```

transformer_engine/pytorch/ops/op.py
```diff
@@ -65,7 +65,12 @@ class FusibleOperation(torch.nn.Module, metaclass=abc.ABCMeta):
...
         """Whether this op is the fusion of one or more basic ops"""
 
-    def pre_forward(self) -> None:
+    def pre_forward(
+        self,
+        *,
+        fp8_enabled: bool,
+        fp8_recipe: Recipe,
+    ) -> None:
         """Preprocessing before forward pass"""
```
timmoon10: `pre_forward` is no longer accurate since we only call it when the recipe changes. How about something like:
```suggestion
    def pre_first_forward(
        self,
        *,
        recipe: Optional[Recipe],
    ) -> None:
        """Preprocessing before first forward pass with a quantization recipe"""
```
It's not quite perfect, but it captures the intent better.

Also, we could implicitly treat `recipe=None` as using no quantization recipe.

timmoon10 (commented):
I wonder if we can extend this work to also support recipe-specific fusions. Instead of having the recipe check in the fuser, how about we put it in `Sequential`: https://github.com/NVIDIA/TransformerEngine/blob/6f4310d700f7445fd12d524c645b1e72fb8886f7/transformer_engine/pytorch/ops/sequential.py#L188-L190
Whenever the global state changes, we just create a new fuser and the fuser would call `pre_forward` within its first `__call__`.

timmoon10 (approved):
LGTM

transformer_engine/pytorch/ops/fuser.py
janekb04: I ultimately removed these arguments after having moved the caching logic to `Sequential` and reverted much of `fuser.py` to its original state.

janekb04:
I tested that `test_layernorm_mlp` (from #1924) also passes (when this branch is locally rebased on `add-sequential-ln-mlp-test`).

timmoon10 (approved):

PR #1921, opened 2025-07-03
[PyTorch] Fuse permute+pad and unpermute+unpad ops for FP8 optimization

xiaoxi-wangfj:
1.Fused `moe_permute_with_probs` + `Fp8Padding` and fused `moe_unpermute` + `Fp8Unpadding`, which removes the explicit padding/unpadding in the MOE experts module, improved performance and reduced peak gpu memory usage.
2.Added tests of fused permute/pad and unpermute/unpad operations.

# Description

This PR optimizes FP8 MoE permute and pad operations by:
1. Fusing `moe_permute_with_probs` + `Fp8Padding` into `moe_permute_and_pad_with_probs`
2. Fusing `moe_unpermute` + `Fp8Unpadding` into `moe_unpermute` with `pad_offsets` argument
3. Thereby removing explicit padding/unpadding steps in the MOE experts module

**Results**:
- **1.1x~1.6x speedup** for fused permute-and-pad operations
- **1.7x~3x speedup** for fused unpermute-and-unpad operations (measured by `tests/pytorch/test_permutation.py`)
- Verified in ene-to-end FP8 model training with Megatron framework, **+0.4% MFU uplift** and **~1GB peak GPU memory reduction** in a typical ~600B paramter setup.

## Performance data

Tests covering a wide range of model training configurations were performed comparing the fused operations ("Fused:") and the original version ("Orig:"). Running time (in milliseconds) are summarized in the table below and the speedup, measured as the reciprocal of the ratio between running times, are also provided. All tests were carried out with the tests/pytorch/test_permutation.py benchmark script.

Fused-perm-pad

## The usage in Megatron-LM

1. `Megatron-LM/megatron/core/transformer/moe/moe_utils.py` : Added Support for Fused Operations

`

    # Added fused function import
    from megatron.core.extensions.transformer_engine import (
        ...,
        fused_permute_and_pad_with_probs,  # [!code ++]
    )

    def permute(
        ...,
        tokens_per_expert: Optional[torch.Tensor] = None,  # [!code ++]
        align_size: int = -1  # [!code ++]
    ):
      ...
      if fused and probs is not None:
          if not HAVE_TE or fused_permute_with_probs is None:
              raise ValueError(
                  "fused_permute_with_probs is not available. Please install TE >= 2.1.0."
              )
          if tokens_per_expert is not None and align_size > 0:  # [!code ++]
              # Use fused permute+pad operation [!code ++]
              return fused_permute_and_pad_with_probs(tokens, probs, routing_map, tokens_per_expert, align_size)   # [!code ++]
          else:
              # Fallback to original implementation
              ...

    def unpermute(
        ...,
        pad_offsets: Optional[torch.Tensor] = None  # [!code ++]
    ):
        return fused_unpermute(
            ...,
            pad_offsets=pad_offsets  # [!code ++]
        )

`

2. `Megatron-LM/megatron/core/transformer/moe/token_dispatcher.py`: Scheduler Integration

`

    class _DeepepManager(_DispatchManager):
        def __init__(...):
            self.pad_offsets = None  # [!code ++] Store padding offsets
        
        def get_permuted_hidden_states_by_experts(...):
            ...
            if self.config.moe_permute_padding_for_fp8:# [!code ++]
                # Use fused path [!code ++]
                (                                                          # [!code ++]
                    hidden_states,                                         # [!code ++]
                    permuted_probs,                                        # [!code ++]
                    self.reversed_mapping_for_combine,                     # [!code ++]
                    self.pad_offsets,                                      # [!code ++]
                    self.tokens_per_expert                                 # [!code ++]
                ) = permute(                                               # [!code ++]
                    hidden_states,                                         # [!code ++]
                    self.dispatched_routing_map,                           # [!code ++]
                    probs=self.dispatched_probs,                           # [!code ++]
                    fused=self.permute_fusion,                             # [!code ++]
                    tokens_per_expert=self.tokens_per_expert,              # [!code ++]
                    align_size=get_fp8_align_size(self.config.fp8_recipe), # [!code ++]
                )                                                          # [!code ++]
            else:
                # Original path
                ...
        
        def get_restored_hidden_states_by_experts(...):
            hidden_states = unpermute(
                ...,
                pad_offsets=self.pad_offsets if self.config.moe_permute_padding_for_fp8 else None, # [!code ++]
            )
            ...

`

## Type of change

Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Added `moe_permute_and_pad_with_probs` api for fused permute and pad, modified `moe_unpermute` api with pad_offsets argument for fused unpermute and unpad in transformer_engine/pytorch/permutation.py
- Added tests in `tests/pytorch/test_permutation.py`

# Checklist:

### DIFF
--- a/transformer_engine/jax/triton_extensions/utils.py
+++ b/transformer_engine/jax/triton_extensions/utils.py
@@ -157,7 +157,7 @@ def compile_triton(
         )
     else:
         kernel = gpu_triton.TritonKernel(
-            compile.name,
+            compiled.name,
             num_warps,
             compiled.metadata.shared,
             compiled.asm["ptx"],


transformer_engine/pytorch/permutation.py
```diff
@@ -191,6 +191,7 @@ def forward(
         routing_map: torch.Tensor,
         num_out_tokens: int,
         probs: torch.Tensor,
+        pad_offsets: torch.Tensor,
```
timmoon10: Nit: `pad_offsets` can be `None`.
```suggestion
        pad_offsets: Optional[torch.Tensor],
```

yaox12:
Thanks for you contribution. 
We also notice there're redundant read and write with the current "permute and then pad" routine. We plan to tackle it by padding the routing map before permutation. Refer to this [commit](https://github.com/NVIDIA/Megatron-LM/commit/92d68dae89af0baab2d4eee092f884902dca4db0) in Megatron-LM. 
For permute fusion, we try to avoid tiny cuda kernels as much as possible. I went through your PR and found that [these lines](https://github.com/NVIDIA/TransformerEngine/pull/1921/files#diff-616c77928f08b7bcd0501442ebc4636434b4961e49f8ff5679559d3346fb1339R629-R633) may introduce some of them. 
~~We prefer padding routing map instead of fusing permutation and padding because we think that `tokens_per_expert` is produced by preprocess and we don't want it to be changed later, otherwise it may cause some confusions.
So we won't merge this PR. Thanks for you contribution again.~~

Never mind, I agree that fusing permutation with padding is a better solution.

xiaoxi-wangfj:
> Thanks for you contribution. We also notice there're redundant read and write with the current "permute and then pad" routine. We plan to tackle it by padding the routing map before permutation. Refer to this [commit](https://github.com/NVIDIA/Megatron-LM/commit/92d68dae89af0baab2d4eee092f884902dca4db0) in Megatron-LM. For permute fusion, we try to avoid tiny cuda kernels as much as possible. I went through your PR and found that [these lines](https://github.com/NVIDIA/TransformerEngine/pull/1921/files#diff-616c77928f08b7bcd0501442ebc4636434b4961e49f8ff5679559d3346fb1339R629-R633) may introduce some of them. We prefer padding routing map instead of fusing permutation and padding because we think that `tokens_per_expert` is produced by preprocess and we don't want it to be changed later, otherwise it may cause some confusions. So we won't merge this PR. Thanks for you contribution again.

@yaox12 
Thank you for your response. 

1. The moe_router_padding_for_fp8 and fused_permute_pad_for_fp8 configurations are compatible. Within the moe_router_padding_for_fp8 logic, if not_enough_tokens_to_pad is triggered, execution will fall back to the fused_permute_pad_for_fp8 computational path. Otherwise, pad_offsets will be set to None."

2. We previously enabled the Megatron-LM [commit](https://github.com/NVIDIA/Megatron-LM/commit/92d68dae89af0baab2d4eee092f884902dca4db0) configuration you mentioned, but found that pading routing may caused loss instability. During pre-training—especially with larger fp8 align size values like 128 (due to our 1*128 blockwise setting)—it frequently triggered the not_enough_tokens_to_pad warning, and this forced a fallback to explicit padding within GroupedMLP, which halved iteration performance when occurring, and sometimes resulted in loss deterioration. Then we disabled pading routing map.
The implementation of Fused_permute_pad will skips any configuration modifications for Fp8Padding/Fp8Unpadding through its update of tokens_per_expert. Regarding refinements to this fused approach, do you have any suggestions for? I hope to continue to refine that optimization."

shcho1118:
We have also implemented permute + pad, unpermute + unpad fusion that works with deepgemm grouped gemm for training, and we are using it.
The implementation is a middle ground between what @yaox12 mentioned and this pr (we pad routing map, but there are some tiny kernels).
The loss instability mentioned by @xiaoxi-wangfj could be due to garbage values in padded gradients from unpermute backward (same as permute gradients).
We had a similar problem, and the solution was to do a zero-init when allocating gradient tensors.

xiaoxi-wangfj:
During pretraining of a 600B-parameter MoE model using the blockwise recipe, I observed that the pad_routing_map behavior leads to performance degradation at larger expert parallelism (EP) scales. Therefore, I evaluated the performance of the fused_permute_pad optimization under these settings. The results are as follows:
I found that fused_permute_pad demonstrates better scalability and delivers consistent performance gains over the baseline across different EP configurations.
fusedpermpad

transformer_engine/pytorch/triton/permutation.py
```diff
@@ -424,16 +435,28 @@ def permute_with_mask_map(
...
+            permuted_probs = torch.empty((num_out_tokens,), dtype=probs.dtype, device="cuda")
     else:
         permuted_probs = None
 
     if scale is not None:
-        permuted_scale = torch.empty(
-            (num_out_tokens, scale_hidden_dim), dtype=scale.dtype, device="cuda"
-        )
+        if pad_offsets is not None:
+            permuted_scale = torch.zeros(
```
tdophung: why do we also need to zero init the padded permuted scales? if the padded input is already zero-init, then we already have 0 * scale = 0. So why the need for this?

transformer_engine/pytorch/triton/permutation.py
tdophung, line 1: There has been some refactoring happening regarding triton kernels in TE. Changes included:
- moving all triton kernels to transformerengine/common/
- keep all pytorch wrapper things in transformerengine/pytorch/triton and transformerengine/pytorch/
- added jax wrappers and primitives that calls these triton kernels from transformerengine/common/

Can you please rebase/merge main into here and add your changes on top of it?

transformer_engine/pytorch/triton/permutation.py
```diff
@@ -424,16 +435,28 @@ def permute_with_mask_map(
...
+                (num_out_tokens, scale_hidden_dim), dtype=scale.dtype, device="cuda"
+            )
+        else:
+            permuted_scale = torch.empty(
```
tdophung:
```suggestion
    alloc = torch.zeros if pad_offsets is not None else torch.empty
    output = alloc((num_out_tokens, hidden_size), dtype=inp.dtype, device="cuda")
    permuted_probs = (
        alloc((num_out_tokens,), dtype=probs.dtype, device="cuda") if probs is not None else None
    )
    permuted_scale = (
        alloc((num_out_tokens, scale_hidden_dim), dtype=scale.dtype, device="cuda")
        if scale is not None
        else None
    )
```

tests/pytorch/test_permutation.py
```diff
@@ -653,6 +655,302 @@ def _test_permutation_mask_map(
...
+        msg=f"Mismatch in te_permute_and_pad bwd",
+        **tols,
+    )
+
+    ###################################################################################################################################
+    #
+    # Benchmark
+    #
+    ###################################################################################################################################
+    if BENCHMARK:
```
tdophung, line 857: Not sure if we usually do benchmarking in transformerengine tests or not.

transformer_engine/pytorch/triton/permutation.py
xiaoxi-wangfj: Thank you for the update. I have rebased my branch onto the latest main, resolved all conflicts, and reapplied my changes on top of the updated Triton kernel structure.

transformer_engine/pytorch/triton/permutation.py
xiaoxi-wangfj: You're right — zero-initializing the padded scales isn’t necessary. I’ve updated the latest commit to remove the zero-init.

tests/pytorch/test_permutation.py
xiaoxi-wangfj: I updated the test and set the default value of BENCHMARK to False in test_permutation_and_padding_mask_map, so pytest will no longer run the benchmark by default

transformer_engine/pytorch/triton/permutation.py
xiaoxi-wangfj: @tdophung ，
It should be noted that I did not modify the calls to _permute_kernel and _unpermute_kernel in transformer_engine/jax/triton_extensions/permutation.py. These should also be updated to pass the new pad_offsets_ptr and FUSION_PAD parameters. Since I cannot validate JAX behavior on my side, I’ll need you to adapt these interface changes accordingly.

transformer_engine/common/triton/permutation.py
tdophung, line 1: What is the reason for not handling both probs and pad fusion in the case of unpermute backwards?

If it was that you simply missed it and this was not intentional, can you please also handle the padding fusion in this kernel: _unpermute_bwd_with_merging_probs_kernel?
It is being used in the bwd rule of moe_unpermute_mask_map in the with_probs case

In addition to that, there needs to be a test case for this as current test_permutation.py does not cover it

transformer_engine/pytorch/permutation.py
```diff
@@ -387,15 +396,15 @@ def forward(
...
         # pylint: disable=missing-function-docstring
         if not unpermuted_act_grad.numel():
-            return unpermuted_act_grad, None, ctx.merging_probs, None
+            return unpermuted_act_grad, None, ctx.merging_probs, None, None
 
         act_grad = None
         probs_grad = None
         if ctx.needs_input_grad[0]:
             if ctx.with_probs:
                 fwd_input, row_id_map, merging_probs = ctx.saved_tensors
```
tdophung: why is with_probs mutually exclusive with padding here

tdophung: if it is intentionally mutually exclusive, as stated in another comment at the beginning of this file, then you should have a warning to users or something in case they pass in a non-None pad ptr and non-None probs, then unpermute_with_mask_map_bwd_with_merging_probs gets called without the right ctx and silently output wrong data.

transformer_engine/pytorch/triton/permutation.py
tdophung: I can modify them separately, I just need you to keep the order of the arguments to the kernels in transformerengine/common/permutation.py the same as HEAD of main now, (then the addition of pad_offsets_tr should be where the other input pointers are, etc.)  so jax should work. The rule for jax triton binding to work is to have the order of args goes like:
input ptrs, scalars, output ptrs, constexpr.

I wonder if I could contribute a few commits to your PR here to cover the jax side, so that when  we merge this PR, it will not break the tests. Can you allow write access for me?

tdophung:
also, since you do not allow write permission, the automatic pre-commit.ci service cannot automatically format the code for you and therefore fail the pre-commit.ci - pr check. Therefore, you could also try running the qa/format.sh script manually to format the code before pushing. Otherwise please enable write permission (as requested for jax compatibility purposes above)

transformer_engine/pytorch/permutation.py
tdophung: I was able to have them wotking together (padding, and merging probs, unpermuting). I made a few commits here on this branch that I rebase from your branch: https://github.com/tdophung/TransformerEngine/tree/teddy/jax-permutation-pad-fuse.

If you let me have write access I could also commit this to this PR

transformer_engine/common/triton/permutation.py
xiaoxi-wangfj: This was an oversight on my part. I did not test or cover the case where both probs and padding fusion occur in the unpermute backward path. I will add the missing padding–fusion handling to _unpermute_bwd_with_merging_probs_kernel and include a corresponding test case to ensure this scenario is properly validated.

transformer_engine/pytorch/permutation.py
xiaoxi-wangfj: It’s likely because the fork is under my organization, so the PR does not allow edits by maintainers by default. Web-based commits should still be mergeable. I’ve already added you as a collaborator with push access, and I’m waiting for you to accept the invitation.

I’ve added fused pad support to unpermute_with_mask_map_bwd_with_merging_probs, and I also included a with_merging_probs test branch in the existing _test_permutation_and_padding_mask_map test. Please review again.

xiaoxi-wangfj:
@tdophung All CI warnings have been addressed. Please let me know if I missed anything.

timmoon10 (approved):
LGTM, pending CI.

transformer_engine/pytorch/permutation.py
xiaoxi-wangfj: @timmoon10 fixed

tdophung:
The change LGTM.

However, though CI passed, it is with a longer than we would like elapsed time for these test_permutation.py in https://gitlab-master.nvidia.com/dl/transformerengine/transformerengine/-/pipelines/40385518

I will remove some test cases here and keep a larger step size between the smallest and largest case to reduce test time in this PR. 

Approve but please hold off on submitting until I committed the changes for JAX + reduce test time here.

transformer_engine/jax/triton_extensions/utils.py
```diff
@@ -142,16 +142,16 @@ def compile_triton(
     )
 
     # Create kernel object for JAX
+    # From jax/jaxlib/gpu/triton_kernels.cc:
+    # Kernel::Kernel(kernel_name, num_warps, num_ctas, shared_mem_bytes, ptx, ttir, compute_capability)
     kernel = gpu_triton.TritonKernel(
```
jberchtold-nvidia: @tdophung This change in signature is because JAX updated their TritonKernelCall API, right? If so, we may need to support both as users do not always use the nightly JAX version

transformer_engine/jax/triton_extensions/permutation.py
```diff
@@ -395,6 +405,167 @@ def lowering(
 register_primitive(PermuteWithMaskMapPrimitive)
 
 
+class PermuteWithMaskMapAndPadPrimitive(BasePrimitive):
```
jberchtold-nvidia: @tdophung Why do we do both of the following? I thought we'd only need to do one option
1. Update existing `PermuteWithMaskMapPrimitive` to take a `pad_offsets` tensors that can be shape `(0,)` when not doing padding
2. Create an separate new primitive `PermuteWithMaskMapAndPadPrimitive` which looks nearly identical but does do padding.

Let me know if I'm missing something. Thanks!

transformer_engine/jax/triton_extensions/utils.py
tdophung: that's true. I was hoping that this will go in after the 2.11 release so the API in the JAX container would have changed by then, but forgot that not everyone uses it

tdophung: issue is how to detect the version to use the correct API ....

xiaoxi-wangfj:
@tdophung , The CI failure (Lint / JAX Python) seems to be triggered by the changes under the jax/ directory.
Would you mind fixing the issue on your side?

transformer_engine/jax/triton_extensions/permutation.py
tdophung: I made it into 1 call

transformer_engine/jax/triton_extensions/utils.py
jberchtold-nvidia: This looks like the commit that changed the triton call API in JAX: https://github.com/jax-ml/jax/commit/81fe5dd8b4413dfb59ec5d1fcd9006377a9293ab

If we can find what version of JAX this corresponds to, we can do something like this:
```
if version.parse(jax.__version__) >= version.parse("0.5.0"):
```
https://github.com/NVIDIA/TransformerEngine/blob/eb8e792b38228e3c221a2d6b69babe3a91acd1f4/build_tools/jax.py#L33

jberchtold-nvidia: It looks like it is `0.8.2`, based on the tag associated with the commit and the version.py value on the particular commit
https://github.com/jax-ml/jax/blob/81fe5dd8b4413dfb59ec5d1fcd9006377a9293ab/jax/version.py#L24

So if it's easy, can you update to:
```
 if version.parse(jax.__version__) >= version.parse("0.8.2"):
        do new API
else:
       do old API
```

If it's hard to support both, we could instead raise an error in the `else` case for the older JAX versions so it's clear to users

transformer_engine/jax/triton_extensions/permutation.py
jberchtold-nvidia: Looks good, thanks!

tests/jax/test_permutation.py
```diff
@@ -16,7 +16,61 @@
...
+]
+DISPATCH_COMBINE_PADDING_CASES = {
+    "L0": ALL_DISPATCH_COMBINE_PADDING_CASES[0:2],
+    "L2": ALL_DISPATCH_COMBINE_PADDING_CASES,
+}
+
+# Dtypes for testing
+ALL_DTYPES = [jnp.float32, jnp.bfloat16]
+DTYPES = {
+    "L0": ALL_DTYPES,
```
jberchtold-nvidia, line 64: SR: If we need to reduce test runtime further, we can move float32 to L2. But if it's reasonably fast with these changes in L0 now, then we can keep it

tdophung: It is less than 5 min for the total test time currently, so I am keeping it.

jberchtold-nvidia: Awesome, thanks!

jberchtold-nvidia (dismissed):
LGTM, thanks!

timmoon10 (dismissed):
LGTM

jberchtold-nvidia (approved):

xiaoxi-wangfj:
@tdophung 
Is the CI failure caused by test_permutation_and_padding_mask_map taking too long?
Perhaps we can reduce the configuration values for "num_tokens, num_expert, hidden_size, topK", making them consistent with the configuration used in test_permutation_mask_map, for example:

@pytest.mark.parametrize("num_tokens", [4096])
@pytest.mark.parametrize("num_expert", [7, 16])
@pytest.mark.parametrize("hidden_size", [4096])
@pytest.mark.parametrize("topK", [2, 5])

In addition, we could remove the test_permutation_and_padding_with_merging_probs test for reduce test time, since the test scenarios for with_merging_probs args are already covered in test_permutation_and_padding_mask_map.
Do you think this is appropriate?

tdophung:
> @tdophung Is the CI failure caused by test_permutation_and_padding_mask_map taking too long? Perhaps we can reduce the configuration values for "num_tokens, num_expert, hidden_size, topK", making them consistent with the configuration used in test_permutation_mask_map, for example:
> 
> @pytest.mark.parametrize("num_tokens", [4096]) @pytest.mark.parametrize("num_expert", [7, 16]) @pytest.mark.parametrize("hidden_size", [4096]) @pytest.mark.parametrize("topK", [2, 5])
> 
> In addition, we could remove the test_permutation_and_padding_with_merging_probs test for reduce test time, since the test scenarios for with_merging_probs args are already covered in test_permutation_and_padding_mask_map. Do you think this is appropriate?

No there are 2 CI failures:

te-ci/L0_jax_distributed_unittest--H100_4GPU: this is just an infra issue where the node cannot be contacted. I rerun the job and  got it passing: https://gitlab-master.nvidia.com/dl/transformerengine/transformerengine/-/jobs/248541463

te-ci/L0_jax_unittest--B200_1GPU](https://github.com/NVIDIA/TransformerEngine/actions/runs/20469149616): there is a mismatch on sort_chunks_by_index test on jax side, which was not touched in this PR, so I doubt it has anything to do with this PR, I also cannot reproduce the issue locally using the same container transformerengine/transformerengine:te_ci-jax-py3.40704327-devel-<arch>. So I am merging

PR #1924, opened 2025-07-03
Add test for `LayerNormMLP` implementation using `te.ops.Sequential` to `test_fusible_ops.py`

janekb04:
# Description

Adds a test checking the validity of a `LayerNormMLP` implementation using `te.ops.Sequential` to `test_fusible_ops.py`. The test doesn't check for numerical accuracy, but only whether execution actually works, as using multiple layers together makes the results incommensurable with the reference implementation.

Additionally, this PR implements some fixes for issues uncovered with the test.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Renamed `TestSequential` to `TestSequentialContainer` in `test_fusible_ops.py`
- Added `TestSequentialModules` with `TestSequentialModules.test_layernorm_mlp` to `test_fusible_ops.py`
- Fixed access to unimplemented methods of `torch.Tensor` when using internal quantized tensors
  - Added missing method `view` to `MXFP8TensorBase`
  - Added check for tensor being quantized before accessing its `dtype` in `BasicLinear._functional_backward`
- Fixed columnwise usage not being set for `dy` when computing `wgrad` GEMM in `BasicLinear._functional_backward`
- Added explicit check in `MXFP8TensorBase.__new__` and `Float8BlockwiseQTensorBase.__new__` to use `object.__new__`, similarly to `Float8TensorBase.__new__`
- Fixed `_ViewFunc` and `_ReshapeFunc` in `mxfp8_tensor.py` to not transpose the size of the columnwise data

# Checklist:

### DIFF
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -2081,6 +2081,7 @@ class TestCheckpointing:
         for x_load, x_save in zip(xs_load, xs_save):
             torch.testing.assert_close(x_load.grad, x_save.grad, **tols)
 
+
 class TestSequentialModules:
     """Test for larger Sequentials with modules commonly used together"""
 
@@ -2115,7 +2116,7 @@ class TestSequentialModules:
     ) -> None:
         """
         LayerNorm/RMSNorm + Linear + GELU + Linear
-        
+
         Note that this test checks only if the module runs
         as when chaining multiple modules it is hard to validate
         numerical accuracy.
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -152,7 +152,7 @@ def make_reference_and_test_tensors(
     return ref, test
 
 
-class TestSequential:
+class TestSequentialContainer:
     """Tests for sequential container"""
 
     def test_modules(self) -> None:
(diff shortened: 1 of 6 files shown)


transformer_engine/pytorch/ops/basic/basic_linear.py
```diff
@@ -665,7 +665,7 @@ def _functional_backward(
         if dtype is None:
             if weight is not None and not is_quantized_tensor(weight):
                 dtype = weight.dtype
-            else:
+            elif not is_quantized_tensor(grad_output):
                 dtype = grad_output.dtype
```
timmoon10: Thinking more about this, I think the right behavior is to take the dtype from `weight` or `grad_output` if they are `QuantizedTensor`s.
```suggestion
            if isinstance(weight, torch.Tensor):
                dtype = weight.dtype
            elif isinstance(grad_output, torch.Tensor):
                dtype = grad_output.dtype
```

tests/pytorch/test_fusible_ops.py
```diff
@@ -2080,3 +2080,109 @@ def test_linear(
...
+        quantized_weight: bool,
+        dtype: torch.dtype,
+        quantization: Optional[str],
+        device: torch.device = "cuda",
+        hidden_size: int = 32,
+        sequence_length: int = 512,
+        batch_size: int = 4,
+        ffn_hidden_size: int = 64,
+        layernorm_epsilon: float = 1e-5,
+    ) -> None:
```
timmoon10, line 2116: Since this is just a sanity test and we already test the ops thoroughly, we should reduce the number of cases:
```suggestion
    @pytest.mark.parametrize("bias", (False, True))
    @pytest.mark.parametrize("quantized_weight", (False, True))
    @pytest.mark.parametrize("quantization", _quantization_list)
    def test_layernorm_mlp(
        self,
        *,
        bias: bool,
        quantized_weight: bool,
        dtype: torch.dtype = torch.bfloat16,
        quantization: Optional[str],
        device: torch.device = "cuda",
        hidden_size: int = 32,
        sequence_length: int = 512,
        batch_size: int = 4,
        ffn_hidden_size: int = 64,
        layernorm_epsilon: float = 1e-5,
    ) -> None:
```
Other changes:
- Only run with `LayerNorm`
- Use `quantized_compute = quantization is not None`

janekb04: I could reduce the number of test cases, but I'm not sure if that is necessary and a good idea. The test doesn't currently take long to run. Running `python3 -m pytest ./tests/pytorch/test_fusible_ops.py -k "test_layernorm_mlp"` takes less than 4 seconds on both H100 and B100.

Furthermore, `LayerNorm` and `RMSNorm` have separate implementations, so it is possible for one to be erroneous while the other passes the test. There were already errors that occurred only with certain `dtype`s like the `inf` in gradients.

As such, I think it is worthwhile to leave all the test cases. If runtime becomes an issue, the `sequence_length` can be reduced instead.

transformer_engine/pytorch/ops/basic/basic_linear.py
janekb04: Yes, that seems like a better way. I applied the change.

PR #1925, opened 2025-07-03
[JAX] Fix grouped GEMM error on CUDA 12.9.1 & later

huanghua1994:
# Description

On CUDA 12.9.1 and later versions, cuBLAS has some breaking changes and makes JAX grouped GEMM fail. This PR fixes the issues.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

1. For FP8 grouped GEMM on Hopper, ensures that the pointer to each scaling factor is 256B aligned (cuBLAS requires at least 16B alignment, use 256B to simplify the code)
2. For MXFP8 on Blackwell, adjust the `value_and_grad` output in unit test to avoid  NaN values due to overflow.

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/gemm.py
+++ b/transformer_engine/jax/cpp_extensions/gemm.py
@@ -98,15 +98,23 @@ class GroupedGemmPrimitive(BasePrimitive):
             A jnp.ndarray containing the result of the grouped GEMM operation
         """
         del lhs_data_aval, rhs_data_aval, bias_aval, group_offset_aval
-        del K, lhs_is_trans, rhs_is_trans, scaling_mode, has_bias
+        del K, lhs_is_trans, rhs_is_trans, has_bias
         # TODO(Phuong): move some shape checks from Cpp to here
         workspace_size = get_cublas_workspace_size_bytes() * num_cublas_streams
         # cuBLAS workspace ptr must be 256 bytes aligned but JAX buffers are not
         # necessarily 256 bytes aligned, we add some padding to ensure alignment.
-        # We also pad scale_inv swizzle buffers size for 256 bytes alignment.
         workspace_size += 256
-        workspace_size += lhs_scale_inv_aval.size + 256
-        workspace_size += rhs_scale_inv_aval.size + 256
+        if scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING or \
+           scaling_mode == ScalingMode.CURRENT_TENSOR_SCALING:
+            # For tensor scaling, each matrix has a single scale value, but it
+            # needs to be aligned to 16 bytes for CUDA 12.9.1 and later. We use
+            # a 256 bytes alignment to simplify the code.
+            workspace_size += lhs_scale_inv_aval.size * 256
+            workspace_size += rhs_scale_inv_aval.size * 256
+        elif scaling_mode == ScalingMode.MXFP8_1D_SCALING:
+            # We also pad scale_inv swizzle buffers size for 256 bytes alignment.
+            workspace_size += lhs_scale_inv_aval.size + 256
+            workspace_size += rhs_scale_inv_aval.size + 256
         workspace_aval = jax.core.ShapedArray(shape=(workspace_size,), dtype=jnp.uint8)
 
         out_shape = (M, N)
--- a/transformer_engine/jax/csrc/extensions/gemm.cpp
+++ b/transformer_engine/jax/csrc/extensions/gemm.cpp
@@ -62,6 +62,14 @@ Error_Type GroupedGemmFFI(cudaStream_t stream, Buffer_Type lhs_data, Buffer_Type
   NVTE_CHECK(group_sizes.dimensions().size() == 1);
   size_t num_gemms = group_sizes.dimensions()[0];
 
+  // It is weird that TE/Common GEMM only use colwise for MXFP8
+  const bool is_fp8_gemm = is_fp8_dtype(lhs_dtype);
+  const bool is_tensor_scaling = scaling_mode == JAXX_Scaling_Mode::DELAYED_TENSOR_SCALING ||
+                                 scaling_mode == JAXX_Scaling_Mode::CURRENT_TENSOR_SCALING;
+  const bool is_mxfp8_scaling = scaling_mode == JAXX_Scaling_Mode::MXFP8_1D_SCALING;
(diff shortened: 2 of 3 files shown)


transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -86,6 +107,23 @@ Error_Type GroupedGemmFFI(cudaStream_t stream, Buffer_Type lhs_data, Buffer_Type
   size_t bias_dtype_bytes = te_dtype_bytes(bias_dtype);
   size_t out_dtype_bytes = te_dtype_bytes(out_dtype);
 
+  if (is_tensor_scaling) {
+    cudaStream_t stream_0 = nvte_get_compute_stream(0);
```
jberchtold-nvidia, line 111: Does JAX allow using a stream separate from the one provided as a param in the FFI interface? Maybe the grouped gemm is already using these streams internally in TE common so it's okay

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -98,15 +98,25 @@ def abstract(
...
         workspace_size += 256
-        workspace_size += lhs_scale_inv_aval.size + 256
-        workspace_size += rhs_scale_inv_aval.size + 256
+        if (
+            scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING
+            or scaling_mode == ScalingMode.CURRENT_TENSOR_SCALING
+        ):
+            # For tensor scaling, each matrix has a single scale value, but it
+            # needs to be aligned to 16 bytes for CUDA 12.9.1 and later. We use
+            # a 256 bytes alignment to simplify the code.
```
jberchtold-nvidia: What is the additional complexity that would be needed to handle 16-byte alignment of scales instead of 256-byte alignment? I'd think it'd be better to align smaller, though at the size of one scale per typically pretty large tensor the difference between 16-byte/256-bytes per scale might be negligible.

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -86,6 +107,23 @@ Error_Type GroupedGemmFFI(cudaStream_t stream, Buffer_Type lhs_data, Buffer_Type
...
+    size_t dpitch = tensor_scaling_sinv_aligment;
+    size_t spitch = lhs_sinv_dtype_bytes;
+    size_t width = lhs_sinv_dtype_bytes;
+    size_t height = lhs_sinv_size;
+    cudaMemcpy2DAsync(lhs_scatter_aligned_ptr, dpitch, lhs_sinv_ptr, spitch, width, height,
+                      cudaMemcpyHostToDevice, stream_0);
```
phu0ngng: I'm surprised that `lhs_sinv_ptr` could be a host pointer as we allocate it as `jax.ndarray`, which should be a device pointer.

transformer_engine/jax/csrc/extensions/gemm.cpp
phu0ngng: No problem if we do it in this way.
However, I suspect that the `nvte_get_compute_stream(0)` is the same stream as the one given by the FFI. In this case, we should simply use the `stream` given by FFI for the code simplification.

transformer_engine/jax/cpp_extensions/gemm.py
huanghua1994: Fixed, now using 16B alignment.

transformer_engine/jax/csrc/extensions/gemm.cpp
huanghua1994: I think all `Buffer_Type` arrays from FFI are device ptr, just like the issue we have on `group_sizes`.

phu0ngng: If `lhs_sinv_ptr` is already a device pointer, why do we need to do a `HostToDevice` here?

huanghua1994: Thanks for reminding me again. I misread the previous comment. Fixed now.

phu0ngng (dismissed):
LGTM

phu0ngng:
Pipeline #31288765 with the nightly internal image.

All the `grouped_gemm` related tests passed. The failures on L40 and the test_multiprocessing_encoder on B200 are not related to grouped_gemm.

Thus, the PR is good to merge.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -98,15 +98,27 @@ def abstract(
...
         # necessarily 256 bytes aligned, we add some padding to ensure alignment.
-        # We also pad scale_inv swizzle buffers size for 256 bytes alignment.
-        workspace_size += 256
-        workspace_size += lhs_scale_inv_aval.size + 256
-        workspace_size += rhs_scale_inv_aval.size + 256
+        workspace_size += workspace_alignment_padding
+        if scaling_mode in (
+            ScalingMode.DELAYED_TENSOR_SCALING.value,
+            ScalingMode.CURRENT_TENSOR_SCALING.value,
+        ):
```
jberchtold-nvidia, line 113: nit SR: `ScalingMode(scaling_mode).is_tensor_scaling()`

transformer_engine/jax/quantize/tensor.py
```diff
@@ -142,9 +142,9 @@ def __post_init__(self):
             pad_width = tuple(
                 (0, a - b) for a, b in zip(expected_scale_shape, expected_unpadded_scale_shape)
             )
-            # This actually pad scale_inv with nan, should we pad it with 127 directly instead?
+            # padding with the smallest number it can present
```
jberchtold-nvidia, line 145: Just for curiosity, was this needed to fix an issue? Or just using this PR as a chance to update it since we're touching gemm padding already

jberchtold-nvidia (approved):

transformer_engine/jax/quantize/tensor.py
phu0ngng: We need this fix so that the MXFP8 GEMM with custom call works correctly.
We did not see this issue before, as without `/ jnp.sqrt(x.size)` the gradient is probably too large that it gets clipped, so all the tests passed in the past. https://github.com/NVIDIA/TransformerEngine/blob/279d0be059c774c146532cc0a09baddab4ccea5e/tests/jax/test_custom_call_compute.py#L1345-L1349

PR #1932, opened 2025-07-07
[BUILD] Exclude ninja from required packages

phu0ngng:
# Description

JAX images do not come with `ninja` by default; hence, one should not enforce ninja as a required package.
Enforcing ninja as a required package causes build failures when ninja is absent.
In fact, when `ninja` is absent, our `build_ext.py` could fall back to use `MakeFile`. 

This PR removes `ninja` from the required package list in `pyproject.toml`.

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -159,7 +159,7 @@ System Requirements
   * Compiler: GCC 9+ or Clang 10+ with C++17 support
   * Python: 3.12 recommended
 
-* **Source Build Requirements:** CMake 3.18+, Ninja, Git 2.17+, pybind11 2.6.0+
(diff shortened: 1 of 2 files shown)


timmoon10 (commented):
If we go this route, we should also update the README: https://github.com/NVIDIA/TransformerEngine/blob/1ae1d228d725a488621deba685bd26d6ee1cdb21/README.rst?plain=1#L162

timmoon10 (changes requested):
I think Ninja is sufficiently useful that the build system should install it by default. If a user really wants to avoid it, they can build with `--no-dependencies` and the build will still work.

PR #1933, opened 2025-07-07
[JAX] Support Flax sharding constraints

jberchtold-nvidia:
# Description

Updates our `with_sharding_constraint_by_logical_axes` helper function to support and prefer Flax logical axis rules when they exist in the current context. If no Flax logical axes rules exist in the current context, it will fall back to TE's hardcoded logical axes, though this functionality is no deprecated and will be removed in the future.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Update `with_sharding_constraint_by_logical_axes` to check if Flax logical axis rules exist in the current context, and if so call the sharding constraint with `flax.linen.with_logical_constraint`
- Update unfused attention in `transformers.py` to remove duplicate SEQLEN_AXIS usage in a single sharding constraint. This was not flagged before since our TE hardcoded axis system didn't check for this and it was always mapped to `None`/replicated.
- Corrected logical axis rule context setup in encoder examples

# Checklist:

### DIFF
--- a/examples/jax/encoder/test_model_parallel_encoder.py
+++ b/examples/jax/encoder/test_model_parallel_encoder.py
@@ -264,8 +264,10 @@ def train_and_evaluate(args):
     device_mesh = mesh_utils.create_device_mesh((num_gpu_dp, num_gpu_tp))
     with jax.sharding.Mesh(
         devices=device_mesh, axis_names=(DEVICE_DP_AXIS, DEVICE_TP_AXIS)
-    ) as mesh, nn_partitioning.axis_rules(
-        ((NAMED_BROADCAST_AXIS, None), (NAMED_TP_AXIS, DEVICE_TP_AXIS))
+    ) as mesh, te.fp8_autocast(
+        enabled=args.use_fp8,
+        fp8_recipe=fp8_recipe,
+        mesh_resource=te.MeshResource(DEVICE_DP_AXIS, DEVICE_TP_AXIS, None, None),
     ):
         rng = jax.random.PRNGKey(args.seed)
         rng, params_rng = jax.random.split(rng)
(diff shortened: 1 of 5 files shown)


transformer_engine/jax/sharding.py
```diff
@@ -133,6 +135,19 @@ def with_sharding_constraint_by_logical_axes(
...
+
+        flax_rules = flax.linen.get_logical_axis_rules()
+        if len(flax_rules) > 0:
+            return flax.linen.with_logical_constraint(
+                x, logical_axis_names, fallback=flax.linen.spmd.RulesFallback.NO_CONSTRAINT
+            )
+    except ImportError:
+        pass
+
+    # If no logical axis rules are available from Flax, fallback to TE's hardcoded logical axis rule table
```
phu0ngng, line 160: Could you raise a warning for soon-to-be deprecated here?

jberchtold-nvidia: Good idea, added a warning

examples/jax/encoder/test_model_parallel_encoder.py
```diff
@@ -261,36 +261,37 @@ def train_and_evaluate(args):
...
-            mesh_resource=te.MeshResource(DEVICE_DP_AXIS, DEVICE_TP_AXIS, None, None),
-        ):
+        # Get the base axis rules and extend them with TE's rules.
+        axis_rules = flax.linen.get_logical_axis_rules()
+        axis_rules += ((NAMED_BROADCAST_AXIS, None), (NAMED_TP_AXIS, DEVICE_TP_AXIS))
+        te_extended_axis_rules = te_flax.extend_logical_axis_rules(axis_rules)
+
+        with jax.sharding.Mesh(
+            devices=device_mesh, axis_names=(DEVICE_DP_AXIS, DEVICE_TP_AXIS)
+        ) as mesh, flax.linen.logical_axis_rules(te_extended_axis_rules):
```
phu0ngng: Hi, 
I'm curious to learn what the difference is between having the `jax.sharding.Mesh` context before vs after the `te.fp8_autocast` context.

jberchtold-nvidia: Afaik it doesn't matter the order of the Mesh vs. fp8_autocast, as long as both are before your model. But the order of `te_flax.extend_logical_axis_rules` and `fp8_autocast` does matter according to the docstring of the former.

So the ordering needs to be:
1. `fp8_autocast` context
2. Create Flax logical axis rule context. To support TE's hardcoded axis system in Flax (which UnfusedDotProductAttention requires), we extend the logical access rules with TE's rule table via `te_flax.extend_logical_axis_rules`, which must be inside an `fp8_autocast`.
3. Create and initialize the model, training loop, etc.

Afaik, the Mesh can come anywhere before item 3.

I just pulled up the `fp8_autocast` to the top and merged the `Mesh` with the `with` block for the logical axis rule context in item 2 to reduce indentation. But if a smaller diff is preferred, I can do Mesh -> fp8_autocast -> logical axis rules

phu0ngng: In this case, I would prefer we keep a minimal diff.

jberchtold-nvidia: Sounds good, updated to a smaller diff. Mesh -> fp8_autocast order is unchanged. fp8_autocast and logical axis rule order is changed as that was not implemented correctly before this PR.

phu0ngng (approved):

PR #1937, opened 2025-07-08
[Common] Skip cuDNN 9.10.0/9.10.1 due to bugs

cyanguwa:
# Description

cuDNN fixed a few bugs in cuDNN 9.10.2 and this PR is to:
- skip cuDNN 9.10.0 for SDPA FP8 related bugs,
- skip cuDNN 9.10.0 and 9.10.1 for SDPA FP16/BF16 related bugs,
- unify the use of `ModelConfig` in all unit tests,
- fix FP8 and certain FP16/BF16 tests so they check backend availability first before running the configs,
- move `test_sanity_attention_extra_state` from `test_sanity.py` to `test_fused_attn.py`,
- remove `test_kv_cache_accuracy` from `test_numerics.py` because it's duplicate to `test_kv_cache` in `test_kv_cache.py`, and
- rename `tests/pytorch/fused_attn` to `tests/pytorch/attention` and other relevant files from `fused_attn` to `attention`.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Please see description.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/base.py
+++ b/transformer_engine/pytorch/module/base.py
@@ -42,7 +42,7 @@ from ..tensor.mxfp8_tensor import MXFP8Quantizer
 from ..tensor.float8_blockwise_tensor import Float8BlockQuantizer
 from ..tensor._internal.float8_tensor_base import Float8TensorBase
 from ..tensor._internal.mxfp8_tensor_base import MXFP8TensorBase
-from ..utils import torch_get_autocast_gpu_dtype
+from ..utils import is_non_tn_fp8_gemm_supported, torch_get_autocast_gpu_dtype
 from ..tensor._internal.float8_blockwise_tensor_base import Float8BlockwiseQTensorBase
 from ...common.recipe import DelayedScaling, Recipe
 from ...debug.pytorch.debug_state import TEDebugState
@@ -1293,21 +1293,29 @@ class TransformerEngineBaseModule(torch.nn.Module, ABC):
 
         # Try getting workspace from cache
         out = None
-
         if cache_name is not None:
             out = self._fp8_workspaces.get(cache_name, None)
-            if quantizer is not None and isinstance(out, MXFP8TensorBase):
-                if quantizer.rowwise_usage and out._rowwise_data is None:
-                    out = None
-                    del self._fp8_workspaces[cache_name]
-                elif quantizer.columnwise_usage and out._columnwise_data is None:
-                    out = None
-                    del self._fp8_workspaces[cache_name]
 
-            is_debug = isinstance(quantizer, DebugQuantizer)
-            is_out_debug_tensor = out is not None and isinstance(out, DebugQuantizedTensor)
-            if is_debug != is_out_debug_tensor:
+        # Reset cache if workspace is invalid
+        if out is not None and quantizer is not None:
+            reset_cache = False
+            if isinstance(out, Float8TensorBase):
+                if (
+                    not is_non_tn_fp8_gemm_supported()
+                    and quantizer.columnwise_usage
+                    and out._transpose is None
+                ):
+                    reset_cache = True
+            elif isinstance(out, MXFP8TensorBase):
+                if quantizer.rowwise_usage and out._rowwise_data is None:
+                    reset_cache = True
+                elif quantizer.columnwise_usage and out._columnwise_data is None:
+                    reset_cache = True
+            if isinstance(out, DebugQuantizedTensor) != isinstance(quantizer, DebugQuantizer):
+                reset_cache = True
+            if reset_cache:
                 out = None
+                del self._fp8_workspaces[cache_name]
 
         # Gather cached Fp8 workspace if it's distributed
         # NOTE: FSDP sharding is supported only for Fp8 buffers and will not work


cyanguwa:
Pipeline 31306522 for 9.10.0 + 25.06 + 12.9
Pipeline 31306722 for 9.10.1 + 25.06 + 12.9

tests/pytorch/test_sanity.py
```diff
@@ -47,6 +47,7 @@
 from transformer_engine.pytorch.tensor.utils import replace_raw_data
 from transformer_engine.pytorch.distributed import checkpoint
 from test_numerics import reset_rng_states, dtype_tols
+from fused_attn.test_fused_attn import ModelConfig, _get_attention_backends
```
timmoon10: `test_fused_attn.py` is kind of a strange place to keep `ModelConfig`. It's indicating that these tests are becoming over-tuned to attention. We could avoid this refactor by moving `test_sanity_attention_extra_state` into the fused attention tests:
https://github.com/NVIDIA/TransformerEngine/blob/7b249ae82d92432936619fbe51f119c5f148a89c/tests/pytorch/test_sanity.py#L1119

Alternatively, we could move `ModelConfig` from `test_fused_attn.py` into `utils.py`. It has many attention-specific options though, so I'd prefer having separate implementations.

zhongbozhu (dismissed):
LGTM

tests/pytorch/test_sanity.py
cyanguwa: I've moved `test_sanity_attention_extra_state` to `test_fused_attn.py`, which is now renamed to `test_attention.py`. I've also moved `ModelConfig`, and a couple of other utility functions, to `tests/pytorch/utils.py`. Please take a look and let me know if the structure is all right! `ModelConfig` has a lot of parameters which seem to be closely related to attention but I think we need them for other tests as well, like `hidden_size`, `head_dim`, and `batch_size`. I plumbed it through and I don't think it's too clunky. Anyway, let me know!

tests/pytorch/utils.py
```diff
@@ -4,12 +4,31 @@
...
+    AttentionLogging,
+)
+from transformer_engine.pytorch.cpp_extensions.fused_attn import FusedAttnBackend
+
+# Initialize RNG state
+seed = 1234
+torch.manual_seed(seed)
+torch.cuda.manual_seed(seed)
+_cpu_rng_state = torch.get_rng_state()
+_cuda_rng_state = torch.cuda.get_rng_state()
```
timmoon10: Importing a module shouldn't have side effects. How about something like:
```python
# Cached RNG state
_rng_state: Optional[Tuple[torch.Tensor, torch.Tensor]] = None

def reset_rng_state() -> None:
    """Revert to deterministic RNG state"""
    global _rng_state
    if _rng_state is None:
        torch.manual_seed(1234)
        torch.cuda.manual_seed(1234)
        _rng_state = (torch.get_rng_state(), torch.cuda.get_rng_state())
    else:
        cpu_rng_state, cuda_rng_state = _rng_state
        torch.set_rng_state(cpu_rng_state)
        torch.cuda.set_rng_state(cuda_rng_state)
```
Each test would need to call `reset_rng_state` explicitly if it wanted deterministic RNG state.

tests/pytorch/test_sanity.py
timmoon10: There are many attention-specific args in `ModelConfig`. For linear layers, the only configs that matter are the BSHD dims so we can initialize the data tensors. I still think it's better to have a basic config in `test_numerics.py` and an attention-specific config in `test_attention.py`. At the least, `ModelConfig` should have defaults so `test_numerics.py` doesn't need much attention-specific knowledge.

tests/pytorch/test_numerics.py
```diff
@@ -124,6 +114,18 @@ def __init__(self, hidden_size, eps, num_attention_heads, embed, num_layers, seq
...
+def is_fused_attn_available(
+    config: ModelConfig, dtype: torch.dtype, qkv_layout="bshd_bshd_bshd", is_training=True
+):
+    available_backends, _, fused_attn_backends = get_available_attention_backends(
+        config,
+        qkv_dtype=dtype,
+        qkv_layout=qkv_layout,
+        is_training=is_training,
+    )
+    return FusedAttnBackend["F16_arbitrary_seqlen"] in fused_attn_backends
```
timmoon10, line 123: I'm uncomfortable that tests in `test_numerics.py` require knowledge of attention backends and would prefer if they were in `test_attention.py` instead. However, I see that this is needed for some of the full GPT model tests, which do belong in this file.

timmoon10 (commented):
Perhaps we should split this PR into two: skipping the buggy cuDNN versions and refactoring the attention tests.

tests/pytorch/utils.py
cyanguwa: I incorporated this in the code. Thanks!

tests/pytorch/test_numerics.py
cyanguwa: Any test that's based on `TransformerLayer` will need some knowledge of attention, and there are a lot of those tests even outside `tests/pytorch/attention`, like `test_numerics.py`, `test_cpu_offload.py`, `test_cuda_graphs.py` and `test_sanity.py`. I feel it's difficult to pull them apart - at the end of the day, we are working with LLMs. I've simplified `ModelConfig` a bit though, so for some of the tests, users only need to specify ~5 params to instantiate a `ModelConfig`.

tests/pytorch/test_sanity.py
cyanguwa: Please see comment below.

tests/pytorch/utils.py
```diff
@@ -4,12 +4,29 @@
...
+    get_attention_backend,
+    AttentionParams,
+    AttentionLogging,
+)
+from transformer_engine.pytorch.cpp_extensions.fused_attn import FusedAttnBackend
+
+# Initialize RNG state
+seed = 1234
+torch.manual_seed(seed)
+torch.cuda.manual_seed(seed)
```
timmoon10: Importing a module shouldn't affect the global RNG state, since that could cause the RNG to change at unexpected times.
```suggestion
```

tests/pytorch/test_numerics.py
```diff
@@ -59,30 +61,18 @@
 torch.manual_seed(seed)
 torch.cuda.manual_seed(seed)
 # Record initial RNG state from script run.
-_cpu_rng_state = torch.get_rng_state()
-_cuda_rng_state = torch.cuda.get_rng_state()
+_rng_states = None
```
timmoon10: Setting the RNG state is redundant with `reset_rng_states`. 
```suggestion
```
There are a few other files where we redundantly set the RNG state.

tests/pytorch/utils.py
```diff
@@ -106,3 +123,184 @@ def make_recipe(name: Optional[str]) -> Optional[Recipe]:
...
+
+class ModelConfig:
+    def __init__(
+        self,
+        batch_size: int,
+        num_heads: int,
+        num_gqa_groups: int,
+        head_dim_qk: int,
+        max_seqlen_q: int,
+        max_seqlen_kv: int,
```
timmoon10: Nit: I think `num_gqa_groups=num_heads` and `max_seqlen_kv=max_seqlen_q` are reasonable defaults. This way we have the minimum number of args needed for a basic Transformer layer.
```suggestion
    def __init__(
        self,
        batch_size: int,
        num_heads: int,
        head_dim_qk: int,
        max_seqlen_q: int,
        num_gqa_groups: Optional[int] = None,
        max_seqlen_kv: Optional[int] = None,
```

tests/pytorch/test_numerics.py
timmoon10: Having good defaults in `ModelConfig` definitely helps a lot. I don't mind having attention-specific stuff internally, as long as it doesn't affect anything non-attention-related.

tests/pytorch/test_numerics.py
```diff
@@ -40,11 +40,13 @@
 from transformer_engine.pytorch.attention.inference import InferenceParams
 from transformer_engine.pytorch.distributed import checkpoint as te_checkpoint
 from transformer_engine.pytorch.cpp_extensions import general_gemm, general_grouped_gemm
+from transformer_engine.pytorch.cpp_extensions.fused_attn import FusedAttnBackend
 from transformer_engine.pytorch.tensor.float8_tensor import Float8Quantizer
 from transformer_engine.pytorch.module.base import get_multi_stream_cublas_workspace, get_workspace
 from transformer_engine.pytorch.utils import get_device_compute_capability, get_cudnn_version
 from transformer_engine.common import recipe
 import transformer_engine_torch as tex
+from utils import ModelConfig, _rng_states, reset_rng_states, get_available_attention_backends
```
timmoon10: No need to import `_rng_states` unless we want to do something with it. It's an implementation detail of `reset_rng_states`, so importing it should be viewed as an evil hack.
```suggestion
from utils import ModelConfig, reset_rng_states, get_available_attention_backends
```
We import it in a few other files.

tests/pytorch/utils.py
cyanguwa: I've shuffled the params a bit and made some kwargs.

timmoon10 (approved):
LGTM

timmoon10 (approved):

tests/pytorch/test_sanity.py
```diff
@@ -105,37 +103,19 @@ def custom_amax_compute(amax_history: torch.Tensor) -> torch.Tensor:
...
-    kv_channels: Optional[int] = None
-
-    def is_fp8_supported(self):
-        if self.seq_len * self.batch_size % 16:
-            return False
-        if self.hidden_size % 16:
-            return False
-        return True
+def is_fp8_supported(config: ModelConfig):
+    if config.max_seqlen_q * config.batch_size % 16 or config.max_seqlen_q * config.batch_size % 16:
```
timmoon10:
```suggestion
    if config.max_seqlen_q * config.batch_size % 16 or config.max_seqlen_kv * config.batch_size % 16:
```

tests/pytorch/utils.py
```diff
@@ -106,3 +118,184 @@ def make_recipe(name: Optional[str]) -> Optional[Recipe]:
...
+    else:
+        cpu_rng_state, cuda_rng_state = _rng_states
+        torch.set_rng_state(cpu_rng_state)
+        torch.cuda.set_rng_state(cuda_rng_state)
+
+
+@pytest.fixture(autouse=True)
+def reset_global_fp8_state():
+    yield
+    fp8.FP8GlobalStateManager.reset()
```
timmoon10: Are we trying to apply this fixture in every test that includes `utils.py`? This is unintuitive and may break some tests. Also, I'm not sure if PyTest is smart enough to detect fixture registration from imported modules.
```suggestion
```
For now, it would probably be easiest to duplicate the code in each test like before.

cyanguwa: Apparently, fixture import is not recommended. If an import happens, the fixture will be registered as defined in that particular module instead of shared by the importing module as well.

PR #1944, opened 2025-07-11
[BUG fix] solve the CI bug of router fusion

Autumn1998:
# Description

Fixes for Router Fusion:
1. Corrected kernel dispatching issue on SM90 architecture.
2. Resolved CI bug by using properly constructed input data in unit tests.
3. Fixed bug occurring when MBS > 1.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/test_fused_router.py
+++ b/tests/pytorch/test_fused_router.py
@@ -148,11 +148,15 @@ def run_comparison(
 ):
     # Set some parameters
     if score_function == "sigmoid":
-        logits = torch.arange(num_experts, device="cuda", dtype=dtype) * 0.1
+        # Construct the special logits to avoid inf in the sigmoid function
+        logits = 1 + torch.arange(num_experts, device="cuda", dtype=dtype) * 0.01
         logits = logits.unsqueeze(0).repeat(num_tokens, 1)
+        random_values = torch.rand(num_tokens, num_experts, device="cuda")
+        _, indices = torch.sort(random_values, dim=1)
+        logits = torch.gather(logits, 1, indices)
     else:
         logits = torch.arange(num_tokens * num_experts, device="cuda", dtype=dtype) * 1e-4
-    logits = logits.view(num_tokens, num_experts)
+        logits = logits.view(num_tokens, num_experts)
     logits.requires_grad = True
     if enable_bias and score_function == "sigmoid":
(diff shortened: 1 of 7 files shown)


tests/pytorch/test_fused_router.py
```diff
@@ -148,11 +148,15 @@ def run_comparison(
 ):
     # Set some parameters
     if score_function == "sigmoid":
-        logits = torch.arange(num_experts, device="cuda", dtype=dtype) * 0.1
+        # Construct the special logits to avoid inf in the sigmoid function
+        logits = 1 + torch.arange(num_experts, device="cuda", dtype=dtype) * 0.01
         logits = logits.unsqueeze(0).repeat(num_tokens, 1)
+        random_values = torch.rand(num_tokens, num_experts, device="cuda")
+        _, indices = torch.sort(random_values, dim=1)
+        logits = torch.gather(logits, 1, indices)
```
timmoon10: Can you explain how this helps avoid infs? Was the problem that `logits` became too large, causing sigmoid to output 1.0 in multiple entries? Also, I don't understand the benefit from forcing `logits >= 1`. Previously we had `logits >= 0` and `sigmoid(0) = 0.5`, so I wouldn't expect us to have had divide-by-zero problems.

timmoon10 (approved):
Overall LGTM, thanks for the fix!

tests/pytorch/test_fused_router.py
Autumn1998: @timmoon10  There are some changes in this part.
In summary, we need to consider two main issues:

1. When using `top-k` on `M` values, the `k-th` and `(k+1)-th` value can sometimes be the same. This may cause fused implementation and PyTorch’s implementation to produce different results. To fix this, we need to ensure that all input values in each row are unique. (Note: using the random API is unreliable even if the seed is fixed.)
2. The sigmoid function can easily produce `inf` if the input is too large, due to the exp(-x)

To fix these, I created a special input matrix that satisfies both conditions and reduced the problem size. There is no need to enforce `logits >=1`; I only used that constraint in the initial version to find a suitable value range to avoid the inf.

yaox12 (approved):

PR #1946, opened 2025-07-11
[JAX] Select cuDNN backend for normalization

phu0ngng:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/include/transformer_engine/normalization.h
```diff
@@ -149,6 +149,10 @@ void nvte_rmsnorm_bwd(const NVTETensor dz, const NVTETensor x, const NVTETensor
 void nvte_enable_cudnn_norm_fwd(bool enable);
 void nvte_enable_cudnn_norm_bwd(bool enable);
 
+/*! \brief Helper to query if cuDNN backend for normalization is enabled */
+int nvte_use_cudnn_norm_fwd();
```
timmoon10, line 158: These functions are nifty. We could use them in the PyTorch extensions as well:
https://github.com/NVIDIA/TransformerEngine/blob/214e2a4a2fef37f14e78bd31701d7a57a1a5488f/transformer_engine/pytorch/csrc/extensions/normalization.cpp#L105
https://github.com/NVIDIA/TransformerEngine/blob/214e2a4a2fef37f14e78bd31701d7a57a1a5488f/transformer_engine/pytorch/csrc/extensions/normalization.cpp#L266

transformer_engine/common/include/transformer_engine/normalization.h
```diff
@@ -149,6 +149,10 @@ void nvte_rmsnorm_bwd(const NVTETensor dz, const NVTETensor x, const NVTETensor
...
+int nvte_use_cudnn_norm_bwd();
```
timmoon10: We should comment on both functions for Doxygen to do the right thing:
```suggestion
/*! \brief Whether cuDNN backend is enabled for LayerNorm/RMSNorm forward pass */
int nvte_use_cudnn_norm_fwd();
/*! \brief Whether cuDNN backend is enabled for LayerNorm/RMSNorm backward pass */
int nvte_use_cudnn_norm_bwd();
```

timmoon10 (commented):
How about we change the default to `NVTE_NORM_FWD_USE_CUDNN=1` (and similar for backward)? We can change:
https://github.com/NVIDIA/TransformerEngine/blob/214e2a4a2fef37f14e78bd31701d7a57a1a5488f/transformer_engine/common/normalization/common.cpp#L504-L507
Into:
```c++
 bool& _cudnn_norm_fwd_flag() { 
   static bool flag = transformer_engine::getenv<bool>("NVTE_NORM_FWD_USE_CUDNN", true); 
   return flag; 
 } 
```
This way JAX users can still disable cuDNN with `NVTE_NORM_FWD_USE_CUDNN=0`.

phu0ngng:
> How about we change the default to `NVTE_NORM_FWD_USE_CUDNN=1` (and similar for backward)?

@timmoon10, I think the question here is whether we want to turn CuDNN backend on by default.

We know that cuDNN backend works faster for MLPerf cases, but we have not conducted a systematic benchmark for all the size configs and dtypes we have in the tuned kernel.

transformer_engine/jax/csrc/extensions/normalization.cpp
```diff
@@ -36,6 +36,9 @@ pybind11::tuple GetNormForwardWorkspaceSizes(size_t batch_size, size_t hidden_si
     output_tensor.set_columnwise_data(static_cast<void *>(&temp), out_dtype, input_shape);
   }
 
+  // Enable norm with cuDNN instead of TE kernels
+  nvte_enable_cudnn_norm_fwd(true);
```
timmoon10, line 40: Is there a reason we are forcing the cuDNN backend? It seems like something that should be controlled with `NVTE_NORM_FWD_USE_CUDNN`. If cuDNN has better perf, it would be better to change the default to `NVTE_NORM_FWD_USE_CUDNN=1` rather than hard-coding it only for JAX.

phu0ngng:
Close as the change is no longer necessary.

PR #1950, opened 2025-07-14
[PyTorch][FP8 CS] Remove the unnecessary torch reciprocal op in fp8 current scaling code path

zhongbozhu:
# Description

`at::reciprocal` was part of FP8 current scaling recipe to make sure TPGEMM overlap works with FP8 current scaling. That's because the scales are not copied into user buffer for communication, but the old TPGEMM overlap codebase relies on the `create_tensor` API to construct the output tensor which requires a scaling factor tensor in it. 

Now that dependency has been removed, we no longer need this op.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/csrc/quantizer.cpp
+++ b/transformer_engine/pytorch/csrc/quantizer.cpp
@@ -220,8 +220,9 @@ std::pair<TensorWrapper, py::object> Float8CurrentScalingQuantizer::create_tenso
   }
   const py::object py_columnwise_data = create_transpose ? py::cast(columnwise_data) : py::none();
 
-  // In current scaling, scale is not known but we initialize it with 1 to avoid division by zero. If scale is already calculated, it can be correctly set.
-  at::Tensor scale_inv = at::reciprocal(scale);
+  // In current scaling, scale_inv is not known but we initialize it as an empty tensor to be filled later.
+  at::Tensor scale_inv =
+      at::empty({1}, at::TensorOptions().dtype(torch::kFloat32).device(torch::kCUDA));
 
   py::object ret;
   if (internal) {
--- a/transformer_engine/pytorch/tensor/float8_tensor.py
+++ b/transformer_engine/pytorch/tensor/float8_tensor.py
@@ -215,7 +215,7 @@ class Float8CurrentScalingQuantizer(Quantizer):
         amax_epsilon: float = 0.0,
     ) -> None:
         super().__init__(rowwise=rowwise, columnwise=columnwise)
-        self.scale = torch.ones(1, dtype=torch.float32, device=device)
+        self.scale = torch.empty(1, dtype=torch.float32, device=device)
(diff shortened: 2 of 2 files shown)


BestJuly:
LGTM.

zhongbozhu:
@timmoon10 Seems like this change will break test_fusible_ops.py, do you know about the reason?

timmoon10:
~It seems like numerical errors in the `tex` functions for activation forward. I wonder if the `scale_inv` isn't being properly updated, so it was relying on the reciprocal in `create_tensor`.~ I expect `LayerNormMLP` also experiences these numerical problems, but its test tolerances are much looser.

The root cause is because the activation backward pass doesn't support FP8 current scaling: https://github.com/NVIDIA/TransformerEngine/blob/e7251f93673a174bb7ead16934ed548d1f6133e3/transformer_engine/pytorch/csrc/extensions/activation.cpp#L76

zhongbozhu:
> ~It seems like numerical errors in the `tex` functions for activation forward. I wonder if the `scale_inv` isn't being properly updated, so it was relying on the reciprocal in `create_tensor`.~ I expect `LayerNormMLP` also experiences these numerical problems, but its test tolerances are much looser.
> 
> The root cause is because the activation backward pass doesn't support FP8 current scaling:
> 
> https://github.com/NVIDIA/TransformerEngine/blob/e7251f93673a174bb7ead16934ed548d1f6133e3/transformer_engine/pytorch/csrc/extensions/activation.cpp#L76

Thanks for the root causing! I don't think layernorm_mlp will have this issue because the place where we call the dact function, the quantizer we gave is None [here ](https://github.com/NVIDIA/TransformerEngine/blob/c0c12e204203208c3a166d73377e63b8f3167cb4/transformer_engine/pytorch/module/layernorm_mlp.py#L986C28-L986C47)=> so it's actually a None quantizer, calculating dact for the high precision, and then we call this line to do the quantization again [here.](https://github.com/NVIDIA/TransformerEngine/blob/c0c12e204203208c3a166d73377e63b8f3167cb4/transformer_engine/pytorch/module/layernorm_mlp.py#L992) So yes, the fusion is not supported, but layernorm_mlp has already taken that into account, in an implicit way of course.

zhongbozhu:
@timmoon10 has the numerical issue in fused_op side been resolved? If it's already done, I can rebase and run the CI again.

timmoon10:
https://github.com/NVIDIA/TransformerEngine/pull/1952 is a comprehensive fix that removes the `at::reciprocal` and properly handles the activation backward with FP8 current scaling.

zhongbozhu:
Update: 

Will close this PR once https://github.com/NVIDIA/TransformerEngine/pull/1952 is merged, which is a superset of this fix.

PR #1951, opened 2025-07-15
Refactor `te.ops`

janekb04:
# Description

This PR refactors some implementation aspects of `te.ops`.

It has no measurable effect on performance, except for a potential benefit when having `input.requires_grad=False` and some frozen parameters.

Additionally, it allows for passing using extra inputs and outputs with `te.ops.Sequential`s compose of multiple module groups.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- (fecadb08caea74eb0801a6b4b1690a777d690310) Remove `is_grad_enabled` parameter from `_OperationFuserAutogradFunction.forward` and use `func_ctx is not None` instead
- (45937b7eafba8f16d02a32ca9550186d94770724) Rename `BackwardBiasActivation` to `BackwardActivationBias` to reflect the order of operations in forward
- (40e7220a2fac5c52861808b2f97c7540548a80f9) Reverse order in which basic operations are stored in `BackwardLinearAdd` and `UserbuffersBackwardLinear` to be the forward order rather than the backward order
- (971772af29930e28ca6e49c0264f0738e20c94c8) Rename `prev_op_grad_input_quantizer` to `prev_op_grad_output_quantizer` to reflect the fact that the quantizer is meant to be used for the incoming gradient of the output
- (65e6367c678d41fa2e2d55224d1794ebb15783d1) Make `OperationFuser` persistent and refuse internally on recipe change instead of recreating it in `te.ops.Sequential`
- (fd5a918279f188e378e11575387919a2fd76fc36) Allow for passing using extra inputs and outputs with `te.ops.Sequential`s compose of multiple module groups. Also added a comprehensive test for this.
- (975ba59d6e5b2217b7592a378103f5b0841d2c1b) Make `OperationFuser` refuse on changes of `requires_grad` to input, parameters, and extra inputs and take them it into account when fusing. This could bring a performance improvement for cases when the input does not require grad and some parameters are frozen as the backward passes of some basic operations will be skipped.
- (55c2d4a610e02bc5e87fc8ba498e806f03b030d8) Make `get_quantizer` return `None` when quantization is not used enabling its users to no longer have to check if quantization is enabled.
  - Also changes `test_fp8_scale_update_with_linear_fuser_op` to get quantizers from `BasicLinear` only after it's quantizers had the chance to be created (as `get_quantizer` no longer creates them implicitly) (as well as make that test also check the amax history, rather than only the scaling factors as it previously did).
  - Also changes `make_graphed_callables` to accept an `Optional[Recipe]` rather than an `Optional[DelayedScaling]`, given it already supported different recipes. Also makes `make_graphed_callables` respect the `fp8_enabled` parameter and use `None` recipe if `fp8_enabled is False`.
- (ecc270dca9f9d4bb24b0fd25b7fcf528a82b0d14) Refactor `pre_first_forward`: split it into `pre_first_fuser_forward` (renamed to emphasize that it is called when the `OperationFuser` thinks is the first iteration) and to `reset_recipe_type` called when the recipe type changes. Made `pre_first_fuser_forward` independent of current recipe type (as that can changed and is supposed to be handled by `reset_recipe_type` instead.

# Checklist:


transformer_engine/pytorch/ops/fuser.py
```diff
@@ -6,18 +6,19 @@
 
 from __future__ import annotations
 from collections.abc import Callable
-from typing import Any, Optional
+from typing import Any, Optional, Iterable
```
timmoon10: Nit: [`typing.Iterable`](https://docs.python.org/3/library/typing.html#typing.Iterable) has been deprecated in favor of [`collections.abc.Iterable`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Iterable). We [import from `__future__`](https://peps.python.org/pep-0585/#implementation), so the modernized version should be safe. 
```suggestion
from collections.abc import Callable, Iterable
from typing import Any, Optional
```

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -36,7 +37,7 @@ def _split_tuple(t: tuple, idx: int) -> tuple[tuple, tuple]:
 
 
 # Lazily imported function used in _is_graph_capturing
-_is_graph_capturing_function: Optional[Callable[[], bool]] = None
+_is_graph_capturing_function: Optional[Callable[[], bool]] | None = None
```
timmoon10: Nit: `Optional[X]` is equivalent to `X | None`, and more explicit about intent.
```suggestion
_is_graph_capturing_function: Optional[Callable[[], bool]] = None
```

transformer_engine/pytorch/ops/fused/backward_activation_bias.py
```diff
@@ -29,7 +29,7 @@
 _fusible_activations = tuple(_fused_activations.keys())
 
 
-class BackwardBiasActivation(FusedOperation):
+class BackwardActivationBias(FusedOperation):
     """Fused backward dbias + dact + quantize
```
timmoon10: Nit:
```suggestion
    """Fused backward dact + dbias + quantize
```

transformer_engine/pytorch/ops/op.py
```diff
@@ -214,120 +210,86 @@ def get_input_quantizer(self) -> Optional[Quantizer]:
...
-            if self._fp8_metas[mode] is None or fp8_meta_key not in self._fp8_metas[mode]:
-                continue
-            recipe_state = self._fp8_metas[mode][fp8_meta_key]
-            need_to_reset_recipe_state = (
-                (recipe.delayed() and not isinstance(recipe_state, DelayedScalingRecipeState))
-                or (recipe.mxfp8() and not isinstance(recipe_state, MXFP8BlockScalingRecipeState))
-                or (
-                    recipe.float8_block_scaling()
-                    and not isinstance(recipe_state, Float8BlockScalingRecipeState)
+            if FP8GlobalStateManager.is_fp8_enabled() and self.num_quantizers(mode):
```
timmoon10: This doesn't match the behavior of the fused modules, which don't register their `fp8_meta`s when capturing a CUDA graph:
https://github.com/NVIDIA/TransformerEngine/blob/86c5097770977542b587f30bf49a20e85886092f/transformer_engine/pytorch/module/base.py#L1012
I see this change is from https://github.com/NVIDIA/TransformerEngine/pull/1951/commits/c9690a42888c5a5b33f054c662566de47d12e16b, which is attempting to fix a CUDA graph test failure. I think the root cause is because `make_graphed_callables` has special logic to handle fused modules, but not `te.Sequential` ops:
https://github.com/NVIDIA/TransformerEngine/blob/86c5097770977542b587f30bf49a20e85886092f/transformer_engine/pytorch/graph.py#L678-L701

I believe it was designed this way because we want to handle `fp8_meta` logic (like amax reductions and misc bookkeeping) outside of the CUDA Graph.

transformer_engine/pytorch/ops/fuser.py
janekb04: Yes, I'm not sure how this change got made.

timmoon10 (approved):
LGTM. This PR definitely makes the code structure more logical and robust. In particular, this will enable `requires_grad`-aware fusions.

PR #1954, opened 2025-07-16
[JAX] Remove unneccessary MXFP8 scale_inv padding 

phu0ngng:
# Description

This PR removes unnecessary padding steps in the MXFP8 workflow, which should improve the recipe performance.

Right now, we do
```
# In layernorm/activation/quantization
1. Allocate padded scale in inner_primitive.abstract() 
2. Slice out the padding in Primitive.impl(), after calling the inner_primitve
3. Add padding in ScaledTensor.__post_init__()

# In GEMM custom call
4. Remove padding before calling the custom call
5. Adding padding in the Primitive.impl() before calling the inner_primitive
```

This PR removes unnecessary padding steps, the new workflow becomes:
```
# In layernorm/activation/quantization
1. Allocate padded scale in inner_primitive.abstract() 
2. Slice out the padding in Primitive.impl(), after calling the inner_primitve

# In GEMM custom call
5. Adding padding in the Primitive.impl() before calling the inner_primitive
```
(3) and (4) are removed.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:

- [] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/transformer_engine/jax/quantize/dequantizer.py
+++ b/transformer_engine/jax/quantize/dequantizer.py
@@ -121,9 +121,6 @@ class BlockScaleDequantizer(Dequantizer):
         scale_shape = scaling_mode.get_scale_shape(
             data_shape, is_colwise, is_padded=False, flatten_axis=flatten_axis
         )
-        scale_inv = jax.lax.slice(
-            scale_inv, [0] * len(scale_shape), scale_shape
-        )  # slice out the padding
 
         data = data.reshape(
             *data_shape[: flatten_axis - 1],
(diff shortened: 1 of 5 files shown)


transformer_engine/jax/quantize/tensor.py
```diff
@@ -136,15 +135,6 @@ def __post_init__(self):
         if self.scaling_mode == ScalingMode.NO_SCALING:
             self.scale_inv = jnp.empty((0,), dtype=jnp.float32)
 
-        else:
-            self.scale_inv = apply_padding_to_scale_inv(
```
jberchtold-nvidia, line 140: Removing this call to `apply_padding_to_scale_inv` means we are no longer checking if the incoming scale_inv is valid based on the data shape (either padded or unpadded).

Since we're now exclusively supporting unpadded scales here, can we add a check to ensure scale_inv.shape == expected_unpadded_shape given self.data.shape

transformer_engine/jax/quantize/dequantizer.py
```diff
@@ -211,28 +208,38 @@ def _grouped_dequantize(grouped_scaled_tensor):
...
             flatten_axis=flatten_axis,
         )
-        scale_shape_i_size = math.prod(scale_shape_i)
-        scale_inv_i = scale_inv[scale_inv_ptr : scale_inv_ptr + scale_shape_i_size]
+        unpadded_scale_shape_i = scaling_mode.get_scale_shape(
+            data_shape_i,
+            grouped_scaled_tensor.is_colwise,
+            is_padded=False,
+            flatten_axis=flatten_axis,
+        )
```
jberchtold-nvidia, line 222: Hi, I'm not familiar on the differences between grouped DQ and non-grouped DQ w.r.t padding. For the non-grouped DQ case above we can just remove the slicing logic, but for the grouped case the slicing logic is still here with some additional logic. Can you help me understand the difference for this case? Thanks!

jberchtold-nvidia (commented):
Overall looks good, will cherry-pick and benchmark. Left a couple small comments

transformer_engine/jax/quantize/tensor.py
phu0ngng: I will add it.

transformer_engine/jax/quantize/dequantizer.py
phu0ngng: When quantizing a group of tensors to a `GroupedScaledTensor` via `grouped_quantize`, we actually don't know the dimension of each individual tensor at the JAX level.
So we need to allocate an extra buffer on the JAX side, go to the C++ side, query the tensor size info, then get the scale_inv and write to the given buffer. All the obtained scale_inv-s are padded.
 
We can't remove this padding in the workflow; thus, we need to remove it in the dequantization so that the dequantized data is correct.
In the past, we removed this padding in the `BlockScaleDequantizer._dequantize_func()`. Now, we no longer need it there as the `ScaledTensor` will host the unpadded scales, I moved this padding removal to the `_grouped_dequantize()`.

jberchtold-nvidia: Got it, makes sense. Thanks for the explanation!

jberchtold-nvidia (approved):

PR #1956, opened 2025-07-16
[PyTorch][Mcore] Fix illegal memory access issue while using Mcore async checkpoint with fp8 tensorwise recipe

zhongbozhu:
# Description

Replace the `tex.dequantize` with a torch implementation fixes the issue for now. Need to root cause why `tex.dequantize` did not work for the async cp, but working for the sync cp. 

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
+++ b/transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py
@@ -20,9 +20,21 @@ from ..quantized_tensor import Quantizer
 
 from ...utils import is_non_tn_fp8_gemm_supported, _empty_tensor
 
+from ...constants import TE_DType_To_Torch
+
 
 class _FromFloat8Func(torch.autograd.Function):
     """Cast from FP8 to other dtype"""
+
+    @staticmethod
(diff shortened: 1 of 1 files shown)


timmoon10 (changes requested):
We should understand the root cause of why the FP8 dequantize is failing.

zhongbozhu:
Update:

tex.dequantize not working for Mcore's async cp is likely because the kernel is launched in a spawned process instead of a forked process. Calling a torch implementation can get away with it since pytorch internally might handle this case well. 

We can either merge this PR but it's likely that we will do the same thing for the dequantize kernel of mxfp8 recipe. Or we can root cause why the kernel launch doesn't work in the spawned process and fix this issue once for all.

timmoon10 (changes requested):
This is a hack and it will hurt perf. We should root cause the issue.

zhongbozhu:
This issue has been fixed in the Mcore side: https://github.com/NVIDIA/Megatron-LM/pull/3845
Closing this PR.

PR #1962, opened 2025-07-17
[JAX] Helper to disable TE custom calls + disable GemmPrimitive for non-MXFP8 recipes.

phu0ngng:
# Description

This PR introduces a helper - `manage_primitive()` - to enable/disable TE custom calls in a recipe.

Besides, from our E2E tests, the TE GEMM custom call shows better performance in MXFP8 while JAX `dot`/`scaled_matmul` performs better in other cases. Thus, this PR sets TE/JAX to use JAX `dot`/`scaled_matmul` for non-MXFP8 recipes.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or new content)
- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:


transformer_engine/jax/cpp_extensions/base.py
```diff
@@ -30,19 +31,41 @@ class BasePrimitive(metaclass=ABCMeta):
...
+        """
+        # Check if environment variable is explicitly set
+        pattern_str = os.getenv("NVTE_JAX_CUSTOM_CALLS_RE")
+        if pattern_str is not None:
+            pattern = re.compile(pattern_str)
+            env_enabled = pattern.fullmatch(cls.__name__) is not None
+            return env_enabled
+
+        # If environment variable is not set, fall back to the internal state
+        enabled_state = getattr(cls, "_is_enabled", True)
```
jberchtold-nvidia: nit: Does this need `getattr`? Since all classes use `BasePrimitive` as a base, they'll all have the `_is_enabled` attr defined unless they explicitly delete it from themselves, right?

transformer_engine/jax/cpp_extensions/base.py
```diff
@@ -30,19 +31,41 @@ class BasePrimitive(metaclass=ABCMeta):
...
+
     @classmethod
     def enabled(cls):
         """
-        A custom call is marked as disabled if the `cls.__name__` does not fully match the
-        `NVTE_JAX_CUSTOM_CALLS_RE` pattern.
-        This uses the Python class name of the primitive definitions that inherit from BasePrimitive.
-        By default, `NVTE_JAX_CUSTOM_CALLS_RE` is set to `.*`, which matches and enables all names.
-        For example, set `NVTE_JAX_CUSTOM_CALLS_RE='^(?!DBiasQuantizePrimitive$).+$'` to disable `DBiasQuantizePrimitive`.
+        Determines if a custom call is enabled based on a state variable and the `NVTE_JAX_CUSTOM_CALLS_RE` environment variable.
```
jberchtold-nvidia: Thanks for updating the docstring here to match the changed behavior, since
> By default, `NVTE_JAX_CUSTOM_CALLS_RE` is set to `.*`

is no longer true.

May be out of scope for this PR, but I think we should explore a more intuitive way to let users turn on/off TE primitives. In my opinion, the regex system we have is a bit tougher to use than it could be.

With this PR, if a user wants to disable the `DBiasQuantizePrimitive`, they'd probably try this ``NVTE_JAX_CUSTOM_CALLS_RE='^(?!DBiasQuantizePrimitive$).+$'``. However, if they're benchmarking this, it could give unintuitive results as this will also enable GemmPrimitive for delayed scaling since we're now using the env var instead of the enabled state. This is clear in your docstring, but it's possible the user doesn't read this fully. Or if they did read fully, they'd need to look through all the current defaults and encode them into a regex and then add additional modifications to that regex.

What if we instead had a key/value approach, `NVTE_JAX_CUSTOM_CALLS="DBiasQuantizePrimitive=false"` would only modify the state of `DBiasQuantizePrimitive`, and keep the state of all others unchanged from default.

If a user wanted to update multiple primitives, they could do `NVTE_JAX_CUSTOM_CALLS="DBiasQuantizePrimitive=false,GemmPrimitive=true"`, the rest would use their default `_is_enabled` state.

To still support an easy way to turn all primitives on or off without enumerating each one, we could support `NVTE_JAX_CUSTOM_CALLS="true"` to turn all on, and `NVTE_JAX_CUSTOM_CALLS="false"` to turn all off.

I think the docs are clear here with your updates, thanks for updating them! I'm more thinking for ease of use and avoiding any subtle cases that could happen if the user skims the docs.

Let me know what you think, I could also handle implementing a key/value or another approach in a follow-up PR if we want to keep this PR moving for perf. Thanks!

transformer_engine/jax/cpp_extensions/base.py
```diff
@@ -145,3 +177,48 @@ def name_of_wrapper_p():
...
+            ):
+                cls.set_enabled(False)
+
+    # Apply enables
+    for name in enable_set:
+        cls = _primitive_registry.get(name)
+        if cls and isinstance(cls, type) and issubclass(cls, BasePrimitive):
+            cls.set_enabled(True)
+        else:
+            warnings.warn(f"Primitive not found in registry: {name}")
```
jberchtold-nvidia: Should this raise an error instead of a warning?

transformer_engine/jax/cpp_extensions/base.py
phu0ngng: > However, if they're benchmarking this, it could give unintuitive results as this will also enable GemmPrimitive for delayed scaling since we're now using the env var instead of the enabled state. 

Thanks for spotting the issue. Agree that the regex system may not be the best in multiple cases. 
I added the `NVTE_JAX_CUSTOM_CALLS` implementation. Please let me know if you have any other suggestions.

PR #1963, opened 2025-07-17
[PyTorch] Debug linear layer when saving original input and using debug quantizer

timmoon10:
# Description

https://github.com/NVIDIA/TransformerEngine/pull/1865 introduced a failure in the distributed debug tests. ~The root cause is because we only all-gather the row-wise data for `DebugQuantizedTensor`:~
https://github.com/NVIDIA/TransformerEngine/blob/f8933bbf6cac743d193ff730e6c61331635e38c1/transformer_engine/pytorch/distributed.py#L1394
~However, if the linear layer is caching its original input tensor and requantizing in the backward pass, the correct behavior is to only quantize the column-wise data.~ This PR is a hacky workaround that only applies the debug quantizer to the gathered input tensor.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Modify linear backward to avoid all-gathering debug tensor with only column-wise data

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/linear.py
+++ b/transformer_engine/pytorch/module/linear.py
@@ -595,7 +595,7 @@ class _Linear(torch.autograd.Function):
                             # All-gather is not supported with FP8 column-wise data
                             quantizer.set_usage(rowwise=True, columnwise=False)
                         else:
-                            quantizer.set_usage(rowwise=False, columnwise=True)
+                            quantizer.set_usage(rowwise=True, columnwise=True)
                         inputmat = quantizer(inputmat)
                 else:
                     if isinstance(inputmat, QuantizedTensorBase):


transformer_engine/pytorch/module/linear.py
```diff
@@ -572,20 +579,26 @@ def backward(ctx, grad_output: torch.Tensor) -> Tuple[Union[torch.Tensor, None],
...
-                                columnwise=not ctx.backward_input_needs_gather,
-                            )
+                        if ctx.backward_input_needs_gather and isinstance(
+                            quantizer, (Float8Quantizer, Float8CurrentScalingQuantizer)
+                        ):
+                            # All-gather is not supported with FP8 column-wise data
+                            quantizer.set_usage(rowwise=True, columnwise=False)
                         else:
-                            quantizer.set_usage(rowwise=False, columnwise=True)
+                            quantizer.set_usage(rowwise=True, columnwise=True)
```
hxbai, line 598: Hi Tim, why do we need the rowwise data here?

pggPL:
I think this line is source of error https://github.com/NVIDIA/TransformerEngine/blob/f8933bbf6cac743d193ff730e6c61331635e38c1/transformer_engine/pytorch/distributed.py#L1408
The rowwise/columnwise in DebugQuantizedTensors are the tensors used in gemms -> they can be both the same Float8Tensor object for example. And `update_usage()` does nothing in debug tensors currently.

transformer_engine/pytorch/module/linear.py
timmoon10: In principle we shouldn't need it, but I was running into issues where FP8 casts were failing without it. ~Actually, I don't think we need it once we skip the debug quantizer case.~

timmoon10:
@pggPL I tried changing https://github.com/NVIDIA/TransformerEngine/blob/f8933bbf6cac743d193ff730e6c61331635e38c1/transformer_engine/pytorch/distributed.py#L1408
to
```python
out_obj.columnwise_gemm_tensor = out_obj.rowwise_gemm_tensor 
```
However, the error reappeared when I applied the debug quantizer to the local input tensor.

For now, I think we should merge this as a quick bugfix and we can fix the edge cases for the debug tensor later.

transformer_engine/pytorch/module/linear.py
hxbai: I think this may be due to that the condition is modified. For `(Float8Quantizer, Float8CurrentScalingQuantizer)` cases, it doesn't support only quantizing the colwise data. But if `backward_input_needs_gather` is False, `(Float8Quantizer, Float8CurrentScalingQuantizer)` tensors would also go into the `else` path, which will cause the error. This hurts the performance of blockwise FP8 and MXFP8. Do you need me to create a fix PR for this, or will you fix this together with the DebugTensor?

hxbai: Created #1978

PR #1970, opened 2025-07-18
[C][PyTorch] NVFP4 forward MXFP8 backward recipe

ksivaman:
# Description

Add support for the NVFP4 forward MXFP8 backward recipe for core and PyTorch.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Add support for the NVFP4 forward MXFP8 backward recipe for core and PyTorch.

# Checklist:

### DIFF
--- a/transformer_engine/common/activation/activation_template.h
+++ b/transformer_engine/common/activation/activation_template.h
@@ -25,15 +25,14 @@ namespace transformer_engine {
 template <typename ComputeType, typename Param, ComputeType (*OP)(ComputeType, const Param &)>
 void act_fn(const NVTETensor input, NVTETensor output, cudaStream_t stream) {
   using namespace detail;
-  constexpr bool IS_DBIAS = false;
-  constexpr bool IS_DACT = false;
-  constexpr bool IS_ACT = true;
-  constexpr NVTETensor dbias = nullptr;
-  constexpr NVTETensor workspace = nullptr;
-  constexpr const NVTETensor grad = nullptr;
+  // constexpr bool IS_DBIAS = false;
+  // constexpr bool IS_DACT = false;
+  // constexpr bool IS_ACT = true;
+  // constexpr NVTETensor dbias = nullptr;
+  // constexpr NVTETensor workspace = nullptr;
+  // constexpr const NVTETensor grad = nullptr;
 
-  quantize_helper<IS_DBIAS, IS_DACT, IS_ACT, Empty, OP>(input, grad, output, dbias, workspace,
-                                                        nullptr, stream);
+  // quantize_helper<IS_DBIAS, IS_DACT, IS_ACT, Empty, OP>(input, grad, output, dbias, workspace, nullptr, stream);
 }
 
 template <typename ComputeType, typename Param, ComputeType (*OP)(ComputeType, const Param &)>
--- a/transformer_engine/common/common.h
+++ b/transformer_engine/common/common.h
@@ -50,6 +50,10 @@ inline bool is_delayed_tensor_scaling(const NVTEScalingMode &mode) {
 
 inline bool is_mxfp_scaling(const NVTEScalingMode &mode) { return mode == NVTE_MXFP8_1D_SCALING; }
 
+inline bool is_nvfp_scaling(const NVTEScalingMode &mode) {
+  return mode == NVTE_FWD_NVFP4_BWD_MXFP8_SCALING;
+}
+
 inline size_t product(const std::vector<size_t> &shape, const size_t begin, const size_t end) {
   NVTE_CHECK(begin <= end && end <= shape.size(), "Attempted to access entries ", begin, " to ",
              end, " in a vector with ", shape.size(), " entries");
@@ -668,7 +672,8 @@ constexpr size_t scale_tensor_alignment_X_colwise = 128;
 constexpr size_t scale_tensor_alignment_Y_colwise = 4;
 
 // Alignment requirements for the Tensor Memory Accelerator (TMA)
-constexpr int TMA_gmem_alignment = 16;  // global memory address alignment
+constexpr size_t TMA_GMEM_ALIGNMENT = 16;    // global memory address alignment
+constexpr size_t TMA_SHMEM_ALIGNMENT = 128;  // shared memory address alignment
 
 inline bool is_aligned_ptr(const void *ptr, size_t alignment) {
   return reinterpret_cast<uintptr_t>(ptr) % alignment == 0;
--- a/transformer_engine/common/gemm/cublaslt_gemm.cu
+++ b/transformer_engine/common/gemm/cublaslt_gemm.cu
@@ -35,6 +35,10 @@ cudaDataType_t get_cuda_dtype(const transformer_engine::DType t) {
       return CUDA_R_8F_E4M3;
     case DType::kFloat8E5M2:
       return CUDA_R_8F_E5M2;
+#if CUDA_VERSION >= 12080
+    case DType::kFloat4E2M1:
+      return CUDA_R_4F_E2M1;
+#endif
     default:
       NVTE_ERROR("Invalid type");
   }
@@ -133,6 +137,14 @@ GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cubla
     ret.Atype = is_A_transposed ? A.data.dtype : A.columnwise_data.dtype;
     ret.A_scale_inv = is_A_transposed ? A.scale_inv.dptr : A.columnwise_scale_inv.dptr;
     ret.lda = is_A_transposed ? k : m;
+  } else if (is_nvfp_scaling(A.scaling_mode)) {
+    NVTE_CHECK(is_A_transposed, "Incorrect GEMM layout for NVFP4");
+    NVTE_CHECK(A.has_data(), "Input A is missing row-wise usage");
+    ret.A = A.data.dptr;
+    ret.transA = CUBLAS_OP_T;
+    ret.Atype = A.data.dtype;
+    ret.A_scale_inv = A.scale_inv.dptr;
+    ret.lda = k;
   } else if (A.scaling_mode == NVTE_BLOCK_SCALING_1D || A.scaling_mode == NVTE_BLOCK_SCALING_2D) {
     // FP8 block scaling
     // Note: Hopper only supports TN GEMMs for FP8. "Column-wise data" is transpose of data.
@@ -192,6 +204,14 @@ GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cubla
     ret.Btype = is_B_transposed ? B.columnwise_data.dtype : B.data.dtype;
     ret.B_scale_inv = is_B_transposed ? B.columnwise_scale_inv.dptr : B.scale_inv.dptr;
     ret.ldb = is_B_transposed ? n : k;
+  } else if (is_nvfp_scaling(B.scaling_mode)) {
+    NVTE_CHECK(!is_B_transposed, "Incorrect GEMM layout for NVFP4");
+    NVTE_CHECK(B.has_data(), "Input B is missing row-wise usage");
+    ret.B = B.data.dptr;
+    ret.transB = CUBLAS_OP_N;
+    ret.Btype = B.data.dtype;
+    ret.B_scale_inv = B.scale_inv.dptr;
+    ret.ldb = k;
   } else if (B.scaling_mode == NVTE_BLOCK_SCALING_1D || B.scaling_mode == NVTE_BLOCK_SCALING_2D) {
     // FP8 block scaling
     // Note: Hopper only supports TN GEMMs for FP8. "Column-wise data" is transpose of data.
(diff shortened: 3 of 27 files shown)


transformer_engine/common/common.h
```diff
@@ -50,6 +50,10 @@ inline bool is_delayed_tensor_scaling(const NVTEScalingMode &mode) {
 
 inline bool is_mxfp_scaling(const NVTEScalingMode &mode) { return mode == NVTE_MXFP8_1D_SCALING; }
 
+inline bool is_nvfp_scaling(const NVTEScalingMode &mode) {
+  return mode == NVTE_FWD_NVFP4_BWD_MXFP8_SCALING;
+}
```
timmoon10: I'm not a fan of this name since it's constraining usage, and would prefer a more descriptive name like `NVTE_HYBRID_NVFP4_MXFP8_SCALING`: https://github.com/NVIDIA/TransformerEngine/pull/1779#discussion_r2220771240

transformer_engine/pytorch/csrc/type_converters.cpp
```diff
@@ -116,6 +116,38 @@ TensorWrapper NVTETensorFromFloat8BlockwiseQTensor(py::handle tensor, Quantizer
...
+
+  NVTE_CHECK(rowwise_usage || columnwise_usage, "No data found for Hybrid NVFP4 Tensor.");
+
+  // Row-scaled data
+  if (rowwise_usage) {
+    const auto &data = tensor.attr("_rowwise_data").cast<at::Tensor>();
+    const auto &scale_inv = tensor.attr("_rowwise_scale_inv").cast<at::Tensor>();
+    ret.set_rowwise_data(data.data_ptr(), DType::kFloat4E2M1, getTensorShape(data));
+    ret.set_rowwise_scale_inv(scale_inv.data_ptr(), DType::kFloat8E4M3, getTensorShape(scale_inv));
+  }
```
timmoon10, line 133: Don't we also need to get `_per_tensor_rowwise_scale_inv`?

transformer_engine/pytorch/csrc/quantizer.cpp
timmoon10, line 1: Be advised I've been working on a refactor of the C++ quantizers at https://github.com/NVIDIA/TransformerEngine/pull/1952

This should allow us to move some of the NVFP4-specific cast logic (e.g. for amax reductions or fused amax kernels) into the C++ quantizer class.

transformer_engine/pytorch/tensor/_internal/hybrid_nvfp4_tensor_base.py
```diff
@@ -0,0 +1,219 @@
...
+
+    """
+
+    _rowwise_data: Optional[torch.Tensor]
+    _columnwise_data: Optional[torch.Tensor]
+    _quantizer: Optional[Quantizer]
+    _fp8_dtype: TE_DType
+    _rowwise_scale_inv: torch.Tensor
+    _columnwise_scale_inv: torch.Tensor
+    _per_tensor_rowwise_scale_inv: torch.Tensor
```
timmoon10, line 66: I would suggest making all the scale-inv names explicit to reduce confusion:
```suggestion
    _rowwise_block_scale_inv: torch.Tensor
    _rowwise_tensor_scale_inv: torch.Tensor
    _columnwise_block_scale_inv: torch.Tensor
```

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -119,8 +123,16 @@ GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cubla
         NVTE_CHECK(!is_fp8_dtype(ret.Atype), "Input A is missing column-wise usage");
       }
     }
-  } else if (is_mxfp_scaling(A.scaling_mode)) {
-    // MXFP8
+  } else if (is_hybrid_nvfp4_scaling(A.scaling_mode) && !grad) {
+    NVTE_CHECK(is_A_transposed, "Incorrect GEMM layout for NVFP4");
```
timmoon10: We aren't actually interested in whether it's the backward pass or not, but rather this GEMM has row-wise usage for A. It would be better to just check `is_A_transposed`:
```suggestion
  } else if (is_hybrid_nvfp4_scaling(A.scaling_mode) && is_A_transposed) {
```

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -83,7 +87,7 @@ struct GemmParam {
  */
 GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cublasOperation_t transA,
                                 const transformer_engine::Tensor &B, const cublasOperation_t transB,
-                                int m, int n, int k) {
+                                int m, int n, int k, bool grad) {
```
timmoon10: It's a bad sign when DL-specific logic like forward/backward is leaking into something as generic as a GEMM. It turns out `grad` is not capturing the important behavior either, which is whether row-wise or column-wise data is required.
```suggestion
                                int m, int n, int k) {
```

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -339,12 +367,20 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
...
 
-  // set fp8 attributes -- input and output types should already be set to fp8 as appropriate
-  // Note: gelu fusion isn't available right now, and we don't need
+  // Set conditions for when MXFP8 and NVFP4 gemm execution.
+  bool hybrid_nvfp4_recipe = is_hybrid_nvfp4_scaling(inputA->scaling_mode) &&
+                             is_hybrid_nvfp4_scaling(inputB->scaling_mode);
+  bool mxfp8_gemm =
+      (is_mxfp_scaling(inputA->scaling_mode) && is_mxfp_scaling(inputB->scaling_mode)) ||
+      (grad && hybrid_nvfp4_recipe);
+  bool nvfp4_gemm = !grad && hybrid_nvfp4_recipe;
```
timmoon10: I don't like how we're incorporating `grad` here since it's logically incorrect (we don't actually care about forward/backward, we care about row-wise/column-wise usage) and it constrains us from using NVFP4 in the backward pass. For example, this hybrid NVFP4-MXFP8 format could be used to store the grad output, with NVFP4 dgrad GEMM and MXFP8 wgrad GEMM. Unfortunately, handling this properly would be messy:
```suggestion
  const auto mode_A = inputA->scaling_mode;
  const auto mode_B = inputB->scaling_mode;
  const auto is_A_mxfp8 = is_mxfp_scaling(mode_A) || (is_nvfp4_scaling(mode_A) && transA == CUBLAS_OP_N);
  const auto is_A_nvfp4 = is_nvfp4_scaling(mode_A) && transA != CUBLAS_OP_N;
    const auto is_B_mxfp8 = is_mxfp_scaling(mode_B) || (is_nvfp4_scaling(mode_B) && transB != CUBLAS_OP_N);
  const auto is_B_nvfp4 = is_nvfp4_scaling(mode_B) && transB == CUBLAS_OP_N;
  const auto mxfp8_gemm = is_A_mxfp8 && is_B_mxfp8;
  const auto nvfp4_gemm = is_A_nvfp4 && is_B_nvfp4;
```
So the choice is between simple and coincidentally correct, or complex and actually correct. We should add a comment if we go for the simple approach.

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -388,6 +424,17 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
...
+      fp8e4m3 *A_scale_inverse = reinterpret_cast<fp8e4m3 *>(param.A_scale_inv);
+      fp8e4m3 *B_scale_inverse = reinterpret_cast<fp8e4m3 *>(param.B_scale_inv);
+      NVTE_CHECK_CUBLAS(cublasLtMatmulDescSetAttribute(operationDesc,
+                                                       CUBLASLT_MATMUL_DESC_A_SCALE_POINTER,
+                                                       &A_scale_inverse, sizeof(A_scale_inverse)));
+      NVTE_CHECK_CUBLAS(cublasLtMatmulDescSetAttribute(operationDesc,
+                                                       CUBLASLT_MATMUL_DESC_B_SCALE_POINTER,
+                                                       &B_scale_inverse, sizeof(B_scale_inverse)));
+      scaling_mode_a = CUBLASLT_MATMUL_MATRIX_SCALE_VEC16_UE4M3;
+      scaling_mode_b = CUBLASLT_MATMUL_MATRIX_SCALE_VEC16_UE4M3;
```
timmoon10: Do we also need to configure the per-tensor scales?

transformer_engine/common/activation/activation_template.h
```diff
@@ -25,15 +25,14 @@ namespace transformer_engine {
...
-  constexpr const NVTETensor grad = nullptr;
+  // constexpr bool IS_DBIAS = false;
+  // constexpr bool IS_DACT = false;
+  // constexpr bool IS_ACT = true;
+  // constexpr NVTETensor dbias = nullptr;
+  // constexpr NVTETensor workspace = nullptr;
+  // constexpr const NVTETensor grad = nullptr;
 
-  quantize_helper<IS_DBIAS, IS_DACT, IS_ACT, Empty, OP>(input, grad, output, dbias, workspace,
-                                                        nullptr, stream);
```
timmoon10: Debugging code:
```suggestion
  constexpr bool IS_DBIAS = false;
  constexpr bool IS_DACT = false;
  constexpr bool IS_ACT = true;
  constexpr NVTETensor dbias = nullptr;
  constexpr NVTETensor workspace = nullptr;
  constexpr const NVTETensor grad = nullptr;

  quantize_helper<IS_DBIAS, IS_DACT, IS_ACT, Empty, OP>(input, grad, output, dbias, workspace,
                                                        nullptr, stream);
```
We shouldn't actually handle activations within the core lib, but should split it up into multiple kernels like we do for FP8 current scaling. That way we can avoid memory allocations in the core lib and perform amax reductions with PyTorch's NCCL communicator.

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -178,7 +190,15 @@ GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cubla
         NVTE_CHECK(!is_fp8_dtype(ret.Btype), "Input B is missing column-wise usage");
       }
     }
-  } else if (is_mxfp_scaling(B.scaling_mode)) {
+  } else if (is_hybrid_nvfp4_scaling(B.scaling_mode) && !grad) {
+    NVTE_CHECK(!is_B_transposed, "Incorrect GEMM layout for NVFP4");
```
timmoon10:
```suggestion
  } else if (is_hybrid_nvfp4_scaling(B.scaling_mode) && !is_B_transposed) {
```

transformer_engine/common/gemm/cublaslt_gemm.cu
zhongbozhu: I think we need to do it, the fix should be done in another NVFP4 stand-alone PR anyway.

transformer_engine/pytorch/csrc/common.h
```diff
@@ -219,6 +219,21 @@ class MXFP8Quantizer : public Quantizer {
...
+ public:
+  DType dtype;
+
+  explicit HybridNVFP4Quantizer(const py::handle& quantizer);
+
+  NVTEScalingMode get_scaling_mode() const override { return NVTE_FWD_NVFP4_BWD_MXFP8_SCALING; }
+
+  void set_quantization_params(TensorWrapper* tensor) const override;
+
+  std::pair<TensorWrapper, py::object> create_tensor(
```
zhongbozhu: please also have a stand-alone get_scale_shape function

transformer_engine/pytorch/csrc/type_converters.cpp
zhongbozhu: yes, I think this part is missing

transformer_engine/common/recipe/__init__.py
```diff
@@ -351,3 +358,41 @@ def __repr__(self) -> str:
             f"fp8_dpa={self.fp8_dpa}, "
             f"fp8_mha={self.fp8_mha}"
         )
+
+
+@dataclass()
+class HybridNVFP4BlockScaling(Recipe):
```
negvet, line 364: What do you think of removing `Hybrid` from recipe name (`HybridNVFP4BlockScaling`) or/and making it more specific?
It may be the case that we will have more hybrid recipes in the future, possibly even NVFP4 plus something else (not MXFP8).

timmoon10: How about something like `HybridNVFP4MXFP8Recipe`, similar to the name used in the core lib:
https://github.com/NVIDIA/TransformerEngine/blob/73fae08c74b10ee1dd0db186ea402d37ac53c652/transformer_engine/common/include/transformer_engine/transformer_engine.h#L94
It describes the data layout, doesn't have the misleading `BlockScaling` (NVFP4 is both block-scaling and tensor-scaling), and doesn't limit usages like `ForwardNVFP4BackwardMXFP8`. However, I don't like how "NVFP4" and "MXFP8" are mashed together in a way that's hard to parse.

negvet: > I don't like how "NVFP4" and "MXFP8" are mashed together in a way that's hard to parse.

Me too, not sure there is a clean way to do it without violating PEP 8. It is hard to come up with another workaround except hardcoding `ForwardNVFP4BackwardMXFP8`.

PR #1974, opened 2025-07-21
Fix the device for cuDNN/cuBLAS handles

cyanguwa:
# Description

[GetHandle](https://github.com/NVIDIA/TransformerEngine/blob/ca7407e3e78bb692ffbec552d1d71e2fd8901259/transformer_engine/common/util/handle_manager.h#L25) returns a handle for cuDNN SDPA, cuDNN Norm, and cuBLAS GEMM ops.  It relies on `cuda::current_device()` to get the device to create the handle on, but this may be a different device than that of the data, which may then lead to execution errors for cuDNN and cuBLAS.

This PRs
- adds a `torch.cuda.device()` context to all the relevant modules in PyTorch: `Linear`, `Grouped_Linear`, `LayerNorm_Linear`, `LayerNorm_MLP`, and `DotProductAttention`
- does not touch TE-Jax since XLA ensures that the computation happens where the data is
- does not touch the C APIs `nvte_xxx`, which require the current stream as an input. Users should make sure to switch to the right current device before generating the stream and calling these C APIs.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Fix current device for cuDNN/cuBLAS handles.

# Checklist:


transformer_engine/pytorch/module/linear.py
```diff
@@ -1340,7 +1340,7 @@ def forward(
             if get_ub(self.ub_name + "_dgrad").is_fp8_ubuf():
                 fp8_grad = True
 
-        with self.prepare_forward(
+        with torch.cuda.device(inp.device), self.prepare_forward(
```
timmoon10: If we go this route, it seems better to use the weight's device. Presumably that's where all the FP8 state and other GPU data has been initialized (and nothing will work if there's a mismatch there).
```suggestion
        with torch.cuda.device(getattr(self, self.weight_names[0])), self.prepare_forward(
```

cyanguwa: I changed to the weight device for linear-based layers. DPA still uses `query_layer.device` because it doesn't have any weights.

cyanguwa:
CI failures are not related to this PR and they will be fixed in #1991.

PR #1980, opened 2025-07-22
[JAX] Fixing GemmPrimitive partitioning rules to handle tensor-parallelism correctly for sequence-parallel inputs

denera:
# Description

`GemmPrimitive` was originally developed to inherit the partitioning rules of the native `jax.nn.scaled_matmul` implementation. However, this causes partitioning rules to effectively kill tensor-parallelism in the MLP block when FFN1 receives a sequence-parallel input.

This PR changes the partitioning rules to avoid all-gathering tensor-parallel RHS operands under any conditions, and instead rely on an extra boolean flag input to convert the `jax.lax.psum` on the GEMM output into a `jax.lax.psum_scatter`.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/gemm.py
+++ b/transformer_engine/jax/cpp_extensions/gemm.py
@@ -34,7 +34,6 @@ from ..quantize import (
     is_fp8_gemm_with_all_layouts_supported,
     apply_padding_to_scale_inv,
 )
-from ..sharding import global_mesh_resource
 from .misc import get_padded_spec
 
 
@@ -581,9 +580,9 @@ class GemmPrimitive(BasePrimitive):
         )
 
         # Extract single leading and contracting dimension specs
-        (lhs_lspec, rhs_lspec, lhs_cspec, rhs_cspec) = map(
+        (lhs_cspec, rhs_cspec) = map(
             lambda specs: None if len(specs) == 0 else specs[0],
-            (lhs_lspec_not_none, rhs_lspec_not_none, lhs_cspec_not_none, rhs_cspec_not_none),
+            (lhs_cspec_not_none, rhs_cspec_not_none),
         )
 
         # Partitioning rules:
--- a/transformer_engine/jax/cpp_extensions/gemm.py
+++ b/transformer_engine/jax/cpp_extensions/gemm.py
@@ -34,6 +34,7 @@ from ..quantize import (
     is_fp8_gemm_with_all_layouts_supported,
     apply_padding_to_scale_inv,
 )
+from ..sharding import global_mesh_resource
 from .misc import get_padded_spec
 
 
@@ -155,7 +156,7 @@ class GemmPrimitive(BasePrimitive):
 
     name = "te_gemm_ffi"
     multiple_results = True
-    impl_static_args = (6, 7, 8, 9, 10, 11, 12, 13, 14, 15)
+    impl_static_args = (6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17)
     inner_primitive = None
     outer_primitive = None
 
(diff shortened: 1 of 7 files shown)


transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -802,12 +808,14 @@ def shardy_sharding_rule(
...
         use_split_accumulator,
+        sequence_parallel_output,
+        sequence_dim,
         mesh,
         operand_types,
         result_types,
     ):
         del lhs_quantized_colwise, rhs_quantized_colwise, out_dtype, grad, use_split_accumulator
-        del mesh, result_types
+        del sequence_parallel_output, sequence_dim, mesh, result_types
```
jberchtold-nvidia, line 818: For this PR, the rules for the GPSMD have been updated in `infer_sharding_from_operands` via `_parse_operand_output_specs`. For Shardy, the rules are unchanged. I wanted to clarify my understanding here, is that because:

a) This more optimal sharding logical isn't compatible with Shardy due to a Shardy limitation?
b) Shardy rules don't need to be updated because Shardy still calls `partition` in the second pass, which does call `_parse_operand_output_specs`?

Thanks!

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1328,6 +1340,14 @@ def gemm(
...
+        Produces an output with the first non-batched non-contracting dimension sharded with the
+        same spec as operand contracting dimensions. This effectively converts the `jax.lax.psum`
+        for the GEMM output into a `jax.lax.psum_scatter`.
+    sequence_dim: int, default = None
+        Index of the sequence dimension for the LHS operand. This controls which dimension of the
+        GEMM output is scattered when `sequence_parallel_output=True`. When `None`, the first
+        non-batched non-contracting dimension is assumed to be the sequence dimension.
 
     Returns
     -------
```
jberchtold-nvidia, line 1355: Can you add an assert below to assert that these new args can't be specified when using the native JAX gemm?
Similar to the existing asserts:
```
assert kwargs.get("gelu_input", None) is None and not fuse_bias, (
            "TE GEMM was invoked with GeLU fusion options that are not supported by the "
            "`jax.lax.dot_general` and `jnp.scaled_matmul` backends used when the custom cuBLAS "
            "GEMM primitive is disabled."
        )
```

transformer_engine/jax/cpp_extensions/gemm.py
denera: The reason is the latter, what you identified in (b).

There is no way for us to detect if we have to execute an all-reduce on the GEMM output (and modify that to a reduce-scatter based on the option flag added in this PR) inside `shardy_sharding_rule()` because JAX does not give us access to the actual `jax.sharding.PartitionSpec` of the operands. Consequently, we are unable to define a rule that aligns with the final sharding of the operands and outputs after the `partition()` call.

This was true before this PR as well. The original `GemmPrimitive` also had Shardy rules that were not entirely aligned with `infer_sharding_from_operands()` for the same reason. This is going to be the case for any GEMM-like custom op that has to conditionally perform a collective in the sharded implementation unless JAX/XLA change what information is available to custom ops when parsing the Shardy rules. At least for the time being though, there doesn't seem to be any adverse consequences from the mismatch.

denera:
NOTE: L1 distributed failures on B200 are due to a runtime limit. Restarted the CI on these tests manually.

transformer_engine/jax/cpp_extensions/gemm.py
jberchtold-nvidia: Got it, thanks for the explanation!

PR #1982, opened 2025-07-22
Use userbuffers for MXFP8 wgrad all-gather overlap

djns99:
# Description

Updates the wgrad comm overlap logic to use userbuffers when doing the all gather gemm/comm overlapping

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Adds new UB manager object for proj/fc2 wgrad 
- Updates code to use new object for copies

# Checklist:


transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -591,6 +597,29 @@ void CommOverlapBase::split_overlap_rs(const TensorWrapper &A, bool transa, cons
...
+void CommOverlapBase::bulk_overlap_external_ag(const TensorWrapper &input, cudaStream_t send_stream,
+                                               cudaStream_t recv_stream, cudaStream_t stream_main) {
+  int comm_bytes = _ubuf.bytes();
+  int comm_bytes_per_rank = comm_bytes / _tp_size;
+
+  // TODO: Check if pull mode syncs before pulling the data
+  NVTE_CHECK(_ub_comm->push == 1,
+             "Userbuffer is populated on the send stream, so push mode must be enabled for proper "
+             "sequencing");
+  NVTE_CHECK(_tp_size == _ub_comm->nvsize,
```
djns99: @timmoon10  Is there a good way to test these two cases to determine if I can remove these checks?

timmoon10: I'm not sure, I would expect these assumptions are deeply embedded. For `_tp_size`, you could try running `run_layer_with_overlap.py`, but initialize the UB comm with `numlocal` that doesn't match the TP size:
https://github.com/NVIDIA/TransformerEngine/blob/7d7ce6b8ff78d7a954fb320ab8e3ddc63ed9e2d7/tests/pytorch/distributed/run_layer_with_overlap.py#L384-L391

djns99: Ok I updated it to use tp_size and tp_id from the communicator instead of relying on the comm->nvsize parameter

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -761,21 +761,30 @@ def backward(
...
-                    dgrad_comm_stream = ub_obj_dgrad.get_communication_stream()
-                    with torch.cuda.stream(dgrad_comm_stream):
-                        # Syncs with the current stream (dgrad_comm_stream) before starting the all-gather
-                        # This ensures that we don't start until all communication for the dgrad GEMM is complete
-                        grad_output, mxfp8_grad_output_work = gather_along_first_dim(
+                    # We use the send stream to copy into the userbuffers.
+                    # This is the same stream that we will use to access the data in the AG,
+                    # so we dont need to add any syncs yet.
+                    with torch.cuda.stream(dgrad_send_stream):
+                        grad_output, _ = fill_userbuffers_buffer_for_all_gather(
```
djns99, line 777: I would like to do the quantization here fused with the rowwise quantization above. I couldn't figure out a nice (quick) way of doing this though so I might come back to it.  
Simply quantizing `grad_outputs[0]` at the top of the function before using it in dgrad, didn't quite work because it hit a bunch of asserts. This is hopefully something we can come back to.  
As it is the quantization and NCCL all gather are also overlapped with the GEMM by virtue of running on the `dgrad_send_stream` here.

denera: This should be easily possible.

Toward the top of the backward pass, we set  `grad_output_quantizer.set_usage(rowwise=True, columnwise=True)` before `TransformerEngineBaseModule.grad_output_preprocess()`, but then flip `columnwise=False` if we see that `ub_overlap_ag` is enabled.

This conditional `ub_overlap_ag` change to the quantizer usage needs to be shifted to *after* `grad_output_preprocess()` in order to ensure that the rowwise and columnwise quantization happen at the same time in the beginning of the backward pass.

I don't believe any further changes are necessary because the `_fill_userbuffers_buffer_for_all_gather()` already avoids re-quantization if it sees that the tensor has been quantized ahead of time. It will simply copy the columnwise-quantized gradient into the communication buffer based on the quantizer usage at the time.

djns99: Unfortunately the code in the preprocess doesn't support quantized tensors (the `reshape()` operation at the top doesnt exist) and `fill_userbuffers_buffer_for_all_gather` (called in the preprocess) doesn't support both being set.

These should be fixable but I had some weird errors so I left it as it was for now

denera: What goes into preprocess is always plain Torch tensors, but what comes out is a QuantizedTensor. That's as expected. So you do not want to explicitly invoke quantization here. That's what preprocess is supposed to do for you based on the usage information you set into the quantizer object.

And on that note, if the `grad_output_quantizer` usage is set to `rowwise=True` and `columnwise=True` before the preprocess, and everything else is passed in as usual, the preprocess should produce a single QuantizedTensor gradient that has both rowwise and columnwise quantizations in it.

At that point, `grad_output_quantizer` usage can be updated again to rowwise-only for DGRAD AG->GEMM and then column wise-only right before invoking `_fill_userbuffers_for_all_gather()` to account for the fact that only one usage can be enabled at a time in either case. This usage information will allow both the GEMM and the UB fill to pluck out the correct quantized data and scale from the same QuantizedTensor.

I will test this later tonight if I get the chance but I'd be surprised if this doesn't work out of the box. The code, in my reading, appears to account for this kind of use case already.

timmoon10: Fusing the casts is messy since it involves outputting into multiple UB buffers. Perhaps we could modify `fill_userbuffers_buffer_for_all_gather` to have separate UB comms for the row-wise and column-wise data. Alternatively, we could separate the logic for constructing the quantized tensor and doing the quantize:
```python
grad_output = make_tensor_with_userbuffers_buffers(grad_outputs[0].size(), rowwise_comm=ub_obj_dgrad, columnwise_comm=ub_obj_wgrad)
grad_output.copy_(grad_outputs[0])
```

timmoon10 (dismissed):
LGTM, barring some linter warnings.

timmoon10 (approved):
LGTM. The L40 failure is spurious and the B200 failure also shows up in the nightly build.

PR #1984, opened 2025-07-23
Fixed double buffering issue for assymetric layers

sanandaraj5597:
Had to deal with assymetric layers when double buffering.


ksivaman (approved):
LGTM

PR #1985, opened 2025-07-23
[PyTorch] Reset recipe state in fusible operations when FP8 amax history length changes

timmoon10:
# Description

https://github.com/NVIDIA/TransformerEngine/pull/1951 changed the behavior of the fusible ops to only change the quantizer recipe state (mostly in `BasicOperation._fp8_metas`) when the recipe class changed. It turns out this was causing problems when the recipe changes but keeps the same type, e.g. if FP8 delayed scaling changes the amax history length. This PR is a quick fix that fixes the case where the FP8 amax history length changes, although it's not robust to other changes (e.g. with the FP8 amax reduction group).

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Reset recipe state in fusible operations when FP8 amax history length changes

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/ops/fuser.py
+++ b/transformer_engine/pytorch/ops/fuser.py
@@ -401,7 +401,7 @@ class OperationFuser:
         need_reset = False
         recipe_type = type(recipe)
         fusion_params = (recipe_type, first_op_requiring_backward)
-        if fusion_params == (self.recipe_type, self.first_op_requiring_backward):
+        if fusion_params != (self.recipe_type, self.first_op_requiring_backward):
             # Recipe type or grad requirmenets have changed
             need_reset = True
         elif (
@@ -415,7 +415,7 @@ class OperationFuser:
 
         # Reset recipe state
         for op in self._basic_ops:
-            op.reset_recipe_type(recipe=recipe)
+            op.reset_recipe_state(recipe=recipe)
 
         # Check if this is the first iteration
         if self.recipe_type is None:
--- a/transformer_engine/pytorch/ops/fuser.py
+++ b/transformer_engine/pytorch/ops/fuser.py
@@ -398,27 +398,29 @@ class OperationFuser:
                     break
 
         # Early exit if fusion parameters haven't changed
+        need_reset = False
         recipe_type = type(recipe)
         fusion_params = (recipe_type, first_op_requiring_backward)
         if fusion_params == (self.recipe_type, self.first_op_requiring_backward):
+            # Recipe type or grad requirmenets have changed
+            need_reset = True
+        elif (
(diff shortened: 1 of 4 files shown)


transformer_engine/pytorch/ops/fuser.py
```diff
@@ -398,27 +398,30 @@ def maybe_fuse_ops(
...
             for op in self._basic_ops:
-                op.reset_recipe_type(recipe=recipe)
-        # Check if amax history was invalidated
-        elif isinstance(recipe, DelayedScaling):
-            if recipe.amax_history_len != self._last_amax_history_len:
-                raise RuntimeError(
-                    "Detected change of amax history length. "
-                    "Changing the length of amax history is currently not supported."
-                )
+                op.pre_first_fuser_forward()
```
janekb04, line 424: I think these two loops should be reversed (`pre_first_fuser_forward` before `reset_recipe_state`). This is required for initializing weights on the `"meta"` device if `fp8_model_init` is used, but `fp8_autocast` is not. Currently, this should trigger an error in [`basic_linear.py:294`](https://github.com/NVIDIA/TransformerEngine/blob/d1967d5504eb96b3be04448ba6d3f0e31f7935a6/transformer_engine/pytorch/ops/basic/basic_linear.py#L294). This happens because if `fp8_model_init` is used, `self._with_quantized_weight=True`, so `self.get_quantizer` must not return `None`. However, if `fp8_autocast` is not used and `reset_recipe_state` is called before the weight initialization, the recipe will be reset to `None` and `get_quantizer` will return `None`.

timmoon10: I think initializing weights with `device="meta"` should be considered an advanced usage, primarily meant for FSDP. If the user initializes quantized weights but does not enable quantization during the first forward pass, I'm fine if that causes an error.

ksivaman (dismissed):
Unblocking

transformer_engine/pytorch/ops/fuser.py
janekb04: Ok, in that case I think the error message in `basic_linear.py:294` could be updated to reflect that. I think something like this could work:
```
                raise RuntimeError(
                    "Tried to quantize weight with deferred initialization "
                    "due to meta device, but no quantizer was available. "
                    "This is most likely because fp8_model_init was called "
                    "with enabled=True, but fp8_autocast was called with "
                    "enabled=False or recipe=None or was not called at all."
                )
```

janekb04: Perhaps the fact that the recipe given to `fp8_model_init` is ignored should also be mentioned somewhere.

PR #1989, opened 2025-07-23
Fix runtime lib loading for cuDNN

ksivaman:
# Description

Make the search paths for cuDNN shared object while more exhaustive when loading libs during runtime.

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Use `ldconfig` to find lib path before defaulting to brute force loading.

# Checklist:

### DIFF
--- a/transformer_engine/common/__init__.py
+++ b/transformer_engine/common/__init__.py
@@ -246,6 +246,18 @@ def _load_cudnn():
     if found:
         return handle
 
+    # Attempt to locate libcudnn via ldconfig
+    libs = subprocess.check_output(
+        f"ldconfig -p | grep 'libcudnn{_get_sys_extension()}'", shell=True
+    )
+    libs = libs.decode("utf-8").split("\n")
+    sos = []
+    for lib in libs:
+        if "libcudnn" in lib and "=>" in lib:
+            sos.append(lib.split(">")[1].strip())
+    if sos:
+        return ctypes.CDLL(sos[0], mode=ctypes.RTLD_GLOBAL)
+
     # If all else fails, assume that it is in LD_LIBRARY_PATH and error out otherwise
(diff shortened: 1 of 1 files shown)


transformer_engine/common/__init__.py
```diff
@@ -246,6 +246,18 @@ def _load_cudnn():
     if found:
         return handle
 
+    # Attempt to locate libcudnn via ldconfig
+    libs = subprocess.check_output(
+        f"ldconfig -p | grep 'libcudnn{_get_sys_extension()}'", shell=True
+    )
```
timmoon10, line 252: `shell=True` is a code smell since it can introduce shell injection vulnerabilities: https://docs.python.org/3/library/subprocess.html#security-considerations

I don't think there's a way to attack that doesn't involve already controlling the Python session, but I don't have a good sense for security.

transformer_engine/common/__init__.py
```diff
@@ -246,6 +246,18 @@ def _load_cudnn():
...
+    libs = libs.decode("utf-8").split("\n")
+    sos = []
+    for lib in libs:
+        if "libcudnn" in lib and "=>" in lib:
+            sos.append(lib.split(">")[1].strip())
+    if sos:
+        return ctypes.CDLL(sos[0], mode=ctypes.RTLD_GLOBAL)
```
timmoon10, line 259: I think the `grep` is redundant:
```suggestion
    # Attempt to locate libcudnn via ldconfig
    libs = subprocess.check_output("ldconfig -p")
    libs = libs.decode("utf-8").split("\n")
    sos = []
    for lib in libs:
        if f"libcudnn{_get_sys_extension()}" in lib and "=>" in lib:
            sos.append(lib.split(">")[1].strip())
    if sos:
        return ctypes.CDLL(sos[0], mode=ctypes.RTLD_GLOBAL)
```

PR #1997, opened 2025-07-25
Remove if-else and torch.tensor to meet cudagraph requirement

katec846:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Remove the if-else statement for enabling cuda graph

# Checklist:


ksivaman:
@sanandaraj5597 Could you take a look?

sanandaraj5597:
This kinda removes a tiny optimization. Which is not compatible with CG?
1) If-else or 2) torch.Tensor?

jiemingz:
`if torch.equal(grad_output, torch.tensor(1.0, device=grad_output.device))`

@sanandaraj5597 the `if` causes a device host sync which is not cudagraphable, because the data is on gpu but the 'if' is executed by host

@katec846 can we guard this check with a flag?

sanandaraj5597:
I see, we never hit a True on this `if` condition in MCore cases. But let's add a `is_cg_capturable` flag from top to CE loss function and guard this `if` statement under that?

We can default it to False, maybe can pass it to True for the MCore case when we use full iteration CG.

katec846:
@sanandaraj5597 can you help review? I don't have permission to add reviewers

sanandaraj5597:
@katec846 Have you turned on CG and matched training loss curves?

katec846:
I've compared the loss with CG on and off with llama8b finetuning, the loss are almost the same for first 20 iters. Do I need to run more iters?

sanandaraj5597:
Thanks for comparing. LGTM! @ksivaman we can merge.

ksivaman (approved):

PR #1998, opened 2025-07-26
[Core] Kernel that swaps first two tensor dimensions

timmoon10:
# Description

When all-gathering transposed data (e.g. FP8 input tensor during linear layer backward on Hopper), it is not possible to perform a basic NCCL all-gather because the data will be interleaved. Our current approach is to make sure the local data is non-transposed, perform an all-gather, and then transpose. This adds a significant amount of complexity since it entangles FP8 quantization and communication. An alternative approach is to perform an all-gather on transposed data and then fix the interleaving by reshaping the tensor and swapping the first two dimensions. This PR implements this kernel, but does not yet integrate into any of the modules.

Running on an H100 NVL, I compare square FP8 transposes with the corresponding swap-first-dims operation that would be needed for TP=2 and TP=8. 

image

At the largest size (32768):

| Kernel | Throughput | Fraction of peak |
|---|---|---|
| Transpose | 3.45 TB/s | 0.88 |
| Swap first dims (TP=2) | 3.45 TB/s | 0.89 |
| Swap first dims (TP=8) | 3.42 TB/s  | 0.88 |

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add a kernel that swaps first two tensor dimensions

# Checklist:


zhongbozhu:
In the case of FP4, the swap dimensions also need to consider that two elements share a single byte. 

I haven't looked into the kernel in great detail, but looks like we haven't considered that.

zhongbozhu:
For NVFP4 AG, suppose TP2, quantize transpose of block shape 16x1, local copy is 512 x 2048. 

Then quantize_transpose gives 2048x512 output (each element is half byte), 2048 x 32 scaling factors (1 byte each).

AG the local data gives 2x2048x512, 2x2048x32 while the targets are supposed to be 2048x1024 data and 2048x64 scales. 

By swapping the dims indeed makes 2x2048x512 become 2048x1024, but that's for elementwise operation, when an element is half byte, it means that two elements living adjacently in one byte needs to be broken apart. So this swap kernel can only be used in scaling factors, but cannot be used for FP4 data.

timmoon10:
> In the case of FP4, the swap dimensions also need to consider that two elements share a single byte.
> 
> I haven't looked into the kernel in great detail, but looks like we haven't considered that.

This is a problem for 2D transposes, but I don't think we need to worry about it for swapping the first two dims. In your TP=2 example:
```
512 x 2048
  ->(quantize_transpose)-> 2048 x 512
  ->(all-gather)-> 4096 x 512
  ->(reshape)-> 2 x 2048 x 512
  ->(swap first dims)-> 2048 x 2 x 512
  ->(reshape)-> 2048 x 1024
```
When we're swapping dims, the last dim remains contiguous and we can just do 256 byte copies.

timmoon10:
Running on a B100, I find that the kernel attains 95% of cudaMemcpyAsync perf and 75% of memory bandwidth (TP=8, seqlen=16384, hidden_size=4096).

image

image

transformer_engine/common/transpose/swap_first_dims.cu
```diff
@@ -0,0 +1,209 @@
...
+  const auto input_shape = input.shape();
+  const auto output_shape = output.shape();
+  NVTE_CHECK(input_shape.size() >= 2, "Invalid input tensor dimensions (shape=", input_shape, ").");
+  NVTE_CHECK(output_shape.size() == input_shape.size(), "Input tensor (shape=", input_shape,
+             ") and output tensor (shape=", output_shape, ") do not match.");
+  NVTE_CHECK(input_shape[0] == output_shape[1], "Input tensor (shape=", input_shape,
+             ") and output tensor (shape=", output_shape, ") do not match.");
+  NVTE_CHECK(input_shape[1] == output_shape[0], "Input tensor (shape=", input_shape,
+             ") and output tensor (shape=", output_shape, ") do not match.");
+  for (size_t i = 2; i < input_shape.size(); ++i) {
```
Oleg-Goncharov, line 143: nit: adding also the index of the mismatch to the error in the NVTE_CHECK would be slightly more informative

PR #2001, opened 2025-07-28
[Common] PDL for Quantization Kernels

yaox12:
# Description

Since the quantization for different input tensors are independent, we can enable PDL for the kernels to overlap them to some extent. Compared to multi-stream launch, PDL gives us a fine-grained control to reduce the interference as much as possible. For example, in the MXFP8 quantization kernel, we let the former kernel's last stage computation overlap with the latter kernel's TMA load.

Here're the Nsys profiles for 8 consecutive MXFP8 quantization kernels (input tensors in shape [4096, 4096]) with 1) launch sequentially, 2) multi-stream launch, 3) PDL launch. Multi-stream launch doesn't show perf gain while PDL launch gives about 1.22x speedup.

1. launch sequentially
image
2. multi-stream launch
image
3. PDL launch
image

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/common/util/cast_kernels.cuh
+++ b/transformer_engine/common/util/cast_kernels.cuh
@@ -203,6 +203,11 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK)
     // Wait for the data to have arrived
     ptx::mbarrier_wait_parity(&mbar[stage], parity);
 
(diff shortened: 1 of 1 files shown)


phu0ngng (approved):
LGTM!

PR #2003, opened 2025-07-28
[JAX] TE GEMM checkpointing policies

jberchtold-nvidia:
# Description

Add JAX checkpointing policies that support the TE GEMM primitives that mirror the existing JAX checkpointing policies that checkpoint `jax.lax.dot_general`'s primitive.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add new JAX checkpoint policies in a new file named `checkpoint_policies.py`. These policies allow users to maintain existing checkpoint policy functionality for GEMMs when switching to use TE GEMMs by using these new policies that support `jax.lax.dot_general` and TE GEMMs. Previously, TE GEMM checkpoints would be ignored and would cause unnecessary rematted computations unless checkpoint names were listed explicitly.
- Expose checkpoint names from fused attention and layernorm MLP so users can change them if they'd like

# Checklist:


transformer_engine/jax/checkpoint_policies.py
```diff
@@ -0,0 +1,31 @@
...
+]
+
+
+def te_gemms_saveable(prim, *_, **__) -> bool:
+    """Checkpoint policy for Transformer Engine GEMMs."""
+    return prim in {GemmPrimitive.outer_primitive, GroupedGemmPrimitive.outer_primitive}
+
+
+# The TE GEMM primitive does not support batched GEMMs, so the batched and non-batched policies are the same.
+te_gemms_with_no_batch_dims_saveable = te_gemms_saveable
```
phu0ngng: Could we raise a warning or an error when batched policies are used instead?

jberchtold-nvidia: Good point, this alias is a bit confusing since the TE GEMM can't batch. I think it's better if I just remove this alias entirely. I want to extend JAX's dot policies to be the same behavior but include TE GEMMs in a single policy, so I'd like `dots_and_te_gemms_with_no_batch_dims` to function like `jax.checkpoint_policies.checkpoint_dots_with_no_batch_dims` + TE GEMMs. But since the TE GEMM doesn't support batching, I can just remove this alias and have

```
dots_and_te_gemms_with_no_batch_dims = jax.checkpoint_policies.save_from_both_policies(
    jax.checkpoint_policies.checkpoint_dots_with_no_batch_dims,
    te_gemms_saveable, # TE GEMMs don't support batching so include all TE GEMMs
)
```

and I should also expose the other version without the batching constraint, like
```
dots_and_te_gemms = jax.checkpoint_policies.save_from_both_policies(
    jax.checkpoint_policies.checkpoint_dots
    te_gemms_saveable, 
)
```

jberchtold-nvidia: Made these changes in latest version

PR #2004, opened 2025-07-29
Fuse amax computation into activation kernel

janekb04:
# Description

This PR fuses amax computation into activation (both forward and backward) kernels for non-quantized outputs when using FP8 current scaling removing a call to `nvte_compute_amax`.

image

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- (646062940a1e768f7738e36e4acc4f814b5abe70) Compute amax in activation kernels when the output pointer is provided, even for non-fp8 outputs
- (234774a38ef04f6b06bff0285ac20dcf492eaa5a) Initialize metatensor values for the `Float8CurrentScalingQuantizer` to avoid using atomic max on uninitialized amax
- (8a686d6c88a6fa2532758f53d30ffcb2e3ec83d1) Fuse computation of amax into the activation kernel for fp8 current scaling. Adds two new member functions to `Float8CurrentScalingQuantizer`.
```cpp
class Float8CurrentScalingQuantizer : public Quantizer {
 public:
  ...

  /*! @brief Convert to a quantized data format avoiding amax computation */
  void quantize_with_amax(TensorWrapper& input, TensorWrapper& out,
                          const std::optional<TensorWrapper>& noop_flag = std::nullopt);

 private:
  void quantize_impl(const TensorWrapper& input, TensorWrapper& out,
                     const std::optional<TensorWrapper>& noop_flag, bool compute_amax);
}
```

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/csrc/extensions/activation.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/activation.cpp
@@ -32,8 +32,17 @@ py::object activation_helper(const at::Tensor& input, py::handle quantizer, int 
     // Compute activation directly
     NVTE_SCOPED_GIL_RELEASE(
         { act_func(input_cpp.data(), out_cpp.data(), at::cuda::getCurrentCUDAStream()); });
+  } else if (detail::IsFloat8CurrentScalingQuantizers(quantizer.ptr())) {
+    // Compute activation in high-precision fused together with amax, then quantize.
+
+    auto quantizer_cpp_cs = dynamic_cast<Float8CurrentScalingQuantizer *>(quantizer_cpp.get());
+    auto [temp_cpp, _] = quantizer_cpp_cs->create_hp_tensor_with_amax(output_shape, fake_dtype);
+    NVTE_SCOPED_GIL_RELEASE(
+        { act_func(input_cpp.data(), temp_cpp.data(), at::cuda::getCurrentCUDAStream()); });
+    quantizer_cpp_cs->quantize_with_amax(temp_cpp, out_cpp);
   } else {
     // Compute activation in high-precision, then quantize
+    
     auto [temp_cpp, _] = NoneQuantizer(py::none()).create_tensor(output_shape, fake_dtype);
     NVTE_SCOPED_GIL_RELEASE(
         { act_func(input_cpp.data(), temp_cpp.data(), at::cuda::getCurrentCUDAStream()); });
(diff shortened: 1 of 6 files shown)


transformer_engine/pytorch/csrc/extensions/activation.cpp
```diff
@@ -32,8 +32,17 @@ py::object activation_helper(const at::Tensor& input, py::handle quantizer, int
     // Compute activation directly
     NVTE_SCOPED_GIL_RELEASE(
         { act_func(input_cpp.data(), out_cpp.data(), at::cuda::getCurrentCUDAStream()); });
+  } else if (detail::IsFloat8CurrentScalingQuantizers(quantizer.ptr())) {
+    // Compute activation in high-precision fused together with amax, then quantize.
+
+    auto quantizer_cpp_cs = dynamic_cast<Float8CurrentScalingQuantizer*>(quantizer_cpp.get());
+    auto [temp_cpp, _] = quantizer_cpp_cs->create_hp_tensor_with_amax(output_shape, fake_dtype);
+    NVTE_SCOPED_GIL_RELEASE(
+        { act_func(input_cpp.data(), temp_cpp.data(), at::cuda::getCurrentCUDAStream()); });
```
timmoon10, line 41: We need to zero out the amax if we reuse the quantizer in multiple steps:
```suggestion
    quantizer_cpp_cs->amax.zeros_();  // or fill_(0)?
    NVTE_SCOPED_GIL_RELEASE(
        { act_func(input_cpp.data(), temp_cpp.data(), at::cuda::getCurrentCUDAStream()); });
```
Similar for the backward pass.

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -229,8 +229,8 @@ def __init__(
         amax_epsilon: float = 0.0,
     ) -> None:
         super().__init__(rowwise=rowwise, columnwise=columnwise)
-        self.scale = torch.empty(1, dtype=torch.float32, device=device)
-        self.amax = torch.empty(1, dtype=torch.float32, device=device)
+        self.scale = torch.ones(1, dtype=torch.float32, device=device)
+        self.amax = torch.zeros(1, dtype=torch.float32, device=device)
```
timmoon10: This will add some kernel launches when constructing the quantizer. The values in these tensors will be wiped out each time the quantizer is used, so the initial value shouldn't have an effect.
```suggestion
        self.scale = torch.empty(1, dtype=torch.float32, device=device)
        self.amax = torch.empty(1, dtype=torch.float32, device=device)
```

transformer_engine/pytorch/csrc/extensions/activation.cpp
janekb04: I changed it to zero out the amax in [`create_hp_tensor_with_amax`](https://github.com/NVIDIA/TransformerEngine/pull/2004/commits/e4c40223b820dd00ec206f2ceb1f140b703ec6b8).

PR #2012, opened 2025-07-30
[PyTorch] Add FP8 attention with current scaling

cyanguwa:
# Description

This PR adds support for FP8 attention with current scaling through the `FusedAttention` backend. 
- requirement: cuDNN 9.13.1+, cudnn-frontend 1.14
- supported arch: Blackwell
- support details: non-CP, CP (P2P, A2A, A2A+P2P)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Add FP8 attention with current scaling.

# Checklist:


zhujian19891203:
Hi, I am trying this PR recently, but 3rdparty cudnn_frontend is your private server, and I donot have access.

tests/pytorch/attention/run_attention_with_cp.py
```diff
@@ -164,14 +168,20 @@ def run_dpa_with_cp(
     qkv_format="bshd",
     kernel_backend="FlashAttention",
     cp_comm_type="p2p",
-    fp8_mha=False,
+    fp8_bwd="True",
+    fp8_dpa="False",
+    fp8_mha="False",
```
ptrendx, line 173: Why string and not boolean values?

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -1672,6 +1673,13 @@ void fused_attn_fp8_fwd_impl_v1(
   auto bias_h = h;
   NVTE_CHECK(~is_bias, "FP8 fused attention does not support pre/post_scale_bias yet!");
   NVTE_CHECK(~is_alibi, "FP8 fused attention does not support ALiBi yet!");
+  bool is_current_scaling = (o_tensor_type == cudnn_frontend::DataType_t::HALF ||
+                             o_tensor_type == cudnn_frontend::DataType_t::BFLOAT16);
+  bool is_delayed_scaling = (o_tensor_type == cudnn_frontend::DataType_t::FP8_E4M3 ||
+                             o_tensor_type == cudnn_frontend::DataType_t::FP8_E5M2);
```
ptrendx, line 1679: Seeing how you only use it to figure out whether to scale O or not, it should be just "is_O_in_F16" - right now the names make it confusing.

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -1978,6 +1999,15 @@ void fused_attn_fp8_bwd_impl_v1(
   auto bias_h = h;
   NVTE_CHECK(~is_bias, "FP8 fused attention does not support pre/post_scale_bias yet!");
   NVTE_CHECK(~is_alibi, "FP8 fused attention does not support ALiBi yet!");
+  bool is_current_scaling = (dqkv_tensor_type == cudnn_frontend::DataType_t::HALF ||
+                             dqkv_tensor_type == cudnn_frontend::DataType_t::BFLOAT16);
+  bool is_delayed_scaling = (dqkv_tensor_type == cudnn_frontend::DataType_t::FP8_E4M3 ||
+                             dqkv_tensor_type == cudnn_frontend::DataType_t::FP8_E5M2);
```
ptrendx, line 2005: That is also brittle and should be just "is_dQKV_in_F16".

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -130,6 +135,58 @@
 
     fa_utils.set_flash_attention_3_params()
 
+# Float8CurrentScaling: fused_attn_bwd takes O in FP8 by default, this flag allows it in F16
+_dpa_fp8_cs_o_in_f16 = os.getenv("NVTE_DPA_FP8CS_O_in_F16", "0") == "1"
```
ptrendx: Considering the convergence experiments we should always have O in FP16, right?

ptrendx: Then why even having this environment variable (and having it off by default)?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -130,6 +135,58 @@
...
+
+
+class FP8EmulationFunc(torch.autograd.Function):
```
ptrendx, line 142: Leftover from debugging?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1026,42 +1158,54 @@ def forward(
...
+            # save appropriate tensors
+            fp8_tensors = (None, None, None, None)
+            qkvo_tensors = (None, None, None, None)
+            if is_bwd_fp8:
+                if fp8_recipe.float8_current_scaling() and _dpa_fp8_cs_o_in_f16:
+                    fp8_tensors = (q_fp8, k_fp8, v_fp8, None)
+                    qkvo_tensors = (None, None, None, out)
+                else:
+                    fp8_tensors = (q_fp8, k_fp8, v_fp8, out_fp8)
+            if not is_bwd_fp8:
```
ptrendx: Just `else:`?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1155,17 +1309,16 @@ def forward(
...
-            ), "Gradient of the DPA output must be in Float8Tensor type for FP8 MHA."
 
-        # FP16/BF16 attn:                  fake_dtype = torch.float16 or torch.bfloat16
-        # FP8 attn, is_output_fp8 = False: fake_dtype = torch.float16 or torch.bfloat16
-        # FP8 attn, is_output_fp8 = True:  fake_dtype = torch.float8_e5m2
-        fake_dtype = d_out.dtype
-
-        d_out = d_out.contiguous()
+        # corner case:
+        # d_out is expected to be in FP8 if is_output_fp8=True,
```
ptrendx, line 1313: That actually is not a corner case - O being in FP8 does not mean that dO will be provided to you in FP8 (and in most cases it will not be).

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1258,40 +1450,47 @@ def backward(ctx, d_out):
...
-                            dv = dv_fp8.dequantize()
-                    else:
-                        dq, dk, dv = dq_fp8, dk_fp8, dv_fp8
+                    # dq_fp8, dk_fp8, dv_fp8: Float8Tensor; dtype = torch.float16 or torch.bfloat16
+                    #                                       fp8_dtype = tex.DType.kFloat8E4M3
+                    # dq, dk, dv:             torch.Tensor; dtype = torch.float16 or torch.bfloat16
+                    dq_fp8, dk_fp8, dv_fp8 = dq_, dk_, dv_
+                    dq, dk, dv = dq_, dk_, dv_
+                    if (
+                        all(isinstance(x, Float8Tensor) for x in [dq_, dk_, dv_])
```
ptrendx: If we know that all of them share the same type then we should check only one. If we don't know that then it should use any rather than all.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1258,40 +1450,47 @@ def backward(ctx, d_out):
...
+                            src_nominal_dtype=dq_fp8.dtype,
+                        )
+                    if (
+                        not all(isinstance(x, Float8Tensor) for x in [dq_, dk_, dv_])
+                        and ctx.is_input_fp8
+                    ):
+                        # return dq_fp8, dk_fp8, dv_fp8
+                        dq, dk, dv = combine_and_quantize(
+                            ctx.qkv_layout, dq, dk, dv, ctx.dQKV_quantizer
+                        )
```
ptrendx, line 1468: This whole part seems strange to me. At the end of this dq, dk, dv are either dequantized or quantized depending on ctx.is_input_fp8. On the other hand we have dq_fp8,... variants of those and those may not actually be fp8 (since they are the same as dq_,...). Maybe we should keep an invariant that e.g. dq_fp8 at the end of this will for sure be fp8?

ptrendx: The comment even says "return dq_fp8". In practice it seems that those _fp8 versions are not actually used anywhere so maybe it would be better to get rid of them?

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -54,6 +59,10 @@
 # NVTE_DEBUG_LEVEL = 0/1/2 # enables more and more verbose debug mode, default = 0
 _NVTE_DEBUG_LEVEL = int(os.getenv("NVTE_DEBUG_LEVEL", "0"))
 _NVTE_FLASH_ATTN = int(os.getenv("NVTE_FLASH_ATTN", "1"))
+# Print debug info
+_to_print = os.getenv("NVTE_PRINT", "0") == "1"
+_to_print_layer = int(os.getenv("NVTE_PRINT_LAYER_NUMBER", "1"))
+_to_print_rank = int(os.getenv("NVTE_PRINT_RANK", "0"))
```
ptrendx: Please fold this with debug logging

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -1897,22 +1908,139 @@ def get_attention_quantizers(fp8, quantizers, cp_specific_quantizers=False):
...
+        "dQKV_quantizer",
+    ]
+    quantizers = [
+        QKV_quantizer,
+        S_quantizer,
+        O_quantizer,
+        dO_quantizer,
+        dP_quantizer,
+        dQKV_quantizer,
+    ]
```
ptrendx: Could you move the creation of those lists inside the if statement to not increase the CPU overhead?

transformer_engine/pytorch/attention/multi_head_attention.py
```diff
@@ -31,7 +32,13 @@
...
 from transformer_engine.pytorch.attention.inference import InferenceParams
 from transformer_engine.pytorch.attention.rope import apply_rotary_pos_emb
-from transformer_engine.pytorch.tensor.quantized_tensor import QuantizedTensor
+
+# Force DotProductAttention to use a different recipe than the fp8_recipe set in fp8_autocast().
+# Useful when GEMMs and attention use different recipes. Supported values are "DelayedScaling"
+# and "Float8CurrentScaling". Use other relevant variables here to define the recipe, e.g. fp8_dpa.
+_dpa_fp8_recipe = os.getenv("NVTE_DPA_FP8_RECIPE", "")
+_dpa_fp8_recipe_dpa = os.getenv("NVTE_DPA_FP8_RECIPE_DPA", "0") == "1"
+_dpa_fp8_recipe_mha = os.getenv("NVTE_DPA_FP8_RECIPE_MHA", "0") == "1"
```
ptrendx, line 41: We should probably say here that it is experimental/temporary and subject to change.

transformer_engine/pytorch/attention/multi_head_attention.py
```diff
@@ -570,15 +577,23 @@ def set_context_parallel_group(
             self.cp_size = get_distributed_world_size(cp_group)
             self.cp_rank = get_distributed_rank(cp_group)
         elif isinstance(cp_group, list):
-            assert len(cp_group) == 2, "Current implementation only supports two-level CP groups!"
+            assert len(cp_group) == 3, (
```
ptrendx: This is a breaking change in the API, no? That is not good. Also, the docstring still says that it should be list of size 2.

transformer_engine/common/fused_attn/fused_attn_fp8.cu
cyanguwa: The data types also change for dQKV: CS - F16, DS - FP8.

tests/pytorch/attention/run_attention_with_cp.py
cyanguwa: I think it's to do with the subprocess launch - we're passing args to a bash script so that might just interpret them as strings. This has been around for a while. I can take a look at changing to booleans if I have time on this PR.

transformer_engine/common/fused_attn/fused_attn_fp8.cu
cyanguwa: There is also the passing pointers or not passing pointers. We are using 1.0f for some amaxes I believe. So I'm creating an illusion that we're just supporting two "recipes" and any other combination wouldn't work.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -385,6 +385,15 @@ std::pair<TensorWrapper, py::object> Float8CurrentScalingQuantizer::create_hp_te
   return {std::move(out_cpp), std::move(out_py)};
 }
 
+std::pair<TensorWrapper, py::object> Float8CurrentScalingQuantizer::create_hp_tensor_with_amax(
```
ptrendx: We could also do it as data being optional parameter to the existing function to not duplicate the logic.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
cyanguwa: I don't know if there is any benefit in using FP8 O in bwd. F16 O kernel is slower than FP8 O, and some models might be ok with FP8 O.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
cyanguwa: Is this meant for a different place in the code? Which part looks like redundant, the _dpa_fp8_cs_o_in_f16?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
cyanguwa: These might be useful when we want to output FP8 dQKV. I can double check on the cleanliness of the code.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
cyanguwa: Sure.

transformer_engine/pytorch/attention/multi_head_attention.py
cyanguwa: Sounds good.

cyanguwa: You might be viewing an older commit of this branch by the way, judging by the lines you commented on.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
cyanguwa: Done.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
cyanguwa: Changed it to just checking one tensor. They are the outputs of bwd and are always in the same type.

transformer_engine/pytorch/attention/multi_head_attention.py
cyanguwa: This is from an older commit. The latest has updated this.

transformer_engine/pytorch/csrc/quantizer.cpp
cyanguwa: Done.

transformer_engine/pytorch/attention/multi_head_attention.py
cyanguwa: Done - labelled it experimental.

transformer_engine/common/fused_attn/fused_attn_fp8.cu
cyanguwa: Will think more about this in a future cleanup PR.

transformer_engine/common/fused_attn/fused_attn_fp8.cu
cyanguwa: Same as the last comment. Will think more about this in a future cleanup PR.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
cyanguwa: Will change the default to 1 for now.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
cyanguwa: Removed _ fp8 variables and now just using dx_ and dx.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
cyanguwa: Done.

cyanguwa:
Pipeline 35778554 for the full run before the last commit.

ptrendx (approved):

PR #2014, opened 2025-07-30
Dropout with 8-bit RNG

vasunvidia:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -1749,7 +1749,7 @@ class TestBasicOps:
         torch.testing.assert_close(y_test, y_ref, **tols)
         torch.testing.assert_close(dx_test, x_ref.grad, **tols)
 
-    @pytest.mark.parametrize("prob", (0.1, 0.5, 0.75))
+    @pytest.mark.parametrize("prob", (0.0625, 0.5, 0.75))
     @pytest.mark.parametrize("is_training", (True, False))
     @pytest.mark.parametrize("quantization", (None, "fp8_current_scaling"))
     @pytest.mark.parametrize("shape", ((101,), (2, 4, 16), (64, 64)))
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -1817,7 +1817,9 @@ class TestBasicOps:
         if is_training:
             prob_observed = 1 - torch.count_nonzero(y_test).item() / y_test.numel()
             z_score = (prob_observed - prob) / math.sqrt(prob * (1 - prob) / y_test.numel())
-            assert abs(z_score) < 2.5758, "Number of zeros is outside 99% confidence interval"
+            assert (
+                abs(z_score) < 2.5758
+            ), f"Number of zeros is outside 99% confidence interval ({prob=}, {prob_observed=})"
 
 
 class TestFusedOps:
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -1752,7 +1752,7 @@ class TestBasicOps:
     @pytest.mark.parametrize("prob", (0.0625, 0.5, 0.75))
     @pytest.mark.parametrize("is_training", (True, False))
     @pytest.mark.parametrize("quantization", (None, "fp8_current_scaling"))
-    @pytest.mark.parametrize("shape", ((101,), (2, 4, 16), (64, 64)))
+    @pytest.mark.parametrize("shape", ((101,), (2, 4, 16), (128, 128)))
     @pytest.mark.parametrize("dtype", _dtypes)
     def test_dropout(
         self,
@@ -1770,6 +1770,7 @@ class TestBasicOps:
         maybe_skip_quantization(quantization, dims=shape, device=device)
 
         # Random data
+        # Note: Shift values to make sure inputs are non-zero
         x_ref, x_test = make_reference_and_test_tensors(
             shape,
             quantization=quantization,
@@ -1777,6 +1778,9 @@ class TestBasicOps:
             test_device=device,
             test_is_quantized=quantized_input,
         )
+        with torch.no_grad():
+            x_test += 1
+            x_ref.copy_(x_test)
         dy_ref, dy_test = make_reference_and_test_tensors(
             shape,
             test_dtype=dtype,
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -1751,23 +1751,38 @@ class TestBasicOps:
 
     @pytest.mark.parametrize("prob", (0.1, 0.5, 0.75))
     @pytest.mark.parametrize("is_training", (True, False))
-    @pytest.mark.parametrize("shape", ((101,), (2, 4, 16)))
+    @pytest.mark.parametrize("quantization", (None, "fp8_current_scaling"))
+    @pytest.mark.parametrize("shape", ((101,), (2, 4, 16), (64, 64)))
     @pytest.mark.parametrize("dtype", _dtypes)
     def test_dropout(
         self,
         *,
         prob: float,
         is_training: bool,
+        quantization: Optional[str],
         shape: Iterable[int],
         dtype: torch.dtype,
         device: torch.device = "cuda",
     ):
 
+        # Skip invalid configurations
+        quantized_input = quantization is not None
+        maybe_skip_quantization(quantization, dims=shape, device=device)
+
         # Random data
-        x_ref = torch.rand(shape, dtype=dtype, device=device) + 0.5
-        x_test = x_ref.clone().requires_grad_()
-        dy_ref = torch.rand(shape, dtype=dtype, device=device) + 0.5
-        dy_test = dy_ref.clone()
+        x_ref, x_test = make_reference_and_test_tensors(
+            shape,
+            quantization=quantization,
+            test_dtype=dtype,
+            test_device=device,
+            test_is_quantized=quantized_input,
+        )
+        dy_ref, dy_test = make_reference_and_test_tensors(
+            shape,
+            test_dtype=dtype,
+            test_device=device,
+            requires_grad=False,
+        )
 
         # Apply dropout
         op = te_ops.Dropout(prob)
@@ -1775,17 +1790,20 @@ class TestBasicOps:
             op.train()
         else:
             op.eval()
-        y = op(x_test)
-        y.backward(dy_test)
+        y_test = op(x_test)
+        y_test.backward(dy_test)
 
         # Check values
+        y_test = y_test.to(dtype=torch.float64, device="cpu")
+        dx_test = x_test.grad.to(dtype=torch.float64, device="cpu")
         if is_training:
-            mask = ((y != 0) / (1 - prob)).to(dtype=dtype)
-            torch.testing.assert_close(y, x_ref * mask)
-            torch.testing.assert_close(x_test.grad, dy_ref * mask)
+            tols = dtype_tols(dtype)
+            mask = ((y_test != 0) / (1 - prob)).to(dtype=dtype)
+            torch.testing.assert_close(y_test, x_ref * mask, **tols)
+            torch.testing.assert_close(dx_test, dy_ref * mask, **tols)
         else:
-            torch.testing.assert_close(y, x_ref, rtol=0, atol=0)
-            torch.testing.assert_close(x_test.grad, dy_ref, rtol=0, atol=0)
+            torch.testing.assert_close(y_test, x_ref, rtol=0, atol=0)
+            torch.testing.assert_close(dx_test, dy_ref, rtol=0, atol=0)
 
         # Hypothesis testing for number of zeros
         # Note: A Bernoulli random variable with probability p has
@@ -1797,8 +1815,8 @@ class TestBasicOps:
         # p-value is less than 1% and we assume that the dropout
         # distribution is incorrect.
         if is_training:
-            prob_observed = 1 - torch.count_nonzero(y).item() / y.numel()
-            z_score = (prob_observed - prob) / math.sqrt(prob * (1 - prob) / y.numel())
+            prob_observed = 1 - torch.count_nonzero(y_test).item() / y_test.numel()
+            z_score = (prob_observed - prob) / math.sqrt(prob * (1 - prob) / y_test.numel())
             assert abs(z_score) < 2.5758, "Number of zeros is outside 99% confidence interval"
 
 
--- a/transformer_engine/common/dropout/dropout.cu
+++ b/transformer_engine/common/dropout/dropout.cu
@@ -1,5 +1,5 @@
 /*************************************************************************
- * Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  *
  * See LICENSE for license information.
  ************************************************************************/
--- a/transformer_engine/common/dropout/dropout.cu
+++ b/transformer_engine/common/dropout/dropout.cu
@@ -25,11 +25,11 @@ constexpr size_t rng_chunk_size = 16;
 constexpr size_t block_size = 128;
 
 // Vector class to help with vectorized memory accesses
-template <typename T, size_t size>
+template <typename T, size_t kSize>
 union Vector {
-  using StorageType = typename BytesToType<sizeof(T) * size>::Type;
+  using StorageType = typename BytesToType<sizeof(T) * kSize>::Type;
   StorageType storage;
-  T entries[size];
+  T entries[kSize];
 };
 
 /* Byte-wise less-than comparison
--- a/transformer_engine/common/dropout/dropout.cu
+++ b/transformer_engine/common/dropout/dropout.cu
@@ -209,8 +209,6 @@ void dropout_fwd(const Tensor &input, Tensor &output, Tensor &mask, Tensor &rng_
              dropout_probability, ").");
   const float scale = 1 / (1 - dropout_probability);
   const uint8_t prob_uint8 = static_cast<uint8_t>(std::floor(dropout_probability * 256));
-  NVTE_CHECK(prob_uint8 == dropout_probability * 256, "Dropout probability (", dropout_probability,
-             ") is not representable in 8 bits");
   uint32_t prob4 = prob_uint8;
   prob4 = (prob4 << 8) | prob4;
   prob4 = (prob4 << 16) | prob4;
--- a/transformer_engine/common/dropout/dropout.cu
+++ b/transformer_engine/common/dropout/dropout.cu
@@ -1,0 +1,356 @@
+/*************************************************************************
+ * Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+
+#include "transformer_engine/dropout.h"
+
+#include <algorithm>
+#include <cmath>
+
+#include <curand.h>
+#include <curand_kernel.h>
+#include <curand_philox4x32_x.h>
+
+#include "../common.h"
+
+namespace transformer_engine {
+namespace {
+
+struct ull2 {
+  uint64_t x;
+  uint64_t y;
+};
+
+inline __device__ uint16_t packed_le_8bit(unsigned A, unsigned B) {
+  unsigned T = A & 0x7F7F7F7F;
+  unsigned D = B | 0x80808080;
+  D = -T + D;
+  T = (A ^ B) | 0x7F7F7F7F;
+  D = ~(T ^ D);
+  asm("lop3.b32 %0, %1, %2, %3, 0x4d;\n\t" : "=r"(D) : "r"(A), "r"(B), "r"(D));
+  D = D & 0x80808080;
+  T = (D & 0x0000FFFF) | (D >> 17);
+  return T;
+}
+
+template <typename T>
+__global__ void dropout_kernel_fwd_f16(const T *input,
+                                       T *output,
+                                       uint16_t *mask,
+                                       int64_t *rng_state,
+                                       int n_rng_blocks,
+                                       unsigned p_dropout_in_uint32_t,
+                                       float inv_prob) {
+  int tid = (blockIdx.y * gridDim.x + blockIdx.x) * blockDim.x + threadIdx.x;
+  int num_threads = gridDim.y * gridDim.x * blockDim.x;
+  uint16_t result = 0;
+  size_t ofs = tid;
+  curandStatePhilox4_32_10_t state;
+  curand_init(rng_state[0], ofs, rng_state[1], &state);
+  ull2 input8;
+  const ull2 *gmem_input_ptr = reinterpret_cast<const ull2 *>(input);
+  ull2 *gmem_output_ptr = reinterpret_cast<ull2 *>(output);
+
+  while (tid < n_rng_blocks) {
+    uint4 random_uint4 = curand4(&state);
+    result = packed_le_8bit(random_uint4.x, p_dropout_in_uint32_t);
+    result |= packed_le_8bit(random_uint4.y, p_dropout_in_uint32_t) >> 2;
+    result |= packed_le_8bit(random_uint4.z, p_dropout_in_uint32_t) >> 4;
+    result |= packed_le_8bit(random_uint4.w, p_dropout_in_uint32_t) >> 6;
+
+    uint16_t result_copy = result;
+#pragma unroll
+    for (int j = 0; j < 2; j++) {
+      input8 = gmem_input_ptr[tid * 2 + j];
+      T *input_h_8 = reinterpret_cast<T *>(&input8);
+
+#pragma unroll
+      for (int i = 0; i < 8; i++) {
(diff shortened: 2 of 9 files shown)


tests/pytorch/test_fusible_ops.py
```diff
@@ -1753,7 +1753,8 @@ def test_dropout(
         else:
             op.eval()
         y = op(x_test)
-        y.backward(dy_test)
+        if is_training:
+            y.backward(dy_test)
```
timmoon10: [Eval mode does not disable autograd](https://docs.pytorch.org/docs/stable/notes/autograd.html#evaluation-mode-nn-module-eval), e.g. if the model is being fine-tuned.
```suggestion
        y.backward(dy_test)
```

tests/pytorch/test_fusible_ops.py
```diff
@@ -1728,8 +1728,8 @@ def test_constant_scale(
 
     @pytest.mark.parametrize("prob", (0.1, 0.5, 0.75))
     @pytest.mark.parametrize("is_training", (True, False))
-    @pytest.mark.parametrize("shape", ((101,), (2, 4, 16)))
-    @pytest.mark.parametrize("dtype", _dtypes)
+    @pytest.mark.parametrize("shape", ((4096, 4096), (1024, 1024)))
+    @pytest.mark.parametrize("dtype", (torch.half, torch.bfloat16))
```
timmoon10: FP32 is TE's default dtype, so we should support it, even if it's suboptimal.
```suggestion
    @pytest.mark.parametrize("dtype", _dtypes)
```

tests/pytorch/test_fusible_ops.py
```diff
@@ -1773,10 +1773,61 @@ def test_dropout(
...
         # p-value is less than 1% and we assume that the dropout
         # distribution is incorrect.
-        if is_training:
-            prob_observed = 1 - torch.count_nonzero(y).item() / y.numel()
-            z_score = (prob_observed - prob) / math.sqrt(prob * (1 - prob) / y.numel())
-            assert abs(z_score) < 2.5758, "Number of zeros is outside 99% confidence interval"
+        # if is_training:
+        #    prob_observed = 1 - torch.count_nonzero(y).item() / y.numel()
+        #    z_score = (prob_observed - prob) / math.sqrt(prob * (1 - prob) / y.numel())
+        #    assert abs(z_score) < 2.5758, "Number of zeros is outside 99% confidence interval"
```
timmoon10: There's a correctness bug if this test is failing:
```suggestion
        if is_training:
           prob_observed = 1 - torch.count_nonzero(y).item() / y.numel()
           z_score = (prob_observed - prob) / math.sqrt(prob * (1 - prob) / y.numel())
           assert abs(z_score) < 2.5758, "Number of zeros is outside 99% confidence interval"
```

tests/pytorch/test_fusible_ops.py
```diff
@@ -1728,8 +1728,8 @@ def test_constant_scale(
 
     @pytest.mark.parametrize("prob", (0.1, 0.5, 0.75))
     @pytest.mark.parametrize("is_training", (True, False))
-    @pytest.mark.parametrize("shape", ((101,), (2, 4, 16)))
-    @pytest.mark.parametrize("dtype", _dtypes)
+    @pytest.mark.parametrize("shape", ((4096, 4096), (1024, 1024)))
```
timmoon10: How difficult is it to support arbitrary dimensions, even suboptimally?

tests/pytorch/test_fusible_ops.py
```diff
@@ -1773,10 +1773,61 @@ def test_dropout(
...
+        # if is_training:
+        #    prob_observed = 1 - torch.count_nonzero(y).item() / y.numel()
+        #    z_score = (prob_observed - prob) / math.sqrt(prob * (1 - prob) / y.numel())
+        #    assert abs(z_score) < 2.5758, "Number of zeros is outside 99% confidence interval"
+
+    @pytest.mark.parametrize("prob", (0.1, 0.5, 0.75))
+    @pytest.mark.parametrize("is_training", (True,))
+    @pytest.mark.parametrize("shape", ((4096, 4096), (1024, 1024)))
+    @pytest.mark.parametrize("dtype", (torch.half, torch.bfloat16))
+    def test_dropout_fp8(
```
timmoon10: I'd prefer including FP8 options in `test_dropout` instead of making a new test function.

transformer_engine/pytorch/ops/basic/dropout.py
```diff
@@ -25,9 +26,10 @@ class Dropout(BasicOperation):
 
     """
 
-    def __init__(self, p: float) -> None:
+    def __init__(self, p: float, otype: Optional[torch.dtype] = None) -> None:
```
timmoon10: This name is non-standard. Most PyTorch ops would name this `dtype`:
```suggestion
    def __init__(self, p: float, dtype: Optional[torch.dtype] = None) -> None:
```
Better yet, could we deduce based on the `Float8Tensor`'s dtype?

transformer_engine/common/dropout/dropout.cu
```diff
@@ -0,0 +1,284 @@
+/*************************************************************************
+ * Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+
+#include <curand.h>
+#include <curand_kernel.h>
+#include <curand_philox4x32_x.h>
```
timmoon10, line 9: If we add cuRAND as a dependency, we may need to add logic to manually load `libcurand.so`, similar to how we load `libcublas.so`:
https://github.com/NVIDIA/TransformerEngine/blob/858755c0c3898d5bc971fcec031b26ff26e7e4fb/transformer_engine/common/__init__.py#L306-L307
This is because users might not have the CUDA Toolkit installed, but rather install CUDA libs with [Pip](https://pypi.org/project/nvidia-curand-cu12/).

We should also make sure cuRAND is linked during the build process: https://github.com/NVIDIA/TransformerEngine/blob/dbdedd30ef5cd78ccbf3c44774b72a2633c1880f/transformer_engine/common/CMakeLists.txt#L120-L123

transformer_engine/common/include/transformer_engine/dropout.h
```diff
@@ -0,0 +1,60 @@
...
+ *  \param[out]    output           Output tensor.
+ *  \param[out]    mask             Mask tensor.
+ *  \param[in]     rng_state        RNG engine inputs.
+ *  \param[in]     dropout_probability Dropout probability.
+ *  \param[in]     stream           CUDA stream used for this operation.
+ */
+void nvte_dropout_fwd(NVTETensor input, NVTETensor output, NVTETensor mask, NVTETensor rng_state,
+                      float dropout_probability, cudaStream_t stream);
+
+/*!  \brief Dropout forward kernel with FP8 input.
```
timmoon10:
```suggestion
/*!  \brief Dropout forward kernel with FP8 input and high-precision output.
```

transformer_engine/common/include/transformer_engine/dropout.h
```diff
@@ -0,0 +1,60 @@
...
+ *
+ *  \param[in]     input            Input tensor (FP8).
+ *  \param[out]    output           Output tensor (dequantized to FP16/BF16).
+ *  \param[out]    mask             Mask tensor.
+ *  \param[in]     rng_state        RNG engine inputs.
+ *  \param[in]     dropout_probability Dropout probability.
+ *  \param[in]     stream           CUDA stream used for this operation.
+ */
+void nvte_dropout_fwd_fp8(NVTETensor input, NVTETensor output, NVTETensor mask,
```
timmoon10: Most of TE's FP8 functions have high precision input and FP8 output. This function is doing something unusual, so we should give it a more explicit name:
```suggestion
void nvte_dropout_fwd_dequantize_fp8(NVTETensor input, NVTETensor output, NVTETensor mask,
```
Also, this function is redundant with `nvte_dropout_fwd`, at least in principle. We could make that function check the inputs and outputs, and call a different kernel if the input is FP8 and the output is high precision.

transformer_engine/common/include/transformer_engine/dropout.h
```diff
@@ -0,0 +1,60 @@
+/*************************************************************************
+ * Copyright (c) 2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+
+/*! \file dropout.h
+ *  \brief Enums and functions for fused attention.
```
timmoon10:
```suggestion
 *  \brief Functions for dropout.
```

transformer_engine/common/include/transformer_engine/dropout.h
```diff
@@ -0,0 +1,60 @@
...
+ */
+
+#ifndef TRANSFORMER_ENGINE_DROPOUT_FP8_H_
+#define TRANSFORMER_ENGINE_DROPOUT_FP8_H_
+
+#include "stdint.h"
```
timmoon10: I don't see anything in this header that needs anything from `stdint.h`. It is good C++ practice to only ["include what you use"](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md).
```suggestion
```

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -242,6 +242,19 @@ std::vector<py::object> dbias_dqgelu(const at::Tensor &grad_output, const at::Te
...
                                      py::handle quantizer);
 
+/***************************************************************************************************
+ * Dropout
+ **************************************************************************************************/
+
+std::vector<py::object> dropout_fwd(const at::Tensor &input, const float dropout_probability,
+                                    const bool is_training);
+
+std::vector<py::object> dropout_fwd_fp8(const py::handle &input, at::Tensor &output,
```
timmoon10: `dropout_fwd_fp8` makes me think the output is FP8:
```suggestion
std::vector<py::object> dropout_fwd_dequantize_fp8(const py::handle &input, at::Tensor &output,
```

transformer_engine/pytorch/csrc/extensions/dropout.cpp
```diff
@@ -0,0 +1,133 @@
...
+  const int rng_block_size = 16;
+  int numel = input.numel();
+
+  // Allocate output tensor
+  std::vector<int> output_shape;
+  for (auto s : input.sizes()) {
+    output_shape.emplace_back(static_cast<int>(s));
+  }
+  auto output = allocateTorchTensor(output_shape[0], output_shape[1], input_tensor.dtype());
+  //auto output = at::empty(output_shape, input_tensor.dtype());
```
janekb04: Why are there some commented out lines of code?

transformer_engine/pytorch/csrc/extensions/dropout.cpp
```diff
@@ -0,0 +1,133 @@
...
+  auto output_tensor = makeTransformerEngineTensor(output);
+
+  // Allocate mask tensor
+  std::vector<size_t> mask_shape{input.numel() / 16};
+  auto mask = allocateTorchTensor(mask_shape[0], DType::kInt16);
+  //auto mask = at::empty(mask_shape, at::CUDA(at::kUInt16));
```
janekb04: Why are there some commented out lines of code?

transformer_engine/pytorch/csrc/extensions/dropout.cpp
```diff
@@ -0,0 +1,133 @@
...
+  auto input_tensor = makeTransformerEngineTensor(input, none);
+  const int rng_block_size = 16;
+  auto input_shape = input_tensor.shape();
+  int numel = input_shape.data[0] * input_shape.data[1];
+
+  auto output_tensor = makeTransformerEngineTensor(output);
+  // Allocate mask tensor
+  std::vector<size_t> mask_shape{numel / 16};
+  auto mask = allocateTorchTensor(mask_shape[0], DType::kInt16);
+  //auto mask = at::empty(mask_shape, at::CUDA(at::kUInt16));
```
janekb04: Why are there some commented out lines of code?

transformer_engine/pytorch/ops/basic/dropout.py
```diff
@@ -38,21 +40,39 @@ def op_forward(
...
+            assert not isinstance(
+                input_, Float8TensorBase
+            ), "fp8 dropout does not support non-training mode"
         # Save context for backward
         if ctx.requires_grad:
             ctx.save_for_backward(mask)
-            ctx.is_training = is_training
-
+            ctx.dropout_probability = self.dropout_probability
+            # ctx.is_training = is_training
```
janekb04: Why are there some commented out lines of code?

transformer_engine/pytorch/ops/basic/dropout.py
```diff
@@ -61,7 +81,7 @@ def op_backward(
         grad_output: torch.Tensor,
     ) -> tuple[torch.Tensor, tuple[()]]:
         (mask,) = ctx.saved_tensors
-        grad_input = grad_output
-        if ctx.is_training:
-            grad_input = grad_input * mask
+        # grad_input = grad_output
+        # grad_input = grad_output * mask
```
janekb04: Why are there some commented out lines of code?

transformer_engine/pytorch/ops/basic/dropout.py
```diff
@@ -38,21 +40,39 @@ def op_forward(
...
+                mask.bernoulli_(keep_prob)
+                mask *= 1 / keep_prob
+                out = out * mask
+            else:
+                if isinstance(input_, Float8TensorBase):
+                    assert self.otype is not None, "otype is not set for fp8 dropout"
+                    assert self.otype in [torch.half, torch.bfloat16], (
+                        "fp8 dropout only supports half and bfloat16 output types but got %s"
+                        % self.otype
+                    )
```
janekb04: This handling of `dtype` is inconsistent with other operations. For example, in `BasicLinear.__init__`:
```python
        dtype = canonicalize_dtype(dtype)
        if dtype not in (torch.float32, torch.float16, torch.bfloat16):
            raise ValueError(f"Supported dtypes are float32, float16, bfloat16 (got {dtype})")
```
Also autocast dtype is not taken into account. For example, from `BasicLinear.op_forward`:
```python
        # Get autocast dtype if needed
        if torch.is_autocast_enabled():
            dtype = torch.get_autocast_dtype("cuda")
        else:
            dtype = self.weight.dtype
```

transformer_engine/pytorch/ops/basic/dropout.py
```diff
@@ -38,21 +40,39 @@ def op_forward(
...
-            keep_prob = 1 - self.dropout_probability
-            mask = torch.empty_like(input_)
-            mask.bernoulli_(keep_prob)
-            mask *= 1 / keep_prob
-            out = out * mask
-
+            if input_.numel() % 16 != 0:
+                assert not isinstance(
+                    input_, Float8TensorBase
+                ), "fp8 dropout does not support non-16-element aligned inputs"
```
janekb04: @timmoon10 Should we fail here or dequantize instead?

transformer_engine/pytorch/ops/basic/dropout.py
```diff
@@ -38,21 +40,39 @@ def op_forward(
...
+                        % self.otype
+                    )
+                    out = torch.empty_like(input_, dtype=self.otype)
+                    _, mask = tex.dropout_fwd_fp8(input_, out, self.dropout_probability)
+                else:
+                    out, mask = tex.dropout_fwd(input_, self.dropout_probability, is_training)
+        else:
+            assert not isinstance(
+                input_, Float8TensorBase
+            ), "fp8 dropout does not support non-training mode"
```
janekb04: ~~@timmoon10 I don't think that failing in inference mode is valid. I think if fp8 dropout cannot support inference, we should dequantize.~~ I forgot to think what dropout is while writing this. We should just pass the input through as Przemek said.

transformer_engine/common/dropout/dropout.cu
```diff
@@ -0,0 +1,284 @@
...
+  unsigned long long y;
+};
+
+inline __device__ uint16_t packed_le_8bit(unsigned A, unsigned B) {
+  unsigned T = A & 0x7F7F7F7F;
+  unsigned D = B | 0x80808080;
+  D = -T + D;
+  T = (A ^ B) | 0x7F7F7F7F;
+  D = ~(T ^ D);
+  asm("lop3.b32 %0, %1, %2, %3, 0x4d;\n\t" : "=r"(D) : "r"(A), "r"(B), "r"(D));
```
janekb04: I think that `asm` should be moved to a function in `ptx.cuh` or `utils.cuh`.

transformer_engine/pytorch/ops/basic/dropout.py
timmoon10: That would be a good approach. If we get unsupported data, e.g. FP8 with bad dims or MXFP8, then we can just dequantize to a plain PyTorch tensor and apply the high-precision kernel.

transformer_engine/common/include/transformer_engine/dropout.h
ptrendx: Right, the type of the input/output is given inside the input and output tensors so we do not need a special API. We should mention in the documentation of the nvte_dropout_fwd which types of the NVTETensors we support there.

transformer_engine/common/dropout/dropout.cu
```diff
@@ -0,0 +1,284 @@
...
+  ull2 input8;
+  ull2 *gmem_input_ptr = reinterpret_cast<ull2 *>(input);
+  ull2 *gmem_output_ptr = reinterpret_cast<ull2 *>(output);
+
+  while (tid < n_rng_blocks) {
+    uint4 random_uint4 = curand4(&state);
+    result = packed_le_8bit(random_uint4.x, p_dropout_in_uint32_t);
+    result |= packed_le_8bit(random_uint4.y, p_dropout_in_uint32_t) >> 2;
+    result |= packed_le_8bit(random_uint4.z, p_dropout_in_uint32_t) >> 4;
+    result |= packed_le_8bit(random_uint4.w, p_dropout_in_uint32_t) >> 6;
```
ptrendx: @vasunvidia could you give some explanation of what those lines do? Especially the shifts by 2/4/6 are strange to me - I would have expected shifts by 4/8/12 (since you take 4 random values in each line?).

Also an idea of the optimization for Blackwell - you should be able to use cvt.rs (stochastic rounding instruction) to do mostly the same thing as what this `packed_le_8bit` instruction is doing, I think.

transformer_engine/pytorch/ops/basic/dropout.py
ptrendx: Float8TensorBase does not have the dtype though.

transformer_engine/pytorch/ops/basic/dropout.py
ptrendx: Dropout in inference is a noop so we should just pass the input through, irrespective of the type.

transformer_engine/pytorch/csrc/extensions/dropout.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+#include <ATen/cuda/CUDAGraphsUtils.cuh>
+
+#include "common.h"
+#include "pybind.h"
+#include "transformer_engine/transformer_engine.h"
+
+namespace transformer_engine::pytorch {
+
+void unpack(at::PhiloxCudaState arg, int64_t *rng_state_ptr) {
```
ptrendx: We already have exactly the same function in attention.cpp, we should just reuse it.

tests/pytorch/test_fusible_ops.py
```diff
@@ -1773,10 +1773,61 @@ def test_dropout(
...
+        x_test = x_ref.clone().requires_grad_()
+        dy_ref = torch.ones(shape, dtype=dtype, device=device)
+        dy_test = dy_ref.clone()
+
+        fp8_recipe = make_recipe("fp8_delayed_scaling")
+        # Apply dropout
+        op = te_ops.Sequential(
+            te_ops.Quantize(forward=True, backward=False),
+            te_ops.Dropout(prob, otype=dtype),
+            te_ops.Quantize(forward=False, backward=False),
```
vasunvidia: Not required.

transformer_engine/common/dropout/dropout.cu
vasunvidia: Each call to packed_le_8bit compares 4 random byte values with p_dropout_in_uint32_t and produces a 16-bit output which has the 4-mask bits in 15,14,7,6 th bit positions. So I shift ith call's output by 2^i to get full 16-bit mask.

transformer_engine/pytorch/csrc/extensions/dropout.cpp
vasunvidia: @timmoon10 can you advise on how to move this function to common.h and reuse it in attention.cpp and dropout.cpp?

tests/pytorch/test_fusible_ops.py
vasunvidia: Added fallback to suboptimal code path for arbitrary shapes.

tests/pytorch/test_fusible_ops.py
vasunvidia: Added fallback to suboptimal code path for FP32 type.

tests/pytorch/test_fusible_ops.py
vasunvidia: Fixed this.

tests/pytorch/test_fusible_ops.py
vasunvidia: As discussed, I changed the p_dropout to 0.125 and the test passes.

tests/pytorch/test_fusible_ops.py
vasunvidia: Add is_fp8_input parameter to test_dropout. Also reduced tolerance for FP8 test.

transformer_engine/common/dropout/dropout.cu
vasunvidia: Need help with this.

transformer_engine/common/include/transformer_engine/dropout.h
vasunvidia: Fixed.

transformer_engine/common/include/transformer_engine/dropout.h
vasunvidia: Fixed.

transformer_engine/common/include/transformer_engine/dropout.h
vasunvidia: Fixed.

transformer_engine/common/include/transformer_engine/dropout.h
vasunvidia: Currently I have changed nvte_dropout_fwd_fp8 to nvte_dropout_fwd_dequantize_fp8. However I'm unsure how to pass nullptr for scale_inv to the kernel when the input dtype is not FP8. Also what Type switch should I use to include all 8- and 16- bit types?

transformer_engine/pytorch/csrc/extensions.h
vasunvidia: Fixed.

transformer_engine/pytorch/csrc/extensions/dropout.cpp
vasunvidia: Fixed.

transformer_engine/pytorch/csrc/extensions/dropout.cpp
vasunvidia: Fixed.

transformer_engine/pytorch/csrc/extensions/dropout.cpp
vasunvidia: Fixed.

transformer_engine/pytorch/ops/basic/dropout.py
timmoon10: TE ops don't actually support arbitrary usage of `Float8TensorBase`. Ops can request internal tensors (by setting `quantizer.internal=True`), but they shouldn't push internal tensors onto ops that don't support them. This is important so since many ops have naive fallback implementations that rely on the automatic dequantization in `__torch_dispatch__`.

transformer_engine/common/include/transformer_engine/dropout.h
timmoon10: I've reorganized into a unified API that chooses the appropriate kernel based on the provided `NVTETensor`s. I've also added checks for all of the kernel requirements.

transformer_engine/common/dropout/dropout.cu
timmoon10: I've added comments so future maintainers can actually understand this. I've also tweaked the bit shift logic so the bit pattern is less weird.

transformer_engine/common/dropout/dropout.cu
timmoon10: This operation is actually quite specific, so there isn't much opportunity to generalize by putting in a header. I've added comments to make its behavior clearer.

transformer_engine/pytorch/csrc/extensions/dropout.cpp
timmoon10: `unpack` is simple enough that I think it's simpler to just include it within `dropout_fwd`.

timmoon10 (approved):

PR #2018, opened 2025-08-01
[PyTorch] Fix cudagraph static_input and static_grad_input reuse

buptzyb:
# Description

- For `static_input`

The `static_input` (sample_args and sample_kwargs) is reused between microbatches that don't overlap their forward and backward scopes. This is based on the assumption that all the callables have the same input tensor signatures (the number of inputs, and their size, shape, etc.), e.g., `hidden_states` tensors with the same shape for all callables as sample_args, and `attention_mask` tensors with the same shape for all callables as sample_kwargs. However, for some models, different layers have different input signatures. One example is the mamba model, where the mamba layer only has one `hidden_states` input, while the attention layer has both `hidden_states` and `attention_mask` inputs. Therefore, this PR makes the reuse only happen between static inputs of the same signature.

- For `static_grad_input`

The `static_grad_input` of the last backward chunk will be sent to another PP rank no later than the finish of this backward chunk, so we can weak ref its buffer at the end of this backward chunk for future reuse. However, the current implementation ignores the case that there can be multiple layers per chunk, for example, two layers per chunk:

`[b1, b2, f1, f2, b3, b4, f3, f4, b5, b6, ...]`

In the current implementation, we weak ref b2 `static_grad_input` after b3 finishes. This is unsafe. It should be after b4 finishes.

This PR makes a new rule for the `static_grad_input` reuse:

1. the `static_grad_input` of the last layer of a backward chunk can be deallocated after its next backward chunk finishes.
2. the `static_grad_input` of other layers of a backward chunk can be deallocated after its next layer in the same chunk finishes.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Allow different input tensor signatures for static input reuse.
- Fix static grad input reuse logic.

# Checklist:


buptzyb:
Hi @timmoon10 , this is a small fix to the cudagraph static memory reuse. Could you help review and make it to r2.6? Thanks!

transformer_engine/pytorch/graph.py
```diff
@@ -513,13 +514,19 @@ def hook_fn(module, inputs, outputs):  # pylint: disable=unused-argument
                         # grad input to another pipeline parallel rank and that the
                         # communication is finished before the end of the next backward
                         # pass.
```
timmoon10: This comment is no longer in the right place and is now misleading. For code this convoluted, we need clear documentation to understand what is going on.

With https://github.com/NVIDIA/TransformerEngine/pull/2018#discussion_r2248634803, we can remove this entire comment.

transformer_engine/pytorch/graph.py
```diff
@@ -513,13 +514,19 @@ def hook_fn(module, inputs, outputs):  # pylint: disable=unused-argument
...
+                            to_deallocate_idx = previous_chunk_last_callable_bwd_idx
+                            previous_chunk_last_callable_bwd_idx = per_callable_bwd_idx
+                        else:
+                            # Weak ref last layer's buffer inside the same chunk.
+                            to_deallocate_idx = previous_per_callable_bwd_idx
+                            previous_per_callable_bwd_idx = per_callable_bwd_idx
+                        if to_deallocate_idx is not None:
+                            per_callable_static_grad_inputs[to_deallocate_idx] = make_weak_ref(
+                                per_callable_static_grad_inputs[to_deallocate_idx]
                             )
```
timmoon10: Won't this miss the weak ref for `l_no == 1`? How about:
```suggestion
                        
                        # Weak ref the static grad inputs of the previous backward pass within the chunk.
                        if previous_per_callable_bwd_idx is not None:
                            idx = previous_per_callable_bwd_idx 
                            per_callable_static_grad_inputs[idx] = make_weak_ref(
                                per_callable_static_grad_inputs[idx]
                            )
                        previous_per_callable_bwd_idx = per_callable_bwd_idx
                        
                        # Weak ref the static grad inputs of the previous chunk's backward pass
                        # Note: After a chunk's backward pass, we assume Mcore will send the
                        # grad input to another pipeline parallel rank and that the
                        # communication is finished before the end of the next chunk's backward
                        # pass.
                        if l_no == 0:
                            if previous_chunk_last_callable_bwd_idx is not None:
                                idx = previous_chunk_last_callable_bwd_idx
                                per_callable_static_grad_inputs[idx] = make_weak_ref(
                                    per_callable_static_grad_inputs[idx]
                                )
                            previous_chunk_last_callable_bwd_idx = per_callable_bwd_idx
```

buptzyb: oh! yes, l1 would never be deallocated in my way. Fixed, thanks!

transformer_engine/pytorch/graph.py

buptzyb:
Hi @timmoon10 , added a new fix to the static_input reuse part. Please see the PR description. Thanks!

timmoon10 (approved):
LGTM, pending CI

PR #2016, opened 2025-07-31
[JAX] Fix for contracting dim specs in GEMM custom calls

phu0ngng:
# Description
In Data Parallism, we have 
```
Wgrad =  input[B{X}, S, D]^T * dY [B{X}, S, H]
``` 
, both the lhs and rhs are sharded in the X mesh axis. 

The [current `_decompose_operand_specs`](https://github.com/phu0ngng/TransformerEngine/blob/main/transformer_engine/jax/cpp_extensions/gemm.py#L525-L526) does not infer the contracting spec correctly when the contracting specs are the batch dimension as it excludes the contracting dimension if the dimension is already in `bdims`. 

This results `None` for both `lhs_cspec` and `rhs_cspec`, leading to `reduce_output=False`, followed by overwriting `lhs_spec` and `rhs_specs`. This removes the `data` axis in the batch dimension for both lhs and rhs. As a result, `AllGather`-s are triggers to gather both the lhs and rhs before the GEMM.

This PR fixes the `_decompose_operand_specs`. The fix was verified in MaxText E2E runs with `tensor_parallelsim`, `tensor_sequence_parallelism`, and `data_parallelism`.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2023, opened 2025-08-01
[JAX] Sharding specs for TE GEMM custom call operands

phu0ngng:
# Description

This PR addresses incorrect sharding specifications in the current TE GEMM custom call implementation, which have previously led to unexpected collective operations—particularly under Data Parallelism (as noted in GH-2016)—and other related issues.

The PR introduces a revised approach for determining sharding specifications, ensuring proper sharding propagation and the correct triggering of collective operations. The updated logic proceeds as follows (note: `AllGather` operations are initiated when a non-None sharding axis becomes None):

1. AllReduce Triggering: An AllReduce is initiated when there is a shared non-None mesh axis within the contracting dimension of both the RHS and LHS.

Upon triggering an AllReduce:

- Sharding specifications for the remaining non-reducing contracting dimensions are set to None.

- Sharding specs for the RHS's non-batch, non-contracting dimension are also reset to None. This prevents the scenario in which the `dgrad` sharding under `FSDP` would have the `'fsdp'` mesh on two axes, which is not allowed in JAX.

2. No AllReduce Case: If no AllReduce is required, the sharding specs of the contracting dimensions for both LHS and RHS are set to None.

3. Sequence Axis Gathering: The sequence axis is always gathered when entering a GEMM, achieved by setting the non-batch, non-contracting dimension of the LHS to None.

Additionally, if the GEMM output should be sharded using a `sharding_constraint` following an AllReduce, XLA is able to substitute this pattern with a `ReduceScatter`. As a result, all previous ReduceScatter-related logic within GEMM has been removed, as it is now unnecessary.

Work on cleaning up `sequence_parallel_output` and `batch_dims` is planned for subsequent PRs.

With this PR, we achieved substantial performance improvements for `MXFP8BlockScaling` with different parallelism schemes, as follows:
image

Note: `DP` = Data Parallelism, `TP` = MaxText Tensor Parallelism = Megatron Tensor Parallelism + Sharding in the Hidden dimension for remaining unsharded activations. `TP-SP` = Megatron Tensor Parallelism + Sequence Parallelism.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -536,111 +520,70 @@ def _parse_operand_output_specs(
...
-        assert len(lhs_lspec_not_none) <= 1 and len(rhs_lspec_not_none) <= 1, (
-            "cuBLAS GEMM operands can have only one sharded non-batched leading dimension: "
-            f"{lhs_specs} @ idx {lhs_ldims} x {rhs_specs} @ idx {rhs_ldims}."
+            (lhs_cdims, rhs_cdims),
         )
-        assert len(lhs_cspec_not_none) <= 1 and len(rhs_cspec_not_none) <= 1, (
-            "cuBLAS GEMM operands can have only one sharded non-batched contracting dimension: "
-            f"{lhs_specs} @ idx {lhs_cdims} x {rhs_specs} @ idx {rhs_cdims}."
+        lhs_non_cspecs, lhs_cspecs, rhs_non_cspecs, rhs_cspecs = map(
+            lambda specs, cdims: tuple(specs[i] for i in cdims),
```
jberchtold-nvidia: nit self-resolve: should `cdims` in `lamda specs, cdims:` be renamed to just `dims` since it covers both contracting and non-contracting dims depending on input

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -536,111 +520,70 @@ def _parse_operand_output_specs(
...
+            for r in rhs_cspecs:
+                if l != None and l == r:
+                    # This if condition is here only to suppress warning, no functional effects
+                    if len(lhs_cspecs) >= 1:
+                        assert (reduce_spec is None, "Multiple reduce dimension is detected!")
+                    reduce_spec = l
+
+        if reduce_spec is not None:
+            # Other non-reduce cdims (if exists) need to be unsharded
+            lhs_cspecs = tuple(s if s == reduce_spec else None for s in lhs_cspecs)
```
jberchtold-nvidia, line 547: It's valid to compare `s == reduce_spec` by spec/axis name instead of dim index because all axes in a spec are guaranteed to be unique (except `None`). In other words, this will never return more than one non-None value in the spec, right?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -536,111 +520,70 @@ def _parse_operand_output_specs(
...
+            rhs_cspecs = tuple(s if s == reduce_spec else None for s in rhs_cspecs)
+            # Non-batched non-contracting dims of RHS needs to be unsharded (i.e. FSDP)
+            rhs_non_cspecs = tuple(
+                None if i not in rhs_bdims else rhs_non_cspecs[i] for i in rhs_non_cdims
```
jberchtold-nvidia: Is this indexing correct?
- `i` is an index into the original rhs tensor dims `[0, ndim)`, not an index into the `rhs_non_cdims` list
- `rhs_bdims` contains indices w.r.t the original tensor, so `[0, ndim)`, which is correct
- `rhs_non_cspecs` contains specs for each dim in `rhs_non_cdims`. If we are indexing by `i`, this could not match up correctly, right?

For example:
- `ndim = 4`
- `rhs_bdims = (1,)`
- `rhs_non_cdims = (1, 2)`
- `rhs_non_cspecs = ('a', 'b')`

```
None if i not in rhs_bdims else rhs_non_cspecs[i] for i in rhs_non_cdims
``` 
For the first iteration of the loop `i = 1`, then `i` is in `rhs_bdims` so we will use `rhs_non_cspecs[i]` which will be `rhs_non_cspecs[1]` which will grab axis `b`, but we really should be using `a` here, right?

Should it be the following? Let me know if I missed something or if this case won't occur in practice
```
None if i not in rhs_bdims else spec for i, spec in zip(rhs_non_cdims, rhs_non_cspecs)
```

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -536,111 +520,70 @@ def _parse_operand_output_specs(
...
             # Otherwise, require contracting dims of both operands to be unsharded
-            lhs_specs = tuple(None if i in lhs_cdims else lhs_specs[i] for i in range(lhs_ndim))
-            rhs_specs = tuple(None if i in rhs_cdims else rhs_specs[i] for i in range(rhs_ndim))
+            lhs_cspecs = (None,) * len(lhs_cspecs)
+            rhs_cspecs = (None,) * len(rhs_cspecs)
+
+        # Non-batched non-contracting dims of LHS to be unsharded, i.e gather SP dim
+        # (This cause MaxText TP (= Megatron TP + activation_hidden sharding) gathering Y1 for dW2 = Y1^T * dY2 which is unexpected)
+        lhs_non_cspecs = tuple(
+            None if i not in lhs_bdims else lhs_non_cspecs[i] for i in lhs_non_cdims
```
jberchtold-nvidia: Same indexing question as above

transformer_engine/jax/cpp_extensions/gemm.py
phu0ngng: Yes, this works only because we know all non-None axes should be unique.

transformer_engine/jax/cpp_extensions/gemm.py
phu0ngng: Thanks for catching this issue. 
Agree that accessing `rhs_non_cspecs[i]` with `i in rhs_non_cdims` could be out of range when rhs contracting dimensions are for the hidden dimension. 
Your suggestion works.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -536,111 +520,70 @@ def _parse_operand_output_specs(
...
-        # Set output scatter dim to the tensor-parallel spec
-        if sequence_parallel_output:
-            out_specs[scatter_dim] = reduce_spec
-
         return (
             (lhs_specs, rhs_specs, bias_specs, gelu_specs),
             (out_specs, bias_specs, gelu_specs),
             reduce_spec,
-            scatter_dim,
+            0,
```
denera, line 589: Since all the scatter logic in the op is being replaced with a sharding constraint outside the op, I think we should return `None` instead of an integer `0` here. This will help expose anything that might still be left behind by intentionally failing any integer-dependent scatter dim logic.

transformer_engine/jax/cpp_extensions/gemm.py
denera: I would argue that we should completely remove all the batched dim logic from the custom op in this PR. 

I originally added this to the primitive under the assumption that we need to *avoid* reducing in the data-parallel axis, and therefore needed a way to distinguish batched contracting dimensions from non-batched ones. That was a false assumption though, and we can treat the batched dims like any other contracting dim under the following two "rules":
- There can only be one contracting dimension with matching not-None sharding between LHS and RHS operands.
- Any other contracting dimension sharded differently between LHS and RHS operands should be all-gathered.

This approach is already implemented anyway in the nested `reduce_spec` double-loop, so I would recommend that we just strip out batched dim stuff everywhere else to match.

Tangentially related to this, I also think we should explicitly disallow RHS operands with more than one sharded non-contracting dimension. This would require BatchedGEMM to handle correctly, which we do not do here. We never attempt this particular GEMM case in any TE/JAX module, but we cannot necessarily guarantee that a user will not try to invoke it, so we should catch that unsupported use case and raise an error.

transformer_engine/jax/cpp_extensions/gemm.py
phu0ngng: Hi, we don't use `scatter_dim` anywhere in the code, so I think it should be fine for now. 

I'm working on a follow-up PR that cleans all the `sequence_parallel_output` and `sequence_dim` logics, so `scatter_dim` will be removed very soon.

jberchtold-nvidia (approved):
LGTM!

transformer_engine/jax/cpp_extensions/gemm.py
phu0ngng: Thanks for the suggestions. Hopefully, we could address them in the follow-up PR. 
This PR will only address urgent functionality and performance issues so that it gets merged asap.
Let me know if there are any cases this fix does not function nor perform as expected.

denera (approved):
LGTM!

PR #2027, opened 2025-08-04
[PyTorch] Register weight and bias params in linear op

timmoon10:
# Description

The [`Linear` op](https://github.com/NVIDIA/TransformerEngine/blob/06947e87b5511f8ad69ccd00286de9227f0fad24/transformer_engine/pytorch/ops/linear.py#L23) is a fused op with an interface similar to [`torch.nn.Linear`](https://docs.pytorch.org/docs/stable/generated/torch.nn.Linear.html). However, the param setters (for `weight` and `bias`) are buggy due to a bad interaction with [`torch.nn.Module.__setattr__`](https://github.com/pytorch/pytorch/blob/50eac811a68e63e96ad56c11c983bfe298a0bb8a/torch/nn/modules/module.py#L1968).

<details>

I would like to be able to replace the linear weight with something like:
```python
op = te.Linear(...)
op.weight = torch.nn.Parameter(...)
```

However, [when `torch.nn.Module.__setattr__` receives a `torch.nn.Parameter`](https://github.com/pytorch/pytorch/blob/50eac811a68e63e96ad56c11c983bfe298a0bb8a/torch/nn/modules/module.py#L1989), it calls [`torch.nn.Module.register_parameter`](https://github.com/pytorch/pytorch/blob/50eac811a68e63e96ad56c11c983bfe298a0bb8a/torch/nn/modules/module.py#L589).  This failed [here](https://github.com/pytorch/pytorch/blob/50eac811a68e63e96ad56c11c983bfe298a0bb8a/torch/nn/modules/module.py#L616) because the property names (`"weight"` and `"bias"`) were already taken by class properties.

</details>

This PR fixes the issue by changing the param logic in `Linear` from using class properties to directly registering params.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Register weight and bias params in linear op

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/ops/linear.py
+++ b/transformer_engine/pytorch/ops/linear.py
@@ -154,18 +154,16 @@ class Linear(FusedOperation):
         Also updates the basic operation that owns the parameter.
 
         """
+        if name == "bias" and self._bias_idx is None and param is not None:
+            raise ValueError(
+                "Attempted to set bias parameter in Linear operation "
+                "that does not have bias enabled"
+            )
         super().register_parameter(name, param)
         if name == "weight":
             self.basic_ops[self._linear_idx].weight = param
-        elif name == "bias":
-            if self._bias_idx is None:
-                if param is not None:
-                    raise ValueError(
-                        "Attempted to set bias parameter in Linear operation "
-                        "that does not have bias enabled"
-                    )
-            else:
-                self.basic_ops[self._bias_idx].bias = param
+        elif name == "bias" and self._bias_idx is not None:
+            self.basic_ops[self._bias_idx].bias = param
 
     def state_dict(self, *, prefix: str = "", **kwargs) -> dict[str, Any]:
         """Save state"""


transformer_engine/pytorch/ops/linear.py
```diff
@@ -111,60 +113,100 @@ def __init__(
...
+            self.basic_ops[self._linear_idx].weight = param
+        elif name == "bias":
+            if self._bias_idx is None:
+                if param is not None:
+                    raise ValueError(
+                        "Attempted to set bias parameter in Linear operation "
+                        "that does not have bias enabled"
+                    )
+            else:
+                self.basic_ops[self._bias_idx].bias = param
```
janekb04: Currently, this does not uphold the strong exception guarantee - if `bias` (which `is not None`) is set on `self` when `self._bias_idx is None`, the superclass will be notified with `super().register_parameter(name, param)`, however a `ValueError` will be raised later, leaving `self` in an inconsistent state. The call to `super().register_parameter` should be placed after the check.

Also, this is not consistent in checking all cases, as it does not handle the cases in which `self._bias_idx is not None`, but `bias is None`, or in which `weight is None`, which are not valid either. Currently, this invalid usage would be detected only in `fuser_forward`, while the `bias is not None` case would be caught immediately.

timmoon10: Moving the check before `super().register_parameter` makes sense.

Setting a param to `None` in `register_parameter` is valid: https://github.com/pytorch/pytorch/blob/8d1cf529229dce7cd5ea04abb0faac83b87ca6d1/torch/nn/modules/module.py#L618-L619
It could easily lead to problems later on, but that's a usage bug beyond the scope of this function.

PR #2028, opened 2025-08-05
Add backward RMSNorm+Add fusion

janekb04:
# Description

Fuse `MakeExtraOutput` with `RMSNorm` in the backward pass. Works through adding `nvte_rmsnorm_bwd_add`.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- (7ef7b578ebc4ce4e0edae218fdaa3a67dabb268a) Add `nvte_rmsnorm_bwd_add` as well as PyTorch binding, test, and update kernel to support the fusion
- (507471894d028cd7e253f1af309438d5ecd1c69b) Add `BackwardAddRMSNorm` fused operation, including test

# Checklist:

### DIFF
--- a/tests/cpp/operator/test_normalization.cu
+++ b/tests/cpp/operator/test_normalization.cu
@@ -27,10 +27,19 @@ namespace {
 
 template <typename InputType, typename OutputType>
 void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
-                 NormType norm_type, bool use_cudnn, const bool zero_centered_gamma_in_weight_dtype) {
+                 const NormType norm_type, const bool use_cudnn,
+                 const bool zero_centered_gamma_in_weight_dtype, const bool fused_bwd_add) {
   if (sizeof(InputType) < sizeof(OutputType)) {
     GTEST_SKIP() << "LN kernel does not support OutputType > InputType";
-    return;
+  }
+
+  if (norm_type == LayerNorm && fused_bwd_add) {
+    GTEST_SKIP() << "Fused LN backward+add not currently supported";
+  }
+
+  if (fused_bwd_add && zero_centered_gamma_in_weight_dtype) {
+    GTEST_SKIP() << "zero_centered_gamma_in_weight_dtype not currently supported "
+                 << "in fused norm backward+add";
   }
 
   if (getDeviceComputeCapability() < hopperComputeCapability && use_cudnn) {
@@ -45,7 +54,6 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
   if ((itype == DType::kBFloat16 && otype == DType::kFloat16) ||
       (itype == DType::kFloat16 && otype == DType::kBFloat16)) {
     GTEST_SKIP() << "LN kernel does not support mixing Float16 and BFloat16";
-    return;
   }
 
   Tensor input("input", std::vector<size_t>{ N, H }, itype);
@@ -55,6 +63,7 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
   Tensor mu("mu", std::vector<size_t>{ N }, DType::kFloat32);
   Tensor rsigma("rsigma", std::vector<size_t>{ N }, DType::kFloat32);
   Tensor dz("dz", std::vector<size_t>{ N, H }, wtype);
+  Tensor bwd_add("bwd_add", std::vector<size_t>{ N, H }, wtype);
   Tensor dx("dx", std::vector<size_t>{ N, H }, itype);
   Tensor dgamma("dgamma", std::vector<size_t>{ H }, wtype);
   Tensor dbeta("dbeta", std::vector<size_t>{ H }, wtype);
@@ -65,6 +74,11 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
   fillUniform(&beta);
   setRandomScale(&z);
   fillUniform(&dz);
+  if (fused_bwd_add) {
+    fillUniform(&bwd_add);
+  } else {
+    fillCase<WeightType>(&bwd_add, zeros);
+  }
 
   std::unique_ptr<OutputType[]> ref_output = std::make_unique<OutputType[]>(N * H);
   std::unique_ptr<float[]> ref_mu = std::make_unique<float[]>(N);
@@ -84,7 +98,6 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
   if (use_cudnn){
     nvte_enable_cudnn_norm_fwd(true);
     nvte_enable_cudnn_norm_bwd(true);
-
 
     // Zero-centered gamma in weight dtype only supported by CuDNN backend currently
     if (zero_centered_gamma_in_weight_dtype) {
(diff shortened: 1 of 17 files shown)


transformer_engine/common/normalization/rmsnorm/rmsnorm_bwd_semi_cuda_kernel.cu
```diff
@@ -17,7 +17,8 @@ void launch_tuned_(LaunchParams<BackwardKernelParams> &launch_params,
                    const bool configure_params) {  // NOLINT(*)
   using Kernel_traits = Kernel_traits<weight_t, input_t, output_t, compute_t, index_t, HIDDEN_SIZE,
                                       CTAS_PER_ROW, WARPS_M, WARPS_N, BYTES_PER_LDG_MAIN>;
-  auto kernel = &rmsnorm_bwd_tuned_kernel<Kernel_traits>;
+  auto kernel = launch_params.params.add ? &rmsnorm_bwd_tuned_kernel<Kernel_traits, true>
+                                         : &rmsnorm_bwd_tuned_kernel<Kernel_traits, false>;
```
timmoon10: We are doubling the number of kernel compilations for RMSNorm. This is fine for now, but it makes an NVRTC impl more compelling.

transformer_engine/common/normalization/rmsnorm/rmsnorm_bwd_kernels.cuh
```diff
@@ -111,6 +118,16 @@ __global__ __launch_bounds__(Ktraits::THREADS_PER_CTA) void rmsnorm_bwd_tuned_ke
       }
     }
 
+    maybe_t<Ovec[LDGS], FusedAdd> add;
```
timmoon10: Future optimization: Load `add` directly into `dx` to avoid extra register usage. However, this would require that `Ivec` and `Ovec` are identical types.

transformer_engine/common/include/transformer_engine/normalization.h
```diff
@@ -24,7 +24,7 @@ extern "C" {
  * y = \frac{x - E[x]}{\sqrt{Var[x] + \varepsilon}} \gamma + \beta
  * @f]
  *
- * Calling this function with workspace set to empty tensor will not perform the operation,
+ * Calling this function with workspace set to the empty tensor will not perform the operation,
```
timmoon10: Pedantic:
```suggestion
 * Calling this function with workspace set to an empty tensor will not perform the operation,
```

tests/pytorch/test_fusible_ops.py
```diff
@@ -2100,6 +2101,106 @@ def test_backward_activation_bias(
...
+        self,
+        *,
+        weight_shape: Iterable[int],
+        in_shape: Iterable[int],
+        dtype: torch.dtype,
+        device: torch.device = "cuda",
+        eps: float = 0.3,
+        zero_centered_gamma: bool,
+        quantization: Optional[str],
+    ) -> None:
```
timmoon10: No need to test the quantized case right now since we only support the unquantized kernel:
```suggestion
    @pytest.mark.parametrize("weight_shape", ((19,), (64,)))
    @pytest.mark.parametrize("in_shape", ((-1,), (6, 16, -1)))
    @pytest.mark.parametrize("dtype", _dtypes)
    @pytest.mark.parametrize("zero_centered_gamma", (False, True))
    def test_backward_add_rmsnorm(
        self,
        *,
        weight_shape: Iterable[int],
        in_shape: Iterable[int],
        dtype: torch.dtype,
        device: torch.device = "cuda",
        eps: float = 0.3,
        zero_centered_gamma: bool,
        quantization: Optional[str] = None,
    ) -> None:
```

timmoon10 (commented):
Overall LGTM, with some nits and future optimizations.

tests/pytorch/test_fusible_ops.py
janekb04: Applied the change.

tests/cpp/operator/test_normalization.cu
```diff
@@ -65,6 +69,11 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
   fillUniform(&beta);
   setRandomScale(&z);
   fillUniform(&dz);
+  if (fused_bwd_add) {
+    fillUniform(&bwd_add);
+  } else {
+    fillCase(&bwd_add, zeros);
```
timmoon10: I'm seeing compilation errors:
```suggestion
    fillCase<WeightType>(&bwd_add, zeros);
```

transformer_engine/common/normalization/rmsnorm/rmsnorm_bwd_kernels.cuh
```diff
@@ -118,20 +118,27 @@ __global__ __launch_bounds__(Ktraits::THREADS_PER_CTA) void rmsnorm_bwd_tuned_ke
       }
     }
 
-    maybe_t<Ovec[LDGS], FusedAdd> add;
+    union {
+      using add_t = maybe_t<Ovec, FusedAdd>;
+      using dx_t = Ivec;
+      struct {
+        char _padding[sizeof(dx_t) > sizeof(add_t) ? sizeof(dx_t) - sizeof(add_t) : 0];
```
timmoon10: I presume this is because `union`s are big endian while arrays are little endian? If the `add` dtype is bigger than `dx`, do we need similar padding for `dx` to ensure the arrays are aligned?

Alternatively, for now could we just force `add` and `dx` to use the same dtype?

transformer_engine/common/normalization/rmsnorm/rmsnorm_bwd_kernels.cuh
```diff
@@ -141,12 +148,12 @@ __global__ __launch_bounds__(Ktraits::THREADS_PER_CTA) void rmsnorm_bwd_tuned_ke
         compute_t y_tmp = y[it * NUM_ELTS + jt];
         compute_t dx_tmp = rs_r * (dy_tmp - (mdyy_local * y_tmp));
         if constexpr (FusedAdd) {
-          compute_t add_tmp = add[it].data.elt[jt];
+          compute_t add_tmp = temp[it].add.data.elt[jt];
           dx_tmp += add_tmp;
         }
-        dx[it].data.elt[jt] = dx_tmp;
+        temp[it].dx.data.elt[jt] = dx_tmp;
```
timmoon10: If the `add` dtype is smaller than the `dx` dtype, couldn't this result in overwriting `add` before it is used? If the `add` dtype is larger than the `dx` dtype, this relies on loop dependencies to avoid corrupting `temp`, which might make it harder for the compiler to do its magic.

transformer_engine/common/normalization/rmsnorm/rmsnorm_bwd_kernels.cuh
janekb04: The padding is added for `add` so that writing `dx` for the current `it` does not overwrite the `add` read by the next `it`:
image

transformer_engine/common/normalization/rmsnorm/rmsnorm_bwd_kernels.cuh
janekb04: I think that actually `dx` cannot have a padding as then it would overwrite `add`.

This indeed forces the loop to be executed sequentially, but I thought that unrolling it would cause it to be executed sequentially anyway, except that it fills the time between fetching `add` from memory and using it with computation from other iterations.

PR #2038, opened 2025-08-06
Ubnext

nv-akorzh:
# Description

Added UBnext fast Allreduce kernels into linear and layernorm_linear layer.
Falls under symmetric_ar_type with new type being 'ub_custom'

# Details
Added UC MC and fast sync - low latency (lamport) allreduce kernel.

Added symmetric allocator which uses pytorch symmetric to allocate pool and suballocate from it.

As pytorch symmetric doesnt support MNNVL yet there is a fallback to use legacy UB code by creating a 11th CommOverlap object. Enabled with env NVTE_USE_UB_FOR_UBNEXT ( requires user to initialize ub by calling initialize_ub)
NVTE_UB_MAXBATCH (default 128) can increase batch size which would have enough memory for fastest kernel. If memory cant be allocated there is gradual fallback: first to UBmain in-place kernel if input could be allocated and output couldnt, and to pytorch symmetric if input couldnt be allocated.

NVTE_UB_SYMM_POOL_SIZE env overrides pool size to given number of megabytes.


nv-akorzh:
closing since will be pushing next more advanced one shortly which would include this one

PR #2039, opened 2025-08-07
[PyTorch] Quantizer as API + CustomRecipe

negvet:
# Description

Expose quantizer as an API.
Main objective - let users build custom quantizers.

Make `Quantizer` general so that it can be used outside of TE.

Usage example:
```
 from transformer_engine.pytorch import Quantizer
 
 class MyCustomQuantizer(Quantizer):
     def quantize_impl(self, tensor, **kwargs):
         # Custom quantization logic, e.g. python-based or pure silicon
         pass
```

Custom quantizers can be used:

- Externally
- In TE, with some TE-required implementations (such as `update_quantized()`)

## Type of change

- [] New feature (non-breaking change which adds functionality)
- [ ] Infra/Build change
- [X] Code refactoring

## Changes

- Introduced `QuantizerBase`
- Exposed `QuantizerBase` as an API

# Checklist:

- [] I have made corresponding changes to the documentation


negvet:
About inconsistencies of `columnwise_usage` and `rowwise_usage`.

Naming: parameters `rowwise` and `columnwise`, but instance attributes: `rowwise_usage` and `columnwise_usage`. Plus, the meaning of "usage" is unclear.
Semantic ambiguity: Currently, rowwise_usage and columnwise_usage are independent boolean flags, but they are different aspects of the tensor layout (semantically close).

I would propose to reconsider this design.

Renaming would already improve the situation.
Another option is to move towards JAX Implementation (enum-based):
```
class QuantizeLayout(Enum):
    ROWWISE = "rowwise"
    COLUMNWISE = "columnwise"  
    ROWWISE_COLWISE = "both"
```
Although this introduces a new class for users to learn.

negvet:
@ptrendx I propose to keep `update_quantized` in the concrete `Quantizer` class, but not in `QuantizerBase`.

This is TE-specific optimization (weight workspace caching + cuda graph support etc.)
Custom quantizers for other use cases might not need in-place updates:
- Research quantizers might only need one-time quantization
- Inference-only quantizers might not need parameter updates
- etc.

timmoon10:
- I agree "usages" is not the best term for the concept we're describing, but we use it consistently throughout the codebase. If we find something better, we should properly document it and change it everywhere.
- There's a semantic difference between "row-wise/column-wise _usage_" and "row-wise/column-wise _data_". "Row-wise/column-wise usages" indicates intent, and each usage is completely orthogonal since you may or may not use the same tensor for multiple operations. "Row-wise/column-wise data" is not orthogonal and is highly recipe-dependent, since some buffers can be used for multiple usages, e.g. FP8 data on Blackwell.
- We should keep in mind that we might add more usages in the future. We are seriously considering adding usages for communication. For example, the FP8 wgrad GEMM currently does an all-gather followed by transpose. Instead of quantizing with row-wise usage, it would be more natural to quantize with all-gather-column-wise usage. We have similar considerations if we want to support MXFP8 with pre-swizzled scales.
- The enum approach is not scalable since its size will grow with 2^n.
- Who knows what GEMM assumptions future architectures will have? Row-wise and column-wise currently rely on the fact that Hopper/Blackwell Tensor Cores use the same data format for A and B in a TN GEMM. Could we require 4 usages in the future (A, A^T, B, B^T)? What if we need to support convolutions?

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -154,6 +154,45 @@ class Quantizer(abc.ABC):
...
+        columnwise: Optional[bool] = None,
+    ) -> None:
+        """Set how the quantized tensor is expected to be used"""
+        if rowwise is not None:
+            self.rowwise_usage = rowwise
+        if columnwise is not None:
+            self.columnwise_usage = columnwise
+
+
+class Quantizer(QuantizerBase):
```
timmoon10: I don't think the distinction between `Quantizer` and `QuantizerBase` is logical. This PR is trying to distinguish between quantizers that call `tex.quantize` and those that call a Python impl, but that's a quantizer-specific implementation detail. `QuantizerBase` is also haphazardly removing parts of the quantizer API, like the ability to construct empty tensors, PyTorch autograd support, etc.

I think the right design is not to add an unnecessary `QuantizerBase` class, but to decouple `Quantizer` from `tex.quantize`. We can add an abstract `quantize_impl` function that is called within `quantize`. The existing quantizers should call `tex.quantize`, but future quantizers could use a pure Python impl.

timmoon10 (commented):
Exposing a pure Python interface will be quite nice. Ideally we'd design the quantizers in such a way that a C++ or Python impl is an implementation detail, and we don't need any special logic in the modules.

One question is how to handle the [C++ quantizer infrastructure](https://github.com/NVIDIA/TransformerEngine/blob/cae1c436027cc028ee49c83463141ba67f0adca0/transformer_engine/pytorch/csrc/common.h#L95), e.g. in `tex.quantize`. Options:
- Only use the C++ quantizer as a perf optimization. We'll need to add checks to avoid passing a pure-Python quantizer into `tex` functions (e.g. for norms or activations).
- Add a C++ quantizer that calls a Python function (or modify the C++ quantizer base class). The C++ quantizer has to deal with both Python tensor classes (as pybind objects) and `NVTETensor`, and handling `NVTETensor`s from Python will be challenging.

I think the pure Python approach in this PR is more straightforward.

timmoon10:
> @ptrendx I propose to keep `update_quantized` in the concrete `Quantizer` class, but not in `QuantizerBase`.
> 
> This is TE-specific optimization (weight workspace caching + cuda graph support etc.) Custom quantizers for other use cases might not need in-place updates:
> 
> * Research quantizers might only need one-time quantization
> * Inference-only quantizers might not need parameter updates
> * etc.

I don't think this is a good reason to change the API. If we want to cut corners by not implementing things (fair enough for experimentation), we can throw `NotImplementedError`s. The problem with this approach is that we'll need to do `isinstance(quantizer, Quantizer)` if we want to use APIs that are not in `QuantizerBase`, which is leaking implementation details out of the class. If it's a `QuantizerBase`, we'll probably throw an exception anyways.

negvet:
I agree, let's keep Quantizer as the base class. Instead, implementations will go one level down to the concrete quantizers.
Please see https://github.com/NVIDIA/TransformerEngine/pull/2039/commits/661697df2f18a49246204a871efbb2ae7f0923fe

negvet:
Let me give a bit more detail about what I am trying to accomplish here.

Main goal - enable custom quantization with user-defined `quantize()` logic, whether in Python, C++, or any other language; this is an implementation detail. The focus is on exposing `Quantizer` and `QuantizedTensorBase`: specifically, `CustomQuantizer(Quantizer)` and `CustomQuantizedTensor(QuantizedTensorBase)`. 
Regarding `QuantizedTensor`, I didn’t end up using it while building basic custom recipes, since I only needed tensor containers like `QuantizedTensorBase`. It might be a good exercise to try using it. However, if we plan to expose specific quantizers like `Float8Quantizer` (which would also require exposing `Float8Tensor`), then we need to expose `QuantizedTensor` as well.

The main idea is that these custom quantizers and quantized tensors can be developed and used both inside and outside of TE. The only initial requirement is a `quantize()` implementation (to keep things simple at first; this might be enough, with the rest being optional). 
Developing within TE would require additional implementations, such as `update_quantized()`, among others. 
This adds extra work for users who want to develop or run code using TE. Users will learn about this TE contract when attempting to integrate a custom quantizer into TE and encountering a `Not Implemented` error.

negvet:
[Support CustomRecipe and CustomRecipeState](https://github.com/NVIDIA/TransformerEngine/pull/2039/commits/7b724a7921d34e23dfc253452b52c01ba8bdd5b5) enables custom quantizers in TE without module or other code modifications. Please let me know what you think about it.

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -187,7 +186,12 @@ def update_quantized(
         noop_flag: Optional[torch.Tensor] = None,
     ) -> QuantizedTensor:
         """Quantize tensor in-place"""
+        raise NotImplementedError(
+            f"{self.__class__.__name__} class does not implement update_quantized function, "
+            " required for weight workspace caching (in-place update of quantized tensor)"
+        )
```
timmoon10, line 214: This is a general function and it's too restrictive to assume it's being used for the weight workspace.
```suggestion
        raise NotImplementedError(
            f"{self.__class__.__name__} class does not implement update_quantized"
        )
```

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -222,15 +220,21 @@ def make_empty(
...
     def calibrate(self, tensor: torch.Tensor) -> None:
         """Calibrate quantizer state
 
         Updates quantization state as if quantizing a tensor, but
         without actually performing the quantization.
 
         """
+        raise NotImplementedError(
+            f"{self.__class__.__name__} class does not implement calibrate function"
+        )
```
timmoon10: Calibration is a no-op for stateless quantizers, so I think that's a reasonable default.
```suggestion
```

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -196,11 +200,6 @@ def quantize(
         dtype: Optional[torch.dtype] = None,  # pylint: disable=unused-argument # used by override
     ) -> QuantizedTensor:
         """Quantize tensor"""
-        if out is not None:
-            return self.update_quantized(tensor, out)
-        if (not self.internal) and torch.is_grad_enabled():
-            return _QuantizeFunc.apply(tensor, self)
-        return _QuantizeFunc.forward(None, tensor, self)
```
timmoon10, line 203: We should keep the existing implementation of `quantize` and internally call a separate `quantize_impl` function. That way we disentangle the out-of-place quantization logic from in-place quantization and autograd. This might look something like:

```python

def quantize(self, tensor, *, out, dtype):
    if out is not None:
        return self.update_quantized(tensor, out)
    if (not self.internal) and torch.is_grad_enabled():
        return _QuantizeFunc.apply(tensor, self.quantize_impl)
    return _QuantizeFunc.forward(None, tensor, self.quantize_impl)

def quantize_impl(self, tensor):  # Implementation in Float8Tensor, MXFP8Tensor, etc
    return tex.quantize(tensor, self)

...

class _QuantizeFunc(torch.autograd.Function):
    """Quantize tensor"""

    @staticmethod
    def forward(
        _ctx: Optional[torch.autograd.function.FunctionCtx],  # unused
        tensor: torch.Tensor,
        quantize_impl: Callable,
    ) -> QuantizedTensor:
        return quantize_impl(tensor)

    @staticmethod
    def backward(
        _ctx: torch.autograd.function.FunctionCtx,  # unused
        grad: torch.Tensor,
    ) -> Tuple[Optional[torch.Tensor], ...]:
        # Assume that we want gradients in full precision
        return grad, None
```

transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -252,17 +256,42 @@ def copy(self) -> Quantizer:
...
+        raise NotImplementedError(
+            f"{self.__class__.__name__} class does not implement _get_compatible_recipe"
+        )
 
     def supports_only_rowwise_all_gather(self) -> bool:
         """Returns True if the quantizer supports only rowwise all-gather"""
-        return False
+        raise NotImplementedError(
+            f"{self.__class__.__name__} class does not implement supports_only_rowwise_all_gather"
+        )
```
timmoon10: This is intended for FP8-specific logic, so I think it makes more sense to return `False` by default:
```suggestion
        False
```

transformer_engine/pytorch/fp8.py
```diff
@@ -1070,3 +1074,63 @@ def make_quantizers(self) -> list:
...
+        if q is None:
+            return None
+        if hasattr(q, "copy"):
+            return q.copy()
+        return copy.copy(q)
+
+    def make_quantizers(self) -> list:
+        qparams = self.recipe.qparams
+        if self.mode == "forward":
+            if self.num_quantizers % 3 != 0:
```
timmoon10: This is not a safe assumption. `te.ops.Linear` has two forward quantizers (input, weight) and one backward quantizer (grad output): https://github.com/NVIDIA/TransformerEngine/blob/05d3b7b511ff142a2a0e6d7b46d56b81d1323461/transformer_engine/pytorch/ops/basic/basic_linear.py#L256-L261

This is more logical than what's being done in `te.Linear`, which has extra quantizers for the output and grad input. The usages are determined by the operation that _consumes_ the data, so if the quantizer lives with the operation that _generates_ the data it might produce the wrong usages. You could work around this by making `te.Linear` peek ahead and configure itself based on the next module, but at that point you might as well just grab the next module's quantizer directly (what `te.ops.Sequential` does).

transformer_engine/common/recipe/__init__.py
```diff
@@ -351,3 +356,35 @@ def __repr__(self) -> str:
             f"fp8_dpa={self.fp8_dpa}, "
             f"fp8_mha={self.fp8_mha}"
         )
+
+
+@dataclass()
+class QLinearParams:
```
timmoon10: We should make it more obvious that this is specific to `CustomRecipe`. The name makes me think it is similar to `MMParams` or `QParams`:
https://github.com/NVIDIA/TransformerEngine/blob/9c67da6e8ca42c2f7627cde4bbd87e9f0569c865/transformer_engine/common/recipe/__init__.py#L44-L62
Maybe:
```suggestion
class CustomRecipeLinearQuantizers:
```

transformer_engine/pytorch/fp8.py
```diff
@@ -1070,3 +1074,63 @@ def make_quantizers(self) -> list:
...
+                raise ValueError(
+                    "CustomRecipeState forward expects num_quantizers multiple of 3, got"
+                    f" {self.num_quantizers}"
+                )
+            out = []
+            for _ in range(self.num_quantizers // 3):
+                out.append(self._clone(qparams.input_quantizer))
+                out.append(self._clone(qparams.weight_quantizer))
+                out.append(self._clone(qparams.output_quantizer))
```
timmoon10: Using a prototype design pattern will work well when the quantizers are independent, but how should we handle the case where the quantizers have shared state (e.g. the input and weight quantizers must have access to the same RNG)?

ptrendx: I have a different issue - here we are doing shallow copy of the quantizers for every op. What would happen if somebody wanted to have an actual state in them, including torch.Tensors? They should actually be copied then. Maybe a solution would be to, rather than taking already made quantizers, to instead ask for a factory function of them?

transformer_engine/pytorch/tensor/quantized_tensor.py
ptrendx: There should be some indication though about what you are losing if you do not implement this function. Maybe that could be included in the documentation?

tests/pytorch/test_custom_recipe.py
```diff
@@ -0,0 +1,113 @@
...
+import transformer_engine_torch as tex
+from transformer_engine.common import recipe
+from transformer_engine.pytorch.fp8 import check_fp8_support, fp8_autocast
+from transformer_engine.pytorch import Linear
+from transformer_engine.pytorch.tensor.float8_tensor import (
+    Float8CurrentScalingQuantizer,
+)
+
+
+def test_custom_recipe_linear():
```
ptrendx: What about the other modules?

transformer_engine/pytorch/tensor/quantized_tensor.py
negvet: Removed from here, let's include in the documentation.

transformer_engine/pytorch/tensor/quantized_tensor.py
negvet: Fixed

transformer_engine/pytorch/tensor/quantized_tensor.py
negvet: Fixed

transformer_engine/pytorch/tensor/quantized_tensor.py
negvet: Fixed

transformer_engine/pytorch/fp8.py
negvet: What about defining a per-mode pool and cycle it to the requested length?

```
    def make_quantizers(self) -> list:
        qfactories = self.recipe.qfactories
        # Build a canonical, per-mode pool of factories and cycle it to match num_quantizers.
        if self.mode == "forward":
            pool = [
                qfactories.input_factory,
                qfactories.weight_factory,
                qfactories.output_factory,
            ]
        else:
            pool = [
                qfactories.grad_output_factory,
                qfactories.grad_input_factory,
            ]

        # Filter out None entries; allow partial factories and reuse as needed.
        pool = [factory for factory in pool if factory is not None]
        if len(pool) == 0: 
            raise ValueError(
                "CustomRecipe requires at least one factory for the requested mode; "
                f"got none for mode={self.mode}."
            )

        out = []
        for i in range(self.num_quantizers):
            factory = pool[i % len(pool)]
            out.append(factory())
        return out
```

`te.Linear` that internally asks for 3/2 continues to get 3/2.
`te.ops.Linear` that asks for 2/1 will get 2/1 without error.
If a user only supplies `input_quantizer`, and an op asks for both `input` and `weight`, the `input` quantizer is reused for the `weight` (although it can raise an exception to be safe).

transformer_engine/common/recipe/__init__.py
negvet: Renamed to `CustomQuantizerFactories`.

transformer_engine/pytorch/fp8.py
negvet: Factories can solve both of the above issues. So let's go this way.

Any internal state such as `torch.Tensor` is newly constructed by the factory.
Shared RNG for input and weight:
```
rng = torch.Generator(device="cuda").manual_seed(1234)

def make_input(args: FactoryArgs, ctx):
    return MyQuantizer(dtype=..., rng=rng)  # same rng shared

def make_weight(args: FactoryArgs, ctx):
    return MyQuantizer(dtype=..., rng=rng)  # same rng shared

def make_output(args: FactoryArgs, ctx):
    return MyQuantizer(dtype=..., rng=torch.Generator(device=args.device))
```

tests/pytorch/test_custom_recipe.py
negvet: Added other modules

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -109,8 +109,8 @@ def _get_act_func_supported_list(recipe: Optional[Recipe] = None):
             "srelu": (tex.srelu, tex.dsrelu, tex.dbias_dsrelu),
         }
     # no activation fusion written yet
-    # Per-tensor current scaling or fp8 blockwise scaling: []
-    if recipe.float8_current_scaling() or recipe.float8_block_scaling():
+    # Per-tensor current scaling, fp8 blockwise scaling, or custom factories: []
+    if recipe.float8_current_scaling() or recipe.float8_block_scaling() or recipe.custom():
```
timmoon10: This is not right since it requires that the quantizer is supported in the `tex` function. However, the problems in the activation func logic are bigger than this PR (activation func should be determined by the quantizer and not by the recipe, activation func logic leaks externally for the DSv3 recipe).

If we decide to fix this now, we should refactor so the code reflects functionality and hides implementation details:
```python
def _activation_forward(
    activation: str,
    input_: torch.Tensor,
    quantizer: Optional[Quantizer],
) -> Union[torch.Tensor, QuantizedTensorBase]:
    func = {
        "gelu": tex.gelu,
        "relu": tex.relu,
        ...
    }[activation]
    with_fused_impl = isinstance(quantizer, (Float8Quantizer, Float8CurrentScalingQuantizer, MXFP8Quantizer))
    out = func(input_, quantizer if with_fused_impl else None)
    if quantizer is not None and not with_fused_impl:
        out = quantizer(out)
    return out

...

act_out = _activation_forward(fc1_out, fc2_input_quantizer)

```

transformer_engine/pytorch/fp8.py
timmoon10: This won't generalize to `GroupedLinear`. `te.GroupedLinear` has `3*num_gemms` forward quantizers. When we implement it for `te.ops`, we'll do it right and have `2*num_gemms` forward quantizers.

I think we can expose options so the module can choose how to cycle. However, the default path should follow `te.ops` and the burden of enabling the non-standard path should be in `te.Linear`.

transformer_engine/pytorch/fp8.py
timmoon10: That's a good idea, this approach could be made very general. Imagine if the user could create a custom recipe with something like:
```python
def _make_recipe_state(recipe, mode, num_quantizers):
    recipe_state = CustomRecipeState(recipe=recipe, mode=mode, num_quantizers=num_quantizers)
    recipe_state.my_weird_shared_state = torch.fill(123)

def _make_quantizer(recipe_state, mode, num_quantizers, quantizer_idx):
    if mode == "linear_forward":
        if quantizer_idx == 0:
            # Input quantizer
            return MyQuantizer1(recipe_state.my_weird_shared_state)
        if quantizer_idx == 1:
            # Weight quantizer
            return MyQuantizer2()
        if quantizer_idx == 2:
            # Output quantizer
            return MyQuantizer3(recipe_state.my_weird_shared_state)
    if mode == "linear_backward":
        # Backward quantizers
        return MyQuantizer3(recipe_state.my_weird_shared_state)
    if mode == "attention_forward" or mode == "attention_backward":
        # Quantizers for attention intermediate tensors
        return MyQuantizer4()

recipe = CustomRecipe()
recipe.make_recipe_state = _make_recipe_state
recipe.make_quantizer = _make_quantizer
```
If we do this right, the user only needs to provide one quantizer factory function.

timmoon10 (commented):
LGTM, once we fix linter warnings and pass CI.

We should consider the custom recipe API experimental and downstream users should not expect it to be stable yet. The biggest issue is that we need quantizers for things other than linear layers, e.g. attention requires quantizers for intermediate tensors. Hard-coding linear-specific logic (`input_factory`, `weight_factory`) is not a good long-term solution. Generalizing the quantizers will require other changes outside the scope of this PR, but I envision the API becoming something like https://github.com/NVIDIA/TransformerEngine/pull/2039#discussion_r2274592844.

transformer_engine/pytorch/fp8.py
negvet: Why won't it generalize? 

The current make_quantizers is flexible for `te.Linear` (3/2), te.ops.Linear (2/1), and `GroupedLinear`.
For `te.GroupedLinear`, forward requests `3*num_gemms` and backward `2*num_gemms` quantizers; `make_quantizers` fills that by cycling the per-mode pool, which is effectively repeating [input, weight, output] (or [grad_output, grad_input]) `num_gemms` times.

There is a sanity test case for that as well, please take a look `test_custom_recipe_grouped_linear_sanity`.

negvet:
Regarding exposing concrete quantizers.
Those should be exposed along with corresponding quantized tensors. 
For custom quantization development, `*Base` quantized tensors are enough for most cases (I didn't use`QuantizedTensor`).
But `Float8TensorBase` and other `*Base` implementations are kind of internal implementation details. We then need to move those out from `_internal` in order to expose them.
Thus, we can expose: `Quantizer` + all `*Quantizer`, `QuantizedTensorBase` + all `*Base`, `QuantizedTensor` + all its sublcasses for those who want `torch.Tensor` sublclass features.
What do you think @timmoon10 @ksivaman @ptrendx ?

timmoon10 (dismissed):
LGTM

transformer_engine/pytorch/fp8.py
```diff
@@ -1112,34 +1112,15 @@ def __init__(
...
-            )
-
+        qfactory = self.recipe.qfactory
         out = []
         for i in range(self.num_quantizers):
-            factory = pool[i % len(pool)]
-            out.append(factory())
+            # TODO(negvet): Support other operations.
+            role = self.recipe.resolve_role("linear", self.mode, i)
+            quantizer = qfactory(role)
```
timmoon10: Separating out different roles is a nice way to handle different operations. Could we go even further by removing `resolve_role` and just passing everything into the factory function?

Pros:
- The factory functions have complete flexibility over what quantizers they produce. They can add new operations/roles without updating TE.
- A naive quantization scheme might just return the same quantizer for everything, in which case resolving the role is unnecessary.
- The role names are linear-specific. I'd expect convolution to also have an input, weights, and output.

Cons:
- The factory functions need to have some awareness of the quantizer order.
- Role names are more descriptive than indices.

transformer_engine/pytorch/fp8.py
timmoon10: When we add support for grouped linear in `te.ops`, we should not include the output quantizers. The forward quantizers should be `[input, weight, input, weight, ...]` and the backward quantizers will be `[grad_output, grad_output, ...]`.

transformer_engine/common/recipe/__init__.py
```diff
@@ -385,13 +369,51 @@ class CustomRecipe(Recipe):
...
+                elif index % 3 == 2:
+                    return "output"
+            elif mode == "backward":
+                if index % 2 == 0:
+                    return "grad_output"
+                elif index % 2 == 1:
+                    return "grad_input"
+        if operation == "attention":
+            raise NotImplementedError("Attention is not supported for now.")
+        return None
```
timmoon10: - The output and grad input quantizers don't logically belong with the linear module (https://github.com/NVIDIA/TransformerEngine/pull/2039#discussion_r2271511000), so a better name would be something like "linear_and_outputs"
- The roles are linear-specific, so names like "input" and "weight" are too general. 
```suggestion
        if operation == "linear":
            if mode == "forward":
                return ("linear_input", "linear_weight")[index % 2]
            elif mode == "backward":
                return "linear_grad_output"
        elif operation == "linear_and_outputs":
            if mode == "forward":
                return ("linear_input", "linear_weight", "linear_output")[index % 3]
            elif mode == "backward":
                return ("linear_grad_output", "linear_grad_input")[index % 2]
        raise ValueError("Failed to resolve mode for operation {operation}, mode {mode}, index {index}")
```

timmoon10 (dismissed):

transformer_engine/pytorch/fp8.py
negvet: Indeed, `resolve_role` can be removed from the `CustomRecipe` to keep API surface smaller.
By passing everything to `qfactory`, users can implement non-TE mappings or custom layouts without subclassing.

To resolve cons, I propose to resolve TE-specific roles internally and pass a role along with additional context to the factory. Users can rely on the role when they don’t want to think about ordering, but still have the knobs when needed.

Please take a look at the implementation.

transformer_engine/common/recipe/__init__.py
negvet: One way to simplify the resolver is to introduce `forward_roles/backward_roles`, provided by the module/op.

forward_roles=("input","weight","output") for `te.Linear`, ("input","weight") for `te.ops.Linear`.
backward_roles=("grad_output","grad_input") vs ("grad_output",).

Resolver becomes role = forward_roles[i % len(forward_roles)]  # for fwd
This scales to grouped GEMMs, 2/1, 3/2 etc.

Please take a look.

negvet: > The roles are linear-specific, so names like "input" and "weight" are too general.

I agree.
Linear case will cover the vast majority of use cases of CustomRecipe.
Given that most users will hack only linear quantization, why not keep its role simple (and aligned with internal TE naming)? 
For the other, more special cases (e.g. attention), we might introduce more specific roles.

negvet:
Are there any restrictions against renaming e.g. `XTensorBase` -> `XTensorStorage` or `XTensorData` (along with `QuantizedTensorBase` -> `QuantizedTensorStorage` and `transformer_engine.pytorch.tensor.base` -> `transformer_engine.pytorch.tensor.storage`)? 

That would be more clear because `XTensorBase` basically holds data attributes.

timmoon10:
Renaming to `XTensorBase` to `XTensorStorage` is not a bad idea. It clarifies a lot of the logic. Let's get this in first though.

transformer_engine/common/recipe/__init__.py
timmoon10: "linear_input" doesn't seem that different than "input". Even if the oververbosity turns out to be a mistake, it's better than the case where we are underverbose.

transformer_engine/pytorch/fp8.py
timmoon10: I think I have an even nicer long-term solution, although we don't need to implement it in this PR. Instead of resolving roles inside `CustomRecipeState`, we should change `RecipeState.make_quantizers` so that it can take in a list of roles. Each operation can then choose whatever order it likes and not be bound to a cyclic order.

For now, let's settle on an external API of just `qfactory(role)`. Since we're keeping the roles, the kwargs are not adding much value. The existing logic in `_resolve_te_role` is fine for now.

transformer_engine/pytorch/fp8.py
```diff
@@ -1084,3 +1105,61 @@ def make_quantizers(self) -> list:
...
+
+    def make_quantizers(self) -> list:
+        qfactory = self.recipe.qfactory
+        out = []
+        for i in range(self.num_quantizers):
+            # TODO(negvet): obtain forward_roles/backward_roles from module/op
+            # configuration (e.g., te.Linear uses (3/2), te.ops.Linear uses (2/1)).
+            # For now, hardcode the te.Linear layout
+            forward_roles = ("input", "weight", "output")
+            backward_roles = ("grad_output", "grad_input")
```
timmoon10: Why pass the forward and backward roles separately into `_resolve_te_role` and `qfactory`? We can fully figure out the roles beforehand:
```suggestion
            roles: List[str]
            if mode == "forward":
                roles = [("input", "weight", "output")[i % 3] for i in range(self.num_quantizers)]
            elif mode == "backward":
                roles = [("grad_output", "grad_input")[i % 2] for i in range(self.num_quantizers)]
            else:
                roles = ["unknown"] * self.num_quantizers
```
No reason for the forward quantizers to know about backward roles, and vice versa. This is one reason I don't like the kwargs in `qfactory`. In the future, we can generalize the quantizer infrastructure so that each operation can pass in its own list of roles.

transformer_engine/pytorch/fp8.py
negvet: This is roughly what I proposed as well, if you check TODO in make_quantizers() in https://github.com/NVIDIA/TransformerEngine/pull/2039/commits/7d6fef5c1ce028a9795f879c1dca1b4d5f12edda#diff-77a87478142dbcec64bbe6fc9f402be4c4c9aa21e375cca6d352545fa8e87166R1140

negvet: > let's settle on an external API of just qfactory(role)

Done

transformer_engine/pytorch/fp8.py
negvet: So if op will pass roles, then we might close this discussion.

negvet:
It's probably time to merge this; we can continue to work on it in the other PRs.

transformer_engine/pytorch/distributed.py
```diff
@@ -1705,8 +1707,8 @@ def _is_te_module(module):
 def prepare_te_modules_for_fsdp(fsdp_root: torch.nn.Module) -> None:
     """
     Inject FSDP process gorup references into FSDP-wrapped TE modules in an FSDP-wrapped root
-    module in order to scatter/gather the Fp8 weight copies at the same time FSDP scatters/gathers
-    its `FlatParameters`.
+    module in order to scatter their
+    activation tensors after each forward pass and gather them before the backward pass.
```
timmoon10: Is this comment correct? FSDP shouldn't have anything to do with activations, it should only affect weights.
```suggestion
    module in order to scatter/gather the Fp8 weight copies at the same time FSDP scatters/gathers
    its `FlatParameters`.
```

transformer_engine/pytorch/tensor/storage/mxfp8_tensor_storage.py
```diff
@@ -49,8 +49,8 @@ def backward(
         return grad, None
 
 
-class MXFP8TensorBase(QuantizedTensorBase):
-    """Mixin class that holds data attributes of MXFP8Tensor.
+class MXFP8TensorStorage(QuantizedTensorStorage):
+    """Mixin/storage class that holds data attributes of MXFP8Tensor.
```
timmoon10: Nit: mix-in classes and storage classes are different things, so "mixin/storage class" doesn't mean much to me.
```suggestion
    """Mixin class that holds data attributes of MXFP8Tensor.
```

transformer_engine/pytorch/tensor/storage/float8_tensor_storage.py
```diff
@@ -52,8 +52,8 @@ def backward(
         return grad, None
 
 
-class Float8TensorBase(QuantizedTensorBase):
-    """Mixin class that holds data attributes of Float8Tensor.
+class Float8TensorStorage(QuantizedTensorStorage):
+    """Mixin/storage class that holds data attributes of Float8Tensor.
```
timmoon10: Nit:
```suggestion
    """Mixin class that holds data attributes of Float8Tensor.
```

transformer_engine/pytorch/tensor/storage/float8_blockwise_tensor_storage.py
```diff
@@ -22,8 +22,8 @@
 from ...utils import _empty_tensor
 
 
-class Float8BlockwiseQTensorBase(QuantizedTensorBase):
-    """Mixin class that holds data attributes of Float8BlockwiseQTensor.
+class Float8BlockwiseQTensorStorage(QuantizedTensorStorage):
+    """Mixin/storage class that holds data attributes of Float8BlockwiseQTensor.
```
timmoon10: Nit:
```suggestion
    """Mixin class that holds data attributes of Float8BlockwiseQTensor.
```

transformer_engine/common/recipe/__init__.py
```diff
@@ -370,31 +370,17 @@ class CustomRecipe(Recipe):
...
-                       quantizer_index: int,
-                       num_quantizers: int,
-                       **kwargs,
+                       role: str,
                    )
 
                Where `role` is one of the following strings for e.g. te.Linear
                (stable public contract):
                - forward:  "input", "weight", "output"
                - backward: "grad_output", "grad_input"
```
timmoon10:
```suggestion
               - forward:  "linear_input", "linear_weight", "linear_output"
               - backward: "linear_grad_output", "linear_grad_input"
```

transformer_engine/common/recipe/__init__.py
```diff
@@ -351,3 +356,37 @@ def __repr__(self) -> str:
...
+               given semantic tensor role.
+               The callable is typically invoked as:
+                   qfactory(
+                       role: str,
+                   )
+
+               Where `role` is one of the following strings for e.g. te.Linear
+               (stable public contract):
+               - forward:  "input", "weight", "output"
+               - backward: "grad_output", "grad_input"
```
timmoon10:
```suggestion
               - forward:  "linear_input", "linear_weight", "linear_output"
               - backward: "linear_grad_output", "linear_grad_input"
```

timmoon10 (dismissed):
LGTM. Lot of back and forth, but I think we've converged to a nice design.

timmoon10 (approved):

PR #2041, opened 2025-08-07
[Draft][JAX] Fix layernorm distributed test sharding and collective assertions

jberchtold-nvidia:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/jax/test_distributed_layernorm.py
+++ b/tests/jax/test_distributed_layernorm.py
@@ -19,6 +19,7 @@ from transformer_engine.jax import fp8_autocast
 from transformer_engine.common import recipe
 from transformer_engine.jax.layernorm import layernorm
 from transformer_engine.jax.quantize import QuantizerFactory, ScalingMode, is_fp8_available
+from transformer_engine.jax.cpp_extensions.base import primitive_context
 
 
 DTYPES = [jnp.bfloat16, jnp.float32]
@@ -38,10 +39,9 @@ if is_fp8_supported:
 if is_mxfp8_supported:
     SUPPORTED_RECIPES.append(pytest.param(recipe.MXFP8BlockScaling(), id="MXFP8BlockScaling"))
 
-
 class TestDistributedLayernorm:
 
-    def generate_inputs(self, shape, mesh_resource, dtype, shard_weights):
+    def generate_inputs(self, shape, mesh_resource, dtype):
         weight_shape = (shape[-1],)
 
         x = random.normal(random.PRNGKey(1124), shape, dtype=dtype)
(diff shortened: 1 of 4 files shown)


tests/jax/test_distributed_layernorm.py
```diff
@@ -55,37 +59,50 @@ def generate_inputs(self, shape, mesh_resource, dtype, shard_weights):
...
+
+        # Determine scaling mode for LN from recipe
+        scaling_mode = ScalingMode.NO_SCALING
+        with fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
+            scaling_mode = QuantizerFactory.create_set().x.scaling_mode
+
+        # JAX norm does AllReduce in FP32 due to intermediate casts for precision
+        wgrad_allreduce_dtype = dtype
+        if not use_te_norm or (
+            not is_norm_zero_centered_gamma_in_weight_dtype(scaling_mode) and ln_type == "rmsnorm"
```
jberchtold-nvidia: I'm still debugging why the AR is performed in float32 here in the case when TE norm is enabled and only for RMSNorm. For TE norm this change is internal to the cuDNN plan on a single device and shouldn't affect JAX gradients dtype

tests/jax/test_distributed_layernorm.py
```diff
@@ -55,37 +59,50 @@ def generate_inputs(self, shape, mesh_resource, dtype, shard_weights):
...
+            return generate_collectives_count(allreduce=0, allgather=0, other=0)
+
+        # Determine scaling mode for LN from recipe
+        scaling_mode = ScalingMode.NO_SCALING
+        with fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
+            scaling_mode = QuantizerFactory.create_set().x.scaling_mode
+
+        # JAX norm does AllReduce in FP32 due to intermediate casts for precision
+        wgrad_allreduce_dtype = dtype
+        if not use_te_norm or (
```
jberchtold-nvidia: Our intermediate casts to float32 to keep precision on the JAX implementation seem to be causing our gradient AR to be performed in float32 regardless of input dtype. I checked this wasn't present when removing our casts to float32 in _jax_layernorm and _jax_rmsnorm

Maybe we have some unnecessary casts, I think we may only need `gamma += 1.0` to be in float32 (assuming zero-centered gamma is not supposed to be in weight dtype). Going to try this out and see if that can fix this issue

tests/jax/test_distributed_layernorm.py
jberchtold-nvidia: Fixed, this was actually a JAX norm issue but I was missing the primitive context in the rmsnorm tests to enable/disable TE Norm primitives

tests/jax/test_distributed_layernorm.py
jberchtold-nvidia: Fixed by adjusting casting pattern in our JAX norm implementations. In my local testing the test_custom_compute.py tests of JAX norm vs. TE norm still pass. Running CI now to confirm

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -828,18 +828,23 @@ def _jax_layernorm(x, gamma, beta, zero_centered_gamma, epsilon, quantizer=None)
     """
     JAX native layernorm implementation
     """
-    x_ = jnp.asarray(x, jnp.float32)
```
phu0ngng, line 831: I'm not sure about this change, as without this casting, the `ctype` of normalization will be `bf16`.

transformer_engine/jax/cpp_extensions/base.py
```diff
@@ -258,3 +259,66 @@ def manage_primitives(enable_names=None, disable_names=None, disable_all_first=F
...
+        primitives = _parse_custom_call_string_to_dict(orig_env_var)
+
+    changes = _parse_custom_call_string_to_dict(primitive_enabling_changes)
+    primitives.update(changes)
+
+    updated_env_var = ",".join(
+        f"{name}={'true' if enabled else 'false'}" for name, enabled in primitives.items()
+    )
+
+    os.environ["NVTE_JAX_CUSTOM_CALLS"] = str(updated_env_var)
```
phu0ngng, line 317: Hi, 

I think we could use `manage_primitives` here and don't need to change the env var.
If you prefer to pass key-value pairs, we could update the `manage_primitives` to do so too.

jberchtold-nvidia: Thanks, just read the full docstring on `manage_primitives` and you're right it will work. I'll update to that instead for this context

I had missed the `disable_all_first` argument and thought `manage_primitives` was always disabling any unspecified primitives

PR #2042, opened 2025-08-08
Fuse linear+scale+add

janekb04:
# Description

This PR adds two new fusions, `ForwardLinearScaleAdd` = `BasicLinear` + `ConstantScale` + `AddExtraInput(in_place=True)` and `BackwardLinearScale` = `BasicLinear` + `ConstantScale`. It also adds a new function to the C++ interface, used to implement those fusions, `nvte_cublas_gemm_scaled`, which allows the user to pass in `alpha` and `beta` parameters into the GEMM.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- (69accf6a3ac617a51b226d2a956fa8c60de518e7) Add `nvte_cublas_gemm_scaled`. Change interface of internal function `cublas_gemm` to accept `alpha` and `beta` instead of `accumulate`. Change existing code using `accumulate` to pass in `beta=1` instead.
- (5a267a6fca1f8b09e25b6fc10e5586bfee0f1ad0) Support use of `alpha` and `beta` in `tex.generic_gemm` by adding them as optional kwargs and making the implementation use `nvte_cublas_gemm_scaled` instead of `nvte_cublas_gemm`.
- (61058b0ebd6bc7d97a3c23a1bdd5cbfa88d81350) Support use of `alpha` and `beta` in `general_gemm`
- (22177d519a4c7c06a28cba9496f21d7a4b5f6423) Support use of `alpha` and `beta` in `BasicLinear._functional_forward` and `BasicLinear._functional_backward`. Separate `alpha`s and `beta`s are allowed for each of the three GEMMs (forward GEMM, `dgrad` GEMM, `wgrad` GEMM).
- (1104c934bce9619012c10398ec9fd44df630186e) Add `ForwardLinearScaleAdd` fusion and `test_forward_linear_scale_add` unit test
- (2b985e3a02e5a9322ff97a543c4946c8634ff14b) Add `BackwardLinearScale` fusion and `test_backward_linear_scale` unit test. Note: this makes the `test_forward_linear_scale_add` unit test also use this fusion implicitly as backward and forward fusions cannot be currently individually toggled. However, as the `ForwardLinearScaleAdd` fusion happens only in `test_forward_linear_scale_add`, the existence of the separate `test_backward_linear_scale` test allows for distinguishing whether an issue stems from the forward or the backward fusion.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/cpp_extensions/gemm.py
+++ b/transformer_engine/pytorch/cpp_extensions/gemm.py
@@ -20,6 +20,14 @@ __all__ = [
     "general_grouped_gemm",
 ]
 
+def validate_gemm_scale(scale: Optional[float], required: bool) -> float:
+    """Validate whether a GEMM scaling factor is consistent with its usage"""
+    if required:
+        return scale if scale is not None else 1.0
+    else:
+        if scale not in (0.0, None):
+            raise ValueError("scale must be zero")
+        return 0.0
 
 def general_gemm(
     A: torch.Tensor,
@@ -29,6 +37,8 @@ def general_gemm(
     quantization_params: Optional[Quantizer] = None,
     gelu: bool = False,
     gelu_in: torch.Tensor = None,
+    alpha: float = 1.0,
+    beta: Optional[float] = None,
     accumulate: bool = False,
     layout: str = "TN",
     out: Optional[torch.Tensor] = None,
@@ -46,6 +56,9 @@ def general_gemm(
     transa = layout[0] == "T"
     transb = layout[1] == "T"
     # assert quantization_params is None, "FP8 output not supported yet"
+
+    alpha = validate_gemm_scale(alpha, True)
+    beta = validate_gemm_scale(beta, accumulate)
 
     if ub_type is not None:
         assert ub is not None, (
@@ -108,6 +121,8 @@ def general_gemm(
         "comm_type": ub_type,
         "extra_output": extra_output,
         "bulk_overlap": bulk_overlap,
+        "alpha": alpha,
+        "beta": beta,
     }
 
     out, bias_grad, gelu_input, extra_output = tex.generic_gemm(*args, **kwargs)
--- a/transformer_engine/pytorch/cpp_extensions/gemm.py
+++ b/transformer_engine/pytorch/cpp_extensions/gemm.py
@@ -18,6 +18,7 @@ from ...debug.pytorch.debug_quantization import DebugQuantizer
 __all__ = [
     "general_gemm",
     "general_grouped_gemm",
+    "validate_gemm_scale",
 ]
 
 def validate_gemm_scale(scale: Optional[float], required: bool) -> float:
--- a/transformer_engine/pytorch/ops/fused/forward_linear_scale_add.py
+++ b/transformer_engine/pytorch/ops/fused/forward_linear_scale_add.py
@@ -1,0 +1,177 @@
+# Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""Fused operation for forward GEMM + scale + add."""
+
+from __future__ import annotations
+from collections.abc import Iterable
+from typing import Any, Optional
+
+import torch
+
+from transformer_engine.pytorch.fp8 import FP8GlobalStateManager
+from transformer_engine.pytorch.ops.basic import AddExtraInput, BasicLinear, ConstantScale
+from transformer_engine.pytorch.ops.op import (
+    FusedOperation,
+    FusibleOperation,
+    OperationContext,
+)
+from transformer_engine.pytorch.tensor import Quantizer
+
+
+class ForwardLinearScaleAdd(FusedOperation):
+    """Fused forward GEMM + scale + add
+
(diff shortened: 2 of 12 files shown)


transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -18,9 +18,19 @@
 __all__ = [
     "general_gemm",
     "general_grouped_gemm",
+    "validate_gemm_scale",
```
timmoon10: Why do we need to expose this externally? I see we call it in the `BasicLinear` op, and then call it again within `general_gemm`.

transformer_engine/pytorch/ops/fused/forward_linear_scale_add.py
```diff
@@ -0,0 +1,176 @@
...
+import torch
+
+from transformer_engine.pytorch.fp8 import FP8GlobalStateManager
+from transformer_engine.pytorch.ops.basic import AddExtraInput, BasicLinear, ConstantScale
+from transformer_engine.pytorch.ops.op import (
+    FusedOperation,
+    FusibleOperation,
+    OperationContext,
+)
+from transformer_engine.pytorch.tensor import Quantizer
```
timmoon10: Nit: Prefer relative imports so this is insensitive to `PYTHONPATH`.
```suggestion
from ...fp8 import FP8GlobalStateManager
from ...tensor import Quantizer
from ..basic import AddExtraInput, BasicLinear, ConstantScale
from ..op import FusedOperation, FusibleOperation, OperationContext
```

transformer_engine/pytorch/ops/fused/forward_linear_scale_add.py
```diff
@@ -0,0 +1,176 @@
...
+        scale_op = self.basic_ops[1]
+
+        # Check which grads are required
+        input_requires_grad = linear_op_ctx.requires_grad
+        weight_requires_grad = linear_op_ctx.requires_grad and linear_op.weight.requires_grad
+
+        # FP8 metadata
+        input_quantizer = linear_op.get_quantizer("forward", 0)
+        weight_quantizer = linear_op.get_quantizer("forward", 1)
+        output_quantizer = next_op_input_quantizer
```
timmoon10: I see we don't support accumulating into an FP8 tensor, so the output quantizer doesn't make sense for this case.
```suggestion
        output_quantizer = None
```

transformer_engine/pytorch/cpp_extensions/gemm.py
janekb04: My intention wasn't exposing it externally. I thought that if a module uses `__all__`, only symbols named in `__all__` can be imported from it from other modules. I will fix this.

transformer_engine/pytorch/ops/fused/forward_linear_scale_add.py
janekb04: Ok, I will change this. I do not understand the logic behind this, though. It seems like code all over TE uses both absolute and relative imports. Some files even mix both (ex. `op.py`). The codebase assumes that `transformer_engine` is a valid top level module from which the submodules can be accessed.

transformer_engine/pytorch/ops/fused/forward_linear_scale_add.py
janekb04: I included this only for forward compatibility, but I realize now that currently this would cause an error in the C++ implementation if the fused operation was followed by a quantize. I will fix this and add a quantization after this fused op to the test to catch this.

transformer_engine/pytorch/cpp_extensions/gemm.py
janekb04: It seems that in this case the import indeed fails as I remembered: `ImportError: cannot import name 'validate_gemm_scale' from 'transformer_engine.pytorch.cpp_extensions'`. This happens because the name is imported not directly from `gemm`, but from `cpp_extensions` through a glob import. I will fix it to import directly from `gemm`, so the symbol doesn't have to be included in `__all__`.

transformer_engine/pytorch/ops/fused/forward_linear_scale_add.py
timmoon10: It's mostly stylistic, and once we pick one we should try to be consistent. For what it's worth, the Google Style Guide prefers absolute imports: https://google.github.io/styleguide/pyguide.html#22-imports

transformer_engine/pytorch/cpp_extensions/gemm.py
timmoon10: Why do we need to call `validate_gemm_scale` in `BasicLinear` if it'll be called again in `general_gemm`?

janekb04: Technically we don’t. We could also technically avoid it in `general_gemm` by passing `Optional[float]` to C++ and handling it there. I just wanted to detect the error as soon as possible. I placed the validation in `BasicLinear`, `general_gemm` and `gemm` in C++, as they could all possibly be called directly. I guess that the validation could be removed from `BasicLinear`. It’s just that it already includes a lot of validation of parameters rather than relying on lower level code to check it and report back, so this seemed consistent.

PR #2044, opened 2025-08-08
[PyTorch] Update amax pointers when reallocating amax history in fusible ops

timmoon10:
# Description

The fusible ops have a bug with FP8 delayed scaling where the amax pointers in the quantizer and `FP8GlobalStateManager` are not updated when the amax history is reallocated. This results in the FP8 scales never being updated.

This hadn't caused problems since we initialized the amax history during the first forward pass and it's uncommon to change the amax history length during training. However, https://github.com/NVIDIA/TransformerEngine/pull/1951 changed behavior to initialize the recipe state when an op is constructed with quantized params (see https://github.com/NVIDIA/TransformerEngine/pull/1985 for a related bug). If `fp8_model_init` and `fp8_autocast` use recipes with different amax history lengths (e.g. Megatron-LM uses default in `fp8_model_init`), then we reallocate the amax history during the first forward pass and the amax pointers get invalidated.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Update amax pointers when reallocating amax history in fusible ops

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/ops/fuser.py
+++ b/transformer_engine/pytorch/ops/fuser.py
@@ -110,14 +110,6 @@ class _OperationFuserAutogradFunction(torch.autograd.Function):
             xs, extra_inputs = _split_tuple(extra_inputs, op.num_extra_inputs)
             basic_op_extra_inputs.append(xs)
 
-        # Get environment state
-        with_quantized_compute = FP8GlobalStateManager.is_fp8_enabled()
-        recipe = FP8GlobalStateManager.get_fp8_recipe() if with_quantized_compute else None
-        is_grad_enabled = func_ctx is not None
-
-        # Attempt to fuse operations if neccesary
-        fuser.maybe_fuse_ops(is_grad_enabled, recipe, input_, basic_op_extra_inputs)
-
         # Apply forward ops
         x = input_
         extra_outputs = [None] * fuser._num_basic_ops
@@ -167,7 +159,7 @@ class _OperationFuserAutogradFunction(torch.autograd.Function):
             extra_outputs_flat.extend(ys)
 
         # Save context for backward pass
-        if is_grad_enabled:
+        if func_ctx is not None:
 
             # Flatten list of saved tensors
             to_save = []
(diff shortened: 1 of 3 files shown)


transformer_engine/pytorch/ops/op.py
```diff
@@ -308,6 +310,25 @@ def reset_recipe_state(
...
+                        if buffer_key in FP8GlobalStateManager.global_amax_buffer:
+                            assert (
+                                buffer_key in FP8GlobalStateManager.global_amax_history_buffer
+                            ), "TE internal error during amax history change."
+                            FP8GlobalStateManager.global_amax_buffer[buffer_key][pos] = (
+                                recipe_state.amax_history[0]
+                            )
+                            FP8GlobalStateManager.global_amax_history_buffer[buffer_key][
+                                pos
+                            ] = recipe_state.amax_history
```
timmoon10, line 330: This logic is similar to `TransformerEngineBaseModule.adjust_amax_history_length`:
https://github.com/NVIDIA/TransformerEngine/blob/cae1c436027cc028ee49c83463141ba67f0adca0/transformer_engine/pytorch/module/base.py#L641-L661

janekb04: I noticed that `self._update_weight_quantizers()` doesn't have a counterpart. #1951 removed this logic ([see here](https://github.com/NVIDIA/TransformerEngine/pull/1951/files#diff-864ad36a21c571fb178499535cfada611df4a82223c9ffbfea872dda39972eaeL355-L361)). Because we pass `weight_quantizer` to `_functional_forward` and `_functional_backward`, perhaps we don't need to update `weight._quantizer`. But I don't know if some other places rely on it being updated. If this has to be added back, I think it would best belong in `BasicLinear.reset_recipe_state`.

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -110,14 +110,6 @@ def forward(
             xs, extra_inputs = _split_tuple(extra_inputs, op.num_extra_inputs)
             basic_op_extra_inputs.append(xs)
 
-        # Get environment state
-        with_quantized_compute = FP8GlobalStateManager.is_fp8_enabled()
-        recipe = FP8GlobalStateManager.get_fp8_recipe() if with_quantized_compute else None
-        is_grad_enabled = func_ctx is not None
-
-        # Attempt to fuse operations if neccesary
-        fuser.maybe_fuse_ops(is_grad_enabled, recipe, input_, basic_op_extra_inputs)
```
timmoon10, line 119: I've moved the op fusion outside of the autograd context because `torch.is_grad_enabled()` is `False` within autograd functions, which was messing up logic for quantizer usages.

janekb04:
LGTM

PR #2045, opened 2025-08-08
[Pytorch] Add Cutlass GroupGEMM Support for fine-grained MoE Model

cassiewilliam:
# Description

Add Cutlass Group GEMM Support for H100(SM90), which provides greater performance advantages compared to the current Multi-Stream implementation in Fine-Grained MoE models. Currently, this PR only supports FP16 and BF16 scenarios, and FP8 support is not yet available. The implementation is limited to the standard MoE Module (Bias and other related features have not been validated yet). Please take note.

Initial performance test results are as follows, and the testing method can be found in file test_group_gemm.py.

run test script with

`python tests/pytorch/test_group_gemm.py`

Shape(g,m,n,k) | TE V2.2 (TFLOPs) | Cutlass-Opt-V1(TFLOPs) | Speed-Up
-- | -- | -- | --
(8, 4096, 768, 2048) | 508.77 | 568.63 | 11.77%
(16, 2048, 768, 2048) | 398.81 | 534.75 | 34.08%

(Update by @yaox12)
Benchmark the performance of `general_grouped_gemm` with TE 2.7 and CUDA 12.9:
Shape(g,m,n,k) | TE V2.7 (TFLOPs) | Cutlass (TFLOPs) | Speed-Up
-- | -- | -- | --
(8, 4096, 768, 2048) | 627.903 | 650.574 | 3.6%
(16, 2048, 768, 2048) | 553.945 | 636.927 | 14.98%

Add the system environment variable `NVTE_USE_CUTLASS_GROUPED_GEMM` to toggle between the two GEMM implementations. Setting `export NVTE_USE_CUTLASS_GROUPED_GEMM=0` selects the original Multi-Stream cuBLAS GEMM, while setting `export NVTE_USE_CUTLASS_GROUPED_GEMM=1` enables the newly added CUTLASS Group GEMM. The default value is 0.

Add a environment variable `NVTE_CUTLASS_GROUPED_GEMM_WARN_FALLBACK`, when it's set, it will give warnings when falling back to the cuBLAS implementaion.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


phu0ngng:
Hi @cassiewilliam ,

Thank you for a great PR - it’s good to see such a clear performance improvement!

I have one suggestion - I think we should refactor the change slightly to minimize modifications in the TE framework extensions.

Currently, we have two separate C APIs: `nvte_multi_stream_cublas_gemm` and `nvte_cutlass_grouped_gemm`. The PyTorch extensions call these individually, and we would need to do the same on the JAX side. Since they share the same function signature, we could unify them into a single API - `nvte_multi_tensor_gemm` - and deprecate `nvte_multi_stream_cublas_gemm`.

Within `nvte_multi_tensor_gemm`, we can determine the GPU architecture and enable CUTLASS GroupedGEMM for FP16/BF16 on Hopper. This way, future changes to the GroupedGEMM implementation or backend would not require modifications to the PyTorch/JAX extensions.

alan-hpc:
> Hi @cassiewilliam ,
> 
> Thank you for a great PR - it’s good to see such a clear performance improvement!
> 
> I have one suggestion - I think we should refactor the change slightly to minimize modifications in the TE framework extensions.
> 
> Currently, we have two separate C APIs: `nvte_multi_stream_cublas_gemm` and `nvte_cutlass_grouped_gemm`. The PyTorch extensions call these individually, and we would need to do the same on the JAX side. Since they share the same function signature, we could unify them into a single API - `nvte_multi_tensor_gemm` - and deprecate `nvte_multi_stream_cublas_gemm`.
> 
> Within `nvte_multi_tensor_gemm`, we can determine the GPU architecture and enable CUTLASS GroupedGEMM for FP16/BF16 on Hopper. This way, future changes to the GroupedGEMM implementation or backend would not require modifications to the PyTorch/JAX extensions.

I fully agree with your suggestion — keeping the code architecture clean is very important. Will you be handling the refactor on your side, or should I go ahead and make the changes directly in the current PR?

yaox12:
Agree with @phu0ngng. We could unify the API and do the dispatch (based on GPU arch/data type/env variable) on the TE/common side.

> Will you be handling the refactor on your side, or should I go ahead and make the changes directly in the current PR?

Please go ahead in this PR.

alan-hpc:
> Agree with @phu0ngng. We could unify the API and do the dispatch (based on GPU arch/data type/env variable) on the TE/common side.
> 
> > Will you be handling the refactor on your side, or should I go ahead and make the changes directly in the current PR?
> 
> Please go ahead in this PR.

Got it — I’ll refactor the code to meet the requirements described above.

alan-hpc:
hello @phu0ngng @yaox12 The nvte_multi_tensor_gemm interface has been fully refactored. Please review the implementation for correctness and compliance with the updated design.

tests/pytorch/test_group_gemm.py
```diff
@@ -0,0 +1,241 @@
...
+
+            # FLOPs 计算
+            total_flops = sum([2 * m * n * k for (m, n, k) in self.group_config])
+            tflops = total_flops / (avg_time_ms * 1e9)
+
+            print(f"🔥 avgerage used time: {avg_time_ms:.3f} ms/iter")
+            print(f"⚡ average throughput: {tflops:.2f} TFLOPs")
+
+
+def run_grouped_gemm(group_config, check_performance, transa, transb, accumulate):
```
phu0ngng: Hi, I think we should be able to unify this test with the existing test suite here. https://github.com/NVIDIA/TransformerEngine/blob/main/tests/pytorch/test_numerics.py#L2520

transformer_engine/common/CMakeLists.txt
```diff
@@ -114,15 +119,25 @@ add_library(transformer_engine SHARED ${transformer_engine_SOURCES})
 target_include_directories(transformer_engine PUBLIC
                            "${CMAKE_CURRENT_SOURCE_DIR}/include")
 
-
+if(${CMAKE_CUDA_COMPILER_VERSION} VERSION_GREATER 12.0)
+  set_source_files_properties(
+    "gemm/cublaslt_gemm.cu"
+    PROPERTIES
+    COMPILE_FLAGS
+    "-gencode arch=compute_90a,code=sm_90a")
```
phu0ngng: Hi, why do we only generate code for SM-90a?

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -718,3 +721,86 @@ using cublasHandleManager = detail::HandleManager<cublasLtHandle_t, CreateCublas
 void nvte_cublas_handle_init() { auto _ = cublasHandleManager::Instance().GetHandle(); }
 
 }  //  namespace transformer_engine
+
+void nvte_cutlass_grouped_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor *D,
+                               const int num_gemms, bool transa, bool transb, bool grad,
+                               NVTETensor *workspace, bool accumulate, int math_sm_count,
+                               cudaStream_t stream) {
+  NVTE_API_CALL(nvte_cutlass_grouped_gemm);
```
phu0ngng: Hi, I think we should remove the `nvte` prefix here as we don't plan to expose this API as a C API.

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -718,3 +721,86 @@ using cublasHandleManager = detail::HandleManager<cublasLtHandle_t, CreateCublas
...
+  using namespace transformer_engine;
+
+  // wait for current stream to finish
+  static cudaEvent_t finish_event;
+  static std::once_flag event_init_flag;
+  std::call_once(event_init_flag, []() { NVTE_CHECK_CUDA(cudaEventCreate(&finish_event)); });
```
phu0ngng: Hi, may I know why we need to wait for the current stream here, and after the CUTLASS call?

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -718,3 +721,86 @@ using cublasHandleManager = detail::HandleManager<cublasLtHandle_t, CreateCublas
...
+  using namespace transformer_engine;
+
+  // NOTE(Alan): Read env flag (default = false)
+  bool cutlass_group_gemm_flag =
+      transformer_engine::getenv<bool>("NVTE_USE_CUTLASS_GROUPGEMM", false);
+
+  if (!cutlass_group_gemm_flag) {
+    nvte_multi_stream_cublas_gemm(A, B, D, bias, pre_gelu_out, num_gemms, transa, transb, grad,
+                                  workspace, accumulate, use_split_accumulator, math_sm_count,
+                                  stream);
```
phu0ngng: It would be great if you could help to add a deprecating warning message inside the `nvte_multi_stream_cublas_gemm`.

transformer_engine/common/gemm/cutlass_groupgemm.cuh
```diff
@@ -0,0 +1,343 @@
...
+  }
+}
+
+inline void check_device_ptr(const void* ptr, const char* name) {
+  cudaPointerAttributes attr;
+  auto err = cudaPointerGetAttributes(&attr, ptr);
+  if (err != cudaSuccess || attr.devicePointer == nullptr) {
+    printf("[Error] Pointer %s is not a valid device pointer\n", name);
+  } else {
+    printf("[OK] Pointer %s is valid on device %d\n", name, attr.device);
```
phu0ngng: Hi, please use `NVTE_CHECK` here instead of `printf`.

transformer_engine/common/gemm/cutlass_groupgemm.cuh
```diff
@@ -0,0 +1,343 @@
...
+
+  using StrideA = typename Gemm::GemmKernel::InternalStrideA;
+  using StrideB = typename Gemm::GemmKernel::InternalStrideB;
+  using StrideC = typename Gemm::GemmKernel::InternalStrideC;
+
+  typename Gemm::Arguments arguments;
+  size_t workspace_size = Gemm::get_workspace_size(arguments);
+
+  transformer_engine::Tensor* wspace = transformer_engine::convertNVTETensor(workspace[0]);
+  char* all_workspace = (char*)(wspace->data.dptr);
```
phu0ngng: Could we add an assertion to check whether the provided workspace buffer is large enough for your required `workspace_size`?

transformer_engine/common/gemm/cublaslt_gemm.cu
alan-hpc: Add Warning Info Like this, OK?
  NVTE_WARN(
      "nvte_multi_stream_cublas_gemm is deprecated and will be removed in a future release. "
      "Please migrate to nvte_multi_tensor_gemm (with CUTLASS Grouped GEMM support where "
      "applicable).");

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -684,6 +686,12 @@ void nvte_multi_stream_cublas_gemm(const NVTETensor *A, const NVTETensor *B, NVT
   NVTE_API_CALL(nvte_multi_stream_cublas_gemm);
   using namespace transformer_engine;
 
+  // Deprecation warning
```
yaox12: If we added a deprecation warning here, then calling it from `nvte_multi_tensor_gemm` would also raise a warning, which is unexpected.

I think a more proper way is to rename the current `nvte_multi_stream_cublas_gemm` to `multi_stream_cublas_gemm`, and call the new function in `nvte_multi_tensor_gemm`.
At the same time, add a new `nvte_multi_stream_cublas_gemm` like
```cpp
void nvte_multi_stream_cublas_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor *D,
                                   const NVTETensor *bias, NVTETensor *pre_gelu_out,
                                   const int num_gemms, bool transa, bool transb, bool grad,
                                   NVTETensor *workspace, bool accumulate,
                                   bool use_split_accumulator, int math_sm_count,
                                   cudaStream_t stream) {
NVTE_WARN(
      "nvte_multi_stream_cublas_gemm is deprecated and will be removed in a future release. "
      "Please migrate to nvte_multi_tensor_gemm (with CUTLASS Grouped GEMM support where "
      "applicable).");
multi_stream_cublas_gemm(
    A, B, D, bias, pre_gelu_out, num_gemms, transa, transb, grad,
    workspace, accumulate, use_split_accumulator, math_sm_count,
    stream);
}
```

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -718,3 +726,46 @@ using cublasHandleManager = detail::HandleManager<cublasLtHandle_t, CreateCublas
...
+
+  auto is_empty_arr = [&](const NVTETensor *p) -> bool {
+    if (p == nullptr) return true;
+    for (int i = 0; i < num_gemms; ++i) {
+      if (transformer_engine::convertNVTETensor(p[i])->has_data()) return false;
+    }
+    return true;
+  };
+
+  // Currently only supports the case when bias is not null, in this case the grad flag can be ignored.
```
yaox12: "when bias is ~~not~~ null" ?

transformer_engine/common/gemm/cutlass_groupgemm.cu
```diff
@@ -0,0 +1,96 @@
+#include "cutlass/bfloat16.h"
```
yaox12: Add a licence here.

transformer_engine/common/gemm/cutlass_groupgemm.cu
```diff
@@ -0,0 +1,96 @@
...
+    case DType::kFloat8E4M3:
+      return CUDA_R_8F_E4M3;
+    case DType::kFloat8E5M2:
+      return CUDA_R_8F_E5M2;
+    default:
+      NVTE_ERROR("Invalid type");
+  }
+}
+
+// 显式实例化，配合 .cuh 中的 extern template
```
yaox12: Comment in English.

transformer_engine/common/gemm/cutlass_groupgemm.cuh
```diff
@@ -0,0 +1,326 @@
...
+  if (gemm.run(stream) != cutlass::Status::kSuccess) {
+    NVTE_CHECK(false, "Failed to run CUTLASS Grouped GEMM");
+  }
+
+  std::free(host_workspace);
+}
+
+}  // namespace grouped_gemm
+
+void nvte_cutlass_grouped_gemm(const NVTETensor* A, const NVTETensor* B, NVTETensor* D,
```
yaox12: Per @phu0ngng 's comments, please remove the `nvte_` prefix as we don't want to expose it as a public API. Only `nvte_multi_tensor_gemm` is exposed.

yaox12:
To pass the DCO check, you have to sign your commits. See [here](https://github.com/NVIDIA/TransformerEngine/pull/2045/checks?check_run_id=48273151332) for more details.

transformer_engine/common/gemm/cublaslt_gemm.cu
alan-hpc: OK

alan-hpc:
> To pass the DCO check, you have to sign your commits. See [here](https://github.com/NVIDIA/TransformerEngine/pull/2045/checks?check_run_id=48273151332) for more details.

Appreciate the feedback — the changes have been made.

yaox12:
Please add a test here. https://github.com/NVIDIA/TransformerEngine/blob/main/tests/pytorch/test_numerics.py#L2520.  
I think you can add an argument `use_cutlass_group_gemm`, and set the environment variable if it's true (remember to unset it at the end of test). I won't expect CUTLASS implementation to be bit-wise match with the baseline, so you need to set tolerances.

yaox12:
@cassiewilliam I enabled some tests in https://github.com/NVIDIA/TransformerEngine/pull/2045/commits/e928062da7747055f29130ec27201a663754d7d8. Please make sure the following tests pass.
```
pytest -v -s tests/pytorch/test_numerics.py::test_grouped_linear_accuracy_cutlass
pytest -v -s tests/pytorch/test_numerics.py::test_grouped_gemm
```
I tested it locally and found tests with `fuse_wgrad_accumulation = True` would fail. When enabling `fuse_wgrad_accumulation`, the output is usually in FP32, but you're using the same data type for C as A/B. This may explain the failures in `test_grouped_linear_accuracy_cutlass`. But in `test_grouped_gemm`, the output is in the same dtype as inputs, so I'm not sure why they failed. Maybe the tolerance is not correctly set.

transformer_engine/common/gemm/cutlass_groupgemm.cuh
```diff
@@ -0,0 +1,332 @@
...
+using GroupedGemmInputBLayout =
+    std::conditional_t<transb, ::cutlass::layout::ColumnMajor, ::cutlass::layout::RowMajor>;
+
+using ProblemShapeType = cute::Shape<int, int, int>;
+using ProblemShape = cutlass::gemm::GroupProblemShape<ProblemShapeType>;  // <M,N,K> per group
+template <typename ScheduleConfig>
+struct GemmGivenSchedule {
+  using ElementA = typename ScheduleConfig::DataType;  // Element type for A matrix operand
+  using ElementB = typename ScheduleConfig::DataType;  // Element type for B matrix operand
+  using ElementC = typename ScheduleConfig::DataType;  // Element type for C and D matrix operands
```
yaox12: When enabling `fuse_wgrad_accumulation`, C/D could have a different data type from A/B.

alan-hpc: Thank you for taking the time to review this. I’ll make the necessary changes and push the fixes as soon as possible.

alan-hpc:
> @cassiewilliam I enabled some tests in [e928062](https://github.com/NVIDIA/TransformerEngine/commit/e928062da7747055f29130ec27201a663754d7d8). Please make sure the following tests pass.
> 
> ```
> pytest -v -s tests/pytorch/test_numerics.py::test_grouped_linear_accuracy_cutlass
> pytest -v -s tests/pytorch/test_numerics.py::test_grouped_gemm
> ```
> 
> I tested it locally and found tests with `fuse_wgrad_accumulation = True` would fail. When enabling `fuse_wgrad_accumulation`, the output is usually in FP32, but you're using the same data type for C as A/B. This may explain the failures in `test_grouped_linear_accuracy_cutlass`. But in `test_grouped_gemm`, the output is in the same dtype as inputs, so I'm not sure why they failed. Maybe the tolerance is not correctly set.

Thank you for taking the time to review this. I’ll make the necessary changes and push the fixes as soon as possible.

zhongbozhu:
Great PR! Thank you so much. CC @timmoon10

transformer_engine/common/gemm/cutlass_groupgemm.cuh
```diff
@@ -0,0 +1,332 @@
...
+  if (gemm.initialize(arguments, workspace_ptr) != cutlass::Status::kSuccess) {
+    NVTE_CHECK(false, "Failed to initialize CUTLASS Grouped GEMM");
+  }
+
+  // Execute the kernel in the current stream.
+  if (gemm.run(stream) != cutlass::Status::kSuccess) {
+    NVTE_CHECK(false, "Failed to run CUTLASS Grouped GEMM");
+  }
+
+  std::free(host_workspace);
```
leefige: Nice work! I would like to remind that this `free` will be dangerous if kernel launch on host is much faster than kernel execution on GPU, because I don't think `gemm.run` implies stream synchronization. In that case, before `cudaMemcpyAsync` could copy `host_workspace` to `all_workspace`, `host_workspace` may have been destroyed.

So I suggest adding a CUDA stream sync before `free`. Otherwise, you can try a different way to maintain ptrs and shapes in device memory.

phu0ngng: Hi, 

I don't think we can add a CUDA stream sync before free, as it will break cudaGraph. 
Alternatively, we can use a smart pointer here.

leefige: Hi, thanks for the reply. I agree with the CUDA graph part. However, I don't think a smart pointer solves the problem. Imagine when the host has returned from this function before the memcpy is executed on the device. After going out of this scope, the smart pointer will be destroyed, and further access to the host workspace will be unsafe. 

A host workspace maintained out of this scope may be an option, but I guess this requires more modification on the framework, even the API.

alan-hpc: Maybe using a static CPU workspace (lifetime tied to the process) could solve this. CUDA stream sync is too costly.

leefige: I agree. It seems possible to maintain a host workspace similar to `_cublas_workspace` or `_multi_stream_cublas_workspace` whose lifetimes are bound to TE.

alan-hpc: > I agree. It seems possible to maintain a host workspace similar to `_cublas_workspace` or `_multi_stream_cublas_workspace` whose lifetimes are bound to TE.

Refactored to use std::call_once with a static host workspace.
This ensures the buffer is allocated only once and safely released at process exit,
while avoiding repeated malloc/free calls during GEMM execution.

yaox12:
@cassiewilliam I'm still seeing some failures in `test_grouped_gemm`.  See https://github.com/NVIDIA/TransformerEngine/actions/runs/17169254634/job/48715768402 (expand the `Log` session). Other failures are irrelevant.

alan-hpc:
> @cassiewilliam I'm still seeing some failures in `test_grouped_gemm`. See https://github.com/NVIDIA/TransformerEngine/actions/runs/17169254634/job/48715768402 (expand the `Log` session). Other failures are irrelevant.



> @cassiewilliam I'm still seeing some failures in `test_grouped_gemm`. See https://github.com/NVIDIA/TransformerEngine/actions/runs/17169254634/job/48715768402 (expand the `Log` session). Other failures are irrelevant.


[test_grouped_linear_accuracy_cutlass.log](https://github.com/user-attachments/files/22000894/test_grouped_linear_accuracy_cutlass.log)
[test_grouped_gemm.log](https://github.com/user-attachments/files/22000933/test_grouped_gemm.log)

It runs fine on my local H100 GPU machine, but the unit tests still fail, which feels a bit odd.
Could you please help take a look?

yaox12:
> > @cassiewilliam I'm still seeing some failures in `test_grouped_gemm`. See https://github.com/NVIDIA/TransformerEngine/actions/runs/17169254634/job/48715768402 (expand the `Log` session). Other failures are irrelevant.
> 
> > @cassiewilliam I'm still seeing some failures in `test_grouped_gemm`. See https://github.com/NVIDIA/TransformerEngine/actions/runs/17169254634/job/48715768402 (expand the `Log` session). Other failures are irrelevant.
> 
> [test_grouped_linear_accuracy_cutlass.log](https://github.com/user-attachments/files/22000894/test_grouped_linear_accuracy_cutlass.log) [test_grouped_gemm.log](https://github.com/user-attachments/files/22000933/test_grouped_gemm.log)
> 
> It runs fine on my local H100 GPU machine, but the unit tests still fail, which feels a bit odd. Could you please help take a look?

I can reproduce the failure. I use the docker image `nvcr.io/nvidia/pytorch:25.06-py3`, install your branch inside it, run `pytest -v -s tests/pytorch/test_numerics.py::test_grouped_gemm` and get 
```
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.float16-shape0] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.float16-shape1] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.float16-shape2] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.float16-shape3] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape0] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape1] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape2] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape3] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.float16-shape0] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.float16-shape1] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.float16-shape2] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.float16-shape3] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape0] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape1] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape2] - AssertionError: Tensor-likes are not close!
FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape3] - AssertionError: Tensor-likes are not close!
```

My installing command is 
```
unset PIP_CONSTRAINT && NVTE_CUDA_ARCHS="90" NVTE_BUILD_THREADS_PER_JOB=8 NVTE_FRAMEWORK=pytorch pip install --no-build-isolation git+https://github.com/cassiewilliam/TransformerEngine.git@feature/cutlass_group_gemm_support
```

alan-hpc:
> pytest -v -s tests/pytorch/test_numerics.py::test_grouped_gemm



> > > @cassiewilliam I'm still seeing some failures in `test_grouped_gemm`. See https://github.com/NVIDIA/TransformerEngine/actions/runs/17169254634/job/48715768402 (expand the `Log` session). Other failures are irrelevant.
> > 
> > 
> > > @cassiewilliam I'm still seeing some failures in `test_grouped_gemm`. See https://github.com/NVIDIA/TransformerEngine/actions/runs/17169254634/job/48715768402 (expand the `Log` session). Other failures are irrelevant.
> > 
> > 
> > [test_grouped_linear_accuracy_cutlass.log](https://github.com/user-attachments/files/22000894/test_grouped_linear_accuracy_cutlass.log) [test_grouped_gemm.log](https://github.com/user-attachments/files/22000933/test_grouped_gemm.log)
> > It runs fine on my local H100 GPU machine, but the unit tests still fail, which feels a bit odd. Could you please help take a look?
> 
> I can reproduce the failure. I use the docker image `nvcr.io/nvidia/pytorch:25.06-py3`, install your branch inside it, run `pytest -v -s tests/pytorch/test_numerics.py::test_grouped_gemm` and get
> 
> ```
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.float16-shape0] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.float16-shape1] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.float16-shape2] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.float16-shape3] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape0] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape1] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape2] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape3] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.float16-shape0] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.float16-shape1] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.float16-shape2] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.float16-shape3] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape0] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape1] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape2] - AssertionError: Tensor-likes are not close!
> FAILED tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape3] - AssertionError: Tensor-likes are not close!
> ```
> 
> My installing command is
> 
> ```
> unset PIP_CONSTRAINT && NVTE_CUDA_ARCHS="90" NVTE_BUILD_THREADS_PER_JOB=8 NVTE_FRAMEWORK=pytorch pip install --no-build-isolation git+https://github.com/cassiewilliam/TransformerEngine.git@feature/cutlass_group_gemm_support
> ```
Very strange, the accuracy seems to be inconsistent in different docker environments

phu0ngng:
@yaox12 could you help resolve the remaining mismatches?
Let's try to get this merge for TE 2.8. Thanks!

yaox12:
We're still seeing mismatch when `accumulate = True` (`beta = 1.0`) and A/B/C dtype = BF16, and the mismatch is significant. I and @QiZhangNV are investigating it. Do you have any ideas? @cassiewilliam 

```
E               Mismatched elements: 327553 / 1283456 (25.5%)
E               Greatest absolute difference: 0.5 at index (14, 4) (up to 0.002 allowed)
E               Greatest relative difference: inf at index (127, 1) (up to 0.001 allowed)
...
=========================== short test summary info ============================
FAILED ../../tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape0]
FAILED ../../tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape1]
FAILED ../../tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape2]
FAILED ../../tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-TN-torch.bfloat16-shape3]
FAILED ../../tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape0]
FAILED ../../tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape1]
FAILED ../../tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape2]
FAILED ../../tests/pytorch/test_numerics.py::test_grouped_gemm[True-True-NN-torch.bfloat16-shape3]
====== 8 failed, 2159 passed, 19 skipped, 7 warnings in 146.96s (0:02:26) ======
```

yaox12:
@phu0ngng  If we want to catch up with v2.8, a WAR is to disable accumulate for the cutlass path, and revisit it later, because

1. AFAIK, accumulating to BF16/FP16 is almost not used.
2. Accumulating to FP32 falls back to cuBLAS already.

yaox12:
Further relaxing the threshold solved the mismatch. We think it's because we're comparing two different BF16 results. Maybe it's better to compare with the FP32 baseline. But the unit tests were originally designed for cuBLAS grouped gemm to bit-wise match the non-grouped version. I think the current version is fine as we mentioned that this accumulating pattern is rarely used.

Renamed the env var to use CUTLASS implementation to `NVTE_USE_CUTLASS_GROUPED_GEMM`. And add another env var `NVTE_CUTLASS_GROUPED_GEMM_WARN_FALLBACK`, we only warn falling back when it's set, otherwise there will be too many warnings.

yaox12 (approved):

yaox12:
CI passed except for some irrelevant failures.

PR #2047, opened 2025-08-08
[JAX] Enable TE GEMM custom call for all recipes

phu0ngng:
# Description

With changes introduced in the GH-2023, the TE GEMM custom calls outperform the built-in JAX `dot_general` in most of the parallelism schemes (except TP + Sharding in the hidden dimension of remaining unsharded activations) for all recipes.
Hence, this PR enabled TE GEMM custom calls as the default. 
In addition, a warning is raised when sharding in the hidden dimension of the MLP activation input is detected.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:

- [ ] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [ ] The functionality is complete
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2054, opened 2025-08-11
[JAX] Cleanup the MLP warning for TE GEMM + TP

phu0ngng:
# Description

~~PR GH-2047 introduces a warning message in the layernorm_mlp when the input is sharded in the hidden dimension. However, it checks `logical_axes`, which can be non-empty even when no real sharding is involved, and as a result, mistakenly issues the warnings. This PR fixes such a check by converting the logical axes to the `pspec` before assertion. In addition, checking against whether JAX dot is already used is added.~~

UPDATE - There is no way to check if the hidden dimension is sharded at this point in the code; thus, the warning is removed. Besides, I also clean up the `generate_pspec` as we only call this function in the TE handcoded mesh axes. 

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:


transformer_engine/jax/sharding.py
```diff
@@ -86,37 +86,14 @@ def get_sharding_map_logic_axis_to_mesh_axis():
     return te_logical_axis_to_mesh_axis
 
 
-def generate_pspec(logical_axis_names, with_flax_rules=False, padded=False):
+def generate_pspec(logical_axis_names):
```
jberchtold-nvidia: Afaict this `generate_pspec` is only used in `with_sharding_constraint_by_logical_axes` in the fallback codepath if no Flax logical axes are present. 

Should we make this function private (`_generate_pspec`) and update the docstring to something like this? To make sure people don't use it for Flax logical axes by mistake

```
""" Convert TransformerEngine logical axes (e.g. BATCH_AXES) to a JAX PartitionSpec. Note, this method does not support Flax logical axes.

Args:
    logical_axis_names: TransformerEngine logical axes to convert to a JAX PartitionSpec.
Returns:
    A JAX PartitionSpec with the mesh axes corresponding to the given TransformerEngine logical axis names
"""
```

transformer_engine/jax/sharding.py
```diff
@@ -86,37 +86,14 @@ def get_sharding_map_logic_axis_to_mesh_axis():
...
-        )
-        rules = get_sharding_map_logic_axis_to_mesh_axis()
-    # mesh_axis_names = [rules[name] for name in logical_axis_names]
-    mesh_axis_names = []
-    for name in logical_axis_names:
-        axis_name = rules[name] if name in rules else None
-        mesh_axis_names.append(axis_name)
+    rules = get_sharding_map_logic_axis_to_mesh_axis()
+
+    mesh_axis_names = [rules[name] for name in logical_axis_names]
```
jberchtold-nvidia: Which part of this was the original code before the warning logic was added?
A) `mesh_axis_names = [rules[name] for name in logical_axis_names]`
B) 
```
    mesh_axis_names = []
    for name in logical_axis_names:
        axis_name = rules[name] if name in rules else None
        mesh_axis_names.append(axis_name)
```

These handle an invalid logical axis differently. A) will raise a `KeyError`, B) will implicitly convert it to `None`.

As long as we didn't have this warning logic + (B) long enough that anyone is relying on the implicit conversion to `None`, I think this change is okay and I prefer being explicit with axes instead of implicit replication. I think this warning logic was pretty recent, so I don't think it's made it into an official release. But just wanted to make a note here in case this could cause backwards compatibility issues

phu0ngng: You are right. Let me update it to `rules.get(name)` so that it can return `None` when the key is not found.

jberchtold-nvidia (approved):
LGTM, thanks!

PR #2056, opened 2025-08-11
Add better ordering enforcment to split_overlap_rs gemms.

chaseblock:
This adds a short delay kernel to the split_overlap_rs function, which ensures that the gemms are properly ordered when run with cuda graphs.

# Description

In some situations, such as when running a TP4 config with comm/gemm overlap and cuda graphs, CG will reorder the GEMM kernels in the CommOverlapP2PBase::split_overlap_rs function, but will _not_ reorder the relevant communication. This leads to exposed communication (see below for a simplified illustration). This is because the GEMMs are issued on (in this case) three separate streams, without dependencies that enforce the proper order.

| Stream | T1 | T2 | T3 | T4 | T5 | T6 |
|--------|--------|--------|--------|--------|---|---|
| 1 |  |  | GemmChunk1 | GemmChunk4 |
| 2 | GemmChunk2 |  |  |  |
| 3 |  | GemmChunk3 |  |  |
| 4 | | | | Comm1 | Comm 2 | Comm 3

We do not want to specify dependencies between each of these GEMM kernels, since that would prevent their prologs/epilogs from overlapping. Additionally, CUBLAS does not support LaunchCompletionEvents. This PR takes a different approach (explained below).

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

This PR adds:
1. Dependencies between every other GEMM in split_overlap_rs (i.e., 1->3, 2->4) to partially order these operations.
2. A small delay kernel, which is issued in parallel with the first gemm chunk, and which has negligible runtime. This way, we can create a dependency between this delay and the second GEMM chunk. This ensures that the first gemm chunk actually issues first. The kernel itself is empty, but launching this empty kernel results in a 2-4 us delay.

This changes the kernel ordering in the table above to:

| Stream | T1 | T2 | T3 | T4 |
|--------|--------|--------|--------|--------|
| 1 | GemmChunk1 |  | | GemmChunk4 |
| 2 | | GemmChunk2 |  |  |
| 3 |  | | GemmChunk3 |  |
| 4 | TinyDelayKernel | Comm1 | Comm2 | Comm3 |

# Checklist:


ptrendx:
@chaseblock One question - what is preventing the CG from reordering the GemmChunk1 with the GemmChunk2 in the new scheme? If I understand correctly there is no dependency between those and that reordering could still lead to (albeit smaller than the original) exposed communication.

chaseblock:
> @chaseblock One question - what is preventing the CG from reordering the GemmChunk1 with the GemmChunk2 in the new scheme? If I understand correctly there is no dependency between those and that reordering could still lead to (albeit smaller than the original) exposed communication.

@ptrendx You're correct that there's no dependency between these two. This is necessary in order to allow neighboring gemm kernels to partially overlap during their prolog/epilog. The ordering between GemmChunk1 and GemmChunk2 in this solution is instead enforced via the "TinyDelayKernel". This TinyDelayKernel and GemmChunk1 both become ready to launch at the same time (i.e., share the same dependencies). However, because GemmChunk2 depends on TinyDelayKernel, it cannot launch immediately, whereas GemmChunk1 can, so GemmChunk1 launches first.

I admit that this is a bit of a hack, but since cublas doesn't provide an API to record launchCompletionEvents, I think this is the best we can do for now.

In practice, this seems to fix the problem, and I see perf uplift.

ptrendx (approved):

PR #2050, opened 2025-08-09
[JAX] Trim dist fused attn tests in L1 

KshitijLakhani:
# Description

Reduce the number of L1 distributed fused attn tests so as to reduce CI nightly time outs (which runs L0 and L1 tests)
**_Preliminary tests suggest reduction of ~12 mins (720 seconds)_** 

**L1 test timings (before changes in this PR)**:
```
================================================================================
TEST RUNTIME SUMMARY (grouped by function)
================================================================================
test                                                         |  36x |    5.23s | avg:   0.15s
test_context_parallel_allgather_attn                         | 480x | 1446.77s | avg:   3.01s
test_context_parallel_allgather_attn_shardy                  |  60x |  204.83s | avg:   3.41s
test_context_parallel_ring_attn                              | 1920x | 2367.42s | avg:   1.23s
test_context_parallel_ring_attn_shardy                       |  60x |   90.06s | avg:   1.50s
test_cross_attn                                              |  12x |   57.77s | avg:   4.81s
test_layernorm                                               | 144x |   58.24s | avg:   0.40s
test_layernorm_mlp_grad                                      |  96x |  134.95s | avg:   1.41s
test_layernorm_mlp_grad_shardy                               |  96x |  108.47s | avg:   1.13s
test_layernorm_mlp_layer                                     |  32x |   53.94s | avg:   1.69s
test_layernorm_mlp_layer_fp8                                 |  96x |   28.72s | avg:   0.30s
test_layernorm_mlp_layer_fp8_shardy                          |  96x |   29.58s | avg:   0.31s
test_layernorm_mlp_layer_shardy                              |  32x |   14.97s | avg:   0.47s
test_rmsnorm                                                 |  72x |   20.51s | avg:   0.28s
test_self_attn                                               |  36x |  193.28s | avg:   5.37s
test_self_attn_shardy                                        |   6x |   10.54s | avg:   1.76s
test_softmax                                                 | 288x |  218.98s | avg:   0.76s
test_softmax_gspmd                                           |  24x |   10.80s | avg:   0.45s
================================================================================
TOTAL RUNTIME                                                |      | 5055.06s |
================================================================================
```

**L1 test timings (after changes in this PR)**:
```
================================================================================
TEST RUNTIME SUMMARY (grouped by function)
================================================================================
test                                                         |  12x |    1.61s | avg:   0.13s
test_context_parallel_allgather_attn                         | 320x | 1233.43s | avg:   3.85s
[...]

PR #2060, opened 2025-08-11
[JAX] Add L2_jax_distributed_unittest

jberchtold-nvidia:
# Description

Create a script for `qa/L2_jax_distributed_unittest` so we can move distributed tests from L1 to L2 with a lower cadence and speed up L1.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Create L2 distributed test script
- Update L1 distributed test script to set test level environment variable `NVTE_JAX_UNITTEST_LEVEL="L1"` which was missing

# Checklist:

### DIFF
--- a/tests/jax/test_distributed_layernorm.py
+++ b/tests/jax/test_distributed_layernorm.py
@@ -25,6 +25,7 @@ DTYPES = [jnp.bfloat16, jnp.float32]
 
 NORM_INPUT_SHAPES = {
     "L0": [[64, 64]],
+    "L1": [[64, 64]],
     "L2": [[64, 64]],
 }
 
--- a/qa/L2_jax_distributed_unittest/test.sh
+++ b/qa/L2_jax_distributed_unittest/test.sh
@@ -8,4 +8,4 @@ set -xe
 : ${XML_LOG_DIR:=/logs}
 mkdir -p "$XML_LOG_DIR"
 
-NVTE_JAX_UNITTEST_LEVEL="L1" python3 -m pytest -c $TE_PATH/tests/jax/pytest.ini -v --junitxml=$XML_LOG_DIR/pytest.xml $TE_PATH/tests/jax/test_distributed_*
+NVTE_JAX_UNITTEST_LEVEL="L2" python3 -m pytest -c $TE_PATH/tests/jax/pytest.ini -v --junitxml=$XML_LOG_DIR/pytest.xml $TE_PATH/tests/jax/test_distributed_*


KshitijLakhani (approved):
LGTM
In preparation for https://github.com/NVIDIA/TransformerEngine/pull/2050
Changes should not affect L1 jax distributed for now. Will only affect (via duplication if L1 and L2 both run) L2 jax distributed, but L2 cadence is higher as discussed offline so safe to merge.

qa/L1_jax_distributed_unittest/test.sh
```diff
@@ -8,4 +8,4 @@ set -xe
 : ${XML_LOG_DIR:=/logs}
 mkdir -p "$XML_LOG_DIR"
 
-python3 -m pytest -c $TE_PATH/tests/jax/pytest.ini -v --junitxml=$XML_LOG_DIR/pytest.xml $TE_PATH/tests/jax/test_distributed_*
+NVTE_JAX_UNITTEST_LEVEL="L1" python3 -m pytest -c $TE_PATH/tests/jax/pytest.ini -v --junitxml=$XML_LOG_DIR/pytest.xml $TE_PATH/tests/jax/test_distributed_*
```
KshitijLakhani, line 11: @jberchtold-nvidia I think this will give an error as `test_distributed_layernorm.py::TestDistributedLayernorm` supports only : 

NORM_INPUT_SHAPES = {
    "L0": [[64, 64]],
    "L2": [[64, 64]],
}

KshitijLakhani (changes requested):

qa/L1_jax_distributed_unittest/test.sh
jberchtold-nvidia: Good catch, thanks! Should be fixed now by adding the corresponding entry to NORM_INPUT_SHAPES for L1, running CI

tests/jax/test_distributed_layernorm.py
```diff
@@ -25,6 +25,7 @@
 
 NORM_INPUT_SHAPES = {
     "L0": [[64, 64]],
+    "L1": [[64, 64]],
```
KshitijLakhani, line 28: So whenever we ran this in the past, I guess because it was run as part of `L1 test.sh` but the env var `NVTE_JAX_UNITTEST_LEVEL="L1"` was not set in the `test.sh` so that meant that it would default to using the values associated with the `L0` key in the dictionary which is `[[64,64]]`.
With this PR, this would just be more explicit and it would use the value associated with the `L1` key in the dictionary, however, the value will be the same so no changes expected.

KshitijLakhani (approved):

PR #2068, opened 2025-08-12
[PyTorch] ONNX export of FP8 Current Scaling

pggPL:
# Description

Adds onnx export support for current scaling recipe.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


transformer_engine/pytorch/onnx_extensions.py
```diff
@@ -166,6 +170,68 @@ def onnx_dequantize_fp8_symbolic(
...
+@onnx_cs_dequantize_fp8_op.register_fake
+def _(tensor: torch.Tensor, _) -> torch.Tensor:
+    return torch.empty(tensor.shape, dtype=torch.float32, device=tensor.device)
+
+
+def onnx_dequantize_fp8_cs_symbolic(
+    tensor: onnxscript.onnx_types.UINT8, scale_inv: onnxscript.onnx_types.TensorType
+) -> onnxscript.onnx_types.TensorType:
+    """Symbolic dequantize with current scaling."""
+    return TRT_FP8DequantizeLinear(tensor, scale_inv)
```
timmoon10: Why do we need a separate dequantize for FP8 current scaling? After an FP8 tensor has been quantized, it doesn't matter whether it was produced with delayed scaling or current scaling or some other weird method.

I see that `fp8_dequantize` uses the scale to dequantize, which only really makes sense for delayed scaling. If we want to keep separate implementations, we should rename the DS impl to `fp8_ds_dequantize` and rename the CS impl to `fp8_dequantize`.

tests/pytorch/test_onnx_export.py
```diff
@@ -592,7 +593,9 @@ def _test_export_layernorm_linear(
                     fname,
                     inp,
                     model,
-                    atol=1e-3,
+                    # For current scaling we use Float8Quantizer in tests + amax computed by hand,
+                    # which has slightly different numerics than Float8CurrentScalingQuantizer.
+                    atol=1e-3 if fp8_recipe.__class__ is not recipe.Float8CurrentScaling else 2e-2,
```
timmoon10, line 599: Shouldn't current scaling have better numerics than delayed scaling?

pggPL: We compare fp8 from TE with fp8 from onnxruntime, not fp8 with high precision. Since onnxruntime will use fp8quantizer it is expected that (float8quantizer vs float8qunatizer) will be closed that (float8CurrentScalingQuantizer vs emulation of current scaling using float8qunatizer.

transformer_engine/pytorch/onnx_extensions.py
pggPL: I changed cs to use the same dequantize as ds. Also, I changed scale to scale_inv.

pggPL:
This PR needs https://github.com/NVIDIA/TransformerEngine/pull/2083 to be merged.

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -177,7 +177,7 @@ def onnx_quantize(self, tensor: torch.Tensor) -> QuantizedTensor:
 
     def onnx_dequantize(self, tensor: QuantizedTensor) -> torch.Tensor:
         """Function using primitives with ONNX defined translations."""
-        out = torch.ops.tex.fp8_dequantize(tensor._data, self.scale.item())
+        out = torch.ops.tex.fp8_dequantize(tensor._data, 1 / self.scale)
```
timmoon10: It would be more robust to use the tensor's scale-inv in case the quantizer's scale changes:
```suggestion
        out = torch.ops.tex.fp8_dequantize(tensor._data, tensor._scale_inv)
```
It also matches the FP8 CS case.

timmoon10 (approved):
LGTM

PR #2069, opened 2025-08-12
[JAX] Manual axis filter in  `with_sharding_constraint`

phu0ngng:
# Description

Maxdiffusion recently [added a `shard_map` wrapper around the TE attention call](https://github.com/AI-Hypercomputer/maxdiffusion/blob/f31d65951479cf202ba039ad95707ec7b262bb53/src/maxdiffusion/models/attention_flax.py#L323), in which we [apply the sharding constraint](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/jax/cpp_extensions/attention.py#L2659) underhood. As a result, users face this runtime error:
```
jax._src.source_info_util.JaxStackTraceBeforeTransformation: ValueError: pspec PartitionSpec(('data', 'fsdp', 'tensor'), None) contains a manual axes ('data', 'fsdp', 'tensor') of mesh
 which is not allowed. If you are using a with_sharding_constraint under a shard_map, only use the mesh axis in PartitionSpec which are not manual.
 ```
This PR adds a mechanism to check and remove the Manual axis (which appears when using the `shard_map`) before applying the sharding constraints. 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### CODE REVIEW

File: transformer_engine/jax/sharding.py [hunk 1]
```diff
@@ -122,16 +122,29 @@ def generate_pspec(logical_axis_names, with_flax_rules=False, padded=False):
 
 def with_sharding_constraint(x: jnp.array, pspec: PartitionSpec):
     """
-    A wrapper function to jax.lax.with_sharding_constraint to
-    support the case that Mesh is empty.
+    A wrapper function to jax.lax.with_sharding_constraint
+        1. Does nothing if mesh is empty.
+        2. If all mesh axes are manual axes, replaces pspec with all Nones.
```
jberchtold-nvidia, line 127: Will setting manual axes to `None` mean they are replicated? Or does `None` just mean they are ignored and can be handled manually with `shard_map`?

transformer_engine/jax/sharding.py
```diff
@@ -122,16 +122,29 @@ def generate_pspec(logical_axis_names, with_flax_rules=False, padded=False):
...
 
     mesh = _PXLA_THREAD_RESOURCES.env.physical_mesh
     if mesh.empty:
         return x
-    return jax.lax.with_sharding_constraint(x, pspec)
+
+    axis_names = mesh.axis_names
+    # We want to exclude the axes that already used by shard_map and shard_map only sets those in the abstract_mesh, not the physical one
+    manual_axis_names = get_abstract_mesh().manual_axes
+    if len(manual_axis_names) == len(axis_names):
```
jberchtold-nvidia: Why do we need this explicit check that all mesh axes are "manual"? Wouldn't it be covered in the case below
```
cleaned_axis_names = tuple(name if name not in manual_axis_names else None for name in pspec)
```
where this tuple would only contain `None` if all axes in the mesh are manual. Or is there a case I'm missing?

jberchtold-nvidia (commented):
Thanks for making this PR! Overall LGTM, left two questions

transformer_engine/jax/sharding.py
phu0ngng: When users use the `shard_map` wrapper with some axes, the data are sharded along those axes before entering the function, and those axes are marked as manual axes. 
So inside the function, the program only sees those tensors as a single-GPU tensor in those axes; thus, when we apply the sharding constraint hints, we should only apply them for non-manual axes.

jberchtold-nvidia (approved):
LGTM

phu0ngng:
Pipeline #199048961 passed except for the linter check, which is addressed in the latest commit.

jberchtold-nvidia (approved):

PR #2075, opened 2025-08-14
[JAX] Fix for TE GEMM - Always AllGather RHS non-contracting dims with FSDP axis

phu0ngng:
# Description

Previously, _parse_operand_output_specs only triggered an AllGather on RHS non-contracting dimensions when an AllReduce was involved.

For example, in the second GEMM of an MLP with FSDP:
```
output = input(fsdp, None, None) x Weight(None, fsdp)
contracting_dims = ((2,), (0,))
```

With the current logic, since no AllReduce occurs, the FSDP axis on the RHS is not gathered — while it is gathered on the LHS — resulting in incorrect behavior.

This PR updates the logic to always gather RHS non-contracting dimensions that contain the FSDP axis, ensuring consistent and correct handling across operands.
## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/.github/workflows/trigger-ci.yml
+++ b/.github/workflows/trigger-ci.yml
@@ -53,7 +53,8 @@ jobs:
            || github.actor == 'lhb8125'
            || github.actor == 'kunlunl'
            || github.actor == 'pstjohn'
-           || github.actor == 'mk-61'
+           || github.actor == 'vcherepanov-nv'
+           || github.actor == 'tdophung'
          )
     steps:
       - name: Check if comment is issued by authorized person
--- a/transformer_engine/jax/flax/module.py
+++ b/transformer_engine/jax/flax/module.py
@@ -15,6 +15,8 @@ from jax import lax
 from jax import random as jax_random
 from jax.ad_checkpoint import checkpoint_name
 
+from transformer_engine.common import recipe
+
 from ..dense import dense
 
 from ..layernorm import canonicalize_norm_type
@@ -366,7 +368,9 @@ class TransformerEngineBase(nn.Module):  # pylint: disable=too-few-public-method
             ).value
             return QuantizeMeta(scale=scale, amax_history=amax_history)
 
-        if QuantizeConfig.SCALING_MODE == ScalingMode.DELAYED_TENSOR_SCALING:
+        if QuantizeConfig.SCALING_MODE == ScalingMode.DELAYED_TENSOR_SCALING or isinstance(
+            fp8_recipe, recipe.DelayedScaling
+        ):
             x_meta = generate_quantize_meta("x")
             kernel_meta = generate_quantize_meta("kernel")
             grad_meta = generate_quantize_meta("grad")
--- a/transformer_engine/pytorch/fp8.py
+++ b/transformer_engine/pytorch/fp8.py
@@ -64,14 +64,26 @@ def check_fp8_block_scaling_support() -> Tuple[bool, str]:
     return False, "FP8 block scaled GEMM requires Hopper and CUDA >= 12.9."
 
 
+def check_recipe_support(recipe: Recipe) -> None:
+    """Check if the given recipe is supported."""
+    recipe_supported = True
+    unsupported_reason = ""
+    if isinstance(recipe, (DelayedScaling, Float8CurrentScaling)):
+        recipe_supported, unsupported_reason = check_fp8_support()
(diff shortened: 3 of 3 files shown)


transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -476,21 +476,20 @@ def _parse_operand_output_specs(
...
             lhs_cspecs = (None,) * len(lhs_cspecs)
             rhs_cspecs = (None,) * len(rhs_cspecs)
 
-        # Non-batched non-contracting dims of LHS to be unsharded, i.e gather SP dim
-        # The spec for batch_dim in lhs_non_cspecs won't ever appear in the rhs_non_cspecs as
-        # rhs_non_cspecs never has batch-dim. Hence, spec for batch_dim of lhs_non_cspecs won't be
-        # overwrite
+        # Non-contracting dims of RHS always needs to be gathered along the FSDP axis
+        rhs_non_cspecs = tuple(
+            None if spec is not None and "fsdp" in spec else spec for spec in rhs_non_cspecs
```
jberchtold-nvidia: This `"fsdp" in spec` is a string comparison, since `spec` is a string spec of a specific axis, right? Are there multiple names we're trying to match here, e.g. `fsdp_type_1`, `fsdp_type_2`? If so, then I'm onboard with this. If not and we're using `"fsdp" in spec` in a way that would be equivalent to `"fsdp" == spec`, then I'd prefer the latter for clarity. Let me know if I missed something, thanks!

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -476,21 +476,20 @@ def _parse_operand_output_specs(
...
         else:
             # Otherwise, require contracting dims of both operands to be unsharded
             lhs_cspecs = (None,) * len(lhs_cspecs)
             rhs_cspecs = (None,) * len(rhs_cspecs)
 
-        # Non-batched non-contracting dims of LHS to be unsharded, i.e gather SP dim
-        # The spec for batch_dim in lhs_non_cspecs won't ever appear in the rhs_non_cspecs as
-        # rhs_non_cspecs never has batch-dim. Hence, spec for batch_dim of lhs_non_cspecs won't be
-        # overwrite
+        # Non-contracting dims of RHS always needs to be gathered along the FSDP axis
```
jberchtold-nvidia: > Non-contracting dims of RHS always needs to be gathered along the FSDP axis
Is this statement true because in the MLP block fwd and bwd the kernels only ever appear on the right-hand side?

transformer_engine/jax/cpp_extensions/gemm.py
phu0ngng: Yep, I was not sure if users ever somehow have other axes with "fsdp_x", so I implemented a "pattern matching" there.
@mingxu1067 What is your suggestion here?

jberchtold-nvidia: Is there a way to get the global "MeshResource" set by the fp8_autocast context or global_shard_guard within partition? I think this is what users are supposed to use to tell TE what their mesh axes names are

```
@dataclass
class MeshResource:
    """A data container for managing mesh resources in distributed training.

    This class defines the mapping between logical axes and physical mesh axes
    for different types of parallelism in distributed training.

    Attributes:
        dp_resource: Axis name for data parallelism (batch sharding), default is None
        tp_resource: Axis name for tensor parallelism (hidden dimension sharding), default is None
        fsdp_resource: Axis name for full-sharded data parallelism, default is None
        pp_resource: Axis name for pipeline parallelism (layer sharding), default is None
        cp_resource: Axis name for context parallelism (sequence sharding), default is None
    """

    dp_resource: str = None
    tp_resource: str = None
    fsdp_resource: str = None
    pp_resource: str = None
    cp_resource: str = None
```

jberchtold-nvidia: Seems like we can use `global_mesh_resource()` defined in `sharding.py`. Maybe it's best to check `global_mesh_resource().fsdp_resource` first and if no global mesh resource is set then fallback to "fsdp" or raise an error that the mesh resource wasn't set

phu0ngng: This will be implemented in a follow-up PR.

PR #2082, opened 2025-08-19
[JAX] Pass mesh_resource context through VJP and into primitives

jberchtold-nvidia:
# Description

Passes `MeshResource` set by `global_shard_guard` through VJP and into TE primitives.

VJP rules are run in a separate pass for the autograd transformation, so any narrowly-scoped `global_shard_guard` contexts will cause `global_mesh_resource()` to be unavailable instead both the forward and backward VJP definitions. This PR updates our VJPs to explicitly pass the mesh resource through and wrap the body in a new `global_shard_guard` context.

Primitive partitioning (`Primitive.partition`) is run in a later pass where narrowly-scoped `global_shard_guard` contexts will also no longer be active. This PR updates each primitive that relies on the `MeshResource` either directly like GemmPrimitive or indirectly via `all_reduce_...` helper functions in `sharding.py`. The PR updates the primitives to take an explicit `MeshResource` as a static argument which will be captured and available in the `partition` function.

This PR allows `global_shard_guard` to be used at any level, including inside JIT, instead of requiring a single `global_shard_guard` at the top-level of the entire program. This can make TE integrations into larger systems easier to isolate to a few minimal files.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Updates VJPs in `transformer_engine/jax/*.py` to explicitly pass through the mesh resource and re-create the context inside
- Updates primitives in `transformer_engine/jax/cpp_extensions/*.py` to take the mesh resource as a static argument
- Updates `sharding.py` to default the global mesh resource to `None` instead of `MeshResource()` with all fields set to `None`. This allows us to differentiate between an incorrectly missing mesh resource context and a user intentionally setting certain `MeshResource` axes to `None` when they are not supported on their physical mesh.
- Error checking in `sharding.py` for whenever the `global_mesh_resource()` is queried it must be set in a context. Unfortunately, we cannot definitively tell at trace-time that we are intending to run replicated or on a single device, so this is slightly overbearing in that it requires an MeshResource context (all axes can be `None`) even when running single-GPU. The error message instructs the user how to fix this issue, but it's not ideal.

# Checklist:

### DIFF
--- a/transformer_engine/jax/activation.py
+++ b/transformer_engine/jax/activation.py
@@ -16,6 +16,7 @@ from . import cpp_extensions as tex
 
 from .quantize.tensor import ScaledTensor
 from .quantize.quantizer import Quantizer
+from .sharding import global_mesh_resource, global_shard_guard
 
 
 def activation(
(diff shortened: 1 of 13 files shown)


jberchtold-nvidia:
Closing per offline discussion that while this is functional it adds unnecessary complexity to unrelated pieces of code that a single global context would solve

PR #2088, opened 2025-08-19
[JAX] Error checking for mesh resource and update GemmPrimitive to use global_mesh_resource().fsdp_resource

jberchtold-nvidia:
# Description

Updates the global MeshResource to `None` instead of an object with all fields set to None. This allows us to identify a user explicitly using a MeshResource with an axis intentionality set to None, for example if they do not support fsdp in their physical mesh. With this change we can now differentiate this case from a user accidentally forgetting to set the MeshResource properly. In the latter case we now raise an error.

Additionally, we now update the GemmPrimitive to use `global_mesh_resource().fsdp_resource` instead of the string pattern `*fsdp*` now that we can reliably error check that this axis is set properly.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Update `_GLOBAL_MESH_RESOURCE` to default to `None` instead of a MeshResource will all empty fields
- Update `global_mesh_resource()` to assert that `_GLOBAL_MESH_RESOURCE` is to a non-None value
- Update `fp8_autocast` to not implicitly create a MeshResource object with empty axes if None is provided
- Update `GemmPrimitive` to use `global_mesh_resource().fsdp_resource` instead of the string pattern `*fsdp*`

# Checklist:


transformer_engine/jax/sharding.py
```diff
@@ -314,6 +314,11 @@ def global_mesh_resource() -> MeshResource:
     Returns:
         The current MeshResource instance
     """
+    assert _GLOBAL_MESH_RESOURCE is not None, (
+        "Global mesh resource is not set. Please set the MeshResource via a global_shard_guard"
+        " context. If you are not using multiple GPUs, you can use an empty MeshResource by"
+        " wrapping your program in 'with global_shard_guard(MeshResource()):'"
+    )
```
phu0ngng, line 321: Great to have!

Issue #1687, opened 2025-04-15
fp8_model_init fails with MXFP8BlockScaling


MaciejBalaNV:
**Describe the bug**

When `fp8_model_init` is used in combination with `MXFP8BlockScaling`, the TE modules in the network have `requires_grad=False` if we cast the network to `bfloat16` precision. If we set `requires_grad=True` manually, the optimizer step fails, but only when using `FusedAdam` from TE, standard Adam from PyTorch works. Still, `FusedAdam` has multiple important features we need to use.

**Steps/Code to reproduce bug**
On B200 system, with 25.03 NVIDIA PyTorch container (TE version '2.1.0+450146a') we can use this code:

```

import os
import sys
import argparse

import transformer_engine.pytorch as te
from transformer_engine.common.recipe import Format, DelayedScaling, MXFP8BlockScaling

import torch
import torch.distributed as dist
import torch.nn.functional as F
from torch import nn, optim

class SimpleNet(nn.Module):
    def __init__(self, input_size, hidden_size, output_size):
        super(SimpleNet, self).__init__()
        self.fc1 = te.Linear(input_size, hidden_size)
        self.fc2 = te.Linear(hidden_size, output_size)
        self.fc3 = torch.nn.Linear(output_size, output_size)

    def forward(self, x):
        x = F.relu(self.fc1(x))
        x = self.fc2(x)
        x = self.fc3(x)
        return x


def _parse_args(argv=None, namespace=None):
    parser = argparse.ArgumentParser(description="Toy example for debugging fully_shard()")
    parser.add_argument("--input-size", type=int, default=2048, help="Input size for the model")
    parser.add_argument("--hidden-size", type=int, default=2048, help="Hidden layer size")
    parser.add_argument("--output-size", type=int, default=2048, help="Output size for the model")
    parser.add_argument("--batch-size", type=int, default=2048, help="Output size for the model")
    parser.add_argument(
        "--fp8-init", action="store_true", default=False, help="Initialize primary weights in FP8."
    )
    parser.add_argument(
        "--mxfp8", action="store_true", default=False, help="Use MXFP8BlockScaling instead of DelayedScaling."
    )
    parser.add_argument(
        "--te-optim", action="store_true", default=False, help="Use FusedAdam from TE."
    )
    parser.add_argument(
        "--fix-requires-grad", action="store_true", default=False, help="Manually set requires_grad=True in the network."
    )
    parser.add_argument(
[...]

PR #2091, opened 2025-08-20
[PyTorch] Propagate `requires_grad` for dequantize and torch dispatch

ksivaman:
# Description

Converting a TE model using `.to` currently does not retain the `requires_grad` field for parameters, more details in [this comment](https://github.com/NVIDIA/TransformerEngine/issues/1687#issuecomment-3199308629).

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Set `requires_grad` in dequantize extension.
- Set `requires_grad` in `__torch_dispatch__`

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/csrc/extensions/cast.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/cast.cpp
@@ -79,6 +79,7 @@ py::object dequantize(const py::handle &input, transformer_engine::DType otype) 
     nvte_dequantize(input_tensor.data(), out_tensor.data(), at::cuda::getCurrentCUDAStream());
   });
 
+  out.attr("requires_grad_")(input.attr("requires_grad"));
   return out;
 }
 
--- a/transformer_engine/pytorch/tensor/float8_tensor.py
+++ b/transformer_engine/pytorch/tensor/float8_tensor.py
@@ -549,7 +549,7 @@ class Float8Tensor(Float8TensorBase, QuantizedTensor):
             return Float8Tensor(
                 shape=out_shape,
                 dtype=tensor.dtype,
-                requires_grad=False,
+                requires_grad=tensor.requires_grad,
                 data=out_data,
                 fp8_scale_inv=tensor._scale_inv,
                 fp8_dtype=tensor._fp8_dtype,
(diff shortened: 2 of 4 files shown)


transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -79,6 +79,7 @@ py::object dequantize(const py::handle &input, transformer_engine::DType otype)
     nvte_dequantize(input_tensor.data(), out_tensor.data(), at::cuda::getCurrentCUDAStream());
   });
 
+  out.attr("requires_grad_")(input.attr("requires_grad"));
```
timmoon10, line 82: This will fail for the internal tensors since they don't have a `requires_grad` attr, and adding `requires_grad` to the internal tensors doesn't make sense because they don't support autograd. How about we put this logic in the dequantize autograd functions? For FP8, we can change this function:
https://github.com/NVIDIA/TransformerEngine/blob/5cbe6b0b6c4b392e7b78dac9d7667ae538d3df10/transformer_engine/pytorch/tensor/_internal/float8_tensor_base.py#L28
```python
    @staticmethod
    def forward(
        _ctx: Optional[torch.autograd.function.FunctionCtx],  # unused
        tensor: Float8TensorBase,
        dtype: torch.dtype,
    ) -> torch.Tensor:
        te_dtype = torch_to_transformer_engine_dtype[dtype]

        # Make sure FP8 data is in expected format
        if tensor._data is not None:
            out = None
            if tensor._data.numel() == 0:
                out = torch.empty_like(tensor._data, dtype=dtype)
            else:
                out = tex.dequantize(tensor, te_dtype)
            if isinstance(tensor, torch.Tensor) and tensor.requires_grad:
                out.requires_grad_()
            return out

        raise NotImplementedError("Casting back from the transpose not implemented yet!")
```

timmoon10 (changes requested):
Does `__torch_dispatch__` also add the op to the autograd graph? If not, then `requires_grad=False` is more logical because we can't actually use the grads. This will also avoid unnecessary work in the backward pass.

I think it should be sufficient to modify the dequantize autograd function to fix the bug in https://github.com/NVIDIA/TransformerEngine/issues/1687#issuecomment-3199308629.

PR #2098, opened 2025-08-21
[PyTorch] Support bf16+fp8 cudagraph

buptzyb:
# Description

The current `make_graphed_callables` API assumes all callables to be in one precision. However, there are some models that have bf16 at the first or last layers while fp8 in the middle. This PR extends the argument `fp8_enabled` to be a tuple. When it's a tuple, it must have the same length as the callables, indicating each callable is in bf16 or fp8.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Accept `fp8_enabled` argument of `make_graphed_callables` to be a tuple.

# Checklist:


buptzyb:
Hi @timmoon10 , could you help review? Thanks!

transformer_engine/pytorch/graph.py
```diff
@@ -896,8 +896,9 @@ def make_graphed_callables(
 
     FP8-related parameters
     ----------------------
-    fp8_enabled: bool, default = `True`
-                 whether or not to enable fp8
+    fp8_enabled: (tuple of) bool, default = `False`
```
ptrendx, line 899: Changing the default is an API change so we should not do that in a minor release (in case we want to change it, which I am not sure about).

buptzyb: No we don't change the default. The default is always False. The original docstring is wrong.

transformer_engine/pytorch/graph.py
```diff
@@ -944,15 +953,19 @@ def wrap_autocast(block):
...
+                fp8_context = contextlib.nullcontext()
+            else:
+                fp8_context = fp8_autocast(
+                    enabled=True,
+                    calibrating=fp8_calibrating,
+                    fp8_recipe=fp8_recipe,
+                    fp8_group=fp8_group,
+                    _graph=True,
+                )
+            with fp8_context:
```
timmoon10: It would be more logical to keep the `fp8_autocast` even when `fp8_enabled=False`:
```suggestion
            with fp8_autocast(
                enabled=module_uses_fp8.get(id(self), False),
                calibrating=fp8_calibrating,
                fp8_recipe=fp8_recipe,
                fp8_group=fp8_group,
                _graph=True,
            ):
```
Unlikely to have a real impact, but it handles edge cases like if the user does `fp8_enabled=False` and `fp8_calibrating=True`.

buptzyb: I thought about this. But this is how it works in mcore https://github.com/NVIDIA/Megatron-LM/blob/core_r0.13.0/megatron/core/fp8_utils.py#L416-L421

buptzyb:
@timmoon10 @ptrendx could you help review again? Thanks!

timmoon10 (approved):
LGTM, pending CI

PR #2099, opened 2025-08-21
[PyTorch] Let `GroupedLinear` accept MXFP8 input

yaox12:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


ptrendx:
@yaox12 What do you mean by the "padding scaling factor kernels"? Is that swizzle? If so, then we definitely need to optimize them, they should take max a few percent of the time to do the quantize.

yaox12:
> @yaox12 What do you mean by the "padding scaling factor kernels"? Is that swizzle? If so, then we definitely need to optimize them, they should take max a few percent of the time to do the quantize.

No, it's not swizzle. What I'm doing here is trying to quantize (or act and quantize) multiple input tensors as a whole versus splitting them into chunks and then quantize them one by one. Both methods produce exact the same quantized data, but the scaling factors may be different due to padding. So for the "quantize as a whole" way, we need to pad the sf manually.

For example, if we have two input tensors (for two experts) both in shape [64, 128]:
- Previous way:
  - Activation function in BF16 on the concat tensor in shape [128, 128]
  - Split it into two tensors in shape [64, 128], and quantize them one by one, then we get **two quantized tensors** with data in shape [64, 128], rowwise scaling factor in shape [128, 4] (padding row to 128), colwise scaling factor in shape [4, 128].

- With this PR:
  - We do act and quantize fusion on the concat tensor in shape [128, 128], we get **one single quantized tensor** with data in shape [128, 128], rowwise scaling factor in shape [128, 4], colwise scaling factor in shape [4, 128]
  - Then we need to split the whole quantized tensor into two. The quantized data can be split directly, while for the rowwise sf , we first split it into two [64, 4] and **pad them both to [128, 4]**. This is the padding I'm talking about.

transformer_engine/common/util/math.h
```diff
@@ -28,7 +30,7 @@ __device__ inline OType dgelu(const IType val, const Empty&) {
 template <typename OType, typename IType>
 __device__ inline OType sigmoid(const IType val, const Empty&) {
   const float cval = val;
-  return 1.f / (1.f + expf(-cval));
+  return sigmoidf(cval);
```
yaox12, line 38: Unify the implementation with that in `cast_gated_kernels.cuh`.

transformer_engine/common/util/vectorized_pointwise.h
```diff
@@ -531,8 +532,18 @@ __launch_bounds__(unary_kernel_threads) __global__
       const ComputeType gelu_in = static_cast<ComputeType>(input_loader0.separate()[i]);
       const ComputeType gate_in = static_cast<ComputeType>(input_loader1.separate()[i]);
 
-      ComputeType after_dgelu = Dactivation(gelu_in, p) * grad_val * gate_in;
-      ComputeType after_dgate = grad_val * Activation(gelu_in, p);
+      ComputeType act_in, dact_in;
+      if constexpr ((Activation == &silu<fp32, fp32>) && (Dactivation == &dsilu<fp32, fp32>)) {
+        const float s = sigmoidf(gelu_in);
```
yaox12: Unify the implementation with that in cast_gated_kernels.cuh.

nvMelissa:
Hi @yaox12 - Would you please check the issues with this PR? It is failing some CI checks. Thank you!

yaox12:
Ready for review. The CI failures are irrelevant. @nvMelissa @timmoon10 @zhongbozhu

PR #2107, opened 2025-08-25
[PyTorch Debug] Fix issue with negative underflow.

pggPL:
# Description

The issue with negative percentage of underflow was observed. The reason is that we count only `0` in `fp8_tensor._data` tensor, but we need to take into account also `-0`, represented by `128` (`10000000` in binary, only sign bit is 1).

This PR also fixes issue with computation of percentage of underflows on one device - I forgot to add `- (x == 0).sum()`.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


tests/pytorch/debug/test_log.py
```diff
@@ -164,8 +164,7 @@ def test_numerics(fp8_recipe, feature_dirs):
             num_quantizers=3,
         )
 
-        tensor = torch.zeros(1024, 1024).cuda()
-        tensor[0, :] = 1000
+        tensor = torch.randn(1024, 1024).cuda()
```
ptrendx, line 170: Should we add some specific zeros here to test the functionality? Or maybe run it through relu or something?

transformer_engine/debug/features/utils/stats_computation.py
```diff
@@ -211,23 +211,28 @@ def add_underflows_stats(recipe_name: str, columnwise: bool = False):
...
+
     STATS[stat_num] = (
-        lambda x, aux_dict: (
+        lambda x, aux_dict: torch.isin(
             aux_dict[recipe_name].get_data_tensors(
                 rowwise_data=not columnwise, columnwise_data=columnwise
-            )
-            == 0
+            ),
+            zero_values,
```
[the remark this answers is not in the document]
ptrendx: I agree here (although not quite for the stated reason) - consider non distributed data parallel case where the tensors in the same process could be coming from different devices. Then we would need to send the zero_values to the device of the first input to torch.isin.

tests/pytorch/debug/test_log.py
pggPL: ok, I inserted these zeros manually

transformer_engine/debug/features/utils/stats_computation.py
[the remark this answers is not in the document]
pggPL: Fixed.

transformer_engine/debug/features/utils/stats_computation.py
```diff
@@ -199,6 +199,12 @@ def _get(buffers, stat_name):
     ),
 }
 
+FP8_NEGATIVE_ZERO = 128  # represnts -0.0 in fp8
+zero_values = {
+    device: torch.tensor([0, FP8_NEGATIVE_ZERO], device=device)
+    for device in [torch.device(f"cuda:{i}") for i in range(torch.cuda.device_count())]
+}
```
timmoon10: I'm worried that it could be a bad idea for every rank to touch every local GPU at import time. At the very least this will incur some unnecessary initialization.

How about we add a helper function that resembles [`torch.count_nonzero`](https://docs.pytorch.org/docs/stable/generated/torch.count_nonzero.html):
```python
def count_nonzero_fp8(fp8_data: torch.Tensor) -> torch.Tensor:
    fp8_data = fp8_data.view(dtype=torch.uint8)
    zero_vals = torch.tensor([0, 128], device=fp8_data.device, dtype=torch.uint8)
    return fp8_data.numel() - torch.isin(fp8_data, zero_vals).sum()
```
Calling it would look something like:
```python
lambda x, aux_dict: (
    x.count_nonzero()
    - count_nonzero_fp8(aux_dict...)
)
```

pggPL: Done

timmoon10 (approved):
LGTM

PR #2109, opened 2025-08-25
[PyTorch Debug] Skip log test on device if it does not support fp8.

pggPL:
# Description

The log tests need fp8, but it is not checked.


timmoon10 (approved):
LGTM

ptrendx (approved):

PR #2113, opened 2025-08-25
[JAX] Add `tpsp_resource` in the `MeshResource` map

phu0ngng:
# Description

This PR is needed for the upcoming CollectiveGEMM feature, as one needs to be able to verify if the tensor sequence parallelism is enabled instead of pure tensor parallelism.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


examples/jax/encoder/test_multigpu_encoder.py
```diff
@@ -264,7 +264,7 @@ def train_and_evaluate(args):
     ) as mesh, te.fp8_autocast(
         enabled=args.use_fp8,
         fp8_recipe=fp8_recipe,
-        mesh_resource=te.MeshResource(DEVICE_DP_AXIS, None, None, None),
+        mesh_resource=te.MeshResource(dp_resource=DEVICE_DP_AXIS),
```
jberchtold-nvidia, line 267: 👍 thanks for cleaning all these up to be clearer named args!

examples/jax/encoder/test_model_parallel_encoder.py
```diff
@@ -267,7 +267,10 @@ def train_and_evaluate(args):
     ) as mesh, te.fp8_autocast(
         enabled=args.use_fp8,
         fp8_recipe=fp8_recipe,
-        mesh_resource=te.MeshResource(DEVICE_DP_AXIS, DEVICE_TP_AXIS, None, None),
+        mesh_resource=te.MeshResource(
+            dp_resource=DEVICE_DP_AXIS,
+            tpsp_resource=DEVICE_TP_AXIS,
```
jberchtold-nvidia, line 272: Are we re-assigning this TP axis to `tpsp` because this model parallel example is doing closer to TPSP and we're reserving TP usage for JAX's TP that shards layernorm's hidden dim?

transformer_engine/jax/sharding.py
```diff
@@ -43,44 +43,44 @@ def _get_mesh_info(resource: str, mesh: jax.sharding.Mesh):
...
+        BATCH_AXES: gsr.fsdp_resource or gsr.dp_resource,
         SEQLEN_AXES: None,
-        SEQLEN_TP_AXES: gsr.tp_resource,
+        SEQLEN_TP_AXES: gsr.tpsp_resource,
         SEQLEN_CP_AXES: gsr.cp_resource,
-        HEAD_AXES: gsr.tp_resource,
+        HEAD_AXES: gsr.tp_resource or gsr.tpsp_resource,
         HIDDEN_AXES: None,
-        HIDDEN_TP_AXES: gsr.tp_resource,
+        HIDDEN_TP_AXES: gsr.tp_resource or gsr.tpsp_resource,
```
jberchtold-nvidia: I think in most settings, a user would set all MeshResource fields to non-None values, even if they're not using it in the current configuration. Should the checks here be changed from
```
gsr.tp_resource or gsr.tpsp_resource
```
to 
```
gsr.tp_resource if is_mesh_axis_active(gsr.tp_resource) else gsr.tpsp_resource
``` 
where 
```
def is_mesh_axis_active(axis_name: Optional[str]):
    physical_mesh = _PXLA_THREAD_RESOURCES.env.physical_mesh
    return axis_name is not None and physical_mesh.shape[axis_name] > 1
```

transformer_engine/jax/sharding.py
```diff
@@ -303,7 +305,11 @@ def global_shard_guard(resource: MeshResource):
     old_resources = _GLOBAL_MESH_RESOURCE
     try:
         _GLOBAL_MESH_RESOURCE = resource
+        _validate_mesh_resource_configuration()
         yield
+    except Exception as e:
+        _GLOBAL_MESH_RESOURCE = old_resources
+        raise e
```
jberchtold-nvidia: I don't think we need this `except` block here. A try/finally without an except should still raise all errors through it

Tried this out in a small sample online here: https://www.online-python.com/ZztpMArsR1
```

try:
    raise ValueError('test')
finally:
    print('hello')
```
and the ValueError is still being raised at the outer scope. But if this is a different case in this example and I missed something, lmk. Thanks!

transformer_engine/jax/sharding.py
```diff
@@ -351,52 +357,3 @@ def all_reduce_max_along_all_axes_except_PP(x: jnp.array, mesh: jax.sharding.Mes
         if axis != global_mesh_resource().pp_resource:
             x = lax_paral_op(x, jax.lax.pmax, axis, mesh)
     return x
-
-
-# Deprecating Items ---------------------------------------------------------------
```
jberchtold-nvidia, line 356: Thanks for removing unused code! 🎉

jberchtold-nvidia (commented):
Overall LGTM, left a few small comments. Thanks for the PR!

transformer_engine/jax/sharding.py
phu0ngng: Good finding! I updated it accordingly. Thanks

examples/jax/encoder/test_model_parallel_encoder.py
phu0ngng: We update to use `tpsp_resource` as we always do TP + SP under the TP name so far in the tests and examples.
`tp_resource` now can be used for TP or TP + Hidden, but should not be used for TP + SP.

phu0ngng:
\te-ci JAX L1

jberchtold-nvidia (approved):
LGTM, thanks!

jberchtold-nvidia (approved):
LGTM, thanks!

PR #2114, opened 2025-08-25
Revert "[Common] PDL for Quantization Kernels"

jberchtold-nvidia:
Reverts NVIDIA/TransformerEngine#2001

This uses PDL unchecked which can cause issues in the case the surrounding kernels have data dependencies and use PDL

I've been able to trigger incorrect outputs by updating our activation to also use PDL and running unfused act+quant

image


jberchtold-nvidia:
CI has same failures as main branch, so unrelated to this PR

PR #2120, opened 2025-08-26
Adds dst.dtype information in copy_ method of quantized tensors.

zobeideThePlayer:
# Description

Fixes a bug that causes precision issues in mix-precision training.

Current implementation of copy_ method in QuantizedTensor class does not properly pass the dst.dtype information when src is a QuantizedTensor and dst is not. This may cause precision concerns under certain circumstances, for instance: 
(1) main-stream precision is bfloat16
(2) model is initialized with FP8 format
(3) master weights in optimizers are kept at high precision, i.e., float32
(4) when continue training from a checkpoint but optimizer stats are not loaded/provided, master weights must be initialized from model weights

In above conditions and alike, model weights will be dequantized to bfloat16 (the dtype recorded in quantizer object) and then copied to master weight (of float32 precision), where the trailing 16 bits info are lost.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- adds dst.dtype information in the dequantize() function call of copy_ method.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/tensor/quantized_tensor.py
+++ b/transformer_engine/pytorch/tensor/quantized_tensor.py
@@ -427,7 +427,7 @@ class QuantizedTensor(torch.Tensor):
                 dst.quantize_(src)
             else:
                 if isinstance(src, QuantizedTensor):
-                    src = src.dequantize()
+                    src = src.dequantize(dtype=dst.dtype)
                 dst.copy_(src)
             return None
 


transformer_engine/pytorch/tensor/quantized_tensor.py
```diff
@@ -427,7 +427,7 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
                 dst.quantize_(src)
             else:
                 if isinstance(src, QuantizedTensor):
-                    src = src.dequantize()
+                    src = src.dequantize(dtype=dst.dtype)
```
timmoon10: TE doesn't have dequantize kernels to FP64:
```suggestion
                    dtype = dst.dtype
                    if dtype not in (torch.float32, torch.float16, torch.bfloat16):
                        dtype = torch.float32
                    src = src.dequantize(dtype=dtype)
```

ptrendx: That logic should probably be handled inside dequantize though, not here @timmoon10.

ksivaman:
Duplicated by #2673 

@zobeideThePlayer

PR #2123, opened 2025-08-26
FP8 Output Quantization for GEMM

vthumbe1503:
# Description
Support for Unfused 
1. MXFP8/CurrentScaling Output Quantization for FP8 inputs
2. Any kind of Output Quantization for FP16/BF16/FP32 Inputs.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Unfused MXFP8/CurrentScaling Output Quantization for FP8 inputs. For delayed scaling, CUBLAS handles it internally.
- Unfused Output Quantization for FP16/BF16 Inputs
- Unit test to verify these use-cases.
- Remove redundant code from quantizer.cpp

# Checklist:


tests/pytorch/test_sanity.py
```diff
@@ -911,6 +911,54 @@ def test_sanity_fp8_gemm_with_unalignment(N, datatype):
...
+    ],
+)
+@pytest.mark.parametrize(
+    "out_quantizer",
+    [
+        Float8CurrentScalingQuantizer(fp8_dtype=tex.DType.kFloat8E4M3, device="cuda"),
+        MXFP8Quantizer(fp8_dtype=tex.DType.kFloat8E4M3),
+    ],
+)
+def test_sanity_fp8gemm_with_quantization(N, datatype, input_quantizer, out_quantizer):
```
timmoon10: Not that critical, but this is a better fit for `test_numerics.py` since we are checking the numerics.

tests/pytorch/test_sanity.py
```diff
@@ -911,6 +911,54 @@ def test_sanity_fp8_gemm_with_unalignment(N, datatype):
...
+        MXFP8Quantizer(fp8_dtype=tex.DType.kFloat8E4M3),
+    ],
+)
+@pytest.mark.parametrize(
+    "out_quantizer",
+    [
+        Float8CurrentScalingQuantizer(fp8_dtype=tex.DType.kFloat8E4M3, device="cuda"),
+        MXFP8Quantizer(fp8_dtype=tex.DType.kFloat8E4M3),
+    ],
+)
```
timmoon10: Does cuBLAS support the mixed FP8/MXFP8 case? If not, then it would be better to have a single arg for the recipe and assume the input and output quantizer type is the same.

tests/pytorch/test_sanity.py
```diff
@@ -911,6 +911,54 @@ def test_sanity_fp8_gemm_with_unalignment(N, datatype):
...
+    )
+    out, *_ = general_gemm(
+        weight_fp8,
+        inp_fp8,
+        get_workspace(),
+        outp_type,
+        quantization_params=None,
+        bias=None,
+        use_split_accumulator=False,
+    )
```
timmoon10: Not critical, but I prefer if the reference implementation uses a completely different code path. There could be bugs in TE or cuBLAS that cause results to be consistent between FP8 and non-FP8, but different from mathematical ground truth. For an example, this test for the linear op uses FP64 CPU data for the reference impl:
https://github.com/NVIDIA/TransformerEngine/blob/84fa28d2477c2243ab32cc02ba83faebc59a9e6b/tests/pytorch/test_fusible_ops.py#L991

tests/pytorch/test_sanity.py
timmoon10: We should include FP8 delayed scaling as well.

tests/pytorch/test_sanity.py
```diff
@@ -911,6 +911,54 @@ def test_sanity_fp8_gemm_with_unalignment(N, datatype):
     torch.cuda.synchronize()
 
 
+@pytest.mark.skipif(not fp8_available, reason=reason_for_no_fp8)
```
timmoon10: The skip logic should handle the fact that FP8 and MXFP8 have different requirements. For an example:
https://github.com/NVIDIA/TransformerEngine/blob/84fa28d2477c2243ab32cc02ba83faebc59a9e6b/tests/pytorch/test_fusible_ops.py#L64

tests/pytorch/test_sanity.py
```diff
@@ -911,6 +911,54 @@ def test_sanity_fp8_gemm_with_unalignment(N, datatype):
...
+    ],
+)
+def test_sanity_fp8gemm_with_quantization(N, datatype, input_quantizer, out_quantizer):
+    # For MXFP8 and CurrentScaling, below unfused quantization should happen
+    # FP8 input --> cublas GEMM --> BF16 output --> Quantize to FP8 --> fp8 Output
+    offset = 32
+    scratchpad = torch.randn(N, N * N + offset, device="cuda", dtype=datatype)
+    scratchpad_fp8 = input_quantizer(scratchpad)
+    inp_fp8 = torch.reshape(scratchpad_fp8[0][:-offset], (N, N))
+    weight_fp8 = torch.reshape(scratchpad_fp8[0][offset:], (N, N))
```
timmoon10: Subviews and reshaping are not well supported for all formats (it's especially tricky with MXFP8 since we could have to handle the case where the input and output don't have matching block scales). How about we do something like:
```python
recipe_state = te.fp8.RecipeState.create(..., num_quantizers=3)
input_quantizer, weight_quantizer, output_quantizer = recipe_state.make_quantizers()
inp_quantized = input_quantizer(inp)
weight_quantized = weight_quantizer(weight)
... = general_gemm(
  inp_quantized,
  weight_quantized,
  ...,
  quantization_params=output_quantizer,
)
```
Instead of creating quantizers directly (and hackily using one quantizer for both input and weights), this uses a `RecipeState` to create the quantizers:
https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/fp8.py#L860

transformer_engine/pytorch/csrc/extensions/gemm.cpp
```diff
@@ -251,23 +274,24 @@ std::vector<py::object> gemm(py::handle A, bool transa, py::handle B, bool trans
...
     }
     if (bias.has_value()) {
       if (bias->numel() != 0 && grad) {
         bias_grad->zero_();
       }
     }
   }
-
+  std::unique_ptr<Quantizer> my_quantizer = convert_quantizer(quantizer);
+  if (unfused_quantization_needed) my_quantizer->quantize(unquantized_D_tensor, D_tensor);
```
timmoon10: It feels a bit neater only construct the C++ quantizer when needed, especially since the constructor calls Python:
```suggestion
  if (unfused_quantization_needed) {
    convert_quantizer(quantizer)->quantize(unquantized_D_tensor, D_tensor);
  }
```

PR #2125, opened 2025-08-27
Fix memory overhead of linear layer when all gather from sequence parallel

yuzhongw-nvidia:
# Description

`gather_along_first_dim` (all gather from sequence parallel region) in `Linear` and `LayerNormLinear` introduces memory overhead:

1. Some of the tensors are not released immediately, e.g., `grad_output`, `inputmat_total`, and `ln_out_total`.
2. When executing `_post_process_fp8_blockwise_gather`, it will call `Float8BlockwiseQTensorBase._transpose_columnwise_data`, whose old implementation is as follows. It will make the old `_columnwise_data` become unreachable and unable to be released.
```
self._columnwise_data = tex.fp8_transpose(
    self._columnwise_data, self._fp8_dtype, out=None
)
````

These two bugs are harmful especially for long context training, because the tensor gathered from SP region will be very large and the overhead is unacceptable. For example, in the MoE layer of DSV3, there will be 6 tensors that were not released in time; and in the dense layer of DSV3, there will be **18 tensors** that were not released in time. 

Suppose `seq_len=65536`, there will be at least `18 * 65536 * 7168 = 7.875 GB` memory allocation overhead, and this overhead will increase proportionally with the sequence length. Furthermore, this usually leads to much more overhead in device memory usage. Still take 64K context length as an example, the device memory usage **decreases from 59187 MB to 40777 MB** for DeepSeek-V3-TP16PP8EP32VPP4CP1-full-recompute with the fix. With this huge improvement, we could further turn into selective recomputation to greatly improve the E2E performance (267 TFLOPS -> 307 TFLOPS).

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### CODE REVIEW

**Comment by zhongbozhu** [comment]
Thanks! It does seem a problem if the tensor GC will happen after a long period of time, while enforcing GC directly brings too much overhead. CC @timmoon10 to also take a look at this PR.

File: transformer_engine/pytorch/module/linear.py [hunk 1]
```diff
@@ -881,6 +886,15 @@ def wgrad_gemm(
                     # Deallocate input tensor if permitted
                     if ctx.owns_input:
                         clear_tensor_data(inputmat_total)
+                    elif ctx.backward_input_needs_gather and not ctx.ub_bulk_dgrad:
```
zhongbozhu: why do we also need `and not ub_bulk_dgrad`? 

EDIT: was it because inputmat_total will be a pointer to the UB memory, not owning its own memory?

transformer_engine/pytorch/module/linear.py
```diff
@@ -881,6 +886,15 @@ def wgrad_gemm(
...
+                        clear_tensor_data(inputmat_total)
+
+                    if (
+                        ctx.parallel_mode == "row"
+                        and ctx.sequence_parallel
+                        and not ctx.ub_overlap_ag
```
zhongbozhu: same question like above

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -215,8 +215,10 @@ def forward(
 
         # Store unquantized layer norm output if we need to return it
         ln_out_return = None
+        ln_out_return_is_ln_out = False
```
zhongbozhu: Nit: better naming. I am thinking `should_free_ln_out` and `should_free_ln_out_total` as two separate flags. When there is no sequence parallel, these two should be the same. Otherwise, it depends on whether we return the layernorm output and whether wgrad is needed.

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -353,8 +356,11 @@ def forward(
 
         # Deallocate GEMM input tensor if no longer needed
         if not weight.requires_grad and not return_layernorm_output:
-            ln_out = ln_out_total = None
             clear_tensor_data(ln_out, ln_out_total)
+            ln_out = ln_out_total = None
+        elif ln_out_total is not ln_out_return and not ub_overlap_ag_fprop:
```
zhongbozhu: same question, propose for better naming and I am also wondering why UB is involved in this logic. Was it because when we use UB, ln_out is actually pointing to the user buffer, which has only one copy for entire model so it shouldn't be removed? Then let's leave some comments about it.

transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py
```diff
@@ -349,9 +349,12 @@ def _create_columnwise(self):
     def _transpose_columnwise_data(self):
         """Plainly transpose the columnwise data and scale inv."""
         if self._columnwise_data is not None:
+            _old_data = self._columnwise_data
             self._columnwise_data = tex.fp8_transpose(
                 self._columnwise_data, self._fp8_dtype, out=None
             )
+            _old_data.data = _empty_tensor()
+            del _old_data
```
timmoon10, line 359: Shouldn't Python refcounting deallocate the old data automatically? If not, then there's a larger bug in [`tex.fp8_transpose`](https://github.com/NVIDIA/TransformerEngine/blob/62a57dd45ad8ec02943214059917ff94b644ae35/transformer_engine/pytorch/csrc/extensions/transpose.cpp#L18) that we should fix. The code will become unmanageable if we have to apply this kind of trick everywhere we call a C++ extension.

I see that we return `at::Tensor` instead of Pybind11 objects. I wonder if it is not properly handling the refcount.

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -894,6 +901,10 @@ def wgrad_gemm(
                     # Deallocate input tensor if permitted
                     if not ctx.return_layernorm_output:
                         clear_tensor_data(ln_out_total)
+                    elif (
+                        (ln_out_total is not ln_out) or not ctx.ln_out_return_is_ln_out
+                    ) and not ctx.ub_overlap_ag:
+                        clear_tensor_data(ln_out_total)
```
timmoon10: Two issues:
- In the backward pass, `ub_overlap_ag` is for the grad output all-gather. We should use `ub_bulk_dgrad` for the input all-gather.
- Isn't `ln_out_return_is_ln_out` redundant? We should double check, but I think the correct logic is:
```suggestion
                    if not ctx.return_layernorm_output and not ctx.return_layernorm_output_gathered:
                        clear_tensor_data(ln_out)
                    elif ctx.return_layernorm_output_gathered and ctx.ln_out_needs_gather:
                        clear_tensor_data(ln_out)
                    if ctx.ln_out_needs_gather and not ctx.ub_bulk_dgrad:
                        clear_tensor_data(ln_out_total)
```

transformer_engine/pytorch/module/layernorm_linear.py
timmoon10: I suspect `ln_out_return_is_ln_out` might be redundant.

transformer_engine/pytorch/module/linear.py
timmoon10: Was this needed to fix an error? Clearing the tensors should not affect the underlying UB buffers, so I wouldn't expect any problems. I prefer if we have the same logic for the NCCL and UB cases as much as possible since it's easier to reason about.

timmoon10 (changes requested):

transformer_engine/pytorch/module/linear.py
yuzhongw-nvidia: > why do we also need `and not ub_bulk_dgrad`?
> 
> EDIT: was it because inputmat_total will be a pointer to the UB memory, not owning its own memory?

Yes, I think we are going to clear the tensor only if it is gathered NOT using the UB buffer.

Do you think this condition is not necessary (because it may not influence the global buffer), and could we remove it?

yuzhongw-nvidia: > Was this needed to fix an error? Clearing the tensors should not affect the underlying UB buffers, so I wouldn't expect any problems. I prefer if we have the same logic for the NCCL and UB cases as much as possible since it's easier to reason about.

If `ctx.backward_input_needs_gather` and NOT `ctx.ub_bulk_dgrad`, we directly call `torch.distributed.all_gather_into_tensor` to gather the tensor from SP region, instead of using the UB buffers, so the memory is newly allocated. If we do not clear it, Python GC would happen after a long period of time, thus bringing the memory overhead.

I'm sorry, but could you please explain more about
> I prefer if we have the same logic for the NCCL and UB cases as much as possible since it's easier to reason about.

, because I do not understand what it means. Thanks!

transformer_engine/pytorch/module/layernorm_linear.py
yuzhongw-nvidia: Thanks for your suggestion. I think you are right. Let me take a test and apply the change if no problem.

transformer_engine/pytorch/module/layernorm_linear.py
yuzhongw-nvidia: I think Tim is right. We do not need this variable. Let me remove it.

transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py
yuzhongw-nvidia: It is a very good question that also confuses me. I have a similar question: why we need to use `clear_tensor_data` to release tensors that unused, but not let Python GC to deallocate them? I guess the root cause may be similar and my current fix is a similar trick with `clear_tensor_data`.

I'm not very familiar with the code and deeper implementation. Could you please share this problem with some TE experts to help solve it? Or do you think we could merge this PR first and try to find out the root cause later, because it is a little bit emergent for the runnability and perf of DSV3 / MLA long context training.

timmoon10: We call `clear_tensor_data` to manually deallocate memory before Python GC. A good example is here in the `LayerNormLinear` backward:
https://github.com/NVIDIA/TransformerEngine/blob/5f2b83100c75fb633ef416d3685efb5e23062f5c/transformer_engine/pytorch/module/layernorm_linear.py#L896
The forward GEMM input tensor is stored within the autograd `ctx`, so GC will not deallocate until after the backward has finished. However, we don't need this buffer after the wgrad GEMM and ideally it would be reused for the LayerNorm grad.

transformer_engine/pytorch/module/linear.py
```diff
@@ -317,6 +317,11 @@ def forward(
         # Finished forward GEMM...
         # ------------------------------------------------------
 
+        # Deallocate GEMM input tensor if no longer needed
+        if with_input_all_gather_nccl:
+            clear_tensor_data(inputmat_total)
+            inputmat_total = None
+
```
timmoon10: Will this have an effect? The only benefit from deallocating the input tensor is that the buffer might be reused for the tensor-parallel reduce-scatter on the output. However, if we've done a tensor-parallel all-gather on the input, we won't do a reduce-scatter.

transformer_engine/pytorch/module/linear.py
timmoon10: The code is already very complicated and hard to understand. We do one thing when using NCCL and another thing when using UB, and bugs keep creeping in because we do something in one case and forget to do it in the other. As a matter of design, it's usually better to write the code in a way that does not add more branching logic if it's not needed. Even if it saves some unnecessary work, the extra complexity is usually not worth it.

timmoon10: @zhongbozhu and I are asking if we can change this to:
```suggestion
                    elif ctx.backward_input_needs_gather:
```
It won't save memory if it's a UB buffer, but as long as it doesn't cause bugs it's fine.

transformer_engine/pytorch/module/linear.py
yuzhongw-nvidia: Yes, it will have an effect. I have just confirmed this by an experiment, and if I remove this line, the inputmat_total will be in the memory for a very long time.

I think that's the most serious problem: the data of FP8 tensors created by make_empty may not be released by Python GC as expected. Do you have any idea about it?

transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py
yuzhongw-nvidia: As I described in [this thread](https://github.com/NVIDIA/TransformerEngine/pull/2125/files#r2336225443), I think it is not the business of C++ extension, but the business of `Float8BlockQuantizer.make_empty` (in this case, the unreleased tensor is also created by `make_empty`, and the C++ extension is only creating the new tensor). In other words, I believe the issue is: tensors created by `Float8BlockQuantizer.make_empty` can only be released by manually deleting `_columnwise_data.data` and `_rowwise_data.data`. (I believe that's why we need to delete the `_columnwise_data.data` and `_rowwise_data.data` in `clear_tensor_data`, but not to delete `_columnwise_data` and `_rowwise_data`.)

transformer_engine/pytorch/module/linear.py
yuzhongw-nvidia: I got your point. Thanks for your advice. Changes done.

transformer_engine/pytorch/module/linear.py
yuzhongw-nvidia: Changes done.

transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py
timmoon10, line 359:
```suggestion
            # TODO(yuzhongw, tmoon): Figure out why _old_data is not automatically
            # deallocated by GC. Manually deallocating is a temporary hack.
            _old_data = self._columnwise_data
            self._columnwise_data = tex.fp8_transpose(
                self._columnwise_data, self._fp8_dtype, out=None
            )
            _old_data.data = _empty_tensor()
            del _old_data
```

transformer_engine/pytorch/module/linear.py
```diff
@@ -317,6 +317,11 @@ def forward(
         # Finished forward GEMM...
         # ------------------------------------------------------
 
+        # Deallocate GEMM input tensor if no longer needed
```
timmoon10, line 320:
```suggestion
        # Deallocate GEMM input tensor if no longer needed
        # TODO(yuzhongw, tmoon): Figure out why inputmat_total is not automatically
        # deallocated by GC. Manually deallocating is a temporary hack.
```

PR #2136, opened 2025-08-28
[JAX] NoScaleTensor wrapper for non-quantized data

jberchtold-nvidia:
# Description

To unify our API between quantized and non-quantized codepaths, as well as supporting high-precision ops to output an amax in use for current scaling, this PR creates a NoScaleTensor wrapper around jnp.ndarray which can also optionally store an amax. This change also restructures the class hierarchy in tensor.py so this new class and the existing ScaledTensor classes all inherit from a new AbstractBaseTensor class with a shared interface.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Introduce new AbstractBaseTensor class that is a base class and interface for the existing ScaledTensor classes as well as the new NoScaleTensor class
- Create NoScaleTensor class which can store non-quantized data along with an optional amax
- Update scaling_modes.py to implement a NoScalingModeMetadataImpl for high-precision which doesn't have scales and always uses a ROWWISE layout
- Simplify VJPs to remove special-case logic that was previously needed for higher-precision data
- Fix issue with TE GEMM and differing non-quantized dtypes (e.g. lhs=bf16, rhs=fp32)
    - Add an assert in TE GEMM primitive to ensure both inputs are the same dtype when using non-fp8 inputs
    - When quantization is disabled, cast kernels to input dtype in VJPs. This casting was previously only present in Flax layers, so could trigger TE GEMM errors when using directly via the VJP.

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/activation.py
+++ b/transformer_engine/jax/cpp_extensions/activation.py
@@ -29,7 +29,7 @@ from .misc import (
 )
 from .quantization import _jax_dbias, _quantize_dbias_impl
 from ..sharding import all_reduce_max_along_all_axes_except_PP, all_reduce_sum_along_dp_fsdp
-from ..quantize import ScaledTensor, ScaledTensorFactory, HighPrecisionTensor
+from ..quantize import ScaledTensor, ScaledTensorFactory, NoScaleTensor
 from ..quantize import (
     Quantizer,
     QuantizeLayout,
@@ -1033,7 +1033,7 @@ def act_lu(
             is_outer=True,
         )
         out = out.reshape(output_shape)
-        out = HighPrecisionTensor(
+        out = NoScaleTensor(
             data=out,
             amax=None,
         )
@@ -1142,7 +1142,7 @@ def quantize_dact_dbias(
         if is_dbias:
             dbias = _jax_dbias(output, dtype=x.dtype, flatten_axis=-2)
 
-        output = HighPrecisionTensor(
+        output = NoScaleTensor(
             data=output,
             amax=None,
         )
--- a/transformer_engine/jax/cpp_extensions/activation.py
+++ b/transformer_engine/jax/cpp_extensions/activation.py
@@ -922,7 +922,7 @@ class DActLuQuantizePrimitive(BaseDActLuDBiasQuantizePrimitive):
     """Subclass of BaseDActLuDBiasQuantizePrimitive for fused activation quantization without dbias. No change in functionality from the base primitive but named differently for use in more granular disabling of primitives via NVTE_JAX_CUSTOM_CALLS."""
 
 
-def _jax_act_lu(inputs, activation_type, quantizer=None) -> Union[jnp.ndarray, ScaledTensor]:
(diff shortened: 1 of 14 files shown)


transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -558,7 +561,9 @@ def _jax_quantize_dbias(
     flatten_axis: int = -1,
 ):
     if quantizer is None:
-        return x, None
```
jberchtold-nvidia, line 561: It seems like before this PR, our `_jax_quantize_dbias` returned `None` for dbias when no quantizer was specific. I've kept the behavior unchanged in this PR, but I think we should be doing dbias here, right? If so, I can fix it in this PR

transformer_engine/jax/quantize/tensor.py
```diff
@@ -93,9 +90,78 @@ def apply_sharding_constraint_by_logical_axes(self, logical_axis_names: Tuple[st
...
+        """This is a no-op for a higher-precision tensor so this simply returns the tensor's data."""
+        return self.data
+
+    def get_tensor(self, usage: TensorUsage):
+        """Returns the tensor based on the tensor usage."""
+        q_layout = ScalingMode.NO_SCALING.get_quantize_layout(usage)
+        assert (
+            q_layout == QuantizeLayout.ROWWISE
+        ), "Only ROWWISE layout is supported for HighPrecisionTensor"
+        return self
```
phu0ngng, line 134: Perhaps adding an assertion that the current tensor is rowwise.

jberchtold-nvidia: HighPrecisionTensor (soon to be named NoScaleTensor) only has data and amax. It doesn't store a layout since it's assumed to always be rowwise.

I've added the assert here for q_layout == ROWWISE just to ensure the implementation here in tensor.py and the scaling mode impl in scaling_modes.py are validly assuming the layout is rowwise. Maybe I should also clarify this in the class docstring?

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -941,11 +941,11 @@ def _jax_act_lu(inputs, activation_type, quantizer=None) -> Union[jnp.ndarray, S
     x = jnp.squeeze(x, axis=-2)
     if quantizer:
         return quantizer.quantize(x, flatten_axis=-1)
-    return x
+    return NoScaleTensor(data=x, amax=None)
```
phu0ngng, line 944: I wonder if `amax=None` will cause any issue as we register amax as a traceable object in the pytree class. 
If it does at one point, then we could have `amax=jnp.empty()` as the default.

transformer_engine/jax/cpp_extensions/quantization.py
phu0ngng: Yes, we should return a valid dbias here.

phu0ngng (approved):
LGTM. Thanks!

phu0ngng (approved):

PR #2139, opened 2025-08-29
[PyTorch] fix cross entropy vanishing gradients

casper-hansen:
# Description

CE is broken causing vanishing gradients. This deploys a fix to compute correct gradients and avoids vanishing gradients. 

Fix https://github.com/NVIDIA/Megatron-LM/issues/1750

Validated with MS-SWIFT Megatron training of Qwen3 30B A3B with 65k context length on 8x H200 CC @Jintao-Huang

image

## Type of change

## Changes

- Fix element_mul_kernel to use the correct pointer for grad_output

# Checklist:


transformer_engine/pytorch/triton/cross_entropy.py
```diff
@@ -252,6 +252,7 @@ def element_mul_kernel(
     X_ptr += program_id * X_stride
 
     # Load the gradient output value
+    grad_output_ptr += program_id
```
timmoon10: This change assumes that `grad_output_ptr` is from a tensor with outer dims that match `grad_output_ptr`. This is valid when the loss outputs are not reduced (equivalent to [`torch.nn.CrossEntropyLoss`](https://docs.pytorch.org/docs/stable/generated/torch.nn.CrossEntropyLoss.html#crossentropyloss) with `reduction='none'`). However, we also need to handle the case where `grad_output_ptr` is a scalar (equivalent to `torch.nn.CrossEntropyLoss` with `reduction='mean'`).

timmoon10 (commented):
This error slipped through the tests because we have an optimization to skip the backward kernel when the grad output is 1:
https://github.com/NVIDIA/TransformerEngine/blob/f378eaf2899f1148c68b567f62595e940556da7f/transformer_engine/pytorch/triton/cross_entropy.py#L350-L352
We should modify the tests to avoid the shortcut, perhaps by multiplying the losses by 2 before launching the backward pass:
https://github.com/NVIDIA/TransformerEngine/blob/f378eaf2899f1148c68b567f62595e940556da7f/tests/pytorch/test_parallel_cross_entropy.py#L60-L64

casper-hansen:
Hi @timmoon10, thanks for your review + fixes! I appreciate your efforts here as I am stuck waiting for a release before I continue my finetuning efforts with Megatron + TE (as I am sure others are as well).

timmoon10 (approved):

Issue #2053, opened 2025-08-11
[PyTorch] High CPU overhead on Grace systems


Marks101:
Hello everyone,

we have been investigating low compute throughput for LLM trainings on GH200 and GB200 systems in comparison to H100 DGX (Intel Xeon Platinum 8480C). This already affects 1 GPU trainings and we tracked this down to CPU overhead of Transformer Engine (TE) modules on Grace systems.

To showcase this, I appended a small benchmarking script, that evaluates GPU kernel runtime and CPU overhead of the Linear module from PyTorch and Transformer Engine on GB200, GH200 and H100 DGX.
Our benchmarking does multiple runs and we evaluate the average time per Linear forward pass both on CPU and GPU. In the following run we look at BF16 precision with a hidden size of 2048 and 8192 tokens, these values are typical for 7B LLM trainings:

|       | backend | CPU runtime [µs] | GPU runtime [µs] | estimated throughput [TFLOP/s] |
|-------|---------|------------------|------------------|--------------------------------|
| GB200 | torch   | 37.08            | 41.79            | 1644.22                        |
| GB200 | TE      | 99.66            | 99.73            | 689.08                         |
| GH200 | torch   | 33.17            | 91.64            | 749.86                         |
| GH200 | TE      | 90.53            | 92.14            | 745.80                         |
| H100 DGX | torch   | 13.48            | 100.06           | 686.79                         |
| H100 DGX | TE      | 49.19            | 101.88           | 674.51                         |

So what we see is that:
- The torch Linear has significantly lower CPU overhead in comparison to TE. This is probably expected as TE offers a lot more features.
- On Grace systems, the CPU overhead is almost a factor of two higher in comparison to H100 DGX

This two factors together can severely limit the throughput as execution becomes CPU bound, see for example the GB200 numbers with TE. Although the GPU has 2.5x the FLOP/s, the scenario does not make use of it. Even the GH200 run with TE is very close to be severly CPU bound, especially if you consider that actual training code adds additional CPU instructions.

Generally, increasing the number of tokens minimizes this issue as it increases GPU runtime but CPU overhead is independent of the problem size. But for many scenarios other constraints like memory, global batch size and pipelining considerations limit the number of tokens per forward/backward pass and as such are infeasible. Putting this all together the CPU overhead can severly limit the compute throughput during training.

[...]

PR #2146, opened 2025-09-02
[PyTorch] CPU Overhead Micro-optimizations

zhongbozhu:
# Description

Motivation: https://github.com/NVIDIA/TransformerEngine/issues/2053 

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/transformer_engine.cpp
```diff
@@ -641,11 +641,15 @@ void nvte_destroy_quantization_config(NVTEQuantizationConfig config) {
 }
 
 int nvte_is_non_tn_fp8_gemm_supported() {
```
timmoon10, line 719: This doesn't handle the case where we have multiple GPUs with different archs. We could add an arg for the device ID, but that just pushes the CPU overhead problem somewhere else.

transformer_engine/pytorch/module/linear.py
```diff
@@ -1387,9 +1388,14 @@ def forward(
             ).is_fp8_ubuf():
                 fp8_grad = True
 
-        with torch.cuda.device(
-            getattr(self, list(self.named_parameters())[0][0]).device
-        ), self.prepare_forward(
+        if is_first_microbatch is None or is_first_microbatch:
```
timmoon10: How do we assume we can skip setting the device if `is_first_microbatch=False`?

zhongbozhu: I assume that the device won't change across microbatches in a global batch.

transformer_engine/common/transformer_engine.cpp
zhongbozhu: Yeah, but we didn't really support this case anyway?

transformer_engine/pytorch/module/linear.py
zhongbozhu: Since in a CPU bounded fwd only case, skipping set device for every single forward pass could account for 10% perf difference.

transformer_engine/common/transformer_engine.cpp
zhongbozhu: For topology like 1 CPU 8/4GPUs with homogenous GPU arch, we can cache the TN layout check.

transformer_engine/pytorch/module/linear.py
timmoon10: This approach is really ad hoc. Personally, I think it would be better to not to support the multi-device case (basically revert https://github.com/NVIDIA/TransformerEngine/pull/1974) than to have inconsistent multi-device support.

zhongbozhu: I agree with it, but not sure if there are any potential impact for customers using this feature?

nvMelissa:
@zhongbozhu - this is out of date with base branch . Woudl you please retry with latest ?

ksivaman:
Closing in favor of #2377  and #2400

PR #2150, opened 2025-09-04
[Common][PyTorch][Rework] PDL for Quantization

yaox12:
# Description

Rework PDL for quantization in #2001 and #2066.

Add two quantization configs
- `pdl_sync`: Add `cudaGridDependencySynchronize` to the first quantization kernel, to make sure the previous unknown kernel has flushed results to the global memory. The following kernels are launched to the same stream, and there's no data dependency between them, so they don't need this sync.
- `pdl_trigger`: Add `cudaTriggerProgrammaticLaunchCompletion` to all but the last quantization kernel, so it won't trigger the next unknown kernel.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -543,26 +571,29 @@ void quantize_transpose_square_blockwise(const SimpleTensor& input, SimpleTensor
...
-                block_scaled_cast_transpose_kernel<kReturnTranspose, float, InputType, OutputType>
-                    <<<grid, THREADS_PER_BLOCK, 0, stream>>>(
-                        reinterpret_cast<const InputType*>(input.dptr),
-                        reinterpret_cast<OutputType*>(output.dptr),
-                        reinterpret_cast<OutputType*>(output_t.dptr),
-                        reinterpret_cast<float*>(scale_inv.dptr),
-                        reinterpret_cast<float*>(scale_inv_t.dptr), row_length, num_rows,
-                        scale_stride_x, scale_stride_y, scale_t_stride_x, scale_t_stride_y, epsilon,
-                        tensor_map_output_trans, pow_2_scale, noop_ptr);
+                NVTE_CHECK_CUDA(cudaLaunchKernelEx(
```
jberchtold-nvidia, line 593: When PDL is not set on the launch config attributes, this behaves identical to the `kernel<<<...>>>` style launch, right?

jberchtold-nvidia:
My understanding is that we have control of both edges between kernels, we can modify the launch of the current kernel with `cudaLaunchKernelEx` and we can modify if the next kernel can launch early via `cudaTriggerProgrammaticLaunchCompletion` inside the current kernel. If this understanding is incorrect, lmk, thanks!

```
kernel1-->kernel2-->kernel3
```

This PR prevents the first kernel in a series of grouped quantize from launching with PDL. The current implementation is blocking kernel1 from launching too early before data is ready, but since the launch and triggering of the next kernel is controlled by the same enable_pdl flag, isn't it also preventing kernel2 from launching early since there won't be a `cudaTriggerProgrammaticLaunchCompletion();` call in kernel1? So we don't get the benefit until kernel3

On the other hand, kernel3 is okay to launch early but do we want to prevent it from calling `cudaTriggerProgrammaticLaunchCompletion()` internally so any unknown subsequent kernels wouldn't launch early. If the subsequent kernels did launch early it'd be a bug in them too, but just to be safe we can avoid the final call to `cudaTriggerProgrammaticLaunchCompletion()` in this sequence of kernels (the call in kernel3) without any performance penalty, right?

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
yaox12: Yes.

yaox12:
> My understanding is that we have control of both edges between kernels, we can modify the launch of the current kernel with `cudaLaunchKernelEx` and we can modify if the next kernel can launch early via `cudaTriggerProgrammaticLaunchCompletion` inside the current kernel. If this understanding is incorrect, lmk, thanks!
> 
> ```
> kernel1-->kernel2-->kernel3
> ```
> 
> This PR prevents the first kernel in a series of grouped quantize from launching with PDL. The current implementation is blocking kernel1 from launching too early before data is ready, but since the launch and triggering of the next kernel is controlled by the same enable_pdl flag, isn't it also preventing kernel2 from launching early since there won't be a `cudaTriggerProgrammaticLaunchCompletion();` call in kernel1? So we don't get the benefit until kernel3
> 
> On the other hand, kernel3 is okay to launch early but do we want to prevent it from calling `cudaTriggerProgrammaticLaunchCompletion()` internally so any unknown subsequent kernels wouldn't launch early. If the subsequent kernels did launch early it'd be a bug in them too, but just to be safe we can avoid the final call to `cudaTriggerProgrammaticLaunchCompletion()` in this sequence of kernels (the call in kernel3) without any performance penalty, right?

I got your point. We need to handle the trigger from the kernel and the launch attribute separately. That makes sense.

yaox12:
@jberchtold-nvidia Now I use two configs to control the behavior of PDL:
- `pdl_sync`: Add `cudaGridDependencySynchronize` to the first kernel, to make sure the previous unknown kernel has flushed results to the global memory. The following kernels are launched to the same stream, and there's no data dependency between them, so they don't need this sync.
- `pdl_trigger`: Add `cudaTriggerProgrammaticLaunchCompletion` to all but the last kernel, so it won't trigger the next unknown kernel.

At the host side, we always set the `cudaLaunchAttributeProgrammaticStreamSerialization` attribute. The behavior still depends the kernel side `cudaGridDependencySynchronize` / `cudaTriggerProgrammaticLaunchCompletion`. If there's no sync nor trigger, the kernel would behave as a normal one without PDL.

jberchtold-nvidia:
> @jberchtold-nvidia Now I use two configs to control the behavior of PDL:
> 
> * `pdl_sync`: Add `cudaGridDependencySynchronize` to the first kernel, to make sure the previous unknown kernel has flushed results to the global memory. The following kernels are launched to the same stream, and there's no data dependency between them, so they don't need this sync.
> * `pdl_trigger`: Add `cudaTriggerProgrammaticLaunchCompletion` to all but the last kernel, so it won't trigger the next unknown kernel.
> 
> At the host side, we always set the `cudaLaunchAttributeProgrammaticStreamSerialization` attribute. The behavior still depends the kernel side `cudaGridDependencySynchronize` / `cudaTriggerProgrammaticLaunchCompletion`. If there's no sync nor trigger, the kernel would behave as a normal one without PDL.

Thanks! Using cudaGridDependencySynchronize seems like a good idea. The PR LGTM from my side but will defer to Przemek or Tim for final approval since I'm not as familiar in using PDL. Thanks for the PR!

nvMelissa:
@yaox12 @timmoon10 - This PR is has conflicts. I don't know if it is because the PR needs to fixed or problems with the CI. Would you please say what is the next step here to fix the CI issues?

yaox12:
> @yaox12 @timmoon10 - This PR is has conflicts. I don't know if it is because the PR needs to fixed or problems with the CI. Would you please say what is the next step here to fix the CI issues?

I'm working on fixing the conflicts and CI issues.

yaox12:
Ready for review. The CI failures are irrelevant.

yaox12:
Close this PR because
1. We prefer to use a grouped quantize to further reduce the CPU overhead.
2. There're parallel work on optimizing the quantization kernels, which makes the PDL hard to maintain.

PR #2157, opened 2025-09-05
[Common][Pytorch] Add support for the FP8 Block Scaling (ie. Deepseek) recipe on Blackwell

janekb04:
# Description

This PR adds support for the FP8 block scaling (ie. DeepSeek) recipe on Blackwell. It exhibits some changes in behavior compared to Hopper.

Addresses [this discussion](https://github.com/NVIDIA/TransformerEngine/pull/1513#discussion_r2023621571) from #1513.

## Motivation

Currently, the FP8 block scaling recipe works only on Hopper. If you try to use it on Blackwell, the `check_fp8_block_scaling_support` function in `fp8.py` will report that is not supported and an exception will prevent further execution. If `check_fp8_block_scaling_support` is changed to instead check that the architecture is Hopper or newer, the failure occurs in `cublas_gemm` instead. Namely, cuBLASLt does not implement `cublasLtMatmul` with a `CUBLASLT_MATMUL_MATRIX_SCALE_VEC128_32F` or `CUBLASLT_MATMUL_MATRIX_SCALE_BLK128x128_32F` input type on Blackwell.

A possible workaround is to simply switch from using the `Float8BlockScaling` recipe to the `MXFP8BlockScaling` recipe. However, this can result in numerical discrepancies. They occur because of differences in how the two recipes quantize tensors and because the MXFP8 recipe performs more operations in low precision than the block scaling recipe.

## Implementation

This PR emulates only the GEMMs with MXFP8. This is done by converting input `NVTE_BLOCK_SCALING_1D` and `NVTE_BLOCK_SCALING_2D` tensors to `NVTE_MXFP8_1D_SCALING` just before a GEMM. The tensors' main data is not touched at all, only the format of the scaling factors is changed to be compatible with MXFP8.

The FP8 block scaling tensors are created (quantized from higher precision) using `quantize_transpose_vector_blockwise` or `quantize_transpose_square_blockwise` - the same as on Hopper. This means that contrary to simply switching to the MXFP8 recipe, the 1x128 and 128x128 quantization block size is preserved (if MXFP8 was used, the scaling factors could be different, as they would correspond to 1x32 blocks).

To make the tensors valid inputs to the MXFP8 GEMM, their scaling factors are converted from the FP8 block scaling format to the MXFP8 format. I take advantage of the fact that when entering the GEMM, the FP8 Block Scaling tensors are guaranteed to already use the `GEMM_READY` scaling factor format. In case of 2D (128x128) block scaling, the scaling factors are simply "unsqueezed" by a factor of 512 - a single scaling factor for a 128x128 block becomes 512 scaling factors for 512 1x32 blocks constituting the 128x128 block. In case of 1D (1x128) block scaling, every 128 scaling factors corresponding to a 128x128 block are swizzled to match the cuBLASLt MXFP8 GEMM format and "unsqueezed" by a factor of 4 to correspond to the 512 1x32 blocks.

## Limitations

1. The conversion from FP8 block scaling scaling factors to MXFP8 scaling factors is lossless if and only if the original FP8 block scaling scaling factors are powers of 2. This is because the MXFP8 scaling factors are not FP32 (like the FP8 block scaling scaling factors), but rather FP8E8M0. The conversion kernels assume this requirement is met and simply extract the FP32 exponent bits and treat them as FP8E8M0. If either the sign bit or any mantissa bits are set, the results will be incorrect, as the exponent bits are not masked out when performing bit shifts. Masking them out would result in numerical discrepancies in the output anyway due to discarding them.

2. Despite losslessly converting the tensors, the GEMM outputs are not identical between Hopper and Blackwell because the Blackwell MXFP8 cuBLASLt GEMM is implemented differently from the Hopper FP8 block scaling cuBLASLt GEMM. However, the numerical error overall should be smaller compared to simply switching to the MXFP8 recipe.

3. Contrary to FP8 Block Scaling on Hopper, GEMM+GELU fusion is not currently supported as cuBLASLt doesn't support it for MXFP8 (with BF16 output, which the FP8 Block Scaling recipe uses).

## Future optimizations to pursue

* **Take advantage of Blackwell support for FP8 non-TN GEMMs**. Contrary to Hopper, Blackwell supports non-TN GEMMs for FP8. This leads to at least the following two optimizations: 
  1. **Don't create columnwise data for weights**. Because weights are 2D-block-scaled, their quantized columnwise data is simply the transpose of the rowwise data. As such, columnwise data doesn't have to be created in the forward pass, as the rowwise data can be used in the dgad GEMM.
  2. **Don't transpose data after all gather**. Currently, `_post_process_fp8_blockwise_gather` in `distributed.py` transposes columnwise data after all gather. This could be avoided.

* **Use a multi kernel for scaling factor swizzling**. Currently, the support for `te_general_grouped_gemm`, which is needed for MoE, is naive. The tensor conversion function `convert_block_scaling_to_mxfp8_tensor` is simply called in a loop for every tensor. This can result in many small kernel launches, which is inefficient. Instead, a multi kernel approach could be used, similar to how MXFP8 scaling factor swizzling is handled.

* **Use preloading to reduce scoreboard stalls**. Currently, the kernels are memory-latency bound. Before they can start loading data, they have to do a bit of address calculation, which prevents latency hiding. To alleviate this, the kernels could prefetch the entire scaling factor tensor to L2 with `cp.async.bulk.prefetch.global.L2` at the very beginning before calculating which part of the tensor they need to read specifically, as the tensors should be small enough to fit in L2, and they will have to be read in their entirety anyway.

## Performance notes

Regarding the performance of the 1D swizzling kernel:
 * Using warp shuffles (as it is currently being done) is 4.7% faster than using shared memory, according to NCU. I think that using a grid-stride loop would improve this benefit even further if shuffles from the different tiles could be interleaved for latency hiding.
 * It may seem like [the first shuffle](https://github.com/NVIDIA/TransformerEngine/pull/2157/files#diff-2d405e136e61a6bc06cb24e1cc8f3b5002224e6ec4f530722632ddf8ec41c7a5R118) is redundant. After all, couldn't we just calculate `lane_load_idx` before the [scaling factors being loaded](https://github.com/NVIDIA/TransformerEngine/pull/2157/commits/aeafe79ad9b25f918b33cd8a704e7038102d083b#diff-2d405e136e61a6bc06cb24e1cc8f3b5002224e6ec4f530722632ddf8ec41c7a5R102-R109) and index `reinterpret_cast<const uint4*>(warp_src)` with `lane_load_idx` rather than `lane`? We could, and this is what I originally did. However, it turned out that it is more performant to start loading the data as soon as possible, minimizing the amount of computations that have to be done before the first load. As such, I changed the kernel to the current approach in aeafe79ad9b25f918b33cd8a704e7038102d083b. After preloading the tensors to L2, this might no longer be neccesary.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

* Add `nvte_swizzle_block_scaling_to_mxfp8_scaling_factors` core function for swizzling the FP8 block scaling `GEMM_READY` scaling factors to MXFP8 swizzled format.
* Add [`swizzle_block_scaling.cu`](https://github.com/NVIDIA/TransformerEngine/pull/2157/files#diff-2d405e136e61a6bc06cb24e1cc8f3b5002224e6ec4f530722632ddf8ec41c7a5) with two new custom kernels for swizzling the 1D and 2D FP8 block scaling scaling factors.
* On Blackwell, when quantizing tensors to the FP8 Block Scaling format (in [`quantize_transpose_square_blockwise.cu`](https://github.com/NVIDIA/TransformerEngine/pull/2157/files#diff-81c7ad7622c58182023a5e23461eb560f5a955e71e283578d98c305761ba87ad) and [`quantize_transpose_vector_blockwise.cu`](https://github.com/NVIDIA/TransformerEngine/pull/2157/files#diff-b4cb14a1a931997eb5e807790215df350234f16a50d75b944127756eb72ba506)) make sure the power-of-two scaling factors are used. The quantization itself does work with non-power-of-two scaling factors, but the check is done here, as it is the only place where this can be checked. In the GEMM, there is no longer a way to determine whether the tensors use power-of-two scaling factors.
* On Blackwell, convert FP8 Block Scaling tensors to MXFP8 tensors right before the GEMM in `gemm` and `te_general_grouped_gemm` in [`gemm.cpp`](https://github.com/NVIDIA/TransformerEngine/pull/2157/files#diff-d539b1df7e0eefd6a2c177a89655516c7276d94537b60aab4594280c69c495f5).
* Change testing for support of the FP8 Block Scaling recipe to allow in on Blackwell in the Pytorch frontend in [`fp8.py`](https://github.com/NVIDIA/TransformerEngine/pull/2157/files#diff-77a87478142dbcec64bbe6fc9f402be4c4c9aa21e375cca6d352545fa8e87166).
* On Blackwell, skip tests with non power of two scaling factors in [`test_cast_float8blockwise.cu`](https://github.com/NVIDIA/TransformerEngine/pull/2157/files#diff-121b5cdd1c86bdd31f36095c51e0ca28f811c31ff171dd6d7b54c8dc9928ce6f) and [`test_float8_blockwise_scaling_exact.py`](https://github.com/NVIDIA/TransformerEngine/pull/2157/files#diff-7402340086ea3fff020fc2c6f022f9aa10103e0a672108d6da8c12c150bdf69e).
* On Blackwell, skip [`test_float8_blockwise_gemm_exact.py`](https://github.com/NVIDIA/TransformerEngine/pull/2157/files#diff-0ae60513e829423110dc6101b1289feece97d171dcda2148cb1ca5a76bc0d458) because the MXFP8 GEMM has some numerical discrepancies with the FP8 Blockwise Scaling GEMM and is not bitwise the same as these tests expect. Other tests like `test_numerics.py` will check that the GEMMs are within acceptable numerical errors.

# Checklist:


transformer_engine/pytorch/csrc/extensions/gemm.cpp
```diff
@@ -222,6 +226,19 @@ std::vector<py::object> gemm(py::handle A, bool transa, py::handle B, bool trans
     swizzled_scale_inverses_list.emplace_back(
         std::move(swizzle_scaling_factors(B_tensor, !transb)));
 
+    // Emulate the FP8 block scaling recipe with MXFP8 on Blackwell and newer
+    // as it is not natively supported by cublasLt
+    if (fp8_block_scaling && transformer_engine::cuda::sm_arch() > 90) {
```
timmoon10:
```suggestion
    if (fp8_block_scaling && transformer_engine::cuda::sm_arch() >= 100) {
```

transformer_engine/common/transpose/quantize_transpose_vector_blockwise.cu
```diff
@@ -529,6 +530,11 @@ void quantize_transpose_vector_blockwise(const SimpleTensor& input, SimpleTensor
                                          cudaStream_t stream) {
   NVTE_API_CALL(quantize_transpose_vector_blockwise);
 
+  if (transformer_engine::cuda::sm_arch() > 90) {
```
timmoon10:
```suggestion
  if (transformer_engine::cuda::sm_arch() >= 100) {
```

transformer_engine/common/transpose/quantize_transpose_square_blockwise.cu
```diff
@@ -485,6 +486,11 @@ void quantize_transpose_square_blockwise(const SimpleTensor& input, SimpleTensor
   NVTE_API_CALL(quantize_transpose_square_blockwise);
   checkCuDriverContext(stream);
 
+  if (transformer_engine::cuda::sm_arch() > 90) {
```
timmoon10:
```suggestion
  if (transformer_engine::cuda::sm_arch() >= 100) {
```

tests/cpp/operator/test_cast_float8blockwise.cu
```diff
@@ -552,6 +558,12 @@ TEST_P(FusedCastFloat8VectorwiseTestSuite, TestFusedCastFloat8Vectorwise) {
   q_opts.amax_epsilon = eps;
   q_opts.block_scaling_dim = 1u;
 
+  // On Blackwell and newer, the FP8 block scaling recipe is emulated with MXFP8,
+  // which requires using power of two scaling factors. Skip unsupported tests.
+  if (getDeviceComputeCapability() > hopperComputeCapability && !force_pow_2) {
```
timmoon10:
```suggestion
  if (getDeviceComputeCapability() >= blackwellComputeCapability && !force_pow_2) {
```

tests/cpp/operator/test_cast_float8blockwise.cu
```diff
@@ -501,6 +501,12 @@ TEST_P(FusedCastFloat8BlockwiseTestSuite, TestFusedCastFloat8Blockwise) {
   q_opts.amax_epsilon = eps;
   q_opts.block_scaling_dim = 2u;
 
+  // On Blackwell and newer, the FP8 block scaling recipe is emulated with MXFP8,
+  // which requires using power of two scaling factors. Skip unsupported tests.
+  if (getDeviceComputeCapability() > hopperComputeCapability && !force_pow_2) {
```
timmoon10:
```suggestion
  if (getDeviceComputeCapability() >= blackwellComputeCapability && !force_pow_2) {
```

tests/pytorch/test_float8_blockwise_gemm_exact.py
```diff
@@ -19,7 +20,8 @@
 
 def fp8_blockwise_gemm_supported() -> bool:
     supported, _ = FP8GlobalStateManager.is_fp8_block_scaling_available()
-    return supported
+    emulated = get_device_compute_capability() > (9, 0)
```
timmoon10:
```suggestion
    emulated = get_device_compute_capability() >= (10, 0)
```

tests/pytorch/test_float8_blockwise_scaling_exact.py
```diff
@@ -32,6 +33,7 @@
 if tensor_dump_dir_env is not None:
     TENSOR_DUMP_DIR = pathlib.Path(tensor_dump_dir_env)
 recipe_available, reason_for_no_recipe = FP8GlobalStateManager.is_fp8_block_scaling_available()
+recipe_emulated = get_device_compute_capability() > (9, 0)
```
timmoon10:
```suggestion
recipe_emulated = get_device_compute_capability() >= (10, 0)
```

transformer_engine/common/swizzle/swizzle_block_scaling.cu
```diff
@@ -0,0 +1,314 @@
...
+//   lane0.row = 0x0004080c
+//   lane1.row = 0x0105090d
+//   lane2.row = 0x02060a0e
+//   lane3.row = 0x03070b0f
+uint32_t __device__ __forceinline__ transpose_4x4_byte_matrix(const uint32_t row,
+                                                              const uint32_t lane,
+                                                              const uint32_t active_mask) {
+  using cu = const uint32_t;
+
+  cu m_0123_4567_89ab_cdef = row;
```
timmoon10, line 45:
```suggestion
  // Threads operate in groups of 4, and each thread stores 4 bytes at a time.
  // The bytes in this 4x4 matrix are labeled in hex. We shuffle around bytes
  // until we have transposed the 4x4 matrix.
  cu m_0123_4567_89ab_cdef = row;
```

transformer_engine/pytorch/csrc/extensions/gemm.cpp
```diff
@@ -462,29 +474,72 @@ std::optional<std::vector<at::Tensor>> te_general_grouped_gemm(
...
-  auto swizzled_scale_inv_A = multi_tensor_swizzle_scaling_factors(te_A_wrappers, transa);
-  auto swizzled_scale_inv_B = multi_tensor_swizzle_scaling_factors(te_B_wrappers, !transb);
+  swizzled_scale_inverses_list.emplace_back(
+      multi_tensor_swizzle_scaling_factors(te_A_wrappers, transa));
+  swizzled_scale_inverses_list.emplace_back(
+      multi_tensor_swizzle_scaling_factors(te_B_wrappers, !transb));
+
+  // Emulate the FP8 block scaling recipe with MXFP8 on Blackwell and newer
+  // as it is not natively supported by cublasLt
+  if (transformer_engine::cuda::sm_arch() > 90) {
```
timmoon10:
```suggestion
  if (transformer_engine::cuda::sm_arch() >= 100) {
```

transformer_engine/pytorch/csrc/util.h
```diff
@@ -27,4 +27,16 @@ std::optional<at::Tensor> swizzle_scaling_factors(transformer_engine::TensorWrap
...
 
+/*! \brief Convert a block scaling tensor to an mxfp8 tensor.
+ *
+ * If rowwise==false, the columnwise data will be reinterpreted as rowwise data to avoid
+ * transposing it in memory. Due to differences in how block scaling and mxfp8 store data,
+ * this requires the calling code to treat the output tensor as having been tranposed in this case.
+ *
+ * Returns the swizzled scaling factor of the converted mxfp8 tensor.
+ * The returned swizzled scaling factor tensor should be kept alive during the GEMM.
+ */
```
timmoon10:
```suggestion
/*! \brief Convert a block scaling tensor to an mxfp8 tensor in-place.
 *
 * If rowwise==false, the columnwise data will be reinterpreted as rowwise data to avoid
 * transposing it in memory. Due to differences in how block scaling and mxfp8 store data,
 * this requires the calling code to treat the output tensor as having been tranposed in this case.
 *
 * Returns the swizzled scaling factor of the converted mxfp8 tensor.
 * The returned swizzled scaling factor tensor should be kept alive during the GEMM.
 */
```

transformer_engine/common/swizzle/swizzle_block_scaling.cu
```diff
@@ -0,0 +1,314 @@
...
+
+  const uint32_t tiles_x = DIVUP(data_cols, 128u);
+  const uint32_t tiles_y = DIVUP(data_rows, 128u);
+  const dim3 grid_dim{DIVUP(tiles_x, WARPS_X_PER_TB), DIVUP(tiles_y, WARPS_Y_PER_TB), 1};
+  const dim3 block_dim{WARP_SIZE, WARPS_Y_PER_TB, WARPS_X_PER_TB};
+
+  const uint32_t input_scale_inv_cols = DIVUP(data_cols, 512u) * 4;
+  const uint32_t in_y_stride = input_scale_inv_cols * sizeof(float);
+
+  const uint32_t out_y_stride = tiles_x * 512;
```
timmoon10:
```suggestion
  // Each 128x128 tile in the data corresponds to a 1x1 tile in the input scales
  // and a 128x4 tile in the output scales.
  const uint32_t input_scale_inv_cols = DIVUP(data_cols, 512u) * 4;
  const uint32_t output_scale_inv_cols = tiles_x * 128 * 4;
  const uint32_t in_y_stride = input_scale_inv_cols * sizeof(float);
  const uint32_t out_y_stride = output_scale_inv_cols * sizeof(uint8_t);
```

transformer_engine/common/swizzle/swizzle_block_scaling.cu
```diff
@@ -0,0 +1,314 @@
...
+
+  const uint32_t tiles_x = DIVUP(data_cols, 128u);
+  const uint32_t tiles_y = DIVUP(data_rows, 128u);
+  const dim3 grid_dim{DIVUP(tiles_x, WARPS_X_PER_TB), DIVUP(tiles_y, WARPS_Y_PER_TB), 1};
+  const dim3 block_dim{WARP_SIZE, WARPS_Y_PER_TB, WARPS_X_PER_TB};
+
+  const uint32_t input_scale_inv_cols = DIVUP(data_rows, 4u) * 4;
+  const uint32_t in_y_stride = input_scale_inv_cols * sizeof(float);
+
+  const uint32_t out_y_stride = tiles_x * 512;
```
timmoon10:
```suggestion
  // Each 128x128 tile in the data corresponds to a 128x1 tile in the input scales
  // and a 128x4 tile in the output scales. The input scales are in transposed order.
  const uint32_t input_scale_inv_cols = DIVUP(data_rows, 4u) * 4;
  const uint32_t output_scale_inv_cols = tiles_x * 128 * 4;
  const uint32_t in_y_stride = input_scale_inv_cols * sizeof(float);
  const uint32_t out_y_stride = output_scale_inv_cols * sizeof(uint8_t);
```

transformer_engine/common/swizzle/swizzle_block_scaling.cu
```diff
@@ -0,0 +1,314 @@
...
+  const size_t output_scale_inv_cols = output->scale_inv.shape[1];
+
+  NVTE_CHECK(output_scale_inv_rows == DIVUP<size_t>(data_rows, 128) * 128,
+             "Expected the output scaling factor matrix to have ",
+             DIVUP<size_t>(data_rows, 128) * 128, " rows, but it has ", output_scale_inv_rows,
+             " rows instead.");
+  NVTE_CHECK(output_scale_inv_cols == DIVUP<size_t>(data_cols, 128) * 4,
+             "Expected the output scaling factor matrix to have ",
+             DIVUP<size_t>(data_cols, 128) * 4, " columns, but it has ", output_scale_inv_cols,
+             " columns instead.");
```
timmoon10, line 287: Be advised that these dimensions are lies. The MXFP8 scales are actually laid out with dims `(data_rows / 128, data_cols / 128, 128*4)`.

timmoon10 (approved):
LGTM, pending CI

yaox12:
Hi @janekb04, just to confirm, with this PR, we still need both rowwise and colwise (transposed) FP8 weights on Blackwell, right? Even though Blackwell supports all layouts for FP8 GEMMs and the colwise quantized weights are just transposed rowwise ones because of the 2D scaling. Am I understanding it correctly?

janekb04:
> Hi @janekb04, just to confirm, with this PR, we still need both rowwise and colwise (transposed) FP8 weights on Blackwell, right? Even though Blackwell supports all layouts for FP8 GEMMs and the colwise quantized weights are just transposed rowwise ones because of the 2D scaling. Am I understanding it correctly?

Yes, this PR still uses both rowwise and columnwise data. Indeed, in the case of 2d block scaling, this could be avoided because the columnwise data is the transpose of the rowwise data. In case of 1d block scaling, however, this could increase numerical differences because the data would be quantized in the other direction, and would not necessarily be the same after dequantization.

In the section on future optimizations I point the two optimizations regarding this I found that could be applicable:

> Don't create columnwise data for weights. Because weights are 2D-block-scaled, their quantized columnwise data is simply the transpose of the rowwise data. As such, columnwise data doesn't have to be created in the forward pass, as the rowwise data can be used in the dgad GEMM.
> Don't transpose data after all gather. Currently, _post_process_fp8_blockwise_gather in distributed.py transposes columnwise data after all gather. This could be avoided.

PR #571, opened 2023-12-18
Activation offloading to CPU's for the Linear, Layernorm Linear and the Layernorm MLP modules

sanandaraj5597:
This PR adds support for offloading all the tensors saved for backward pass by the Linear, Layernorm Linear and the Layernorm MLP modules except the weight tensors.

The _cpu_offloading_ switch will be sent from the M-LM module construction which will enable calling the PyTorch hooks during tensor saving and retrieval. When these PyT hooks are called, the weight.main_grad isn't saved, so we save them separately on a need basis when we need to fuse gradients. All these layer executions will be done under a context which is placed at M-LM repo.

Please review and let me know if you have any questions.

PR #2158, opened 2025-09-06
[PyTorch] Support activation CPU offloading in fusible ops

timmoon10:
# Description

This PR modifies the fusible ops to support CPU offloading for activations (see https://github.com/NVIDIA/TransformerEngine/pull/571). It also simplifies the test and makes it more precise.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Support activation CPU offloading in fusible ops
- Debug and refactor CPU offloading test

# Checklist:

### DIFF
--- a/tests/pytorch/test_cpu_offloading.py
+++ b/tests/pytorch/test_cpu_offloading.py
@@ -109,9 +109,9 @@ def _estimate_cached_weight_size(
 
     # FP8 tensor-scaling caches one byte per element
     if quantization_recipe.delayed() or quantization_recipe.float8_current_scaling():
-        if (
-            not is_non_tn_fp8_gemm_supported()
-            and model_name not in ("linear_op", "layernorm_mlp_ops")
+        if not is_non_tn_fp8_gemm_supported() and model_name not in (
+            "linear_op",
+            "layernorm_mlp_ops",
         ):
             # Modules do not deallocate FP8 transpose for weights
             return 2 * param_elements / 1024**2
@@ -162,8 +162,7 @@ def _measure_cached_memory(
     memory_before_forward = torch.cuda.memory_allocated() / (1024**2)
     for module in modules:
         with te.fp8_autocast(
-            enabled=quantization_recipe is not None,
-            fp8_recipe=quantization_recipe
+            enabled=quantization_recipe is not None, fp8_recipe=quantization_recipe
         ), offload_context:
             tensor = module(tensor)
         tensor = sync_function(tensor)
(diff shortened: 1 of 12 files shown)


tests/pytorch/test_cpu_offloading.py
```diff
@@ -2,24 +2,30 @@
...
+mxfp8_available, _ = FP8GlobalStateManager.is_mxfp8_available()
 
-fp8_recipes = [None]
+quantization_recipes: Optional[recipe.Recipe] = [None]
 if fp8_available:
-    fp8_recipes.append(recipe.Float8CurrentScaling())
-    fp8_recipes.append(recipe.DelayedScaling())
+    quantization_recipes.extend((recipe.Float8CurrentScaling(), recipe.DelayedScaling()))
+if mxfp8_available:
+    quantization_recipes.append(recipe.MXFP8BlockScaling())
```
timmoon10: It seems CPU offloading with MXFP8 is not yet handled properly.
```suggestion
```

transformer_engine/pytorch/ops/_common.py
```diff
@@ -29,7 +29,9 @@ def maybe_dequantize(
     if is_quantized_tensor(tensor):
         return tensor.dequantize(dtype=dtype)
     if dtype is not None and tensor.dtype != dtype:
-        return tensor.to(dtype)
+        tensor = tensor.to(dtype)
```
pggPL, line 32: I see that you use it multiple times in the code on tensors after .contiguous(), so this may be inconsistent. And if this function will be used frequently, we may want to use .to(dtype=..., memory_format=torch.contiguous).

If I understand correctly you want to make sure that we offload contiguous tensors, which makes sense.

pggPL (approved):
LGTM with one minor remark, which is only partially related to this PR

Anyways all of this will not be needed after https://github.com/NVIDIA/TransformerEngine/pull/1762 is merged. But I get it may be needed faster for mlperf submission.

transformer_engine/pytorch/ops/_common.py
timmoon10: This was actually fixing an unrelated bug that was exposed in the unit test. The TE ops assume contiguous data, but they weren't handling cases when they recieved non-contiguous grads in the backward pass.

Also, PyTorch has a bug where `tensor.to(memory_format=torch.contiguous_format)` doesn't do anything: https://github.com/pytorch/pytorch/issues/132020.

PR #2161, opened 2025-09-08
[Pytorch] Support for Swiglu Activation used in GPT OSS

vthumbe1503:
# Description

Pytorch integration for Swiglu Activation used in [GPT OSS](https://github.com/openai/gpt-oss/blob/a0a84273e9e0c14a233cb9befdfd159c2bcfa6cd/gpt_oss/torch/model.py#L250)

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes
- nvte functions for gpt oss's swiglu and dswiglu added. This is different from other nvte functions for activations in sense that it accepts float parameters for limit(used in clamping the input) and alpha(used in scaling sigmoid's input).
- math.h has the activation and deactivation device functions defined for oss based silu that includes clamping
- All the gated/dgated kernels accepts a generic Parameter p so that it can be passed to our gated_activation_kernels. Parameter name had already been templatized in most places. 
- Typically gate is directly multiplied by the activation in gated activation kernels. However, in GPT OSS, gate is clamped and added 1 before multiplying it with clamped silu activation. To acheive this in gated activation kernels, we now check for the parameter type as to whether they are GPT OSS based. If they are, gates are clamped and added 1.
- Activation Helper that does quantization after applying activation, originally assumed no parameter apart from input, quantizer is passed to the nvte functions. However since the GPT OSS activation also takes in limit and alpha, activation helper also templatizes and takes care of activation functions that can take in variable arguments apart from inputs and quantizer. Similar templatization is applied for dactivation helper functions.
- Unit test added that verifies end to end integration from python for GPT OSS based swiglu.

# Checklist:

### DIFF
--- a/transformer_engine/common/util/nvfp4_transpose.cuh
+++ b/transformer_engine/common/util/nvfp4_transpose.cuh
@@ -1433,7 +1433,8 @@ void nvfp4_quantize_transpose(const Tensor &input, const Tensor *noop, Tensor *o
   const size_t block_size = THREADS_NUM;
 
   const size_t scale_stride = output->scale_inv.shape[1];
-  const size_t scale_stride_transpose = output->columnwise_scale_inv.shape[1];
+  const size_t scale_stride_transpose =
+      return_transpose ? output->columnwise_scale_inv.shape[1] : 0;
 
   nvfp4_scale_t *const scales_ptr = reinterpret_cast<nvfp4_scale_t *>(output->scale_inv.dptr);
   nvfp4_scale_t *const scales_transpose_ptr =


transformer_engine/common/include/transformer_engine/activation.h
```diff
@@ -173,6 +173,24 @@ void nvte_geglu(const NVTETensor input, NVTETensor output, cudaStream_t stream);
  */
 void nvte_swiglu(const NVTETensor input, NVTETensor output, cudaStream_t stream);
 
+/*! \brief Computes the gated Swish activation of the input used in GPT OSS.
+ *        https://github.com/openai/gpt-oss/blob/a0a84273e9e0c14a233cb9befdfd159c2bcfa6cd/gpt_oss/torch/model.py#L250
```
timmoon10: Nit: The Doxygen `\brief` tag should be followed by an empty line so that Doxygen knows where to put the break.
```suggestion
/*! \brief Computes the gated Swish activation of the input used in GPT OSS.
 *
 *        See https://github.com/openai/gpt-oss/blob/a0a84273e9e0c14a233cb9befdfd159c2bcfa6cd/gpt_oss/torch/model.py#L250
 *
```
The other comment should also be changed.

transformer_engine/pytorch/ops/basic/activation.py
```diff
@@ -389,3 +390,36 @@ def _activation_forward_impl(self, *args, **kwargs) -> torch.Tensor:
...
+    ----------
+    limit: float
+        The clamp limit.
+    alpha: float
+        The scaling factor for the sigmoid function used in the activation.
+    cache_quantized_input: bool, default = False
+        Quantize input tensor when caching for use in the backward pass.
+    """
+
+    def __init__(self, *, limit: float, alpha: float, cache_quantized_input: bool = False):
```
timmoon10: Might as well use the same defaults as GPT-OSS (https://github.com/openai/gpt-oss/blob/a0a84273e9e0c14a233cb9befdfd159c2bcfa6cd/gpt_oss/torch/model.py#L249).
```suggestion
    def __init__(
        self,
        *,
        limit: float = 7,
        alpha: float = 1.702,
        cache_quantized_input: bool = False,
    ):
```
It may also be worth mentioning in the docs that alpha=1.702 is taken from [the original GELU paper](https://arxiv.org/pdf/1606.08415) for an approximation to GELU.

transformer_engine/pytorch/ops/basic/activation.py
```diff
@@ -389,3 +390,36 @@ def _activation_forward_impl(self, *args, **kwargs) -> torch.Tensor:
 
     def _activation_backward_impl(self, *args, **kwargs) -> torch.Tensor:
         return tex.dswiglu(*args, **kwargs)
+
+
+class GptOssSwiglu(_ActivationOperation):
```
timmoon10: Nit: [The original SwiGLU paper](https://arxiv.org/pdf/2002.05202) calls it "SwiGLU".
```suggestion
class GptOssSwiGLU(_ActivationOperation):
```
Stylistically I don't like that we've camel-cased "GPT" and "OSS", but I guess "GPTOSSSwiGLU" would be overwhelming.

transformer_engine/common/include/transformer_engine/activation.h
```diff
@@ -173,6 +173,24 @@ void nvte_geglu(const NVTETensor input, NVTETensor output, cudaStream_t stream);
...
+ *         the block quantization (MXFP8) of the specified shape of the block will be used.
+ *
+ *  \param[in]     input     Input tensor of shape [N, H * 2].
+ *  \param[in,out] output    Output tensor of shape [N, H].
+ *                           It computes Act(input[N, :H]) x input[N, H:]
+ *  \param[in]     limit     Clipping limits for gate and pre-activation.
+ *  \param[in]     alpha     Scaling factor for the sigmoid function used in the activation.
+ *  \param[in]     stream    CUDA stream used for the operation.
+ */
+void nvte_gptoss_swiglu(const NVTETensor input, NVTETensor output, float limit, float alpha,
```
timmoon10: Nit: We separate "gpt" and "oss" elsewhere, so we should do so here as well for consistency.
```suggestion
void nvte_gpt_oss_swiglu(const NVTETensor input, NVTETensor output, float limit, float alpha,
```
Similar for `nvte_gpt_oss_dswiglu`.

tests/pytorch/test_fusible_ops.py
```diff
@@ -1707,6 +1707,75 @@ def test_swiglu(
...
+        out_glu = x_glu * torch.sigmoid(1.702 * x_glu)
+        y_ref = out_glu * (x_linear + 1)
+        y_ref.backward(dy_ref)
+
+        # Implementation with fusible operation
+        recipe = make_recipe(quantization)
+
+        forward = te_ops.Sequential(
+            te_ops.Quantize(forward=False, backward=quantize_backward),
+            te_ops.GptOssSwiglu(limit=0.1, alpha=1.702),
```
timmoon10: `limit=0.1` seems a bit small. Won't that kill a lot of the grads?

timmoon10 (commented):
The functionality and APIs look good.

I'm not a fan of how it makes the activation implementation even more of a rat-nest. Instead of having these monstrous templated functions that have to keep track of all feature combinations (cast fusion, dbias fusion, and now user-provided args), it would be better to have separate implementations that are more focused. That refactor may be beyond the scope of this PR though.

tests/pytorch/test_fusible_ops.py
vthumbe1503: I was using this limit to check for correctness of dgrad implementation. Turns out when I keep a higher value like 7.0 which is used in the paper, the dgrad values from pytorch and TE were always matching. When I then changed to a small value like 0.1 it started failing and the reason was incorrect clamping implementation for the dgrad of swiglu. So most likely the gradients were not getting killed at all disallowing us from actually checking whether the clamping is correctly implemented.

transformer_engine/common/include/transformer_engine/activation.h
vthumbe1503: Done

transformer_engine/pytorch/ops/basic/activation.py
vthumbe1503: Done

transformer_engine/common/include/transformer_engine/activation.h
vthumbe1503: Have changed everything to clamped swiglu based on @ptrendx 's suggestion

transformer_engine/common/util/math.h
```diff
@@ -75,7 +75,7 @@ __device__ inline OType silu(const IType val, const Empty& e) {
 }
 
 template <typename OType, typename IType>
-__device__ inline OType oss_silu(const IType val, const GptOssParam& p) {
+__device__ inline OType oss_silu(const IType val, const ClampedSwiGLUParam& p) {
```
timmoon10: Nit: `clamped_silu` and `clamped_dsilu` would be more consistent with the naming scheme.
```suggestion
__device__ inline OType oss_silu(const IType val, const ClampedSwiGLUParam& p) {
```

tests/pytorch/test_fusible_ops.py
timmoon10: I can see `limit=7` is way too high, but `limit=0.1` would also mean that all the grads are small (even the ones that are not killed), so we might miss some errors. The inputs are generated with `torch.rand`, so how about an intermediate value like `limit=0.75`?

timmoon10:
LGTM, although we should wait to merge until after https://github.com/NVIDIA/TransformerEngine/pull/2177 is merged.

timmoon10 (approved):

PR #2163, opened 2025-09-08
[JAX] Scale swizzling via JAX transpose op

phu0ngng:
# Description

XLA often inserts transpose ops before and after the `custom_partitioning` in the custom call to transform and partition the data and scale_inv.
This PR fuses the swizzle with the transpose after the `custom_partitioning`, resulting in up to 1.4% speedup in the E2E training of LLama3.1 8B on GB200x4.  

Profiling with TPSP2 + FSDP2, focusing on the First GEMM in the MLP.
 - Before fusing swizzling:
Screenshot 2025-09-08 at 12 29 27
- After fusing swizzling:
Screenshot 2025-09-08 at 12 29 45

The time between AG and launching the GEMM kernel reduces from 70us to 53us.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/gemm.py
+++ b/transformer_engine/jax/cpp_extensions/gemm.py
@@ -379,35 +379,22 @@ class GemmPrimitive(BasePrimitive):
         grad,
         use_split_accumulator,
     ):
-        lhs_cdims, rhs_cdims = map(sanitize_dims, (lhs.ndim, rhs.ndim), contracting_dims)
-        lhs_transposed, rhs_transposed = _get_gemm_layout(
-            (lhs.ndim, rhs.ndim), (lhs_cdims, rhs_cdims)
-        )
+        if scaling_mode.is_1d_block_scaling():
+            lhs_cdims, rhs_cdims = map(sanitize_dims, (lhs.ndim, rhs.ndim), contracting_dims)
+            lhs_transposed, rhs_transposed = _get_gemm_layout(
+                (lhs.ndim, rhs.ndim), (lhs_cdims, rhs_cdims)
+            )
+            lhs_flatten_axis = max(lhs_cdims) + 1 if lhs_transposed else min(lhs_cdims)
+            rhs_flatten_axis = min(rhs_cdims) if rhs_transposed else max(rhs_cdims) + 1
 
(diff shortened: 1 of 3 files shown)


transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -408,7 +424,39 @@ def impl(
             grad=grad,
             use_split_accumulator=use_split_accumulator,
         )
-        return outputs[:-3]  # discard workspace arrays
+        return outputs[:-1]  # discard workspace array
+
+    @staticmethod
+    def outer_impl(
```
jberchtold-nvidia, line 417: Why do we need a new `outer_impl` here that just passes through args? Is that because if we didn't have the outer_impl we'd be running the swizzle operation twice, once for the outer primitive and once for the inner primitive and getting an incorrect result?

jberchtold-nvidia:
Overall LGTM pending CI, just one question on outer_impl. Great idea for this perf improvement!

transformer_engine/jax/cpp_extensions/gemm.py
phu0ngng: In the single-GPU case, `impl` is invoked directly.
In the multi-GPU case, the flow looks like this:
```
outer_impl --> partitioning --> impl
```
We should not do a swizzle in the `outer_impl`. Otherwise, we will partition the scale wrongly.

Adding the arg `is_outer` as we did for other primitives could not serve the purpose because we still can't distinguish if it is the outer or inner call, thus we need to add the new method `outer_impl`.

jberchtold-nvidia: Sounds good, thanks!

I guess in theory we could use this (or is_outer) for the scale slicing in primitives like quantization, as we don't need to slice scales to be unpadded in the outer_impl. Assuming my understanding is correct that for other primitives the outer_impl is bound to just `impl` as there's only a single def'n. Though in practice I don't think this matters much as it will be a no-op if we try to slice out the unpadded shape from scales that are already unpadded.

jberchtold-nvidia:
LGTM

jberchtold-nvidia (approved):
LGTM

PR #2166, opened 2025-09-09
[JAX] CollectiveGemm

phu0ngng:
# Description

This PR adds CollectiveGEMM, which allows communication to be fully overlapped with the computation of the GEMM in TPSP parallelism.

For now, it only supports:
- BF16 GEMM
- Single device per process

The following supports are planned for follow-up PRs:
- Multiple devices per process
- Flax module integration
- FP8
- Reused the gathered grad from the DGRAD GEMM for WGRAD (only if there is E2E perf benefit).

Usage:
```
from transformer_engine.jax.cpp_extensions.gemm import collective_gemm_bootstrap, CollectiveOp
import transformer_engine.jax.cpp_extensions as tex

# args from command line
jax.distributed.initialize(
    coordinator_address=args.coordinator_address,
    num_processes=args.num_processes,
    process_id=args.process_id,
    local_device_ids=global_device_ids_for_this_process
)
    
# Bootstrap distributed communicators, once per program, after jax.distributed.initialize
collective_gemm_bootstrap(
    num_total_devices=args.num_total_devices,
    num_devices_per_process=1,  # Single device per process
    process_id=args.process_id,
    tensor_parallel_size=args.tp_size,
)

# AllGather CGEMM
output = tex.gemm(
    x, weight, bias=bias,
    contracting_dims=((2,), (0,)),
    collective_op=CollectiveOp.ALL_GATHER
)

# ReduceScatter CGEMM
output = tex.gemm(
    x, weight, bias=bias,
    contracting_dims=((2,), (0,)),
    collective_op=CollectiveOp.REDUCE_SCATTER
)
```

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


build_tools/jax.py
```diff
@@ -87,4 +87,5 @@ def setup_jax_extension(
         sources=[str(path) for path in sources],
         include_dirs=[str(path) for path in include_dirs],
         extra_compile_args=cxx_flags,
+        libraries=["nccl"],
```
phu0ngng, line 90: Hi @ksivaman,

I need to link the TE-JAX build with NCCL. Hope that it is ok to do it in this way.

examples/jax/collective_gemm/test_dense_grad.py
```diff
@@ -0,0 +1,215 @@
...
+        gathered_grads = []
+        gathered_ref_grads = []
+        for ref_grad, grad in zip(ref_grads, sharded_grads):
+            gathered_grads.append(
+                jax.lax.with_sharding_constraint(grad, NamedSharding(mesh, PartitionSpec(None)))
+            )
+            gathered_ref_grads.append(
+                jax.lax.with_sharding_constraint(ref_grad, NamedSharding(mesh, PartitionSpec(None)))
+            )
+        jax.block_until_ready(gathered_grads)
```
jberchtold-nvidia, line 147: Why do we need `block_until_ready` here? Is this because we have multiple processes and only one process is checking the result with assert_allclose?

Is anything here specific to custom communications for the comm+gemm overlap or is this blocking required for any multiprocess usage? Thanks!

examples/jax/collective_gemm/test_layernorm_mlp_grad.py
```diff
@@ -0,0 +1,275 @@
...
+        jax.block_until_ready(gathered_grads)
+        jax.block_until_ready(gathered_ref_grads)
+
+    if args.enable_result_check and args.process_id == 0:
+        assert_allclose(ref_output, output, dtype=jnp.bfloat16)
+        for ref_grad, gathered_grad in zip(gathered_ref_grads, gathered_grads):
+            assert_allclose(ref_grad, gathered_grad, dtype=jnp.bfloat16)
+
+
+class TestCollectiveDenseGradient(unittest.TestCase):
```
jberchtold-nvidia: self-resolve: `TestCollectiveLayernormMLPGradient`?

examples/jax/collective_gemm/test_layernorm_mlp_grad.py
```diff
@@ -0,0 +1,275 @@
...
+    """Collective Dense Gradient unittests"""
+
+    # is_fp8_supported, fp8_reason = is_fp8_available(ScalingMode.DELAYED_TENSOR_SCALING)
```
jberchtold-nvidia: self-resolve: Unused comment code

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1259,6 +1627,7 @@ def gemm(
             "`jax.lax.dot_general` and `jax.nn.scaled_matmul` backends used when the custom cuBLAS "
             "GEMM primitive is disabled."
         )
+        assert collective_op.is_none, "JAX GEMM does not support collective GEMM"
```
jberchtold-nvidia, line 1634: Was looking to make sure this wouldn't silently do a regular gemm without overlap for the JAX gemm and you already have a check for that, nice!

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -302,9 +489,12 @@ def _dims_are_consecutive(dims):
...
+        workspace_size = get_cublas_workspace_size_bytes()
+        if not collective_op.is_none:
+            workspace_size *= get_cgemm_num_max_streams()
         # cuBLAS workspace ptr must be 256 bytes aligned but JAX buffers are not
         # necessarily 256 bytes aligned, we add some padding to ensure alignment.
-        workspace_size = get_cublas_workspace_size_bytes() + 256
+        workspace_size += 256
         workspace = jax.core.ShapedArray(shape=(workspace_size,), dtype=jnp.uint8)
 
         return output, bias_grad, pre_gelu_out, workspace
```
jberchtold-nvidia, line 500: This `pre_gelu_out` is not fully supported now and we're going to remove it, right? So we don't need to update it in the same way as `output` is updated for comm+gemm overlap

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -492,22 +766,13 @@ def batcher(
...
         contracting_dims,
+        transpose_batch_sequence,
+        collective_op,
     ):
         lhs_specs, _, rhs_specs, *_ = map(get_padded_spec, arg_infos)
 
         gsr = global_mesh_resource()
 
-        # Ensure that tensor sequence parallelism is not used via setting tp_resource
-        if gsr.tp_resource is not None:
```
jberchtold-nvidia, line 501: Don't we still want this check when comm+gemm overlap is disabled?

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -0,0 +1,262 @@
...
+  return unique_id;
+}
+
+void CommunicatorHandler::init(int num_total_devices, int num_devices_per_process, int process_id,
+                               int tp_size) {
+  // Validate inputs
+  NVTE_CHECK(num_devices_per_process <= MAX_DEVICES,
+             "num_devices_per_process exceeds MAX_DEVICES=", MAX_DEVICES,
+             ", got num_devices_per_process=", num_devices_per_process);
+  NVTE_CHECK(num_devices_per_process >= 1,
```
jberchtold-nvidia: Do we need to assert `num_devices_per_process` is strictly equal to `1` for now since that's all we support? Or is the reason for not supporting multi-device per process purely on the TE/JAX/Python side and this assert is okay as is.

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -0,0 +1,262 @@
...
+  for (int local_idx = 0; local_idx < num_devices_per_process; local_idx++) {
+    NVTE_CHECK_CUDA(cudaSetDevice(handler.local_device_ids_within_process[local_idx]));
+    int tp_local_rank = handler.local_device_ids_within_tp_domain[local_idx];
+    NVTE_CHECK_NCCL(
+        ncclCommInitRank(&handler.tp_comms[local_idx], handler.tp_size, tp_id, tp_local_rank));
+  }
+  NVTE_CHECK_NCCL(ncclGroupEnd());
+
+  // Allocate device memory for barrier operations
+  NVTE_CHECK_CUDA(cudaMalloc(&handler._barrier, sizeof(int)));
```
jberchtold-nvidia: This is fine to allocate outside of JAX's allocator because it's just 4-bytes per GPU, right?

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -0,0 +1,262 @@
...
+
+  int device_idx = get_local_device_idx_for_current_device();
+  ncclComm_t tp_comm = tp_comms[device_idx];
+
+  size_t expected_output_bytes = input_bytes * tp_size;
+  NVTE_CHECK(output_bytes == expected_output_bytes, "TP allgather buffer size mismatch: expected ",
+             expected_output_bytes, ", got ", output_bytes);
+
+  NVTE_CHECK_NCCL(ncclAllGather(input_buf, output_buf, input_bytes, ncclChar, tp_comm, nullptr));
+  cudaDeviceSynchronize();
```
jberchtold-nvidia, line 232: We need a full device sync here not just a stream sync?

transformer_engine/jax/csrc/extensions/cgemm_helper.h
```diff
@@ -0,0 +1,193 @@
...
+#include "common/comm_gemm_overlap/userbuffers/userbuffers.h"
+#include "common/util/cuda_runtime.h"
+#include "common/util/logging.h"
+#include "transformer_engine/comm_gemm_overlap.h"
+
+namespace transformer_engine {
+namespace jax {
+
+#ifndef MAX_DEVICES
+#define MAX_DEVICES 8
```
jberchtold-nvidia: Should we increase this constant for production scenarios?

transformer_engine/jax/csrc/extensions/cgemm_helper.h
```diff
@@ -0,0 +1,193 @@
...
+class CommunicatorHandler {
+ public:
+  int num_total_devices = -1;
+  int num_devices_per_process = -1;
+  int process_id = -1;
+  int num_processes = -1;
+
+  int tp_size = -1;
+  int tp_num_domains = -1;
+  int local_device_ids_within_tp_domain[MAX_DEVICES] = {-1};
```
jberchtold-nvidia: Why do these need to be C-style fixed size arrays determined at compile time? Could these instead be an std::vector whose size is dictated by an environment variable, so each process will still have the same array size, but folks using lots of GPUs won't need to recompile this code with a new MAX_DEVICES?

Lmk if there's some inherent limitation here that requires the fixed C-style arrays, thanks!

transformer_engine/jax/csrc/extensions/cgemm_helper.h
```diff
@@ -0,0 +1,193 @@
...
+
+  CommunicatorHandler(const CommunicatorHandler &) = delete;
+  CommunicatorHandler &operator=(const CommunicatorHandler &) = delete;
+
+ private:
+  CommunicatorHandler();
+  ~CommunicatorHandler();
+
+  bool _initialize = false;
+  int *_barrier = nullptr;
```
jberchtold-nvidia: self-resolve: Can we add a comment to indicate that this is a device ptr?

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -133,9 +195,63 @@ Error_Type GemmFFI(cudaStream_t stream, Buffer_Type lhs, Buffer_Type lhs_scale_i
...
+                 "cuBLAS GEMM output buffer size is incorrect, expected ", out_.numel(),
+                 " elements ", to_string_like(out_shape), " but got ", output->element_count(),
+                 " elements ", to_string_like(output->dimensions()));
+
+      // Launch GEMM+RS
+      executor->split_overlap_rs(rhs_, rhs_transposed, lhs_, lhs_transposed, ubuf_out_, bias_,
+                                 pre_gelu_, workspace_, grad, false, use_split_accumulator, out_,
+                                 stream);
+
+      // TODO: Don't we need to copy the output back to the original buffer?
```
jberchtold-nvidia: Reminder about this TODO

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -288,7 +302,9 @@ CommOverlapBase::CommOverlapBase(const std::vector<size_t> &buffer_shape, DType
   size_t buffer_bytes = get_buffer_size_bytes(buffer_shape[0], buffer_shape[1], buffer_dtype);
   void *buffer_ptr;
   _ub_reg = register_user_buffer_collective(&buffer_ptr, buffer_bytes, _ub_comm, true);
```
jberchtold-nvidia, line 304: Will this allocate memory outside of JAX's allocation pool? From looking at the impl of this `register_user_buffer_collective` it seems like it will, but it may be required to due to special permissions for NCCL set up

Is my understanding correct? And if so, is there anything we can do to make sure JAX's allocation pool doesn't compete with these allocations and lead to OOM, such as recommending users to not set JAX's pool default usage to e.g. 99% GPU memory.

examples/jax/collective_gemm/test_dense_grad.py
phu0ngng: Hi, 
`jax.block_until_ready` actually does not wait for all processes. It simply forces the Python program to wait for the completion of JAX computations on a device within a single process. It's a good practice to do in JAX in general before inspecting any tensors.

transformer_engine/jax/cpp_extensions/gemm.py
phu0ngng: Yes, I will remove `pre_gelu_out` as we don't ever test it and `bias_grad` as we don't need it at all, after FP4 work is concluded.

transformer_engine/jax/cpp_extensions/gemm.py
phu0ngng: Was mistakenly removed. I added them back. Thanks

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
phu0ngng: For now, it only supports a single device per process. I updated the assertion to check for `num_devices_per_process == 1`. Thanks

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
phu0ngng: I think it's fine to do this for now, as this bootstrap is called only once per program.

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
phu0ngng: We need all processes barriered (achieved via an AllGather) and device-to-host sync. 
We can do either `cudaStreamSynchronize()` to sync all streams or `cudaDeviceSynchronize()` to sync all devices. I feel like `cudaDeviceSynchronize` is more strict. It has the worst performance, but it won't matter much as this barrier is used only in the bootstrap, which happens once per program.

transformer_engine/jax/csrc/extensions/cgemm_helper.h
phu0ngng: Right now, we only support a single device per process, so it won't be a problem yet.
When we support multiple devices per process, we should consider what the best `MAX_DEVICES` would be.

transformer_engine/jax/csrc/extensions/cgemm_helper.h
phu0ngng: Good point. I did it via a C array because of the `ncclComm_t tp_comms[MAX_DEVICES]`. Let me check if it is possible to use std::vector here.

transformer_engine/jax/csrc/extensions/cgemm_helper.h
phu0ngng: Good point. I renamed it to `_device_barrier`.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
phu0ngng: From my understanding, this `register_user_buffer_collective()` allocates buffers in a special memory segment which is accessible by NCCL/NVLINK. When JAX supports symmetric memory buffers in the future, perhaps we can try to allocate them as JAX buffers, but there is no solution right now.

More E2E tests need to be done before we can make any recommendations to users.

Fortunately, we only allocate 4 extra buffers with a total size of ~6 times the input matrix for the decoder, and these buffers are reused for later layers, so hopefully it won't cause much trouble.

transformer_engine/jax/csrc/extensions/cgemm_helper.h
jberchtold-nvidia: Ah got it, didn't realize this was per-process but that makes sense. Thanks!

examples/jax/collective_gemm/test_dense_grad.py
jberchtold-nvidia: Thanks! Does that mean that we should update our `assert_allclose` in our main tests to call `block_until_ready` on the inputs before passing the arrays to `np.testing.assert_allclose`?
https://github.com/NVIDIA/TransformerEngine/blob/3f875fb57fcf2872d238f8c7cb199b171c424536/tests/jax/utils.py#L1465-L1466

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
jberchtold-nvidia: Sounds good, thanks for the explanation!

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
jberchtold-nvidia: Sounds good, thanks for the explanation!

transformer_engine/jax/csrc/extensions/cgemm_helper.h
denera: Unless I'm misunderstanding what this is for, won't this incorrectly restrict MNNVL runs to TP=8 even though Userbuffers supports up to 32?

transformer_engine/jax/csrc/extensions/misc.h
```diff
@@ -87,5 +87,31 @@ constexpr struct Alignment {
 
 std::vector<size_t> get_mxfp8_scale_shape(size_t M, size_t N, bool is_colwise);
 
+template <typename T, typename... Rest>
+void hash_combine(int64_t &seed, const T &v, Rest... rest) {
+  seed ^= std::hash<T>{}(v) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+  (hash_combine(seed, rest), ...);
+}
+
+enum class JAXX_Collective_Op : int64_t {
```
denera, line 96: This enum seems redundant when `CommOverlapType` is already exposed to Python (see `transformer_engine/common/util/pybind_helper.h`). Is there a reason why you don't want to use what's already available in TE/common?

transformer_engine/jax/dense.py
```diff
@@ -130,10 +140,13 @@ def _dense(
         kernel: Weight matrix
         bias: Optional bias tensor
         contracting_dims: Contracting dimensions specification
+        batch_sequence_transpose: Transpose the batch and sequence dimensions of the input tensor.
         input_axes: Logical axes for sharding the activation input
+        output_axes: Logical axes for sharding the output_axes
```
denera, line 145: Can we infer this from input and kernel axes using the contracting dims?

examples/jax/collective_gemm/test_dense_grad.py
phu0ngng: We probably should.

transformer_engine/jax/csrc/extensions/cgemm_helper.h
phu0ngng: Hi, 

This is max device per process, not per TP. 
Right now, we only support a single device per process, so it won't be an issue yet. 

Just curious, when we have NVL36/72, is it correct that we still have the max 8 devices per process?  

I'm looking into whether it is possible to replace the `ncclComm_t tp_comms[MAX_DEVICES]` with `std::vector<ncclComm_t>`. If it works, we can eliminate this `MAX_DEVICE`.

transformer_engine/jax/csrc/extensions/misc.h
phu0ngng: Hi, 

For the same reason as the `JAXX_Scaling_Mode`, we want to decouple it from TE/Common so that we don't need to change the code too much if there is a rename in TE/Common.

Additionally, we want to have an `enum class` in JAX so that it can be auto-decoded by XLA. Even though the `CommOverlapType` is an `enum class` right now, we should not forget that TE/Common is supposed to provide a C APIs.

transformer_engine/jax/dense.py
phu0ngng: We could not do that in the case that the output needs to be ReduceScatter. 

This arg is for the normal GEMM, not the CGEMM. Users either need to pass this arg or apply a sharding constraint on the output after calling our VJP op if they want a ReduceScatter.

transformer_engine/jax/csrc/extensions/cgemm_helper.h
phu0ngng: Hi, 

Using `std::vector<ncclComm_t>` does not cause any issue, so I removed the `MAX_DEVICE`.

jberchtold-nvidia: Awesome, thanks!

transformer_engine/jax/csrc/extensions/cgemm_helper.h
jberchtold-nvidia: Resolved with std::vector

transformer_engine/jax/cpp_extensions/gemm.py
phu0ngng: Actually, I recalled why I removed this check now. 
This check does not do anything as the `tpsp` won't ever appear twice in the LHS. 

I will update it so that it raises a warning if the `tp` is found in LHS instead.

jberchtold-nvidia: Got it, thanks!

phu0ngng:
Pipeline 35601160 passed except the `L0_jax_unittest--A100_1GPU` runs out of time. 
This PR does not add any tests to the L0_jax_unittest.

Issue #1880, opened 2025-06-13
Storage in fp8


dxqbYD:
Related to https://github.com/NVIDIA/TransformerEngine/issues/1261 and https://github.com/NVIDIA/TransformerEngine/issues/1764 but it is not entire clear there:

TransformerEngine could support storage in fp8 and could be dropping storage of weights in native precision after initialization. This might seem counterintuitive in a training environment, but please consider LoRA and other adapter trainings. Most of your weights you never need at their original precision again - you just want to use TransformerEngine for its efficient calculations.

The LoRA weights you keep at a higher precision.

The vram usage of TransformerEngine is currently prohibitive for training a small adapter to a large transformer.

**Describe alternatives you've considered**

Continue to use a custom Linear layer that stores in fp8, but doesn't have the efficient calculations performed by TransformerEngine

timmoon10:
You can store weights in low precision by initializing the model within an [`fp8_model_init`](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/api/pytorch.html#transformer_engine.pytorch.fp8_model_init) context. The expected memory usage will depend on the GPU arch:
- Blackwell: FP8 and MXFP8 inference only require 1 byte/param. MXFP8 training requires 2 bytes/param because the forward and backward pass require different scaling formats (1x32 blocks vs 32x1 blocks).
- Hopper: FP8 inference requires 1 byte/param. FP8 training requires 2 bytes/param because the forward and backward pass require different data layouts (non-transposed vs transposed). If there is demand, we could support an alternative implementation that reduces training memory usage to 1 byte/param, in exchange for some extra compute (see https://github.com/NVIDIA/TransformerEngine/pull/1827).

dxqb:
Thank you for pointing this out, I've missed that!

So far I have not been successful in using it though.

1. It fails in an `accelerate.init_empty_weights()` context
Maybe this is expected, but difficult to find because of the error message `TypeError: Float8TensorBase.__new__() missing 3 required keyword-only arguments: 'data', 'fp8_scale_inv', and 'fp8_dtype'`

After removing that, 

2. Apparently moving the model to GPU causes dequantization for some reason and uses large amounts of VRAM. Note that the model *without* fp8_model_init, storing both precisions, still barely fit into 32 GB. *With* fp8_model_init, it needs more VRAM while the model is moved?
Possibly a `diffusers` incompatibility with transformer engine
```
    self.transformer.to(device=device)
[...]

PR #2168, opened 2025-09-10
blockwise fp8 weight memory optimization: on-demand columnwise fp8 weight creation

skydoorkai:
# Description

When blockwise fp8 is used, weights are quantized into rowwise_data and columnwise_data. rowwise_data is used in forward pass and columnwise_data is used in backward pass.

Current implementation will create columnwise_data in advance for training, so the blockwise fp8 weight GPU memory consumption is similar to bf16 (plus two scale tensors). 

Since weights are 2D-quantized for blockwise fp8, columnwise_data can be created from rowwise_data with a blockwise-quantization-aware transpose without loss of accuracy, this PR adds a GPU memory optimization method by only creating columnwise_data when needed (in backward pass) from rowwise_data. This will reduce fp8 weight GPU memory in half.
Similar to activation checkpointing, this is a tradeoff between GPU memory and computation.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add FP8_BLOCKWISE_WEIGHT_ON_DEMAND_TRANSPOSE in FP8GlobalStateManager. Default if False, when users can set it by setting an env variable NVTE_ON_DEMAND_FP8_WEIGHT_TRANSPOSE=1 and blockwise fp8 recipe is used.

- Add get_columnwise_fp8_tensor function to create columnwise fp8 data from rowwise fp8 data by a triton kernel.

- Add support for GroupedLinear, Linear, LayerNormLinear.

- Add test_fp8_weight_on_demand_transpose test

# Checklist:


yaox12:
In fact you don't need a standalone kernel to do the transpose. Just call `weight_fp8.update_usage(columnwise_usage=True)`.

BestJuly:
> In fact you don't need a standalone kernel to do the transpose. Just call `weight_fp8.update_usage(columnwise_usage=True)`.

If using `update_usage`, it will generate the transposed data when it does not have transposed one. And after transpose, in the next micro batch backward pass within the same global batch, the columnwise data is always there, which may not save memory as what they want. I think they want to use this kernel to generate the transposed weight and release the memory right after the usage.

But yes, not sure why it does not use existing kernel like using [tex.fp8_transpose](`https://github.com/NVIDIA/TransformerEngine/blob/4903f947d6de871cd92c3478c8b5b78f835d5b7f/transformer_engine/pytorch/tensor/_internal/float8_blockwise_tensor_base.py#L326C33-L326C50`) as what has been done in `_create_columnwise()`.

yaox12:
> > In fact you don't need a standalone kernel to do the transpose. Just call `weight_fp8.update_usage(columnwise_usage=True)`.
> 
> If using `update_usage`, it will generate the transposed data when it does not have transposed one. And after transpose, in the next micro batch backward pass within the same global batch, the columnwise data is always there, which may not save memory as what they want. I think they want to use this kernel to generate the transposed weight and release the memory right after the usage.
> 
> But yes, not sure why it does not use existing kernel like using tex.fp8_transpose as what has been done in `_create_columnwise()`.

If you want to release the column data after use, you can set `columnwise_usage=False` after the GEMM. And I agree, at least we don't need to add another kernel to do the transpose.

skydoorkai:
> > > In fact you don't need a standalone kernel to do the transpose. Just call `weight_fp8.update_usage(columnwise_usage=True)`.
> > 
> > 
> > If using `update_usage`, it will generate the transposed data when it does not have transposed one. And after transpose, in the next micro batch backward pass within the same global batch, the columnwise data is always there, which may not save memory as what they want. I think they want to use this kernel to generate the transposed weight and release the memory right after the usage.
> > But yes, not sure why it does not use existing kernel like using tex.fp8_transpose as what has been done in `_create_columnwise()`.
> 
> If you want to release the column data after use, you can set `columnwise_usage=False` after the GEMM. And I agree, at least we don't need to add another kernel to do the transpose.

Yes, no new kernel is needed. Updated to use tex.fp8_transpose.

1tex:
Hi, is this PR still active? I've been working on a related implementation for frozen FP8 weights in PEFT/LoRA fine-tuning and wanted to coordinate before opening another PR.

It follows @yaox12's suggestion above: right after the dgrad GEMM, frozen 2D-block-scaled weights release their columnwise copy through the existing `update_usage` path (the compute-for-memory trade-off @timmoon10 mentioned in #1880). It's opt-in, covers `Linear`, `GroupedLinear`, `LayerNormLinear` and `LayerNormMLP`, and has guards for CUDA graphs and FSDP2 columnwise-only tensors.

Tested on 8× H800 including CUDA graphs, FSDP2 and TP/SP. On a 16-layer, 4096-wide frozen stack it reduced peak memory by the expected 0.25 GiB with ~3.9% fwd+bwd overhead under `torch.no_grad()` initialization.

Would you prefer this as a contribution to this PR, or as a separate frozen-only PR based on current main? Happy to share the patch and test details.

PR #2177, opened 2025-09-16
[Core][PyTorch] NVFP4 recipe

ksivaman:
# Description

Add support for the NVFP4 training recipe.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Adds support for NVFP4 block scaling training recipe, including the required kernels and framework infrastructure.

# Checklist:


transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -0,0 +1,859 @@
...
+            columnwise_scale_inv=columnwise_scale_inv,
+            amax_rowwise=amax_rowwise,
+            amax_columnwise=amax_columnwise,
+            fp4_dtype=self.dtype,
+            quantizer=self,
+            requires_grad=requires_grad,
+        )
+
+    def calibrate(self, tensor: torch.Tensor) -> None:
+        # TODO(ksivamani): No calibration?
```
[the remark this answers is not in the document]
timmoon10:
```suggestion
```

transformer_engine/common/common.h
```diff
@@ -174,6 +183,26 @@ struct Tensor {
...
+            for (size_t i = 1; i < columnwise_data.shape.size() - 1; i++) {
+              ret.push_back(columnwise_data.shape[i]);
+            }
+            ret.push_back(columnwise_data.shape[columnwise_data.shape.size() - 1] * 2);
+            ret.push_back(columnwise_data.shape.front());
+          }
+          return ret;
+        } else {
+          auto data_shape_nvfp4 = data.shape;
+          data_shape_nvfp4.back() *= 2;
```
phu0ngng: Hi, 
Could we make sure to move this shape multiplication to TE/PyTorch before merging this PR? @ptrendx @ksivaman

phu0ngng (dismissed):

transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu
```diff
@@ -0,0 +1,792 @@
...
+  NVTE_CHECK(!pow2_scale, "No support for pow2_scale for MXFP4 for now");
+
+  if (!return_identity && !return_transpose) {
+    return;
+  }
+
+  if (use_2d_quantization && !return_identity) {
+    return;
+  }
+
```
jberchtold-nvidia, line 726: This function is missing shape checks on the input tensors

jberchtold-nvidia (changes requested):

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1116,4 +1143,499 @@ std::vector<size_t> MXFP8Quantizer::get_scale_shape(const std::vector<size_t>& s
...
+                              const std::optional<TensorWrapper>& noop_flag) {
+  if (input.numel() == 0) {
+    return;
+  }
+  QuantizationConfigWrapper quant_config;
+  if (noop_flag) {
+    quant_config.set_noop_tensor(noop_flag->data());
+  }
+  quant_config.set_nvfp4_2d_quantization(this->with_2d_quantization);
+  quant_config.set_stochastic_rounding(this->stochastic_rounding);
```
jberchtold-nvidia, line 1449: @ksivaman was looking thru this with @ptrendx and it looks like we don't set the SR rng_seed or rng_sequence here

transformer_engine/common/common.h
```diff
@@ -261,12 +289,20 @@ struct QuantizationConfig {
   NVTETensor noop_tensor = nullptr;
   Float8BlockScaleTensorFormat float8_block_scale_tensor_format =
       Float8BlockScaleTensorFormat::GEMM_READY;
+  size_t rng_seed = 0;
+  size_t rng_sequence = 0;
```
ptrendx: This is not compatible with cuda graphs and will need to be changed to NVTETensor.

ptrendx (changes requested):

build_tools/utils.py
```diff
@@ -234,15 +234,16 @@ def get_cuda_include_dirs() -> Tuple[str, str]:
...
 @functools.lru_cache(maxsize=None)
 def cuda_archs() -> str:
-    version = cuda_version()
-    if os.getenv("NVTE_CUDA_ARCHS") is None:
-        if version >= (13, 0):
-            os.environ["NVTE_CUDA_ARCHS"] = "75;80;89;90;100;120"
+    archs = os.getenv("NVTE_CUDA_ARCHS")
+    if archs is None:
+        version = cuda_version()
+        if version >= (12, 9):
```
timmoon10: SM 70 is not supported in CUDA 13:
```suggestion
        if version >= (13, 0):
            arch = "75;80;89;90;100;100a;103a;120"
        elif version >= (12, 9):
```

build_tools/utils.py
```diff
@@ -234,15 +234,18 @@ def get_cuda_include_dirs() -> Tuple[str, str]:
...
 @functools.lru_cache(maxsize=None)
 def cuda_archs() -> str:
-    version = cuda_version()
-    if os.getenv("NVTE_CUDA_ARCHS") is None:
+    archs = os.getenv("NVTE_CUDA_ARCHS")
+    if archs is None:
+        version = cuda_version()
         if version >= (13, 0):
-            os.environ["NVTE_CUDA_ARCHS"] = "75;80;89;90;100;120"
+            arch = "75;80;89;90;100;100a;103a;120"
```
timmoon10: Typo:
```suggestion
            archs = "75;80;89;90;100;100a;103a;120"
```

transformer_engine/common/common.h
timmoon10: This is absolutely right, it was a design mistake to represent the shape of FP4 tensors based on the underlying uint8 storage type rather than based on the logical tensor shape.

Unfortunately, that mistake was made very early and it's spread throughout the code base. https://github.com/ksivaman/TransformerEngine-1/commit/f8b2a2d0111d9af690b43bb98ae448d9a430a185 attempts to move the tensor reshape logic to the PyTorch extensions, but we're seeing many test failures. I fear we need to rework more logic, especially when launching NVFP4 kernels.

Deadlines looming over us, I propose we merge as-is and fix this in the 2.9 release.

transformer_engine/common/common.h
timmoon10: I don't think this is possible within deadlines. We should defer to 2.9. We shouldn't make any API-stability commitments for FP4-related features in this release.

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -0,0 +1,896 @@
...
+    Used for random Hadamard transform.
+
+    """
+    mask = 0
+    for i, v in enumerate(vector):
+        mask |= (v == -1) << i
+    return mask
+
+
+def get_wgrad_sign_vector() -> torch.Tensor:
```
tsengalb99, line 47: Is this the only source of "randomness" for the RHT or is this a python sanity check function? If you're actually using this to generate signs, surely you could do `(torch.rand(16) > 0).float()*2 - 1` instead without adding too many ops?

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -0,0 +1,896 @@
...
+@functools.lru_cache(maxsize=None)
+def get_rht_matrix(with_random_sign_mask: bool) -> torch.Tensor:
+    """Construct matrix used in random Hadamard transform."""
+    hadamard_dimension = 16
+    if with_random_sign_mask:
+        signs = get_wgrad_sign_vector()
+    else:
+        signs = get_no_random_sign_vector()
+    sign_matrix = signs * torch.eye(hadamard_dimension, dtype=torch.float32)
+    rht_matrix = sign_matrix @ get_hadamard_matrix(hadamard_dimension)
```
tsengalb99, line 98: you can write this as `get_hadamard_matrix(16) * signs`, which should avoid manifesting a 16x16 identity matrix.

transformer_engine/common/util/dequantize_kernels.cuh
```diff
@@ -331,6 +334,83 @@ void mxfp8_dequantize(const Tensor &input, Tensor *output, cudaStream_t stream)
...
+    out.data.elt[0] = static_cast<OType>(current.x * final_scale);
+    out.data.elt[1] = static_cast<OType>(current.y * final_scale);
+    out.data.elt[2] = static_cast<OType>(current.z * final_scale);
+    out.data.elt[3] = static_cast<OType>(current.w * final_scale);
+    output_vec[my_output_index + i] = out;
+  }
+}
+#endif  // CUDA_VERSION
+
+void fp4_dequantize(const Tensor &input, Tensor *output, cudaStream_t stream) {
```
zhongbozhu, line 378: Technically we should ban the dequantize if the quantized data experienced RHT, we will have to de-RHT and then we can dequantize, even though that's not really happening since technically dequantize is only useful for weights, which we are not choosing to apply RHT for now.

timmoon10: Printing a warning would be a good idea.

transformer_engine/pytorch/tensor/nvfp4_tensor.py
zhongbozhu: Yes, we are not really using random generated RHT matrix for every input micro-batch and already found RHT useful in loss curves and evals. 

In terms of the sign vector, there are two possible inputs to the hadamard kernels: either a 16x16 matrix directly or a uint16 value which is then populated as matrix in registers. Using 16x16 matrix directly is an optimization in the fusion path. TLDR, this sign vector is not really passed into the kernel, but will be converted to a uint16 in this case. 
https://github.com/ksivaman/TransformerEngine-1/blob/nvfp4_recipe/transformer_engine/pytorch/tensor/nvfp4_tensor.py#L152-L153

transformer_engine/pytorch/tensor/_internal/nvfp4_tensor_base.py
```diff
@@ -30,12 +43,41 @@ def forward(
...
     ) -> torch.Tensor:
         # pylint: disable=missing-function-docstring
-        dtype = torch_to_transformer_engine_dtype[dtype]
 
-        # Make sure FP8 data is in expected format
+        # Dequantize row-wise data
         if tensor._rowwise_data is not None:
-            return tex.dequantize(tensor, dtype)
-        raise NotImplementedError("Casting back from the transpose not implemented yet!")
+            ### TODO(tmoon): Debug dequantize kernel and remove unfused impl
```
zhongbozhu, line 49: Why we are not using the dequantize cuda kernel here?

transformer_engine/pytorch/fp8.py
```diff
@@ -1146,74 +1146,46 @@ def __init__(
...
-                                stochastic_rounding=self.recipe.fp4_quant_fwd_inp.stochastic_rounding,
-                            ),
-                        ]
-                        for _ in range(self.num_quantizers // 3)
-                    ]
+
+            def _make_quantizer(idx: int) -> NVFP4Quantizer:
+                qparams = (
+                    self.recipe.fp4_quant_fwd_weight
+                    if idx % 3 == 1
```
zhongbozhu, line 1162: Will this break the future grouped linear support? So then 6 quantizers could mean 3 experts for fused ops but 2 experts for legacy logic because we used to assume a place holder quantizer for gemm output.

transformer_engine/common/util/dequantize_kernels.cuh
zhongbozhu: Warnings tend to get ignored though, we should do runtime error, better inside python code not a runtime error raised in C++ layer.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1116,4 +1143,499 @@ std::vector<size_t> MXFP8Quantizer::get_scale_shape(const std::vector<size_t>& s
...
+        NVTE_SCOPED_GIL_RELEASE({
+          nvte_quantize_v2(rht_output_t_cpp.data(), out_transpose.data(), quant_config, stream);
+        });
+      } else {
+        // RHT cast fusion kernel.
+        NVTE_CHECK(this->rht_matrix.defined() && this->rht_matrix.numel() > 0,
+                   "RHT matrix is not set");
+        auto rht_matrix_nvte = makeTransformerEngineTensor(this->rht_matrix);
+        NVTE_SCOPED_GIL_RELEASE({
+          nvte_hadamard_transform_cast_fusion_columnwise(
```
jberchtold-nvidia, line 1628: @ksivaman @ptrendx I don't think we use `this->rht_matrix_random_sign_mask_t` for the RHT-cast fusion case anymore. 

Previously `rht_matrix_random_sign_mask_t` was on the quant_config, but looks like it has been moved out to a separate arg. For the unfused RHT cast case above, `nvte_hadamard_transform` does take in `this->rht_matrix_random_sign_mask_t` as an explicit arg, but the fused API `nvte_hadamard_transform_cast_fusion_columnwise` does not seem to take this as an input. Is this intended?

zhongbozhu: Yes, because the nvte_hadamard_transform_cast_fusion_columnwise prefers taking a 16x16 matrix directly for max perf.

jberchtold-nvidia: Got it, thanks! So this 16x16 matrix should have the random sign mask pre-multiplied into the hadamard matrix directly?

jberchtold-nvidia: @zhongbozhu can we document this assumption that the hadamard matrix is pre-multiplied with the random sign mask? I didn't see it mentioned in the docstring for this function.

From testing this function with a pre-multiplied hadamard matrix I am now getting the correct results that match my reference impl, tho please lmk if my interpretation of this is not correct. Thanks!

transformer_engine/pytorch/fp8.py
timmoon10: This matches the current behavior where we assume that the forward quantizers follow the pattern `input, weight, output, input, weight, output, ...`, so I don't expect problems with grouped linear.

Not quite relevant, but for completeness, I should bring up that I don't like how modules own the output quantizer (https://github.com/NVIDIA/TransformerEngine/pull/2039#discussion_r2271511000). https://github.com/NVIDIA/TransformerEngine/pull/2039 introduces the idea of "quantizer roles", which may provide a more flexible way for modules to specify quantizer types without forcing them to cycle in a specific order.

charmin161:
Is there any plan supporting transformer_engine.common.recipe.NVFP4BlockScaling ?
Since the megatron lm has such assert:
image

qijiaxing:
> Is there any plan supporting transformer_engine.common.recipe.NVFP4BlockScaling ? Since the megatron lm has such assert: image

this PR does support `NVFP4BlockScaling`, see 
https://github.com/NVIDIA/TransformerEngine/blob/9981058f30b5ad2018a817c609a0287a49026626/transformer_engine/common/recipe/__init__.py#L382
You can use this PR in latest Megatron.

transformer_engine/pytorch/tensor/_internal/nvfp4_tensor_base.py
timmoon10: The dequantize kernel fails because it assumes FP4 tensors to match the logical tensor dims, while we are currently using the uint8 tensor dims (see https://github.com/NVIDIA/TransformerEngine/pull/2177#discussion_r2355415532).

transformer_engine/common/util/dequantize_kernels.cuh
timmoon10: In the core C++ lib, I don't think we should treat RHT any different than any other activation function. There's nothing special that makes it different than GeLU or LayerNorm or whatever. The result is an NVFP4 tensor, and consumers of that data don't need to care about how that data was generated.

In the Python level, I think there's a good argument for erroring out. The RHT logic is part of the quantizer, so the quantizer might mark that the NVFP4 tensor can't be dequantized naively.

transformer_engine/common/common.h
ptrendx: It was fixed in the latest version of the pr, resolving.

transformer_engine/common/common.h
timmoon10: Fixed in 9dc682b345256c1242f35ddb2d20fca7c12e24ae.

transformer_engine/pytorch/csrc/quantizer.cpp
timmoon10: Fixed in 8ce814023468b8e2e7cf469d985e7fb2d6c6b2cd.

tests/pytorch/distributed/run_numerics.py
```diff
@@ -100,13 +116,14 @@ def main(argv=None, namespace=None):
         SEQ_LEN = 32
         BATCH_SIZE = 32
         HIDDEN_SIZE = 128
-    elif QUANTIZATION == "fp8_block_scaling":
-        SEQ_LEN = 128
-        BATCH_SIZE = 128
-        HIDDEN_SIZE = 512
+    elif QUANTIZATION == "fp8_block_scaling" or QUANTIZATION == "nvfp4":
```
zhongbozhu: Why are we using a smaller shape for DSv3 shape here? 
We need a larger shape for test the correctness of FP8 AG of DSv3 recipe.

tests/cpp/operator/test_cast_nvfp4_transpose.cu
```diff
@@ -0,0 +1,741 @@
...
+    for (size_t i = 0; i < rows; ++i) {
+        for (size_t j = 0; j < cols; ++j) {
+            const size_t idx = i * cols + j;
+            amax = fmaxf(amax, static_cast<float>(input_dptr[idx]));
+        }
+    }
+    // Set 2nd stage NVFP4 scaling factor
+    output.set_scale(amax);
+
+    bool use_2d_quantization = false;
```
jberchtold-nvidia, line 570: @ptrendx @timmoon10 do we test 2D quantization anywhere? Afaict these cpp unit tests look like they have support for testing 2d along with a reference 2d impl, but this flag is always false so it won't run 2D

timmoon10 (dismissed):

muse-coder:
I encountered an unsupported operator issue when using the current PR for nvfp4 training. The scenario is training an MoE (Mixture of Experts) model, and the error message is: "Unable to find suitable cublas gemm algorithm."
It seems the problem occurs with the following descriptors:
Adesc = [type=R_4F_E2M1 rows=1712 cols=1408 ld=1712]
Bdesc = [type=R_4F_E2M1 rows=1712 cols=2048 ld=1712]
Cdesc = [type=R_32F rows=1408 cols=2048 ld=1408]
Ddesc = [type=R_32F rows=1408 cols=2048 ld=1408]
preference = [maxWavesCount=0.0 maxWorkspaceSizeinBytes=33554428]
computeDesc = [computeType=COMPUTE_32F scaleType=R_32F pointerMode=MODE_DEVICE transa=OP_T aScalePointer=0x7f84ed9d9000 bScalePointer=0x7f6e49fa2400 aScaleMode=VEC16_UE4M3 bScaleMode=VEC16_UE4M3].
error

zhongbozhu:
@muse-coder MoE support will be the next PR :)

transformer_engine/pytorch/csrc/extensions/normalization.cpp
```diff
@@ -123,7 +123,8 @@ std::vector<py::object> layernorm_fwd(py::handle input, py::handle weight, Maybe
     if (nvfp4_quantizer_cpp->with_rht && nvfp4_quantizer_cpp->with_post_rht_amax) {
       // Post-RHT amax is handled within NVFP4 quantizer
       impl = Impl::UNFUSED;
-    } else {
+    } else if (!transformer_engine::getenv<bool>("NVTE_NORM_FWD_USE_CUDNN")) {
+      // TE kernel supports amax output
```
zhongbozhu, line 128: I kinda remember CUDNN layernorm also supports outputting AMAX right? I could be wrong.

transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
```diff
@@ -0,0 +1,836 @@
...
+  auto tma_load_b =  make_tma_copy_B_sm100(
+        SM90_TMA_LOAD{},
+        tensorB,
+        sB(_,_,_,0),
+        cga_tile_shape,
+        mma);
+
+  // Assert checks on tile sizes -- no predication
+  assert(M % size<0>(cga_tile_shape) == 0);
+  assert(N % size<1>(cga_tile_shape) == 0);
```
jberchtold-nvidia: Can we use NVTE_CHECK here? C asserts will be disabled unless we're in a debug build, right?

jberchtold-nvidia (changes requested):

transformer_engine/pytorch/csrc/extensions/normalization.cpp
timmoon10: For FP8, but I'm not aware they support it for BF16.

phu0ngng: They don't for BF16. We should put in a feature request.

yash3056:
does nvfp4 only work with sm100 or all arch above sm100?

PR #2183, opened 2025-09-17
[Draft] Local-Amax for Current-Scaling

mingxu1067:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/normalization.py
+++ b/transformer_engine/jax/cpp_extensions/normalization.py
@@ -886,6 +886,8 @@ def layernorm_fwd(
     zero_centered_gamma: bool,
     epsilon: float,
     quantizer: Optional[Quantizer],
+    amax_across_tpsp=False,
+    amax_across_fsdp=False,
 ) -> tuple[Union[jnp.ndarray, ScaledTensor], jnp.ndarray, jnp.ndarray]:
     """Layer normalization forward pass with optional quantization.
 
@@ -899,6 +901,8 @@ def layernorm_fwd(
         zero_centered_gamma: If True, gamma is zero-centered.
         epsilon: Small constant for numerical stability.
         quantizer: Optional quantizer for FP8 quantization of the output.
+        amax_across_tpsp: Indicate if running all-reduce along TP/SP mesh axes for amax. Only works when using current-scaling. Default is False.
+        amax_across_fsdp: Indicate if running all-reduce along FSDP mesh axes for amax. Only works when using current-scaling. Default is False.
 
     Returns:
         A tuple containing:
@@ -958,7 +962,14 @@ def layernorm_fwd(
             epsilon=epsilon,
             quantizer=None,
         )
-        out, _ = _quantize_dbias_impl(out, is_dbias=False, quantizer=quantizer, dq_dtype=x.dtype)
+        out, _ = _quantize_dbias_impl(
+            out,
+            is_dbias=False,
+            quantizer=quantizer,
+            dq_dtype=x.dtype,
+            amax_across_tpsp=amax_across_tpsp,
+            amax_across_fsdp=amax_across_fsdp,
+        )
         return out, mu, rsigma
 
     is_2x2x = quantizer.is_2x2x()
@@ -1088,6 +1099,8 @@ def rmsnorm_fwd(
     zero_centered_gamma: bool,
     epsilon: float,
     quantizer: Optional[Quantizer],
+    amax_across_tpsp=False,
+    amax_across_fsdp=False,
 ) -> tuple[Union[jnp.ndarray, ScaledTensor], jnp.ndarray]:
     """Root mean square normalization forward pass with optional quantization.
 
@@ -1099,6 +1112,8 @@ def rmsnorm_fwd(
         zero_centered_gamma: If True, gamma is zero-centered.
         epsilon: Small constant for numerical stability.
         quantizer: Optional quantizer for FP8 quantization of the output.
+        amax_across_tpsp: Indicate if running all-reduce along TP/SP mesh axes for amax. Only works when using current-scaling. Default is False.
+        amax_across_fsdp: Indicate if running all-reduce along FSDP mesh axes for amax. Only works when using current-scaling. Default is False.
 
     Returns:
         A tuple containing:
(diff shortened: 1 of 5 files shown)


transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -532,6 +537,134 @@ class QuantizePrimitive(BaseDBiasQuantizePrimitive):
...
 
 
+class AmaxCalculationPrimitive:
+    """
+    Amax Calculation Primitive with custom_partitioning
+    """
+
+    name = "jax_local_amax"
+    multiple_results = False
+    impl_static_args = (1, 2)  # use_global_amax
```
jberchtold-nvidia: Static arg names should be `# amax_across_tpsp, amax_across_fsdp`, right?

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -532,6 +537,134 @@ class QuantizePrimitive(BaseDBiasQuantizePrimitive):
...
+                gmesh = global_mesh_resource()
+                amax = lax_paral_op(amax, jax.lax.pmax, gmesh.fsdp_resource, mesh)
+
+            return amax
+
+        arg_shardings = tuple(arg_i.sharding for arg_i in arg_infos)
+        return mesh, sharded_impl, amax_sharding, arg_shardings
+
+
+def register_primitive_only_outter(cls):
```
jberchtold-nvidia: Self-resolve: `outter` -> `outer`

transformer_engine/jax/layernorm_mlp.py
```diff
@@ -272,13 +272,12 @@ def _layernorm_mlp_fwd_rule(
         epsilon,
         norm_type,
         quantizer=ffn1_quantizer_set.x,
+        amax_across_tpsp=True,
```
jberchtold-nvidia: Can you help me understand why `amax_across_tpsp` is being set to `True` in these cases? Is this because we will be all-gathering these GEMM inputs?

transformer_engine/jax/dense.py
```diff
@@ -175,13 +183,15 @@ def _dense_fwd_rule(
...
         quantizer=quantizer_set.x,
+        amax_across_tpsp=using_global_amax_of_x,
     )
     casted_x = with_sharding_constraint_by_logical_axes(casted_x, input_axes)
 
     casted_kernel = tex.quantize(
         kernel,
         flatten_axis=flatten_axis_k,
         quantizer=quantizer_set.kernel,
+        amax_across_fsdp=True,
```
jberchtold-nvidia: Same question here on why `amax_across_fsdp=True` here always? Is that because this will be All-Gathered in the GEMM for FSDP?

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -532,6 +537,134 @@ class QuantizePrimitive(BaseDBiasQuantizePrimitive):
...
+    ):
+        """
+        amax calcuation abstract
+        """
+        del amax_across_tpsp, amax_across_fsdp
+
+        dtype = dtypes.canonicalize_dtype(x_aval.dtype)
+        assert dtype in [jnp.float32, jnp.float16, jnp.bfloat16]
+
+        out_aval = jax.core.ShapedArray(shape=(1,), dtype=jnp.float32)
```
jberchtold-nvidia, line 571: This is only storing a single-value as the global shape, so from JAX/XLA's perspective it will think this is the same value replicated across devices, rather than a unique value per device, so things could break easily if we have any resharding, right?

@phu0ngng last we synced I thought we were going with your idea for an amax tensor on each device that has the shape of the mesh. Did this plan change? Thanks!

jberchtold-nvidia: Resolved per offline discussion on plan

transformer_engine/jax/cpp_extensions/quantization.py
phu0ngng: Hi, may I know why we can't simply use the existing registration?

transformer_engine/jax/layernorm_mlp.py
mingxu1067: Yes, as the recipe suggested, we only need to run activation AR for those run AG right before GEMM, which is colmun-parallel layers in SP.

transformer_engine/jax/dense.py
mingxu1067: Yes, it is because we AG the weight then run the GEMM, therefore GEMM would need weight scale to unique one across all devices.

transformer_engine/jax/cpp_extensions/quantization.py
mingxu1067: Makes sense. Added an arg to contol the scope of registration.

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -1099,6 +1108,7 @@ def rmsnorm_fwd(
         zero_centered_gamma: If True, gamma is zero-centered.
         epsilon: Small constant for numerical stability.
         quantizer: Optional quantizer for FP8 quantization of the output.
+        amax_scope: Indicate the scope to run amax calculation. Default is AmaxScope.LOCAL.
```
jberchtold-nvidia: Can we expand this comment to indicate that this arg will have no effect if we are not using a form of current scaling?

transformer_engine/jax/dense.py
```diff
@@ -238,6 +249,7 @@ def _dense_bwd_rule(
         is_dbias=use_bias,
         flatten_axis=flatten_axis_k,
         quantizer=quantizer_set.dgrad,
+        amax_scope=AmaxScope.TPSP,
```
phu0ngng: Hi, 
I think we only do AG the `grad` only if the input in the FWD is not AG.
```suggestion
        amax_scope=AmaxScope.LOCAL if using_global_amax_of_x else AmaxScope.TPSP,
```

transformer_engine/jax/cpp_extensions/normalization.py
mingxu1067: Done

transformer_engine/jax/dense.py
mingxu1067: Thank you for pointing out. Fixed.

PR #2184, opened 2025-09-17
Fix incorrect TP rank calculation when using data parallel

djns99:
# Description

Fixes the issue with TP + DP and communication overlap where DP ranks are not properly offset

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add world rank parameter to UB communicator to offset the send/recv rank correctly

# Checklist:


timmoon10 (changes requested):
We can't easily change `release_v2.7` at this point.

PR #2191, opened 2025-09-21
FSDP grad fusion support

sanandaraj5597:
This PR adds support gradient fusion for MCore FSDP.

### DIFF
--- a/transformer_engine/pytorch/module/grouped_linear.py
+++ b/transformer_engine/pytorch/module/grouped_linear.py
@@ -402,7 +402,7 @@ class _GroupedLinear(torch.autograd.Function):
                     use_bias=ctx.use_bias if grad_biases[0] is None else None,
                     bias=biases,
                     use_split_accumulator=wgrad_gemm_use_split_accumulator,
-                    accumulate=accumulate_wgrad_into_param_main_grad,
+                    accumulate=accumulate_wgrad_into_param_main_grad if not hasattr(weights[0], "__fsdp_param__") else False,
                 )
                 # WGRAD
                 if ctx.wgrad_store is not None and ctx.wgrad_store.delay_wgrad_compute():
--- a/transformer_engine/pytorch/ops/basic/basic_linear.py
+++ b/transformer_engine/pytorch/ops/basic/basic_linear.py
@@ -997,6 +997,7 @@ class BasicLinear(BasicOperation):
             weight_param = self.weight
             if hasattr(weight_param, "__fsdp_param__"):
                 weight_param.main_grad = weight_param.get_main_grad()
+                accumulate_into_main_grad = False
             if not hasattr(weight_param, "main_grad"):
                 raise RuntimeError(
                     "BasicLinear op is configured with "
--- a/transformer_engine/pytorch/ops/fused/backward_linear_add.py
+++ b/transformer_engine/pytorch/ops/fused/backward_linear_add.py
@@ -59,6 +59,7 @@ class BackwardLinearAdd(FusedOperation):
             weight_param = linear_op.weight
             if hasattr(weight_param, "__fsdp_param__"):
                 weight_param.main_grad = weight_param.get_main_grad()
+                accumulate_into_main_grad = False
             if not hasattr(weight_param, "main_grad"):
                 raise RuntimeError(
                     "BasicLinear op is configured with "
--- a/transformer_engine/pytorch/ops/fused/backward_linear_scale.py
+++ b/transformer_engine/pytorch/ops/fused/backward_linear_scale.py
@@ -60,6 +60,7 @@ class BackwardLinearScale(FusedOperation):
             weight_param = linear_op.weight
             if hasattr(weight_param, "__fsdp_param__"):
                 weight_param.main_grad = weight_param.get_main_grad()
+                accumulate_into_main_grad = False
             if not hasattr(weight_param, "main_grad"):
                 raise RuntimeError(
                     "BasicLinear op is configured with "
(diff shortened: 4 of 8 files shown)


timmoon10:
The root problem is that Mcore DDP and FSDP have different behaviors and require different contracts with TE:
- DDP uses persistent `main_grad` buffers and it expects TE to accumulate into it. To adhere to this contract, Mcore zeros out the `main_grad` before the first microbatch step.
- FSDP uses temporary `main_grad` buffers and it expects TE to overwrite it.

I don't like this PR's approach of switching between these two cases based on whether Mcore is using DDP or FSDP, since that's not actually the important thing. It also needlessly blocks some possible optimizations (DDP might want to overwrite `main_grad`s in the first microbatch, FSDP might want to accumulate into `main_grad`s if a weight is shared).

There are a few possible redesigns:

1. Deprecate the `fuse_wgrad_accumulation` kwarg in favor of something like `output_wgrad_to_main_grad`. Then check param flags to decide whether to overwrite or accumulate into the `main_grad`:
```python
    grad_weight: torch.Tensor
    accumulate: bool = False
    if output_wgrad_to_main_grad:
        if getattr(weight, "get_main_grad", None) is not None:
            grad_weight = weight.get_main_grad()
        else:
            grad_weight = weight.main_grad
        accumulate = getattr(weight, "_overwrite_main_grad", True)
    else:
        grad_weight = torch.empty(...)

    gemm(..., out=grad_weight, accumulate=accumulate)
```
Ensuring backward compatibility will be tricky.

2. Have separate kwargs for `fuse_wgrad_accumulation` and `overwrite_wgrad_main_grad`. This means that the two cases are separate code paths and backward compatibility is easier to maintain. However, it also means we can't change behavior between steps.
3. Keep the `fuse_wgrad_accumulation` kwarg and purely control behavior with param flags. This is basically the approach used in this PR, although it could be improved by using better names rather than just checking `weight.__fsdp_param__`. One problem is that `fuse_wgrad_accumulation` will no longer be an accurate name.

transformer_engine/pytorch/ops/basic/basic_linear.py
```diff
@@ -1019,6 +1019,7 @@ def op_backward(
             weight_param = self.weight
             if hasattr(weight_param, "__fsdp_param__"):
                 weight_param.main_grad = weight_param.get_main_grad()
+                accumulate_into_main_grad = not hasattr(weight_param, "overwrite_main_grad")
```
timmoon10: Accumulating into `main_grad` should be orthogonal to FSDP:
```suggestion
            accumulate_into_main_grad = not hasattr(weight_param, "overwrite_main_grad")
```

transformer_engine/pytorch/ops/fused/backward_linear_add.py
```diff
@@ -59,6 +59,7 @@ def fuser_backward(
             weight_param = linear_op.weight
             if hasattr(weight_param, "__fsdp_param__"):
                 weight_param.main_grad = weight_param.get_main_grad()
+                accumulate_into_main_grad = not hasattr(weight_param, "overwrite_main_grad")
```
timmoon10:
```suggestion
            accumulate_into_main_grad = not hasattr(weight_param, "overwrite_main_grad")
```

transformer_engine/pytorch/ops/fused/backward_linear_scale.py
```diff
@@ -60,6 +60,7 @@ def fuser_backward(
             weight_param = linear_op.weight
             if hasattr(weight_param, "__fsdp_param__"):
                 weight_param.main_grad = weight_param.get_main_grad()
+                accumulate_into_main_grad = not hasattr(weight_param, "overwrite_main_grad")
```
timmoon10:
```suggestion
            accumulate_into_main_grad = not hasattr(weight_param, "overwrite_main_grad")
```

transformer_engine/pytorch/ops/fused/userbuffers_backward_linear.py
```diff
@@ -523,6 +523,7 @@ def fuser_backward(
             weight_param = linear_op.weight
             if hasattr(weight_param, "__fsdp_param__"):
                 weight_param.main_grad = weight_param.get_main_grad()
+                accumulate_into_main_grad = not hasattr(weight_param, "overwrite_main_grad")
```
timmoon10:
```suggestion
            accumulate_into_main_grad = not hasattr(weight_param, "overwrite_main_grad")
```

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -402,7 +402,11 @@ def backward(ctx, grad_output: torch.Tensor) -> Tuple[Union[torch.Tensor, None],
                     use_bias=ctx.use_bias if grad_biases[0] is None else None,
                     bias=biases,
                     use_split_accumulator=wgrad_gemm_use_split_accumulator,
-                    accumulate=accumulate_wgrad_into_param_main_grad,
+                    accumulate=(
+                        accumulate_wgrad_into_param_main_grad
+                        if not hasattr(weights[0], "overwrite_main_grad")
```
timmoon10: What if the user sets `overwrite_main_grad=False`?
```suggestion
                        if not getattr(weights[0], "overwrite_main_grad", False)
```

sanandaraj5597: I was thinking it should be okay if the attribute is present instead of having a value. Do you want to check the value?

PR #2194, opened 2025-09-22
[JAX] Clamped Swiglu Integration

vthumbe1503:
# Description

Jax integration for clamped swiglu. This is the continuation of [PR](https://github.com/NVIDIA/TransformerEngine/pull/2161) which added Clamped Swiglu(used in GPT OSS) support in TE along with Pytorch integration. This PR hooks up the clamped swiglu and dswiglu's nvte APIs to TE Jax.

## Type of change

- [ ] Documentation change (change only to the documentation)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes
- TE Common: Adds addition enum value for NVTE_Activation_Type for Clamped Swiglu since that enum is py-binded to JAX Python/
-  TE Common: Gated Kernel Functions for Clamped Swiglu used parameters as pass by reference. However pass by value would be more efficient since our parameter structs are either Empty or too small.
- TE JAX: Passing a generic nested dict named ActivationParams which at the moment just have parameters for clamped Swiglu. It is passed all the way down to XLA FFI Interfaces for Activation and Dactivation in python. From C++ side there is a nested struct for Activation Params which is mapped to the parameters received from XLA FFI interface.
- TE JAX reference implementation: Also adds reference implementation for clamped swiglu using jax.nn APIs
- TE JAX unit test: Modifies the unit test to also check the correctness of TE JAX implementation by including clamped swiglu in the activation list. Also modifies the test function's interfaces to accept activation parameters since previously none of the activations accepted parameters but clamped swiglu does.

# Checklist:


tests/jax/test_custom_call_compute.py
```diff
@@ -274,9 +294,13 @@ def test_act_forward_with_tensor_scaling_fp8(
             q_layout=q_layout,
         )
 
-        te_output = tex.act_lu(x, activation_type, te_quantizer)
-        jax_output = _jax_act_lu(x, activation_type, jax_quantizer)
-
+        act_params = (
+            ClampedSwigluParams.create(limit=1.0, alpha=1.702)
```
jberchtold-nvidia: Is there a particular reason we use a different limit for fp8 tensor scaling here and block scaling below? Or just to test a variety of limit values?

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -13,9 +13,10 @@
 from jax.experimental.custom_partitioning import SdyShardingRule
 from jax.sharding import PartitionSpec
 
+import numpy as np
 import transformer_engine_jax
 from transformer_engine_jax import NVTE_Activation_Type
-
+from dataclasses import dataclass
```
jberchtold-nvidia: nit self-resolve: Can we move this dataclass import up above the TE/JAX imports?

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -56,17 +57,45 @@
...
 
 
-def _convert_to_activation_function(fn_or_string):
+@dataclass(frozen=True)
+class ClampedSwigluParams:
+    limit: float = 7.0
+    alpha: float = 1.702
+
+    @staticmethod
+    def create(limit: float = 7.0, alpha: float = 1.702):
```
jberchtold-nvidia: Style: Do we need a `create` method here? With the default values specified above this would be identical to calling the `ClampedSwigluParams(limit=X, alpha=Y)`, right?

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -56,17 +57,45 @@
...
+        return ClampedSwigluParams(limit=limit, alpha=alpha)
+
+    def __hash__(self):
+        return hash(self.limit)
+
+    def __eq__(self, value):
```
jberchtold-nvidia: I think for dataclasses they will automatically generate an `__eq__` method, assuming your individual fields all have `__eq__` methods, which the float numbers will have

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -180,12 +217,14 @@ def impl(
...
         scale_dtype,
         is_outer,
+        act_params,
     ):
         """
         to describe implementation
         """
         del is_outer
         assert ActLuPrimitive.inner_primitive is not None
+        import numpy as np
```
jberchtold-nvidia: nit self-resolve: Remove numpy import here

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -236,6 +277,7 @@ def batcher(
         x, scale = batched_args
         x_bdim, scale_bdim = batch_dims
         amax_bdim = scale_bdim
+        import numpy as np
```
jberchtold-nvidia: nit self-resolve: remove numpy import here

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -611,10 +663,13 @@ def impl(
         act_enum,
         act_len,
         is_outer,
+        act_params,
     ):
         """
         te_dact_dbias_quantize_p impl
         """
+        import numpy as np
```
jberchtold-nvidia: nit self-resolve: Remove numpy import

transformer_engine/jax/activation.py
```diff
@@ -16,12 +16,14 @@
 
 from .quantize.tensor import NoScaleTensor
 from .quantize.quantizer import Quantizer
+from .cpp_extensions.activation import ClampedSwigluParams
```
jberchtold-nvidia: I'm hesitant to require users to important anything in `cpp_extensions` for our API as we like to keep some of that complexity hidden and abstracted from the user. 

What do you think about moving `ClampedSwigluParams` into this outer `activation.py` and we can import it in the cpp extensions from here? I'm pretty sure this won't result in a cyclic import as we don't depend on anything in cpp extensions in this file at import-time, only when the functions are called

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -1045,13 +1116,13 @@ def act_lu(
     if quantizer.scaling_mode == ScalingMode.CURRENT_TENSOR_SCALING:
         # Current scaling does not support fused operations. Perform dact in higher precision then quantize after.
         out = act_lu(
-            x=x,
+            x=x.astype(jnp.float32),
```
jberchtold-nvidia: This diff might be a rebase issue, we recently merged a PR to avoid these float32 casts as a performance improvement so we should make sure this PR isn't adding these back in.

transformer_engine/jax/csrc/extensions/activation.cpp
```diff
@@ -18,7 +18,10 @@ Error_Type ActLuFFI(cudaStream_t stream, Buffer_Type input_buf, Buffer_Type scal
                     Result_Type output_buf, Result_Type colwise_output_buf,
                     Result_Type scale_inv_buf, Result_Type colwise_scale_inv_buf,
                     Result_Type amax_buf, int64_t act_enum, JAXX_Scaling_Mode scaling_mode,
-                    bool is_2x_int) {
+                    bool is_2x_int, ClampedSwigluConfig act_params) {
+  // parameters for clamped swiglu used in GPT OSS
+  auto swiglu_limit = act_params.limit;
```
jberchtold-nvidia: nit self-resolve: What do you think about moving this struct unwrapping into the `case NVTE_Activation_Type::CLAMPED_SWIGLU` block right before the call to `nvte_clamped_swiglu`?

transformer_engine/jax/csrc/extensions/activation.cpp
```diff
@@ -216,7 +226,11 @@ Error_Type DActLuDBiasQuantizeFFI(cudaStream_t stream, Buffer_Type input_buf,
                                   Result_Type scale_inv_buf, Result_Type colwise_scale_inv_buf,
                                   Result_Type amax_buf, Result_Type dbias_buf,
                                   Result_Type workspace_buf, JAXX_Scaling_Mode scaling_mode,
-                                  int64_t act_enum, bool is_2x, bool is_dbias) {
+                                  int64_t act_enum, bool is_2x, bool is_dbias,
+                                  ClampedSwigluConfig act_params) {
+  // parameters for clamped swiglu used in GPT OSS
+  auto swiglu_limit = act_params.limit;
+  auto swiglu_alpha = act_params.alpha;
```
jberchtold-nvidia: nit self-resolve: Same suggestion as above about moving this struct unwrapping into the case block for clamped swiglu

transformer_engine/jax/flax/module.py
```diff
@@ -955,7 +961,8 @@ class LayerNormMLP(TransformerEngineBase):
     bias_axes_1: Tuple[str, ...] = ("act", "mlp")
     bias_axes_2: Tuple[str, ...] = ("embed",)
     return_layernorm_output: bool = True
-    activations: Sequence[Union[str, Callable]] = ("relu",)
+    activations: Sequence[Union[str, Callable]] = (("relu",),)
```
jberchtold-nvidia: Why was this default value changed to wrap the values in an additional tuple?

transformer_engine/jax/flax/module.py
```diff
@@ -955,7 +961,8 @@ class LayerNormMLP(TransformerEngineBase):
...
+    activation_params: dict = (None,)
```
jberchtold-nvidia: Why is this default value `(None,)` instead of `None` without being wrapped in a tuple?

transformer_engine/jax/layernorm_mlp.py
```diff
@@ -203,6 +206,7 @@ def _layernorm_mlp(
         ffn1_ckpt_name,
         ffn2_ckpt_name,
         activation_type,
+        activation_params
```
jberchtold-nvidia: I think this is missing a trailing comma

jberchtold-nvidia (commented):
Overall looks good pending CI, left some comments and questions, mostly minor things. Thanks for updating the Flax modules as well, great work on this PR!

tests/jax/test_custom_call_compute.py
vthumbe1503: Ah, this was just an artifact of testing. Made it consistent everywhere now. Thanks

transformer_engine/jax/cpp_extensions/activation.py
vthumbe1503: Done

transformer_engine/jax/cpp_extensions/activation.py
vthumbe1503: Done

transformer_engine/jax/cpp_extensions/activation.py
vthumbe1503: Removed it now

transformer_engine/jax/cpp_extensions/activation.py
vthumbe1503: Done

transformer_engine/jax/flax/module.py
vthumbe1503: sometimes copilot does things I dont ask to Lol :). Fixed.

transformer_engine/jax/flax/module.py
vthumbe1503: Again copilot's artifact. Fixed.

transformer_engine/jax/layernorm_mlp.py
vthumbe1503: Fixed. Thanks!

transformer_engine/jax/csrc/extensions/activation.cpp
vthumbe1503: decalaration of these variables cant be done inside the switch statement as C++ doesn't allow to. So, would have to keep declaration outside but unwrapping can be moved in. But I guess that might be slightly more unclean than this. So kept it outside so that definition and declaration is done together.

transformer_engine/jax/csrc/extensions/activation.cpp
vthumbe1503: Replied above.

transformer_engine/jax/cpp_extensions/activation.py
vthumbe1503: Done

transformer_engine/jax/cpp_extensions/activation.py
vthumbe1503: Done

transformer_engine/jax/cpp_extensions/activation.py
vthumbe1503: Fixed

transformer_engine/jax/activation.py
vthumbe1503: Great idea! Definitely makes sense and user is expected to use ClampedSwigluParams to pass to TE jax interface so makes sense to keep the Parameter class in the outer activation rather than inner activation.py which is not intended to be exposed to user much.

Made that change and it doesnt result in any circular imports. Thanks!

transformer_engine/jax/csrc/extensions/activation.cpp
jberchtold-nvidia: I think you can if you use braces around the switch case, or if this is the last case in a switch block. But given we just have two params here I don't think we need to add braces for all the switch cases just for this change. So I think it's fine as is, and in the future if we have lots more args we can update the switch cases to have braces around them and move the declarations+unwrapping inside.

For example, this compiled for me
```c++
#include <vector>
#include <cmath>
#include <cstdio>

struct MyStruct {
    float a;
    float b;
};

void my_func(MyStruct const& my_struct, int some_value) {
    switch (some_value) {
        case 0: {
            auto a = my_struct.a;
            auto b = my_struct.b;
            printf("a+b: %f\n", a+b);
            break;
        }
        case 1: {
            auto a = my_struct.a;
            auto b = my_struct.b;
            printf("a-b: %f\n", a-b);
        }
    }
}

int main() {
    my_func(MyStruct{1.f, 2.f}, 0);
    return 0;
}
```

transformer_engine/jax/activation.py
jberchtold-nvidia: Thanks!

transformer_engine/jax/csrc/extensions/activation.cpp
vthumbe1503: Oh interesting! Sure, lets make that change when we have a lot more parameters needed :)

transformer_engine/jax/activation.py
```diff
@@ -8,20 +8,36 @@
...
-
+import numpy as np
 from . import cpp_extensions as tex
 
 from .quantize.tensor import NoScaleTensor
 from .quantize.quantizer import Quantizer
 
 
+@dataclass(frozen=True)
+class ClampedSwigluParams:
```
phu0ngng: Hi, let's move this class to `cpp_extensions.py` and expose it as a part of the `tex`.

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -922,20 +959,23 @@ class DActLuQuantizePrimitive(BaseDActLuDBiasQuantizePrimitive):
     """Subclass of BaseDActLuDBiasQuantizePrimitive for fused activation quantization without dbias. No change in functionality from the base primitive but named differently for use in more granular disabling of primitives via NVTE_JAX_CUSTOM_CALLS."""
 
 
-def _jax_act_lu(inputs, activation_type, quantizer=None) -> Union[NoScaleTensor, ScaledTensor]:
+def _jax_act_lu(
+    inputs, activation_type, quantizer=None, act_params: Optional[ClampedSwigluParams] = None
```
phu0ngng: Hi, should we make a default ActParam instead of None here?

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -980,10 +1023,14 @@ def _jax_quantize_dact_dbias(
     return dx, dbias
 
 
+from dataclasses import dataclass
```
phu0ngng: Move to top.

transformer_engine/jax/csrc/extensions.h
```diff
@@ -38,6 +38,11 @@ XLA_FFI_REGISTER_ENUM_ATTR_DECODING(transformer_engine::jax::JAXX_Scaling_Mode);
 namespace transformer_engine {
 namespace jax {
 
+struct ClampedSwigluConfig {
+  float limit;
+  float alpha;
+};
```
phu0ngng, line 42: Hi, 
Since we will lower this struct for all activations (even if we don't use it), should we have a more general name for it instead?
Something like:
```
struct ActParams {
  float swiglu_limit;
  float swiglu_alpha;
 };
 ```

File: transformer_engine/jax/flax/module.py [hunk 19]
```diff
@@ -898,6 +898,12 @@ class LayerNormMLP(TransformerEngineBase):
     activations: Sequence[Union[str, Callable]], default = ('relu',)
         The sequence of activation functions to apply after the first dense layer transformation.
         Each activation has its own transformation layer.
+    activation_params: dict, default = None
+        The parameters for the ClampedSwiglu activation used in GPT OSS. This is only
+        used when ('clamped_silu', 'clamped_linear') is in :attr:`activations`. At the moment
+        ClampedSwiglu is the only activation that requires parameters. If there is more activation
+        functions that require parameters in the future, we might need to change it to a more gerneric
+        parameter container.
```
phu0ngng: This docstring is for users, so I think we should split this dev note into a TODO.

transformer_engine/jax/activation.py
jberchtold-nvidia: He had it there originally and I told him to move it here. Since this param will be user facing, they'll need to create it themselves and import from `cpp_extensions`. My understanding is that we wanted to encapsulate and hide those internals from users of the higher-level VJPs or Flax layers, so I suggested moving it here.

But do you think it's more important to prevent the dependency from cpp_extensions on this type in the higher-level VJP  even tho users will need to use cpp_extensions types directly? Lmk what you think, thanks!

transformer_engine/jax/csrc/extensions.h
jberchtold-nvidia: Varun made a different version here with a generic outer struct and a nested inner struct for ClampedSwiglu, which I believe is what you have in mind. I haven't reviewed this other version yet, planning to do it today
https://github.com/vthumbe1503/TransformerEngine/pull/2

transformer_engine/jax/activation.py
phu0ngng: The reasons I suggested moving it to the tex are that:
- Users don't really use this activation VJP, so they will still need to import from this file even if they don't use other stuff. 
- We have other Params/Configs that are placed in the tex (i.e., `AmaxScope` is in `cpp_extensions/quantization.py`, `CollectiveOp` is in `cpp_extensions/gemm.py`).

At the same time, I agree that exposing it here may be a cleaner approach. If we decide to keep it here, perhaps we should start a follow-up work on which functions/symbols we want to export from each VJP file so that users can do `import transformer_engine.jax as te_jax`.

jberchtold-nvidia: Sounds good, I'm okay with reverting back to having it in `cpp_extensions` (`tex`) for now then to be consistent with other structures until we decide on a way to organize these structures. Apologies for the churn, Varun

vthumbe1503: No issues @jberchtold-nvidia

transformer_engine/jax/flax/module.py
vthumbe1503: Modified the docstring now. ActivationParams is a generic dict now. So no TODO required

transformer_engine/jax/csrc/extensions.h
vthumbe1503: Have moved the generic approach to this PR itself

transformer_engine/jax/cpp_extensions/activation.py
vthumbe1503: I am actually following generally safer practice of not providing default values other than None for user defined data type in the function definition. Although ActivationParams Object today is immutable, it might be dangerous in future in case it isnt  anymore. Since the default values created are just instantiated and shared during import time. And accidental change of the param in one function can impact the other which can be harder to debug. Although this can be solved with default factory and such a things to happen is very unrealistic, so let me know what you think.

transformer_engine/jax/cpp_extensions/activation.py
vthumbe1503: DOne

transformer_engine/jax/layernorm_mlp.py
```diff
@@ -306,10 +311,18 @@ def _layernorm_mlp_fwd_rule(
     dot_1_output = checkpoint_name(dot_1_output, ffn1_ckpt_name)
 
     # (batch..., hidden_in) -> (batch..., hidden)
+    # At the moment the act_params is only used for ClampedSwiglu
```
jberchtold-nvidia: Can remove this comment as the work to make it a generic container is already done :)

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -923,20 +1023,22 @@ class DActLuQuantizePrimitive(BaseDActLuDBiasQuantizePrimitive):
     """Subclass of BaseDActLuDBiasQuantizePrimitive for fused activation quantization without dbias. No change in functionality from the base primitive but named differently for use in more granular disabling of primitives via NVTE_JAX_CUSTOM_CALLS."""
 
 
-def _jax_act_lu(inputs, activation_type, quantizer=None) -> Union[NoScaleTensor, ScaledTensor]:
+def _jax_act_lu(
+    inputs, activation_type, quantizer=None, act_params: Optional[ActivationParams] = None
```
phu0ngng, line 1027: Hi, 

How about we have a default empty ActivationParams, i.e., `ActivationParams()`, so that we don't need to check for None later?

transformer_engine/jax/csrc/extensions/activation.cpp
```diff
@@ -133,20 +140,23 @@ Error_Type ActLuFFI(cudaStream_t stream, Buffer_Type input_buf, Buffer_Type scal
...
+        .Ret<Buffer_Type>()      // output
+        .Ret<Buffer_Type>()      // colwise output
+        .Ret<Buffer_Type>()      // scale_inv
+        .Ret<Buffer_Type>()      // scale_inv colwise
+        .Ret<Buffer_Type>()      // amax
+        .Attr<int64_t>("act_enum")
+        .Attr<JAXX_Scaling_Mode>("scaling_mode")
+        .Attr<bool>("is_2x")
+        .Attr<ActivationConfig>(
+            "act_params"),  // Can generalize the config later if we have more activations that need params
```
phu0ngng: Old comment.

phu0ngng (approved):
Hi, 
Left a small comment. LGTM otherwise. Thanks

transformer_engine/jax/csrc/extensions/activation.cpp
vthumbe1503: Removed

transformer_engine/jax/cpp_extensions/activation.py
vthumbe1503: Hi Phuong. I am actually following generally safer practice of not providing default values other than None for user defined data type in the function definition. Although ActivationParams Object today is immutable, it might be dangerous in future in case it isnt anymore. Since the default values created are just instantiated and shared during import time. And accidental change of the param in one function can impact the other which can be harder to debug. Although this can be solved with default factory and such a things to happen is very unrealistic, so let me know what you think.

tests/jax/test_custom_call_compute.py
```diff
@@ -170,29 +170,35 @@ def assert_dequantized_grouped_scaled_tensor(
...
     ("squared_relu",),
     ("squared_relu", "linear"),
+    ("clamped_silu", "clamped_linear"),
 ]
 
 ACTIVATION_TYPES = {
     "L0": [
         ("gelu",),
         ("gelu", "linear"),
+        ("clamped_silu", "clamped_linear"),
```
jberchtold-nvidia: Let's keep L0 as just gelu and geglu.  Btw I may have forgot to mention, you can test L1 or L2 locally by setting the env var `NVTE_JAX_UNITTEST_LEVEL="L2"`

transformer_engine/common/util/cast_gated_kernels.cuh
```diff
@@ -924,7 +924,7 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK)
 
 template <bool IS_DGATED, typename ParamOP, float (*ActOP)(float, const ParamOP &),
           float (*DActOP)(float, const ParamOP &)>
-void cast_fp8_gated(const Tensor &grad, const Tensor &gated_input, Tensor *output, ParamOP &p,
```
jberchtold-nvidia, line 927: Why is this change from pass by ref to value needed? I would've assumed the TE common changes would have merged with the PyTorch version of this PR. Curious if there was some change needed here for the JAX bindings, thanks!

transformer_engine/jax/layernorm_mlp.py
vthumbe1503: Done

tests/jax/test_custom_call_compute.py
vthumbe1503: Done

transformer_engine/common/util/cast_gated_kernels.cuh
vthumbe1503: This was interstingly comment from copilot in other JAX PR that I had abandoned. So I had added the change here, but forgot to add the change in the Pytorch PR. Both are equally functional from both Pytorch/JAX Bindings perspective. And the perf implications are not much noticable as well. But given that we have most of our activations consume Empty struct in TE common, passing by reference is definitely an additional overhead for most cases.
And for clamped swiglu as well the struct has two floats which is 8 bytes and still equal to size of a pointer on 64 bit systems.

jberchtold-nvidia: Makes sense, thanks! And I do feel better about it being pass by value than a non-const mutable reference

jberchtold-nvidia (approved):
LGTM, thanks!

PR #2197, opened 2025-09-23
[JAX] Update JAX version requirement in pyproject.toml

phu0ngng:
# Description

We should not require `jax[cuda12]` in general.

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/tests/jax/test_distributed_layernorm.py
+++ b/tests/jax/test_distributed_layernorm.py
@@ -76,8 +76,6 @@ class TestDistributedLayernorm:
             all_reduce_loss_bytes + weight_count * shape[-1] * jax_dtype.itemsize
         )
         other_bytes = 0
-        if fp8_recipe == recipe.Float8CurrentScaling():
-            allreduce_total_bytes += jax_dtype.itemsize  # 1 * dtype for the amax reduction
         return generate_collectives_count(
             allreduce=allreduce_total_bytes * int(is_dp_enabled), allgather=0, other=other_bytes
         )
--- a/transformer_engine/jax/cpp_extensions/activation.py
+++ b/transformer_engine/jax/cpp_extensions/activation.py
@@ -26,7 +26,7 @@ from .misc import (
     should_apply_1x_fused_dbias_war_for_arch_l_100,
     NamedSharding,
 )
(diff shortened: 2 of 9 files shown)


jberchtold-nvidia:
Thanks for making this PR!

pyproject.toml
```diff
@@ -3,7 +3,8 @@
 # See LICENSE for license information.
 
 [build-system]
-requires = ["setuptools>=61.0", "cmake>=3.21", "wheel", "pybind11[global]", "ninja", "pip", "torch>=2.1", "jax[cuda12]", "flax>=0.7.1"]
+requires = ["setuptools>=61.0", "cmake>=3.21", "wheel", "pybind11[global]", "ninja", "pip",
+"torch>=2.1", "jax>=0.5.0", "flax>=0.7.1"]
```
jberchtold-nvidia, line 7: Do we need to use `jax[cuda12]>=0.5.0` instead of `jax>=0.5.0`? The previous version uses jax[cuda12] 

CI is passing but I'm wondering if it passes because `jax[cuda12]` is already present in all our base containers. Other than this one question, the PR LGTM!

phu0ngng: We don't need `jax[cuda12]` in a CUDA 13 image. Right? 
I have no clue why `cuda12` was there in the first place.

jberchtold-nvidia: Latest installation instructions from here say to use `jax[cuda13]` when installing JAX for NVIDIA GPUs
https://docs.jax.dev/en/latest/installation.html

How we handle both cuda 12 and cuda 13, I don't know.

This will likely work as intended in our containers since they all have base JAX + cuda13. But if someone tries to install base jax without jax[cuda13] outside of our containers, it may pass these pyproject.toml check but get an error at runtime that GPU devices aren't supported. Though I'm not sure on this

phu0ngng: Hi, 

I tested and confirmed that with this change, even if we build it in an isolated environment (i.e. `NVTE_FRAMEWORK=jax pip install . -v` without `--no-build-isolation`), there is no issue. All the tests worked as normal, utilizing GPUs.
With this, I think we can merge the PR.

jberchtold-nvidia (approved):

PR #2201, opened 2025-09-25
[JAX] Remove unnecessary SWA calculation in _segment_ids_pos_to_seqlens_offsets()

KshitijLakhani:
# Description

I noticed that in `_segment_ids_pos_to_seqlens_offsets()`, a temp / intermediate mask is created, the purpose of which is only to help with the calculation of seqlens and offsets which is then passed onto the cuDNN backend. 
This intermediate attn_mask should only need to use the segment ids and pos to create the attn_mask followed by a layer of masking based on the application of the type of mask (causal or brcm) as that would influence the final seqlens and offsets generated.
However, SW mask calculation and application to attn_mask in no way should affect the final seqlens and offsets generated hence this is an unnecessary operation. I tested this out as well and it seems to check out.

I had put in a TODO for this observation and am addressing the same in this PR

## Type of change

- [X] Infra/Build change
- [ ] Code refactoring

# Checklist:

- [X] I have made corresponding changes to the documentation

### DIFF
--- a/transformer_engine/jax/attention.py
+++ b/transformer_engine/jax/attention.py
@@ -531,7 +531,7 @@ def _segment_ids_pos_to_seqlens_offsets(
     # This fast path avoids expanding the mask to Q * KV matrix and instead allows us to
     # examine only O(Q+KV) elements.
 
-    # For seqlens and seqoffsets calculations, the intermediate(temp) attn_mask creation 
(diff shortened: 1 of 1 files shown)


KshitijLakhani:
CI pipeline passes: 38909687

PR #2204, opened 2025-09-25
Support fused amax kernels with NVFP4 quantization

timmoon10:
See https://github.com/NVIDIA/TransformerEngine/pull/2177.


timmoon10:
Flattened into 38055db081a86a96959ad870a857abff4e588bfb in https://github.com/NVIDIA/TransformerEngine/pull/2177.

PR #2211, opened 2025-09-29
[JAX] Address tolerance check for current scaling dact dbias

jberchtold-nvidia:
# Description

Address L2 JAX unit test failures for dact with current scaling due to recent changes to removed our intermediate cast to fp32 because calling the primitive.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Do lower precision comparison `dbias` comparison for particular activation type for current scaling dact, similar to existing `dbias` tolerance check for MXFP8

# Checklist:

### DIFF
--- a/tests/jax/test_custom_call_compute.py
+++ b/tests/jax/test_custom_call_compute.py
@@ -780,9 +780,15 @@ class TestFusedQuantize:
             assert_allclose(te_output.data, jax_output.data)
 
         if is_dbias:
-            # TE kernels cast the intermediate results to the input dtype which reduces precision compared to the JAX implementation, for dbias this typically only affects bfloat16.
             precise_comparison = not (
-                in_dtype == jnp.bfloat16 and scaling_mode.is_1d_block_scaling()
+                # TE kernels cast the intermediate results to the input dtype which reduces precision compared to the JAX implementation, for dbias this typically only affects bfloat16.
+                (in_dtype == jnp.bfloat16 and scaling_mode.is_1d_block_scaling())
+                # Due to the amax dependency, current scaling is unfused. In TE we store the activation results in bf16 which reduces precision compared to JAX implementation which will implicitly promote to float32 for the intermediate results when JIT'd. This only produces a tolerance issue when using squared_relu currently.
+                or (
+                    activation_type == ("squared_relu",)
+                    and in_dtype == jnp.bfloat16
+                    and scaling_mode == ScalingMode.CURRENT_TENSOR_SCALING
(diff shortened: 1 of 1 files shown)


tests/jax/test_custom_call_compute.py
```diff
@@ -780,9 +780,15 @@ def _test_quantize_dact_dbias(
             assert_allclose(te_output.data, jax_output.data)
 
         if is_dbias:
-            # TE kernels cast the intermediate results to the input dtype which reduces precision compared to the JAX implementation, for dbias this typically only affects bfloat16.
             precise_comparison = not (
-                in_dtype == jnp.bfloat16 and scaling_mode.is_1d_block_scaling()
+                # TE kernels cast the intermediate results to the input dtype which reduces precision compared to the JAX implementation, for dbias this typically only affects bfloat16.
+                (in_dtype == jnp.bfloat16 and scaling_mode.is_1d_block_scaling())
+                # Due to the amax dependency, current scaling is unfused. In TE we store the activation results in bf16 which reduces precision compared to JAX implementation which will implicitly promote to float32 for the intermediate results when JIT'd. This only produces a tolerance issue when using squared_relu currently.
+                or (
```
phu0ngng, line 787: Hi, 

A minor note that this is a WAR for 2.8. We should probably rework the JAX reference implementation in the future to match this BF16 intermediate type.

phu0ngng (approved):
LGTM. Thanks

tests/jax/test_custom_call_compute.py
jberchtold-nvidia: Good point, I was looking at the forward act instead of dact for fp32 casts. Looks like we do have fp32 casts in the JAX dact impl here: https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/jax/cpp_extensions/activation.py#L965

I'll try casting back to the input dtype for current scaling and see if I can avoid this tolerance change

jberchtold-nvidia: I tried a couple things and wasn't able to get a matching result to TE.

If I compute dact fully in the input dtype for JAX, I get mismatching quantized results. My guess is this is because the intermediate computations of the activation itself are in bf16.

If I compute dact in fp32 then add an intermediate cast to the input_dtype before quantization, I get the original dbias issue. My guess is this is because XLA is optimizing out the intermediate cast.

For v2.8, I think this WAR works but longer-term we can figure out a solution to match the precision here.

PR #2213, opened 2025-09-29
[JAX] Async issuing D2H memcpy for grouped_gemm group_sizes array

huanghua1994:
# Description

This is a draft PR for saving some work and discussion.

Recently we used TE/JAX's `grouped_gemm()` interface for a MoE model's inference. Nsys shows a GPU bubble when `grouped_gemm()` is copying the `group_sizes` array from device to host. This is a known issue when we were designing the `grouped_gemm()` interface. It's performance impact for training / inference prefill stage is relatively small but cannot be ignored in inference decode stage. This draft aims to partially address the bubble issue.

Our target model uses MLP-MoE, i.e., each expert is a MLP layer. After fusing GEMMs, each MLP-MoE layer needs two `grouped_gemm()` with the same `group_sizes` array. This PR allows issuing an async D2H copy of the `group_size` array before entering `grouped_gemm()`, then `grouped_gemm()` can reuse the downloaded `group_sizes`. We have validated the correctness of the implementation in this PR in our target model.

This PR does not solve the issue of breaking CUDA graph in `grouped_gemm()` since in the async copy mode it still needs to call `cudaEventSynchronize()`. Furthermore, the D2H memcpy does not overlap with other operations for copying and dispatching tokens to experts in our implementation for the target model, since those JAX-native operations are captured and executed in CUDA graph, while the async D2H copy does not support CUDA graph.

@phu0ngng @mingxu1067 Please let me know your comments and suggestions. Much appreciated!

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Added `GroupedGemmCopySizesPrimitive` for async copying of `group_sizes` from GPU to host
- Added optional argument `use_async_d2h_group_sizes` for `grouped_gemm()`, the default value is `False` so the original code path will be used by default

# Checklist:

### DIFF
--- a/docs/api/common.rst
+++ b/docs/api/common.rst
@@ -12,6 +12,8 @@ Common API
 
 .. autoapiclass:: transformer_engine.common.recipe.MXFP8BlockScaling(fp8_format=Format.E4M3)
 
+.. autoapiclass:: transformer_engine.common.recipe.NVFP4BlockScaling(fp4_format=Format.E2M1)
+
 .. autoapiclass:: transformer_engine.common.recipe.Float8CurrentScaling(fp8_format=Format.HYBRID)
 
 .. autoapiclass:: transformer_engine.common.recipe.Float8BlockScaling(fp8_format=Format.E4M3)
--- a/tests/cpp_distributed/test_comm_gemm.cu
+++ b/tests/cpp_distributed/test_comm_gemm.cu
@@ -69,6 +69,34 @@ bool IsMulticastSupported(int device_id) {
   return supported;
 }
 
+int GetDeviceComputeCapability(int device_id) {
+  int major{};
+  int minor{};
+  CHECK_CU(cuDeviceGetAttribute(&major, CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MAJOR, device_id));
+  CHECK_CU(cuDeviceGetAttribute(&minor, CU_DEVICE_ATTRIBUTE_COMPUTE_CAPABILITY_MINOR, device_id));
+  return major * 10 + minor;
+}
+
+template <typename T>
+bool IsDTypeSupported(int /* device_id */) {
+  return true;
+}
+
+template <>
+bool IsDTypeSupported<test::fp8e5m2>(int device_id) {
+  return GetDeviceComputeCapability(device_id) >= 89;
+}
+
+template <>
+bool IsDTypeSupported<test::fp8e4m3>(int device_id) {
+  return GetDeviceComputeCapability(device_id) >= 89;
+}
+
+template <typename... Ts>
+bool AllDTypesSupported(int device_id) {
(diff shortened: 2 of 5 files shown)


transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -280,12 +280,71 @@ XLA_FFI_DEFINE_HANDLER_SYMBOL(GemmHandler, GemmFFI,
...
+
+  NVTE_CHECK(dev_group_sizes == nullptr || host_group_sizes == nullptr,
+             "Only one of dev_group_sizes and host_group_sizes can be non-nullptr.");
+
+  if (dev_group_sizes != nullptr) {
+    NVTE_CHECK(num_gemms <= max_num_gemms, "num_gemms ", num_gemms, " exceeds the maximum ",
+               "supported number ", max_num_gemms, " to be downloaded in advance.");
+    host_num_gemms = num_gemms;
+    // Wait for current compute stream to finish
+    cudaStream_t compute_stream_0 = nvte_get_compute_stream(0);
```
phu0ngng, line 309: @mingxu1067 could you check if this causes the same stream sync issue as last time when we used the `compute_stream(0)` instead of the stream given by XLA?

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -280,12 +280,71 @@ XLA_FFI_DEFINE_HANDLER_SYMBOL(GemmHandler, GemmFFI,
...
+  static std::once_flag init_flag;
+  static cudaEvent_t d2h_event;
+  static size_t host_num_gemms;
+  static const size_t max_num_gemms = 1024;
+  //static int32_t host_group_sizes_internal[max_num_gemms];
+  static int32_t *host_group_sizes_internal = nullptr;
+  auto init = [&]() {
+    NVTE_CHECK_CUDA(cudaEventCreate(&d2h_event));
+    NVTE_CHECK_CUDA(cudaMallocHost(&host_group_sizes_internal, sizeof(int32_t) * max_num_gemms));
+  };
```
phu0ngng, line 298: If this causes any issues, we could consider moving this allocation into the FFI prepare phase.

phu0ngng (approved):
I think it is a good improvement for now. 

We should probably provide a `GroupedLayerNormMLP` VJP op, which encloses the `grouped_gemm_copy_group_sizes` function and the `use_async_d2h_group_sizes` option so that we don't expose these two to users as they can be pretty bug-prone.

transformer_engine/jax/csrc/extensions/gemm.cpp
huanghua1994: Just a note: this part follows the logic in https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/common/gemm/cublaslt_gemm.cu#L915

phu0ngng (approved):
LGTM. Thanks

PR #2215, opened 2025-09-30
[Draft][PyTorch][MOE] Support NVFP4 Grouped Linear

zhongbozhu:
# Description

NVFP4 Group Linear Support.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## 

- [ ] Pipe clean, fix NVFP4 padding
- [ ] Numerical test pass
- [ ] Fused Bulk Alloc 
- [ ] Fused multi-swizzle kernel

# Checklist:


transformer_engine/common/swizzle/swizzle.cu
```diff
@@ -661,8 +661,11 @@ void multi_tensor_swizzle_scaling_factors(const std::vector<Tensor*>& input,
       int num_tiles_k = k / SF_TILE_DIM_K;
       int vec_load_size_i = (num_tiles_k - 1) % 4 + 1;
       // We use the minimum vec_load_size across all tensors.
-      // vec_load_size = std::min(vec_load_size, vec_load_size_i);
-      vec_load_size = 1;
+      // TODO(zhongbo): fix vec_load_size for NVFP4
+      // Current unit test won't capture this issue, but in E2E
+      // using vec_load_size = 1 other than 1 will lead to mis-aligned
+      // address error in MOE training
+      vec_load_size = all_nvfp4 ? 1 : std::min(vec_load_size, vec_load_size_i);
```
zhongbozhu, line 668: @yaox12 do you have any idea why? Note that NVFP4 is TN only, ie. there will be transpose happening, unlike MXFP8. Plus that the error only happens for WGRAD. Leaving me to believe maybe it's because padding to 32 for NVFP4 in `m_splits` is not enough if we want `vec_load_size` more than 1. 

So then I increased the padding from 32 to 64, 128. I found that only 128 works. However, I haven't really figured out why the `vec_load_size` calculation logic is wrong, so I am overriding it to 1 as hack.

transformer_engine/common/common.h
```diff
@@ -154,7 +154,8 @@ struct Tensor {
     return acc;
   }
 
-  bool has_data() const noexcept { return data.dptr != nullptr; }
+  // Check for size (not just pointer) for 0-dim or no token cases.
+  bool has_data() const noexcept { return data.dptr != nullptr || data.shape.size() != 0; }
```
timmoon10: Mathematically, a 0-D tensor is a scalar with 1 entry.
```suggestion
  bool has_data() const noexcept { return data.dptr != nullptr; }
```

transformer_engine/pytorch/csrc/extensions/recipe.cpp
```diff
@@ -20,10 +20,11 @@ void compute_amax(const at::Tensor& tensor, at::Tensor& amax) {
 
   TORCH_CHECK(amax.scalar_type() == at::kFloat, "amax must be a float tensor");
   TORCH_CHECK(amax.numel() == 1, "amax must have exactly one element");
+  auto* amax_ptr = amax.data_ptr<float>();
   TensorWrapper fake_te_output(
-      nullptr, te_input.shape(),
+      amax_ptr, te_input.shape(),
       DType::kFloat8E4M3,  // It doesn't matter because we only compute amax.
-      amax.data_ptr<float>());
+      amax_ptr, nullptr, amax_ptr);
```
timmoon10: This horrifying hack is needed because the tensor checking functions assume that the output tensor requires data: https://github.com/NVIDIA/TransformerEngine/blob/dd9433e7ad28c12f27da9770be54c9c584e85fa0/transformer_engine/common/transformer_engine.cpp#L270

The right answer is to modify the API for `nvte_compute_amax` so that the output tensor is an FP32 tensor with one entry. We might use that amax value later to compute an FP8 tensor, an NVFP4 tensor, whatever, but that is completely irrelevant.

transformer_engine/common/common.h
zhongbozhu: We are also checking for data shape for `has_columnwise_data` too.

Removing this won't affect the group linear functionality, but will cause issues with our existing test case comparing grouped gemm with sequential linear layers. 

We can mitigate this issue by largely refactoring the `.shape()` function by not speculating the shape by rowwise & columnwise data shape, but returning a property directly.

timmoon10: If we check dims, the correct impl is:
```c++
bool has_data() const noexcept { return data.dptr != nullptr || product(data.shape) != 0; }
```

A 0-D tensor has 1 entry for the same reason that `2^0=1`. This is the convention used in NumPy and PyTorch and [Wikipedia](https://en.wikipedia.org/wiki/Tensor#Examples):
```
>>> x = np.zeros([])
>>> x.shape
()
>>> x.size
1
>>> y = torch.zeros([])
>>> y.shape
torch.Size([])
>>> y.numel()
1
```

transformer_engine/pytorch/module/fp8_padding.py
yaox12, line 81: Update the docstring.

transformer_engine/pytorch/module/fp8_unpadding.py
yaox12, line 79: Update the docstring.

transformer_engine/pytorch/csrc/util.cpp
```diff
@@ -130,39 +134,44 @@ std::optional<at::Tensor> multi_tensor_swizzle_scaling_factors(
...
+    // auto input_shape = nvte_shape_to_vector(tensor.shape());
+    NVTEShape nvte_input_shape;
+    if (rowwise) {
+      nvte_input_shape = tensor.shape();
+    } else {
+      nvte_input_shape = tensor.get_columnwise_data().shape;
+    }
+    auto input_shape = nvte_shape_to_vector(nvte_input_shape);
     // Reconstruct input only to avoid swizzling both directions if not needed.
     // Use any 8 bit type, it's irrelevant.
```
yaox12, line 154: Perhaps remove this comment.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -491,6 +491,207 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>> bulk_allocate_mx
   return retval;
 }
 
+// allocate fp4 data, fp8 scalings, and amax values
+// layout: [fp4_data0, ..., fp4_dataN, fp8_scaling0, ..., fp8_scalingN, amax0, ..., amaxN]
+// amax values need to be zeroed out, use cudaMemsetAsync for the last few bytes for amax
```
yaox12: Is this comment still accurate? I don't see the `cudaMemsetAsync`.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -491,6 +491,207 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>> bulk_allocate_mx
...
+  const auto rowwise_usage = quantizer_cpp_list[0]->rowwise_usage;
+  const auto columnwise_usage = quantizer_cpp_list[0]->columnwise_usage;
+  const auto scaling_mode = quantizer_cpp_list[0]->get_scaling_mode();
+  const auto fp4_dtype = quantizer_cpp_list[0]->dtype;
+  constexpr size_t scale_elem_size = 1;
+
+  // Helper function to construct tensor view
+  // Note: Deleter holds a shared_ptr for the buffer, so the buffer
+  // will survive until all views are deleted.
+  auto make_torch_view = [](std::shared_ptr<at::Tensor> &buffer, const std::vector<size_t> &shape,
```
yaox12, line 521: May be better to move this `make_torch_view` function to utils, since it's a common helper function for all  `bulk_allocate_*` methods.

transformer_engine/pytorch/csrc/extensions/cast.cpp
zhongbozhu: Yes, this comment needs to be removed, there is already zero amax kernels to clear the amax values in 1. hadamard amax kernel 2. compute amax kernel 
https://github.com/NVIDIA/TransformerEngine/blob/fd2f589f26dfedeb4b3705f0137ec89fe8e5d8cf/transformer_engine/common/hadamard_transform/hadamard_transform.cu#L338
https://github.com/NVIDIA/TransformerEngine/blob/fd2f589f26dfedeb4b3705f0137ec89fe8e5d8cf/transformer_engine/common/recipe/current_scaling.cu#L75

zhongbozhu:
There are some merge conflicts caused by recent TE changes in the mainline that refactored FP8/FP4 into quantization. However, it's unclear how MegatronCore will be impacted so I will pause any changes to this PR as of now.

transformer_engine/common/common.h
zhongbozhu: this would fail the test unfortunately

timmoon10: I tried changing the default behavior so uninitialized tensors have `shape=[0]`, but that quickly became non-trivial. I think we would want to make that change in a standalone, non-functional PR. For now I've reverted this change and moved the logic into the `shape` function. That way the bad design is restricted to one place and we don't expose it even further.

timmoon10 (approved):

PR #2218, opened 2025-09-30
[JAX] Rework amax reduction over TPSP

phu0ngng:
# Description

In this PR, we add an additional check to detect whether the input is partitioned in the sequence dimension and only trigger the amax reduction across TPSP in that case.
With this change, the `using_global_amax_of_x` is no longer needed.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:

### DIFF
--- a/tests/pytorch/attention/run_attention_with_cp.py
+++ b/tests/pytorch/attention/run_attention_with_cp.py
@@ -12,13 +12,17 @@ from transformer_engine.pytorch.attention import DotProductAttention
 from transformer_engine.pytorch.attention.dot_product_attention.context_parallel import (
     get_cu_seqlens_on_cp_rank,
 )
+from transformer_engine.pytorch.attention.dot_product_attention.utils import combine_and_quantize
 import transformer_engine_torch as tex
 from test_attention_with_cp import model_configs_flash_attn, model_configs_fused_attn
 from transformer_engine.pytorch.fp8 import fp8_autocast
-from transformer_engine.pytorch.tensor.float8_tensor import Float8Tensor, Float8Quantizer
-from transformer_engine.common.recipe import DelayedScaling
+from transformer_engine.pytorch.tensor.float8_tensor import (
+    Float8Tensor,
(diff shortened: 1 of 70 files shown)


transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -613,25 +620,26 @@ def partition(
...
+                batch_sequence_transpose=batch_sequence_transpose,
             )
-            if amax_scope is AmaxScope.TPSP:  # Run AR across TP/SP
-                gmesh = global_mesh_resource()
-                amax = lax_paral_op(amax, jax.lax.pmax, gmesh.tp_resource, mesh)
+            gmesh = global_mesh_resource()
+            sequence_dim = 0 if batch_sequence_transpose else 1
+            # Run AR across TPSP only when tensor-sequence is detected in the input spec
+            if amax_scope is AmaxScope.TPSP and x_spec[sequence_dim] == gmesh.tpsp_resource:
                 amax = lax_paral_op(amax, jax.lax.pmax, gmesh.tpsp_resource, mesh)
```
jberchtold-nvidia, line 640: Previously, we would also AR in the `gmesh.tp_resource` dim. To keep the previous behavior, do we still need to reduce on `TP` if `amax_scope is TPSP` but the sequence dim is `TP` not `TPSP`? Or is it okay to only reduce when `tpsp` is active?

jberchtold-nvidia (approved):
Left one question, pending that LGTM. Thanks!

transformer_engine/jax/cpp_extensions/quantization.py
phu0ngng: We only do AG the sequence dimension before the GEMM in TPSP, thus amax reduction is needed.

If users use TP for TPSP, there are existing warnings in the GEMM op to warn users to switch to TPSP.

phu0ngng:
pipeline 35814193 passed.

PR #2222, opened 2025-09-30
[PyTorch] Set usages for linear op quantizers before forward

timmoon10:
# Description

This fixes a bug where a quantizer's usages are set in the backward pass, but are not properly reset in the forward pass. More specifically, the input quantizer might be set with column-wise usage in the backward pass, which causes it to not produce row-wise data in the next forward pass. The solution is to set usages within a `pre_fuser_forward` function that is called before each forward pass. This PR also fixes some bugs where quantizer state is not properly set when the linear op is fused.

Note that the ops used to have a `pre_forward` function, but it was refactored into `pre_first_fuser_forward` and `reset_recipe_type ` in https://github.com/NVIDIA/TransformerEngine/pull/1917 and https://github.com/NVIDIA/TransformerEngine/pull/1951. This was done to avoid unnecessary overheads in every forward pass, mostly related to initializing params and quantization. This PR should not reintroduce significant overhead because setting quantizer usages is cheap. 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add `pre_fuser_forward` function that fusible ops call before each forward pass.
- Fix bugs where linear op might not set quantizer usages or configure FP8 current scaling.
- Add distributed test with MLP block.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/csrc/extensions/bias.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/bias.cpp
@@ -54,10 +54,25 @@ std::vector<py::object> bgrad_quantize(const at::Tensor &grad_output, py::handle
     return {py::cast(std::move(grad_bias_torch)), std::move(grad_input_py)};
   }
 
-  // Unfused impl if quantizer is not supported
-  const bool with_fused_dbias_quantize_kernel =
-      detail::IsFloat8Quantizers(quantizer.ptr()) || detail::IsMXFP8Quantizers(quantizer.ptr());
-  if (!with_fused_dbias_quantize_kernel) {
+  // Check if fused kernel is supported
+  bool with_fused_kernel = false;
+  if (detail::IsFloat8Quantizers(quantizer.ptr())) {
+    auto prop = at::cuda::getCurrentDeviceProperties();
+    const size_t sm_arch = 10 * prop->major + prop->minor;
+    if (sm_arch >= 100) {
+      // Fused kernel for dbias + FP8 cast on SM arch 10.0+
+      with_fused_kernel = true;
+    } else if (quantizer_cpp->rowwise_usage && quantizer_cpp->columnwise_usage) {
+      // Fused kernel for dbias + FP8 cast + FP8 transpose
+      with_fused_kernel = true;
+    }
+  } else if (detail::IsMXFP8Quantizers(quantizer.ptr())) {
+    // Fused kernel for dbias + MXFP8 quantize
+    with_fused_kernel = true;
+  }
+
+  // Apply unfused impl if fused kernel is not supported
+  if (!with_fused_kernel) {
     at::sum_out(grad_bias_torch, grad_output_torch.reshape({-1, bias_size}), {0});
     quantizer_cpp->quantize(grad_output_nvte, grad_input_nvte);
     return {py::cast(std::move(grad_bias_torch)), std::move(grad_input_py)};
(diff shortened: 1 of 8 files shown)


transformer_engine/pytorch/ops/basic/basic_linear.py
```diff
@@ -352,6 +366,35 @@ def reset_recipe_state(self, *, recipe: Optional[Recipe]) -> None:
...
+        # Note: This function may be called in base class constructor,
+        # before any basic linear attrs have been set.
+        if recipe is not None:
+            if recipe.float8_current_scaling():
+                input_quantizer.force_pow_2_scales = recipe.fp8_quant_fwd_inp.power_2_scale
+                input_quantizer.amax_epsilon_scales = recipe.fp8_quant_fwd_inp.amax_epsilon
+                weight_quantizer.force_pow_2_scales = recipe.fp8_quant_fwd_inp.power_2_scale
+                weight_quantizer.amax_epsilon_scales = recipe.fp8_quant_fwd_inp.amax_epsilon
+                grad_output_quantizer.force_pow_2_scales = recipe.fp8_quant_fwd_inp.power_2_scale
+                grad_output_quantizer.amax_epsilon_scales = recipe.fp8_quant_fwd_inp.amax_epsilon
```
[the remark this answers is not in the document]
ptrendx: Actually @timmoon10 should all of those use the same `fp8_quant_fwd_inp`? Recipe has pieces for grad and weight so probably those should be used instead?

transformer_engine/pytorch/csrc/extensions/bias.cpp
```diff
@@ -54,10 +54,25 @@ std::vector<py::object> bgrad_quantize(const at::Tensor &grad_output, py::handle
...
+  bool with_fused_kernel = false;
+  if (detail::IsFloat8Quantizers(quantizer.ptr())) {
+    auto prop = at::cuda::getCurrentDeviceProperties();
+    const size_t sm_arch = 10 * prop->major + prop->minor;
+    if (sm_arch >= 100) {
+      // Fused kernel for dbias + FP8 cast on SM arch 10.0+
+      with_fused_kernel = true;
+    } else if (quantizer_cpp->rowwise_usage && quantizer_cpp->columnwise_usage) {
+      // Fused kernel for dbias + FP8 cast + FP8 transpose
+      with_fused_kernel = true;
```
ksivaman, line 67: Won't this case also trigger the failure in `fp8_quantize_arch_l_100` that checks for unfused bias?

ksivaman: Oh I guess for the 2x direction quantize it won't go through that path because it'll call `cast_transpose` instead, so I think the logic is okay but not tight. Ideally this should be in one place instead of here + core but that's not easy here.

ksivaman (approved):
LGTM

transformer_engine/pytorch/csrc/extensions/bias.cpp
ptrendx: Yeah, that should be cleaned in common, but not today :-).

transformer_engine/pytorch/ops/basic/basic_linear.py
```diff
@@ -322,6 +322,20 @@ def pre_first_fuser_forward(self) -> None:
         if self.weight.device.type == "meta":
             self.reset_parameters()
 
+    def pre_fuser_forward(self, *, requires_grad: bool) -> None:
+        super().pre_fuser_forward(requires_grad=requires_grad)
+        if FP8GlobalStateManager.is_fp8_enabled():
+            # Configure quantizer usages
+            # Note: We cache the quantized input for backward pass,
+            # but discard the quantized weights.
```
timmoon10, line 330: This comment was made incorrect in https://github.com/NVIDIA/TransformerEngine/pull/1817. Since then we've cached both the quantized inputs and weights.

PR #2223, opened 2025-10-01
Fix the cuBLAS workspace alignment

ptrendx:
# Description

In some cases the pointer alignment we got from pyTorch is not 256 B. This PR is a (potentially temporary) solution while we figure out why that happens. It overallocates the  workspace buffer and then makes sure that the pointer given to cuBLAS is aligned to 256 B.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

### DIFF
--- a/transformer_engine/common/gemm/cublaslt_gemm.cu
+++ b/transformer_engine/common/gemm/cublaslt_gemm.cu
@@ -703,7 +703,8 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
   NVTE_CHECK_CUBLAS(cublasLtMatmulPreferenceSetAttribute(
       preference, CUBLASLT_MATMUL_PREF_MIN_ALIGNMENT_D_BYTES, &D_alignment, sizeof(D_alignment)));
   NVTE_CHECK(new_workspace_alignment % 256 == 0,
-             "cuBLAS workspace pointer must be aligned to 256 bytes, got ", new_workspace_alignment);
+             "cuBLAS workspace pointer must be aligned to 256 bytes, got ",
+             new_workspace_alignment);
 
   const auto status =
       cublasLtMatmulAlgoGetHeuristic(handle, operationDesc, Adesc, Bdesc, Cdesc, Ddesc, preference,


transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -679,14 +679,21 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
 #endif
   }
 
+  // align the workspace to 256 B
+  const int required_alignment = 256;
+  const auto original_workspace_alignment = _getAlignment(reinterpret_cast<uintptr_t>(workspace));
+  uint8_t *aligned_workspace_ptr =
+      reinterpret_cast<uint8_t *>(workspace) + required_alignment - original_workspace_alignment;
```
phu0ngng, line 686: We did something similar in JAX.
https://github.com/phu0ngng/TransformerEngine/blob/main/transformer_engine/jax/csrc/extensions/gemm.cpp#L30C17-L34

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -679,14 +679,21 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
...
+  workspaceSize = workspaceSize - required_alignment + original_workspace_alignment;
+  const auto new_workspace_alignment =
+      _getAlignment(reinterpret_cast<uintptr_t>(aligned_workspace_ptr));
```
timmoon10, line 689: Nits:
- This implementation relies on a hidden assumption that the max alignment from `_getAlignment` is 256 B, and a change could result in underflows.
- We don't do anything with the original workspace pointer, so no harm in modifying it in-place? No reason that the code below needs to know exactly what we did with the pointer, especially if we do more manipulations and carveouts in the future.
```suggestion
  // align the workspace to 256 B
  auto workspace_alignment = _getAlignment(reinterpret_cast<uintptr_t>(workspace));
  const size_t required_workspace_alignment = 256;
  if (workspace_alignment % required_workspace_alignment != 0) {
    workspace = reinterpret_cast<uint8_t *>(workspace) + required_workspace_alignment - workspace_alignment;
    workspaceSize -= required_workspace_alignment - workspace_alignment
    workspace_alignment = _getAlignment(reinterpret_cast<uintptr_t>(workspace));
  }
```

timmoon10 (approved):

PR #2230, opened 2025-10-02
[JAX] Fix for GEMM + fuse bias + AllReduce

phu0ngng:
# Description

Currently, when we need to do AllReduce on the GEMM output, TE custom call fuses the bias addition on each GPU, then we sum up the partial outputs. As a result, the bias is added N-GPU times. 
This PR introduces a logic for this case to not fuse the bias so that we can add the bias only once after the AllReduce.
A simple unit test file for distributed gemm and distributed dense vjp is added.

UPDATE: This PR also fixed the bug we have in the NormBwdPrimitive that we missed AllReduce the dgamma along the TPSP axes.
UPDATE2: This PR also fixed the byte count bug in the `test_distributed_layernorm.py`.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


tests/jax/distributed_test_base.py
```diff
@@ -16,27 +16,24 @@
...
-        configs.append(
-            pytest.param(2, (2,), ("dp",), MeshResource(dp_resource="dp"), id="n2_dp2_tp1")
-        )
-        configs.append(
-            pytest.param(2, (2,), ("tpsp",), MeshResource(tpsp_resource="tpsp"), id="n2_dp1_tp2")
-        )
-
     if is_devices_enough(4):
-        configs.append(
+        return [
```
phu0ngng: I think we don't necessarily need to test the 2-GPU config when 4-GPUs are available.

tests/jax/test_distributed_layernorm_mlp.py
```diff
@@ -129,7 +148,7 @@ def layernorm_fp8_mlp_prim_func(
         quantizer_sets = QuantizerFactory.create_set(n_quantizer_sets=2)
 
         # out = ((x * kernel_1) + bias_1) * kernel_2 + bias_2
-        return jnp.mean(
+        return jnp.sum(
```
phu0ngng: Switch to sum so that the errors could be amplified more if any.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1039,13 +1038,17 @@ def _sharded_impl(lhs, lhs_scale_inv, rhs, rhs_scale_inv, bias, gelu_input):
...
+                if not collective_op.is_reduce_scatter:
+                    if is_all_reduce_in_float32():  # For unittest only
+                        outputs[0] = jax.lax.psum(outputs[0].astype(jnp.float32), reduce_spec).astype(
+                            out_dtype
+                        )
+                    else:
+                        outputs[0] = jax.lax.psum(outputs[0], reduce_spec)
+
+                if fuse_bias:  # TODO(Phuong): rename fuse_bias to has_bias
+                    outputs[0] += bias
```
jberchtold-nvidia, line 1049: Not for this PR, but long-term would it be possible to fuse the bias on one device per group of reducing devices?

If we have a Mesh such that for a particular gemm we had, (non_contracting_axes: 4 GPUs, contracting_axes: 2 GPUs), then in our CPP impl we'd only fuse the bias with cuBLAS GEMM for only one device in each group of contracting_axes block of GPUs. This might be pretty hard to do correctly in JAX, I'm not sure if we can assume any jax.devices() to cudaDeviceId mapping or assignment in the Mesh

tests/jax/distributed_test_base.py
jberchtold-nvidia: Can you move these 2 GPU tests to L2 instead of removing them? Check for the `NVTE_JAX_UNITTEST_LEVEL == "L2"`.

 I think we should keep these 2 GPU tests in some form because it will help triage failures faster by isolating tests to just DP or just TPSP, but I agree it doesn't have to be L0 or L1 since it is semi-redundant with the 4 GPU case

transformer_engine/jax/cpp_extensions/gemm.py
phu0ngng: I don't think we can do this in this way, as there is no way to tell which device is the first device in the TP domain if we don't do CGEMM. 
The other approach, which is easy to implement, is to fuse bias with bias = bias / TP_size().

jberchtold-nvidia (approved):

tests/jax/test_distributed_layernorm_mlp.py
phu0ngng: I changed it back to mean that now we have bigger shapes the sum makes the gradient range too large for BF16 tolerances.

tests/jax/test_distributed_layernorm_mlp.py
```diff
@@ -404,9 +430,10 @@ def _test_layernorm_mlp(
         # within tolerance to the float32 ground truth.
         jax_triton_gemm_precision_tolerance_update = (
             with_jax_gemm
-            and isinstance(fp8_recipe, recipe.Float8CurrentScaling)
-            and dtype == jnp.bfloat16
-            and activation_type == ("gelu", "linear")
+            and fp8_recipe is not None
+            and (fp8_recipe.delayed() or fp8_recipe.float8_current_scaling())
+            and dtype in (jnp.bfloat16, jnp.float16)
```
phu0ngng, line 435: I need to add `float16` and `DelayedScaling` into this jax_gemm tol update as we now have: 1. Bigger test case, 2. AllReduce in dbias, 3. AllReduce the dgamma in the TPSP axis, which introduces more diffs. 
This should not be an issue, as we have verified that the diffs still stay within FP32 tolerances.

tests/jax/test_distributed_layernorm.py
```diff
@@ -66,18 +66,19 @@ def generate_collectives_count_ref(
...
-        # TODO(Jeremy): debug this check because layernorm should always have 2x weights regardless of dp
-        weight_count = 2 if (ln_type == "layernorm" and "dp" in mesh_axes) else 1
-        allreduce_total_bytes = (
-            all_reduce_loss_bytes + weight_count * shape[-1] * jax_dtype.itemsize
-        )
-        other_bytes = 0
+        # loss, 1 FP32
+        allreduce_total_bytes = 4 if is_dp_enabled else 0
+        # dgamma and dbeta
+        weight_count = 2 if ln_type == "layernorm" else 1
```
jberchtold-nvidia, line 76: LGTM, thanks for updating this!

I have in my backlog to finish this PR that will run the JAX norm in addition to the TE norm, will continue this PR after I finish the current items: https://github.com/NVIDIA/TransformerEngine/pull/2041/files

phu0ngng:
pipeline #36123624 passed.

PR #2238, opened 2025-10-06
[JAX] Activation/Normalization to output amax for later quantization in CurrentScaling

phu0ngng:
# Description

Activation/Normalization to output amax for later quantization in CurrentScaling

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/activation.py
+++ b/transformer_engine/jax/cpp_extensions/activation.py
@@ -191,6 +191,7 @@ class ActLuPrimitive(BasePrimitive):
         dtype = dtypes.canonicalize_dtype(x_aval.dtype)
         assert dtype in [jnp.float32, jnp.float16, jnp.bfloat16]
         assert scale_aval is None or scale_aval.dtype == jnp.float32
+        assert amax_aval is None or amax_aval.dtype == jnp.float32
         assert x_aval.shape[-2] == act_len, (
             "activation input should be replicated by act_len in the -2 axis, got input shape"
             f" {x_aval.shape} and act_len {act_len}"
@@ -242,7 +243,7 @@ class ActLuPrimitive(BasePrimitive):
         """
         te_gated_act_lu_p lowering rules
         """
-        del out_dtype, scale_dtype, act_len, is_outer
+        del out_dtype, scale_dtype, act_len, is_outer, amax_scope, transpose_batch_sequence
         x_aval, scale_aval, amax_aval = ctx.avals_in
         assert x_aval.dtype in [jnp.float32, jnp.float16, jnp.bfloat16]
         assert scale_aval is None or scale_aval.dtype == jnp.float32
@@ -647,7 +648,7 @@ class BaseDActLuDBiasQuantizePrimitive(BasePrimitive):
         """
         te_dact_dbias_quantize_p abstract
         """
-        del act_enum, act_params
+        del act_enum, act_params, amax_scope, transpose_batch_sequence, output_amax_when_no_scaling
         dz_dtype = dtypes.canonicalize_dtype(dz_aval.dtype)
         assert dz_dtype in [jnp.float32, jnp.float16, jnp.bfloat16]
         assert x_aval.dtype == dz_dtype
@@ -656,6 +657,7 @@ class BaseDActLuDBiasQuantizePrimitive(BasePrimitive):
             f" {x_aval.shape} and act_len {act_len}"
         )
         assert scale_aval.dtype == jnp.float32
+        assert amax_aval.dtype == jnp.float32
 
         assert scaling_mode != ScalingMode.CURRENT_TENSOR_SCALING.value, (
             "Current tensor scaling is not supported for fused dact and quantization. Please do"
@@ -1254,6 +1256,8 @@ def act_lu(
         If quantizer is provided:
             A ScaledTensor containing the quantized activated input.
     """
+    # TODO(Phuong): remove the output_amax_when_no_scaling exposure by introducing _act_lu_impl()
+    # Do the same with dact_dbias_quantize() and layernorm_fwd()
     act_type_id = ActivationEnum[activation_type].value
     act_len = len(activation_type)
     assert x.shape[-2] == act_len, (
@@ -1418,7 +1422,6 @@ def quantize_dact_dbias(
             scale,
             amax,
             # outputs float32 for dbias accumulation
-            # TODO: remove this upcast
             out_dtype=(jnp.float32 if is_dbias else x.dtype),
             # default value for no scaling, TE/common ignore this value when scale is unset
             scaling_mode=ScalingMode.NO_SCALING.value,
--- a/transformer_engine/jax/cpp_extensions/activation.py
+++ b/transformer_engine/jax/cpp_extensions/activation.py
@@ -148,7 +148,6 @@ class ActLuPrimitive(BasePrimitive):
     name = "te_act_lu_ffi"
     multiple_results = True
     impl_static_args = (
-        2,
         3,
         4,
         5,
@@ -156,7 +155,11 @@ class ActLuPrimitive(BasePrimitive):
         7,
         8,
         9,
-    )  # out_dtype, act_enum, act_len, scaling_mode, is_2x, scale_dtype, is_outer, act_params
+        10,
+        11,
+        12,
+        13,
+    )  # out_dtype, act_enum, act_len, scaling_mode, is_2x, scale_dtype, act_params, amax_scope, transpose_batch_sequence, output_amax_when_no_scaling, is_outer
     inner_primitive = None
     outer_primitive = None
 
@@ -164,6 +167,7 @@ class ActLuPrimitive(BasePrimitive):
     def abstract(
         x_aval,
         scale_aval,
+        amax_aval,
         *,
         out_dtype,
         act_enum,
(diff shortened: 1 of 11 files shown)


transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -148,36 +148,43 @@ class ActLuPrimitive(BasePrimitive):
...
+        amax_scope,
+        transpose_batch_sequence,
+        output_amax_when_no_scaling,
+        is_outer,
     ):
         """
         te_act_lu_p abstract
         """
-        del act_enum, act_params
+        del act_enum, act_params, amax_scope, transpose_batch_sequence, output_amax_when_no_scaling
```
jberchtold-nvidia: Can we add an assert here that we are using `ScalingMode.NO_SCALING` if `output_amax_when_no_scaling` is true?

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -452,26 +495,40 @@ def partition(
...
+                is_outer=True,
             )
 
             if scaling_mode == ScalingMode.DELAYED_TENSOR_SCALING.value:
-                global_updated_amax = all_reduce_max_along_all_axes_except_PP(local_amax, mesh)
+                global_updated_amax = all_reduce_max_along_all_axes_except_PP(
+                    local_updated_amax, mesh
+                )
+            elif scaling_mode == ScalingMode.NO_SCALING.value and output_amax_when_no_scaling:
+                global_updated_amax = amax_scope.all_reduce_amax_along_TPSP_and_FSDP(
```
jberchtold-nvidia, line 531: Thanks for refactoring this so the logic for amax reduction isn't duplicated between primitives!

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -1232,19 +1400,22 @@ def quantize_dact_dbias(
         f" {x.shape} and act_len {act_len}"
     )
 
-    scale = jnp.empty((), jnp.float32)
+    scale = jnp.empty((1,), jnp.float32)
+    amax = jnp.zeros((1,), jnp.float32)  # need to init with zero and shape=(1,)
```
jberchtold-nvidia, line 1411: Why does this amax need to be an input to the primitive? We only every output an amax from fused act, right?  Is this just to allow JAX to zero out the buffer instead of nvte_memset?

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -110,15 +111,20 @@ def abstract(
...
         scale_dtype,
+        amax_scope,
+        transpose_batch_sequence,
+        output_amax_when_no_scaling,
         is_outer,
     ):
         """
         LayerNorm fwd inner primitive abstract
         """
+        del amax_scope, transpose_batch_sequence, output_amax_when_no_scaling
```
jberchtold-nvidia: Can we assert here in the abstract that `output_amax_when_no_scaling` is only allowed to be True when the scaling mode is NO_SCALING and we are not using cuDNN norm? Same for norm bwd abstract

transformer_engine/jax/csrc/extensions/normalization.cpp
```diff
@@ -123,8 +132,6 @@ Error_Type NormForwardFFI(cudaStream_t stream, Buffer_Type x_buf, Buffer_Type sc
 
   if (scaling_mode == JAXX_Scaling_Mode::DELAYED_TENSOR_SCALING && is_fp8_dtype(out_dtype)) {
     output_tensor.set_scale(scale, DType::kFloat32, std::vector<size_t>{1});
-    nvte_memset(amax, 0, sizeof(float), stream);
```
jberchtold-nvidia, line 126: Do we know if `jnp.zeros` is faster or is this just a simplification for refactoring purposes? Previously, we saw a cudaMemset being slower than a fill kernel, but I am unsure which `jnp.zeros` is doing or if JAX/XLA can combine multiple zeros into a single op.

transformer_engine/jax/flax/module.py
jberchtold-nvidia, line 1081: Can we add a TODO to utilize `output_amax_if_no_scaling` here for the unfused norm codepath in the modules? I agree, that it is not as high priority as our fused VJP path as that is the most optimal path we want to optimize

jberchtold-nvidia: Update: I think `output_amax_if_no_scaling` should not be in the public API and should instead be internal to `cpp_extensions`. In the future, when we want to support this unfused cases like here, we can add it. But for now we don't, so we don't need to do anything here

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -1091,7 +1178,9 @@ def rmsnorm_fwd(
     zero_centered_gamma: bool,
     epsilon: float,
     quantizer: Optional[Quantizer],
-    amax_scope: AmaxScope = AmaxScope.LOCAL,
+    amax_scope: AmaxScope = AmaxScope.TPSP,
+    transpose_batch_sequence: bool = False,
+    output_amax_when_no_scaling: bool = False,
```
jberchtold-nvidia, line 1190: I don't think `output_amax_when_no_scaling` should be exposed publicly outside of `cpp_extensions` for now as afaict we don't use it anywhere in this PR. The logic to use this can be determined internally in `cpp_extensions` as is already done for the current scaling case. If I'm missing something, lmk, thanks!

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -1209,6 +1374,9 @@ def quantize_dact_dbias(
     is_dbias: bool = True,
     quantizer: Optional[Quantizer] = None,
     act_params: Optional[ActivationParams] = None,
+    amax_scope: AmaxScope = AmaxScope.LOCAL,
+    transpose_batch_sequence: bool = False,
+    output_amax_when_no_scaling: bool = False,
```
jberchtold-nvidia, line 1386: I don't think `output_amax_when_no_scaling` should be exposed publicly outside of `cpp_extensions` for now as afaict we don't use it anywhere in this PR. The logic to use this can be determined internally in `cpp_extensions` as is already done for the current scaling case. If I'm missing something, lmk, thanks!

transformer_engine/jax/cpp_extensions/normalization.py
```diff
@@ -882,6 +937,8 @@ def layernorm_fwd(
     epsilon: float,
     quantizer: Optional[Quantizer],
     amax_scope: AmaxScope = AmaxScope.LOCAL,
+    transpose_batch_sequence: bool = False,
+    output_amax_when_no_scaling: bool = False,
```
jberchtold-nvidia, line 948: I don't think `output_amax_when_no_scaling` should be exposed publicly outside of `cpp_extensions` for now as afaict we don't use it anywhere in this PR. The logic to use this can be determined internally in `cpp_extensions` as is already done for the current scaling case. If I'm missing something, lmk, thanks!

transformer_engine/jax/cpp_extensions/activation.py
phu0ngng: Yes, I added amax to activation and normalization primitives for two reasons:
1. Eliminate the nvte_memset
2. Stay consistent with quantization primitives.

transformer_engine/jax/cpp_extensions/normalization.py
phu0ngng: Done for NormFWD. We don't quantize the NormBWD output thus this change is irrelevant.

transformer_engine/jax/csrc/extensions/normalization.cpp
phu0ngng: XLA should be able to combine zero initialization.

transformer_engine/jax/flax/module.py
phu0ngng: It's probably tricky if we want to support this unfuse case even in the future because if you output the amax in the fwd, you will need to deal with its gradient in the bwd.
In fact, we have this unfuse code path here because the VJPs did not support BF16 in the past. Now it is, so we should be able to refactor the code to use VJP.

transformer_engine/jax/cpp_extensions/normalization.py
phu0ngng: I expose `output_amax_when_no_scaling` in `normalization.py` and `activation.py` because we recursively call the function inside the function itself, for example, here https://github.com/phu0ngng/TransformerEngine/blob/main/transformer_engine/jax/cpp_extensions/normalization.py#L950-L957. 

In general, I don't anticipate users using cpp_extensions functions as they are not differential. I would be more cautious if we need to change VJP APIs.

phu0ngng: Another possible way is to split the implementation into an internal `_impl` function, as we did in the `_quantize_dbias_impl`. Let me add a todo for that.

jberchtold-nvidia: I see, yes I think an internal `_impl` method is a good idea for this but just adding a TODO in this PR sounds good to me.

phu0ngng:
Pipeline 36141587 passed.

PR #2245, opened 2025-10-07
FSDP2 Weight Update Fix

vthumbe1503:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/distributed/run_fsdp2_model.py
+++ b/tests/pytorch/distributed/run_fsdp2_model.py
@@ -9,7 +9,7 @@ import sys
 import argparse
 
 import transformer_engine.pytorch as te
-from transformer_engine.common.recipe import Format, DelayedScaling
+from transformer_engine.common.recipe import Format, DelayedScaling, Float8CurrentScaling
 
 import torch
 import torch.distributed as dist
@@ -104,24 +104,19 @@ def _train(args):
     # FP8 Configuration
     fp8_format = Format.HYBRID
     fp8_recipe = DelayedScaling(fp8_format=fp8_format, amax_history_len=16, amax_compute_algo="max")
-
+    build_model_context_args = {}
     if not args.fp8_init:
         # Build model context (FP8 init)
         build_model_context = nullcontext
-        build_model_context_args = {}
-
+    else:
         from transformer_engine.pytorch import fp8_model_init
-
         build_model_context = fp8_model_init
         build_model_context_args["enabled"] = True
-
-        # Build the model with the specified context
-        with build_model_context(**build_model_context_args):
-            model = SimpleNet(args.input_size, args.hidden_size, args.output_size)
-    else:
-        model = SimpleNet(args.input_size, args.hidden_size, args.output_size)
+        build_model_context_args["recipe"] = fp8_recipe
     # Move the model to the correct device
-
+    # Build the model with the specified context
+    with build_model_context(**build_model_context_args):
+        model = SimpleNet(args.input_size, args.hidden_size, args.output_size)
     model.to(device)
 
     if LOCAL_RANK == 0:
@@ -163,13 +158,14 @@ def _train(args):
         # Zero the parameter gradients
         optimizer.zero_grad()
         input_data = torch.randn(args.batch_size, args.input_size).to(device)
-        output = model(input_data)
+        with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
+            output = model(input_data)
         target = torch.randn(args.batch_size, args.output_size).to(device)
         loss = F.mse_loss(output, target)
         loss.backward()
         optimizer.step()
         if LOCAL_RANK == 0:
-            print(f"Rank {LOCAL_RANK}: Iteration {iteration} completed.")
+            print(f"Rank {LOCAL_RANK}: Iteration {iteration} completed with loss {loss.item()}")
 
     dist.destroy_process_group()
     if LOCAL_RANK == 0:
--- a/transformer_engine/pytorch/tensor/float8_tensor.py
+++ b/transformer_engine/pytorch/tensor/float8_tensor.py
@@ -4,6 +4,7 @@
 
 """Tensor class with FP8 data"""
 from __future__ import annotations
+import os
 from typing import Optional, Tuple, Iterable, Union
 import warnings
 
@@ -537,8 +538,34 @@ class Float8Tensor(Float8TensorStorage, QuantizedTensor):
         self._transpose = None
 
     @classmethod
-    def __torch_dispatch__(cls, func, types, args, kwargs=None):
+    def make_like(
+        cls,
+        tensor: QuantizedTensor,
+        *,
+        shape: Optional[Iterable[int]] = None,
+        dtype: Optional[torch.dtype] = None,
+        requires_grad: bool = False,
+        data: Optional[torch.Tensor] = None,
+    ) -> Float8Tensor:
+        """Create new Float8 tensor
+        By default, new tensor has the same attributes and underlying
+        data.
 
+        """
+        if shape is None:
+            shape = data.shape if data is not None else tensor.shape
+        dtype = dtype if dtype is not None else tensor.dtype
+        kwargs = tensor.get_metadata()
+        if data is not None:
+            kwargs["data"] = data
+            # TODO: Move common code to Base class and just update metadata here.
+            fp8_scale_inv = tensor._scale_inv.detach().clone()
+            kwargs["fp8_scale_inv"] = fp8_scale_inv
+
+        return cls(shape=shape, dtype=dtype, requires_grad=requires_grad, **kwargs)
+
+    @classmethod
+    def __torch_dispatch__(cls, func, types, args, kwargs=None):
         # View op
         if func == aten.view.default:
             tensor = args[0]
@@ -594,16 +621,19 @@ class Float8Tensor(Float8TensorStorage, QuantizedTensor):
                 Float8Tensor.make_like(tensor, data=split_tensor, shape=split_tensor.shape)
                 for split_tensor in func_out
             ]
-        if func == aten.new_zeros.default:
+
+        if func == aten.new_zeros.default or func == aten.zeros_like.default:
             tensor = args[0]
             data = tensor._data
+            extra_args = list(args[1:]) if len(args) > 1 else []
             func_out = data.__torch_dispatch__(
                 func,
                 types,
-                [data] + list(args[1:]),
+                [data] + extra_args,
                 kwargs,
             )
             return Float8Tensor.make_like(tensor, data=func_out, shape=func_out.shape)
+
         if func == torch.ops.aten.as_strided.default:
             tensor = args[0]
             data = tensor._data
@@ -614,6 +644,7 @@ class Float8Tensor(Float8TensorStorage, QuantizedTensor):
                 kwargs,
             )
             return Float8Tensor.make_like(tensor, data=func_out, shape=func_out.shape)
+
         if func == torch.ops.aten.detach.default:
             return cls.detach(args[0])
         if func == torch.ops.aten.clone.default:
--- a/transformer_engine/pytorch/tensor/quantized_tensor.py
+++ b/transformer_engine/pytorch/tensor/quantized_tensor.py
@@ -5,6 +5,7 @@
 """Tensor with quantized data"""
 
 from __future__ import annotations
+import os
 from typing import Callable, Optional, Tuple, Iterable, Any, Dict, Union
 import abc
 import copy
@@ -495,6 +496,10 @@ class QuantizedTensor(torch.Tensor):
                 and schema_arg.alias_info.is_write
             ):
                 arg.quantize_(new_arg)
+            elif isinstance(arg, list) and isinstance(new_arg, list):
+                # Recursively handle for lists of tensors
+                for a, na in zip(arg, new_arg):
+                    maybe_update_inplace(a, na, schema_arg)
 
         # In-place op: dequantize, perform op, and quantize
         if func._schema.is_mutable:


transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -635,9 +733,92 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+            metadata (Any): metadata sent out in fsdp_pre_all_gather used for reconstructing the Float8Tensor.
+            param_dtype (torch.dtype):
+            out (Optional[torch.Tensor], optional): _description_. Defaults to None.
+
+        Returns:
+            Tuple[Float8Tensor, Tuple[torch.Tensor, ...]]: Allgathered Float8Tensor and tuple of internal tensors
+            used by the Float8Tensor that was actually allgathered
+        """
+        # Import here to avoid circular imports
+        from ..fp8 import FP8GlobalStateManager
```
timmoon10: `transformer_engine.fp8` was deprecated in https://github.com/NVIDIA/TransformerEngine/pull/2256: 
```suggestion
        from ..quantization import FP8GlobalStateManager
```

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -635,9 +733,92 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+                  "dtype": fake_dtype,
+                  "fp8_scale_inv": fp8_scale_inv,
+                  "fp8_dtype": fp8_dtype,
+                  "quantizer": quantizer,
+                  "requires_grad": False,
+                  "data": None,
+                  "data_transpose": None,
+                  }
+        fp8_args.update({"data" if is_in_fp8_autocast else "data_transpose": data})
+        return Float8Tensor(**fp8_args), (data, )
```
timmoon10: This manual transpose logic will break on Blackwell because FP8 transposes are only for Hopper. Also, it's unnecessary because we coerce the weight tensor to have the required data:
https://github.com/NVIDIA/TransformerEngine/blob/2712bb95cb4a4f7d1f2b8b473a2240ac3d6e7e58/transformer_engine/pytorch/module/linear.py#L263
https://github.com/NVIDIA/TransformerEngine/blob/2712bb95cb4a4f7d1f2b8b473a2240ac3d6e7e58/transformer_engine/pytorch/module/linear.py#L685

It would be simpler just to populate `_data` and to let the module lazily compute `_transpose` if needed.
```suggestion
        if out is not None:
            out._data = data
            out._transpose_invalid = True
            out.update_usage(
                rowwise_usage=out.quantizer.rowwise_usage,
                columnwise_usage=out.quantizer.columnwise_usage,
            )
            return
        fp8_args = {
            "shape": orig_shape,
            "dtype": fake_dtype,
            "fp8_scale_inv": fp8_scale_inv,
            "fp8_dtype": fp8_dtype,
            "quantizer": quantizer,
            "requires_grad": False,
            "data": data,
            "data_transpose": None,
        }
        return Float8Tensor(**fp8_args), (data, )
```

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -635,9 +733,92 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+        Returns:
+            Tuple[Float8Tensor, Tuple[torch.Tensor, ...]]: Allgathered Float8Tensor and tuple of internal tensors
+            used by the Float8Tensor that was actually allgathered
+        """
+        # Import here to avoid circular imports
+        from ..fp8 import FP8GlobalStateManager
+        # Detect if we're within fp8_autocast scope. We'll be in 
+        # forward pass when within the fp8_autocast scope. Backward
+        # pass when outside of the fp8_autocast scope.
+        is_in_fp8_autocast = FP8GlobalStateManager.is_fp8_enabled()
```
timmoon10: Assuming that the forward pass is within an autocast and that the backward pass is outside is quite tenuous. I don't see a logical reason why this has to be true. For an example, `te.ops.Linear` supports the case with FP8 params and non-FP8 compute.

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -635,9 +733,92 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+        quantizer = self._quantizer.copy()
+        if isinstance(quantizer, Float8CurrentScalingQuantizer) and mesh is not None:
+            # When sharded weight is updated after reduce scattering the gradients in FSDP2,
+            # we need to do amax reduction across the mesh to make sure all weight shards are
+            # updated with same scale inverse. Setting the state below in the quantizer will make
+            # sure that updated Quantized weight tensor have same scale inverse across all shards.
+            quantizer.amax_reduction_group = mesh.get_group()
+            quantizer.with_amax_reduction = True
+        is_transpose_needed = not self._transpose_invalid and self._transpose is not None
+        quantizer.set_usage(rowwise=not is_transpose_needed, columnwise=is_transpose_needed)
```
timmoon10: I don't think this is correct. I would expect the sharded weights to be unchanged between the forward and backward pass, but the transpose requirements are different in the forward and backward pass.

If we remove transpose logic from `fsdp_post_all_gather`, we can simply remove this logic:
```suggestion
```

transformer_engine/pytorch/module/base.py
```diff
@@ -1371,6 +1371,7 @@ def get_weight_workspace(
         # Note: Make sure weights have required usages, but do not
         # destroy unnecessary usages since they may be used later.
         if isinstance(tensor, QuantizedTensor):
+            quantizer = tensor._quantizer
```
timmoon10: This doesn't really make sense since there's no reason to force the tensor to match its internal quantizer. This logic is there to force the output tensor to match the user-provided quantizer. If we sometimes follow the quantizer and sometimes ignore it, then that makes it hard to reason about this function and requires knowledge of its internal implementation.

If the tensor is doing something wrong from following the quantizer, the right fix is to change the quantizer.

transformer_engine/pytorch/module/linear.py
```diff
@@ -240,7 +240,8 @@ def forward(
         weightmat = weight
         if fp8 or debug:
             # Configure quantizer
-            if weight_quantizer is not None:
+            # No need to set the quantizer states if weight is already quantized
+            if weight_quantizer is not None and not isinstance(weight, QuantizedTensor):
```
timmoon10, line 244: We want `get_weight_workspace` to actually respect the quantizer that's passed into it. Otherwise, you need to know its internal details to reason out the cases where it follows the quantizer and the cases where it ignores it. We should add logic so that `columnwise_usage=False` if FSDP is enabled.

tests/pytorch/distributed/run_fsdp2_model.py
```diff
@@ -103,25 +110,23 @@ def _train(args):
 
     # FP8 Configuration
     fp8_format = Format.HYBRID
-    fp8_recipe = DelayedScaling(fp8_format=fp8_format, amax_history_len=16, amax_compute_algo="max")
+    # fp8_recipe = DelayedScaling(fp8_format=fp8_format, amax_history_len=16, amax_compute_algo="max")
+    # fp8_recipe = Float8CurrentScaling(fp8_format=fp8_format)
+    fp8_recipe = MXFP8BlockScaling(fp8_format=fp8_format)
```
timmoon10: The quantization recipe should be configurable so that we can test all of them.

tests/pytorch/distributed/run_fsdp2_model.py
```diff
@@ -157,19 +166,34 @@ def _train(args):
...
+        print("Memory after materialization:", torch.cuda.memory_allocated(device)/1e6, "MB")
+
     optimizer = optim.Adam(model.parameters(), lr=1e-3)
 
     for iteration in range(args.iter):
         # Zero the parameter gradients
         optimizer.zero_grad()
         input_data = torch.randn(args.batch_size, args.input_size).to(device)
-        output = model(input_data)
+        with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
```
timmoon10: We've deprecated `fp8_autocast` in https://github.com/NVIDIA/TransformerEngine/pull/2256 since we also support FP4:
```suggestion
        with te.autocast(enabled=True, recipe=fp8_recipe):
```

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -1,3 +1,4 @@
+
```
timmoon10:
```suggestion
```

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -317,33 +318,249 @@ def contiguous(
...
+            scale_invs = [tensor._rowwise_scale_inv, tensor._columnwise_scale_inv]
+            split_sizes_for_scale = [split_size, split_size // MXFP8_BLOCK_SCALING_SIZE]
+            for (scale_inv, split_size) in zip(scale_invs, split_sizes_for_scale):
+                scale_inv_out = scale_inv.__torch_dispatch__(
+                    func,
+                    types,
+                    [scale_inv, split_size] + list(args[2:]),
+                    kwargs,
+                )
+                out_data.append(scale_inv_out)
```
timmoon10: We should take the scale-inv padding into account. Alternatively, we could fall back to the dequantize-requantize impl if the dims are not multiples of 512.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -317,33 +318,249 @@ def contiguous(
...
+
+        Returns:
+            shareded_tensors: Tuple[torch.Tensor, ...]: Tuple of tensors
+            that need to be all-gathered.
+            metadata: Tuple[Any]: Metadata needed for reconstructing the
+            MXFP8Tensor after all-gather.
+        """
+        # Import here to avoid circular imports
+        from ..fp8 import FP8GlobalStateManager
+        is_forward_pass = FP8GlobalStateManager.is_fp8_enabled()
```
timmoon10: This assumption is brittle and sometimes incorrect (e.g. `te.ops.Linear` with MXFP8 params and BF16 compute). Is there any other way we can detect if we are in the backward pass?

transformer_engine/pytorch/tensor/float8_tensor.py
vthumbe1503: Fixed this for Blackwell now.

transformer_engine/pytorch/module/linear.py
vthumbe1503: I think weight_quantizer doesn't make sense in the case of an already quantized tensor. Since we are never going to use that at all in the course of training.

My understanding is the code in this if statement is needed for the case when weights are in high precision and we need to quantize it appropriately for the forward and backward passes.

In case of Quantized weight tensor, whenever the weights go through a dequantization/requantization phase for a op, it is the weight._quantizer that is going to be used. Is my understanding right? If so, in alignment with your previous comment. I can set weight_quantizer = weight._quantizer in the case weight is the quantized tensor, so even the get_weight_workspace gets the right quantizer.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
vthumbe1503: Yes, found another way, there is another function signature for pre_fsdp_all_gather that takes in module as argument which contains the current training state. Updated the code now

transformer_engine/pytorch/module/base.py
vthumbe1503: Agreed. Changed the quantizer that needs to be passed to the interface now.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
vthumbe1503: Shouldn't the dims multiple be 128 instead of 512? I am already doing that at the moment and in that case it does default back to dequantize-requantize phase.

transformer_engine/pytorch/tensor/float8_tensor.py
vthumbe1503: Yes you are right sharded weights should have the quantizer config not changed. But allgathered weights does have different requirements for forward and backward pass. Fixed that logic now.

transformer_engine/pytorch/tensor/float8_tensor.py
vthumbe1503: Fixed it now. Using FSDP module's training state to decide to as to whether we are in forward pass.

tests/pytorch/distributed/run_fsdp2_model.py
vthumbe1503: Fixed

transformer_engine/pytorch/tensor/float8_tensor.py
vthumbe1503: not using fp8 module for deciding forward pass anymore.

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -501,8 +505,6 @@ def make_like(
             shape = data.shape if data is not None else tensor.shape
         dtype = dtype if dtype is not None else tensor.dtype
         kwargs = tensor.get_metadata()
-        if data is not None:
-            kwargs["data"] = data
```
timmoon10, line 505: This change is logical. As far as I can tell, `data` is specific to FP8 (other data formats have different tensors for row-wise and column-wise data, as well as scales), so it doesn't belong in an abstract quantized tensor class. We should go further and remove the `data` kwarg from `make_like`.

transformer_engine/pytorch/module/linear.py
timmoon10: Hm, I see your argument. If the weights are quantized then the damage is already done. No benefit trying to fix usages here, and we might as well update usages lazily. I think your approach is reasonable.

I'll just nit the comment below ("In the case of FSDP2...") since this is not specific to FSDP, but for any quantized weights. The comment above ("No need to set the quantizer states...") is more accurate.

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -632,9 +731,107 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+            # Transpose is not needed at all for Blackwell+
+            tensor_has_transpose = not self._transpose_invalid and self._transpose is not None
+            # When module is wrapped with torch no_grad, the training state
+            # will be IDLE even in forward pass
+            is_forward_pass = fsdp_state._training_state in (
+                TrainingState.FORWARD,
+                TrainingState.IDLE,
+            )
+            transpose_needed = tensor_has_transpose and not is_forward_pass
+            quantizer.set_usage(rowwise=not transpose_needed, columnwise=transpose_needed)
```
timmoon10: No need to manually keep track of transposes. We can specify usages and let `Float8Tensor.update_usages` handle that logic internally:
```suggestion

        # Weights need row-wise usage for forward pass, column-wise usage for backward pass
        fsdp_state = module._get_fsdp_state()
        need_rowwise = fsdp_state._training_state in (
            TrainingState.FORWARD,
            TrainingState.IDLE,  # Forward within no_grad context
        )
        if fsdp_state._fsdp_param_group._reshard_after_forward:
            need_columnwise = fsdp_state._training_state == TrainingState.PRE_BACKWARD
        else:
            # Without resharding, weights in forward pass will be reused in backward pass
            need_columnwise = fsdp_state._training_state in (
                TrainingState.FORWARD,
                TrainingState.PRE_BACKWARD,
            )

        # Quantizer to be used for all-gathered weights
        quantizer = self._quantizer.copy()
        quantizer.set_usage(rowwise=need_rowwise, columnwise=need_columnwise)

```

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -314,33 +314,327 @@ def contiguous(
...
+                TrainingState.IDLE,
+            )
+            # Allgather only the necessary tensors based on forward/backward pass
+            quantizer.set_usage(rowwise=is_forward_pass, columnwise=not is_forward_pass)
+            sharded_tensors = (
+                sharded_tensors
+                if is_forward_pass
+                else (self._columnwise_data, self._columnwise_scale_inv)
+            )
+        metadata = (self._fp8_dtype, quantizer, reshard_after_forward)
```
timmoon10: There's a bug if FSDP does not reshard after the forward pass. Making this consistent with my suggestion for `Float8Tensor.fsdp_pre_all_gather`:
```suggestion

        # Weights need row-wise usage for forward pass, column-wise usage for backward pass
        fsdp_state = module._get_fsdp_state()
        need_rowwise = fsdp_state._training_state in (
            TrainingState.FORWARD,
            TrainingState.IDLE,  # Forward within no_grad context
        )
        reshard_after_forward = fsdp_state._fsdp_param_group._reshard_after_forward
        if reshard_after_forward:
            need_columnwise = fsdp_state._training_state == TrainingState.PRE_BACKWARD
        else:
            # Without resharding, weights in forward pass will be reused in backward pass
            need_columnwise = fsdp_state._training_state in (
                TrainingState.FORWARD,
                TrainingState.PRE_BACKWARD,
            )

        # Quantizer to be used for all-gathered weights
        quantizer = self._quantizer.copy()
        quantizer.set_usage(rowwise=need_rowwise, columnwise=need_columnwise)
        
        # Return data to be all-gathered
        sharded_tensors = []
        if quantizer.rowwise_usage:
            sharded_tensors.extend((self._rowwise_data, self._rowwise_scale_inv))
        if quantizer.columnwise_usage:
            sharded_tensors.extend((self._columnwise_data, self._columnwise_scale_inv))
        sharded_tensors = tuple(sharded_tensors)
        metadata = (self._fp8_dtype, quantizer)
```
I'm not very happy with the messy logic for `sharded_tensors`/`all_gather_outputs`. If `fsdp_pre_all_gather`/`fsdp_post_all_gather` can handle `None`s, then we could make things slightly nicer:
```python
# fsdp_pre_all_gather
sharded_tensors = (
    self._rowwise_data if quantizer.rowwise_usage else None,
    self._rowwise_scale_inv if quantizer.rowwise_usage else None,
    self._columnwise_data if quantizer.columnwise_usage else None,
    self._columnwise_scale_inv if quantizer.columnwise_usage else None,
)

# fsdp_post_all_gather
(
    rowwise_data,
    rowwise_scale_inv,
    columnwise_data,
    columnwise_scale_inv,
) = all_gather_outputs
```

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -314,33 +314,327 @@ def contiguous(
...
+            columnwise_scale_inv = mxfp8_tensor._columnwise_scale_inv
+
+        if out is not None:
+            out._rowwise_data = rowwise_data
+            out._rowwise_scale_inv = rowwise_scale_inv
+            out._columnwise_data = columnwise_data
+            out._columnwise_scale_inv = columnwise_scale_inv
+            out._quantizer = quantizer
+
+        return mxfp8_tensor, all_gather_outputs
```
timmoon10: Matching changes in `fsdp_pre_all_gather`, avoiding unnecessary tensor creation when `out` is provided, and removing unnecessary logic for `reshard_after_forward`:
```suggestion
        # Extract all-gathered data from args
        fp8_dtype, quantizer = metadata
        if quantizer.rowwise_usage:
            rowwise_data, rowwise_scale_inv = all_gather_outputs[:2]
        else:
            rowwise_data, rowwise_scale_inv = None, None
        if quantizer.columnwise_usage:
            columnwise_data, columnwise_scale_inv = all_gather_outputs[-2:]
        else:
            columnwise_data, columnwise_scale_inv = None, None

        # Initialize all-gathered tensor
        if out is None:
            out = MXFP8Tensor(
                rowwise_data=rowwise_data,
                rowwise_scale_inv=rowwise_scale_inv,
                columnwise_data=columnwise_data,
                columnwise_scale_inv=columnwise_scale_inv,
                fp8_dtype=fp8_dtype,
                dtype=param_dtype,
                shape=rowwise_data.shape if rowwise_data is not None else columnwise_data.shape,
                quantizer=quantizer,
            )
        else:
            out._rowwise_data = rowwise_data
            out._rowwise_scale_inv = rowwise_scale_inv
            out._columnwise_data = columnwise_data
            out._columnwise_scale_inv = columnwise_scale_inv
            out._quantizer = quantizer
            
        return out, all_gather_outputs
```

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -478,10 +772,12 @@ def forward(
...
                     break
         if shape[-1] != ctx.shape[-1]:
-            raise RuntimeError(
-                "MXFP8Tensor does not support reshaping inner dimension "
+            warnings.warn(
+                "MXFP8Tensor does not support reshaping inner dimension."
                 f"(attempted to reshape dims={tuple(tensor.shape)} to {tuple(shape)})"
+                "If you are using this for FSDP2, then ignore this warning."
             )
+            return tensor.detach()
```
timmoon10: This is very suspicious. What is FSDP2 trying to do here, flatten the tensor to 1D? If it calls `view(-1)` and we don't return a flattened tensor, how do we know it won't cause problems?

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -684,75 +684,74 @@ def backward(
...
-            gemm_out = None
-            reduce_scatter_out = None
-            if ctx.ub_overlap_rs_dgrad:
-                reduce_scatter_out = torch.empty(
-                    dgrad_shape, dtype=ctx.activation_dtype, device=grad_outputs[0].device
-                )
-            elif ctx.ub_bulk_wgrad:
-                gemm_out = ub_obj_wgrad.get_buffer(local_chunk=False)
+            dgrad = None
+            if ctx.requires_dgrad:
```
timmoon10: Even if the input doesn't require grad, we still need to do the dgrad GEMM if the LayerNorm params require grads. For now we should keep the existing behavior where we always do the dgrad GEMM.

tests/pytorch/distributed/test_torch_fsdp2.py
```diff
@@ -36,7 +40,8 @@ def _run_test(fp_init, sharding_dims):
 @pytest.mark.skipif(not te.torch_version() >= (2, 4, 0), reason="Requires PyTorch 2.4.0+")
 @pytest.mark.parametrize("sharding_dims", ([NUM_PROCS], [2, NUM_PROCS // 2]))
 @pytest.mark.parametrize("fp8_init", (False, True))
-def test_distributed(fp8_init, sharding_dims):
+@pytest.mark.parametrize("recipe", ("delayed_scaling", "current_scaling", "mx_fp8_block_scaling"))
```
timmoon10, line 43: We should skip recipe test cases if the recipe is not supported. See the logic in `distributed/test_numerics.py`: https://github.com/NVIDIA/TransformerEngine/blob/0e80c847845be58dbb88f46a0975786ddc823798/tests/pytorch/distributed/test_numerics.py#L55-L69

tests/pytorch/distributed/test_torch_fsdp2.py
```diff
@@ -36,7 +40,8 @@ def _run_test(fp_init, sharding_dims):
...
+def test_distributed(fp8_init, sharding_dims, recipe):
```
timmoon10: Beyond the scope of this PR, but we should consider reorganzing the tests so that we reduce the number of times we start parallel jobs. That process is slow (on the order of seconds), so it becomes painful if we have many test cases. See `distributed/test_fusible_ops.py` for an example where we only launch parallel jobs a few times: https://github.com/NVIDIA/TransformerEngine/blob/0e80c847845be58dbb88f46a0975786ddc823798/tests/pytorch/distributed/test_fusible_ops.py#L1043-L1044
Within each job, we run many test cases:
https://github.com/NVIDIA/TransformerEngine/blob/0e80c847845be58dbb88f46a0975786ddc823798/tests/pytorch/distributed/test_fusible_ops.py#L962

transformer_engine/pytorch/tensor/float8_tensor.py
vthumbe1503: This goes back to having the problem for blackwell GPUs right? Where transpose might not be needed at all for backward pass. That is the reason I had that logic for checking whethere transpose was not present. And if it is not present in sharded weights itself, then I was assuming it is Blackwell and not setting the columnwise usage

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -632,9 +731,107 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+                "quantizer": quantizer,
+                "requires_grad": False,
+                "data": data,
+            }
+            out = Float8Tensor(**fp8_args)
+            out.update_usage(
+                rowwise_usage=quantizer.rowwise_usage,
+                columnwise_usage=quantizer.columnwise_usage,
+            )
+        return out, (data,)
```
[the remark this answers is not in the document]
vthumbe1503: doesnt matter since FSDP ignores the return statement if not Needed. So can have both the cases return the same

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -632,9 +731,107 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+        (data,) = all_gather_outputs
+        (fp8_scale_inv, fp8_dtype, quantizer) = metadata
+        orig_shape = data.size()
+        # Quantizer has only columnwise usage set for backward pass
+        # with pre-hopper architectures if weights are resharded after forward pass.
+        if out is not None:
+            out.update_usage(
+                rowwise_usage=quantizer.rowwise_usage,
+                columnwise_usage=quantizer.columnwise_usage,
+            )
```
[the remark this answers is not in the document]
vthumbe1503: Yes always idempotent

transformer_engine/pytorch/quantized_tensor.py
vthumbe1503: Done

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -600,17 +652,63 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+                    types,
+                    [transpose, t_shape] + list(args[2:]),
+                    kwargs,
+                )
+            # deep copy the scale inverse tensor and quantizer as well.
+            scale_inv = tensor._scale_inv.detach().clone()
+            quantizer = tensor._quantizer.copy()
+            out_tensor = Float8Tensor(
+                data=func_out,
+                shape=data.shape,
```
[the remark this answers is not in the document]
vthumbe1503: fixed

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -587,11 +613,37 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+                    func,
+                    types,
+                    [transpose, args[1], t_dim],
+                    kwargs,
+                )
+            outs = [
+                Float8Tensor.make_like(
+                    tensor,
+                    data=split_tensor,
+                    data_transpose=split_tranpose_tensor,
```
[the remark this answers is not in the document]
vthumbe1503: fixed

transformer_engine/pytorch/module/base.py
```diff
@@ -1324,8 +1354,11 @@ def clear(self):
                 param._high_precision_init_val = high_precision_init_val
                 param.get_high_precision_init_val = MethodType(get, param)
                 param.clear_high_precision_init_val = MethodType(clear, param)
+                # Update the parameter based on its type
 
-            setattr(self, name, param)
+            (
+                setattr(self, name, param) if not is_dtensor else setattr(self, name, dtensor_param)
+            )  # pylint: disable=expression-not-assigned
```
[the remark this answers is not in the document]
vthumbe1503: no

transformer_engine/pytorch/tensor/mxfp8_tensor.py
vthumbe1503: allgather cant handle None unfortunately. since fsdp needs the size of the sharded tensor to determine the shape of allgathered dtensor

transformer_engine/pytorch/tensor/mxfp8_tensor.py
vthumbe1503: Nice idea :) Added the code.

transformer_engine/pytorch/module/base.py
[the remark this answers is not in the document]
vthumbe1503: Done

transformer_engine/pytorch/distributed.py
```diff
@@ -1885,6 +1885,37 @@ def allreduce(
     return inp, handle
 
 
+@lru_cache
+def _get_module_fsdp_state(module):
```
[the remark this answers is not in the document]
vthumbe1503: No it wont be stale ever. Since the same state is used again and again throughout the course of training.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -478,10 +767,12 @@ def forward(
...
```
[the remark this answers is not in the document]
vthumbe1503: Added better warning

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -314,33 +314,322 @@ def contiguous(
...
+        out: Optional[torch.Tensor] = None,
+    ):
+        """Functions FSDP2 calls after all-gather of the
+        weights for both forward and backward passes.
+        Args:
+            all_gather_outputs (Tuple[torch.Tensor, ...]): sharded_tensors sent out in fsdp_pre_all_gather from each rank
+            are all-gathered and received here as a tuple.
+            metadata (Any): metadata sent out in fsdp_pre_all_gather used for reconstructing the MXFP8Tensor.
+            param_dtype (torch.dtype):
+            out (Optional[torch.Tensor], optional): _description_. Defaults to None.
```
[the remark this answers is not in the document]
vthumbe1503: DOne

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -632,9 +730,105 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+        if reshard_after_forward:
+            # Transpose is not needed at all for Blackwell+
+            tensor_has_transpose = not self._transpose_invalid and self._transpose is not None
+            training_state = fsdp_state._fsdp_param_group._training_state
+            is_backward_pass = training_state == TrainingState.PRE_BACKWARD
+            transpose_needed = tensor_has_transpose and not is_backward_pass
+            quantizer.set_usage(rowwise=not transpose_needed, columnwise=transpose_needed)
+        sharded_tensors = (self._data,)
+        metadata = (self._scale_inv, self._fp8_dtype, quantizer)
+        return sharded_tensors, metadata
```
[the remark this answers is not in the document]
vthumbe1503: Done

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -478,10 +759,12 @@ def forward(
                     shape[i] = d_inferred
                     break
         if shape[-1] != ctx.shape[-1]:
-            raise RuntimeError(
-                "MXFP8Tensor does not support reshaping inner dimension "
+            warnings.warn(
+                "MXFP8Tensor does not support reshaping inner dimension."
```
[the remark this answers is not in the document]
vthumbe1503: Done

tests/pytorch/distributed/run_fsdp2_model.py
```diff
@@ -74,10 +90,160 @@ def _parse_args(argv=None, namespace=None):
...
+            kwargs["input_layernorm"] = True
+    kwargs["device"] = config.device
+    model = nn.Sequential(*[layer_type(*args, **kwargs) for _ in range(config.num_layers)])
+    return model, inp_shape, out_shape
+
+
+def get_device_mesh(world_size, sharding_dims):
+    dist_print(f"sharding-dims:{sharding_dims}")
+    device_ids = list(range(world_size))
+    if sharding_dims == None:  # FSDP
```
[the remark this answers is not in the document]
vthumbe1503: Done

tests/pytorch/distributed/test_torch_fsdp2.py
vthumbe1503: Noted. Will do this in a follow up PR.

tests/pytorch/distributed/test_torch_fsdp2.py
vthumbe1503: Done

transformer_engine/pytorch/module/layernorm_linear.py
vthumbe1503: Fixed it. Removed those changes. Now always computing the dgrad

transformer_engine/pytorch/tensor/mxfp8_tensor.py
vthumbe1503: Added clearer warning message now and changed view to go down dequantization route instead(for sanity, although it wont be used at the moment) . Turns out, view(-1) is used if atleast one of the two conditions are satisfied
1)  fsdp_pre_all_gather methods is not defined or
2) compiled autograd is enabled.
For us both of them is not true, since we have the method defined for mxfp8_tensor now and also we arent using torch compile for backward pass. However in case torch compile is needed for backward pass in the future, then the behavior right now is we are dequantizing and viewing the tensor.

FSDP2 essentially uses view(-1) to eventually define the sharded parameter and uses that to use allgather instead if any one of the conditions are met. Since fsdp_pre_all_gather is defined for us, in case torch compile is enabled for backward pass, allgather would happen in fp32(so from fidelity standpoint it isnt an issue)

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -314,33 +314,314 @@ def contiguous(
...
+
+        else:
+            out = MXFP8Tensor(
+                rowwise_data=rowwise_data,
+                rowwise_scale_inv=rowwise_scale_inv,
+                columnwise_data=columnwise_data,
+                columnwise_scale_inv=columnwise_scale_inv,
+                fp8_dtype=fp8_dtype,
+                dtype=param_dtype,
+                shape=rowwise_data.shape if rowwise_data is not None else columnwise_data.shape,
```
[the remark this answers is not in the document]
vthumbe1503: Nope thats not gonna happen

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -552,9 +575,11 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
                 if (
                     out_transpose_shape[0] != out_shape[-1]
-                    or out_transpose_shape[1:] != out_shape[:-1]
-                ):
-                    out_transpose = None
+                else:
+                    view_shape_for_transpose = [out_shape[-1]] + list(out_shape[:-1])
+                    if out_transpose is not None:
+                        out_transpose = out_transpose.view(*view_shape_for_transpose)
+                    out_transpose = out_transpose.view(*view_shape_for_transpose)
```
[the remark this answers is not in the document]
vthumbe1503: Fixed now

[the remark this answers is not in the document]
vthumbe1503: removed now

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -314,33 +314,331 @@ def contiguous(
...
```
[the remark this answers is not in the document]
vthumbe1503: Would never happen

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -314,33 +314,331 @@ def contiguous(
...
+                        data.__torch_dispatch__(
+                            func,
+                            types,
+                            [data] + list(args[1:]),
+                            kwargs,
+                        )
+                        if data is not None
+                        else None
+                    )
+                    out_data.append(func_out)
```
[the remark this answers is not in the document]
vthumbe1503: Fixed

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -314,33 +314,330 @@ def contiguous(
...
+        Returns:
+            Tuple[MXFP8Tensor, Tuple[torch.Tensor, ...]]: Allgathered MXFP8Tensor and tuple of internal tensors
+            used by the MXFP8Tensor that was being computed after allgather.
+        """
+        fp8_dtype, quantizer = metadata
+        rowwise_data, rowwise_scale_inv = (
+            all_gather_outputs[:2] if quantizer.rowwise_usage else (None, None)
+        )
+        columnwise_data, columnwise_scale_inv = (
+            all_gather_outputs[-2:] if quantizer.columnwise_usage else (None, None)
```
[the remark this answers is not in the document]
vthumbe1503: If both usage flags are true. Then both rowwise and columnwise data must be present and hence 4 tensors must be present.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -314,33 +314,321 @@ def contiguous(
...
+                fp8_dtype=tensor._fp8_dtype,
             )
-            out_shape = out_data.size()
+        if func == torch.ops.aten.copy_.default:
+            dst, src = args[0], args[1]
+            if isinstance(src, MXFP8Tensor) and isinstance(dst, MXFP8Tensor):
+                dst._rowwise_data.copy_(src._rowwise_data.detach())
+                dst._rowwise_scale_inv.copy_(src._rowwise_scale_inv.detach())
+                dst._columnwise_data.copy_(src._columnwise_data.detach())
+                dst._columnwise_scale_inv.copy_(src._columnwise_scale_inv.detach())
```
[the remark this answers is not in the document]
vthumbe1503: not gonna happen

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -314,33 +314,330 @@ def contiguous(
...
+                        [scale_inv, scale_split_size] + list(args[2:]),
+                        kwargs,
+                    )
+                    if scale_inv is not None
+                    else None
+                )
+                out_data.append(scale_inv_out)
+            return [
+                MXFP8Tensor(
+                    shape=splitted_tensor_data[0].size(),
```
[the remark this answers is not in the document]
vthumbe1503: one of them has to be present. Cant happen

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -478,10 +786,14 @@ def forward(
                     shape[i] = d_inferred
                     break
         if shape[-1] != ctx.shape[-1]:
-            raise RuntimeError(
-                "MXFP8Tensor does not support reshaping inner dimension "
+            warnings.warn(
+                "MXFP8Tensor does not support reshaping inner dimension. "
                 f"(attempted to reshape dims={tuple(tensor.shape)} to {tuple(shape)})"
```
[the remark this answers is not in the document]
vthumbe1503: Done

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -338,9 +337,318 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+            out._quantizer = quantizer
+        else:
+            out = MXFP8Tensor(
+                rowwise_data=rowwise_data,
+                rowwise_scale_inv=rowwise_scale_inv,
+                columnwise_data=columnwise_data,
+                columnwise_scale_inv=columnwise_scale_inv,
+                fp8_dtype=fp8_dtype,
+                dtype=param_dtype,
+                shape=rowwise_data.shape if rowwise_data is not None else columnwise_data.shape,
```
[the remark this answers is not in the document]
vthumbe1503: both cant be None

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -278,10 +278,13 @@ def forward(
         weightmat = weight
         quantized_weight = False
         if fp8 or debug:
-            quantized_weight = not isinstance(weight, QuantizedTensorStorage)
```
timmoon10, line 281: Oof, this variable name was aggressively bad. Just to make things extremely explicit, perhaps we should rename to `is_weight_param_quantized`.

If we change this variable, we need to make sure to update the other places it is used:
https://github.com/NVIDIA/TransformerEngine/blob/dcaca2a67ee0c390cb900a4c29168aef6ac198d5/transformer_engine/pytorch/module/layernorm_linear.py#L432
https://github.com/NVIDIA/TransformerEngine/blob/dcaca2a67ee0c390cb900a4c29168aef6ac198d5/transformer_engine/pytorch/module/layernorm_linear.py#L566

transformer_engine/pytorch/distributed.py
```diff
@@ -1885,6 +1885,37 @@ def allreduce(
...
+    from torch.distributed._composable_state import _module_state_mapping
+
+    # this will return correct fsdp state if module itself is an fsdp module
+    if hasattr(module, "_get_fsdp_state"):
+        fsdp_state = module._get_fsdp_state()
+    # Otherwise get the fsdp state of lca of module in the module hierarchy
+    else:
+        min_nodes_in_parent = float("inf")
+        closest_parent_fsdp_mod = None
+        for fsdp_mod in _module_state_mapping.keys():
```
timmoon10, line 1908: I don't see any good alternatives to this O(n^2) search, but we should switch if we ever find a more natural way to traverse parent modules.

transformer_engine/pytorch/distributed.py
```diff
@@ -1885,6 +1885,37 @@ def allreduce(
     return inp, handle
 
 
+@lru_cache
```
timmoon10: Using the default of `maxsize=128`, then a Transformer model with >32 layers will not fit in the LRU cache. Since this logic is per-module, it's more natural to cache the FSDP state within the module rather than in a global cache. How about something like:
```python
def _get_module_fsdp_state(module):
    if hasattr(module, "_get_fsdp_state"):
        return module._get_fsdp_state()
    if getattr(module, "_parent_fsdp_state", None) is not None:
        return module._parent_fsdp_state

    # Traverse to find parent FSDP

    module._parent_fsdp_state = fsdp_state
    return fsdp_state
```

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -314,7 +314,42 @@ def contiguous(
 
     @classmethod
     def __torch_dispatch__(cls, func, types, args, kwargs=None):
-        # FSDP2 related functions
+        # View op
+        if func == aten.view.default:
```
timmoon10, line 318: This implementation is incorrect because it doesn't handle scales or column-wise data. We should prefer using the existing view impl: https://github.com/NVIDIA/TransformerEngine/blob/5a40f2b3179e27a967d21eef01c1234be6ee2a4c/transformer_engine/pytorch/tensor/mxfp8_tensor.py#L756

I assume that this case is because FSDP does something like `view(-1)`? If so, we should check the view dims and only use this impl for that case:
```python
if func == aten.view.default:
    tensor = args[0]
    dims = args[1]  # ?
    if len(dims) == 1:
        if dims[0] not in (-1, tensor.numel()):
            raise ValueError(...)  # Invalid dims
        warnings.warn(...)  # Warn non-FSDP users
        return MXFP8Tensor(
            rowwise_data=tensor._rowwise_data.view(-1),
            columnwise_data=tensor._columnwise_data.view(-1),
            ...
        )
    return tensor.view(dims)
```

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -314,7 +314,42 @@ def contiguous(
...
+                rowwise_data=out_data,
+                rowwise_scale_inv=tensor._rowwise_scale_inv,
+                columnwise_data=tensor._columnwise_data,
+                columnwise_scale_inv=tensor._columnwise_scale_inv,
+                quantizer=tensor._quantizer,
+                requires_grad=False,
+                fp8_dtype=tensor._fp8_dtype,
+            )
+
+        if func == torch.ops.aten.copy_.default:
```
timmoon10, line 340: The base implementation respects the quantizer usages of the `dst` tensor, so it's incorrect for this impl to wipe everything out based on the quantizer usages of the `src` tensor.

The use case I am thinking about is a model that has been initialized with `quantized_model_init`. The modules have configured the weight quantizers, possibly with very complicated internal logic. Then we load a checkpoint and update the weights with `param.copy_(checkpoint_param)`. Overwriting the quantizer would undo all of the hard work we have already done in the configuration.

transformer_engine/pytorch/distributed.py
vthumbe1503: Yep, reasonable. Will do that!

transformer_engine/pytorch/tensor/mxfp8_tensor.py
vthumbe1503: When FSDP calls view its rather the view method of the MXFP8 tensor class that is called (which does ViewFunc.apply) instead of going into the torch dispatch mechanism. This view implementation was what had been there before this pr itself, I just moved it above. But it makes sense to correct it, How about I use _ViewFunc.apply here itself, rather than repeating the view and FSDp specific logic?

transformer_engine/pytorch/distributed.py
vthumbe1503: Actually if we are guranteed python3.7 or higher with dict order same as insertion order, we can stop at the first fsdp module which is the parent, since the fsdp modules in the dict are supposed to be put bottom up in FSDP2. In that case it is O(n). And it seems like for TE, we have requirement of python3.10+ and so I can remove the logic of  finding the parent with minimum number of submodules, but rather return the first parent enciuntered. Does that sound reasonable?

transformer_engine/pytorch/tensor/mxfp8_tensor.py
vthumbe1503: In that case checkpoint param would be high precision tensor right? I have a check, to copy things only if both src and dst are MXFP8 Tensors and so this code wont be executed and rather fall back to the base implementation.

transformer_engine/pytorch/module/layernorm_linear.py
vthumbe1503: Sure, will do that.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
vthumbe1503: But I see your argument. That philosophy is consistent even with per tensor scaling class. So would remove the quantizer assignment.

transformer_engine/pytorch/distributed.py
timmoon10: The existing impl is already somewhat delicate (e.g. fails in weird edge cases like if a module is wrapped by two independent FSDP modules). Relying on insertion order seems too fragile and non-obvious. An alternative is to skip modules in `_module_state_mapping` if the number of submodules is more than `min_nodes_in_parent`, although I suspect that it won't help much because iterating through submodules is the expensive part.

That said, since we are caching then this is a one-time cost. I'd say it would be better to prioritize robustness over speed.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
timmoon10: I see that this buggy impl has been around since MXFP8 support was added, maybe because `torch.autograd.Function` creates tensor views of inputs? In any case, it's orthogonal to this PR and we can hold off of fixing it until we understand it better.

vthumbe1503: I am not sure about its use-case. So will leave it as it is

transformer_engine/pytorch/tensor/mxfp8_tensor.py
vthumbe1503: Done

transformer_engine/pytorch/distributed.py
vthumbe1503: Done

transformer_engine/pytorch/distributed.py
vthumbe1503: Fair enough

transformer_engine/pytorch/module/layernorm_linear.py
vthumbe1503: Done

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -338,9 +337,320 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+            dst, src = args[0], args[1]
+            if isinstance(src, MXFP8Tensor) and isinstance(dst, MXFP8Tensor):
+                if src._rowwise_data is not None and dst._rowwise_data is not None:
+                    dst._rowwise_data.copy_(src._rowwise_data.detach())
+                    dst._rowwise_scale_inv.copy_(src._rowwise_scale_inv.detach())
+                if src._columnwise_data is not None and dst._columnwise_data is not None:
+                    dst._columnwise_data.copy_(src._columnwise_data.detach())
+                    dst._columnwise_scale_inv.copy_(src._columnwise_scale_inv.detach())
+                return dst
```
timmoon10: Fixing a correctness bug if `src` and `dst` have different usages:
```suggestion
            if (
                isinstance(src, MXFP8Tensor) and isinstance(dst, MXFP8Tensor)
                and (src._rowwise_data is not None or dst._rowwise_data is None)
                and (src._columnwise_data is not None or dst._columnwise_data is None)
            ):
                if dst._rowwise_data is not None:
                    dst._rowwise_data.copy_(src._rowwise_data.detach())
                if dst._columnwise_data is not None:
                    dst._columnwise_data.copy_(src._columnwise_data.detach())
                return dst
```

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -276,12 +276,15 @@ def forward(
         # Prepare weight tensor
         # ------------------------------------------------------
         weightmat = weight
-        quantized_weight = False
+        is_weight_param_quantized = False
         if fp8 or debug:
-            quantized_weight = not isinstance(weight, QuantizedTensorStorage)
+            is_weight_param_quantized = isinstance(weight, QuantizedTensorStorage)
```
timmoon10, line 281: We are changing the definition of this variable with a `not`, but its usages are not changed (see https://github.com/NVIDIA/TransformerEngine/pull/2245#discussion_r2496657079). Is this correct?

vthumbe1503: Yes it was defined wrong and not even used anywhere. And so we werent facing issues. I decided to use that variable. But for that I needed to define it correctly.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -338,9 +337,329 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+            Tuple[MXFP8Tensor, Tuple[torch.Tensor, ...]]: Allgathered MXFP8Tensor and tuple of internal tensors
+            used by the MXFP8Tensor that was being computed after allgather.
+        """
+        fp8_dtype, quantizer = metadata
+        rowwise_data, rowwise_scale_inv = (
+            all_gather_outputs[:2] if quantizer.rowwise_usage else (None, None)
+        )
+        columnwise_data, columnwise_scale_inv = (
+            all_gather_outputs[-2:] if quantizer.columnwise_usage else (None, None)
+        )
```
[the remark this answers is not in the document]
vthumbe1503: no it would hold

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -338,9 +337,330 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+
+        else:
+            out = MXFP8Tensor(
+                rowwise_data=rowwise_data,
+                rowwise_scale_inv=rowwise_scale_inv,
+                columnwise_data=columnwise_data,
+                columnwise_scale_inv=columnwise_scale_inv,
+                fp8_dtype=fp8_dtype,
+                dtype=param_dtype,
+                shape=rowwise_data.shape if rowwise_data is not None else columnwise_data.shape,
```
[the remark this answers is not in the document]
vthumbe1503: both would never be none together

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -338,9 +337,325 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+            out._quantizer = quantizer
+        else:
+            out = MXFP8Tensor(
+                rowwise_data=rowwise_data,
+                rowwise_scale_inv=rowwise_scale_inv,
+                columnwise_data=columnwise_data,
+                columnwise_scale_inv=columnwise_scale_inv,
+                fp8_dtype=fp8_dtype,
+                dtype=param_dtype,
+                shape=rowwise_data.shape if rowwise_data is not None else columnwise_data.shape,
```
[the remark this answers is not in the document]
vthumbe1503: both cant be none together

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -428,7 +428,7 @@ def forward(
                 fsdp_group,
                 mu,
                 rsigma,
-                weightmat if is_weight_param_quantized else None,
+                weightmat if not is_weight_param_quantized else None,
```
timmoon10: Change to match previous behavior and behavior in `Linear`:
```suggestion
                weightmat if fp8 and not is_weight_param_quantized else None,
```

PR #964, opened 2024-06-25
[JAX] Add experimental internal used THD(packed) fused attn API

zlsh80826:
# Description

This PR exposes underly THD format API to JAX side and wrap it to be supported on multi-gpus.

Fixes # (issue)

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refractor

## Changes

- Unify `fused_attn_qkvpacked`, `fused_attn_kvpacked`, `fused_attn` to be a single function
- Add experimental `fused_attn_thd` API with the corresponding unit tests, this API requires users to configure a static `max_segments_per_seq` argument and ensure the "number of segments" in a seqeunce less and equal than this value
    - workspace memory consumptions: ~batch_size * max_sequence_length * max_segments_per_seq * num_heads
- Reduce `test_fused_attn` runtime
    - Test only 1 FP16 shape
    - Disable forward tests (backward tests include forward, we kept forward for the development and debugging).

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2247, opened 2025-10-08
Update minimum python version to 3.10 and add checks in CI

ksivaman:
# Description

We have a minimum python version requirement for our build (currently v3.8) but no way of checking whether new changes introduced violate this rule, e.g:

- #964 and #2012 used Structural Pattern Matching, support for which is only available from python 3.10 ([pep](https://peps.python.org/pep-0634/)).
- Several PRs also added the use of `functools.cache` as a decorator which is a 3.9+ feature.

Instead of changing the functionality to not use these features, I've taken the opportunity to update the minimum required python version to v3.10 since python v3.8 is at [end of life](https://devguide.python.org/versions/) (and v3.9 will also be by the end of the month) .

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Update minimum python version requirement to 3.10.
- Add python version requirements to the framework's `setup.py`s.
- Use a pre-commit hook to catch changes that violate the minimum version requirement.

# Checklist:

### DIFF
--- a/build_tools/utils.py
+++ b/build_tools/utils.py
@@ -12,10 +12,28 @@ import re
 import shutil
 import subprocess
 import sys
+import platform
 from pathlib import Path
 from importlib.metadata import version as get_version
 from subprocess import CalledProcessError
 from typing import List, Optional, Tuple, Union
+
+
+# Needs to stay consistent with .pre-commit-config.yaml
+python_min_version = (3, 10, 0)
+python_min_version_str = ".".join(map(str, python_min_version))
+if sys.version_info < python_min_version:
+    print(
+        f"You are using Python {platform.python_version()}. Python >={python_min_version_str} is"
+        " required."
+    )
+    sys.exit(-1)
+
+
+@functools.lru_cache(maxsize=None)
+def get_min_python_version_str() -> str:
+    """Returns the minimum supported python version"""
+    return python_min_version_str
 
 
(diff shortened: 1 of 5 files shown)


ksivaman:
Pipeline 36220067

build_tools/utils.py
```diff
@@ -12,12 +12,30 @@
...
+        f"You are using Python {platform.python_version()}. Python >={python_min_version_str} is"
+        " required."
+    )
+    sys.exit(-1)
+
+
+@functools.lru_cache(maxsize=None)
+def get_min_python_version_str() -> str:
+    """Returns the minimum supported python version"""
+    return python_min_version_str
```
timmoon10: Nits:
- Throwing an exception is more Pythonic than forcefully exiting. Also, I worry if this will mess up Pip's infrastructure for build failures.
- The version functions are fast, so memoizing seems like overkill.
```suggestion
def min_python_version() -> Tuple[int]:
    """Minimum supported Python version."""
    return (3, 10, 0)


def min_python_version_str() -> str:
    """String representing minimum supported Python version."""
    return ".".join(map(str, min_python_version()))


if sys.version_info < min_python_version():
    raise RuntimeError(
        f"Transformer Engine requires Python {min_python_version_str()} or newer, "
        f"but found Python {platform.python_version()}."
    )
```

PR #2248, opened 2025-10-08
[Common] Split cast/gated kernels by scaling mode

Oleg-Goncharov:
# Description

Breaks up the large `cast_kernels.cuh` and `cast_gated_kernels.cuh` into smaller headers organized by scaling mode.
No functional or behavior changes: code is moved, not modified. This improves structure, readability, and maintainability (easier to navigate/extend specific scaling paths). Build includes/exports updated accordingly; tests unaffected.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Broke up the large `cast_kernels.cuh` and `cast_gated_kernels.cuh` into smaller headers organized by scaling mode.

# Checklist:

### DIFF
--- a/transformer_engine/common/CMakeLists.txt
+++ b/transformer_engine/common/CMakeLists.txt
@@ -332,7 +332,7 @@ list(APPEND nvte_sources_with_fast_math fused_softmax/scaled_masked_softmax.cu
                                         fused_attn/context_parallel.cu
                                         fused_attn/kv_cache.cu)
 
-option(NVTE_BUILD_ACTIVATION_WITH_FAST_MATH "Compile activation kernels with --use_fast_math option" ON)
+option(NVTE_BUILD_ACTIVATION_WITH_FAST_MATH "Compile activation kernels with --use_fast_math option" OFF)
 if (NVTE_BUILD_ACTIVATION_WITH_FAST_MATH)
   list(APPEND nvte_sources_with_fast_math activation/gelu.cu
                                           activation/relu.cu


transformer_engine/common/cast/nvfp4/quantize_nvfp4.cuh
```diff
@@ -0,0 +1,681 @@
...
+
+  destroy_barriers<STAGES>(mbar, is_master_thread);
+#endif  // #if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
+}
+}  // namespace quantize_kernel
+
+// This kernel supports only two scaling cases:
+// 1. r16c0  - Rowwise NVFP4
+// 2. r16c32 - Rowwise NVFP4 AND Colwise MXFP8
+template <bool COMPUTE_ACTIVATIONS, typename ParamOP, float (*OP)(float, const ParamOP &)>
```
timmoon10: Do we actually support fused activation-cast kernels for NVFP4? If not, we should remove these template arguments so that we don't compile unnecessary kernels and so we prevent users from accidentally calling them. We should also remove them from the kernel, and modify `quantize_helper` so it errors out if you attempt something invalid.
```suggestion
```

tests/cpp/operator/test_cast_nvfp4_transpose.cu
```diff
@@ -663,12 +663,7 @@ std::vector<std::vector<size_t>> tensor_dims = {
 
 // Only GeLU activation tests are supported
 std::vector<ActivationType> Activation_types = {
-    ActivationType::Identity,
-    ActivationType::GeLU,
-    ActivationType::SiLU,
-    ActivationType::ReLU,
-    ActivationType::QGeLU,
-    ActivationType::SReLU,
```
timmoon10: If we support activation-cast kernels for NVFP4, we should test them somewhere. Maybe not in this file, but somewhere.

transformer_engine/common/cast/nvfp4/quantize_transpose_nvfp4.cuh
```diff
@@ -1389,8 +1159,10 @@ __global__ void __launch_bounds__(THREADS_NUM)
 
 template <bool COMPUTE_ACTIVATIONS, typename ParamOP, float (*OP)(float, const ParamOP &),
           bool use_2d_quantization>
```
timmoon10: If we don't support activation-quantize kernels for NVFP4, we should remove these template arguments. There are similar changes needed throughout this file.
```suggestion
template <bool use_2d_quantization>
```

transformer_engine/common/cast/nvfp4/quantize_transpose_nvfp4.cuh
```diff
@@ -4,40 +4,35 @@
...
 namespace transformer_engine {
+namespace dispatch {
+namespace nvfp4 {
 
-#if CUDA_VERSION > 12080
-namespace nvfp4_transpose {
-
-using RNG = decltype(curanddx::Generator<curanddx::philox4_32>() + curanddx::PhiloxRounds<10>() +
-                     curanddx::SM<800>() + curanddx::Thread());
+namespace quantize_transpose_kernel_ns {
```
timmoon10: Nit: `_ns` is both over-obvious (`quantize_transpose_kernel_ns::something` is obviously accessing a namespace) and non-obvious (I type "what does ns mean" into Google and it tells me "no skill", "national security", "nervous system"). Stylistically you should prefer to avoid acronyms, especially non-standard ones like this, so that your code is easy and _pleasant_ to read.
```suggestion
namespace quantize_transpose_kernel {
```

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -0,0 +1,186 @@
...
+          nvfp4::quantize_transpose<IS_ACT, ParamOP, OP, true>(
+              *input_tensor, noop_tensor, output_tensor, &quant_config_cpp, stream);
+        } else {
+          nvfp4::quantize_transpose<IS_ACT, ParamOP, OP, false>(
+              *input_tensor, noop_tensor, output_tensor, &quant_config_cpp, stream);
+        }
+      } else {
+        auto &global_amax = (output_tensor->amax.dptr != nullptr) ? output_tensor->amax
+                                                                  : output_tensor->columnwise_amax;
+        NVTE_CHECK((!IS_DBIAS && !IS_DACT && !IS_ACT),
```
timmoon10: If we don't support activation-cast kernels for NVFP4, we should move this check so it handles the case with optimized kernels.

transformer_engine/common/cast/nvfp4/quantize_nvfp4.cuh
Oleg-Goncharov: I intentionally left activation template arguments and all the activation related logic untouched, so we can easily enable it when/if it becomes the part of the FP4 recipe.
@ptrendx, should we keep it, or I just go ahead and clean up the kernel?
I also didn't want to add any functionality related modifications to this PR to not overwhelm it, and to do it separately in a following PRs. Since there are some parts of the NVFP4 code that need to be reviewed/changed anyways

transformer_engine/common/cast/nvfp4/quantize_transpose_nvfp4.cuh
Oleg-Goncharov: Yeah, it is a bit unclear to have it like this, but it generated an ambiguity error during the compilation due to the conflict with names; I'll rename it.

transformer_engine/common/cast/nvfp4/quantize_nvfp4.cuh
timmoon10: If we don't support them, we should at least error out if you attempt to run them. Avoiding unnecessary compilations would also be useful so we don't blow up compile time and binary size.

I'm fine deferring this if we want this PR to minimize functional changes, but we should aim to catch more of these errors.

ptrendx: @timmoon10 @Oleg-Goncharov Let's minimize changes in this PR and just do the code movement here. Otherwise it will be very hard to properly review if the functionality was not altered.

transformer_engine/common/cast/nvfp4/core_nvfp4.cuh
```diff
@@ -33,12 +33,12 @@ namespace nvfp4 {
...
 
 namespace quantization_and_transposition_SF {
-#if CUDA_VERSION > 12080
-#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
 // Used in transpose variant
 // Compute per-block E4M3 encoding/decoding scaling factor
 __device__ __forceinline__ nvfp4_scale_t compute_decoding_scaling_factor(const float block_amax,
                                                                          const float S_enc) {
+#if CUDA_VERSION > 12080
+#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
```
timmoon10: When doing these compile-time checks, we should throw an error in the "else" path rather than silently doing a noop. Here's an example from one of the `cvt_*` device functions:

https://github.com/NVIDIA/TransformerEngine/blob/4c572f040d3b8e9a29f512c9537b4f4a526f843e/transformer_engine/common/util/nvfp4_transpose.cuh#L188-L192

We should make a similar change in `quantization_SF::compute_decoding_scaling_factor`

timmoon10 (commented):
Overall LGTM once we iron out the test failures.

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -0,0 +1,167 @@
...
+  const auto scaling_mode = output->scaling_mode;
+  if ((scaling_mode != NVTE_DELAYED_TENSOR_SCALING) && !is_supported_by_CC_100()) {
+    NVTE_ERROR("Not supported by the Arch < 10.0");
+  }
+
+  constexpr bool allow_empty = false;
+  CheckInputTensor(input, "input");
+  CheckOutputTensor(*output, "output", allow_empty);
+
+  NVTE_CHECK(input.flat_last_dim() % 2 == 0, "Number of columns must be even.");
```
ptrendx:
```suggestion
  NVTE_CHECK(input.flat_last_dim() % 2 == 0, "Number of columns must be even, got ",
             input.flat_last_dim(), ".");
```

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -0,0 +1,167 @@
...
+  if (output->has_data()) {
+    is_fp8_rowwise_output = is_fp8_dtype(output->data.dtype);
+    NVTE_CHECK(output->flat_first_dim() == rows, "Wrong dimension of the output.");
+    NVTE_CHECK(output->flat_last_dim() == cols, "Wrong dimension of the output.");
+  }
+  if (output->has_columnwise_data()) {
+    is_fp8_colwise_output = is_fp8_dtype(output->columnwise_data.dtype);
+    NVTE_CHECK(output->flat_first_dim() == rows, "Wrong dimension of the output.");
+    NVTE_CHECK(output->flat_last_dim() == cols, "Wrong dimension of the output.");
+  }
```
ptrendx:
```suggestion
  if (output->has_data()) {
    is_fp8_rowwise_output = is_fp8_dtype(output->data.dtype);
    NVTE_CHECK(output->flat_first_dim() == rows, "Wrong dimension of the output. Expected first "
               "dimension ", rows, ", but got ", output->flat_first_dim(), ".");
    NVTE_CHECK(output->flat_last_dim() == cols, "Wrong dimension of the output. Expected last "
               "dimension ", cols, ", but got ", output->flat_last_dim(), ".");
  }
  if (output->has_columnwise_data()) {
    is_fp8_colwise_output = is_fp8_dtype(output->columnwise_data.dtype);
    NVTE_CHECK(output->flat_first_dim() == rows, "Wrong dimension of the output. Expected first "
               "dimension ", rows, ", but got ", output->flat_first_dim(), ".");
    NVTE_CHECK(output->flat_last_dim() == cols, "Wrong dimension of the output. Expected last "
               "dimension ", cols, ", but got ", output->flat_last_dim(), ".");
  }
```

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -0,0 +1,167 @@
...
+
+  const bool use_tma_kernels = is_fp8_rowwise_output && is_fp8_colwise_output && (cols % 32 == 0) &&
+                               is_supported_by_CC_100();
```
ptrendx: I also really don't like the name of this function, but let's fix that in the

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -0,0 +1,167 @@
...
+      break;
+    }
+    case NVTE_MXFP8_1D_SCALING: {
+      if (use_tma_kernels) {
+        Tensor dummy_tensor;  // grad
+        mxfp8::quantize_gated_dgated<false, ParamOP, ActOP, nullptr>(dummy_tensor, input, output, p,
+                                                                     stream);
+      } else {
+        NVTE_ERROR("Invalid input shape. Expected the last dimension to be divisible ",
+                   "by 32, got input of shape ", input.data.shape);
```
ptrendx: This error is potentially misleading since it is possible that the reason for use_tma_kernels to be false is lack of rowwise or columnwise output not being there. Also this is actually wrong since I believe the mxfp8 kernels can support only 1 of those outputs, right? Looking at it the second time it seems that the logic of setting the variables for rowwise and columnwise output is convoluted and not at all understandable. I will make a comment there.

transformer_engine/common/cast/dispatch/gated.cuh
ptrendx: The logic of is_fp8_rowwise_output (and columnwise too) is very confusing - they may be true even if the corresponding output does not exist. Also I'm not even sure if we necessarily need them - don't TMA kernels support BF16 output too?

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -0,0 +1,167 @@
...
+      if (use_tma_kernels) {
+        Tensor dummy_tensor;  // grad
+        fp8::cast_gated_dgated_tma<false, ParamOP, ActOP, nullptr>(dummy_tensor, input, output, p,
+                                                                   stream);
+      } else {
+        fp8::cast_gated<ParamOP, ActOP>(input, output, p, stream);
+      }
+      break;
+    }
+    case NVTE_MXFP8_1D_SCALING: {
```
ptrendx, line 58: You can make more checks based on this scaling mode - e.g. you could here make sure that the output dtypes are FP8 rather than doing those confusing variables earlier.

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -0,0 +1,167 @@
...
+
+  const size_t rows = gated_input.flat_first_dim();
+  const size_t cols = gated_input.flat_last_dim() / 2;
+  const size_t output_cols = 2 * cols;
+
+  CheckInputTensor(grad, "grad");
+  NVTE_CHECK(!is_fp8_dtype(grad.data.dtype), "Grad input must be in higher precision.");
+  NVTE_CHECK(grad.data.dtype == gated_input.data.dtype, "Types of both inputs must match.");
+  NVTE_CHECK(grad.flat_first_dim() == rows, "Wrong dimension of the grad input.");
+  NVTE_CHECK(grad.flat_last_dim() == cols, "Wrong dimension of the grad input.");
```
ptrendx: Please make the error message more easily actionable by telling the user what was expected and what was provided, similar to the suggestion above.

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -0,0 +1,167 @@
...
+  if (output->has_data()) {
+    is_fp8_rowwise_output = is_fp8_dtype(output->data.dtype);
+    NVTE_CHECK(output->flat_first_dim() == rows, "Wrong dimension of the output.");
+    NVTE_CHECK(output->flat_last_dim() == output_cols, "Wrong dimension of the output.");
+  }
+  if (output->has_columnwise_data()) {
+    is_fp8_colwise_output = is_fp8_dtype(output->columnwise_data.dtype);
+    NVTE_CHECK(output->flat_first_dim() == rows, "Wrong dimension of the output.");
+    NVTE_CHECK(output->flat_last_dim() == output_cols, "Wrong dimension of the output.");
+  }
```
ptrendx: Same issues here as in gated activation logic.

transformer_engine/common/cast/dispatch/dequantize.cuh
```diff
@@ -0,0 +1,53 @@
...
+namespace transformer_engine {
+namespace dispatch {
+
+inline void dequantize_helper(const Tensor &input, Tensor *output, cudaStream_t stream) {
+  CheckInputTensor(input, "cast_input");
+  CheckOutputTensor(*output, "cast_output");
+
+  switch (input.scaling_mode) {
+    case NVTE_DELAYED_TENSOR_SCALING: {
+      fp8::dequantize(input, output, stream);
```
ptrendx, line 33: We should test that the input is in FP8 before calling this function.

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -0,0 +1,186 @@
...
+        } else {
+          cast_transpose_fused<IS_DBIAS, IS_DACT, IS_ACT, float, ParamOP, OP>(
+              *input_tensor, activation_input_tensor, output_tensor, dbias_tensor, workspace_tensor,
+              stream);
+        }
+      } else if (output_tensor->has_data()) {
+        fp8::quantize<IS_DBIAS, IS_DACT, IS_ACT, ParamOP, OP>(
+            *input_tensor, activation_input_tensor, noop_tensor, output_tensor, dbias_tensor,
+            workspace_tensor, stream);
+      }
```
ptrendx, line 78: We should have else here to error out if the output does not have either of data/columnwise_data.

transformer_engine/common/cast/fp8/gated_fp8.cuh
```diff
@@ -0,0 +1,424 @@
...
+#include "../../util/math.h"
+#include "../../util/ptx.cuh"
+#include "../../util/vectorized_pointwise.h"
+#include "../../utils.cuh"
+
+namespace transformer_engine {
+namespace dispatch {
+namespace fp8 {
+namespace kernel {
+__device__ inline float sigmoidf(const float x) { return __frcp_rn(1.0f + __expf(-x)); }
```
ptrendx: Why do we have this function here and not in the math.h?

transformer_engine/common/cast/fp8/quantize_fp8.cuh
```diff
@@ -0,0 +1,580 @@
...
+          if constexpr (IS_DBIAS) {
+            common::reduce_dbias<IType>(workspace_ptr, dbias, dbias_rows, dbias_cols, stream);
+          });  // NOLINT(*)
+  );           // NOLINT(*)
+}
+
+namespace detail {
+using Empty = transformer_engine::Empty;
+__device__ inline float identity(float value, const Empty &) { return value; }
+}  // namespace detail
```
ptrendx, line 461: Do we need this?

transformer_engine/common/cast/nvfp4/core_nvfp4.cuh
```diff
@@ -0,0 +1,299 @@
...
+        "}"
+        : "=h"(out_4x)
+        : "l"(in_4x), "l"(reinterpret_cast<const uint64_t &>(scale)), "r"(rbits));
+  } else {
+    NVTE_DEVICE_ERROR(
+        "FP4 cvt PTX instructions are architecture-specific. "
+        "Try recompiling with sm_XXXa instead of sm_XXX.");
+  }
+  return *reinterpret_cast<fp4e2m1x4 *>(&out_4x);
+}
```
ptrendx: Functions like this should land in ptx.cuh, right?

transformer_engine/common/cast/dispatch/dequantize.cuh
Oleg-Goncharov: Moved it out from the `fp8::dequantize(input, output, stream)` to make it consistent

transformer_engine/common/cast/fp8/gated_fp8.cuh
Oleg-Goncharov: Ah, yes, moved it to `math.h`

transformer_engine/common/cast/fp8/quantize_fp8.cuh
Oleg-Goncharov: Yes, it is currently used by the `CastVectorizedUnaryKernelLauncher` and `CastVectorizedUnaryGradKernelLauncher` below

transformer_engine/common/cast/nvfp4/core_nvfp4.cuh
Oleg-Goncharov: Yes, moved it to the `ptx.cuh`

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -0,0 +1,163 @@
...
+#include "../fp8/gated_fp8.cuh"
+#include "../mxfp8/gated_mxfp8.cuh"
+
+namespace transformer_engine {
+namespace dispatch {
+
+template <typename ParamOP, float (*ActOP)(float, const ParamOP &)>
+void quantize_gated_helper(const NVTETensor nvte_input, NVTETensor nvte_output, ParamOP &p,
+                           cudaStream_t stream) {
+  using namespace dispatch;
```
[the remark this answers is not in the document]
ptrendx: Agreed.

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -0,0 +1,163 @@
...
+        NVTE_CHECK(is_fp8_dtype(output->data.dtype),
+                   "The type of the output tensor should be FP8.");
+      }
+      if (output->has_columnwise_data()) {
+        NVTE_CHECK(is_fp8_dtype(output->columnwise_data.dtype),
+                   "The type of the columnwise output tensor should be FP8.");
+      }
+      NVTE_CHECK(is_supported_by_CC_100(),
+                 "Gated FWD NVTE_MXFP8_1D_SCALING is only supported on SM 10.0+");
+      Tensor dummy_tensor;  // grad
```
ptrendx: Why not name it dummy_grad_tensor then?

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -0,0 +1,163 @@
...
+      mxfp8::quantize_gated</*IS_BWD=*/false, ParamOP, ActOP, nullptr>(dummy_tensor, input, output,
```
ptrendx: Putting the gradient as the first input in quantize_gated (which has a name that doesn't even suggest dealing with the backward pass) feels deeply wrong to me. Could we make it at least the second parameter?

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -0,0 +1,163 @@
...
+    default:
+      NVTE_ERROR("Not supported scaling mode: " + to_string(output->scaling_mode) + ".");
+  }
+}
+
+template <typename ParamOP, float (*ActOP)(float, const ParamOP &),
+          float (*DActOP)(float, const ParamOP &)>
+void quantize_dgated_helper(const NVTETensor nvte_grad, const NVTETensor nvte_gated_input,
+                            NVTETensor nvte_output, ParamOP &p, cudaStream_t stream) {
+  using namespace dispatch;
```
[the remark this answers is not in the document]
ptrendx: Agreed.

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -0,0 +1,189 @@
...
+  const Tensor *activation_input_tensor;
+  if constexpr (IS_DBIAS || IS_DACT) {
+    // backward - input is incoming gradient
+    input_tensor = convertNVTETensorCheck(grad);
+    activation_input_tensor = convertNVTETensor(input);
+  } else {
+    // forward = input is activation input
+    input_tensor = convertNVTETensorCheck(input);
+    activation_input_tensor = nullptr;
+  }
```
ptrendx: To be honest I really hoped that as part of this refactor we could fix this part of this API since it is super confusing and actually have consistent naming of the function arguments.

transformer_engine/common/cast/fp8/dequantize_fp8.cuh
```diff
@@ -0,0 +1,54 @@
...
+
+inline void dequantize(const Tensor &input, Tensor *output, cudaStream_t stream) {
+  const size_t N = product(input.data.shape);
+  TRANSFORMER_ENGINE_TYPE_SWITCH_FP8ONLY(
+      input.data.dtype, IType,
+      TRANSFORMER_ENGINE_TYPE_SWITCH_NON_FP8ONLY(
+          output->data.dtype, OType,
+
+          constexpr int nvec = 32 / sizeof(OType);
+          DequantizeParam p; p.scale_inv = reinterpret_cast<const fp32 *>(input.scale_inv.dptr);
```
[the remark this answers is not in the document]
ptrendx: Yup.

transformer_engine/common/cast/nvfp4/core_nvfp4.cuh
```diff
@@ -0,0 +1,112 @@
...
+  // const float S_dec_b = block_amax * rcp_6f;
+  // const nvfp4_scale_t S_dec_b_fp8 = static_cast<nvfp4_scale_t>(S_dec_b * S_enc);
+  // return S_dec_b_fp8;
+  // NOTE: Divide by 6.0f is not elegant and not efficient.
+  // However, this is part of the emulation code to ensure exact match.
+  using namespace detail;
+  constexpr float fp4_max = TypeExtrema<fp4e2m1>::max;  // 6.0f;
+  const float S_dec_b = block_amax / fp4_max * S_enc;
+  return static_cast<nvfp4_scale_t>(fminf(S_dec_b, TypeExtrema<float>::max));
+}
```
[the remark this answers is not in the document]
ptrendx: Yup, it seems to be basically the same function written twice.

transformer_engine/common/cast/nvfp4/core_nvfp4.cuh
```diff
@@ -0,0 +1,112 @@
...
+// Compute the global encode scale factor for a given global amax
+__device__ __forceinline__ float compute_global_encode_scaling_factor_FP4(const float global_amax) {
+  using namespace detail;
+  constexpr float fp8_max = TypeExtrema<fp8e4m3>::max;  // 448.0f;
+  constexpr float fp4_max = TypeExtrema<fp4e2m1>::max;  // 6.0f;
+  float global_encode_scale = fp8_max * fp4_max / global_amax;
+  // If scale is infinity, return max value of float32
+  global_encode_scale = fminf(global_encode_scale, TypeExtrema<float>::max);
+  // If global amax is 0 or infinity, return 1
+  if (global_amax == 0.0f || global_encode_scale == 0.0f) {
```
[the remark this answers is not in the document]
ptrendx: While this is wrong comment, it actually points the correct fact that something is skippable here - the global_amax == 0 check. This is because if amax is 0 then the scale is infinity and would be clamped by the fminf already (and if the amax is 0 we do not really care what the value of the scale is as long as it is finite so multiplied by 0 would not produce NaN). This kernel is so tiny though that it doesn't matter.

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
```diff
@@ -0,0 +1,111 @@
...
+    float4 current = static_cast<float4>(value.small_vec[i]);
+    OVec out;
+    out.data.elt[0] = static_cast<OType>(current.x * final_scale);
+    out.data.elt[1] = static_cast<OType>(current.y * final_scale);
+    out.data.elt[2] = static_cast<OType>(current.z * final_scale);
+    out.data.elt[3] = static_cast<OType>(current.w * final_scale);
+    output_vec[my_output_index + i] = out;
+  }
+}
+#endif  // CUDA_VERSION
```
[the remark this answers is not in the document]
ptrendx: Agreed.

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
```diff
@@ -0,0 +1,111 @@
...
+namespace dequantize_kernel {
+#if FP4_TYPE_SUPPORTED
+template <typename OType>
+__global__ void __launch_bounds__(512)
+    dequantize_fp4_kernel(const void *const input, OType *output, const fp8e4m3 *const scales,
+                          const float *const tensor_amax, const size_t N, const size_t M,
+                          const size_t scale_stride) {
+  const size_t thread_idx = blockIdx.x * blockDim.x + threadIdx.x;
+  const size_t x = thread_idx % M;
+  const size_t y = thread_idx / M;
```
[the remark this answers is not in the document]
ptrendx: That's maybe correct and I did intend to look into this kernel anyway since it doesn't work in some cases, but that is out of scope for this PR.

Oleg-Goncharov:
Regarding the `compute_decoding_scaling_factor` function used in NVFP4, I intentionally left two previously existing versions of the scaling factor computation (one is used in the more performant kernel which quantizes in the rowwise only; another is used in the quantize_transpose variant), to not introduce any changes that can impact numerics. But it definitely needs to be refactored in a separate PR.

transformer_engine/common/cast/dispatch/quantize.cuh
Oleg-Goncharov: I refactored the `dispatch/quantize.cuh` in a similar way as `dispatch/gated.cuh`, splitting into two helpers (FWD and BWD), there is a bunch of code duplicated, but now the logic of the code should be more intuitive.

transformer_engine/common/cast/cast.cu
```diff
@@ -184,9 +153,8 @@ void nvte_multi_tensor_quantize(const NVTETensor *inputs, NVTETensor *outputs,
   }
 
   for (int i = 0; i < num_tensors; i++) {
-    detail::quantize_helper<IS_DBIAS, IS_DACT, IS_ACT, Empty, nullptr>(
-        inputs[i], grad, outputs[i], dbias, workspace, nullptr,
-        detail::get_compute_stream(i % num_streams));
+    dispatch::quantize_fwd_helper<IS_ACT, Empty, nullptr>(
+        inputs[i], outputs[i], nullptr, detail::get_compute_stream(i % num_streams));
```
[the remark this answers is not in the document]
ptrendx: Agreed, we should pass the quant_config (quant_configs is actually misleading name since it suggests there are multiple of those) to those quantize functions.

transformer_engine/common/cast/cast.cu
```diff
@@ -77,11 +56,10 @@ void nvte_quantize_dbias(const NVTETensor input, NVTETensor output, NVTETensor d
...
-  constexpr bool IS_ACT = false;
   constexpr const NVTETensor activation_input = nullptr;
 
-  detail::quantize_helper<IS_DBIAS, IS_DACT, IS_ACT, Empty, nullptr>(
-      activation_input, input, output, dbias, workspace, nullptr, stream);
+  dispatch::quantize_bwd_helper<IS_DBIAS, IS_DACT, Empty, nullptr>(
+      input, activation_input, output, dbias, workspace, nullptr, stream);
 }
 
 void nvte_quantize_dbias_dgelu(const NVTETensor input, const NVTETensor activation_input,
```
ptrendx: Could we move those functions involving activations to the activation-specific files? That way we could make sure that we use fast math only for the activations (and then maybe actually turn it on by default?) and not for the entire cast.cu file.

ptrendx (approved):

PR #2256, opened 2025-10-10
Generalize quantization APIs for FP8/FP4/.. recipes

ksivaman:
# Description

#2177 and #2254 add support for the NVFP4 recipe. This PR is a follow-up to generalize the user-facing APIs to be agnostic of the recipe dtype and deprecate the old functions/arguments that explicitly mention FP8.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- `fp8_autocast` &rarr; `autocast` for PyTorch and Jax.
- `fp8_model_init` &rarr; `quantized_model_init` for PyTorch.
- Deprecate fp8 specific arguments to PyTorch's `make_graphed_callables` for more general versions.
- Introduce some utility functions into pytorch extensions public API (e.g. `is_fp8_available`). This PR starts off the process but will likely not be exhaustive. We want to expose as many necessary utils as possible so that users are discouraged to utilize internal TE functionality.

# Checklist:

### DIFF
--- a/tests/pytorch/nvfp4/test_nvfp4_module_exact.py
+++ b/tests/pytorch/nvfp4/test_nvfp4_module_exact.py
@@ -230,13 +230,13 @@ def check_nvfp4_module_versus_reference(
         grad_output = grad_output_val.clone().detach()
 
         # Native forward/backward
-        with autocast(enabled=True, recipe=nvfp4_recipe):
+        with te.autocast(enabled=True, recipe=nvfp4_recipe):
             # enable weight cache by giving is_first_microbatch
             y_native = native_module(x_native, is_first_microbatch=(step == 0))
         y_native.backward(grad_output)
 
         # Reference forward/backward
-        with autocast(enabled=True, recipe=nvfp4_ref_recipe):
+        with te.autocast(enabled=True, recipe=nvfp4_ref_recipe):
             y_ref = ref_module(x_ref)
         y_ref.backward(grad_output)
 
@@ -454,12 +454,12 @@ def check_nvfp4_layernorm_linear_versus_reference(
         grad_output = grad_output_val.clone().detach()
 
         # Native forward/backward
-        with autocast(enabled=True, recipe=nvfp4_recipe):
+        with te.autocast(enabled=True, recipe=nvfp4_recipe):
             y_native, ln_out_native = native_module(x_native, is_first_microbatch=(step == 0))
         y_native.backward(grad_output)
 
         # Reference forward/backward
-        with autocast(enabled=True, recipe=nvfp4_ref_recipe):
+        with te.autocast(enabled=True, recipe=nvfp4_ref_recipe):
             y_ref, ln_out_ref = ref_module(x_ref)
         y_ref.backward(grad_output)
 


transformer_engine/pytorch/quantization.py
timmoon10, line 1: Nit: "quantization" sounds more natural to me than "quantize". "Quantize" makes me think it's a quantize op, while "quantization" seems more about quantization infrastructure or configuration.

tests/pytorch/distributed/run_cast_master_weights_to_fp8.py
```diff
@@ -439,7 +439,7 @@ def _test_fsdp_cast_master_weights_to_fp8(quantization, dp_group):
     }
 
     # Create model with FP8 weights
-    with te.fp8.fp8_model_init(
+    with te.fp8.quantized_model_init(
```
timmoon10:
```suggestion
    with te.quantized_model_init(
```

tests/pytorch/distributed/run_cast_master_weights_to_fp8.py
```diff
@@ -573,7 +573,7 @@ def _test_cast_master_weights_to_fp8(quantization, dp_group):
     linear_kwargs = {"params_dtype": torch.bfloat16, "bias": False, "fuse_wgrad_accumulation": True}
 
     # Create model with FP8 weights
-    with te.fp8.fp8_model_init(
+    with te.fp8.quantized_model_init(
```
timmoon10:
```suggestion
    with te.quantized_model_init(
```

tests/pytorch/test_checkpoint.py
```diff
@@ -12,13 +12,15 @@
...
+
 import transformer_engine.pytorch as te
 
 from utils import make_recipe
 
 # Check supported quantization schemes
-fp8_available, reason_for_no_fp8 = te.fp8.FP8GlobalStateManager.is_fp8_available()
-mxfp8_available, reason_for_no_mxfp8 = te.fp8.FP8GlobalStateManager.is_mxfp8_available()
+fp8_available, reason_for_no_fp8 = te.quantize.FP8GlobalStateManager.is_fp8_available()
+mxfp8_available, reason_for_no_mxfp8 = te.quantize.FP8GlobalStateManager.is_mxfp8_available()
```
timmoon10:
```suggestion
fp8_available, reason_for_no_fp8 = te.is_fp8_available()
mxfp8_available, reason_for_no_mxfp8 = te.is_mxfp8_available()
```

transformer_engine/pytorch/quantize.py
```diff
@@ -0,0 +1,1365 @@
...
+@contextmanager
+def autocast(
+    enabled: bool = True,
+    calibrating: bool = False,
+    recipe: Optional["Recipe"] = None,
+    amax_reduction_group: Optional["dist_group_type"] = None,
+    _graph: bool = False,
+) -> None:
+    """
+    Context manager for FP8 or FP4 usage.
```
timmoon10: Nit: Generalizing to FP2, FP1, FP0, FP-1.
```suggestion
    Context manager for quantization schemes like FP8 or FP4.
```

tests/pytorch/test_checkpoint.py
ksivaman: The plan is to make a lot of the utility functions (including these) public APIs in order to discourage use and import of internal TE functions.

timmoon10: Yes, makes sense. We should use the public APIs here.

transformer_engine/pytorch/graph.py
```diff
@@ -855,6 +856,11 @@ def make_graphed_callables(
     fp8_recipe: Optional[Recipe] = None,
     fp8_group: Optional[dist_group_type] = None,
     fp8_weight_caching: bool = False,
+    enabled: SingleOrTuple[bool] = False,
+    calibrating: bool = False,
+    recipe: Optional[Recipe] = None,
+    amax_reduction_group: Optional[dist_group_type] = None,
+    cache_quantized_params: bool = False,
```
timmoon10: For backward compatibility, we should make the old and new kwargs have `None` as the default. The kwarg checking logic should look something like:
```
if fp8_enabled is not None:
    if enabled is not None:
        raise ValueError(
            "make_graphed_callables has deprecated `fp8_enabled` kwarg "
            "in favor of `enabled`, but both kwargs are set."
        )
    warnings.warn(
        "make_graphed_callables has deprecated `fp8_enabled` kwarg in favor of `enabled`. "
        "`fp8_enabled` will be removed in a future release."
        category=DeprecationWarning,
        stacklevel=2,
    )
    enabled = fp8_enabled
if enabled is None:
    enabled = False
```

transformer_engine/pytorch/graph.py
```diff
@@ -894,30 +905,73 @@ def make_graphed_callables(
...
+                            the `is_first_microbatch` boolean argument must be passed into the forward
+                            method for TransformerEngine modules. When storing primary weights in low precision
+                            using TE's `quantized_model_init` API and using an quantization aware optimizer,
+                            this arg must be set to `False` if calculating weight transposes' outside TE, e.g.,
+                            in the optimizer step.
 
     """
+
+    # Handle deprecated args.
+    if fp8_enabled is not False and enabled is False:
```
timmoon10: This won't raise a warning if the user passes in `fp8_enabled=False`.

transformer_engine/pytorch/graph.py
```diff
@@ -894,30 +905,73 @@ def make_graphed_callables(
...
+        )
+    if fp8_calibrating is not False and calibrating is False:
+        calibrating = fp8_calibrating
+        warnings.warn(
+            "Argument 'fp8_calibrating' is deprecated and will be removed"
+            " in a future release. Use argument 'calibrating' instead.",
+            category=DeprecationWarning,
+            stacklevel=2,
+        )
+    if fp8_recipe is not False and recipe is False:
```
timmoon10: `recipe` and `amax_reduction_group` should check `None` instead of `False` (or better, everything should use `None`).

ksivaman: Oops, copy/paste error

transformer_engine/pytorch/graph.py
ksivaman: For old kwargs, changing to `None` makes sense but not for the new kwargs. It would make the deprecation/warning logic cleaner but generally a bool is more natural type for an argument like `enabled`.

timmoon10: Eventually, but I think backward-compatibility takes precedence. Passing in `fp8_enabled=True` and `enabled=False` should result in an error.

tests/pytorch/test_checkpoint.py
```diff
@@ -12,13 +12,15 @@
...
+
 import transformer_engine.pytorch as te
 
 from utils import make_recipe
 
 # Check supported quantization schemes
-fp8_available, reason_for_no_fp8 = te.fp8.FP8GlobalStateManager.is_fp8_available()
-mxfp8_available, reason_for_no_mxfp8 = te.fp8.FP8GlobalStateManager.is_mxfp8_available()
+fp8_available, reason_for_no_fp8 = te.quantization.FP8GlobalStateManager.is_fp8_available()
+mxfp8_available, reason_for_no_mxfp8 = te.quantization.FP8GlobalStateManager.is_mxfp8_available()
```
timmoon10:
```suggestion
fp8_available, reason_for_no_fp8 = te.is_fp8_available()
mxfp8_available, reason_for_no_mxfp8 = te.is_mxfp8_available()
```

ksivaman:
For the thunder CI failure: https://github.com/Lightning-AI/lightning-thunder/pull/2635

transformer_engine/pytorch/fp8.py
```diff
@@ -2,1247 +2,52 @@
...
-from transformer_engine.common.recipe import (
-    Recipe,
-    DelayedScaling,
-    Format,
-    MXFP8BlockScaling,
-    Float8CurrentScaling,
-    Float8BlockScaling,
-    NVFP4BlockScaling,
-    CustomRecipe,
-)
```
timmoon10, line 26: I see you've removed the Thunder dependency on these imports (https://github.com/Lightning-AI/lightning-thunder/pull/2635), but we should restore the imports so we don't break older versions of Thunder.

transformer_engine/pytorch/utils.py
```diff
@@ -463,6 +463,13 @@ def is_bf16_compatible() -> None:
     return torch.cuda.get_device_capability()[0] >= 8
 
 
+def is_bf16_available() -> None:
```
timmoon10: Nit: This function does not match the convention for `is_fp8_available` and related functions. Those functions return both a bool and the reason string.

Since we are exposing `is_fp8_available` in the public API, it may be worth putting effort into making it nice. I actually prefer the way this `is_bf16_available` is implemented since it's more natural for an `is_*` function to return a bool. I see a few options:
- Change `is_fp8_available` to `check_fp8_available` so that it's harder to confuse with `is_bf16_available`.
- Change `is_fp8_available` to return a bool, and implement a separate function for `reason_fp8_is_not_available`.
- Change `is_bf16_available` to include a reason string.
- Remove `is_fp8_available` from the public API so we can delay making commitments. The tests can import them from `transformer_engine.pytorch.utils`.

If we choose to keep the current design, then we should at least fix the type hint:
```python
def is_bf16_available() -> bool:
```

transformer_engine/pytorch/quantization.py
```diff
@@ -143,47 +148,77 @@ def get_fp8_max(fp8_recipe: Recipe, fprop_tensor: bool = True) -> tex.DType:
     return Format.E5M2.value.max_fwd
 
 
-def is_fp8_available() -> Tuple[bool, str]:
+def is_fp8_available(return_reason: bool = False) -> Union[bool, Tuple[bool, str]]:
```
timmoon10, line 151: I think this design is a big improvement, especially since `if is_fp8_available(): ...` does what you expect and since the pattern is consistent between recipes. I do get some code smells from the variable return type. The alternative would be to expose `check_fp8_support` in the public API and use that when we want reason strings, but that also means expanding the functions we are committing to maintain. I don't have a strong opinion on which approach is better.

ksivaman: Yeah I went with lower public maintenance while still having the option to do return reason.

tests/pytorch/distributed/run_numerics_exact.py
```diff
@@ -47,7 +52,9 @@ def get_nvfp4_quantizer_factory():
     """
     Create a quantizer factory for NVFP4 reference implementation.
 
-    This factory returns NVFP4QuantizerRef instances with RHT and 2D quantization
+    This factory returns
+
+    Ref instances with RHT and 2D quantization
```
timmoon10:
```suggestion
    This factory returns NVFP4QuantizerRef instances with RHT and 2D quantization
```

tests/pytorch/distributed/run_gemm_with_overlap.py
```diff
@@ -171,12 +174,12 @@ def _parse_args(argv=None, namespace=None):
         opts.p2p = True
 
     if opts.atomic:
-        if not te.fp8.check_fp8_support():
+        if not te.quantization.check_fp8_support():
```
timmoon10: Hm, looks like the original implementation has a bug. This if statement would always evaluate `False`. 
```suggestion
        if not te.is_fp8_available():
```

PR #2264, opened 2025-10-10
Added support for DistOpt with offloading with MoE's

sanandaraj5597:
This PR adds support for CPU offloading with DistOpt for grouped linear module. This adds to the old path of CPU offloading.


transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -209,6 +209,19 @@ def forward(
                     if isinstance(weight, QuantizedTensorStorage):
                         weight.update_usage(columnwise_usage=True)
 
+            if cpu_offloading:
+                ctx.grad_added_to_main_grad = hasattr(weights[0], "grad_added_to_main_grad")
+
+                if ctx.grad_added_to_main_grad:
+                    # If you are passing torch.nn.Parameter through the Torch hooks, you will
```
pggPL, line 216: This is not True iirc, the problem was only when we offloaded weights - then making copy by using empty_like removed torch.parameter wrapper.

The fix in the backward code rebuilds parameter from tensor, which works for main grad, but may destroy sth in optimizer. Since we do not support weight offloading anymore, I suggest to remove problematic code from the backward.

sanandaraj5597: When using distributed optimizer, we should not change the weight object reference even when we do activation offloading. Passing weight objects to the torch hooks changes the weight object address, so we store the weight object pointers and then restoring it during backward for the distributed optimizer.

Without this fix, the runs are crashing for grouped linear usage with activation offloading.

PR #2270, opened 2025-10-13
[JAX] NVFP4 recipe with option to enable/disable SR, RHT, and 2D quantization

jberchtold-nvidia:
# Description

Support recipe flags for disabling SR, RHT, and 2D quantization for the NVFP4 recipe in TE/JAX.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Support `disable_stochastic_rounding` by not returning any stochastic rounding state in the QuantizeMeta returned from NVFP4 quantize config if `disable_stochastic_rounding` is set
- Support `disable_rht` by moving the `should_use_rht` check into the quantizer itself and check the quantize config for disabling RHT. A new `uses_rht` field has been added to the `ScaledTensor1x` to further consolidate this logic and avoid duplicate downstream re-checking of RHT, such as in the gemm.
- Support `disable_2d_quantization` by updating the NVFP4 quantize config to always return 1D quantization if `disable_2d_quantization` is set

- Fix bug where SR rng state on the `NVFP4Quantizer` was being lost across VJP boundaries due to missing pytree flatten and unflatten overrides resulting in not applying SR. Specialized `tree_flatten` and `tree_unflatten` methods have now been implemented on `NVFP4Quantizer` to fix this issue.
- Add checks to `test_helper.py` to test that NVFP4 quantizer's SR state is preserved across a VJP boundary. Confirmed the test fails before this fix to pytree flattening and after the fix it passes.
- Fix SR state sharding assertion when the input data is replicated along all or some axes.
- Fix usage of Python's default `hash` function, which is not deterministic across runs, in SR state creation. Now we use `hashlib` for consistent hashes and SR rng state with the same top-level seed.

# Checklist:

### DIFF
--- a/examples/jax/encoder/test_multiprocessing_encoder.py
+++ b/examples/jax/encoder/test_multiprocessing_encoder.py
@@ -672,7 +672,7 @@ class TestEncoder(unittest.TestCase):
     def test_te_nvfp4(self):
         """Test Transformer Engine with NVFP4"""
         result = self.exec(True, "NVFP4BlockScaling")
-        assert result[0] < 0.451 and result[1] > 0.79
+        assert result[0] < 0.451 and result[1] > 0.788
 
     @unittest.skipIf(not is_bf16_supported(), "Device compute capability 8.0+ is required for BF16")
     def test_te_bf16_shardy(self):
@@ -710,7 +710,7 @@ class TestEncoder(unittest.TestCase):
     def test_te_nvfp4_shardy(self):
         """Test Transformer Engine with NVFP4"""
         result = self.exec(True, "NVFP4BlockScaling", enable_shardy=True)
-        assert result[0] < 0.451 and result[1] > 0.79
+        assert result[0] < 0.451 and result[1] > 0.788
 
 
 if __name__ == "__main__":
--- a/examples/jax/encoder/test_single_gpu_encoder.py
+++ b/examples/jax/encoder/test_single_gpu_encoder.py
@@ -390,7 +390,7 @@ class TestEncoder(unittest.TestCase):
         self.args.use_fp8 = True
         self.args.fp8_recipe = "NVFP4BlockScaling"
         actual = train_and_evaluate(self.args)
-        assert actual[0] < 0.477 and actual[1] > 0.772
+        assert actual[0] < 0.477 and actual[1] > 0.769
 
 
 if __name__ == "__main__":


transformer_engine/jax/quantize/tensor.py
```diff
@@ -175,6 +175,7 @@ class ScaledTensor1x(AbstractBaseTensor1x, ScaledTensor):
         is_colwise: Whether the tensor uses column-wise quantization
         data_layout: The data_layout specification for the tensor
         flatten_axis: The quantization axis for the tensor
+        uses_rht: Whether the tensor uses the Randomized Hadamard Transform (RHT)
```
phu0ngng: Nitpick: I suggest naming it `applied_rht` or `with_rht` as RHT is already applied to the data when the tensor is created.

transformer_engine/jax/quantize/tensor.py
```diff
@@ -364,6 +368,7 @@ def __init__(
             is_colwise=is_colwise,
             data_layout=data_layout,
             flatten_axis=flatten_axis,
+            uses_rht=False,
```
phu0ngng: Could you add a TODO here that we will need to update this when we work on making GroupedGEMM + NVFP4 work?

transformer_engine/jax/quantize/tensor.py
```diff
@@ -626,6 +636,8 @@ def create_2x(
             group_sizes: Array containing the size of each group (default: None)
             original_shape: The original shape of the tensor before grouping (default: None)
             group_axis: The axis along which grouping is performed (default: 0)
+            rowwise_uses_rht: Whether the row-wise tensor uses the Randomized Hadamard Transform (RHT) (default: False)
```
phu0ngng: We don't support rowwise + RHT at all, right? If we leave this here, I think we should probably add an assertion.

transformer_engine/jax/quantize/quantizer.py
```diff
@@ -603,6 +607,30 @@ def __post_init__(self):
...
+
+        Args:
+            aux_data: Auxiliary data containing quantizer parameters
+            children: Unused children data
+
+        Returns:
+            A reconstructed Quantizer instance
+        """
+        stochastic_rounding_rng_state = children[0]
+        return cls(*aux_data, stochastic_rounding_rng_state=stochastic_rounding_rng_state)
```
phu0ngng, line 630: I would reorder the attrs in the NVFP4Quantizer class so that all the traceable objects, i.e the `sr_rng_state`, are defined before the static attributes, then no need to overload the `tree_unflatten`.

transformer_engine/jax/quantize/quantizer.py
```diff
@@ -688,8 +716,11 @@ def _quantize_func(self, x, is_colwise=False, dq_dtype=None, flatten_axis=-1) ->
             flatten_axis = x.ndim - flatten_axis
         x_shape = x.shape
 
-        if should_use_rht(self.scaling_mode, is_colwise=is_colwise):
-            # We only apply RHT for 1D colwise nvfp4
+        # Split q_layout into rowwise and colwise components for JAX impl so we don't apply RHT incorrectly to the wrong axis
+        use_rht = self.should_use_rht(
+            q_layout=QuantizeLayout.COLWISE if is_colwise else QuantizeLayout.ROWWISE
+        )
+        if use_rht:
```
phu0ngng, line 719: I suggest making the Quantizer class store `use_rht` enum, and this attribute is set when the Quantizer is created in the QuantizerFactory. 
The Quantizer should simply apply RHT when it is told to do so and raise an error if it is not supported.

transformer_engine/jax/quantize/quantizer.py
```diff
@@ -790,8 +821,22 @@ def repeat_to_shape(x, target_shape):
             scaling_mode=self.scaling_mode,
             dq_dtype=dq_dtype,
             flatten_axis=rowwise_flatten_axis,
+            uses_rht=use_rht,
         )
 
+    def should_use_rht(self, q_layout=None):
```
phu0ngng: Same here, the quantizer probably should not have the right to decide whether to apply RHT.

tests/jax/test_helper.py
```diff
@@ -171,5 +233,16 @@ def test_autocast_nvfp4_block_scaling(self):
...
+
+        bs = NVFP4BlockScaling(
+            disable_stochastic_rounding=True,
+            disable_rht=True,
+            disable_2d_quantization=True,
+        )
+        with fp8_autocast(enabled=True, fp8_recipe=bs, mesh_resource=MeshResource()):
+            self.assertTrue(get_quantize_config().is_fp8_enabled())
+            self._compare_nvfp4_scaling(bs)
+            self._compare_nvfp4_scaling_quantizers(bs)
```
phu0ngng, line 253: Great to have this test! Thanks!

transformer_engine/jax/quantize/quantizer.py
jberchtold-nvidia: I'd like to encapsulate the RHT logic on a per-scaling mode or per-recipe basis. So I'm hesitant to put it directly in the QuantizerFactory where we'd need to do `if is_nvfp4 and ...: use_rht=True`

But I'll see if I can move this to the recipe logic in `helper.py` and use the quantization config to pass that as the default to QuantizerFactory. For unit tests, we can manually set this `use_rht` field when creating quantizers

transformer_engine/jax/quantize/tensor.py
jberchtold-nvidia: Updated

transformer_engine/jax/quantize/tensor.py
jberchtold-nvidia: Updated

transformer_engine/jax/quantize/tensor.py
jberchtold-nvidia: Added an assert here

transformer_engine/jax/quantize/quantizer.py
jberchtold-nvidia: The ordering of fields is base class + child class. In this particular case, NVFP4 only has a single q_dtype so we could redefine that with a default value. But if the base class adds in a new field that isn't redefined on the child class here, then this could break again, or we get type errors when unflattening. This should be caught by the new test I added, but I'm not sure if relying on all child classes to re-define all the parent fields if they define any traceable fields is the best solution. Unless there's a way to re-order the child class fields that I'm not aware of. Thanks!

transformer_engine/jax/quantize/quantizer.py
jberchtold-nvidia: Done

transformer_engine/jax/quantize/quantizer.py
jberchtold-nvidia: I've moved this decision on whether to use RHT to helper.py so it is decided at the recipe level. I've also added dedicated tests for testing RHT quantization and gemms now that it is not enabled by default in the quantizer class. I've also updated `test_helper.py` to ensure `use_rht` is passed through correctly through VJPs.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -168,13 +168,13 @@ def _quantize_gemm_operands(lhs, rhs, lhs_quantizer, rhs_quantizer, contracting_
...
+        return isinstance(q, ScaledTensor1x) and q.has_rht_applied
 
-    assert uses_rht(lhs_q) == uses_rht(rhs_q), (
-        "With NVFP4_1D_SCALING, if one operand is colwise quantized, the other must be colwise"
-        " quantized as well. This is to ensure the RHT is applied to both and will cancel out in"
-        " the GEMM."
+    assert has_rht_applied(lhs_q) == has_rht_applied(rhs_q), (
+        "With NVFP4_1D_SCALING, if one operand is quantized with RHT, the other must be quantized"
+        " with RHT as well. This is to ensure the RHT is applied to both and will cancel out in the"
+        " GEMM."
```
phu0ngng, line 177: Could we simplify this block to 
```
assert lhs_q.has_rht_applied == rhs_q.has_rht_applied, msg
```
?

phu0ngng (approved):
I like the new design very much!
Left a small comment but please feel free to merge it.

transformer_engine/jax/cpp_extensions/gemm.py
jberchtold-nvidia: The lhs_q and rhs_q are not guaranteed to be `ScaledTensor1x`. If we are running a non-quantized GEMM, they could be `NoScaleTensor` which doesn't have `has_rht_applied`.

PR #2276, opened 2025-10-16
[PyTorch] Decouple python quantization classes and refactor custom quantization

negvet:
# Description

Decouple python base quantization classes (aka API).
Keep "custom" instead of "experimental" naming convention.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
```diff
@@ -12,10 +12,11 @@
 import transformer_engine.pytorch as te
 import transformer_engine_torch as tex
 from transformer_engine.pytorch import NVFP4Quantizer
-from transformer_engine.common.recipe import NVFP4BlockScaling
+from transformer_engine.pytorch.custom_recipes.quantization_nvfp4 import NVFP4QuantizerRef
+from transformer_engine.pytorch.custom_recipes import utils
 from transformer_engine.pytorch.constants import TE_DType
-from transformer_engine.pytorch.experimental.quantization_nvfp4 import NVFP4QuantizerRef
-from transformer_engine.pytorch.experimental import utils
+from transformer_engine.pytorch.fp8 import fp8_autocast, get_fp4_te_dtype
```
ksivaman:
```suggestion
```

ksivaman: Deprecated functions. They also seem unused here.

tests/pytorch/test_custom_recipe.py
```diff
@@ -17,6 +17,49 @@
     Float8CurrentScalingQuantizer,
 )
 import transformer_engine.pytorch.ops as te_ops
+from transformer_engine.pytorch.module.grouped_linear import GroupedLinear
```
ksivaman:
```suggestion
```

ksivaman: Duplicate import

ksivaman (commented):
The main blocking issue is with some of the imports.

Could you detail in the description the motivation behind this change aside from the naming from `experimental` &rarr; `custom`? Why do we want to have a separate `quantization_base.py` outside of the `tensor` dir and why the decoupling of that from the `quantization_helpers` file which is inside the `tensor` dir?

negvet:
> Why do we want to have a separate quantization_base.py outside of the tensor dir and why the decoupling of that from the quantization_helpers file which is inside the tensor dir?

The main idea is to help users more easily create custom recipes.
Here is the benefit: when subclassing, users need to learn only classes from quantization_base.py (three classes and two functions as of now).
This is an attempt to abstract away all TE complexity from the user.

For us, this can enable setting guards for this API change - if this file is getting changes, certain people might be alarmed etc. 
This will help to keep API stable. Usually API is located separately, quantization_base.py is not an API because it contains some implementations (there are no pure abstractions), but still it is good to keep it separate at a pytorch top level.

For the quantization_helpers, it contains some TE internal implementations that a general user might not need to know about, so I kept it in tensor submodule. With this PR, tensor submodule becomes a place for concrete implementations, but not base classes.

ksivaman:
Fair. I think the name `quantized_tensor.py` might be more apt instead of `quantization_base.py` then. Because it is more representative of the methods in it and we already have a `quantization.py` file in the same dir which is mostly unrelated to this.

PR #2277, opened 2025-10-16
fall back after failing ldconfig-based lib loading for cuDNN

getim:
# Description
This PR is a follow-up to #1989 and applies the suggested changes from the comments on that PR.

This change accomplishes two things:
* It avoids the `shell=True` as suggested by the original comment
* The `ldconfig -p | grep <pattern>` fails when the library we're searching for is not found. Since this command is executed with `check_output`, a `CalledProcessError` is raised. That means the final line of the function that uses `LD_LIBRARY_PATH` is effectively unreachable.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes
- Remove the early filtering with `grep` to avoid `shell=True` and command failures. 

# Checklist:

### DIFF
--- a/transformer_engine/common/__init__.py
+++ b/transformer_engine/common/__init__.py
@@ -252,9 +252,7 @@ def _load_cudnn():
         return handle
 
     # Attempt to locate libcudnn via ldconfig
-    libs = subprocess.check_output(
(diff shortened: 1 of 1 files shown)


timmoon10 (approved):
LGTM, thanks for the fix.

PR #2278, opened 2025-10-16
Wheels for cuda 13

ksivaman:
# Description

Support PyPI wheels for cuda13 and make required changes to build scripts and dependency structure.

## Type of change

- [x] Infra/Build change
- [x] Code refactoring

## Changes

- Change dependency structure of the meta package such that the core package is now a dependency of the framework extensions. This allows us to detect the available cudart version dynamically via the frameworks when the sdists are compiled on the user machine.
- Add extra dependency for the core package so that we retain being able to use TE without any framework extensions. `te[core_cu12]` and `te[core_cu13]` will install the required package for the specified cuda version and `te[core]` will default to a version of our choice, for now cuda 12. This is a breaking change for users with this workflow as `pip install transformer-engine` now installs just the empty package with no functionality.
- Fix script to build wheel in order to accommodate changes introduced in #2177.
- Miscellaneous changes to wheel related scripts for ease of use and compiling all needed artifacts for PyPI.

# Checklist:

### DIFF
--- a/transformer_engine/common/gemm/cublaslt_gemm.cu
+++ b/transformer_engine/common/gemm/cublaslt_gemm.cu
@@ -140,6 +140,16 @@ GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cubla
       } else {
         NVTE_CHECK(!is_fp8_dtype(ret.Atype), "Input A is missing column-wise usage");
       }
+    } else if (nvte_is_non_tn_fp8_gemm_supported() && !A.has_data()) {
+      // Blackwell supports any GEMM layout for FP8, so we can use column-wise/transposed
+      // data  with the mirrored transpose-flag if we don't have row-wise data.
+      NVTE_CHECK(A.has_columnwise_data() && is_fp8_dtype(A.columnwise_data.dtype),
+                 "Input A is missing column-wise usage");
+      ret.A = A.columnwise_data.dptr;
+      ret.transA = is_A_transposed ? CUBLAS_OP_N : CUBLAS_OP_T;
+      ret.Atype = A.columnwise_data.dtype;
+      ret.A_scale_inv = A.columnwise_scale_inv.dptr;
+      ret.lda = is_A_transposed ? m : k;
     }
 
     if (is_fp8_dtype(ret.Atype)) {
@@ -221,6 +231,16 @@ GemmParam CanonicalizeGemmInput(const transformer_engine::Tensor &A, const cubla
       } else {
         NVTE_CHECK(!is_fp8_dtype(ret.Btype), "Input B is missing column-wise usage");
       }
+    } else if (nvte_is_non_tn_fp8_gemm_supported() && !B.has_data()) {
+      // Blackwell supports any GEMM layout for FP8, so we can use column-wise/transposed
+      // data with the mirrored transpose-flag if we don't have row-wise data.
+      NVTE_CHECK(B.has_columnwise_data() && is_fp8_dtype(B.columnwise_data.dtype),
+                 "Input B is missing column-wise usage");
+      ret.B = B.columnwise_data.dptr;
+      ret.transB = is_B_transposed ? CUBLAS_OP_N : CUBLAS_OP_T;
+      ret.Btype = B.columnwise_data.dtype;
+      ret.B_scale_inv = B.columnwise_scale_inv.dptr;
+      ret.ldb = is_B_transposed ? k : n;
     }
 
     if (is_fp8_dtype(ret.Atype)) {


ksivaman:
pipeline 36803310

transformer_engine/common/__init__.py
```diff
@@ -8,35 +8,53 @@
...
     try:
-        metadata(package)
+        distribution(package)
     except PackageNotFoundError:
         return False
     return True
 
 
+@functools.lru_cache(maxsize=None)
+def _is_package_installed_via_pypi(package) -> bool:
```
timmoon10: This function is misleading. It only checks TE and it will return `True` if TE was installed from a non-PyPI wheel.
```suggestion
def _is_transformer_engine_installed_from_wheel(package) -> bool:
```

transformer_engine/common/__init__.py
```diff
@@ -130,46 +161,66 @@ def load_framework_extension(framework: str) -> None:
...
         extra_dep_name = "pytorch"
 
+    # Find the TE packages. The core and framework packages can only be installed via PyPI.
+    # For the `transformer-engine` package, we need to check explicity.
+    te_core_installed, te_core_package_name, te_core_version = get_te_core_package_info()
+    te_framework_installed = _is_package_installed(module_name)
+    te_installed = _is_package_installed("transformer_engine")
+    te_installed_via_pypi = _is_package_installed_via_pypi("transformer_engine")
+
+    assert te_installed, "Could not find `transformer-engine`."
```
timmoon10: This is checking the output of `_is_package_installed("transformer_engine")`:
```suggestion
    assert te_installed, "Could not find `transformer_engine`."
```

transformer_engine/common/__init__.py
```diff
@@ -8,35 +8,53 @@
...
 
 
+@functools.lru_cache(maxsize=None)
+def _is_package_installed_via_pypi(package) -> bool:
+    """Check if the given package is installed via PyPI."""
+
+    if not _is_package_installed(package):
+        return False
+
+    te_dist = distribution("transformer-engine")
```
timmoon10: Do we need to check that this `transformer-engine` package matches the currently loaded `transformer_engine` module? It's possible that the user has multiple TE installs (e.g. the system comes with a PyPI system install in `/usr/local/lib/python3.12/dist-packages` and a non-root user has built TE locally in `~/.local/lib/python3.12/site-packages`).

transformer_engine/common/__init__.py
```diff
@@ -130,46 +161,66 @@ def load_framework_extension(framework: str) -> None:
...
-            "transformer_engine_cu12"
-        ), "Could not find `transformer-engine-cu12`."
-        assert (
-            version(module_name)
-            == version("transformer-engine")
-            == version("transformer-engine-cu12")
-        ), (
-            "TransformerEngine package version mismatch. Found"
+    # extension are all installed via PyPI and have matching versions.
+    if te_framework_installed:
```
timmoon10, line 176: This check matches the existing behavior, so not critical to fix in this PR. However, if the user has multiple TE builds, it's perfectly possible that there is an irrelevant `transformer_engine_torch` package installed somewhere that we will never import, so version checking would cause unnecessary failures.

transformer_engine/common/__init__.py
```diff
@@ -130,46 +161,66 @@ def load_framework_extension(framework: str) -> None:
...
-            )
-
     # After all checks are completed, load the shared object file.
     spec = importlib.util.spec_from_file_location(module_name, _get_shared_object_file(framework))
     solib = importlib.util.module_from_spec(spec)
     sys.modules[module_name] = solib
     spec.loader.exec_module(solib)
 
 
+def sanity_checks_for_pypi_installation() -> None:
```
timmoon10, line 195: Similar questions as in `load_framework_extension`.

transformer_engine/common/__init__.py
```diff
@@ -112,6 +130,19 @@ def _get_shared_object_file(library: str) -> Path:
     )
 
 
+def get_te_core_package_info() -> Tuple[bool, str, str]:
+    """
+    Check if Tranformer Engine core package is installed.
+    Returns the module name and version if found.
+    """
+
+    te_core_packages = ("transformer-engine-cu12", "transformer-engine-cu13")
```
timmoon10, line 139: I worry about an edge case when there are multiple TE installs with different CUDA versions. This logic prioritizes CUDA 12, so we might load the wrong one. Not sure the beset way to handle this though. Is there a way to interrogate a PyPI package's upstream dependencies?

transformer_engine/common/__init__.py
ksivaman: With the package structure that we have, it is probably a good thing in this case in order to avoid loading the wrong .so or possibly some other bug that could come with having multiple installs. I presume that this is a rare use case.

General note and for the questions below as well: We ship 5 packages in PyPI and have a different structure for source vs wheel installation. Dealing with validating multiple installations of TE given the combination of all of these different packages and versions is a non-trivial problem out of the scope of this PR.

fzyzcjy:
Hi, may I know when will the cu13 be able to be installed? I would like to install TE on cuda13.0 hardware and thus would be great to have it. or, is there any workaround for it?

ksivaman:
@fzyzcjy Wheels for TE core cuda 13 will be available from `release_v2.9` which will be in ~2 weeks. Until then, TE can be used on cuda13 via [source installation](https://github.com/NVIDIA/TransformerEngine/blob/main/docs/installation.rst#installation-from-source).

yiakwy-xpu-ml-framework-team:
I build local transformer engine (becuase megatron-core and megatron bridge have been broken)

```
# v2.9 70f536662ae10a62a54f4ed1ba92e3314c5cfd69, cuda 12.8, pytroch 2.9, FA 2.8.3, FA3, nvshmem supported, H100 DGX
export NVTE_FRAMEWORK=pytorch
export NVTE_RELEASE_BUILD=1
export NVTE_UB_WITH_MPI=1
export MPI_HOME=/opt/openmpi/
export NVTE_ENABLE_NVSHMEM=1
export NVSHMEM_HOME=/usr/local/nvshmem
export NVTE_FRAMEWORK=pytorch
pip install -v -e .[pytorch]
```

截屏2025-11-20 21 27 13

here is what installed :

截屏2025-11-20 21 34 24

However when I import transformer engine

截屏2025-11-20 21 36 36

Whay are we checking this ? Is this necessary ?

截屏2025-11-20 21 43 15

> te_core_packages = ("transformer-engine-cu12", "transformer-engine-cu13")

NOTE : pip3 install --no-build-isolation transformer-engine[pytorch] is already broken

and this line should be removed :

> assert te_installed_via_pypi, "Could not find `transformer-engine` PyPI package."

 @ksivaman @fzyzcjy @timmoon10

yiakwy-xpu-ml-framework-team:
Here is right version after fix :

截屏2025-11-20 22 08 18

PR #2279, opened 2025-10-16
Overhaul the compilation for the arch-specific features

ptrendx:
# Description

This pull request refactors and improves the CUDA architecture handling and build process for the Transformer Engine, particularly in relation to arch-specific feature support.

Key changes include:

**Build System and CUDA Architecture Handling:**

* The logic for setting `CMAKE_CUDA_ARCHITECTURES` is refactored, with a new mechanism to separate the compilation for generic and specific architectures. This allows for more precise control over which files are compiled for which GPU architectures.
* The source files are now grouped into `cpp`, `cuda`, and `cuda_arch_specific` categories, allowing for targeted compilation flags and better build parallelism.

**Architecture-Specific Code Improvements:**

* The use of architecture-specific PTX instructions for FP4 conversions is now guarded with more targeted `constexpr` checks (using `ARCH_HAS_STOCHASTIC_ROUNDING` and `ARCH_BLACKWELL_FAMILY` macros) instead of purely preprocessor macros, leading to better error reporting if the wrong architecture is targeted (e.g. compiling the file that needs 100a features for 100 architecture will now fail at compile time).
* The CUDA version check for enabling FP4 features is corrected (`#if CUDA_VERSION >= 12080` instead of `> 12080`).

**Other Notable Changes:**

* The list of default CUDA architectures in `build_tools/utils.py` is updated to remove certain architectures (`100a`, `103a`) for CUDA 13.0+, aligning with the new build logic.
* The application of the `--use_fast_math` compile option is fixed to append to the compilation options, rather than replacing them.

## Type of change

- [x] Infra/Build change
- [x] Code refactoring

### DIFF
--- a/transformer_engine/common/utils.cuh
+++ b/transformer_engine/common/utils.cuh
@@ -16,8 +16,8 @@
 #endif
 
 #if !defined(__CUDACC_RTC__)
-#include <cstdint>
 #include <cassert>
+#include <cstdint>
 #else
 // Importing C++ standard headers is a pain with NVRTC
 using uint8_t = unsigned char;


transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
```diff
@@ -97,22 +97,23 @@ cutlass::Array<cutlass::float_e2m1_t, 8>
...
-#if CUDA_ARCH_HAS_FEATURE_SM10X_ALL
-  auto output_ptr = reinterpret_cast<uint16_t *>(&output);
-  asm volatile( \
-      "{\n" \
-      "cvt.rs.satfinite.e2m1x4.f32   %0, {%5, %4, %3, %2}, %10;\n" \
-      "cvt.rs.satfinite.e2m1x4.f32   %1, {%9, %8, %7, %6}, %11;\n" \
-      "}" \
-      : "=h"(output_ptr[0]),
+  constexpr bool has_rs = ARCH_HAS_STOCHASTIC_ROUNDING;
+  if constexpr (has_rs) {
```
jberchtold-nvidia, line 101: This will only trigger an error at runtime if we try to build a kernel with SR on an unsupported arch, right?

Could we use a helper function that wraps this in a template here so we can trigger a static_assert at compile time if this function is attempted to be instantiated in any kernels on unsupported archs?

Haven't tested it, but maybe something like this:
```
template <bool STOCHASTIC_ROUNDING>
CUTLASS_DEVICE
cutlass::Array<cutlass::float_e2m1_t, 8>
StochasticNumericConverterBaseTemplate(cutlass::Array<float, 8> const &input, cutlass::Array<uint32_t, 2> const &rbits);

template <>
CUTLASS_DEVICE
cutlass::Array<cutlass::float_e2m1_t, 8>
StochasticNumericConverterBaseTemplate<true>(cutlass::Array<float, 8> const &input, cutlass::Array<uint32_t, 2> const &rbits) {
  using result_type = cutlass::Array<cutlass::float_e2m1_t, 8>;
  result_type output;
  auto output_ptr = reinterpret_cast<uint16_t *>(&output);
  asm volatile( \
      "{\n" \
      "cvt.rs.satfinite.e2m1x4.f32   %0, {%5, %4, %3, %2}, %10;\n" \
      "cvt.rs.satfinite.e2m1x4.f32   %1, {%9, %8, %7, %6}, %11;\n" \
      "}" \
      : "=h"(output_ptr[0]),
        "=h"(output_ptr[1])
      : "f"(input[0]), "f"(input[1]), "f"(input[2]), "f"(input[3]),
      "f"(input[4]), "f"(input[5]), "f"(input[6]), "f"(input[7]),
      "r"(rbits[0]), "r"(rbits[1]));
  return output;
}

template <>
CUTLASS_DEVICE
cutlass::Array<cutlass::float_e2m1_t, 8>
StochasticNumericConverterBaseTemplate<false>(cutlass::Array<float, 8> const &input, cutlass::Array<uint32_t, 2> const &rbits) {
  static_assert(false, "FP4 cvt PTX instructions are architecture-specific. "
        "Try recompiling with sm_XXXa instead of sm_XXX.");
}


// This wrapper helper may trigger instantiation of the template, even if the wrapper isn't called. So we may need to invoke StochasticNumericConverterBaseTemplate<ARCH_HAS_STOCHASTIC_ROUNDING> at the call sites, I'm unsure
CUTLASS_DEVICE
cutlass::Array<cutlass::float_e2m1_t, 8>
StochasticNumericConverterBase(cutlass::Array<float, 8> const &input, cutlass::Array<uint32_t, 2> const &rbits) {
  return StochasticNumericConverterBaseTemplate<ARCH_HAS_STOCHASTIC_ROUNDING>(input, rbits);
}
```

ptrendx: There are 3 possibilities:
 - you compile this file with 100a or 103a - the `if` branch passes and you get the good PTX
 - you compile this file with e.g. 90 - the `if` branch failes and you get the runtime error
 - you compile this file with 100 - static_assert triggers in the ARCH_HAS_STOCHASTIC_ROUNDING macro.

Ultimately we would like to remove the runtime assert to add the emulation path, but that is something I would like in the follow up to this PR.

jberchtold-nvidia: Oh nice! I missed the logic in the ARCH_HAS_STOCHASTIC_ROUNDING macro, I thought it was a simple bool. Then this is already addressed in a more generic way at compile time, awesome! Resolving this thread

transformer_engine/common/util/ptx.cuh
```diff
@@ -313,6 +449,7 @@ __device__ __forceinline__ void mul_cvt_4x(fp4e2m1x4 &out, const Tx2 &in01, cons
 // SIMD like "Fused" cast + multiplication (x2)
 __device__ __forceinline__ void mul_cvt_2x(fp8e4m3x2 &out, const floatx2 &in,
                                            const floatx2 &scale) {
+#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
```
Oleg-Goncharov: mul.f32x2 instruction is supported from the sm100 arch and higher
```suggestion
#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
```

transformer_engine/common/util/ptx.cuh
```diff
@@ -325,10 +462,12 @@ __device__ __forceinline__ void mul_cvt_2x(fp8e4m3x2 &out, const floatx2 &in,
       : "=h"(reinterpret_cast<uint16_t &>(out))
       : "l"(reinterpret_cast<const uint64_t &>(in)),
         "l"(reinterpret_cast<const uint64_t &>(scale)));
+#endif  // (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
 }
 
 __device__ __forceinline__ void mul_cvt_2x(fp8e5m2x2 &out, const floatx2 &in,
                                            const floatx2 &scale) {
+#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
```
Oleg-Goncharov:
```suggestion
#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
```

transformer_engine/common/util/ptx.cuh
```diff
@@ -341,9 +480,11 @@ __device__ __forceinline__ void mul_cvt_2x(fp8e5m2x2 &out, const floatx2 &in,
       : "=h"(reinterpret_cast<uint16_t &>(out))
       : "l"(reinterpret_cast<const uint64_t &>(in)),
         "l"(reinterpret_cast<const uint64_t &>(scale)));
+#endif  // (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
 }
 
 __device__ __forceinline__ void mul_cvt_2x(fp8e4m3x2 &out, const bf16x2 &in, const floatx2 &scale) {
+#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
```
Oleg-Goncharov:
```suggestion
#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
```

transformer_engine/common/util/ptx.cuh
```diff
@@ -363,9 +504,11 @@ __device__ __forceinline__ void mul_cvt_2x(fp8e4m3x2 &out, const bf16x2 &in, con
       : "=h"(reinterpret_cast<uint16_t &>(out))
       : "r"(reinterpret_cast<const uint32_t &>(in)),
         "l"(reinterpret_cast<const uint64_t &>(scale)));
+#endif  // (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
 }
 
 __device__ __forceinline__ void mul_cvt_2x(fp8e5m2x2 &out, const bf16x2 &in, const floatx2 &scale) {
+#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
```
Oleg-Goncharov:
```suggestion
#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
```

transformer_engine/common/util/ptx.cuh
```diff
@@ -385,9 +528,11 @@ __device__ __forceinline__ void mul_cvt_2x(fp8e5m2x2 &out, const bf16x2 &in, con
       : "=h"(reinterpret_cast<uint16_t &>(out))
       : "r"(reinterpret_cast<const uint32_t &>(in)),
         "l"(reinterpret_cast<const uint64_t &>(scale)));
+#endif  // (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
 }
 
 __device__ __forceinline__ void mul_cvt_2x(fp8e4m3x2 &out, const fp16x2 &in, const floatx2 &scale) {
+#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
```
Oleg-Goncharov:
```suggestion
#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
```

transformer_engine/common/util/ptx.cuh
```diff
@@ -407,9 +552,11 @@ __device__ __forceinline__ void mul_cvt_2x(fp8e4m3x2 &out, const fp16x2 &in, con
       : "=h"(reinterpret_cast<uint16_t &>(out))
       : "r"(reinterpret_cast<const uint32_t &>(in)),
         "l"(reinterpret_cast<const uint64_t &>(scale)));
+#endif  // (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
 }
 
 __device__ __forceinline__ void mul_cvt_2x(fp8e5m2x2 &out, const fp16x2 &in, const floatx2 &scale) {
+#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
```
Oleg-Goncharov:
```suggestion
#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
```

Oleg-Goncharov (dismissed):
LGTM, just found a few places where ptx instructions require some higher arch generation  (although those are already guarded in kernels).

transformer_engine/common/util/ptx.cuh
```diff
@@ -19,43 +19,164 @@
...
+#define ARCH_HAS_STOCHASTIC_ROUNDING ARCH_SPECIFIC(Arch<100>, Arch<103>)
+
+namespace ptx {
 
 // https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#parallel-synchronization-and-communication-instructions-mbarrier-init
 __device__ __forceinline__ void mbarrier_init(uint64_t *mbar, const uint32_t count) {
+#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
   uint32_t mbar_ptr = __cvta_generic_to_shared(mbar);
   asm volatile("mbarrier.init.shared.b64 [%0], %1;" ::"r"(mbar_ptr), "r"(count) : "memory");
+#endif  // #if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
```
timmoon10, line 133: It would be safer for these device functions to raise errors rather than silently skipping.
```suggestion
#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
  uint32_t mbar_ptr = __cvta_generic_to_shared(mbar);
  asm volatile("mbarrier.init.shared.b64 [%0], %1;" ::"r"(mbar_ptr), "r"(count) : "memory");
#else
  NVTE_DEVICE_ERROR("mbarrier_init is only supported on SM 10.0+.");
#endif  // #if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
```

transformer_engine/common/util/nvfp4_transpose.cuh
```diff
@@ -30,7 +30,7 @@
 
 namespace transformer_engine {
 
-#if CUDA_VERSION > 12080
+#if CUDA_VERSION >= 12080
```
timmoon10: Should also update comment for `#endif`.

transformer_engine/common/util/nvfp4_transpose.cuh
```diff
@@ -1382,16 +1382,11 @@ __global__ void __launch_bounds__(THREADS_NUM)
...
-#ifndef USE_2D_NVFP4_KERNEL
-#define USE_2D_NVFP4_KERNEL 0
-#endif
-
 template <bool COMPUTE_ACTIVATIONS, typename ParamOP, float (*OP)(float, const ParamOP &),
           bool use_2d_quantization>
 void nvfp4_quantize_transpose(const Tensor &input, const Tensor *noop, Tensor *output,
                               const QuantizationConfig *quant_config, cudaStream_t stream) {
-#if CUDA_VERSION > 12080
+#if CUDA_VERSION >= 12080
```
timmoon10: Should also update comment for `#endif`.

transformer_engine/common/util/ptx.cuh
```diff
@@ -19,43 +19,164 @@
 #endif  // CUDA_VERSION >= 12080
 
 namespace transformer_engine {
-namespace ptx {
 
-#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
+template <int N>
+struct Arch {
```
timmoon10: It feels like we're trying to do metaprogramming with the `Arch` and `Family` classes. `Arch` feels like a C++11-style type trait class and `ArchHelper::compatible` feels like a C++20 concept. Not sure if we can use that connection to inform the design.

The biggest complaints I have are with naming and exposure:
- `Family` is exposed in the main `transformer_engine` namespace, but its name is not obvious. It would be better to have names like `DeviceArchFamily`/`CUDAArchFamily`/`SMArchFamily`. Alternatively, we could stash these classes within a more specific namespace.
- `Arch` is not obviously related to arch-specific compilation like `sm_100a`. If we wanted to handle logic to check for `sm_100`, it would be more natural for `Arch`/`DeviceArch` to refer to `sm_100` and `ArchSpecific`/`DeviceArchSpecific`/`DeviceSpecificArch` to refer to `sm_100a`. Seeing how many `__CUDA_ARCH__` macros we have in this header, this may be something we want to support soon.

transformer_engine/common/util/ptx.cuh
```diff
@@ -19,43 +19,164 @@
...
+  constexpr static int id = N * 10;
+  constexpr static bool is_arch = true;
+  constexpr static bool is_family = false;
+};
+
+template <int CurrentArch, int ArchSpecific, int FamilySpecific, int N>
+struct ArchHelper {
+  constexpr static bool compatible() {
```
timmoon10: It seems more natural to define `ArchHelper` as a function rather than as a class. Also, it may be worth putting this within a `detail` or anonymous namespace because it's highly entangled with the implementation of `is_supported_arch`.

transformer_engine/common/util/ptx.cuh
```diff
@@ -19,43 +19,164 @@
...
+                    "features. Please compile for smXXXf architecture instead of smXXX "
+                    "architecture.");
+      return true;
+    } else {
+      return false;
+    }
+  }
+};
+
+template <int Arch, int ArchSpecific, int FamilySpecific, class T, class... U>
```
timmoon10, line 63: This function is hard to use because the template arguments are entangled with the `ARCH_SPECIFIC` macro. Some of it is unavoidable because the current arch has to be deduced from macros, but it would be nice if the compatibility checking logic was standard C++. What if we separated the arch detection and compatiblity checking:
```
template <int Arch, int ArchSpecific, int ArchFamily>
struct CurrentDeviceArch {
constexpr static int arch_id= Arch;
constexpr static int arch_specific_id = ArchSpecific;
constexpr static int arch_family_id = ArchFamily;

template <class FirstArg, class... LastArgs>
constexpr bool any_compatible(const FirstArg &first, const LastArgs... &last) {
  // Check if first arg matches current arch
  if constexpr (first::is_arch_specific) {
    if constexpr (arch_specific_id == first::arch_specific_id) {
      return true;
    }
  } else if constexpr (first::is_arch_family) {
    if constexpr (arch_family_id / 100 == first::arch_family_id / 100) {
      return true;
    }
  } else {
    if constexpr (arch_id >= first::arch_id) {
      return true;
    }
  }

  // First arg doesn't match current arch, recurse to other args
  if constexpr (sizeof...(LastArgs) == 0) {
    return false;
  }
  return any_compatible(last...);
}

};

#define CURRENT_DEVICE_ARCH \
  CurrentDeviceArch<__CUDA_ARCH__, __CUDA_ARCH_SPECIFIC__, __CUDA_ARCH_FAMILY_SPECIFIC__>
```
We could use this like:
```
__device__ void stochastic_rounding_func() {
  constexpr auto current_arch = CURRENT_DEVICE_ARCH();
  if constexpr (current_arch.any_compatible(STOCHASTIC_ROUNDING_ARCHS) {
    ...
  } else {
    ...
  }
}

__device__ void blackwell_func() {
  constexpr auto current_arch = CURRENT_DEVICE_ARCH();
  if constexpr (current_arch.any_compatible(DeviceArch<100>()) {
    ...
  } else {
    ...
  }
}
```
Let me work this out in more detail.

transformer_engine/common/CMakeLists.txt
```diff
@@ -139,12 +186,58 @@ list(APPEND transformer_engine_SOURCES
      recipe/delayed_scaling.cu
      recipe/fp8_block_scaling.cu
      recipe/nvfp4.cu
+     comm_gemm_overlap/userbuffers/userbuffers.cu)
+
+list(APPEND transformer_engine_cuda_arch_specific_sources
+     gemm/cutlass_grouped_gemm.cu
```
timmoon10, line 192: The CUTLASS kernel is confusing our logic. It requires arch-specific compilation, but only for Hopper. Moving it to `transformer_engine_cuda_sources` would avoid some unnecessary compilation.

transformer_engine/common/util/ptx.cuh
```diff
@@ -121,107 +248,125 @@ __device__ __forceinline__ float exp2f(e8m0_t biased_exp) {
...
-    return 0x00;
-  }
-  uint32_t val_u32 = *reinterpret_cast<uint32_t *>(&val);
-  e8m0_t exponent = (val_u32 >> FP32_MANTISSA_BITS);
-  uint32_t mantissa = val_u32 & 0x7FFFFF;
-  // Round up exponent and deal with satfinite.
-  if ((mantissa > 0 && exponent != 0xFE) && !(exponent == 0 && mantissa <= 0x400000)) {
-    ++exponent;
+  constexpr bool is_blackwell = ARCH_BLACKWELL_FAMILY;
+  if constexpr (is_blackwell) {
```
Oleg-Goncharov, line 261: Sanity check on the MXFP8 convert path.
Would `cvt.rp.satfinite.ue8m0x2.f32` emit a hardware opcode on Blackwell when building with `CUDA <12.9` for `sm_100a`?
(PTX ISA lists this op as supported on arch-specific target `sm_100a` from `CUDA >=12.7`; and on the family-specific targets `sm_100f` from `CUDA >=12.9`.)
With the new `ARCH_BLACKWELL_FAMILY / is_blackwell constexpr`, will the HW path (not the emulated fallback) be taken with toolkits `<12.9` and `sm_100a`.

transformer_engine/common/CMakeLists.txt
ptrendx: My understanding is that we would want to compile it also for Blackwell specific - there are otherwise some warnings it produces about possible performance issues.

transformer_engine/common/util/ptx.cuh
ptrendx: That should be the case  - for CUDA < 12.9 we set the proper macros ourselves.

transformer_engine/common/util/nvfp4_transpose.cuh
ptrendx: Will change to FP4_TYPE_SUPPORTED macro that is already defined in common.h.

transformer_engine/common/CMakeLists.txt
timmoon10: At the moment we only run the kernel on Hopper, although I'm not aware of anything preventing us from using it on Blackwell:
https://github.com/NVIDIA/TransformerEngine/blob/e90582f2010deae477a71bad0aacf278dd5abfa4/transformer_engine/common/gemm/cublaslt_gemm.cu#L1044-L1048

transformer_engine/common/util/ptx.cuh
ptrendx: I think that moving those classes to the ptx namespace is better than changing the names.
As for the general architecture support and not just the a or f variant - not sure to be honest. I mean, sure, having it all in one system could be ok, but not sure we really need it since the generic features are forward compatible so the maintenance burden is not big. This pr was needed to simplify the management of the a and f specific features since those require a lot more attention.

transformer_engine/common/util/ptx.cuh
ptrendx: I will think about it. As I was writing it I thought that this is for some reason necessary, but now I do not quite see why. Maybe it was an artifact of the code at the time?

transformer_engine/common/util/ptx.cuh
ptrendx: Not sure I understand the "hard to use" argument and I would not complicate things to be honest - could you elaborate? This function is not something you would use directly but via ARCH_SPECIFIC macro, which seems pretty easy to use.

transformer_engine/common/util/ptx.cuh
ptrendx: Sure, will add that.

transformer_engine/common/util/ptx.cuh
timmoon10: Here's a working prototype: https://github.com/timmoon10/TransformerEngine/commit/7116157c4933f87bf047a6dcf6bb236a2d8c678b
It's a little overengineered, but it's easier to specify archs without creating a new macro:
```c++
__device__ void my_new_kernel() {
  using CurrentCUDAArch = NVTE_CURRENT_CUDA_ARCH;
  if constexpr (CurrentCUDAArch::any_compatible<cuda_arch::ArchSpecific<103>, cuda_arch::ArchFamily<130>>()) {
    // fancy new feature that is supported in 103a and 130f
  }
}
```

transformer_engine/common/util/ptx.cuh
timmoon10: This prototype handles general archs: https://github.com/timmoon10/TransformerEngine/commit/7116157c4933f87bf047a6dcf6bb236a2d8c678b

transformer_engine/common/util/ptx.cuh
ptrendx: You can do the same in the current code by just using ARCH_SPECIFIC(Arch<100>) for example.

timmoon10: If we treat this as an internal implementation for `NVTE_CUDA_ARCHS_MATCHES`, then it would be nicer to discourage direct calls by putting it within a namespace or by naming with a leading underscore or something. That said, moving it out of the main `transformer_engine` namespace and into `transformer_engine::ptx` does address the bulk of my uneasiness.

timmoon10 (approved):
Overall LGTM.

I'm still a little uneasy that we are defining objects that can't really stand alone, but are entangled with macros. For example, `ArchSpecific::compatible` and `is_supported_arch` are templated on the current CUDA arch, so we need a macro to resolve `__CUDA_ARCH__`/`__CUDA_ARCH_SPECIFIC__`/`__CUDA_ARCH_FAMILY_SPECIFIC__` in order to call them. In effect, the macros are the real interfaces and these functions are internal implementation, which doesn't feel like great practice. My prototype at https://github.com/timmoon10/TransformerEngine/commit/7116157c4933f87bf047a6dcf6bb236a2d8c678b attempted to concentrate the macro-dependent logic to a single point and to use normal code for everything else, but it didn't really make usage any cleaner and it added template metaprogramming messiness. Ultimately, I think the existing logic is straightforward enough and we should be careful not to overengineer.

PR #2272, opened 2025-10-15
[common] Misc improvements for attention

cyanguwa:
# Description

This PR:
- removes `nvte_fused_attn_fwd_qkvpacked`, `nvte_fused_attn_bwd_qkvpacked`, `nvte_fused_attn_fwd_kvpacked`, `nvte_fused_attn_bwd_kvpacked` APIs, and leaves `nvte_fused_attn_fwd` and `nvte_fused_attn_bwd` for uniformity and easier maintainance,
- improves error messaging in `nvte_get_fused_attn_backend` (ultimately the backend selection logic in this function will be replaced by cudnn-frontend's `check_support()` when it's ready).

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

See description.

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2287, opened 2025-10-20
[common] Remove kvpacked and qkvpacked attention functions for every kernel type.

pggPL:
# Description

There are 3 variants of fused_attention functions: for separate QKV, KV packed and QKV packed, which differ only by pointers to qkv. This results in code duplication for each type of the fused attention kernel: arbitrary seqlen, max 512 and fp8. This PR deduplicates the code and moves pointer computation one abstraction layer - from the functions like `fused_attn_max_512_fwd_qkvpacked` into the functions like `nvte_fused_attn_fwd_qkvpacked` in common c++ api.

These packed versions of common attention api functions are used by JAX, so I think running JAX CI is good test of that changes. PyTorch uses only non-packed function.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:


cyanguwa:
I think this is similar to #2272 :) Yes, Jax needs a bit of fixing in order to get its attention working.

cyanguwa:
Could you add the deprecation note for these qkvpacked/kvpacked APIs as we discussed offline please? Thanks.

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -782,10 +972,30 @@ void nvte_fused_attn_bwd_kvpacked(
   if (fused_attention_backend == NVTE_Fused_Attn_Backend::NVTE_F16_max512_seqlen) {
 #if (CUDNN_VERSION >= 8901)
     Tensor *output_S = convertNVTETensorCheck(Aux_CTX_Tensors->tensors[0]);
-    fused_attn_max_512_bwd_kvpacked(
-        b, h_q, max_seqlen_q, max_seqlen_kv, d, attn_scale, dropout, qkv_layout, bias_type,
-        attn_mask_type, input_Q, input_KV, input_dO, output_S, output_dQ, output_dKV, output_dBias,
-        input_cu_seqlens_q, input_cu_seqlens_kv, wkspace, stream, handle);
+
+    // Unpack KV and dKV and call the non-packed function
+    size_t stride = 2 * h_q * d;
```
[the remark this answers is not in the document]
pggPL: That's good point - this is some leftover from my previous change. Weird that tests didn't catch it. Fixed.

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -216,6 +216,9 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
     int64_t window_size_right, bool return_max_logit);
 
 /*! \brief Compute dot product attention with packed QKV input.
+ *
+ * \warning This API is **deprecated**.
```
cyanguwa: Would `\deprecated` be a better marking for this message in the document?

Could we add something like this to give user warnings at compile time as well?
```
[[deprecated("nvte_fused_attn_fwd_qkvpacked() is deprecated. Please use nvte_fused_attn_fwd() instead.")]]
void nvte_fused_attn_fwd_qkvpacked() {
}
```

cyanguwa (dismissed):

transformer_engine/common/include/transformer_engine/fused_attn.h
pggPL: Done, renders correctly and docs and compiles locally with warnings/

cyanguwa (approved):

PR #2294, opened 2025-10-22
[JAX] Make SR rng state always 2D (num_devices, 4) to fix partitioning issue

jberchtold-nvidia:
# Description

Fixes partitioning issue where a 2D sharding constraint was applied when a JAX Mesh was available, but the Mesh only container a single-GPU. Now even if we are on a single-GPU or without a Mesh, the SR rng state will always be 2D `(num_devices, 4)`

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- In the case of a JAX Mesh being present but only containing a single-GPU, we now make the shape `(1, 4)` instead of `(4,)` so partitioning constraints don't fail.

# Checklist:

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,1 @@
+recursive-include transformer_engine/common/include *.*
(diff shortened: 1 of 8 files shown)


phu0ngng (dismissed):
LGTM. Thanks

phu0ngng (dismissed):

PR #2301, opened 2025-10-24
Docs fix

pggPL:
# Description

Our documentation returned a lot of warnings and it seems that some of them were rational. It turned out that half of our PyTorch API was not rendered. This PR fixes all the warnings and forces Github workflow to error out if any docs warning will appear.

The most problematic error was related to cyclic imports - this resulted in part of our PyTorch API not being rendered. Other were mostly related to wrong formatting and some warnings didn't result in anything wrong.

Our docs for 2.9 contains only part of PyTorch API exposed in 2.8, so this will need urgent fix I will update in separate PR.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:

- [ x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)


docs/api/pytorch.rst
```diff
@@ -37,16 +37,22 @@ pyTorch
 .. autoapiclass:: transformer_engine.pytorch.CudaRNGStatesTracker()
   :members: reset, get_states, set_states, add, fork
 
-.. autoapifunction:: transformer_engine.pytorch.fp8_autocast
-
-.. autoapifunction:: transformer_engine.pytorch.fp8_model_init
```
ptrendx, line 40: We do want to have them still in the documentation though, just marked as deprecated.

docs/api/pytorch.rst
```diff
@@ -37,16 +37,22 @@ pyTorch
...
 .. autoapifunction:: transformer_engine.pytorch.checkpoint
 
+
+.. autoapifunction:: transformer_engine.pytorch.make_graphed_callables
+
+.. autoapifunction:: transformer_engine.pytorch.get_cpu_offload_context
+
+
+Recipe availability
+------------------------
```
ptrendx: In the past I had some trouble with the titles where the underline was not exact same length as the text.

docs/api/pytorch.rst
```diff
@@ -79,9 +84,22 @@ pyTorch
 
 .. autoapifunction:: transformer_engine.pytorch.moe_sort_chunks_by_index_with_probs
 
+
+GEMM Comm overlap
```
ptrendx: Maybe "Communication - computation overlap"?

docs/api/pytorch.rst
ptrendx: Ok, I see that you just moved it.

docs/examples/attention/attention.ipynb
```diff
@@ -16,8 +16,8 @@
     "\n",
     "[Transformer Engine](https://github.com/NVIDIA/TransformerEngine.git) supports the calculation of dot product attention in two frameworks, [PyTorch](https://github.com/pytorch/pytorch) and [JAX](https://github.com/google/jax). The API for each framework is\n",
     "\n",
-    "- [transformer_engine.pytorch.DotProductAttention](../../api/pytorch.rst#transformer_engine.pytorch.DotProductAttention)\n",
-    "- [transformer_engine.jax.flax.DotProductAttention](../../api/jax.rst#transformer_engine.jax.flax.DotProductAttention)"
+    "- [transformer_engine.pytorch.DotProductAttention](../../api/pytorch.rst#transformer_engine.pytorch.dotproductattention)\n",
+    "- [transformer_engine.jax.flax.DotProductAttention](../../api/jax.rst#transformer_engine.jax.flax.dotproductattention)"
```
ptrendx: Are you sure about that? Just looked at the current docs page and the link works properly there.

transformer_engine/pytorch/cross_entropy.py
```diff
@@ -87,4 +89,55 @@ def backward(ctx, grad_output):
...
+    FP32 only. The returned loss is always in FP32, the input gradients are upcasted
+    to the datatype of the input.
+
+    If ``dist_process_group`` is passed for distributed loss calculation, the input to each
+    distributed rank should be ``(*, V/world_size)``. Note that each of the ranks should
+    get equal shards along the V dimension.
+
+    Parameters
+    ----------
+    _input : torch.Tensor
```
ptrendx: Why _input and not just input?

docs/examples/attention/attention.ipynb
pggPL: Oh, some debugging leftover. It does also work, but I will remove this change.

transformer_engine/pytorch/cross_entropy.py
pggPL: I also noticed this. This function already uses _input as an argument name, so I documented what is in the code rather than changing that - it would be breaking change. This function does not render now due to error btw.

docs/api/pytorch.rst
pggPL: If the line is shorter, then warning is generated. In that case there is no warning and it renders properly, so I guess it's ok now.

transformer_engine/pytorch/cross_entropy.py
pggPL: I put _input into kwargs and changed it into input in args. I removed _input entirely from the docs, because I doubt anyone uses it. In MCore they use only non-keyword arguments.

docs/api/jax.rst
```diff
@@ -4,10 +4,10 @@
     See LICENSE for license information.
 
 Jax
-=======
+===
 
 Pre-defined Variable of Logical Axes
-------------------------------------
+-------------------------------------
```
ksivaman: Was this needed to fix the rendering? The earlier version seems fine and of the correct length

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -245,14 +245,17 @@ class DotProductAttention(TransformerEngineBaseModule):
...
-                 'learnable': S[:,j,:,i] = exp(S[:,j,:,i])/(exp(alpha[j]) + sum(exp(S[:,j,:,:]), dim=-1)),
-                 where alpha is a learnable parameter in shape [h].
+
+                 For a given attention score ``S = Q x K^T``, of shape ``[b, h, s_q, s_kv]``:
+
+                 * 'vanilla': ``S[:,:,:,i] = exp(S[:,:,:,i])/sum(exp(S[:,:,:,:]), dim=-1)``
+                 * 'off-by-one': ``S[:,:,:,i] = exp(S[:,:,:,i])/(1 + sum(exp(S[:,:,:,:]), dim=-1))``
+                 * 'learnable': ``S[:,j,:,i] = exp(S[:,j,:,i])/(exp(alpha[j]) + sum(exp(S[:,j,:,:]), dim=-1))``
+
+                 where ``alpha`` is a learnable parameter in shape ``[h]``.
```
ksivaman:
```suggestion
                 where ``alpha`` is a learnable parameter of shape ``[h]``.
```

transformer_engine/pytorch/__init__.py
```diff
@@ -14,13 +14,8 @@
...
 
-@functools.lru_cache(maxsize=None)
-def torch_version() -> tuple[int, ...]:
-    """Get PyTorch version"""
-    return PkgVersion(str(torch.__version__)).release
-
-
-assert torch_version() >= (2, 1), f"Minimum torch version 2.1 required. Found {torch_version()}."
+torch_version = PkgVersion(str(torch.__version__)).release
+assert torch_version >= (2, 1), f"Minimum torch version 2.1 required. Found {torch_version}."
```
ksivaman: Please change the instances at the call site as well. For e.g. `linear.py` still calls `torch_version()`

transformer_engine/pytorch/utils.py
```diff
@@ -10,15 +10,21 @@
...
 from ..debug.pytorch.debug_quantization import DebugQuantizedTensor
 
 
 __all__ = ["get_device_compute_capability", "get_cudnn_version", "is_bf16_available"]
 
 
+@functools.lru_cache(maxsize=None)
+def torch_version() -> tuple[int, ...]:
+    """Get PyTorch version"""
+    return PkgVersion(str(torch.__version__)).release
```
ksivaman: Is this redefined here because of cyclic import issues when import directly from `__init__.py`? We should have it defined only in one location.

ksivaman: If the __init__ one is causing issues with import, please remove it and change all instance to import from utils instead

transformer_engine/pytorch/cross_entropy.py
```diff
@@ -87,4 +90,65 @@ def backward(ctx, grad_output):
...
+    _input: Optional[torch.Tensor] = None,
+) -> torch.Tensor:
+    """
+    Cross Entropy loss with optional distributed reduction.
+
+    The input tensor can be in BF16/FP32, the loss and gradient calculation happens in
+    FP32 only. The returned loss is always in FP32, the input gradients are upcasted
+    to the datatype of the input.
+
+    If ``dist_process_group`` is passed for distributed loss calculation, the input to each
```
ksivaman, line 109: There are multiple instances where we use a single backtick in the docstrings instead of double backticks, this PR seems like a good place to change all those instances as well.

transformer_engine/pytorch/cross_entropy.py
```diff
@@ -23,7 +26,7 @@ class CrossEntropyFunction(torch.autograd.Function):
     @staticmethod
     def forward(
         ctx,
-        _input,
+        input,
```
ksivaman: I saw @ptrendx's [comment](https://github.com/NVIDIA/TransformerEngine/pull/2301#discussion_r2467462025) above suggesting we should change it, but `input` is also a python keyword and we should avoid using it. Maybe we can go with `inp` as we do in our other modules. I'm surprised that the linter didn't complain about this change

docs/api/jax.rst
pggPL: Yeah, that's incorrect. Fixed.

transformer_engine/pytorch/cross_entropy.py
ptrendx: That's a good point actually.

transformer_engine/pytorch/utils.py
pggPL: Yes, that is the case. And I cannot import utils before dynamically loading c++ lib, because it imports some other torch classes. So I guess there are 2 ways of doing that - create some file torch_version.py with only that or do what I did. I do not want to add file for one line function, what do you think?

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
pggPL: i changed all occurrences in TE (around 10)

transformer_engine/pytorch/cross_entropy.py
pggPL: Changed to inp

transformer_engine/pytorch/cross_entropy.py
pggPL: Ok, I did it and it took a lot of time and a lot of files were changed. Also, I discovered more formatting errors, so this PR is not that small now.

transformer_engine/pytorch/utils.py
timmoon10: I think creating a standalone `torch_version.py` file would be the most straightforward way to untangle these issues with import ordering.

An alternative would be changing `__init__.py` to manually import in the right order:
```python
from transformer_engine.common import load_framework_extension
load_framework_extension("torch")

from transformer_engine.pytorch.utils import torch_version
```
However, this is not ideal because we don't check the PyTorch version before loading `transformer_engine_torch`, so we might get linker errors instead of a good error message.

ksivaman: I think we can create a `torch_version.py` even if it is a single function @pggPL

ksivaman (dismissed):
LGTM, will check artifacts after CI

ksivaman (dismissed):

transformer_engine/pytorch/cpu_offload.py
```diff
@@ -657,60 +657,64 @@ def get_cpu_offload_context(
...
             If True, the synchronization is done manually by the user.
             Additional argument manual_controller is returned. See more in manual control section.
-    offload_stream: torch.cuda.Stream, default = `None`
-            If provided, the offload stream is used for offloading and reloading.
-            Otherwise, a new stream is allocated internally. It can be other than None
-            only if manual_synchronization is True.
+    offload_stream: torch.cuda.Stream, default = None
+        If provided, the offload stream is used for offloading and reloading.
+        Otherwise, a new stream is allocated internally. It can be other than None
+        only if manual_synchronization is True.
```
ptrendx: Why change in the indentation?

transformer_engine/pytorch/distributed.py
```diff
@@ -812,7 +812,9 @@ def set_states(self, states: Dict[str, torch.Tensor]) -> None:
         Set the rng states. For efficiency purposes, we do not
         check the size of seed for compatibility.
 
-        states: Dict[str, torch.Tensor]
+        Parameters
+        ----------
+        states : Dict[str, torch.Tensor]
```
ptrendx, line 817: There are other cases where there is no space between the parameter and the ':' - we should use one consistently (preferably the right one :-) ).

transformer_engine/pytorch/__init__.py
```diff
@@ -7,22 +7,14 @@
...
 
 from transformer_engine.common import load_framework_extension
-
-
-@functools.lru_cache(maxsize=None)
-def torch_version() -> tuple[int, ...]:
-    """Get PyTorch version"""
-    return PkgVersion(str(torch.__version__)).release
-
+from transformer_engine.pytorch.torch_version import torch_version
```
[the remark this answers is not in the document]
ksivaman: CC @pggPL

transformer_engine/pytorch/distributed.py
pggPL: Zrzut ekranu 2025-11-26 o 15 06 14

official numpy docs guide have space, so i changed this to have space everywhere

transformer_engine/pytorch/cpu_offload.py
pggPL: fixed

transformer_engine/pytorch/__init__.py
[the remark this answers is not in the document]
pggPL: It is outdated, I haven't commited the torch_version.py at first, but then I fixed the issue.

ksivaman (approved):

PR #2307, opened 2025-10-27
[JAX] Fix mesh resource requirement when no mesh

jberchtold-nvidia:
# Description

When not using Flax logical axes, a global MeshResource context was required by TE even when no JAX mesh was active. This PR fixes this case and avoids raising an unnecessary error.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Update TE's logical axis system to return an empty mapping and avoid raising an error when no JAX mesh is active. Use case, new user to TE trying it out on single-GPU. Want to avoid unnecessary complexity in cases where it isn't required

# Checklist:


transformer_engine/jax/sharding.py
```diff
@@ -80,6 +80,11 @@ def get_sharding_map_logic_axis_to_mesh_axis():
         # If no mesh is defined, return an empty dict and do not require a MeshResource context to be present
         return {}
 
+    abstract_mesh = get_abstract_mesh()
+    if abstract_mesh is None or sorted(abstract_mesh.manual_axes) == sorted(mesh.axis_names):
```
phu0ngng: Hi, why do we need to check for `if abstract_mesh is None` here?

jberchtold-nvidia: I wasn't exactly sure how the abstract mesh relates to the PXLA physical mesh above. If they both are always set via `with jax.sharding.Mesh():`, then this `abstract_mesh is None` check is redundant.

phu0ngng: Looks like the `get_abstract_mesh()` won't return None.
https://github.com/jax-ml/jax/blob/eace2086ab710c2b95b40e926a44212d73cfd7ae/jax/_src/mesh.py#L625

jberchtold-nvidia: Thanks @phu0ngng! I've removed the `abstract_mesh is None` check in the latest version

PR #2308, opened 2025-10-27
Change the pyTorch installation to CUDA 13 in Build All GitHub action 

ptrendx:
# Description

With this change it will match the version in the JAX container

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -83,7 +83,9 @@ jobs:
       options: --user root
     steps:
       - name: 'Dependencies'
-        run: pip install torch pybind11[global] einops onnxscript
+        run: |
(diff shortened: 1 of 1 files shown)


ptrendx:
This PR does not actually do what is claimed (and that will only be possible once we can compile TE with nvcc from pip, so that we could start from a bare container and install JAX, pyTorch and then TE with just pip dependencies, cc @ksivaman), since the jax container has CUDA installed system-wide and pytorch does not want to use that. But somehow it does resolve the original problem of the build all github action failing with disk space issue, so I still think it is worth to merge it.
Tagging @jberchtold-nvidia

PR #2310, opened 2025-10-28
[JAX] Quickstart documentation

tdophung:
# Description

This change adds the new quick start notebook for Jax, mirroring the same Transformer architecture as the PyTorch guide, for users familiar with the PyTorch guide can also easily follow. Contains an 4 iterations of the layer with different training step time durations: 
1.  pure JAX/Flax implemtation
2. Basic TE implementation without any fused layer
3. TE implementation with mixed fused and unfused layers
4. Full TransformerLayer from TE

This might also include changes to how sphinx display this content on the HTML docs page in later commits

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Added 1 quickstart guide written in Jax

# Checklist:

- [ DOES NOT APPLY] I have added tests that prove my fix is effective or that my feature works
- [ DOES NOT APPLY] New and existing unit tests pass locally with my changes


docs/examples/quickstart_jax_utils.py
```diff
@@ -0,0 +1,537 @@
...
+import transformer_engine.jax as te
+import transformer_engine.jax.flax as te_flax
+from transformer_engine.common import recipe
+from transformer_engine.jax.sharding import global_shard_guard, MeshResource
+
+
+@lru_cache
+def is_bf16_supported():
+    """Return if BF16 has hardware supported"""
+    gpu_arch = get_device_compute_capability(0)
```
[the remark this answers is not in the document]
tdophung: I have already resolved this in latest commits

docs/examples/quickstart_jax_utils.py
```diff
@@ -0,0 +1,433 @@
...
+        x = nn.Dense(features=self.hidden_size, use_bias=True)(x)
+        return x
+
+
+def share_parameters_with_basic_te_model(basic_model_params, te_params_template):
+    """Initialize parameters for a TE Transformer layer using basic JAX modules.
+    Parameter values are copied from another JAX-based model.
+    """
+
+    new_te_params = te_params_template.copy()
```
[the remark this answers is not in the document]
tdophung: this should not matter much since we only use the new_te_params anyways in the notebook. But for a more easy to understand read of the code, deepcopy works better since it highlights the intention

docs/examples/quickstart_jax_utils.py
```diff
@@ -0,0 +1,439 @@
...
+    def train_step_fn(variables: Any, inp: jnp.ndarray, grad_target: jnp.ndarray, dropout_key):
+        """Compute forward pass and VJP in one step"""
+
+        # Define forward function that closes over grad_target and dropout_key
+        def forward_fn(variables: Any, inp: jnp.ndarray):
+            """Pure forward function for VJP computation"""
+            rngs = {"dropout": dropout_key}
+            with te.fp8_autocast(**fp8_autocast_kwargs):
+                call_kwargs = {**forward_kwargs, "rngs": rngs}
+                return model_apply_fn(variables, inp, **call_kwargs)
```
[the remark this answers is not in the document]
tdophung: this is just the comment being confusing, but the functionality should be correct. Will change the comment

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,830 @@
...
+    "        \n",
+    "        # Attention self-implemented. Comment out if not used\n",
+    "        attention = utils.DotProductAttention(\n",
+    "            num_attention_heads=self.num_attention_heads,\n",
+    "            kv_channels=self.kv_channels,\n",
+    "            attention_dropout=self.attention_dropout,\n",
+    "        )\n",
+    "        x = attention(q, k, v, attention_mask, deterministic=deterministic)\n",
+    "        \n",
+    "        # Attention built-in. Comment out if not used\n",
```
jberchtold-nvidia: Can we add a bit more detail to this comment, something like: If you would like to use Flax's built-in attention implementation, you can use this instead

Want to clarify to the users' that this is an alternative option and not some debug code we left by mistake

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,830 @@
...
+    "te_params = basic_te_transformer.init(key, x, attention_mask=None, deterministic=False)\n",
+    "\n",
+    "# Extract the 'params' pytrees\n",
+    "basic_params = params['params']\n",
+    "te_params_template = te_params['params']\n",
+    "\n",
+    "print(f\"Basic TE parameter shapes: {jax.tree_util.tree_map(lambda x: x.shape, te_params_template)}\")\n",
+    "\n",
+    "shared_te_params = {}\n",
+    "shared_te_params['params'] = utils.share_parameters_with_basic_te_model(basic_params, te_params_template)\n",
```
jberchtold-nvidia: This causes a KeyError for me
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[15], [line 19](vscode-notebook-cell:?execution_count=15&line=19)
     16 print(f"Basic TE parameter shapes: {jax.tree_util.tree_map(lambda x: x.shape, te_params_template)}")
     18 shared_te_params = {}
---> [19](vscode-notebook-cell:?execution_count=15&line=19) shared_te_params['params'] = utils.share_parameters_with_basic_te_model(basic_params, te_params_template)
     21 # Test forward pass
     22 y = basic_te_transformer.apply(te_params, x, attention_mask=None, deterministic=True)

File transformerengine/docs/examples/quickstart_jax_utils.py:325, in share_parameters_with_basic_te_model(basic_model_params, te_params_template)
    323 new_te_params = copy.deepcopy(te_params_template)
    324 # Layer Norms
--> 325 new_te_params["LayerNormDenseGeneral_0"]["scale"] = basic_model_params["LayerNorm_0"]["scale"]
    326 new_te_params["LayerNormDenseGeneral_0"]["ln_bias"] = basic_model_params["LayerNorm_0"]["bias"]
    328 # QKV and Projection

KeyError: 'LayerNormDenseGeneral_0'
```

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,830 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Mean time: 18.0991792678833 ms\n"
+     ]
+    }
+   ],
+   "source": [
+    "shared_fused_te_params = {}\n",
+    "shared_fused_te_params['params'] = utils.share_fused_parameters_with_basic_te_model(basic_params, fused_te_params_template)\n",
```
jberchtold-nvidia: This gives me
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[13], [line 2](vscode-notebook-cell:?execution_count=13&line=2)
      1 shared_fused_te_params = {}
----> [2](vscode-notebook-cell:?execution_count=13&line=2) shared_fused_te_params['params'] = utils.share_fused_parameters_with_basic_te_model(basic_params, fused_te_params_template)
      4 # Test forward pass
      5 y = fused_te_transformer.apply(fused_te_params, x, attention_mask=None, deterministic=True)

AttributeError: module 'quickstart_jax_utils' has no attribute 'share_fused_parameters_with_basic_te_model'
```

jberchtold-nvidia: Figured it out. There are mismatching ''' and """ quotes in the docstrings in the utils file so some methods are basically commented out

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,830 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Mean time: 7.96757698059082 ms\n"
+     ]
+    }
+   ],
+   "source": [
+    "utils.speedometer(\n",
```
jberchtold-nvidia, line 749: I think it'd be slightly clearer to the user to just wrap this speedometer call in `with te.fp8_autocast` and remove the inner `fp8_autocast`s inside the utils file. 

However, I briefly tried this and I got a runtime of 12ms instead of the 8ms I was getting originally, so there may be something interfering with the context here.

Can you try to see if there's just something small I missed in my attempt? Thanks!
```
with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):
    utils.speedometer(
        model_apply_fn=te_transformer.apply,
        model_init_fn=te_transformer.init,
        variables=shared_te_transformer_params,  # Includes both params and fp8_metas
        input=x,
        output_grad=dy,
        dropout_key=dropout_key,
        forward_kwargs={"attention_mask": None, "deterministic": False},
    )
```

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,830 @@
...
+    "\n",
+    "with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):\n",
+    "    te_transformer_params = te_transformer.init(key, x, deterministic=False)\n",
+    "    \n",
+    "    # When using FP8, we need to preserve the fp8_metas collection\n",
+    "    # that was created during initialization within the fp8_autocast context.\n",
+    "    # Only the 'params' are shared from basic_params, but fp8_metas must come from\n",
+    "    # the FP8-initialized model.\n",
+    "    shared_te_transformer_params = {}\n",
+    "    shared_te_transformer_params['params'] = utils.share_parameters_with_transformerlayer_te_model(basic_params, te_transformer_params_template)\n",
```
jberchtold-nvidia: Does the PyTorch getting started give any reasoning for why they want to share params here? Is it just to convince users it is functionally equivalent and uses the same number of params?

I feel like with the additional params state tracking in JAX to integrate `fp8_metas`, this complicates the getting started and in my opinion I don't think we need to include this param sharing. Param sharing would be a feature for advanced users and they may prefer to use our TE/JAX functional API instead of sharing params between two layer implementations.

docs/examples/quickstart_jax.ipynb
tdophung: There is this pre-commit-ci bot thing that keep pushing commits to change my ''' to """ and it messed up all the long comments I guarded with ''' as it only changes 1 end of the comment and leave the other unchanged. leading to the part you are executing is part of the NEXT function and not the current function where it was supposed to be. I'll fix it in next commit

docs/examples/quickstart_jax.ipynb
tdophung: It does not explicitly give any reasons into why params are shared. I assumed the same as you, to convince users that it is functionally equivalent. However maybe @ptrendx or @timmoon10  or @ksivaman can shed some light here as one of them wrote it (I saw all 3 committing changes to the utils file).

Maybe we can optionally share parameters by having it in a separate code cell and tell users that they can go read it in utils and don't bother with it for the quickest start?

jberchtold-nvidia: Sounds good, if Przemek, Tim, or Kirthi don't have any strong motivation for it on the JAX side besides convincing users, I would say we leave out the weight sharing entirely for now. I'm not sure if this form of weight sharing is good practice in JAX as when we scale to automatic sharding this becomes difficult to maintain to ensure both models use the same sharding. It would be easier in JAX to either have the TE model allocate it's own weights or call TE/JAX as functions, the latter of which is what we do to integrate with an existing LLM framework called MaxText.

tdophung: Seems like Kirthi also answered on slack that it is okay to leave out the weight copying so I am removing it as you suggested

docs/examples/quickstart_jax.ipynb
tdophung: it might have been that the since jax.jit() only look at the scope of the function it is  tracing, it does not see the outer "with te.fp8_autocast" that we wrap speedometer with, so it compiled a version of the loss_fn that does not have fp8 enabled, hence falling back to bf16 and hence 12ms

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+   "execution_count": 27,
+   "id": "a4d1cfdc",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "class BasicTransformerLayer(nn.Module):\n",
+    "    \"\"\"Basic Transformer layer using plain JAX/Flax modules\n",
+    "    \n",
+    "    This is the JAX/Flax equivalent of the PyTorch BasicTransformerLayer\n",
+    "    from the quickstart.ipynb notebook.\n",
```
phu0ngng: Not sure if we should emphasize PyTorch everywhere in this tutorial.

This tutorial should be for JAX users who want to experience TE/JAX, not PyTorch users who want to try TE/JAX as there is no reason for them to do so.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    "\n",
+    "<div class=\"alert alert-info\">\n",
+    "\n",
+    "<b>Summary</b>\n",
+    "    \n",
+    "We build a basic Transformer layer using regular JAX modules. This will be our baseline for later comparisons with Transformer Engine.\n",
+    "\n",
+    "</div>\n",
+    "\n",
+    "Let's start with creating the transformer layer using plain JAX/Flax. Figure 1 shows the overall structure.\n",
```
phu0ngng: `JAX/FLAX` to `FLAX Linen`. You can link to this site if needed https://flax.readthedocs.io/en/stable/.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    "import transformer_engine.jax.flax as te_flax\n",
+    "from transformer_engine.jax.quantize import is_fp8_available, ScalingMode"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f28cb444",
+   "metadata": {},
+   "source": [
+    "TE provides a set of JAX modules that can be used to build Transformer layers. The simplest of the provided modules are the `Linear ` and `LayerNorm` layers, which we can use instead of `flax.linen.Linear` and ` flax.linen.LayerNorm`. Let's modify our `BasicTransformLayer`:"
```
phu0ngng: Flax Linen modules.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+   "cell_type": "code",
+   "execution_count": 37,
+   "id": "6b0c705e",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Mean time: 18.087706565856934 ms\n"
```
phu0ngng: Looks like this is slower than the Unfused one. Could you have a look at which part is the slower part?

Besides, this `FusedTETransformerLayer` is also slower than the `te_flax.TransformerLayer` while they are using the same modules. Could you have a look?

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    "\n",
+    "<div class=\"alert alert-info\">\n",
+    "\n",
+    "<b>Summary</b>\n",
+    "    \n",
+    "We configure a TE module to perform compute in FP8.\n",
+    "\n",
+    "</div>\n",
+    "\n",
+    "Enabling FP8 support is very simple in Transformer Engine. We just need to wrap the modules within an [fp8_autocast](.../api/jax.rst#transformer_engine.jax.fp8_autocast) context manager. Note that fp8_autocast should only be used to wrap the forward pass and must exit before starting a backward pass. See the [FP8 tutorial](fp8_primer.ipynb) (currently only available in PyTorch) for a detailed explanation of FP8 recipes and the supported options.\n",
```
phu0ngng: I think we just need to rename `fp8_autocast` to `autocast` https://github.com/NVIDIA/TransformerEngine/pull/2256.

> Note that fp8_autocast should only be used to wrap the forward pass and must exit before starting a backward pass.

I suspect this is not true in TE-JAX.

ptrendx:
There seems to be some issue with the commits mixed up. Could you rebase on top of the current main @tdophung ?

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    "\n",
+    "This is specifically a guide for getting started to use Transformer Engine with JAX. We recommend you to try understanding the basics of JAX first, using these resources:\n",
+    "- Thinking in JAX: https://docs.jax.dev/en/latest/notebooks/thinking_in_jax.html\n",
+    "- JAX 101: https://docs.jax.dev/en/latest/jax-101.html\n",
+    "- Key concepts in JAX: https://docs.jax.dev/en/latest/key-concepts.html#jax-arrays-jax-array\n",
+    "- Flax 101: https://flax-linen.readthedocs.io/en/latest/guides/flax_fundamentals/index.html\n",
+    "\n",
+    "If you do not wish to learn/use JAX, there is another guide in this same director, called quickstart.ipynb, that is to get started with PyTorch.\n",
```
ptrendx: I don't think this sentence is needed.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+   "id": "b48b8ac0",
+   "metadata": {},
+   "source": [
+    "# Getting Started\n",
+    "\n",
+    "## Overview\n",
+    "\n",
+    "Transformer Engine (TE) is a library for accelerating Transformer models on NVIDIA GPUs, providing better performance with lower memory utilization in both training and inference. It provides support for 8-bit floating point (FP8) precision on Hopper GPUs, implements a collection of highly optimized building blocks for popular Transformer architectures, and exposes an automatic-mixed-precision-like API that can be used seamlessly with your PyTorch code. It also includes a framework-agnostic C++ API that can be integrated with other deep learning libraries to enable FP8 support for Transformers.\n",
+    "\n",
+    "This is specifically a guide for getting started to use Transformer Engine with JAX. We recommend you to try understanding the basics of JAX first, using these resources:\n",
```
ptrendx:
```suggestion
    "This guide shows how to start using Transformer Engine with JAX. Similar tutorial for pyTorch is available [here](quickstart.ipynb) \n",
    "\n", 
    "We recommend you to try understanding the basics of JAX first, using these resources:\n",
```

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    "- Thinking in JAX: https://docs.jax.dev/en/latest/notebooks/thinking_in_jax.html\n",
+    "- JAX 101: https://docs.jax.dev/en/latest/jax-101.html\n",
+    "- Key concepts in JAX: https://docs.jax.dev/en/latest/key-concepts.html#jax-arrays-jax-array\n",
+    "- Flax 101: https://flax-linen.readthedocs.io/en/latest/guides/flax_fundamentals/index.html\n",
+    "\n",
+    "If you do not wish to learn/use JAX, there is another guide in this same director, called quickstart.ipynb, that is to get started with PyTorch.\n",
+    "\n",
+    "## Let's build a Transformer layer (*)!\n",
+    "<small>(*) _This was based upon the GPT decoder layer, but for the sake of simplicity and mirroring the PyTorch tutorial whose defaults are without any attention mask, we are setting attention mask here also to 0, making the attention basically an encoder, which does not exist in the GPT architecture. However, since the code support any attention mask here in the TransformerLayer (later in this guide), we will leave it to the audience to try experimenting with different attention masks._</small>\n",
```
ptrendx: This is not needed also - let's just say that we are building an encoder layer in the first real sentence of this section.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    "</figure>\n",
+    "\n",
+    "We construct the components as follows:\n",
+    "\n",
+    "- `LayerNorm`: `nn.LayerNorm` (JAX/Flax)\n",
+    "- `QKV Projection`: `nn.Dense` (conceptually three `Dense` layers for Q, K, and V separately, but we fuse into a single `Dense` layer that is three times larger)\n",
+    "- `DotProductAttention`: `DotProductAttention` from [quickstart_jax_utils.py](quickstart_jax_utils.py)\n",
+    "- `Projection`: `nn.Dense` (JAX/Flax)\n",
+    "- `Dropout`: `nn.Dropout` (JAX/Flax)\n",
+    "- `MLP`: `BasicMLP` from [quickstart_jax_utils.py](quickstart_jax_utils.py)\n",
```
ptrendx: @ksivaman putting on your radar that this will require more substitutions when we create the docs package.

docs/examples/quickstart_jax.ipynb
ptrendx: I don't think this comment really got resolved. Let me make a suggestion how I think it could look like instead.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    "        \n",
+    "        # Attention built-in. Comment out if not used\n",
+    "        # attention = flax.nnx.MultiheadAttention(\n",
+    "        #     num_heads=self.num_attention_heads,\n",
+    "        #     in_features=self.hidden_size,\n",
+    "        #     qkv_features=self.kv_channels,\n",
+    "        #     dropout_rate=self.attention_dropout,\n",
+    "        #     deterministic=True\n",
+    "        # )\n",
+    "        # x = attention(q, k, v, attention_mask, deterministic=deterministic)\n",
```
ptrendx:
```suggestion
    "        # We implement two variants for the attention implementation - from scratch implementation\n",
    "        # for illustrative purposes and one using FLAX. Please choose the one you would like to use\n",
    "        # by commenting out the other one.\n",
    "        # From scratch attention implementation.\n",
    "        attention = utils.DotProductAttention(\n",
    "            num_attention_heads=self.num_attention_heads,\n",
    "            kv_channels=self.kv_channels,\n",
    "            attention_dropout=self.attention_dropout,\n",
    "        )\n",
    "        x = attention(q, k, v, attention_mask, deterministic=deterministic)\n",
    "        \n",
    "        # FLAX Attention implementation.\n",
    "        # attention = flax.nnx.MultiheadAttention(\n",
    "        #     num_heads=self.num_attention_heads,\n",
    "        #     in_features=self.hidden_size,\n",
    "        #     qkv_features=self.kv_channels,\n",
    "        #     dropout_rate=self.attention_dropout,\n",
    "        #     deterministic=True\n",
    "        # )\n",
    "        # x = attention(q, k, v, attention_mask, deterministic=deterministic)\n",
```

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    "The JAX/Flax BasicTransformerLayer above is equivalent to the PyTorch version in the main quickstart.ipynb notebook. It uses:\n",
+    "\n",
+    "- `nn.LayerNorm`: JAX/Flax LayerNorm\n",
+    "- `nn.Dense`: JAX/Flax Dense layer for QKV projection  \n",
+    "- `DotProductAttention`: Custom attention from [quickstart_jax_utils.py] (**)(quickstart_jax_utils.py)\n",
+    "- `nn.Dense`: JAX/Flax Dense layer for projection\n",
+    "- `nn.Dropout`: JAX/Flax Dropout\n",
+    "- `BasicMLP`: Custom MLP from [quickstart_jax_utils.py](quickstart_jax_utils.py)\n",
+    "\n",
+    "<small> (**) _The code below also shows how to use the built-in attention sub-layer from either pure Flax or TE Flax in commented code if you wish to use those instead of the custom attention in [quickstart_jax_utils.py]. The implementation is there for your reference of how attention is roughly implemented in our source_</small>\n",
```
ptrendx: I don't like this sentence - it sounds too informal. A quick pass with copilot gave this instead:

The code snippet below also includes commented examples demonstrating how to utilize the built-in attention sub-layer from either pure Flax or TE Flax. These alternatives are provided should you prefer them over the custom attention implementation found in quickstart_jax_utils.py. This reference is intended to offer insight into the general structure and approach used for implementing attention mechanisms in our source code.

I don't really like that one either (although it sounds better), but maybe we could use it to get something inbetween.

docs/examples/quickstart_jax.ipynb
ptrendx: And "BasicTransformLayer" -> "BasicTransformerLayer"

ptrendx: And DenseGeneral instead of Linear :-).

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+    "        #     head_dim=self.kv_channels,\n",
+    "        #     attention_dropout=self.attention_dropout,\n",
+    "        #     attention_mask_type='no_mask',\n",
+    "        # )\n",
+    "\n",
+    "         # x = attention(q, k, v, attention_mask, deterministic=deterministic)\n",
+    "\n",
+    "        # Projection concatenated QKVoutput back to hidden size.\n",
+    "        # Delete this if use the buiil-in MultiheadAttention module from either flax or te_flax\n",
+    "        x = te_flax.DenseGeneral(features=self.hidden_size, use_bias=True)(x)\n",
```
ptrendx: What would be better (and as I'm thinking about it probably would be better in the previous case too) would be to just create a variable `attention_implementation` being an enum CustomAttention/TEAttention/FLAXAttention which the user could set here to choose the path rather than doing this commenting/uncommenting business. That way you would also be able to include the projection gemm inside the same branch rather than telling the users to comment that out too.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,763 @@
...
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Mean time: 17.390952110290527 ms\n"
+     ]
+    }
+   ],
+   "source": [
+    "import quickstart_jax_utils\n",
+    "importlib.reload(quickstart_jax_utils)\n",
```
ptrendx: Why do we need to reload it?

docs/examples/quickstart_jax.ipynb
ptrendx: I also don't think we need to have the (currently only available in pyTorch) comment - the information is universal, even if the code snippets are not.

docs/examples/quickstart_jax.ipynb
tdophung: I added the word encoder to "Let's build a Transformer encoder layer"

docs/examples/quickstart_jax.ipynb
tdophung: we don't. I was updating the utils file as I was working on the notebook for parameters sharing and accidentally left this behine. Removing

pggPL:
1. It's not being rendered in the docs now, so I cannot see it - you need to put it into some proper .rst file.
2. I had a discussion with @ptrendx yesterday and he said that you considered using tabs like in d2l.ai. I did something like this [here](https://github.com/NVIDIA/TransformerEngine/pull/2343/files#diff-ca48d6c2a4137dd16c8cf699458568ddd1a74d2a5a01047b9e74cacd248512d5R124-R173) using sphinx-tabs package.  You can download the docs from [here](https://github.com/NVIDIA/TransformerEngine/actions/runs/19076147193?pr=2343) and go to section features -> low precision training -> custom recipe to see it. Tbh it is 100% cursor generated, so the code may not make sense, but it looks nice. Idk if you want to use it, just letting you know.
Zrzut ekranu 2025-11-4 o 17 57 00

docs/examples/quickstart_jax.ipynb
tdophung: After taking a look yesterday it was because the dropout rate is not default to 0. And even setting return_layernorm_output=False, would also not eliminate the Drop out step.
After explicitly setting intermediate_dropout_rate to 0.0, I finally observed the same run time between the fuse and unfused versions (when both using the same DPA). 

Adding here for completeness to resolve

docs/examples/quickstart_jax.ipynb
tdophung: I put in the attention wrapper to switch between different implementations.

docs/examples/quickstart_jax.ipynb
tdophung: I put in the attention wrapper to switch between different implementations.

tdophung:
Wonder why I cannot reply to Pawel's suggestion to do tabbing in html. But I intend for this MR to just be the notebook first and work on a separate PR for displaying on html since that will also touch the Python quickstart guide. This PR is just for a functional ipynb

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,699 @@
...
+    "- Flax 101: https://flax-linen.readthedocs.io/en/latest/guides/flax_fundamentals/index.html\n",
+    "\n",
+    "## Let's build a Transformer encoder layer (*)!\n",
+    "<small>(*) _This was based upon the GPT decoder layer, but for the sake of simplicity and mirroring the PyTorch tutorial whose defaults are without any attention mask, we are setting attention mask here also to 0, making the attention basically an encoder, which does not exist in the GPT architecture. However, since the code support any attention mask here in the TransformerLayer (later in this guide), we will leave it to the audience to try experimenting with different attention masks._</small>\n",
+    "\n",
+    "<div class=\"alert alert-info\">\n",
+    "\n",
+    "<b>Summary</b>\n",
+    "    \n",
+    "We build a basic Transformer layer using regular JAX modules. This will be our baseline for later comparisons with Transformer Engine.\n",
```
phu0ngng: Flax modules.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,699 @@
...
+    "        \n",
+    "        return x + res\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3a69efd5",
+   "metadata": {},
+   "source": [
+    "That's it! We now have a simple Transformer layer in JAX/Flax. Let's test it:\n"
```
phu0ngng:
```suggestion
    "That's it! We now have a simple Transformer layer in JAX. Let's test it:\n"
```

docs/examples/quickstart_jax.ipynb
phu0ngng: What is the purpose of having both the `flax.linen.MultiHeadDotProductAttention` and the self-built attention? If the self-built attention is implemented correctly, they should have exactly the same ops as the linen one, right? 

I think it will be less confusing to simply start with the linen attention module directly, then swap it with the TE one. In that way, we can expose those two modules in this tutorial directly rather than hiding their details in the util file.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,699 @@
...
+     "output_type": "stream",
+     "text": [
+      "Mean time: 16.173095703125 ms\n"
+     ]
+    }
+   ],
+   "source": [
+    "y = fused_te_transformer.apply(fused_te_params, x, attention_mask=None, deterministic=True)\n",
+    "\n",
+    "# with jax.profiler.trace(\"/tmp/fused2\"):\n",
```
phu0ngng: cleanup

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,699 @@
...
+   "cell_type": "code",
+   "execution_count": 19,
+   "id": "6b0c705e",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Mean time: 16.173095703125 ms\n"
```
phu0ngng: I suggest we add an option to switch between the two Attention modules in the call function, then we have 2 blocks of time measurement, one for the FusedTETransformerLayer + linen.DPA, one for FusedTETransformerLayer + te_flax.DPA.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,699 @@
...
+    "    forward_kwargs={\"attention_mask\": None, \"deterministic\": False},\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "a45c12c8",
+   "metadata": {},
+   "source": [
+    "Finally, the `TransformerLayer` module is convenient for creating standard Transformer architectures and it provides the highest degree of performance optimization:"
```
phu0ngng: This is quite misleading, as it is right now. 

It shows that the TE Transformer layer is 11ms while the FusedTETransformerLayer is 16ms (which was measured with linen.DPA). So with my above suggestions, we should be able to show that TE Transformer and FusedTETransformerLayer have the same runtime.

I don't think the TransformerLayer could be any more performant than the FusedTETransformerLayer, as the performant part comes from the fact that we try to fuse kernels together in the fused modules. In other words, it should not be different if users use our TransformerLayer or make their own FusedTETransformerLayer by stacking the fused modules.

So the TransformerLayer is there just for convenient usage.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,699 @@
...
+    "\n",
+    "<div class=\"alert alert-info\">\n",
+    "\n",
+    "<b>Summary</b>\n",
+    "    \n",
+    "We optimize the example Transformer layer with TE modules for fused operations.\n",
+    "\n",
+    "</div>\n",
+    "\n",
+    "The `DenseGeneral` layer is enough to build any Transformer model and it enables usage of Transformer Engine even for very custom Transformers. However, having more knowledge about the model allows for additional optimizations like kernel fusion, increasing the achievable speedup.\n",
```
phu0ngng:
```suggestion
    "The `DenseGeneral` layer is enough to build any Transformer model, and it enables usage of the Transformer Engine even for very custom Transformers. However, having more knowledge about the model allows for additional optimizations, such as kernel fusions in mixed-precision training recipes, increasing the achievable speedup.\n",
```

.github/workflows/build.yml
jberchtold-nvidia: This is unrelated and will go away after rebasing, right?

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,699 @@
...
+  {
+   "cell_type": "markdown",
+   "id": "b48b8ac0",
+   "metadata": {},
+   "source": [
+    "# Getting Started\n",
+    "\n",
+    "## Overview\n",
+    "\n",
+    "Transformer Engine (TE) is a library for accelerating Transformer models on NVIDIA GPUs, providing better performance with lower memory utilization in both training and inference. It provides support for 8-bit floating point (FP8) precision on Hopper GPUs, implements a collection of highly optimized building blocks for popular Transformer architectures, and exposes an automatic-mixed-precision-like API that can be used seamlessly with your JAX code. It also includes a framework-agnostic C++ API that can be integrated with other deep learning libraries to enable FP8 support for Transformers.\n",
```
jberchtold-nvidia: > It provides support for 8-bit floating point (FP8) precision on Hopper GPUs

We should update this to a newer message. I cc'd you on the same discussion about our TE GitHub description now that we support nvfp4. You can copy whichever wording we decide there into here

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,699 @@
...
+    "import transformer_engine.jax.flax as te_flax\n",
+    "from transformer_engine.jax.quantize import is_fp8_available, ScalingMode"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f28cb444",
+   "metadata": {},
+   "source": [
+    "TE provides a set of FLAX Linen modules that can be used to build Transformer layers. The simplest of the provided modules are the `DenseGeneral ` and `LayerNorm` layers, which we can use instead of `flax.linen.Dense` and ` flax.linen.LayerNorm`. Let's modify our `BasicTransformerLayer`:"
```
jberchtold-nvidia: nit: FLAX -> Flax

docs/examples/quickstart_jax.ipynb
jberchtold-nvidia: I'm debating how we should explain these fused modules to the user. One the one hand, the fused blocks are recommended if possible as they should in general give TE more information to enable potential speed ups. This is especially true for PyTorch with CPU overhead and I'd imagine the PyTorch getting started does show a speedup here.

However, for JAX with JIT and without quantization they will be equivalent speeds since it's all lowered to the same kernels anyways. So users may think our speedups aren't good

@tdophung @phu0ngng what do you think about moving the FP8 section up before this, so we can used fused layers to show a perf improvement?

So the new ordering would be:
1. Basic pure-JAX/Flax
2. TE unfused layers
3. TE unfused layers + FP8
4. TE fused layers + FP8
5. TE's TransformerLayer

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,699 @@
...
+    "\n",
+    "</div>\n",
+    "\n",
+    "Enabling FP8 support is very simple in Transformer Engine. We just need to wrap the modules within an [autocast](.../api/jax.rst#transformer_engine.jax.fp8_autocast) context manager. See the [FP8 tutorial](fp8_primer.ipynb) for a detailed explanation of FP8 recipes and the supported options.\n",
+    "\n",
+    "<div class=\"alert alert-warning\">\n",
+    "\n",
+    "<b>Important: FP8 Metadata Initialization</b>\n",
+    "\n",
+    "When using FP8, the model **must be initialized within the `fp8_autocast` context**. This creates a special collection called `fp8_metas` that contains scaling factors and other metadata required for FP8 computation. If you initialize a model outside of `fp8_autocast` and then try to use it with FP8, you will get a `ScopeCollectionNotFound` error because the `fp8_metas` collection was never created.\n",
```
jberchtold-nvidia: nit: `fp8_autocast`  context -> `autocast` context

jberchtold-nvidia: Is this description of fp8_metas copied from the PyTorch tutorial? I think PyTorch needs this metadata to store all their scaling factors for every recipe since it's not easily supported in PyTorch. But in TE/JAX, we don't need this for scaling factors. We only need this for recipes that store state across iterations, which is only DelayedScaling's amax history and scale

We also do have some partial support for creating quantizers outside of `autocast` contexts, so I don't think the `ScopeCollectionNotFound` error applies to JAX. But if you have observed, lmk and I can take a look. Thanks!

tdophung: This was not copied from PyTorch tutorial. 
I actually observed this while creating the JAX tutorial, hence the warning. Please take a look!

jberchtold-nvidia: Good to know, I'll take a look

docs/examples/quickstart_jax.ipynb
tdophung: I wonder if it is helpful to have the self-built attention since it shows how attention works and also stay close to the Pytorch tutorial

docs/examples/quickstart_jax.ipynb
tdophung: Currently the FusedTETransformerLayer is 16ms, and the BasicTETransformerLayer are both at 16ms and they both already use te_flax.DPA `from transformer_engine.jax.flax.transformer import DotProductAttention`, NOT linen.DPA.  
I was not aware that TE Transformer layer (11ms) is not supposed to be much faster than the other 2 before this. I can try to take a look at why that is the case (that the other 2 are 16ms even with te_flax.DPA). But please comment if you see anything more apparent

docs/examples/quickstart_jax.ipynb
phu0ngng: I think this is probably a better structure. 
Fused modules only have a performance advantage in FP8 recipes, as the quantization can be fused without ops.

.github/workflows/build.yml
tdophung: right

docs/examples/quickstart_jax.ipynb
tdophung: I have figured out why this was the case. Simply put I passed the wrong expected shape of input into the TransformerLayer while passing the correct expected shape into the Fused version.
I initiated the input in the same shape as what pytorch would expect (seq_len, batch, hidden] while the TransformerLayer takes in [batch, seq_len, hidden] by default. in te_flax.DPA I correctly set transpose_batch_sequence=True in the argument to account for this, but in TransformerLayer I did not. Therefore the max_seqlen was set wrong and TransformerLayer did a lot less work than it supposed to (batch=4, seqlen=2048)

After I fixed this in the most recent commit, this is fixed

docs/examples/quickstart_jax.ipynb
tdophung: In the latest version, I changed the structure of the tutorial to bring FP8 up and Fused down as suggested before, and added the comparison between BasicTETransformerLayer + linen.DPA and BasicTETransformerLayer + te_flax.DPA already, so I do not think it is necessary to do the same for Fused.

Also, since Fused is now only under the FP8 section, there is no XLA ffi for linen.DPA with fp8 autocast so I could not have done this anyways. Please me know what you think

docs/examples/quickstart_jax.ipynb
jberchtold-nvidia: This is the wording that the GitHub repo description got updated to

> 8-bit and 4-bit floating point (FP8 and FP4) precision on Hopper, Ada and Blackwell GPUs

tdophung: I have already updated it to 
"It provides support for 8-bit floating point (FP8) precision on Hopper, Ada, as well as 8-bit and 4-bit floating point (NVFP4) precision on Blackwell GPUs"

 which is more precise because we do not have 4-bit precision on Hopper or Ada?

jberchtold-nvidia: Oh nice! Missed that. I agree, that sounds good, thanks!

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,749 @@
...
+    "This guide shows how to start using Transformer Engine with JAX. Similar tutorial for pyTorch is available [here](quickstart.ipynb).\n",
+    "We recommend you to try understanding the basics of JAX first, using these resources:\n",
+    "\n",
+    "- Thinking in JAX: https://docs.jax.dev/en/latest/notebooks/thinking_in_jax.html\n",
+    "- JAX 101: https://docs.jax.dev/en/latest/jax-101.html\n",
+    "- Key concepts in JAX: https://docs.jax.dev/en/latest/key-concepts.html#jax-arrays-jax-array\n",
+    "- Flax 101: https://flax-linen.readthedocs.io/en/latest/guides/flax_fundamentals/index.html\n",
+    "\n",
+    "## Let's build a Transformer encoder layer (*)!\n",
+    "<small>(*) _This was based upon the GPT decoder layer, but for the sake of simplicity and mirroring the PyTorch tutorial whose defaults are without any attention mask, we are setting attention mask here also to 0, making the attention basically an encoder, which does not exist in the GPT architecture. However, since the code support any attention mask here in the TransformerLayer (later in this guide), we will leave it to the audience to try experimenting with different attention masks._</small>\n",
```
jberchtold-nvidia: > This was based upon the GPT decoder layer, but for the sake of simplicity and mirroring the PyTorch tutorial whose defaults are without any attention mask, we are setting attention mask here also to 0, making the attention basically an encoder, which does not exist in the GPT architecture

I don't think we need to compare to PyTorch here. In my opinion, the goal of this tutorial shouldn't be to convince users to use JAX. But for users who are already using JAX, we should show them why using JAX+TE is a win.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,785 @@
...
+    "The JAX/Flax BasicTransformerLayer above uses:\n",
+    "\n",
+    "- `nn.LayerNorm`: JAX/Flax LayerNorm\n",
+    "- `nn.Dense`: JAX/Flax Dense layer for QKV projection  \n",
+    "- `DotProductAttention`: Custom attention from [quickstart_jax_utils.py] (**)(quickstart_jax_utils.py)\n",
+    "- `nn.Dense`: JAX/Flax Dense layer for projection\n",
+    "- `nn.Dropout`: JAX/Flax Dropout\n",
+    "- `BasicMLP`: Custom MLP implemented from `nn.Dense\n",
+    "\n",
+    "<small> (**) _The code below also shows how to use the built-in attention sub-layer from either pure Flax or TE Flax in commented code if you wish to use those instead of the custom attention in [quickstart_jax_utils.py]. The implementation is there for your reference of how attention is roughly implemented in our source_</small>\n",
```
jberchtold-nvidia: >     "<small> (**) _The code below also shows how to use the built-in attention sub-layer from either pure Flax or TE Flax in commented code if you wish to use those instead of the custom attention in [quickstart_jax_utils.py]. The implementation is there for your reference of how attention is roughly implemented in our source_</small>\n",


This can be removed now that we've removed the custom attention impl

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,785 @@
...
+    "</div>\n",
+    "\n",
+    "The JAX/Flax BasicTransformerLayer above uses:\n",
+    "\n",
+    "- `nn.LayerNorm`: JAX/Flax LayerNorm\n",
+    "- `nn.Dense`: JAX/Flax Dense layer for QKV projection  \n",
+    "- `DotProductAttention`: Custom attention from [quickstart_jax_utils.py] (**)(quickstart_jax_utils.py)\n",
+    "- `nn.Dense`: JAX/Flax Dense layer for projection\n",
+    "- `nn.Dropout`: JAX/Flax Dropout\n",
+    "- `BasicMLP`: Custom MLP implemented from `nn.Dense\n",
```
jberchtold-nvidia: nit: missing closing backtick ` on nn.Dense here

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,785 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": 7,
+   "id": "bed20d6b",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import transformer_engine.jax as te\n",
+    "import transformer_engine.jax.flax as te_flax\n",
+    "from transformer_engine.jax.quantize import is_fp8_available, ScalingMode"
```
jberchtold-nvidia: > `from transformer_engine.jax.quantize import is_fp8_available, ScalingMode`

I don't see `is_fp8_available` or `ScalingMode` used later in the notebook. Let's remove them

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,785 @@
...
+    "    forward_kwargs={\"attention_mask\": None, \"deterministic\": False},\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c9a101d3",
+   "metadata": {},
+   "source": [
+    "## Enabling FP8\n",
```
jberchtold-nvidia: nit: Here and below can we replace FP8 -> Quantization unless specifically talking about an fp8-specific recipe. For the first mention of quantization, this title, we can mention FP8 and FP4 explicitly "Enabling Quantization (FP8 or FP4)"

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,785 @@
...
+     "text": [
+      "Mean time: 10.065140724182129 ms\n"
+     ]
+    }
+   ],
+   "source": [
+    "with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):\n",
+    "    basic_te_params = basic_te_transformer.init(key, x, attention_mask=None, deterministic=False)\n",
+    "\n",
+    "    y = basic_te_transformer.apply(basic_te_params, x, attention_mask=None, deterministic=True)\n",
```
jberchtold-nvidia: Why do we need this line? I don't see `y` used later
```
y = basic_te_transformer.apply(basic_te_params, x, attention_mask=None, deterministic=True)
```

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,785 @@
...
+   "source": [
+    "fused_te_transformer = FusedTETransformerLayer(\n",
+    "    hidden_size, \n",
+    "    ffn_hidden_size, \n",
+    "    num_attention_heads\n",
+    ")\n",
+    "\n",
+    "with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):\n",
+    "    fused_te_params = fused_te_transformer.init(key, x, attention_mask=None, deterministic=False)\n",
+    "    y = fused_te_transformer.apply(fused_te_params, x, attention_mask=None, deterministic=True)\n",
```
jberchtold-nvidia: Again here, I don't see `y` used later. Can this be removed or is there other reasoning I'm missing, thanks!

docs/examples/quickstart_jax.ipynb
jberchtold-nvidia: @tdophung now that the performance matches, let's remove "it provides the highest degree of performance optimization:" since TransformerLayer and individual usage of fused LayerNormMLP, etc. are equivalent

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,785 @@
...
+    "    use_bias=True,\n",
+    "    intermediate_dropout=0.0,\n",
+    "    enable_relative_embedding=False,\n",
+    "    self_attn_bias_type='no_bias',\n",
+    "    hidden_dropout=0.0\n",
+    ")\n",
+    "\n",
+    "with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):\n",
+    "    te_transformer_params = te_transformer.init(key, x, deterministic=False)\n",
+    "    y = te_transformer.apply(te_transformer_params, x, attention_mask=None, deterministic=True)"
```
jberchtold-nvidia, line 731: Same question as above for `y = ...`

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,785 @@
...
+    "This guide shows how to start using Transformer Engine with JAX. Similar tutorial for pyTorch is available [here](quickstart.ipynb).\n",
+    "We recommend you to try understanding the basics of JAX first, using these resources:\n",
+    "\n",
+    "- Thinking in JAX: https://docs.jax.dev/en/latest/notebooks/thinking_in_jax.html\n",
+    "- JAX 101: https://docs.jax.dev/en/latest/jax-101.html\n",
+    "- Key concepts in JAX: https://docs.jax.dev/en/latest/key-concepts.html#jax-arrays-jax-array\n",
+    "- Flax 101: https://flax-linen.readthedocs.io/en/latest/guides/flax_fundamentals/index.html\n",
+    "\n",
+    "## Let's build a Transformer encoder layer (*)!\n",
+    "<small>(*) _This was based upon the GPT decoder layer, but for the sake of simplicity and mirroring the PyTorch tutorial whose defaults are without any attention mask, we are setting attention mask here also to 0, making the attention basically an encoder, which does not exist in the GPT architecture. However, since the code support any attention mask here in the TransformerLayer (later in this guide), we will leave it to the audience to try experimenting with different attention masks._</small>\n",
```
jberchtold-nvidia: Since we are no longer tied to the PyTorch tutorial and no longer have the custom attention, how hard would it be to use a causal mask so we can remove this caveat? Looks like Flax linen has a built-in function for creating the mask here: https://flax.readthedocs.io/en/v0.6.11/api_reference/_autosummary/flax.linen.make_causal_mask.html

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,785 @@
...
+    "Let's start with creating the transformer layer using plain [FLAX Linen](https://flax.readthedocs.io/en/stable/) . Figure 1 shows the overall structure.\n",
+    "\n",
+    "<figure align=\"center\">\n",
+    "<img src=\"transformer_layer.png\" width=\"20%\">\n",
+    "<figcaption> Figure 1: Structure of a GPT encoder layer.</figcaption>\n",
+    "</figure>\n",
+    "\n",
+    "We construct the components as follows:\n",
+    "\n",
+    "- `LayerNorm`: `nn.LayerNorm` (JAX/Flax)\n",
```
phu0ngng:
```suggestion
    "- `LayerNorm`: `nn.LayerNorm` (Flax)\n",
```
and other places.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,785 @@
...
+    "- `QKV Projection`: `nn.Dense` (conceptually three `Dense` layers for Q, K, and V separately, but we fuse into a single `Dense` layer that is three times larger)\n",
+    "- `DotProductAttention`: `DotProductAttention` from [quickstart_jax_utils.py](quickstart_jax_utils.py)\n",
```
phu0ngng: I think we are no longer using the utils.DPA?

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,857 @@
...
+    "    \n",
+    "Now that we have a basic Transformer layer in JAX/Flax, let's use Transformer Engine to speed up the training. The following examples show how to use TE modules.\n",
+    "\n",
+    "</div>\n",
+    "\n",
+    "The JAX/Flax BasicTransformerLayer above is equivalent to the PyTorch version in the main quickstart.ipynb notebook. It uses:\n",
+    "\n",
+    "- `nn.LayerNorm`: JAX/Flax LayerNorm\n",
+    "- `nn.Dense`: JAX/Flax Dense layer for QKV projection  \n",
+    "- `DotProductAttention`: Custom attention from [quickstart_jax_utils.py] (**)(quickstart_jax_utils.py)\n",
```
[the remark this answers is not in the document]
phu0ngng: Same here, Flax and Flax MHA.

docs/examples/quickstart_jax.ipynb
phu0ngng: I think we don't have commented code anymore?

docs/examples/quickstart_jax.ipynb
phu0ngng: @tdophung Great finding!

docs/examples/quickstart_jax.ipynb
tdophung: These are  to show that calling it in 1 direction (forward only) and take away obscurity in the speedometer before we show performance using speedometer.
I also got this from the Pytorch tutorial and ^ is what I infer about their purpose of having this line

jberchtold-nvidia: I see, I like that! Can you add a small comment here and the instances below like `# Example usage of forward` to clarify to users that this isn't required for functionality? But I agree that we should keep it as a useful example

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,784 @@
...
+   "cell_type": "code",
+   "execution_count": 8,
+   "id": "56105579",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from transformer_engine.jax.flax.transformer import DotProductAttention as TEDotProductAttention\n",
+    "\n",
+    "\n",
+    "class BasicTEMLP(nn.Module):\n",
```
jberchtold-nvidia: `TEUnfusedMLP`

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,784 @@
...
+    "This guide shows how to start using Transformer Engine with JAX. Similar tutorial for pyTorch is available [here](quickstart.ipynb).\n",
+    "We recommend you to try understanding the basics of JAX first, using these resources:\n",
+    "\n",
+    "- Thinking in JAX: https://docs.jax.dev/en/latest/notebooks/thinking_in_jax.html\n",
+    "- JAX 101: https://docs.jax.dev/en/latest/jax-101.html\n",
+    "- Key concepts in JAX: https://docs.jax.dev/en/latest/key-concepts.html#jax-arrays-jax-array\n",
+    "- Flax 101: https://flax-linen.readthedocs.io/en/latest/guides/flax_fundamentals/index.html\n",
+    "\n",
+    "## Let's build a Transformer encoder layer (*)!\n",
+    "<small>(*) _This was based upon the GPT decoder layer, but for the sake of simplicity, we are setting attention mask here also to 0, making the attention basically an encoder, which does not exist in the GPT architecture. However, since the code support any attention mask here in the TransformerLayer (later in this guide), we will leave it to the audience to try experimenting with different attention masks._</small>\n",
```
ptrendx: I don't think this is really needed to be honest. It is fine to just say that we are building an encoder layer, no need to make extra explanations :-).

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,784 @@
...
+    "\n",
+    "<figure align=\"center\">\n",
+    "<img src=\"transformer_layer.png\" width=\"20%\">\n",
+    "<figcaption> Figure 1: Structure of a GPT encoder layer.</figcaption>\n",
+    "</figure>\n",
+    "\n",
+    "We construct the components as follows:\n",
+    "\n",
+    "- `LayerNorm`: `nn.LayerNorm` (Flax)\n",
+    "- `QKV Projection`: `nn.Dense` (conceptually three `Dense` layers for Q, K, and V separately, but we fuse into a single `Dense` layer that is three times larger)\n",
```
ptrendx:
```suggestion
    "- `QKV Projection`: `nn.Dense` (conceptually there are three separate `Dense` layers for Q, K, and V, but we fuse them together into a single `Dense` layer that is three times larger)\n",
```

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,784 @@
...
+    "- `DotProductAttention`: `nn.MuliheadDotProductAttention` (Flax)\n",
+    "- `Projection`: `nn.Dense` (Flax)\n",
+    "- `Dropout`: `nn.Dropout` (Flax)\n",
+    "- `MLP`: `FlaxMLP` implemented from `nn.Dense` and `nn.gelu`\n",
```
ptrendx:
```suggestion
    "- `MLP`: `FlaxMLP` implemented using `nn.Dense` and `nn.gelu`\n",
```

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,784 @@
...
+    "        \n",
+    "        return x + res\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3a69efd5",
+   "metadata": {},
+   "source": [
+    "That's it! We now have a simple Transformer layer in JAX. Let's test it:\n"
```
ptrendx: Ending the section with a colon is strange. Did you intend to put some code to show it working? Or are you just doing a segue to the next section about performance bechmarking? If so, I think it is not really needed - that section starts with basically the same thing anyway.

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,784 @@
...
+    "\n",
+    "<div class=\"alert alert-info\">\n",
+    "\n",
+    "<b>Summary</b>\n",
+    "    \n",
+    "Now that we have a basic Transformer layer in Flax, let's use Transformer Engine to speed up the training. The following examples show how to use TE modules.\n",
+    "\n",
+    "</div>\n",
+    "\n",
+    "The Flax FlaxTransformerLayer above uses:\n",
```
ptrendx:
```suggestion
    "As a reminder, the Flax FlaxTransformerLayer above used:\n",
```

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,784 @@
...
+    "\n",
+    "<div class=\"alert alert-info\">\n",
+    "\n",
+    "<b>Summary</b>\n",
+    "    \n",
+    "We configure a TE module to perform compute in FP8.\n",
+    "\n",
+    "</div>\n",
+    "\n",
+    "Enabling FP8 support is very simple in Transformer Engine. We just need to wrap the modules within an [autocast](.../api/jax.rst#transformer_engine.jax.fp8_autocast) context manager. See the [FP8 tutorial](fp8_primer.ipynb) for a detailed explanation of FP8 recipes and the supported options.\n",
```
ptrendx, line 506: fp8_autocast is deprecated - we should switch to the new naming (autocast).

docs/examples/quickstart_jax.ipynb
```diff
@@ -0,0 +1,784 @@
...
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "Mean time: 9.978384971618652 ms\n"
+     ]
+    }
+   ],
+   "source": [
+    "with te.fp8_autocast(enabled=True, fp8_recipe=fp8_recipe):\n",
```
ptrendx: fp8_autocast is deprecated.

docs/examples/quickstart_jax.ipynb
tdophung: done!

docs/examples/quickstart_jax.ipynb
tdophung: I have changed this to decoder per Jeremy comment, so this is no longer needed

jberchtold-nvidia (approved):
LGTM, thanks!

PR #2325, opened 2025-10-31
[Pytorch] change fused cross entropy backward grad to fp32 and reduce one read/…

RandMist:
# Description

The fused cross entropy kernel in Transformer Engine uses 16-bit floating point (BF16) for the backward pass when the input is in BF16, whereas Megatron's VocabParallelCrossEntropy performs its computations in FP32. This discrepancy may lead to divergence in some cases.

This PR also reduces one read of `logits`, which improves the performance by up to 1.25x.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Changed the fused cross entropy backward gradient computation to fp32 for consistency with Megatron's VocabParallelCrossEntropy.
- Optimized the computation logic to reduce one read/write operation of the logits.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/cross_entropy.py
+++ b/transformer_engine/pytorch/cross_entropy.py
@@ -80,7 +80,14 @@ class CrossEntropyFunction(torch.autograd.Function):
         label_smoothing = ctx.label_smoothing
         reduce_loss = ctx.reduce_loss
         _input = triton_cross_entropy.cross_entropy_backward(
-            _input, target, m_d_X_y, grad_output, label_smoothing, reduce_loss, dist_process_group, ctx.is_cg_capturable
+            _input,
+            target,
+            m_d_X_y,
+            grad_output,
+            label_smoothing,
+            reduce_loss,
+            dist_process_group,
+            ctx.is_cg_capturable,
         )
         return (
             _input,
--- a/transformer_engine/pytorch/triton/cross_entropy.py
+++ b/transformer_engine/pytorch/triton/cross_entropy.py
@@ -201,6 +201,7 @@ def cross_entropy_forward_kernel(
 # The optimal maximum block size depends on your hardware, your kernel, and your dtype
 MAX_FUSED_SIZE = 65536 // 2
 
+
 @triton.jit
 def cross_entropy_backward_kernel(
     X_ptr,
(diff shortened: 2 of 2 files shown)


yaox12:
@sanandaraj5597 @timmoon10 Could you please review? The previous BF16 backward may lead to divergence in some cases (reported by several customers).

yaox12:
@RandMist You need to sign-off your commits (`git commit -s`). See [this](https://github.com/NVIDIA/TransformerEngine/pull/2325/checks?check_run_id=54164971208).

DXZDXZ:
@RandMist @yaox12 Hello~ Our experiments found that after applying this change, the output can sometimes become **inf**. This subsequently leads to NaN gradients for the corresponding token during the RMS backward pass. Furthermore, this is a **non-deterministic** issue that cannot be reliably reproduced on a specific input.

sanandaraj5597:
The goal of this kernel was to avoid saving the input for backward. The goal is to write the gradients on the input tensor itself to reduce the peak memory usage. This change defeats that purpose and break our upstream workloads using this kernel. 

I understand you have improved performance but it comes at the cost of extra memory usage which we don't want.

RandMist:
> @RandMist @yaox12 Hello~ Our experiments found that after applying this change, the output can sometimes become **inf**. This subsequently leads to NaN gradients for the corresponding token during the RMS backward pass. Furthermore, this is a **non-deterministic** issue that cannot be reliably reproduced on a specific input.

Hi, I will try to reproduce this phenomenon, or you can provide a script/data for reproduction.

RandMist:
> The goal of this kernel was to avoid saving the input for backward. The goal is to write the gradients on the input tensor itself to reduce the peak memory usage. This change defeats that purpose and break our upstream workloads using this kernel.
> 
> I understand you have improved performance but it comes at the cost of extra memory usage which we don't want.

Hi, the main objective of this new kernel is to keep the backward propagation computations in fp32.

The original kernel stored gradients in the input, whereas the new kernel directly stores the input and additionally saves the target and m_d_X_y. The shapes of these extra tensors are n_rows × 1 and n_rows × 4, respectively. Compared to the input shape of n_rows × n_cols (where n_cols is typically >100,000), this additional memory consumption should therefore be negligible.

RandMist:
> @RandMist @yaox12 Hello~ Our experiments found that after applying this change, the output can sometimes become **inf**. This subsequently leads to NaN gradients for the corresponding token during the RMS backward pass. Furthermore, this is a **non-deterministic** issue that cannot be reliably reproduced on a specific input.

We did not observe this phenomenon in our internal experiments. We trained the 16B MoE model for 20k steps, and the loss/grad norm of the model using the new fused kernel fully matched that of the model using mcore's original unfused loss in fp32 (the old fused kernel with bf16 backward showed an upward trend in grad norm after 15k steps).

PR #2330, opened 2025-11-01
[Core] Fix inconsistent logic in C++ tensor class

timmoon10:
# Description

This PR fixes some hacky logic in the C++ `Tensor` class:
- Construct uninitialized tensors with `shape=[0]`. Previously we constructed them as 0-D tensors, which should have one entry (see https://github.com/NVIDIA/TransformerEngine/pull/2215#discussion_r2427408835).
- Make `Tensor::has_data` and `Tensor::has_columnwise_data` have a consistent meaning: ~"is it safe to touch the data pointer?"~ "is the buffer non-default in any way?"
- Fix code that use `Tensor::has_data`/`Tensor::has_columnwise_data` inappropriately (`Tensor::shape`, swizzling). These are cases where the data is initialized but the pointer is not safe to touch, e.g. when the tensor has a zero dim. Note that I've only fixed places that were causing problems, and I haven't fully gone through the entire code base.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Construct uninitialized tensors with zero entries.
- Fix logic in `Tensor::has_data` and `Tensor::has_columnwise_data`.
- Debug inappropriate usage of `Tensor::has_data`/`Tensor::has_columnwise_data`.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/csrc/extensions/recipe.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/recipe.cpp
@@ -22,8 +22,8 @@ void compute_amax(const at::Tensor& tensor, at::Tensor& amax) {
   TORCH_CHECK(amax.numel() == 1, "amax must have exactly one element");
   auto* amax_ptr = amax.data_ptr<float>();
   TensorWrapper fake_te_output(
-      nullptr, te_input.shape(),
-      DType::kFloat8E4M3,  // It doesn't matter because we only compute amax.
+      /*dptr=*/nullptr, te_input.shape(),
+      DType::kFloat32,  // It doesn't matter because we only compute amax.
       amax_ptr);
 
   nvte_compute_amax(te_input.data(), fake_te_output.data(), at::cuda::getCurrentCUDAStream());


transformer_engine/common/common.h
```diff
@@ -154,11 +155,10 @@ struct Tensor {
...
   }
 
-  bool has_data() const noexcept { return data.dptr != nullptr; }
+  bool has_data() const noexcept { return data.dptr != nullptr && data.numel() != 0; }
 
-  // Check for size (not just pointer) for 0-dim or no token cases.
   bool has_columnwise_data() const noexcept {
-    return columnwise_data.dptr != nullptr || columnwise_data.shape.size() != 0;
+    return columnwise_data.dptr != nullptr && columnwise_data.numel() != 0;
   }
```
ptrendx: Considering both the constructor for SimpleTensor and the clear method, isn't it enough to just check the first element (if it exists?). Then we would avoid the loop in most cases where we do have valid shape.

transformer_engine/common/common.h
```diff
@@ -169,34 +169,52 @@ struct Tensor {
...
+    bool use_columnwise_shape = false;
+    if (data.dptr != nullptr) {
+      use_columnwise_shape = false;
+    } else if (columnwise_data.dptr != nullptr) {
+      use_columnwise_shape = true;
+    } else if (!is_shape_trivial(data.shape)) {
+      use_columnwise_shape = false;
+    } else if (!is_shape_trivial(columnwise_data.shape)) {
+      use_columnwise_shape = true;
+    }
```
ptrendx: And so this is a change in behavior, since now if all of these fail (and so there is no data nor columnwise data and both shapes are trivial) then you end up with dim = 1 instead of 0. Is that expected?

transformer_engine/common/common.h
```diff
@@ -169,34 +169,52 @@ struct Tensor {
...
           use_columnwise_shape = false;
         } else if (columnwise_data.dptr != nullptr) {
           use_columnwise_shape = true;
-        } else if (data.shape.size() != 0) {
+        } else if (!is_shape_trivial(data.shape)) {
           use_columnwise_shape = false;
-        } else if (columnwise_data.shape.size() != 0) {
+        } else if (!is_shape_trivial(columnwise_data.shape)) {
           use_columnwise_shape = true;
         }
```
ptrendx: Why do we need to have the copies of this part for every scaling mode? We could do that before the switch.

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -626,7 +633,8 @@ class TensorWrapper {
    */
   const NVTEShape shape() const noexcept {
     if (tensor_ == nullptr) {
-      return nvte_make_shape(nullptr, 0);
+      const size_t zero = 0;
+      return nvte_make_shape(&zero, 1);
```
ptrendx: Why not just reture empty_shape instead?

transformer_engine/common/swizzle/swizzle.cu
```diff
@@ -332,68 +332,120 @@ __global__ void multi_tensor_swizzle_col_scaling_kernel(MultiSwizzleArgs kernel_
...
+  }
 
-  auto& scaling_mode = input->scaling_mode;
-  NVTE_CHECK(scaling_mode == NVTE_MXFP8_1D_SCALING || scaling_mode == NVTE_NVFP4_1D_SCALING,
-             "Unsupported scaling mode for swizzling.");
-
-  bool nvfp4 = scaling_mode == NVTE_NVFP4_1D_SCALING;
+  // Check if scaling factors are non-trivial
+  const bool has_rowwise_scale_inv =
+      (input->scale_inv.dptr != nullptr && input->scale_inv.numel() != 0);
```
ptrendx: This check seems like something that should be an util function by itself in the SimpleTensor to be honest.

transformer_engine/common/common.h
timmoon10: Yes, an uninitialized tensor now has `shape=[0]`. A tensor with `shape=[]` has 1 entry, so it should not be considered uninitialized. I've propagated this change everywhere, as far as I can tell.

transformer_engine/common/common.h
timmoon10: If a tensor has `shape=[128,0]`, then it is not safe to touch the data even if it is not uninitialized.

transformer_engine/common/swizzle/swizzle.cu
timmoon10: Good point, https://github.com/NVIDIA/TransformerEngine/pull/2388 adds `SimpleTensor::has_data`.

transformer_engine/common/common.h
timmoon10: I prefer not to entangle all of the tensor formats, since there's no logical reason they need to interpret `data` or `columnwise_data` in a consistent way (and in fact, they do not). In practice we only have two ways that these dims are interpreted, so I'll consolidate to reduce the bloat. ~I'll also remove the checks for null pointers since they seem unnecessary~ Turns out we see some errors in the NVFP4 group quantize kernels without the null pointer checks.

transformer_engine/common/include/transformer_engine/transformer_engine.h
timmoon10: `shape=[]` has 1 entry. I've added a `nvte_make_zero_shape` function to make it more convenient to return `shape=[0]`.

transformer_engine/common/common.h
timmoon10: I've removed the loop by chainging the definition to match the original definition of `has_columnwise_data`. Basically, the meaning is now "is the buffer uninitialized?".

transformer_engine/common/common.h
timmoon10: After changing the meaning of `has_data`/`has_columnwise_data` to "is the buffer uninitialized?", it turns out we can simplify this logic.

transformer_engine/common/swizzle/swizzle.cu
```diff
@@ -404,18 +454,30 @@ void swizzle_scaling_factors(const Tensor* input, Tensor* output, cudaStream_t s
 
     int original_M, original_K;
     void *input_scale_inv_ptr, *output_scale_inv_ptr;
```
timmoon10: Nit from @greptile-apps:
```suggestion
    int original_M{0}, original_K{0};
    void *input_scale_inv_ptr{nullptr}, *output_scale_inv_ptr{nullptr};
```

transformer_engine/common/common.h
```diff
@@ -74,38 +74,44 @@ inline size_t product(const std::vector<size_t> &shape) {
...
-      acc *= dim;
-    }
-    return acc;
-  }
-  bool has_data() const noexcept { return dptr != nullptr && numel() > 0; }
+  /*! Number of tensor elements. */
+  size_t numel() const { return product(shape); }
+
+  /*! Whether the tensor is not uninitialized. */
+  bool has_data() const { return !(dptr == nullptr && shape.size() == 1 && shape[0] == 0); }
```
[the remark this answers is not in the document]
timmoon10: I have updated the PR description. Now the meaning is "is the buffer non-default in _any_ way?" and null pointers are perfectly valid.

timmoon10:
Force-pushing to unblock https://github.com/NVIDIA/TransformerEngine/issues/2446.

PR #2266, opened 2025-10-11
Add post-processing API for FP8 primary weights to support CUDA Graph

kunlunl:
# Description

Please include a brief summary of the changes, relevant motivation and context.

Fixes # (issue)

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Please list the changes introduced in this PR:

- Change A
- Change B

# Checklist:

- [ ] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [ ] The functionality is complete
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2342, opened 2025-11-04
Make FP8 weights compatible with older MCore version

kunlunl:
# Description

The old MCore version cannot call the `post_all_gather function` in `transformer_engine/pytorch/tensor/utils.py`. This MR modifies the default behavior of `cast_master_weights_to_fp8` back to remove transpose/columnwise data (making it consistent with before https://github.com/NVIDIA/TransformerEngine/pull/2266 is merged), and adds a flag to manually enable retaining transpose/columnwise data. This prevents users from unknowingly invoking the new `cast_master_weights_to_fp8` without manually calling `post_all_gather_processing`.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Modifies the default behavior of `cast_master_weights_to_fp8` back to remove transpose/columnwise data (making it consistent with before https://github.com/NVIDIA/TransformerEngine/pull/2266 is merged)
- Adds a flag to manually enable retaining transpose/columnwise data.

# Checklist:

### DIFF
--- a/examples/jax/encoder/common.py
+++ b/examples/jax/encoder/common.py
@@ -3,6 +3,9 @@
 # See LICENSE for license information.
 """Shared functions for the encoder tests"""
 from functools import lru_cache
+import os
+import pathlib
+import zipfile
 
 import jax
 import jax.numpy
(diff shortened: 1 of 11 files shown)


transformer_engine/pytorch/tensor/utils.py
```diff
@@ -69,6 +74,9 @@ def cast_master_weights_to_fp8(
     fsdp_shard_model_weights : list of FSDP shard model weights. If None, it means that the model weights are
                              not sharded. Otherwise, it means that the model weights are sharded and we get
                              target model weights data storage using the FSDP shard model weights.
+    keep_columnwise: bool. If True, keep the columnwise data of model weights, and the
+                     post_all_gather_processing function must be called manually by user after the
+                     weights all-gather. Default is False.
```
ksivaman:
```suggestion
    keep_columnwise: bool, default = `False`
                    If True, keep the columnwise data of model weights, and the
                    post_all_gather_processing function must be called manually by user after the
                    weights all-gather. Default is False.
```

transformer_engine/pytorch/tensor/utils.py
```diff
@@ -48,7 +48,12 @@ def replace_raw_data(tensor: QuantizedTensor, new_raw_data: torch.Tensor):
 
 
 def cast_master_weights_to_fp8(
-    model_weights, master_weights, start_offsets, group, fsdp_shard_model_weights=None
+    model_weights,
+    master_weights,
+    start_offsets,
+    group,
+    fsdp_shard_model_weights=None,
+    keep_columnwise=False,
```
timmoon10: This name is not quite correct. If we generalize to MXFP8 and NVFP4, it would be wrong to deallocate the column-wise data when `keep_columnwise=False`. This arg isn't really about column-wise data, but whether we need to post-process after the AG. Maybe a name like `with_post_all_gather_processing` would make this contract more explicit.

tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
```diff
@@ -24,17 +24,20 @@
...
     test_cmd = LAUNCH_CMD + [str(test_path)] + ["--quantization", quantization]
+    if keep_columnwise:
+        test_cmd.append("--keep-columnwise")
     result = subprocess.run(test_cmd, env=os.environ, check=False)
     assert result.returncode == 0
 
 
 @pytest.mark.parametrize("quantization", ["fp8", "fp8_cs", "fp8_block"])
-def test_cast_master_weights_to_fp8(quantization):
+@pytest.mark.parametrize("keep_columnwise", [False, True])
```
timmoon10: Not critical yet, but as we add test cases we should consider reorganizing the distributed launches. Each call to `torchrun` takes ~seconds. If we have many test cases, it's better to call `torchrun` once and perform multiple tests within the same session.

See `distributed/test_fusible_ops.py` for an example where we only launch parallel jobs a few times: https://github.com/NVIDIA/TransformerEngine/blob/0e80c847845be58dbb88f46a0975786ddc823798/tests/pytorch/distributed/test_fusible_ops.py#L1043-L1044
Within each job, we run many test cases:
https://github.com/NVIDIA/TransformerEngine/blob/0e80c847845be58dbb88f46a0975786ddc823798/tests/pytorch/distributed/test_fusible_ops.py#L962

transformer_engine/pytorch/tensor/utils.py
kunlunl: Done.

transformer_engine/pytorch/tensor/utils.py
kunlunl: OK. I've changed the name to `manual_post_all_gather_processing`, does it sound good to you? I kind of feel `with_post_all_gather_processing` is a bit ambiguous to me.

tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
kunlunl: I refactored `cast_master_weights_to_fp8.py` based on the structure of `test_fusible_ops.py`:
* Now, `torchrun` is only launched once, with different test cases executed internally via a for loop.
* The `run_cast_master_weights_to_fp8.py` and `test_cast_master_weights_to_fp8.py` have been merged, with only latter one retained.

PR #2343, opened 2025-11-04
More detailed documentation for recipes

pggPL:
# Description

This PR adds detailed documentation for the Low Precision Training feature in Transformer Engine, covering FP8, MXFP8, NVFP4, and other quantization recipes for both PyTorch and JAX frameworks.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
+..
+    Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+    See LICENSE for license information.
+
+Introduction
+===================================
+
+Transformer Engine accelerates deep learning by leveraging low precision formats on NVIDIA GPUs.
+This chapter introduces mixed precision training and FP8 support.
```
ptrendx: I would change that a little to say that low precision training is only one of the ways TE accelerates training (although very important one :-) ).

docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
...
+
+.. raw:: html
+   :file: img/fp_formats_comparison.svg
+
+*Figure 1: Comparison of FP32, BF16, and FP16 floating-point formats showing bit allocation for sign, exponent, and mantissa.*
+
+The key differences between these formats are:
+
+* **FP32** (32 bits total): 1 sign bit + 8 exponent bits + 23 mantissa bits – standard single-precision format
+* **BF16** (16 bits total): 1 sign bit + 8 exponent bits + 7 mantissa bits – maintains FP32's exponent range but reduced precision
```
ptrendx:
```suggestion
* **BF16** (16 bits total): 1 sign bit + 8 exponent bits + 7 mantissa bits – maintains FP32's exponent range but has reduced precision
```

docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
...
+* **FP16** (16 bits total): 1 sign bit + 5 exponent bits + 10 mantissa bits – reduced range but higher precision than BF16
+
+BF16's advantage is that it shares the same exponent range as FP32, 
+making it easier to convert between the two formats without overflow/underflow issues. 
+FP16 offers better precision for smaller values but has a more limited dynamic range,
```
ptrendx:
```suggestion
FP16 offers better precision for smaller values but has a limited dynamic range,
```

docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
...
+which results in the need to perform loss scaling to avoid overflow/underflow—see `this paper on loss scaling <https://arxiv.org/pdf/1710.03740>`__ for more details.
+
+**Mixed precision**
+
+Not all operations can run in reduced precision.
```
ptrendx:
```suggestion
Not all operations should be run in reduced precision to preserve accuracy.
```

docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
...
+Modern deep learning frameworks use *mixed precision training*, where:
+
+* *Low precision* is used for matrix multiplications and other compute-heavy operations, which remain numerically stable at lower precision,
+* *High precision (FP32)* must be used for numerically sensitive operations to maintain training stability. These include layer normalization, softmax, and loss computations—operations that involve division or exponentiation, where small rounding errors can amplify and propagate through the network, leading to gradient instability or degraded convergence.
```
ptrendx: This is not quite correct, although I don't have the exact suggestion how to change it. Basically there are actually 3 types of operations in mixed precision: the ones that want low precision (like gemms), the ones that require high precision (like loss, sum or exponent), and ones that don't mind the low precision and use high precision internally to perform the calculation. LayerNorm and SoftMax are in that 3rd category - they need high precision but only internally, so if they are implemented as a single kernel then they are actually ok to have inputs/outputs in low precision (they may still need to have the weights in high precision though - this is what we do with LayerNorm for example) - e.g. softmax's output will always be <= 1, so there is no danger that it will overflow.

docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
...
+
+**Master weights**
+
+Mixed precision training also raises the question of how to store model weights.
```
ptrendx:
```suggestion
Another consideration in mixed precision training is how to store the model weights.
```

docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
...
+Lower precision formats like FP16 and BF16 have limited representational granularity, 
+which becomes problematic during gradient updates. 
+When a small gradient is added to a not so small weight stored in low precision, 
+the result may round back to the original value if the update falls below the format's precision threshold.
+Moreover, some elements of the gradient itself can be too small to be represented in low precision.
```
ptrendx:
```suggestion
Moreover, some elements of the gradient itself can be too small to be represented in low precision, especially after the accumulation from multiple GPUs in the data parallel training setting.
```

docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
...
+This ensures that even small gradients accumulate correctly over time.
+
+There are two common software approaches to storing master weights:
+
+* *In the optimizer*: 
+  The model holds low-precision weights, 
+  while the optimizer maintains FP32 copies alongside momentum and other state. 
+  During each step, 
+  the optimizer updates its FP32 copy and casts the result back to the model's low-precision weights. 
+  This makes it easier to shard master weights together with other optimizer state, for example in ZeRO optimizer.
```
ptrendx:
```suggestion
  This approach makes it easier to shard master weights together with other optimizer state, for example in ZeRO optimizer.
  Since the casting happens only during the optimizer step, this approach is also faster when optimizer runs less frequently than the model, e.g. when performing gradient accumulation or pipeline parallel training.
```

docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
...
+*Figure 2: Three approaches to weight storage—low precision only (no master weights), master weights stored in the model, and master weights stored in the optimizer.*
+
+.. tabs::
+
+   .. tab:: PyTorch
+
+      The PyTorch API of Transformer Engine provides two mechanisms to control precision:
+      
+      * **Weight precision**: Use the ``params_dtype`` argument in any TE layer constructor.
+      * **Computation precision**: Use the ``torch.autocast`` context manager.
```
ptrendx: We also use the input dtype to set the computation precision - see set_activation_dtype in base.py

docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
...
+
+   .. tab:: PyTorch
+
+      The PyTorch API of Transformer Engine provides an ``autocast`` context manager to control precision.
+      It's similar to the ``torch.autocast`` context manager, but tailored for low precision training.
+      The most important argument is the ``recipe`` argument, which accepts objects inheriting from
+      :class:`~transformer_engine.common.recipe.Recipe`.
+
+      Forward computations need to be performed inside the ``autocast`` context manager,
+      while the ``.backward()`` call should be outside of it.
```
ptrendx:
```suggestion
      while the ``.backward()`` call should be outside of it (it inherits the setting from the
      corresponding forward pass).
```

docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
...
+
+**Mixed precision with 8- or 4-bit precisions**
+
+From now on, we will refer to FP8/MXFP8/NVFP4 etc. as *low precision*
+and to FP32/BF16/FP16 as *high precision*. This terminology will be
+used throughout the rest of the documentation.
+
+Not all operations run in low precision:
+
+- **Non-attention linear operations**: run in low precision.
```
ptrendx: This is a little unclear, since people have different understanding of what "attention" means - it could be understood as dot product attention or as multihead attention (and so "non attention linear operations" could be interpreted as just mlp, which is not correct). Maybe we should provide examples or just drop the "non-attention" altogether and just talk about the linear operations.

docs/features/low_precision_training/introduction/introduction.rst
```diff
@@ -0,0 +1,277 @@
...
+- **Attention computations**: run in high precision by default (some recipes allow low precision as an option).
+- **Other operations** (layer normalization, softmax, etc.): run in high precision.
+
+Within high-precision operations, there are two categories:
+
+- **Configurable precision**: most operations run in parameter precision (FP32/BF16/FP16) or the precision specified by ``torch.autocast``.
+- **Fixed FP32 precision**: some operations, or parts of operations—such as the division in layernorm—always run in FP32, regardless of other settings.
+
+.. raw:: html
+   :file: img/mixed_precision_operations.svg
```
ptrendx, line 255: I'm not sure how I feel about splitting the attention into 3 blocks in this picture, especially the block saying "attn * V" is not great - those should be scores instead of "attn" I think?

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+and the j-th column of ``B`` (elements ``B_0j, B_1j, ...``). Thus, ``A`` is accessed rowwise 
+and ``B`` is accessed columnwise.
+
+For NT (non-transposed, transposed) multiplication ``C = A * B^T``, the formula changes to ``C_ij = sum_k(A_ik * B_jk)``.
+Now we iterate over the i-th row of ``A`` and the j-th row of ``B`` (elements ``B_j0, B_j1, ...``).
+Both tensors are accessed rowwise.
+
+The figure below illustrates these access patterns:
+
+.. figure:: img/gemm_access_pattern.svg
```
ptrendx, line 40: I think it is better to use square matrices here since as-is the second picture does not make sense - the dimensions of A and B don't match.

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+
+2. *Computation of quantized tensor in rowwise usage in forward pass and fused quantization to produce columnwise usage in backward pass*. 
+
+   This is usually slower than the previous one, since high precision tensor needs to be read twice.
+   It is used for example when high precision tensor is gathered both in forward and in backward 
+   and quantized tensor gather is not implemented for such recipe.
+
+3. *Computation of quantized tensor in rowwise usage in forward pass and transpose to columnwise usage in backward pass*. 
+
+   This is not possible for all recipes, but if it is possible it is more memory efficient than Option 1.
```
ptrendx:
```suggestion
   It is more memory efficient than Option 1, but not all recipes can utilize it (otherwise
   the quantization accuracy would drop due to double quantization errors).
```

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+
+Transformer Engine uses the best possible fusion internally, so users do not need to worry about the details.
```
ptrendx: That is pretty informal.

```suggestion
Transformer Engine chooses the best possible fusion internally taking the recipe and the operation into account.
```

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+We showcase this issue in the documentation to understand memory consequences of different fusion scenarios.
```
ptrendx: I would remove this sentence.

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+
+Memory usage
+------------
+
+This section discusses memory usage in low precision training. 
+Contrary to intuition, FP8 training does not always reduce memory compared to BF16/FP16.
+
+*Master weights*
+
+Transformer Engine stores weights in high precision and quantizes them to low precision before each GEMM.
```
ptrendx:
```suggestion
Transformer Engine by default stores weights in high precision and quantizes them to low precision before each GEMM.
```

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+Moreover, one can specify the precision of the weights stored in the model - if this can be FP32 or 
+BF16/FP16 -- or do not store high precision weights in the model at all. There are multiple scenarios to consider,
```
ptrendx:
```suggestion
Moreover, one can specify which high precision should be used to store the weights in the
model (FP32/BF16/FP16) -- or choose not to store high precision weights in the model at all.
There are multiple scenarios to consider,
```

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+three of them are listed below:
+
+1. model weights are in FP32, quantized to low precision before each GEMM,
+2. model weights are in BF16/FP16, quantized to low precision before each GEMM, master weights in optimizer are in FP32.
+3. model weight are stored directly in low precision, and master weights in optimizer are in FP32.
```
ptrendx:
```suggestion
3. model weights are stored directly in low precision, and master weights in optimizer are in FP32.
```

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+
+Note that all these scenarios may have different memory footprints.
```
ptrendx:
```suggestion
Note that each of these scenarios may have different memory footprint.
```

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+
+*Activations saved for backward*
+
+Unlike weights, activations do not require a high precision copy for optimizer updates. 
+As shown in Table 2, the input needs rowwise usage in forward and columnwise usage 
+for weight gradient computation in backward — so it must be saved between passes.
+
+The memory impact depends on which scenario from Figure 3.
```
ptrendx: Incomplete sentence?

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+Additionally, on architectures where rowwise and columnwise share the same memory layout
```
ptrendx:
```suggestion
Additionally, on architectures where rowwise and columnwise tensors share the same memory layout 
```

ptrendx: or usages?

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+         .. literalinclude:: memory_usage_2_pytorch.out
+            :language: text
+            :start-after: # START_MEMORY_USAGE_2
+            :end-before: # END_MEMORY_USAGE_2
+      
+      Total memory usage is ``2 MB (weight) + 1 MB (weight in FP8) + 2 MB (input) + 1 MB (input in FP8) + 2 MB (output) = 8 MB``.
+      
+      **3. FP8 training with model weights stored directly in low precision**
+
+      When model weights are stored directly in low precision, master weights are not needed.
```
ptrendx: Master weights are not needed? This sentence is probably wrong - you should say that this is because we are doing inference or something?

Also, the numbers are slightly misleading since input in high precision is saved only because of the user script rather than the model saving it for backward. So a better example would delete input from the user script and show that in this case the memory usage is actually lower than in BF16.

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+Fused layers
+------------
+
+
+Transformer Engine provides fused layers such as ``LayerNormLinear`` and ``LayerNormMLP`` 
+that enable kernel fusion optimizations. One key optimization is fusing layer normalization 
+with quantization.
+
+In a typical Transformer architecture, LayerNorm precedes a Linear layer. Without fusion, 
+the LayerNorm outputs in FP32, and the Linear layer must then quantize this input before
```
ptrendx:
```suggestion
the LayerNorm outputs in high precision, and the Linear layer must then quantize this input before 
```

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+         
+         JAX does not currently support storing model weights directly in FP8 format
+         like PyTorch's ``quantized_model_init``. Weights are always stored in high precision
+         (BF16/FP32) and quantized to FP8 during computation.
+
+Fused layers
+------------
+
+
+Transformer Engine provides fused layers such as ``LayerNormLinear`` and ``LayerNormMLP``
```
ptrendx:
```suggestion
Transformer Engine provides fused layers such as ``LayerNormLinear`` (``LayerNormDenseGeneral`` in JAX) and ``LayerNormMLP`` 
```

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+         </div>
+
+      .. literalinclude:: fused_layers_pytorch.py
+         :language: python
+         :start-after: # START_FUSED_LAYERS
+         :end-before: # END_FUSED_LAYERS
+      
+      The fused ``LayerNormLinear`` layer is particularly efficient in FP8 training because 
+      it avoids an intermediate quantization step. The LayerNorm output is directly quantized 
+      for the GEMM operation, reducing memory bandwidth and improving performance.
```
ptrendx:
```suggestion
      for the GEMM operation, reducing memory movement and improving performance.
```

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+         </div>
+
+      .. literalinclude:: fused_layers_jax.py
+         :language: python
+         :start-after: # START_FUSED_LAYERS
+         :end-before: # END_FUSED_LAYERS
+      
+      The fused ``LayerNormDenseGeneral`` layer is particularly efficient in FP8 training because 
+      it avoids an intermediate quantization step. The LayerNorm output is directly quantized 
+      for the GEMM operation, reducing memory bandwidth and improving performance.
```
ptrendx:
```suggestion
      for the GEMM operation, reducing memory movement and improving performance.
```

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,480 @@
...
+the interaction between communication and low precision computation.
+
+Recall that each Linear layer involves six tensors: weight, input, output, and their gradients. 
+Of these, output and gradients are returned in high precision, and weights are generally not 
+communicated (except in FSDP, which is outside the scope of this section). This leaves two 
+tensors where low precision communication matters: **input** and **output gradient**.
+
+For sequence parallelism, TE supports all-gather of quantized tensors. This provides several benefits:
+
+1. *Reduced memory* — no need to store high precision tensors for backward pass.
```
ptrendx:
```suggestion
1. *Reduced memory usage* — no need to store high precision tensors for backward pass.
```

docs/features/low_precision_training/performance_considerations/save_original_input_pytorch.out
```diff
@@ -0,0 +1,12 @@
+/usr/local/lib/python3.12/dist-packages/torch/library.py:356: UserWarning: Warning only once for all operators,  other operators may also be overridden.
```
ptrendx: Do we need this warning :-)?

docs/features/low_precision_training/fp8_current_scaling/fp8_current_scaling.rst
```diff
@@ -0,0 +1,178 @@
+..
+    Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+    See LICENSE for license information.
+
+FP8 Current Scaling
+===================================
+
+FP8 current scaling is the simplest low precision recipe provided by Transformer Engine.
```
ptrendx:
```suggestion
FP8 current scaling recipe is the simplest low precision recipe provided by Transformer Engine. 
```

docs/features/low_precision_training/fp8_current_scaling/fp8_current_scaling.rst
```diff
@@ -0,0 +1,178 @@
...
+   :file: img/fp8_formats.svg
+
+*Figure 1: Structure of the floating point datatypes. All of the values shown (in FP16, BF16, FP8 E4M3 and FP8 E5M2) are the closest representations of value 0.3952.*
+
+
+**E4M3 and E5M2 usage in training**
+
+By default, Transformer Engine uses a hybrid approach:
+
+* *Forward pass* - activations and weights require more precision, so E4M3 datatype is best used.
```
ptrendx:
```suggestion
* *Forward pass* - activations and weights require more precision, so E4M3 datatype is used to store them.
```

docs/features/low_precision_training/fp8_current_scaling/fp8_current_scaling.rst
```diff
@@ -0,0 +1,178 @@
...
+* *Backward pass* - gradients are less susceptible to precision loss but require higher dynamic range, so E5M2 datatype is preferred. 
+
+The user can configure this behavior via the ``fp8_format`` parameter of the recipe.
+
+
+Scaling factors
+---------------
+
+
+FP8's limited dynamic range is insufficient for many tensors.
```
ptrendx:
```suggestion
Limited dynamic range of FP8 datatype is insufficient for many tensors. 
```

docs/features/low_precision_training/fp8_current_scaling/fp8_current_scaling.rst
```diff
@@ -0,0 +1,178 @@
...
+
+The user can configure this behavior via the ``fp8_format`` parameter of the recipe.
+
+
+Scaling factors
+---------------
+
+
+FP8's limited dynamic range is insufficient for many tensors. 
+To address this, scaling factors are used. In FP8 Current Scaling there is one **FP32** scale factor per tensor.
```
ptrendx:
```suggestion
To address this, values in the tensor are scaled. FP8 Current Scaling recipe uses one **FP32** scale factor per tensor.
```

docs/features/low_precision_training/fp8_current_scaling/fp8_current_scaling.rst
```diff
@@ -0,0 +1,178 @@
...
+    x = x_fp8 * s
+
+where
+
+* ``x_fp8`` is the FP8 value (E4M3 or E5M2),
+* ``s`` is a global **FP32** scaling factor applied to the entire tensor.
+
+**FP8 Current Scaling quantization**
+
+Let's look more closely at how quantization to FP8 with scaling factor is implemented in
```
ptrendx:
```suggestion
Let's take a closer look at how quantization to FP8 with scaling factor is implemented in
```

docs/features/low_precision_training/fp8_current_scaling/fp8_current_scaling.rst
```diff
@@ -0,0 +1,178 @@
...
+the FP8 Current Scaling recipe.
+
+.. raw:: html
+   :file: img/fp8_scaling_concept.svg
+
+*Figure 3: Quantization to FP8 consists of amax computation, scaling to fit the FP8 range and casting to the respective FP8 format.*
```
ptrendx:
```suggestion
*Figure 3: Quantization to FP8 consists of amax (absolute maximum) computation, scaling to fit the FP8 range and casting to the respective FP8 format.*
```

docs/features/low_precision_training/fp8_current_scaling/fp8_current_scaling.rst
```diff
@@ -0,0 +1,178 @@
...
+
+Quantization to FP8 consists of 3 steps:
+
+1. Computation of the absolute maximum value of the tensor - we refer to it as ``amax``.
+2. Applying the scaling factor of ``fp8_max / amax`` to the tensor, to fit it into the FP8 range
+3. Casting into the respective FP8 format using *Round To Nearest Even (RTNE)*. Values round to the nearest representable FP8 value. When exactly halfway between two values, rounds to the one with even mantissa to minimize systematic bias.
```
ptrendx, line 67: Not sure if we need to explain how round to nearest even works here, but I'm ok with leaving this if you think it is useful.

docs/features/low_precision_training/fp8_current_scaling/fp8_current_scaling.rst
```diff
@@ -0,0 +1,178 @@
...
+   :file: img/fp8_cast_process.svg
+
+*Figure 4: FP8 quantization with current scaling recipe - two tensor reads are needed, one to compute amax and one to apply the scaling factor and cast to FP8.*
+
+Hardware support
+----------------
+
+The Hopper architecture introduced FP8 support in Tensor Cores, enabling efficient low-precision computation. 
+Tensor Cores support every combination of E4M3 and E5M2 formats as inputs, allowing flexible precision choices for different operands.
+The inputs to an FP8 Tensor Core operation consist of chunks of FP8 tensors along with their corresponding scaling factors.
```
ptrendx: That is not actually correct - only Blackwell introduced passing scaling factors to Tensor Cores. Hopper applies scaling via alpha in the epilogue of the GEMM. I don't think we need to go into details about it, so suggesting getting rid of this sentence enitrely.


```suggestion
```

docs/features/low_precision_training/fp8_current_scaling/fp8_current_scaling.rst
ptrendx: Also, I don't like the "FP8 range min" in the picture - this is ambiguous, considering that we are mostly talking about the dynamic range in the text - in that context, it should be something like "- FP8 range max". Maybe we could do something like:
 - first picture: add "Original range" spanning all values
 - second picture: rather than having a single arrow show a trapezoid that compacts that original range to the FP8 range
 - third picture: no changes

docs/features/low_precision_training/fp8_current_scaling/fp8_current_scaling.rst
```diff
@@ -0,0 +1,178 @@
...
+The Tensor Core performs the matrix multiplication in FP8 precision and produces output in higher precision (FP16, BF16, or FP32).
+
+.. raw:: html
+   :file: img/fp8_tensor_core.svg
+
+*Figure 5: FP8 Tensor Cores process two input tensors (A and B) with their respective scaling factors and perform matrix multiplication to accumulate higher-precision output.*
```
ptrendx: Yeah, so as noted in the previous comment it is not quite how it works, but maybe it is close enough that it is fine?

docs/features/low_precision_training/fp8_current_scaling/fp8_current_scaling.rst
```diff
@@ -0,0 +1,178 @@
...
+
+
+Distributed training 
+--------------------
+
+**All-gather of columnwise tensors**
+
+Supported for Blackwell and later, since rowwise and columnwise tensors share the same memory layout.
+For Hopper and Ada, all-gather of transposed FP8 tensors is not supported. 
+The rowwise tensor is gathered and then it is transposed to columnwise tensor.
```
ptrendx: I don't think this section makes sense to be honest - people should not really care if we support allgather of the columnwise tensors, but rather if we support allgather in the low precision (and we do, for all architectures).

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
```diff
@@ -0,0 +1,172 @@
+..
+    Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+    See LICENSE for license information.
+
+FP8 Delayed Scaling
+===================================
+
+FP8 Delayed Scaling estimates scaling factors from historical amax values rather than computing them
```
ptrendx:
```suggestion
FP8 Delayed Scaling recipe estimates scaling factors from historical amax values rather than computing them
```

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
```diff
@@ -0,0 +1,172 @@
...
+for each tensor. This reduces tensor reads per quantization from two to one, improving memory efficiency.
```
ptrendx:
```suggestion
for each tensor. Compared to Current Scaling recipe, this reduces tensor reads per quantization from two to one, improving memory efficiency.
```

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
```diff
@@ -0,0 +1,172 @@
...
+
+Both this recipe and :doc:`FP8 Current Scaling <../fp8_current_scaling/fp8_current_scaling>` use
```
ptrendx:
```suggestion
Both this and :doc:`FP8 Current Scaling <../fp8_current_scaling/fp8_current_scaling>` recipe use 
```

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
```diff
@@ -0,0 +1,172 @@
...
+    See LICENSE for license information.
+
+FP8 Delayed Scaling
+===================================
+
+FP8 Delayed Scaling estimates scaling factors from historical amax values rather than computing them
+for each tensor. This reduces tensor reads per quantization from two to one, improving memory efficiency.
+
+Both this recipe and :doc:`FP8 Current Scaling <../fp8_current_scaling/fp8_current_scaling>` use 
+the same FP8 formats (E4M3/E5M2) with one float32 scaling factor per tensor.
```
ptrendx:
```suggestion
the same FP8 formats (E4M3/E5M2) with one FP32 scaling factor per tensor. 
```

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
```diff
@@ -0,0 +1,172 @@
...
+from historical amax values - hence *delayed* (using past values) versus *current* (using present values).
+
+The quantization process works as follows:
+
+1. **Compute scaling factor from history** (no tensor read needed):
+   The scaling factor is derived from stored ``amax_history`` using the formula:
+   
+   ``scaling_factor = FP8_MAX / amax``
+   
+   where ``amax`` is computed from history using either ``max`` (default) or ``most_recent`` algorithm.
```
ptrendx:
```suggestion
   where ``amax`` is computed from history using either ``max`` (maximum over window, default) or ``most_recent`` algorithm.
```

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
```diff
@@ -0,0 +1,172 @@
...
+Position 0 serves as a staging area for the current amax, while positions 1 to N-1 
+store the history from oldest to newest. Each quantization writes the observed amax 
+to position 0, and after the pass completes, the history is rotated:
+
+.. code-block:: text
+
+   Before rotation: [amax_N, amax_1, amax_2, ..., amax_N-1]   (amax_N = current, amax_1 = oldest)
+   After rotation:  [0,      amax_2, ..., amax_N-1, amax_N]   (amax_1 dropped, amax_N appended)
+
+The effective history length is ``amax_history_len - 1`` since position 0 is reserved
```
ptrendx: I don't believe that is true statement - scaling factor for the future iteration is calculated before the rotation, so it sees all N values (from 1 to N).

ptrendx: (and if that is not the case then I would say this is a bug actually :-) )

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
```diff
@@ -0,0 +1,172 @@
...
+      - Forward: shape ``(amax_history_len, num_gemms * 3)`` — three FP8 tensors per GEMM (input, weight, output)
+      - Backward: shape ``(amax_history_len, num_gemms * 2)`` — two FP8 tensors per GEMM (grad_output, grad_input)
+      
+      During the first forward pass, modules register their ``amax_history`` tensors 
+      to a **global buffer** associated with the autocast context. When the context exits,
+      a single CUDA kernel processes all registered tensors at once - performing both 
+      amax reduction across GPUs and history rotation.
+      
+      This batched approach (one kernel for all tensors instead of one kernel per tensor)
+      minimizes kernel launch overhead.
```
ptrendx: I don't know if this is really needed here - that would be a good thing to put in a developer documentation (maybe comment in the code?) but not necessarily in the user-facing documentation.

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
```diff
@@ -0,0 +1,172 @@
...
+
+   .. tab:: JAX
+
+      Each quantizer maintains its own ``amax_history`` as a Flax variable with shape ``(amax_history_len,)``.
+      There is no global buffer - each quantizer updates independently.
+      
+      The rotation is performed per-quantizer using ``jnp.roll``:
```
ptrendx: Same comment here (developer vs user audience).

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
```diff
@@ -0,0 +1,172 @@
...
+         :language: python
+         :start-after: # START_DELAYED_SCALING_EXAMPLE
+         :end-before: # END_DELAYED_SCALING_EXAMPLE
+
+
+Distributed Training
+--------------------
+
+Since FP8 Delayed Scaling uses the same data formats as FP8 Current Scaling,
+transpose gather is not supported. However, amax reduction works slightly differently in different frameworks.
```
ptrendx: What do you mean by "transpose gather is not supported"? Just like int current scaling, I think what is more relevant is the fact that allgather is performed in low precision.

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
```diff
@@ -0,0 +1,172 @@
...
+         :language: python
+         :start-after: # START_AMAX_REDUCTION_EXAMPLE
+         :end-before: # END_AMAX_REDUCTION_EXAMPLE
+
+      In data parallel training, some modules may not execute on certain ranks 
+      (e.g., MoE experts that receive no tokens). This is handled as follows:
+      
+      - **First iteration**: All modules must execute on all ranks to register 
+        their ``amax_history`` tensors in the global buffer. Mismatched registration
+        causes the ``all_reduce`` to hang due to different tensor sizes across ranks.
```
ptrendx:
```suggestion
        would cause the ``all_reduce`` to hang due to different tensor sizes across ranks.
```

docs/features/low_precision_training/fp8_blockwise_scaling/fp8_blockwise_scaling.rst
```diff
@@ -0,0 +1,237 @@
+..
+    Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+    See LICENSE for license information.
+
+FP8 Blockwise Scaling
+===================================
+
+FP8 Blockwise Scaling is inspired by the quantization scheme used to train the `DeepSeek-v3 model <https://arxiv.org/abs/2412.19437>`__ –
```
ptrendx:
```suggestion
FP8 Blockwise Scaling recipe is inspired by the quantization scheme used to train the `DeepSeek-v3 model <https://arxiv.org/abs/2412.19437>`__ –
```

docs/features/low_precision_training/fp8_blockwise_scaling/fp8_blockwise_scaling.rst
```diff
@@ -0,0 +1,237 @@
...
+* ``s_block`` is a local **FP32** scaling factor shared by a block of elements.
+
+
+.. raw:: html
+   :file: img/combined_scaling.svg
+
+*Figure 1. Top: Comparison of standard FP8 scaling (left) using a single scaling factor per tensor versus 
+FP8 blockwise scaling in 1 dimension (right) using multiple scaling factors, one per block of 128 elements.
+Bottom: FP8 blockwise scaling in 2 dimensions where each 128×128 block in the data tensor has a corresponding
+scaling factor, providing fine-grained spatial control over quantization precision.*
```
ptrendx: "providing fine-grained spatial control over quantization precision" - ? This sounds LLM-y and does not really provide much information I think.

docs/features/low_precision_training/fp8_blockwise_scaling/fp8_blockwise_scaling.rst
```diff
@@ -0,0 +1,237 @@
...
+
+**FP8 format**
+
+Unlike FP8 Current/Delayed Scaling, E4M3 is used by default for both forward and backward passes.
+Previous recipes used E5M2 for gradients due to its higher dynamic range,
```
ptrendx:
```suggestion
Tensor-scaled recipes used E5M2 for gradients due to its higher dynamic range,
```

docs/features/low_precision_training/fp8_blockwise_scaling/fp8_blockwise_scaling.rst
```diff
@@ -0,0 +1,237 @@
...
+but with multiple scaling factors per tensor, E4M3 is usually sufficient.
```
ptrendx:
```suggestion
but with multiple scaling factors per tensor the dynamic range requirement is lowered, so E4M3 is usually sufficient.
```

docs/features/low_precision_training/fp8_blockwise_scaling/fp8_blockwise_scaling.rst
```diff
@@ -0,0 +1,237 @@
...
+
+* the tensor must have at least 2 dimensions,
+* the last dimension must be divisible by 128,
+* the product of all dimensions except the last must be divisible by 128.
+
+**Scaling factors**
+
+Scaling factors are stored as 32-bit floating point numbers.
+By default, they are constrained to powers of 2 (utilizing the 8 exponent bits of FP32).
+This constraint can be relaxed by setting the environment variable ``NVTE_FP8_BLOCK_SCALING_FP32_SCALES=1``.
```
ptrendx: We should note that this is applicable only to Hopper - on Blackwell only powers of 2 are supported.

docs/features/low_precision_training/fp8_blockwise_scaling/fp8_blockwise_scaling.rst
```diff
@@ -0,0 +1,237 @@
...
+
+Each block's scaling factor is computed through the following steps:
+
+1. Find the maximum absolute value (``amax_block``) across all elements in the block
+   (128 consecutive values for 1D blocks, or 128×128 values for 2D blocks).
+2. Calculate ``s_block = max_fp8 / amax_block``, where ``max_fp8`` is
+   the maximum representable value in the FP8 format (448 for E4M3, 57344 for E5M2).
+3. If the power-of-2 constraint is enabled, round down to the nearest power of 2
```
ptrendx, line 92: We should double check that - in MXFP8 we do rounding up (and that is better for accuracy), so we should probably revisit this if we round down here.

ptrendx: Although depending on what exactly is mentioned here (encode scale vs decode scale) rounding up or rounding down could mean the same thing actually - anyway, we should double check.

docs/features/low_precision_training/fp8_blockwise_scaling/fp8_blockwise_scaling.rst
```diff
@@ -0,0 +1,237 @@
...
+* *Rowwise access*: 1 scaling factor per 128 consecutive elements in a row.
+* *Columnwise access*: 1 scaling factor per 128 consecutive elements in a row of the transposed tensor,
+  corresponding to 128 consecutive elements in a column of the original tensor.
+
+For 2D scaling, each 128×128 tile has one scaling factor regardless of access direction.
+
+This is illustrated below:
+
+.. raw:: html
+   :file: img/transpose_handling.svg
```
ptrendx, line 114: Considering that we said earlier that things need to be padded to 128 then having descriptions like 120x240 tensor doesn't make sense. Those values do not make sense also because they are too small considering the picture (where each block is 128 wide, so if it really was just 240 then it should contain only 2 blocks)

docs/features/low_precision_training/fp8_blockwise_scaling/fp8_blockwise_scaling.rst
```diff
@@ -0,0 +1,237 @@
...
+
+Unlike FP8 Current/Delayed Scaling, transposing a 1D quantized tensor is not supported.
+Rowwise and columnwise blocks cover different sets of elements, so their scaling factors differ.
+Both versions must be quantized separately from the high-precision source.
+
+For 2D scaling, columnwise data can be created from rowwise data by transposing 
+both the quantized data and the scaling factors. Each 128×128 block covers the same 
+elements regardless of access direction, so the scaling factors remain valid.
+
+Swizzle of scaling factors
```
ptrendx, line 189: I'm also not sure how relevant this section is for the general audience to be honest - it is a good source of info for the developers but user does not really need to worry about this I think.

docs/features/low_precision_training/mxfp8/mxfp8.rst
```diff
@@ -0,0 +1,199 @@
...
+The representation of an FP8 tensor element ``x`` in MXFP8 precision is given by:
+
+.. code-block:: python
+
+    x = x_fp8 * s_block
+
+where
+
+* ``x_fp8`` is the FP8 value in E4M3 format,
+* ``s_block`` is a local **E8M0** scaling factor shared by a block of 32 elements.
```
ptrendx, line 28: We should say what E8M0 is, just for completeness

docs/features/low_precision_training/mxfp8/mxfp8.rst
```diff
@@ -0,0 +1,199 @@
...
+
+
+**FP8 format**
+
+Like FP8 Blockwise Scaling, E4M3 is used by default for both forward and backward passes.
+The finer-grained scaling provides sufficient dynamic range without requiring the E5M2 format.
+The ``fp8_format`` parameter also supports ``HYBRID`` mode (E4M3 for forward, E5M2 for backward).
```
ptrendx, line 36: Really? cuBLAS supports E5M2 MXFP8 gemms? We should test that

docs/features/low_precision_training/mxfp8/mxfp8.rst
```diff
@@ -0,0 +1,199 @@
...
+* the last dimension must be divisible by 32,
+* the product of all dimensions except the last must be divisible by 32.
+
+
+**Scaling factors**
+
+Scaling factors are stored as E8M0 (8 exponent bits, 0 mantissa bits), which inherently represents
+powers of 2. This differs from FP8 Blockwise Scaling, which uses 32-bit floating point numbers
+optionally constrained to powers of 2. Note that FP32 also has 8 exponent bits, so the representable
+ranges are similar when the power-of-2 constraint is enabled.
```
ptrendx: "similar"? I think they are exactly the same - what is the difference?

docs/features/low_precision_training/mxfp8/mxfp8.rst
```diff
@@ -0,0 +1,199 @@
...
+Handling transposes
+-------------------
+
+Blackwell architecture supports multiple FP8 GEMM layouts (TN, NT, NN), so columnwise usage
+does not require explicit transposition. However, rowwise and columnwise quantizations are different:
+
+- *Rowwise* - 1 scaling factor per 32 consecutive elements along a row (1×32 blocks).
+- *Columnwise* - 1 scaling factor per 32 consecutive elements along a column (32×1 blocks).
+
+Because the scaling factor blocks have different orientations, rowwise and columnwise MXFP8 tensors
```
ptrendx:
```suggestion
Since the scaling factor blocks have different orientations, rowwise and columnwise MXFP8 tensors
```

docs/features/low_precision_training/mxfp8/mxfp8.rst
```diff
@@ -0,0 +1,199 @@
...
+are numerically different — one cannot derive one from the other. Both must be quantized
+independently from full-precision data.
```
ptrendx:
```suggestion
independently from the full-precision data.
```

docs/features/low_precision_training/mxfp8/mxfp8.rst
```diff
@@ -0,0 +1,199 @@
...
+
+.. raw:: html 
+   :file: img/mxfp8_row_col.svg
+
+*Figure 2. MXFP8 rowwise vs columnwise quantization layout.*
+
+
+Swizzling scaling factors
```
ptrendx: Same comment as in the Blockwise scaling. Actually, maybe we could have a special subsection "Developer notes" or something like that to put this in?

docs/features/low_precision_training/mxfp8/mxfp8.rst
```diff
@@ -0,0 +1,199 @@
...
+-------------------------
+
+Like :doc:`FP8 Blockwise Scaling <../fp8_blockwise_scaling/fp8_blockwise_scaling>`, MXFP8 uses different data layouts for communication and computation.
+MXFP8 GEMMs require scaling factors in a specific hardware layout
+(see `cuBLAS documentation <https://docs.nvidia.com/cuda/cublas/index.html#block-scaling-factors-layout>`__).
+The conversion to this GEMM-ready layout is called *swizzling*. Because swizzled scaling factors
+cannot be communicated across devices, Transformer Engine performs swizzling after any required
+communication, just before each GEMM operation.
```
ptrendx: We actually just merged the support from @timmoon10 to handle swizzling in quantization as well - this text is pretty much correct still, just pointing it out.

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -0,0 +1,261 @@
...
+
+
+.. raw:: html
+   :file: img/nvfp4_hierarchical_scaling.svg
+
+*Figure 2. NVFP4 hierarchical scaling structure showing the combination of block-level and global scaling factors.*
+
+This hierarchical structure uses fine-grained block scaling 
+to adapt to local magnitude variations and global scaling 
+to handle the overall dynamic range.
```
ptrendx: "global scaling to handle the overall dynamic range" seems wrong to me. Handling of the (tensor) dynamic range is done by the block scaling factors and (block-level) dynamic range is handled by the FP4 values themselves. The global scaling factor has similar role to the loss scale in traiditonal AMP - it is a "bias" that moves the whole range to the center, so that the values can be represented by E4M3 * E2M1 combo.

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -0,0 +1,261 @@
...
+``disable_stochastic_rounding=True`` in the recipe configuration.
+
+
+Random Hadamard Transform
+--------------------------
+
+Random Hadamard Transform (RHT) applies an orthogonal rotation to the tensor **before quantization**,
+smoothing outliers in the tensor distributions and making them easier to represent accurately in NVFP4.
+RHT is applied to columnwise quantization of inputs and gradients, which are operands
+for the **wgrad GEMM**. This GEMM – according to the paper – is particularly sensitive
```
ptrendx:
```suggestion
for the **wgrad GEMM**. This GEMM is particularly sensitive
```

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -0,0 +1,261 @@
...
+to quantization errors, hence the additional outlier smoothing.
+RHT is supported only for BF16 inputs/gradients; other dtypes will raise an error.
```
ptrendx:
```suggestion
RHT is supported only for BF16 inputs/gradients.
```

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -0,0 +1,261 @@
...
+      .. literalinclude:: jax_nvfp4_example.py
+         :language: python
+         :start-after: # START_NVFP4_EXAMPLE
+         :end-before: # END_NVFP4_EXAMPLE
+
+
+Supported devices
+-----------------
+
+Blackwell and later (SM 10.0+)
```
ptrendx: This is more complicated - e.g. stochastic rounding hardware instruction does not exist on SM12 (which is also technically Blackwell), so I would say here 10.x (10.0/10.3) instead for training and SM10+ for inference.

docs/features/low_precision_training/introduction/introduction.rst
pggPL: I prepared some modification to describe what you mean.

docs/features/low_precision_training/introduction/introduction.rst
pggPL: I added the description.

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
pggPL: i think it makes sense, changed picture a little, we may discuss it.

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
pggPL: it seems to be docs bug, everything in code is ok

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
pggPL: I made it simpler, but I think that the user may want to look at the profile and he should understand where this rotation happens.

docs/features/low_precision_training/fp8_delayed_scaling/fp8_delayed_scaling.rst
pggPL: same answer as above

docs/features/low_precision_training/fp8_blockwise_scaling/fp8_blockwise_scaling.rst
pggPL: it seems that it is equivalent

docs/features/low_precision_training/mxfp8/mxfp8.rst
pggPL: one support +inf, the other not

docs/features/low_precision_training/mxfp8/mxfp8.rst
pggPL: it works

docs/features/low_precision_training/fp8_delayed_scaling/jax_delayed_scaling_distributed_example.py
```diff
@@ -0,0 +1,15 @@
...
+
+# START_AMAX_REDUCTION_EXAMPLE
+import transformer_engine.jax as te
+from transformer_engine.common.recipe import DelayedScaling
+
+# Amax reduction scope is managed internally
+recipe = DelayedScaling(reduce_amax=True)  # Must be True in JAX
+
+with te.autocast(enabled=True, recipe=recipe):
+    output = layer.apply(params, inp)
```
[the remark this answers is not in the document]
pggPL: this examlpe will not be run, just a visualizaiton

docs/features/low_precision_training/fp8_delayed_scaling/jax_delayed_scaling_example.py
```diff
@@ -0,0 +1,51 @@
...
+    or "RTX 5" in cc
+    or "Ada" in cc
+    or "L40" in cc
+    or "H100" in cc
+    or "H200" in cc
+    or "GH" in cc
+    or "B100" in cc
+    or "B200" in cc
+    or "GB" in cc
+), "This example requires SM89 (Ada) or newer"
```
jberchtold-nvidia: Should we use this helper that gets the compute capability of the device directly rather than matching based on the name. What do you think?
https://github.com/NVIDIA/TransformerEngine/blob/69636a08171d162a223ae3a01e4b36902b64dba1/transformer_engine/jax/quantize/device_utils.py#L23

docs/features/low_precision_training/introduction/autocast_jax.py
```diff
@@ -0,0 +1,95 @@
...
+with te.autocast(enabled=True, recipe=recipe):
+    layer = TransformerLayer(
+        hidden_size=1024,
+        mlp_hidden_size=4096,
+        num_attention_heads=16,
+    )
+
+    init_key, dropout_key = jax.random.split(jax.random.PRNGKey(0))
+    x = jax.random.normal(init_key, (32, 128, 1024), dtype=jnp.bfloat16)
+    params = layer.init({"params": init_key, "dropout": dropout_key}, x)
```
jberchtold-nvidia: For JAX model init can you use the name `var_collect` instead of `params`? It is not always just model params. For recipes like DelayedScaling, it becomes  `{"params": <model params>, "<state_key_set_by_user_config>": <delayed scaling state>}`. Thanks!

docs/features/low_precision_training/introduction/autocast_jax.py
```diff
@@ -0,0 +1,95 @@
...
+
+# END_AUTOCAST_SEQUENTIAL
+
+
+# START_AUTOCAST_NESTED
+
+outer_recipe = DelayedScaling(fp8_format=Format.E4M3)
+inner_recipe = DelayedScaling(fp8_format=Format.HYBRID)
+
+with te.autocast(enabled=True, recipe=outer_recipe):
```
jberchtold-nvidia, line 66: Can you add a note here to be careful with JIT when dealing with Python context such as `autocast` and mention something like "for finer-grained control in JAX, the recommended approach is directly passing the recipe, more details can be found here" and link to https://github.com/NVIDIA/TransformerEngine/blob/main/docs/examples/te_jax_integration.ipynb

I'm pretty sure this example will work, even if it were to be JIT'd. But JIT+Python contexts is a bit tricky since many things get deconstructed and reconstructed in different passes in JAX, (such as vjp rules) and it's easy to let a Python context accidentally go out of scope during a later JIT pass and get confusing behavior.

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
```diff
@@ -0,0 +1,481 @@
...
+         </div>
+      
+      .. container:: program-output
+      
+         .. literalinclude:: memory_usage_2_jax.out
+            :language: text
+            :start-after: # START_MEMORY_USAGE_2
+            :end-before: # END_MEMORY_USAGE_2
+      
+      In JAX, unlike PyTorch, FP8 weights are not cached between forward passes.
```
jberchtold-nvidia: Can you elaborate slightly to say the model weights memory usage doesn't change with JAX currently, but the memory usage of saved activations and other intermediate results is reduced with quantization, plus the perf benefit of fp8 GEMM

docs/features/low_precision_training/fp8_blockwise_scaling/jax_blockwise_scaling_example.py
```diff
@@ -0,0 +1,43 @@
...
+# START_BLOCKWISE_SCALING_EXAMPLE
+
+import jax
+import jax.numpy as jnp
+import transformer_engine.jax as te
+from transformer_engine.jax.flax import DenseGeneral
+from transformer_engine.common.recipe import Float8BlockScaling
+
+# Create FP8 Blockwise Scaling recipe
+recipe = Float8BlockScaling(
```
jberchtold-nvidia: Does TE/JAX support Float8BlockScaling? Afaik we haven't implemented support for it so this shouldn't work, but if trying this doesn't give a clear error message that it wasn't supported, lmk and I can look until fixing it. Thanks!

docs/features/low_precision_training/nvfp4/jax_nvfp4_example.py
```diff
@@ -0,0 +1,39 @@
...
+    key = jax.random.PRNGKey(0)
+    x = jax.random.normal(key, (32, 128, 1024), dtype=jnp.bfloat16)
+    params = layer.init(key, x)
+
+    # Forward and backward pass
+    def loss_fn(params):
+        output = layer.apply(params, x)
+        return output.sum()
+
+    loss, grads = jax.value_and_grad(loss_fn)(params)
```
jberchtold-nvidia: Does running this give any errors? For NVFP4, it should be required to specify an additional Flax RNG for `sr_rng` for stochastic rounding, like here:  https://github.com/NVIDIA/TransformerEngine/blob/69636a08171d162a223ae3a01e4b36902b64dba1/docs/examples/te_jax_integration.ipynb#L310

jberchtold-nvidia (commented):
Fantastic work! The diagrams and explanations are great and along with the code examples will hopefully make things a lot clearer to users

docs/features/low_precision_training/fp8_delayed_scaling/jax_delayed_scaling_example.py
pggPL: fixed

docs/features/low_precision_training/fp8_blockwise_scaling/jax_blockwise_scaling_example.py
pggPL: oh, I was not running this on hopper and blackwell, so I tested only first 2 recipes and I forgot about that. big thanks, it would be really bad bug in docs!

docs/features/low_precision_training/introduction/autocast_jax.py
pggPL: fixed

docs/features/low_precision_training/nvfp4/jax_nvfp4_example.py
pggPL: fixed

docs/features/low_precision_training/performance_considerations/performance_considerations.rst
pggPL: I changed this section a little, so the user will understand that.

docs/features/low_precision_training/introduction/autocast_jax.py
pggPL: done

docs/features/low_precision_training/fp8_blockwise_scaling/fp8_blockwise_scaling.rst
```diff
@@ -128,8 +132,61 @@ For 2D scaling, columnwise data can be created from rowwise data by transposing
...
+   .. tab:: JAX
+
+      ``Float8BlockScaling`` is **not currently supported** in JAX.
+
+Supported devices
+-----------------
+
+Hopper (SM 9.0)
+
+Blackwell and later (SM >= 10.0) – recipe is emulated with MXFP8. Note that this is done mainly for compatibility, MXFP8 is the preferred recipe on Blackwell.
```
pggPL: - emulated = power of 2 scaling factor only
- remove the compatibility part.

docs/features/low_precision_training/mxfp8/mxfp8.rst
```diff
@@ -197,4 +147,67 @@ Here's how to use MXFP8 recipe in PyTorch and JAX:
 Supported devices
 -----------------
 
-Blackwell and later (SM 10.0+)
\ No newline at end of file
+Blackwell and later (SM 10.0+)
```
pggPL: Comment about SM 12

ptrendx (approved):

PR #2344, opened 2025-11-04
Disable cuDNN attention for known IMA and NaNs

ksivaman:
# Description

Disable cuDNN attention backends for cases with known bugs.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Disable cuDNN backend for versions with a known IMA with cuda graphs (under specific cases).
- Disable cuDNN backend for versions known to produce NaN outputs (under specific cases).

# Checklist:


ksivaman:
Awaiting update from cuDNN FE

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -207,13 +207,14 @@ NVTE_QKV_Format nvte_get_kv_format(NVTE_QKV_Layout qkv_layout);
...
  *  \param[in]     return_max_logit    Whether to produce Max and Sum_Exp, or Stats.
+ *  \param[in]     cuda_graph          Whether cuda graph capture is enabled or not.
  */
 NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
     bool is_training, NVTEDType q_dtype, NVTEDType kv_dtype, NVTE_QKV_Layout qkv_layout,
     NVTE_Bias_Type bias_type, NVTE_Mask_Type attn_mask_type, NVTE_Softmax_Type softmax_type,
     float dropout, size_t num_attn_heads, size_t num_gqa_groups, size_t max_seqlen_q,
     size_t max_seqlen_kv, size_t head_dim_qk, size_t head_dim_v, int64_t window_size_left,
-    int64_t window_size_right, bool return_max_logit);
+    int64_t window_size_right, bool return_max_logit, bool cuda_graph);
```
ptrendx, line 217: This is a breaking change.

ksivaman: Yes, I did note that in the original PR checklist.

This set of attention APIs hasn't been stable and is often broken.

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -407,6 +407,27 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
...
+        (max_seqlen_kv > 1024) && (attn_mask_type != NVTE_Mask_Type::NVTE_CAUSAL_MASK)) {
+      backend = NVTE_Fused_Attn_Backend::NVTE_No_Backend;
+      std::cout << "Warning: Given combination of attention mask (non-causal) and "
+                   "max_seqlen_kv (> 1024) does not support cuda graph capture with "
+                   "fused attention for cuDNN 9.13.* or 9.14.*."
+                   " Please upgrade your cuDNN version if possible."
+                << std::endl;
+    }
+    if ((cudnn_runtime_version == 91500) && is_training &&
+        (qkv_format == NVTE_QKV_Format::NVTE_BSHD) &&
```
cyanguwa: Is it just BSHD or both BSHD and SBHD? Usually they have the same support matrix (and issues).

transformer_engine/common/include/transformer_engine/fused_attn.h
cyanguwa: Can we do the CUDA graph check on the framework side instead?

ksivaman: It is possible, but we do the fused attention specific checks (e.g. related to cuDNN version) in core in order to exhaustively cover all cases. Otherwise there could be situations where the core api returns a different backend than the framework.

ksivaman: The framework backend selection is used for selecting between fused, unfused, and flash attentions

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -407,6 +407,27 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
                    " Please upgrade your cuDNN version if possible."
                 << std::endl;
     }
+    if ((cudnn_runtime_version >= 91300) && (cudnn_runtime_version < 91500) && cuda_graph &&
```
vedaanta: - please remove cuda_graph from this one, and instead add to the next one.
- this only fails with 9.14.0
- and only for FWD psas
- plus, this fails whenever there is no causal/brcm/sliding (meaning cases like padding or no mask at all will work)

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -407,6 +407,27 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
...
+        (max_seqlen_kv > 1024) && (attn_mask_type != NVTE_Mask_Type::NVTE_CAUSAL_MASK)) {
+      backend = NVTE_Fused_Attn_Backend::NVTE_No_Backend;
+      std::cout << "Warning: Given combination of attention mask (non-causal) and "
+                   "max_seqlen_kv (> 1024) does not support cuda graph capture with "
+                   "fused attention for cuDNN 9.13.* or 9.14.*."
+                   " Please upgrade your cuDNN version if possible."
+                << std::endl;
+    }
+    if ((cudnn_runtime_version == 91500) && is_training &&
```
vedaanta: - please change to <= 91500
- this only fails for bwd graph
- this only fails if there is NO padding mask
- only fails for s_kv%128!=0 (no need to check s_q)
- this fails for all layouts like BHSD/SBHD/BSHD (THD works because it inherently has padding mask)

transformer_engine/common/fused_attn/fused_attn.cpp
ksivaman: There is no way for us to disable it during forward pass and re-enable it in the backward pass with the current method of checking things. Right? @cyanguwa

ksivaman: > plus, this fails whenever there is no causal/brcm/sliding (meaning cases like padding or no mask at all will work)

Is this condition on top of the previous conditions or as a separate case where it also fails?

transformer_engine/common/fused_attn/fused_attn.cpp
ksivaman: It's both

transformer_engine/common/fused_attn/fused_attn.cpp
cyanguwa: > There is no way for us to disable it during forward pass and re-enable it in the backward pass with the current method of checking things. Right? @cyanguwa

No, there isn't.

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -407,6 +407,26 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
...
                 << std::endl;
     }
+    if ((cudnn_runtime_version == 91400) && (max_seqlen_kv > 1024) &&
+        (attn_mask_type != NVTE_Mask_Type::NVTE_CAUSAL_MASK)) {
+      backend = NVTE_Fused_Attn_Backend::NVTE_No_Backend;
+      std::cout << "Warning: Given combination of attention mask (non-causal) and "
+                   "max_seqlen_kv (> 1024) does not support cuda graph capture with "
+                   "fused attention for cuDNN 9.13.* or 9.14.*."
+                   " Please upgrade your cuDNN version if possible."
+                << std::endl;
```
cyanguwa, line 417: - can you remove 9.13.* from the error string?
- could you please add this to the mask constraint as well? `(attn_mask_type != NVTE_Mask_Type::NVTE_CAUSAL_BOTTOM_RIGHT_MASK) && (window_size_left != -1)`

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -407,6 +407,26 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
...
+        (qkv_format == NVTE_QKV_Format::NVTE_BSHD || qkv_format == NVTE_QKV_Format::NVTE_SBHD) &&
+        (max_seqlen_kv % 128 != 0) && cuda_graph &&
+        (attn_mask_type != NVTE_Mask_Type::NVTE_PADDING_MASK)) {
+      backend = NVTE_Fused_Attn_Backend::NVTE_No_Backend;
+      std::cout << "Warning: Given combination of attention mask (non-padding),"
+                   " max_seqlen_kv (not divisible by 128), and qkv_format (BSHD/SBHD) for"
+                   " backward fused attention with graph capture requires cuDNN 9.15.1+. "
+                   "Please upgrade your cuDNN version if possible."
+                << std::endl;
+    }
```
cyanguwa, line 431: - I think these conditions were working for pre-9.13? Could you please add `(cudnn_runtime_version > 91300)`?
- Could you also add to the mask constraint `(attn_mask_type != NVTE_Mask_Type::NVTE_PADDING_CAUSAL_MASK) &&  (attn_mask_type != NVTE_Mask_Type::NVTE_PADDING_CAUSAL_BOTTOM_RIGHT_MASK)` please?

cyanguwa (commented):
Other than those two comments, I think it looks good!

transformer_engine/common/fused_attn/fused_attn.cpp
cyanguwa: I think you can ignore the comment about adding `> 91300` based on our offline discussion. It affects all versions <= 9.15.0.

PR #2347, opened 2025-11-05
[JAX] Fused layers argument default values changed

tdophung:
# Description

Changing default arguments in some JAX layers

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Changes some default argument  values to match with the Pytorch implementation + enable fused codepath to be activated by default.:
- Default activation to be gelu instead of relu
- Return_layernorm_output to False to enable fused layernorm and MLP/DenseGeneral by default
- Intermediate_dropout_rate to 0.0: this dropout layer does not exist on the PyTorch side, so we should have it turned off by default also on Jax side. Will be removed completely when rubin comes

# Checklist:

### DIFF
--- a/tests/jax/test_distributed_layernorm_mlp.py
+++ b/tests/jax/test_distributed_layernorm_mlp.py
@@ -389,7 +389,7 @@ class TestDistributedLayernormMLP:
                     intermediate_dim=INTERMEDIATE,
                     activations=activation_type,
                     use_bias=use_bias,
-                    return_layernorm_output=True
+                    return_layernorm_output=True,
                 )
                 params_single = ln_mlp_single.init(init_rngs, x, deterministic=True)
                 mlp_out_single, ln_out_single = ln_mlp_single.apply(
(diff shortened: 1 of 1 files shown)


transformer_engine/jax/flax/transformer.py
```diff
@@ -1753,12 +1753,12 @@ class TransformerLayer(nn.Module):  # pylint: disable=too-few-public-methods
...
     hidden_dropout_dims: Sequence[int] = ()
     attention_dropout: float = 0.1
-    intermediate_dropout: float = 0.1
+    intermediate_dropout: float = 0.0
     intermediate_dropout_dims: Sequence[int] = ()
     dropout_rng_name: str = "dropout"
     mha_kernel_init: Initializer = None
     mlp_kernel_init: Initializer = None
-    mlp_activations: Sequence[str] = ("relu",)
+    mlp_activations: Sequence[str] = ("gelu",)
```
jberchtold-nvidia, line 1761: TE/PyTorch seems to not have a default activation type and afaict requires the user to explicitly set it
https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/module/layernorm_mlp.py#L204

For now, we can keep this PR as is to make the activation default consistent across TE/JAX, but in future we could try making it an required argument with no default.

jberchtold-nvidia:
LGTM pending CI, thanks for making this change! Since this PR is smaller it is likely to ready to merge before the getting started tutorial, so you may want to rebase this on main instead

PR #2351, opened 2025-11-06
[PyTorch][NVFP4][MOE] NVFP4 Grouped Hadamard Amax Kernel

zhongbozhu:
# Description

This PR is one of the many on-going grouped kernels for NVFP4 to reduce CPU overhead and reduce quantization cost. 

**This PR is ready for code review**

Action items: 
- [x] Clean up code, more testing
- [x] Enable optionally disabling RHT, only get amax (@timmoon10 FP8 CS can potentially use this one too).
- [x] Fix stochastic rounding in the grouped kernel 
- [x] Let MegatronLM also bump up the NVFP4 M dim padding from 32 to 64, check out [PR to MCore main](https://github.com/NVIDIA/Megatron-LM/pull/1985) & [PR to Mcore dev](https://github.com/NVIDIA/Megatron-LM/pull/2036)

Further optimization ideas:
- Made a persistent RHT amax kernel [here](https://github.com/zhongbozhu/TransformerEngine/commit/423eed64fc970eaf5f4924704b4237e81fa921fc), not seeing speedup for target problem shape, need more investigation 
- Explore some hyper-params

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/csrc/extensions/cast.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/cast.cpp
@@ -560,14 +560,18 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>, bool> bulk_alloc
     for (size_t i = 0; i < num_tensors; ++i) {
       // FP4 data is aligned to 256B
       const auto offset = roundup(buffer_size, 256);
-      if (offset != buffer_size) { contiguous_data_and_scale = false; }
+      if (offset != buffer_size) {
+        contiguous_data_and_scale = false;
+      }
       data_offsets.push_back(offset);
       buffer_size = offset + (product(rowwise_data_shapes[i]) + 1) / 2;
     }
     for (size_t i = 0; i < num_tensors; ++i) {
       // Scales are aligned to 16B
       const auto offset = roundup(buffer_size, 16);
-      if (offset != buffer_size) { contiguous_data_and_scale = false; }
+      if (offset != buffer_size) {
+        contiguous_data_and_scale = false;
+      }
       scale_offsets.push_back(offset);
       buffer_size = offset + product(rowwise_scale_shapes[i]) * scale_elem_size;
     }
@@ -618,14 +622,18 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>, bool> bulk_alloc
     for (size_t i = 0; i < num_tensors; ++i) {
       // FP4 data is aligned to 256B
       const auto offset = roundup(buffer_size, 256);
-      if (offset != buffer_size) { contiguous_data_and_scale = false; }
+      if (offset != buffer_size) {
+        contiguous_data_and_scale = false;
+      }
       data_offsets.push_back(offset);
       buffer_size = offset + (product(columnwise_data_shapes[i]) + 1) / 2;
     }
     for (size_t i = 0; i < num_tensors; ++i) {
       // Scales are aligned to 16B
       const auto offset = roundup(buffer_size, 16);
-      if (offset != buffer_size) { contiguous_data_and_scale = false; }
+      if (offset != buffer_size) {
+        contiguous_data_and_scale = false;
+      }
       scale_offsets.push_back(offset);
       buffer_size = offset + product(columnwise_scale_shapes[i]) * scale_elem_size;
     }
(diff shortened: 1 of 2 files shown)


transformer_engine/common/include/transformer_engine/hadamard_transform.h
```diff
@@ -61,6 +61,38 @@ void nvte_hadamard_transform_cast_fusion_columnwise(const NVTETensor input, NVTE
...
+ *  This function is experimental and the API is not stable.
+ *
+ *  \param[in]      input           Input tensor, assumed contiguous in memory and split on dimension 0.
+ *  \param[in,out]  outputs         Array of output tensors.
+ *  \param[in]      split_sections  Array specifying splits in dimension 0 for each output tensor.
+ *  \param[in]      num_tensors     Number of output tensors, must be > 0.
+ *  \param[in]      stream          CUDA stream used for the operation.
+ */
+void nvte_multi_tensor_amax(const NVTETensor input, NVTETensor* outputs, const int* split_sections,
+                            const size_t num_tensors, cudaStream_t stream);
```
timmoon10: Let's move this to `multi_tensor.h` since it's not related to Hadamard transforms.

Also, we should make the API clearer and disentangle it from tensor splitting. I propose:
```c++
/*!
 *  \brief Compute absolute maximum (amax) of multiple tensors.
 *
 *  \param[in]  inputs      Input tensors.
 *  \param[out] amaxes      Output tensors. Must be an FP32 tensors with one entry.
 *  \param[in]  num_tensors Number of tensors.
 *  \param[in]  stream      CUDA stream used for the operation.
 */
void nvte_multi_tensor_amax(const NVTETensor *inputs, NVTETensor *amaxes, size_t num_tensors, cudaStream_t stream);
```
I think this is fine when the split points are on CPU. The inputs will be a high-precision tensor, so creating tensor splits will be cheap.

If we have GPU split points, then the tensor split will be an intrinsic part fo the compute. We should then give it a more descriptive name like `nvte_split_amaxes`:
```c++
/*!
 *  \brief Split tensor along first dim and compute absolute maximum (amax) for each.
 *
 *  \param[in]  input        Input tensor.
 *  \param[in]  split_points Tensor split points (cumulative indices, not split sizes). Must be an INT64 tensor.
 *  \param[out] amaxes       Output tensors. Must be FP32 tensors with one entry and the number of tensors must match the number of split points.
 *  \param[in]  stream       CUDA stream used for the operation.
 */
void nvte_split_amaxes(const NVTETensor input, const NVTETensor split_points, NVTETensor *amaxes, cudaStream_t stream);
```
Better, we could use the group tensor infrastructure from https://github.com/NVIDIA/TransformerEngine/pull/2388.

transformer_engine/common/include/transformer_engine/hadamard_transform.h
```diff
@@ -61,6 +61,38 @@ void nvte_hadamard_transform_cast_fusion_columnwise(const NVTETensor input, NVTE
...
+ *
+ *  \param[in]      input                Input tensor to apply Hadamard transform, assumed contiguous in memory and split on dimension 0.
+ *  \param[in,out]  outputs              Array of output tensors.
+ *  \param[in]      split_sections       Array of splits in dimension 0 for each output tensor.
+ *  \param[in]      num_tensors          Number of output tensors, must be > 0.
+ *  \param[in]      random_sign_mask     16-bit (int) sign mask for transform.
+ *  \param[in]      random_sign_mask_t   16-bit (int) sign mask for transform (transposed).
+ *  \param[in]      stream               CUDA stream used for the operation.
+ */
+void nvte_multi_hadamard_transform_amax(const NVTETensor input, NVTETensor* outputs,
```
timmoon10: We should make the API clearer, making similar changes as in https://github.com/NVIDIA/TransformerEngine/pull/2351#discussion_r2547406912. For one, this function makes me think that we will populate `outputs` with the RHT values, but we're only filling the amaxes. Also, if we're incorporating tensor splits in the compute, then that should be reflected in the name.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -100,10 +100,224 @@ py::object dequantize(const py::handle &input, transformer_engine::DType otype)
...
+        if (input_list[i].numel() == 0) {
+          continue;
+        }
+        nvte_quantize_v2(input_list[i].data(), output_list[i].data(), quant_config_list[i], stream);
+      }
+    });
+  }
+}
+
+void multi_tensor_quantize_impl(const TensorWrapper &single_input,
```
timmoon10: I think we should separate the logic for multi-tensor quantize and split + multi-tensor quantize. It's a bad sign that we're fighting against the design here: https://github.com/NVIDIA/TransformerEngine/blob/e0a47141b461a2c55f033156e94e016e7b859943/transformer_engine/pytorch/csrc/extensions/cast.cpp#L454-L460

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -246,12 +526,18 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>> bulk_allocate_fp
     size_t buffer_size = 0;
     std::vector<size_t> data_offsets, scale_offsets;
     for (size_t i = 0; i < num_tensors; ++i) {
+      // Check if roundup changed buffer_size to determine if data buffer is contiguous
+      size_t prev_buffer_size = buffer_size;
       buffer_size = roundup(buffer_size, 256);  // align to 256B
+      contiguous_data_and_scale &= (prev_buffer_size == buffer_size);
```
timmoon10: Nit: `&=` is not safe since it does a bit-wise AND. It'll technically work since the [C++ standard guarantees that `bool`s are cast to 0 or 1](https://en.cppreference.com/w/cpp/language/implicit_cast.html#Integral_conversions), but it's vulnerable to bugs if the RHS evaluates to anything other than `bool`.
```suggestion
      contiguous_data_and_scale = contiguous_data_and_scale && (prev_buffer_size == buffer_size);
```

PR #2355, opened 2025-11-06
[PyTorch] Fix amax computation using output_t data in normalization

negvet:
# Description

Fused amax was computed for compute_t data (fp32), not the output_t (e.g. bf16), which is not exactly correct.
Now amax is computed on output_t data (adding one more conversion).

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/common/normalization/layernorm/ln_fwd_kernels.cuh
+++ b/transformer_engine/common/normalization/layernorm/ln_fwd_kernels.cuh
@@ -123,7 +123,14 @@ __global__ __launch_bounds__(Ktraits::THREADS_PER_CTA) void ln_fwd_tuned_kernel(
 
         if (requires_amax) {
           __builtin_assume(amax >= 0);
-          amax = fmaxf(amax, fabsf(temp_output));
+          if (params.fp8_out) {
+            // For fp8_out, keep amax on pre-scale compute_t
+            amax = fmaxf(amax, fabsf(temp_output));
+          } else {
+            // Otherwise compute amax on the value converted to output_t (e.g., bf16)
+            output_t out_t_val = output_t(temp_output);
+            amax = fmaxf(amax, fabsf(compute_t(out_t_val)));
+          }
         }
         if (params.fp8_out) {
           temp_output = temp_output * scale;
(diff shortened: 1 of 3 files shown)


transformer_engine/common/normalization/layernorm/ln_fwd_kernels.cuh
```diff
@@ -290,7 +297,14 @@ __global__ __launch_bounds__(Ktraits::THREADS_PER_CTA) void ln_fwd_general_kerne
...
             compute_t z_ij = z.data.elt[jt];
             __builtin_assume(amax >= 0);
-            amax = fmaxf(amax, fabsf(z_ij));
+            if (params.fp8_out) {
+              // For fp8_out, keep amax on pre-scale compute_t
+              amax = fmaxf(amax, fabsf(z_ij));
+            } else {
+              // Otherwise compute amax on the value converted to output_t (e.g., bf16)
+              output_t out_t_val = output_t(z_ij);
+              amax = fmaxf(amax, fabsf(compute_t(out_t_val)));
```
ksivaman, line 306: Why do we want to do this? We apply the final scalar on `compute_t` dtype, which is `fp32`, so it makes sense to also calculate that scale based on the full `fp32` amax value instead of narrowing it down first?

Also, why only norms? Is it a recipe related thing?

timmoon10 (commented):
This PR looks reasonable, but can you explain why it's better than the existing impl? I could buy it, but it's not obvious to me.

Let's consider two cases:
1. LayerNorm+amax+cast kernel, FP8 scale update kernel
2. LayerNorm+amax kernel, FP8 scale update kernel, cast kernel

My first instinct is that these two cases should produce results that are as close as possible, so amaxes should be identical. However, in case 2 the scale is never applied to FP32 values, so perhaps having full FP32 accuracy in amax is unnecessary?

negvet:
> This PR looks reasonable, but can you explain why it's better than the existing impl? I could buy it, but it's not obvious to me.
> 
> Let's consider two cases:
> 
> 1. LayerNorm+amax+cast kernel, FP8 scale update kernel
> 2. LayerNorm+amax kernel, FP8 scale update kernel, cast kernel
> 
> My first instinct is that these two cases should produce results that are as close as possible, so amaxes should be identical. However, in case 2 the scale is never applied to FP32 values, so perhaps having full FP32 accuracy in amax is unnecessary?

Right, in case 2 there is a down-cast at the end of LayerNorm+amax kernel before scale-from-amax and cast kernels, so amax should be for this down-casted tensor rather than intermediate FP32 data after normalization.

timmoon10 (approved):
Okay, so for the non-FP8 case we should consider the BF16 cast as part of the LayerNorm compute. Since the FP8 and non-FP8 cases are doing different math, it makes sense that the amaxes are slightly different.

PR #2360, opened 2025-11-07
Add device-Initiated Grouped GEMM supporting m_splits on device

QiZhangNV:
# Description

Introduces a CUTLASS-based grouped GEMM implementation that reads m_splits directly on the device. 

This optimization removes the need for device-to-host data transfers and synchronization in MCore, while allowing the number of quantization kernels to be reduced to one. 

The kernel is fully compatible with CUDA Graphs.

Key points:
•	Does not break the existing API. The operator now accepts m_splits as either a torch.Tensor (on CPU or GPU) or a Python list.
•	Reduces CPU overhead, especially for large expert counts, by using a single quantization kernel instead of one per GEMM.
•	Currently supports only MXFP8.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change `m_splits` from `List[int]` to `torch.Tensor`, but can still run correctly with `List[int]` (will be internally converted to a tensor)
- Add `te_general_device_initiated_grouped_gemm`

## Unit Test
`pytest -v -s tests/pytorch/test_numerics.py::test_grouped_linear_accuracy_cutlass_device`

# Checklist:

### DIFF
--- a/tests/pytorch/test_numerics.py
+++ b/tests/pytorch/test_numerics.py
@@ -1758,6 +1758,7 @@ def _test_grouped_linear_accuracy(
     fp8,
     fuse_wgrad_accumulation,
     delay_wgrad_compute=False,
+    m_splits_on_device=False,
 ):
     reset_rng_states()
     if fp8:
@@ -1769,6 +1770,11 @@ def _test_grouped_linear_accuracy(
         device="cuda",
         requires_grad=True,
     )
+    # for i in range(config.max_seqlen_q):
+    #     for j in range(config.hidden_size):
+    #         inp_hidden_states[i, :, j] = 1
+    # print("inp_hidden_states:", inp_hidden_states)
+    # inp_hidden_states.requires_grad_()
     inp_hidden_states.retain_grad()
 
     if num_gemms > 1:
@@ -1776,20 +1782,24 @@ def _test_grouped_linear_accuracy(
         if fp8:
             split_size = 16
             if recipe.mxfp8() or recipe.nvfp4():
-                split_size = 32
+                split_size = 128
         m = config.max_seqlen_q // split_size
         dist = torch.sort(torch.randint(0, m, (num_gemms - 2,))).values.tolist()
         dist.append(dist[-1])  # Manually add a zero
+        # dist = torch.sort(torch.randint(0, m, (num_gemms - 1,))).values.tolist()
         m_splits = torch.tensor(dist + [m]) - torch.tensor([0] + dist)
         m_splits = m_splits * split_size
         assert m_splits.sum() == config.max_seqlen_q and len(m_splits) == num_gemms
     else:
         m_splits = torch.tensor([config.max_seqlen_q])
+    # print("m_splits:", m_splits)
 
     with autocast(enabled=fp8, recipe=recipe):
+        if m_splits_on_device:
+            m_splits = m_splits.to("cuda")
         if isinstance(block, GroupedLinear):
             m_splits = m_splits * bs
-            out = block(inp_hidden_states, m_splits.tolist())
+            out = block(inp_hidden_states, m_splits)
         else:
             out = torch.cat(
                 [
@@ -1797,7 +1807,12 @@ def _test_grouped_linear_accuracy(
                     for i, inp in enumerate(torch.split(inp_hidden_states, m_splits.tolist()))
                 ]
             )
-    loss = out.sum()
+    target = torch.rand_like(out, device=out.device, dtype=out.dtype)
+    # for i in range(out.shape[0]):
+    #     for j in range(out.shape[1]):
+    #         for k in range(out.shape[2]):
+    #             target[i, j, k] = 1
+    loss = (out * target).sum()
     loss.backward()
     if delay_wgrad_compute:
         if isinstance(block, GroupedLinear):
@@ -1815,6 +1830,7 @@ def _test_grouped_linear_accuracy(
                 assert p.grad is None  # grad should be None if fuse_wgrad_accumulation is True
             else:
                 outputs.append(p.grad)
+    # outputs = [out]
     return outputs
 
 
@@ -1839,6 +1855,8 @@ def test_grouped_linear_accuracy(
     delay_wgrad_compute,
     parallel_mode=None,
     use_cutlass=False,
+    m_splits_on_device=False,
+    num_unfuse_wgrad_accumulation=0,
 ):
     fp8 = recipe is not None
     if fp8 and fp8_model_params and NVTE_TEST_NVINSPECT_ENABLED:
@@ -1853,7 +1871,17 @@ def test_grouped_linear_accuracy(
             pytest.skip(
                 f"Input dtype {dtype} not supported for NVFP4 Recipe {recipe.__class__.__name__}"
             )
-
+    if num_unfuse_wgrad_accumulation > 0 and not m_splits_on_device:
+        pytest.skip("Partial accumulate is not supported when m_splits_on_device is False")
+    wgrad_accumulation_mask = None
+    if fuse_wgrad_accumulation and num_unfuse_wgrad_accumulation > 0 and num_unfuse_wgrad_accumulation < num_gemms:
+        wgrad_accumulation_mask = torch.ones(num_gemms, dtype=torch.bool)
+        indices = list(range(num_gemms))
+        random.shuffle(indices)
+        for idx in indices[:num_unfuse_wgrad_accumulation]:
+            wgrad_accumulation_mask[idx] = False
+    # print("fuse_wgrad_accumulation:", fuse_wgrad_accumulation)
+    # print("wgrad_accumulation_mask:", wgrad_accumulation_mask)
     with quantized_model_init(enabled=fp8 and fp8_model_params, recipe=recipe):
         grouped_linear = GroupedLinear(
             num_gemms,
@@ -1864,9 +1892,12 @@ def test_grouped_linear_accuracy(
             parallel_mode=parallel_mode,
             device="cuda",
             fuse_wgrad_accumulation=fuse_wgrad_accumulation,
+            wgrad_accumulation_mask=wgrad_accumulation_mask,
             delay_wgrad_compute=delay_wgrad_compute,
             save_original_input=False,
         ).eval()
+        if wgrad_accumulation_mask is None:
+            wgrad_accumulation_mask = torch.full((num_gemms,), fuse_wgrad_accumulation, dtype=torch.bool)
         sequential_linear = torch.nn.ModuleList(
             [
                 Linear(
@@ -1876,9 +1907,9 @@ def test_grouped_linear_accuracy(
                     params_dtype=dtype,
                     parallel_mode=parallel_mode,
                     device="cuda",
-                    fuse_wgrad_accumulation=fuse_wgrad_accumulation,
+                    fuse_wgrad_accumulation=wgrad_accumulation_mask[i],
                 ).eval()
-                for _ in range(num_gemms)
+                for i in range(num_gemms)
             ]
         )
 
@@ -1888,11 +1919,12 @@ def test_grouped_linear_accuracy(
             sequential_linear[i].weight = Parameter(getattr(grouped_linear, f"weight{i}").clone())
             if bias:
                 sequential_linear[i].bias = Parameter(getattr(grouped_linear, f"bias{i}").clone())
-            if fuse_wgrad_accumulation:
+            if wgrad_accumulation_mask[i]:
                 weight_i = getattr(grouped_linear, f"weight{i}")
                 weight_i.main_grad = torch.rand_like(weight_i, dtype=torch.float32)
                 sequential_linear[i].weight.main_grad = weight_i.main_grad.clone()
-
+    import torch.cuda.nvtx as nvtx
+    nvtx.range_push("sequential_gemm")
     outputs_ref = _test_grouped_linear_accuracy(
         sequential_linear,
         num_gemms,
@@ -1903,7 +1935,11 @@ def test_grouped_linear_accuracy(
         fp8,
         fuse_wgrad_accumulation,
         delay_wgrad_compute,
+        m_splits_on_device
     )
+    nvtx.range_pop()
+    torch.cuda.synchronize()
+    nvtx.range_push("grouped_gemm")
     outputs = _test_grouped_linear_accuracy(
         grouped_linear,
         num_gemms,
@@ -1914,7 +1950,10 @@ def test_grouped_linear_accuracy(
         fp8,
         fuse_wgrad_accumulation,
         delay_wgrad_compute,
+        m_splits_on_device
     )
+    nvtx.range_pop()
+   
 
     for o, o_ref in zip(outputs, outputs_ref):
         if use_cutlass:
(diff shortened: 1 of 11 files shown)


transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -845,8 +920,8 @@ def backward_dw(self):
             wgrad_list = tensor_list[2]
             weight_params = [getattr(self, f"weight{i}") for i in range(self.num_gemms)]
             bias_params = [getattr(self, f"bias{i}") for i in range(self.num_gemms)]
-            if not self.fuse_wgrad_accumulation:
-                for i in range(self.num_gemms):
+            for i in range(self.num_gemms):
+                if not self.fuse_wgrad_accumulation or (self.wgrad_accumulation_mask is not None and not self.wgrad_accumulation_mask[i]):
```
zhongbozhu, line 919: what is wgrad_accumulation_mask?

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -403,20 +456,26 @@ def backward(ctx, grad_output: torch.Tensor) -> Tuple[Union[torch.Tensor, None],
...
+                        assert ctx.fp8 and ctx.fp8_recipe.mxfp8(), "Only MXFP8 is supported when m_splits is on devie"
+                        # Cannot split because the m_splits is not available on host.
+                        inputmats = tex.split_quantize(inp_view, [inp_view.size(0)], ctx.input_quantizers[:1])
                 grouped_gemm_wgrad = functools.partial(
                     general_grouped_gemm,
                     out_dtype=ctx.activation_dtype,
-                    workspaces=get_multi_stream_cublas_workspace(),
-                    layout="NT",
+                    workspaces=get_multi_stream_cublas_workspace() if not ctx.m_splits_on_devie else get_cutlass_grouped_gemm_workspace(),
+                    layout="NT" if not ctx.m_splits_on_devie else "TN",
```
zhongbozhu: why should we change the layout back to TN for backward pass?

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -124,11 +129,15 @@ def forward(
...
+            if fp8:
+                inputmats = tex.split_quantize(inp_view, m_splits.tolist(), input_quantizers)
+            else:
+                inputmats = torch.split(cast_if_needed(inp_view, activation_dtype), m_splits.tolist())
         else:
-            inputmats = torch.split(cast_if_needed(inp_view, activation_dtype), m_splits)
-
+            assert fp8 and FP8GlobalStateManager.get_fp8_recipe().mxfp8(), "Only MXFP8 is supported when m_splits is on devie"
+            # Cannot split because the m_splits is not available on host.
+            inputmats = tex.split_quantize(inp_view, [inp_view.size(0)], input_quantizers[:1])
```
zhongbozhu: I think we can support this dispatch in C++ split quantize. I made the dispatching for NVFP4 here: https://github.com/NVIDIA/TransformerEngine/pull/2351

tests/pytorch/test_numerics.py
```diff
@@ -1776,7 +1777,7 @@ def _test_grouped_linear_accuracy(
         if fp8:
             split_size = 16
             if recipe.mxfp8() or recipe.nvfp4():
-                split_size = 32
+                split_size = 128
```
zhongbozhu, line 1780: cc @Oleg-Goncharov Can we modify the MXFP8 quantize kernel to respect the 128x4 padding of mxfp8 scaling factors? Or do you think it's better if we should do a larger padding here.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -63,7 +64,7 @@ class _GroupedLinear(torch.autograd.Function):
     def forward(
         ctx,
         inp: torch.Tensor,
-        m_splits: List[int],
+        m_splits: torch.Tensor,
```
zhongbozhu, line 66: Do we have any PRs to Mcore that will support this API change? It's also a bit headache that Mcore might have to check TE versions and dispatch for compatibility.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -170,13 +186,15 @@ def forward(
...
         # Perform GEMM
         _ = general_grouped_gemm(
-            weights_fp8,
-            inputmats,
+            weights_fp8 if not m_splits_on_devie else inputmats,
+            inputmats if not m_splits_on_devie else weights_fp8,
             [out],
             activation_dtype,
-            get_multi_stream_cublas_workspace(),
+            get_multi_stream_cublas_workspace() if not m_splits_on_devie else get_cutlass_grouped_gemm_workspace(),
```
zhongbozhu: let's make it a more general API to get workspace

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -170,13 +186,15 @@ def forward(
...
             single_output=True,
+            layout="TN" if not m_splits_on_devie else "NT",
```
zhongbozhu: Same question about the layout, why is NT no longer needed? TE cublas path requires this layout to select whether rowwise data or columnwise data to participate in the GEMM.

transformer_engine/pytorch/module/base.py
```diff
@@ -81,6 +82,8 @@ def get_cublas_workspace_size_bytes() -> None:
         return 32 * 1024 * 1024 + 1024
     return 4_194_304
 
+def get_cutlass_grouped_gemm_workspace_size_bytes() -> None:
+    return 16_777_216
```
zhongbozhu, line 86: Can you explain this size requirement for cutlass grouped gemm?

transformer_engine/pytorch/module/grouped_linear.py
timmoon10: I agree backward compatibility is important. How about adding logic like:
```python
if not isinstance(m_splits, torch.Tensor):
    m_splits = torch.tensor(m_splits, dtype=torch.int, device="cpu")
```
This could be refined further so we don't keep going back-and-forth between `List[int]` and `torch.Tensor`.

transformer_engine/pytorch/module/grouped_linear.py
timmoon10: Can we just call `input_quantizers[0](inp_view)`? If we're not doing a split, then no need to deal with the extra overhead of `split_quantize`. In general, I don't see much value in forcing the device-initiated and host-initiated impls to use the same internal APIs.

zhongbozhu: Sounds good, we should either call `input_quantizers[0](inp_view)`, or if we can tweak the mxfp8 quantize kernel to respect the 128x4 scaling factor padding (so that we don't need 128 aligned padding), then we can use the split_quantize.

transformer_engine/pytorch/module/grouped_linear.py
QiZhangNV: It is for [MoE Echo](https://docs.google.com/document/d/1zXSyUXCsFXfW4vnug82R5usnptNl_dtVS9mnl8edQAk/edit?tab=t.0), which is designed to avoid dropping tokens when a fixed buffer size is used for CUDA graph. When MoE Echo is enabled, each GroupedLinear layer contains both home experts and echo experts. Echo experts correspond to weights from other ranks, whose `main_grad `are not available on the current rank. Therefore, their wgrad should be excluded from accumulation, which is controlled by the `wgrad_accumulation_mask`.

transformer_engine/pytorch/module/grouped_linear.py
QiZhangNV: I didn't swapAB for cutlass backend

transformer_engine/pytorch/module/grouped_linear.py
QiZhangNV: Thanks for the comment. I’ll go with  `input_quantizers[0](inp_view)` for now. Also noticed you looped in Oleg to discuss making the MXFP8 quantize kernel respect the 128×4 scaling-factor padding, we can wait for his suggestions

transformer_engine/pytorch/module/grouped_linear.py
QiZhangNV: Already have the check in `GroupedLinear forward()`: https://github.com/QiZhangNV/TransformerEngine/blob/cutlass_device_grouped_gemm/transformer_engine/pytorch/module/grouped_linear.py#L832-L833
Does it still needed in `_GroupedLinear forward()`?

transformer_engine/pytorch/module/base.py
QiZhangNV: Just allocate a sufficiently large workspace so the CUTLASS grouped GEMM kernel can initialize and run safely; the exact size depends on the kernel config.

transformer_engine/pytorch/module/grouped_linear.py
QiZhangNV: Done

transformer_engine/pytorch/csrc/extensions/gemm.cpp
```diff
@@ -552,4 +552,221 @@ std::optional<std::vector<at::Tensor>> te_general_grouped_gemm(
...
+    // To enable CUDA Graph, we need to use a pinned host buffer to avoid illegal memory access during H2D copy.
+    at::Tensor inputB_and_SF_addrs;
+    if (at::cuda::currentStreamCaptureStatusMayInitCtx() != at::cuda::CaptureStatus::None) {
+      NVTE_CHECK(pinned_host_buffer_index + num_gemms * 2 <= workspace[1].size(0),
+                 "Pinned host buffer out of bounds, please increase the capacity by setting "
+                 "NVTE_CUTLASS_HOST_PINNED_U64_CAPACITY. "
+                 "Current buffer size: ",
+                 workspace[1].size(0));
+      inputB_and_SF_addrs = workspace[1].narrow(0, pinned_host_buffer_index, num_gemms * 2);
+      pinned_host_buffer_index += num_gemms * 2;
```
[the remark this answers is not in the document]
QiZhangNV: CUDA Graph capture records only one iteration; after capture, the graph is replayed each step, so those lines (including the pinned_host_buffer_index increment) are not reached again outside capture.

transformer_engine/pytorch/module/base.py
```diff
@@ -103,6 +106,21 @@ def get_multi_stream_cublas_workspace() -> List[torch.Tensor]:
...
 
+def get_cutlass_grouped_gemm_workspace() -> List[torch.Tensor]:
+    """Returns workspace for cutlass grouped gemm."""
+    global _cutlass_grouped_gemm_workspace
+    if not _cutlass_grouped_gemm_workspace:
+        _cutlass_grouped_gemm_workspace = [
+            # Device buffer for cutlass arguments and kernel
+            torch.empty(get_cutlass_grouped_gemm_workspace_size_bytes(), dtype=torch.uint8, device="cuda"),
+            # TODO: Only allocate pinned buffer when cuda graph is enabled
+            # Host pinned buffer for the source of H2D copy of cutlass arguments
```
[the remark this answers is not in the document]
QiZhangNV: The device buffer will always be used. The pinned buffer cannot be lazy allocation since cuda graph capture does not support .pinned_memory()

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -205,24 +210,55 @@ def general_grouped_gemm(
...
+    else:
+        assert isinstance(A[0], MXFP8TensorStorage) and isinstance(
+            B[0], MXFP8TensorStorage), "Only MXFP8 A and B are supported when m_splits is on device"
+        assert out[0].dtype == torch.bfloat16 or out[0].dtype == torch.float16 or (
+            wgrad and out[0].dtype == torch.float32), "Only BF16, FP16 or FP32(only for wgrad accumulation) output is supported when m_splits is on device"
+        assert not use_bias, "Bias is not supported when m_splits is on device"
+        assert not gelu, "GELU is not supported when m_splits is on device"
+        assert TE_DType[out[0].dtype] == out_dtype, "Output dtype mismatch: out[0].dtype=" + \
+            str(out[0].dtype) + ", out_dtype=" + str(out_dtype)
+        bias = tex.te_general_device_initiated_grouped_gemm(
```
yaox12: Can we do all the dispatch in CPP?

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -228,6 +229,74 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
...
+ *  \param[out]    workspace             List of workspace tensors.
+ *  \param[in]     workspaceSize         Workspace size.
+ *  \param[in]     accumulate            Whether to accumulate the result into the D matrix.
+ *  \param[in]     accumulate_mask       Mask to indicate which GEMMs to accumulate.
+ *  \param[in]     use_split_accumulator Whether to use split accumulator in the FP8 GEMM.
+ *  \param[in]     math_sm_count         Number of GPU SMs to use (default=0: use cuBLAS heuristics)
+ *  \param[in]     stream                CUDA stream to wait on.
+ */
+
+void nvte_device_cutlass_grouped_gemm_wgrad(const NVTETensor* A, const NVTETensor* B, void** D,
```
yaox12: Can we use one function for wgrad and other GEMMs?

transformer_engine/pytorch/module/grouped_linear.py
QiZhangNV: Aligned the A/B order and layout to make it clear

transformer_engine/pytorch/module/grouped_linear.py
QiZhangNV: Aligned the A/B order and layout to make it clear

transformer_engine/pytorch/module/base.py
zhongbozhu: It's interesting that it only requires 16MB, while we already used 32MB for a single GEMM for cublas

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -205,24 +210,55 @@ def general_grouped_gemm(
...
+            A,
```
zhongbozhu, line 243: Is it true that this A (maybe B), despite its type is a list of inputs, it's actually a list with a single element in it? Does it make more sense if we still calculate the pointer offsets on device in API `split_quantize` so that the input to `general_grouped_gemm` are still two lists with the same length already splitted?  cc @timmoon10

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -228,6 +229,74 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
...
+ * this function is only used for moe fprop and dgrad.
+ *
+ * Computes:
+ *  - `D = AB` if both `bias` and `pre_gelu_out` are empty tensors
+ *  - `D = AB + bias` if `pre_gelu_out` is empty and `bias` is not empty
+ *  - `D = GELU(AB + bias)` if both `bias` and `pre_gelu_out` are not empty tensors
+ *  Note: bias and GELU are not supported yet.
+ *
+ *  \param[in]     A                     The list of A matrices.
+ *  \param[in]     B_and_SF_addrs        The list of B and SF matrices addresses.
```
huanghua1994:
```suggestion
 *  \param[in]     A_and_SF_addrs        The list of A and SF matrices addresses.
 *  \param[in]     B                     The list of B matrices.
```
Match with the API signature.

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -228,6 +229,74 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
...
+ *  \param[out]    workspace             List of workspace tensors.
+ *  \param[in]     workspaceSize         Workspace size.
+ *  \param[in]     accumulate            Whether to accumulate the result into the D matrix.
+ *  \param[in]     accumulate_mask       Mask to indicate which GEMMs to accumulate.
+ *  \param[in]     use_split_accumulator Whether to use split accumulator in the FP8 GEMM.
+ *  \param[in]     math_sm_count         Number of GPU SMs to use (default=0: use cuBLAS heuristics)
+ *  \param[in]     stream                CUDA stream to wait on.
+ */
+
+void nvte_device_cutlass_grouped_gemm_wgrad(const NVTETensor* A, const NVTETensor* B, void** D,
```
huanghua1994, line 294: My understanding is that wgrad and dgrad have different transpose targets, this API should be just for wgrad only?

transformer_engine/pytorch/module/base.py
QiZhangNV: I think 32MB is just a safe upper bound for the workspace allocation in cublas(see https://docs.nvidia.com/cuda/cublas/index.html#cublassetworkspace); actual usage may typically much lower. For CUTLASS, 16MB is sufficient in my experience, and I’ve added an assert to catch insufficient workspace. 
Since you’re concerned about this,  would it be better to allow setting this value via an environment variable?

transformer_engine/pytorch/cpp_extensions/gemm.py
QiZhangNV: Yes, activations is a single-element list. It would be good if `split_quantize` can handle the 'split' logical. However, can it work well when the m_splits is not available on Host? How will you do that?

transformer_engine/pytorch/module/grouped_linear.py
QiZhangNV: Hi @timmoon10, after discussing with zhongbozhu, we’d like to introduce a new API called `batch_quantize_device`. This API would support quantization not only for mxfp8, but also for other narrow precision types such as nvfp4 in device-initiated mode. For mxfp8, its behavior would be similar to calling `input_quantizers[0](inp_view)`. For nvfp4, in addition to quantized outputs, it would also return the necessary information like amax tensors and the prefix sum of m_split tensors. How do you think?

transformer_engine/common/include/transformer_engine/gemm.h
QiZhangNV: Yes, and for fprop and dgrad, A consists of separate tensors while B is a single contiguous tensor. For wgrad, both A and B are single contiguous tensors, so the processing logic is a bit different.

ptrendx:
Closing this PR since we added the GroupedLinear support via the cuBLASLt and cudnn Frontend.

PR #2361, opened 2025-11-07
[JAX] cuBlasMp integration for CollectiveGemm custom op

denera:
# Description

This PR integrates TE/common cuBlasMp bindings into the TE/JAX CollectiveGemm custom op.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
+++ b/transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
@@ -137,9 +137,8 @@ void CommunicatorHandler::init(int num_total_devices, int num_devices_per_proces
   handler._initialize = true;
 
   // Bootstrap UB via creating a dummy CommOverlapP2PBase object
-  auto _ = CollectiveGemmPlanRegistry::getInstance().get_plan(
-      {1, 1}, DType::kFloat32, JAXX_Collective_Op::ALL_GATHER
-  );
+  auto _ = CollectiveGemmPlanRegistry::getInstance().get_plan({1, 1}, DType::kFloat32,
+                                                              JAXX_Collective_Op::ALL_GATHER);
 }
 
 void InitializeCgemmCommunicator(int num_total_devices, int num_devices_per_process, int process_id,
@@ -185,8 +184,9 @@ void *CollectiveGemmPlan::get_context() {
   }
 }
 
-CollectiveGemmPlan *CollectiveGemmPlanRegistry::get_plan(
-    std::vector<size_t> buffer_shape, DType dtype, JAXX_Collective_Op collective_op) {
+CollectiveGemmPlan *CollectiveGemmPlanRegistry::get_plan(std::vector<size_t> buffer_shape,
+                                                         DType dtype,
+                                                         JAXX_Collective_Op collective_op) {
   auto &comm_handler = CommunicatorHandler::get();
   auto &cgemm_config = CgemmConfig::get();
 
@@ -228,10 +228,11 @@ CollectiveGemmPlan *CollectiveGemmPlanRegistry::get_plan(
         buffer_shape, dtype, comm_handler.get_global_rank(), comm_handler.num_total_devices,
         comm_handler.get_local_device_id_within_tp_domain(), comm_handler.tp_size,
         comm_handler.get_tp_domain_id(), comm_handler.get_tp_num_domains(), comm_handler.tp_size,
-        comm_handler.allgather_func, comm_handler.barrier_func, get_nvte_collective_op(collective_op),
-        cgemm_config.num_max_streams, 1 /*comm_cga_size*/, cgemm_config.gemm_priority,
-        cgemm_config.comm_priority, cgemm_config.num_comm_sm, true /*set_sm_margin*/,
-        cgemm_config.use_ce, false /*atomic_gemm*/, cgemm_config.aggregate_ag);
+        comm_handler.allgather_func, comm_handler.barrier_func,
+        get_nvte_collective_op(collective_op), cgemm_config.num_max_streams, 1 /*comm_cga_size*/,
+        cgemm_config.gemm_priority, cgemm_config.comm_priority, cgemm_config.num_comm_sm,
+        true /*set_sm_margin*/, cgemm_config.use_ce, false /*atomic_gemm*/,
+        cgemm_config.aggregate_ag);
     plan = std::make_unique<CollectiveGemmPlan>(reinterpret_cast<void *>(ctx));
   }
   CollectiveGemmPlan *plan_ptr = plan.get();
--- a/transformer_engine/jax/csrc/extensions/cgemm_helper.h
+++ b/transformer_engine/jax/csrc/extensions/cgemm_helper.h
@@ -158,7 +158,7 @@ class CommunicatorHandler {
 };
 
 struct CollectiveGemmPlan {
-  NVTECommGemmCtx* cublasmp_context;
+  NVTECommGemmCtx *cublasmp_context;
   std::unique_ptr<CommOverlapCore> userbuffers_context;
 
   CollectiveGemmPlan(void *ctx);
@@ -178,7 +178,8 @@ class CollectiveGemmPlanRegistry {
     return instance;
   }
 
-  CollectiveGemmPlan *get_plan(std::vector<size_t> buffer_shape, DType dtype, JAXX_Collective_Op collective_op);
+  CollectiveGemmPlan *get_plan(std::vector<size_t> buffer_shape, DType dtype,
+                               JAXX_Collective_Op collective_op);
 
  private:
   CollectiveGemmPlanRegistry() {}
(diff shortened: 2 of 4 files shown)


phu0ngng:
Hi, could you add some unit tests?

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -157,9 +157,9 @@ int GetCgemmNumMaxStreams() {
   return config.num_max_streams;
 }
 
-CommOverlapCore *CollectiveGemmPlanRegistry::get_executor(std::vector<size_t> buffer_shape,
-                                                          DType dtype,
-                                                          JAXX_Collective_Op collective_op) {
+CollectiveGemmCtx *CollectiveGemmPlanRegistry::get_context(std::vector<size_t> buffer_shape,
```
jberchtold-nvidia: Let's restructure this to instead of a compile-time flag, we compile both backend implementations, similar to TE's support for NormPlanRegistry that can use cuDNN or TE kernels. This way a user can easily switch between the two without needing to rebuild TE/JAX

nvMelissa:
Hi @denera - This needs to go into TE2.10. I see @phu0ngng 's last comment to add some tests. Please let us know if that was already done and if you think it will take much longer. @KshitijLakhani  to keep an eye on this to cherry pick into TE2.10 when ready.

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
jberchtold-nvidia: Hi, just wanted to follow up on this. Thanks!

denera:
> Hi, could you add some unit tests?

The same unit tests we have for Comm+GEMM overlap are still functional for this, but it just requires TE to be compiled with `NVTE_WITH_CUBLASMP=1`. I'll look into how we to make that work with the pipelines, but I don't think it's something we can implement via the unit test system in TE source code.

phu0ngng (changes requested):

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -191,19 +195,27 @@ CommOverlapCore *CollectiveGemmPlanRegistry::get_executor(std::vector<size_t> bu
...
+    auto ctx = CommOverlapP2PBase(
+        buffer_shape, dtype, comm_handler.get_global_rank(), comm_handler.num_total_devices,
+        comm_handler.get_local_device_id_within_tp_domain(), comm_handler.tp_size,
+        comm_handler.get_tp_domain_id(), comm_handler.get_tp_num_domains(), comm_handler.tp_size,
+        comm_handler.allgather_func, comm_handler.barrier_func,
+        get_nvte_collective_op(collective_op), cgemm_config.num_max_streams, 1 /*comm_cga_size*/,
+        cgemm_config.gemm_priority, cgemm_config.comm_priority, cgemm_config.num_comm_sm,
+        true /*set_sm_margin*/, cgemm_config.use_ce, false /*atomic_gemm*/,
+        cgemm_config.aggregate_ag);
+    plan = std::make_unique<CollectiveGemmPlan>(reinterpret_cast<void *>(ctx));
```
phu0ngng: I don't think this code works at all.
How can it be possible to cast an object into a pointer and construct the unique pointer from another pointer?

As long as unit test is not enabled, we should not merge this PR.

phu0ngng (changes requested):

transformer_engine/common/CMakeLists.txt
```diff
@@ -270,14 +270,12 @@ if (NVTE_WITH_CUBLASMP)
     target_compile_definitions(transformer_engine PRIVATE NVTE_WITH_CUBLASMP)
     target_include_directories(transformer_engine PRIVATE ${CUBLASMP_DIR}/include ${NVSHMEM_DIR}/include)
     find_library(CUBLASMP_LIB
-                 NAMES cublasmp libcublasmp
-                 PATHS ${CUBLASMP_DIR}
-                 PATH_SUFFIXES lib
+                 NAMES cublasmp libcublasmp.so.0
```
ptrendx, line 273: Why do you need those changes?

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -157,9 +156,37 @@ int GetCgemmNumMaxStreams() {
...
+
+CollectiveGemmPlan::CollectiveGemmPlan(ncclComm_t comm, int nranks, int rank) {
+  NVTE_CHECK(getenv<bool>("NVTE_WITH_CUBLASMP", false) == true,
+             "Internal TE/JAX error: CollectiveGemmPlan() initializer called for cuBlasMp ",
+             "backend when NVTE_WITH_CUBLASMP=0.");
+  cublasmp_context = nvte_comm_gemm_ctx_create(comm, nranks, rank);
+}
+
+CollectiveGemmPlan::~CollectiveGemmPlan() {
+  if (getenv<bool>("NVTE_WITH_CUBLASMP", false)) {
```
ptrendx, line 174: This seems brittle - what if the environment changes in the middle of the run for whatever reason? The object should keep track whether it was created with UB or cuBLASMp support in some member variable instead.

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -157,9 +156,37 @@ int GetCgemmNumMaxStreams() {
   return config.num_max_streams;
 }
 
-CommOverlapCore *CollectiveGemmPlanRegistry::get_executor(std::vector<size_t> buffer_shape,
-                                                          DType dtype,
-                                                          JAXX_Collective_Op collective_op) {
+CollectiveGemmPlan::CollectiveGemmPlan(void *ctx) {
+  NVTE_CHECK(getenv<bool>("NVTE_WITH_CUBLASMP", false) == false,
```
ptrendx, line 160: I don't really like that - this makes the env variable the only way somebody could run with cuBLASMp. What if I wanted to mix UB and cuBLASMp in some context (e.g. even for benchmarking one vs the other)?

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -157,9 +156,37 @@ int GetCgemmNumMaxStreams() {
...
+}
+
+CollectiveGemmPlan::~CollectiveGemmPlan() {
+  if (getenv<bool>("NVTE_WITH_CUBLASMP", false)) {
+    nvte_comm_gemm_ctx_destroy(cublasmp_context);
+  }
+}
+
+void *CollectiveGemmPlan::get_context() {
+  if (getenv<bool>("NVTE_WITH_CUBLASMP", false)) {
```
ptrendx, line 180: Same issue as in the destructor.

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -157,9 +156,37 @@ int GetCgemmNumMaxStreams() {
...
+    return reinterpret_cast<void *>(cublasmp_context);
+  } else {
+    return reinterpret_cast<void *>(userbuffers_context.get());
+  }
+}
+
+CollectiveGemmPlan *CollectiveGemmPlanRegistry::get_plan(std::vector<size_t> buffer_shape,
```
ptrendx, line 187: Does this call always create a new plan? If not, then probably the buffer_shape should be passed by reference to avoid the copy.

transformer_engine/jax/csrc/extensions/cgemm_helper.h
```diff
@@ -156,6 +157,19 @@ class CommunicatorHandler {
   std::vector<std::string> _nccl_id_file_name;
 };
 
+struct CollectiveGemmPlan {
```
ptrendx, line 160: The way I see it is that this should either be a base class with inheritance or a tagged union, since you will never have both contexts at the same time, right?

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -191,19 +218,26 @@ CommOverlapCore *CollectiveGemmPlanRegistry::get_executor(std::vector<size_t> bu
...
-  CommOverlapCore *executor_ptr = executor.get();
-  plan_map[plan_id] = std::move(executor);
-  return executor_ptr;
+  std::unique_ptr<CollectiveGemmPlan> plan;
+  if (getenv<bool>("NVTE_WITH_CUBLASMP", false)) {
+    plan = std::make_unique<CollectiveGemmPlan>(comm_handler.get_comm_for_current_device(),
+                                                comm_handler.num_total_devices,
+                                                comm_handler.get_global_rank());
+  } else {
+    auto ctx = new CommOverlapP2PBase(
```
jberchtold-nvidia, line 227: Why do we use `new` here then `make_unique`? Can we not also use smart pointers for this `ctx` here?

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
phu0ngng: I think this logic, i.e., selecting the backend, should happen in TE/Common rather than replicated across frameworks.

denera:
Closing in favor of #2443 that includes the changes.

PR #2364, opened 2025-11-10
[JAX] Shardy rule + QuantizeLayout Rework

phu0ngng:
# Description

- Rework the Shardy rules so that they can handle NVFP4 + 2D 1x1x quantization, which will be followed in another PR.
- Unify QuantizeLayout and quantizer.is_2x2x logics.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/gemm.py
+++ b/transformer_engine/jax/cpp_extensions/gemm.py
@@ -533,6 +533,9 @@ class GemmPrimitive(BasePrimitive):
 
         # Declare cuBLAS workspace
         workspace_size = get_cublas_workspace_size_bytes()
+        # NVFP4 swizzling happen in via nvte kernel instead of JAX transposes
+        if scaling_mode.is_nvfp4_scaling:
+            workspace_size += lhs_scale_inv.size + rhs_scale_inv.size
         if not collective_op.is_none:
             workspace_size *= get_cgemm_num_max_streams()
         # cuBLAS workspace ptr must be 256 bytes aligned but JAX buffers are not
@@ -662,6 +665,8 @@ class GemmPrimitive(BasePrimitive):
             rhs_scale_inv = apply_padding_to_scale_inv(
                 rhs_scale_inv, scaling_mode, rhs.shape, not rhs_transposed, rhs_flatten_axis
             )
+        # Only perform JAX-based swizzle for MXFP8, NVFP4 swizzle will go though nvte kernel
+        if scaling_mode.is_mxfp8_scaling:
             lhs_scale_inv = swizzled_scale(lhs_scale_inv, lhs_flatten_axis, lhs_transposed)
             rhs_scale_inv = swizzled_scale(rhs_scale_inv, rhs_flatten_axis, not rhs_transposed)
 
--- a/transformer_engine/jax/csrc/extensions/gemm.cpp
+++ b/transformer_engine/jax/csrc/extensions/gemm.cpp
@@ -34,8 +34,8 @@ static uint8_t *move_ptr_to_next_256B_aligned(uint8_t *ptr) {
 }
 
 std::tuple<TensorWrapper, std::vector<size_t>> xla_buffer_to_nvte_gemm_operand(
-    cudaStream_t stream, Buffer_Type buffer, Buffer_Type scale_inv, JAXX_Scaling_Mode scaling_mode,
-    size_t axis_boundary, bool rowwise) {
+    cudaStream_t stream, Buffer_Type buffer, Buffer_Type scale_inv, uint8_t *swizzle_scale_ptr,
+    JAXX_Scaling_Mode scaling_mode, size_t axis_boundary, bool rowwise) {
   // Set tensor data with collapsed 2D shape
   auto buffer_dims = buffer.dimensions();
   std::vector<size_t> input_shape = {product(buffer_dims, 0, axis_boundary),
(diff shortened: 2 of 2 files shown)


transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -272,7 +270,7 @@ def lowering(
             post_rht_amax,
             rht_matrix,
             scaling_mode=scaling_mode.value,
-            q_layout=q_layout,
+            q_layout=q_layout.value.value,
```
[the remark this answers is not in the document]
phu0ngng: This is correct, as we want to lower q_layout as an int here.

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -259,7 +259,7 @@ def lowering(
             amax,
             act_enum=act_enum,
             scaling_mode=scaling_mode.value,
-            is_2x=is_2x,
+            quantize_layout=quantize_layout.value.value,
```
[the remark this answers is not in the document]
phu0ngng: This is correct, as we want to lower q_layout as an int here.

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -419,7 +419,7 @@ def infer_sharding_from_operands(
         elif scaling_mode == ScalingMode.MXFP8_1D_SCALING.value:
             scale_inv_spec = out_spec
 
-        if is_2x:
+        if quantize_layout.is_rowwise_colwise:
```
jberchtold-nvidia, line 429: Can you add an assert inside this if block that `assert not ScalingMode(scaling_mode).is_colwise_transposed(), "Transpose layout scaling modes are not supported here yet"`

This logic is correct for now as we only ever use fused act with tensor scaling or MXFP8. NVFP4 can't use fused activations because it requires an amax for current scaling. But if we ever introduce a new recipe that is supported by fused act but has a transposed layout for the colwise, this wouldn't work.

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -643,39 +641,37 @@ def shardy_sharding_rule(
...
-
-        dbias = x_axes[flatten_axis:] if is_dbias else (prefix + "dbias",)
-        amax = (prefix + "amax",)
-        sr_rng_state = (prefix + "sr_rng_state_partition_axis", prefix + "sr_rng_state_data_axis")
+        input_spec = scale_rules.input_spec
+        dbias = input_spec[flatten_axis:] if is_dbias else (prefix + "_dbias",)
+        amax = (BATCHING + prefix + "_amax",)
+        scale = (BATCHING + prefix + "_scale",)
+        sr_rng_state = (
+            BATCHING + prefix + "_sr_rng_state_partition_axis",
```
jberchtold-nvidia, line 658: What does this `BATCHING` prefix do? Unlike amax, tensor scale, and the rht_matrix, this SR rng state is not replicated and we do want it split across devices

transformer_engine/jax/cpp_extensions/activation.py
```diff
@@ -173,7 +173,7 @@ def abstract(
         act_enum,
         act_len,
         scaling_mode,
-        is_2x,
+        quantize_layout,
         scale_dtype,
         act_params,
         amax_scope,
```
jberchtold-nvidia, line 179: Can you add an assert in the abstract here that a  colwise-only layout isn't supported in this primitive yet? In theory, out guard in the wrapper function should catch that but might as well add an assert to be safe

transformer_engine/jax/cpp_extensions/activation.py
phu0ngng: I did not rework the `infer_sharding_from_operands` as we will remove it pretty soon.
I can add the assertion for now.

transformer_engine/jax/cpp_extensions/quantization.py
phu0ngng: This BATCHING is for the case we want to batch TE ops, i.e., wrapping an op with vmap. 
We will need to update the GEMM batching rule at one point as well.

transformer_engine/jax/cpp_extensions/activation.py
phu0ngng: I would say we simply assert for not NVFP4.

jberchtold-nvidia: I think it'd be good to add both assertions. If we try to change our logic or something slips thru by mistake and we do invoke colwise-only here for a non-NVFP4 recipe, we may get unexpected behavior because we will be calling the rowwise or 2x kernels where the rowwise buffers aren't properly allocated (they'll be some (1,) tensor) since our C++ FFI isn't handling colwise-only.

transformer_engine/jax/cpp_extensions/quantization.py
jberchtold-nvidia: Oh I see these are set on all Shardy strings now either here or via the scaling mode impl. Thanks for the explanation!

transformer_engine/jax/cpp_extensions/activation.py
jberchtold-nvidia: The same check is present in `partition` . Sounds good, thanks, the assertion is all we need for now. Just want to make sure it's easy to notice in the future if we add a new recipe that needs this functionality, but we don't need to pre-emptively implement any support for it.

Also ack that this logic without the assertion was present before this PR

phu0ngng: I added the assertions to the partitioning rules instead.

transformer_engine/jax/cpp_extensions/activation.py
phu0ngng: Done.

PR #2366, opened 2025-11-11
[JAX] Improve support and testing for direct recipe usage without autocast contexts

jberchtold-nvidia:
# Description

This PR improves the support for directly passing a recipe when creating quantizers from `QuantizerFactory` or `TransformerEngineModuleBase.generate_quantizer_set`.

Known Limitation:
This limitation was not introduced by this PR, but is a part that is not yet addressed but will be improved in a future PR.
Under the specific condition of:
* An outer scope using an `autocast` context with a quantization recipe enabled
* An inner scope wanting to prevent quantization on one particular operation via directly passing the recipe None
This will not operate correctly as the quantizer factory cannot distinguish between not passing a recipe (recipe=None) and explicitly intending to disable quantization. We need an explicit NoScalingRecipe recipe here.

Though this case will be addressed in the future, fortunately, this case is not common and can be avoided by any of:
* (Recommended) Replacing the wide-scoped `autocast` context with direct recipe passing for finer granularity of control of where the recipe is applied
* Using a more narrow-scoped `autocast` context to wrap only the particular pieces of the model that are intended to be quantized. Be careful with Python contexts across JIT and VJP boundaries. The first solution of directly passing the recipe without a context may be simpler to use for this reason
* Creating quantizers directly with `ScalingMode.NO_SCALING`

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Remove `get_quantize_config()` and the global `_QUANTIZE_CONFIG` and instead replace with a global `get_global_quantize_recipe()`
- Remove existing usages of `get_quantize_config()` and ensure a recipe or quantize config is passed through explicitly to ensure both `autocast` context and directly passing in the recipe both work the same way. This ensures there are no hidden assumptions where we only read global state and we don't properly pass through configurations in the direct recipe approach
- Rework `test_recipe_characteristics.py` to test the following for all recipes:
     - Test recipe data is passed through correctly when using an `autocast` context
     - Within a sample model layer and VJP, ensure the quantizers are correctly created under both an `autocast` context and when directly passing the recipe to the quantizer factory

# Checklist:


transformer_engine/jax/quantize/quantizer.py
```diff
@@ -343,17 +343,35 @@ class DelayedScaleQuantizer(CurrentScaleQuantizer):
...
+        amax_history_len: Length of the amax history
         scale: Current scaling factor
         amax_history: History of maximum absolute values
     """
 
-    scaling_mode: ScalingMode = ScalingMode.DELAYED_TENSOR_SCALING
-    q_layout: QuantizeLayout = QuantizeLayout.ROWWISE_COLWISE
+    margin: float = 0.0
+    amax_compute_algo: AmaxComputeAlgo = AmaxComputeAlgo.MAX
+    amax_history_len: int = 1024
```
phu0ngng: Should we store the `amax_history_len` or should we simply have an init function and init to allocate the amax_history directly?

transformer_engine/jax/quantize/quantizer.py
```diff
@@ -1122,6 +1152,7 @@ def _create_set(
         bwd_dtype,
         is_2x2x,
         n_groups,
+        _is_inference_mode=False,
```
phu0ngng: Nit: I would remove the `_` so that the arg name is consistent with others.

transformer_engine/jax/quantize/quantizer.py
```diff
@@ -1200,25 +1232,34 @@ def create_set(
...
 
+        # TODO(jberchtold): Currently this is a limitation because we only support automatically populating quantizer fields based on a given recipe when using Flax. In the generic quantizer logic, we cannot assume Flax is being used, so we require the user to provide the quantize_meta_set created by quantize_config.get_quantize_flax_meta() or the same data created by themselves if they are passing a recipe here directly.
+        assert (
+            fp8_recipe is None or "quantize_meta_set" in kwargs
+        ), "When fp8_recipe is specified, quantize_meta_set must be provided in kwargs."
+
+        if fp8_recipe is None:
+            fp8_recipe = get_global_quantize_recipe()
+
         if fp8_recipe is not None:
```
phu0ngng, line 1251: Here, we no longer respect the case when a scaling mode is given but not the recipe, as the fp8_recipe is set to the current/default global recipe.

It is not necessarily a bad thing, however. 
I think it is actually cleaner to support creating quantizer sets based on an input recipe only, and remove the support for creating the quantizer sets based on the scaling modes. 
For individual unit tests for each scaling mode, we can still use the `create()`, which supports creating quantizers based on a given scaling mode.

transformer_engine/jax/quantize/quantizer.py
jberchtold-nvidia: Done

transformer_engine/jax/quantize/quantizer.py
jberchtold-nvidia: I've reverted back to the original where it is default field factory initialized unless specified. The tests added in `test_recipe_charateristics.py` test that the amax history has the proper shape based on the given recipe used

transformer_engine/jax/quantize/quantizer.py
jberchtold-nvidia: Good point. If you don't have a global recipe set then you can still specify the scaling mode directly here. However, I agree that this is not properly checked and functionality we wouldn't recommend users using. For now, I've added asserts that if a recipe is set directly or indirectly from an autocast context, the `scaling_mode`, `fwd_dtype` and `bwd_dtype` arguments must be `None`. In the future, I think we can remove these args completely.

PR #2370, opened 2025-11-12
FSDP2 Allgather Perf improvement and support for FusedAdam with FSDP2

vthumbe1503:
# Description
FSDP2 Allgather Perf improvement and support for FusedAdam with FSDP2

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- For FSDP2, FP8/MXFP8 tensors implement fsdp_pre_all_gather method that splits our tensors into uint8 tensor and metadata, so that it can be reconstructed into unsharded fp8/mxfp8 tensor after allgatherting Uint8 data. 
While constructing the metadata, we create a copy of the quantizer, for pre_allgather call in each iteration of training. This  PR fixes that to create copy only once to reduce CPU overhead.
  - **Before change:** FSDP::pre_forward time = 2.1ms,  **After change:** FSDP::pre_forward time = 1.9ms

 

- Using FusedAdam with FP8 FSDP2 results in segfault currently. Since we werent taking care of DTensor FP8 tensors(FSDP2 specific) while doing optimizer step and rather only FP8 tensors. This is fixed now and FusedAdam works as expected with fp8 and FSDP2.

# Checklist:

### DIFF
--- a/benchmarks/linear/benchmark_grouped_linear.py
+++ b/benchmarks/linear/benchmark_grouped_linear.py
@@ -45,6 +45,16 @@ nsys profile \
     --trace=cuda,nvtx,cudnn,cublas \
     python benchmarks/linear/benchmark_grouped_linear.py --profile --recipe nvfp4
 
+# Example for jagged input benchmark to simulate unbalanced token splits
+python benchmarks/linear/benchmark_grouped_linear.py --recipe nvfp4 --jagged-input "15296,8960,14656,14784,11712,7936,14080,10880"
+
+# Example to look at a single kernel target with NCU, like the fused hadamard amax kernel for NVFP4 recipe
+ncu -f -o ./benchmarks/linear/ncu_b200_numgemm_8_nvfp4_rht_amax \
+    --set=full \
+    --kernel-name "GroupHadamardAmaxTmaKernel" \
+    -s 5 -c 5 \
+    python benchmarks/linear/benchmark_grouped_linear.py --profile --recipe nvfp4 --profile
+
 """
 
 RECIPES = {
@@ -163,7 +173,7 @@ def benchmark_linear(
     return timing_ms
 
 
-def run_benchmark_linear(mkns, recipe_name, use_bias, num_gemms=4):
+def run_benchmark_linear(mkns, recipe_name, use_bias, num_gemms=4, m_splits=None):
     data = []
     assert not use_bias, "Bias is not supported for GroupedLinear benchmark"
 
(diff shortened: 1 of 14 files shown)


transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -715,7 +715,7 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
                 )
             # deep copy the scale inverse tensor and quantizer as well.
             scale_inv = tensor._scale_inv.detach().clone()
-            quantizer = tensor._quantizer.copy()
```
timmoon10, line 718: I'm uncomfortable with this change since it means that an FP8 tensor created with `zeros_like` will share the same quantizer, and so could have strange non-local interactions. I assume that we are trying to avoid CPU overhead in `Quantizer.copy`, but do we understand exactly why that function is slow?

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -552,7 +552,7 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
                 rowwise_scale_inv=rowwise_scale_inv,
                 columnwise_data=columnwise_data,
                 columnwise_scale_inv=columnwise_scale_inv,
-                quantizer=tensor._quantizer.copy(),
```
timmoon10, line 555: We should think more carefully before committing to this change. It doesn't make sense for the input and output of `zeros_like` to be entangled.

transformer_engine/pytorch/optimizers/fused_adam.py
```diff
@@ -11,6 +11,7 @@
 import warnings
 
 import torch
+from torch.distributed.tensor import DTensor
```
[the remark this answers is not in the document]
timmoon10: This is a nit, but still a good point. We nominally support PyTorch 2.1:
https://github.com/NVIDIA/TransformerEngine/blob/f8cb598c1c9b6228d39bf77cec3ea27c7ab6a1d7/build_tools/pytorch.py#L17

PyTorch 2.5 moved `DTensor` to `torch.distributed.tensor` and deprecated `torch.distributed._tensor`: https://github.com/pytorch/pytorch/pull/134203

Better to use the backward-compatible API for now, and we can revisit when we bump the minimum requirements to PyTorch 2.5+.
```suggestion
from torch.distributed._tensor import DTensor
```

transformer_engine/pytorch/tensor/mxfp8_tensor.py
vthumbe1503: The constructor already creates a copy of the quantizer before assigning to object. And that is why a duplicate copy is unecessary,

vthumbe1503: https://github.com/NVIDIA/TransformerEngine/blob/f8cb598c1c9b6228d39bf77cec3ea27c7ab6a1d7/transformer_engine/pytorch/tensor/storage/mxfp8_tensor_storage.py#L84

transformer_engine/pytorch/tensor/float8_tensor.py
vthumbe1503: Similar to the mxfp8_tensor comment. We are already creating a copy in the constructor of float8_tensor within the __new__ method, So no need for another copy.
https://github.com/NVIDIA/TransformerEngine/blob/f8cb598c1c9b6228d39bf77cec3ea27c7ab6a1d7/transformer_engine/pytorch/tensor/storage/float8_tensor_storage.py#L87

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -715,7 +715,7 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+            quantizer = tensor._quantizer
```
timmoon10: We should make sure comments are accurate:
```suggestion
            scale_inv = tensor._scale_inv.detach().clone()
            quantizer = tensor._quantizer  # Deep-copied within constructor
```

PR #2379, opened 2025-11-13
[JAX] Add CP + THD + AG + Striped>1 + SWA support

KshitijLakhani:
## Description

Add support for Striped>1 Reordering in TE JAX
Add support for CP + THD + AG + Striped>1 + SWA in TE JAX
Mask type: PADDING_CAUSAL_MASK

Below is an example of what the pattern would deconstruct into after load balancing and striping

Unbalanced matrix with 4 segments (64*64)
Screenshot 2025-11-25 at 10 02 30 PM

Balanced matrix with stripe_size=4, cp_size=4 (each rank works on 16*64 pattern) and SW
Screenshot 2025-11-25 at 10 05 31 PM

CP0: THD with 5 segments using PBRCM and SW
CP1: THD with 5 segments using PBRCM and SW
CP2: THD with 4 segments using PBRCM and SW
CP3: THD with 5 segments using PBRCM and SW

## Testing

Below are the timings for the minimal new test cases added 
- Added 320 L1 tests (52 passed, 268 skipped) for CP + THD + AG + Striped>1 + SWA
`NVTE_JAX_UNITTEST_LEVEL="L1" NVTE_JAX_TEST_TIMING=1 pytest -k "test_context_parallel_allgather_striped_attn" tests/jax/test_distributed_fused_attn.py`
```
================================================================================
TEST RUNTIME SUMMARY (grouped by function)
================================================================================
test_context_parallel_allgather_striped_attn                 | 320x |  388.31s | avg:   1.21s
================================================================================
TOTAL RUNTIME                                                |      |  388.31s |
================================================================================
```

- Added 64 L2 tests (24 passed, 136 skipped) for CP + THD + AG + Striped>1 + SWA
`NVTE_JAX_UNITTEST_LEVEL="L2" NVTE_JAX_TEST_TIMING=1 pytest -k "test_context_parallel_allgather_striped_attn" tests/jax/test_distributed_fused_attn.py`
```
================================================================================
TEST RUNTIME SUMMARY (grouped by function)
================================================================================
test_context_parallel_allgather_striped_attn                 | 160x |   77.24s | avg:   0.48s
================================================================================
TOTAL RUNTIME                                                |      |   77.24s |
================================================================================
```

- Added 12 L1 tests for Reordering with Striped>1
 `NVTE_JAX_TEST_TIMING=1 NVTE_JAX_UNITTEST_LEVEL="L1" pytest -k "TestReorderCausalLoadBalancing" tests/jax/test_distributed_fused_attn.py`
```
================================================================================
TEST RUNTIME SUMMARY (grouped by function)
================================================================================
test                                                         |  27x |    3.32s | avg:   0.12s
================================================================================
TOTAL RUNTIME                                                |      |    3.32s |
================================================================================
```

- I ran an additional testing sweep with randomized num_segments_per_seq = [2,3,4,9,11,14] and randomized seed = [12, 30, 42] and max_seqlens = [64, 1024, 2048] with all-to-all mapping between these three set of of params to ensure robustness and I saw no failures on B200x8

## Changes

- Added load balancing/reordering logice for stripe_size > 1
- For the CP+THD+AG+Striped>1+SWA primitive, the different types of deconstructed patterns per CP could not be fully expressed correctly by passing sharded seg ids and pos to get_seqlens_and_offsets() and hence helper functions had to be created for constructing the seqlens and offsets directly from the sharded seg ids and seg pos

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## TODO

- Optimize and compare this primitive perf to CP+THD+RingP2P+Striped+SWA for different data layouts

# Checklist:

### DIFF
--- a/docs/examples/attention/cp_ag_thd_dpa_jax_deep_dive.ipynb
+++ b/docs/examples/attention/cp_ag_thd_dpa_jax_deep_dive.ipynb
@@ -6,8 +6,20 @@
    "metadata": {},
    "source": [
     "## Deep Dive into CP + THD + AG + Striped>1 + SWA support for Transformer Engine JAX\n",
-    "This feature was merged as part of [PR 2379](https://github.com/NVIDIA/TransformerEngine/pull/2379/) and was made available Transformer Engine v2.11 onwards. This document will try to address 3 fundamental questions pertaining to the design considerations and the implementation logic for this feature"
+    "This feature was merged as part of [PR 2379](https://github.com/NVIDIA/TransformerEngine/pull/2379/) and was made available in Transformer Engine v2.11. This document addresses 3 fundamental questions about the design considerations and the implementation logic for this feature."
    ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "16f738c7",
+   "metadata": {
+    "vscode": {
+     "languageId": "plaintext"
+    }
+   },
+   "outputs": [],
+   "source": []
   },
   {
    "cell_type": "markdown",
@@ -16,11 +28,11 @@
    "source": [
     "### Question 1: Why choose Striped>1 ?\n",
     "\n",
-    "Prior to the addition of this feature, Transformer Engine JAX attention already supported load balancing via striping pattern, i.e., `stripe_size=1` for `CP + THD + P2P(Ring) + Striped + SWA`. However, this reordering technique does not lend itself well to an all gathered (post AG) pattern. Let's compare the two examples below to better understand this. For this example, `cp_size=4`, `num_segments=4`, `window_size=(8,0)`, and the pattern is for a single rank after striped reordering has been performed: \n",
+    "Prior to the addition of this feature, Transformer Engine JAX attention already supported load balancing via a striping pattern, i.e., `stripe_size=1` for `CP + THD + P2P(Ring) + Striped + SWA`. However, this reordering technique does not lend itself well to an all-gathered (post-AG) pattern. The following example illustrates this distinction. For this example, `cp_size=4`, `num_segments=4`, `window_size=(8,0)`, and the pattern is for a single rank after striped reordering has been performed: \n",
     "\n",
-    "#### I. Striped (`stripe_size = 1`)\n",
+    "#### I. Striped (`stripe_size=1`)\n",
     "- Such a staggered pattern is not supported by cuDNN\n",
-    "- One possible way to express this with cuDNN support is by treating each q token as a segment, thereby producing 16 segments with varying kv token counts. However, this is very inefficient and does not scale well when max_seqlens increases\n",
+    "- One possible way to express this with cuDNN support is by treating each `q` token as a segment, thereby producing 16 segments with varying `kv` token counts. However, this is very inefficient and does not scale well as max_seqlens increases\n",
     "\n",
     "```\n",
     "1 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -\n",
@@ -46,9 +58,9 @@
     "\n",
     "\n",
     "#### II. Striped > 1 (`stripe_size > 1`)\n",
-    "- This pattern is supported by cuDNN, with suggested `stripe_size=128`\n",
-    "- The mask type supported by `CP + THD + AG + Striped>1 + SWA` is `PADDING_CASUAL_MASK`, however, to express the below pattern, each rank executes THD + SWA using `PADDING_BOTTOM_RIGHT_CAUSAL_MASK`\n",
-    "- `max_num_segments_for_rank` needs to be estimated. The estimation formula used is: `max_seqlen // (stripe_size*cp_size) + max_num_segments`\n",
+    "- This pattern is supported by cuDNN, with a suggested `stripe_size=128`\n",
+    "- The mask type supported by `CP + THD + AG + Striped>1 + SWA` is `PADDING_CAUSAL_MASK`; however, to express the pattern below, each rank executes THD + SWA using `PADDING_BOTTOM_RIGHT_CAUSAL_MASK`\n",
+    "- `max_num_segments_for_rank` needs to be estimated. The estimation formula used is: `max_seqlens // (stripe_size * cp_size) + max_num_segments`\n",
     "\n",
     "```\n",
     "1 - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -\n",
@@ -108,7 +120,7 @@
     "<figcaption> Figure 3: Example 1 for problem using mask path in get_seqlens_and_offsets() for attention pattern (post striping and AG) . </figcaption>\n",
     "</figure>\n",
     "\n",
-    "Here, ideally, we should split the two sections of the segment 3 into two different segments (segment 3_1 formed using rows 9-12 and segment 3_2 formed using rows 13-16) as cuDNN does not support segment 3's entire staggered shape, as discussed earlier, however, the mask route is unable to make this distinction, and it ends up treating it as one large segment thereby performing unnecessary compute of padded regions in segment 3(rows 9-12 )\n",
+    "Here, ideally, the two sections of the segment 3 should be split into two different segments (segment 3_1 formed using rows 9-12 and segment 3_2 formed using rows 13-16) as cuDNN does not support segment 3's entire staggered shape (as discussed earlier) , however, the mask route is unable to make this distinction, and it ends up treating it as one large segment thereby performing unnecessary computations of the padded regions in segment 3(rows 9-12 )\n",
     "\n",
     "In the below example, the mask route takes the `kv_seqlens` for segment 1 to be 6 and masks it using Bottom Right Causal Mask rather than taking `kv_seqlens` of 4 and masks it using Bottom Right Causal Mask, resulting in incorrect results\n",
     "\n",
@@ -144,7 +156,7 @@
    "source": [
     "### Question 3: What is the implementation logic for the separate helper functions ?\n",
     "\n",
-    "We'll discuss the implementation logic for 2 of these 4 helper functions to use as a reference, as the other two are using similar principles. Lets use the below test example, for which, `cp_size=4`, `stripe_size=4`, `max_seqlens=64`, `num_segments=2`and no SWA for simplicity. seg_1 has 8 valid tokens + 13 padded tokens ad seg_2 has 31 valid tokens and 1 padding token. The 0 is used to explicitly show the padded region of seg_1 which as reordered, but for computation purposes it is equivalent to the any of the `-` marked elements.\n",
+    "This section discusses the implementation logic for two of these four helper functions which serve as a reference, as the other two are using similar principles. Consider the test example in the code block, for which, `cp_size=4`, `stripe_size=4`, `max_seqlens=64`, `num_segments=2` and no SWA for simplicity. seg_1 has 8 valid tokens + 13 padded tokens and seg_2 has 31 valid tokens + 1 padded token. The 0 is used to explicitly show the padded region of seg_1 which is reordered, but for computation purposes it is equivalent to any of the `-` marked elements.\n",
     "\n",
     "```\n",
     "segment_ids_q_0_reordered = segment_ids_kv_0_reordered = jnp.array([[1, 1, 1, 1, 0, 0, 0, 0, 2, 2, 2, 2, 2, 2, 2, 2]])\n",
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -56,3 +56,4 @@ Transformer Engine documentation
    api/c/index
    debug
    examples/attention/attention.ipynb
+   examples/attention/cp_ag_thd_dpa_jax_deep_dive.ipynb


KshitijLakhani:
> /te-ci jax L0 L1 L2

Pipeline: 39181941 passes all tests (except lint test - to be fixed in a separate commit)

huanghua1994 (approved):

KshitijLakhani:
> /te-ci jax L0 L1 L2

Pipeline: 39192121 passes all L0, L1 and L2 tests

tests/jax/test_distributed_fused_attn.py
```diff
@@ -327,7 +327,7 @@ def test_cross_attn(
 ]
 
 DISTRIBUTED_CONTEXT_SELF_ATTN_DATA_SHAPES = [
-    # Sequence lengths will be scaled by CP so that we don't run with tiny sizes.
+    # Sequence lengths will be scaled by CP*2 so that we don't run with tiny sizes.
     pytest.param([2, 128, 8, 128], id="2-128xCP-8-128"),
     pytest.param([4, 256, 16, 64], id="4-256xCP-16-64"),
```
jberchtold-nvidia: nit: update pytest param id to include `CPx2`

tests/jax/test_distributed_fused_attn.py
```diff
@@ -351,12 +351,14 @@ def impl_test_context_parallel_attn(
...
     ):
         if qkv_layout.is_thd():
-            if cp_strategy == CPStrategy.ALL_GATHER:
-                pytest.skip("THD doesn't support all gather context parallelism.")
-            if not load_balanced and cp_strategy == CPStrategy.RING:
-                pytest.skip("THD + ring doesn't support unbalanced context parallelism.")
+            if not load_balanced and (
+                cp_strategy == CPStrategy.RING or cp_strategy == CPStrategy.ALL_GATHER
+            ):
+                pytest.skip(f"THD + {cp_strategy=} doesn't support unbalanced context parallelism.")
```
jberchtold-nvidia, line 361: today I learned `{var=}` does `var={var}` in f-strings. useful!

tests/jax/test_distributed_fused_attn.py
```diff
@@ -457,7 +460,7 @@ def check_has_backend_for_mask(mask_type):
     @pytest.mark.parametrize("dtype", [pytest.param(jnp.bfloat16, id="BF16")])
     @pytest.mark.parametrize(
         "qkv_layout, attn_mask_type",
-        DISTRIBUTED_CONTEXT_SELF_ATTN_LAYOUTS_MASKS,
+        DISTRIBUTED_CONTEXT_SELF_ATTN_LAYOUTS_MASKS[:-1],
```
jberchtold-nvidia: Why are we skipping a mask here?

jberchtold-nvidia: Is this because the new test below is explicitly testing the index=-1 mask case? If so, can we remove this last mask from the `DISTRIBUTED_CONTEXT_SELF_ATTN_LAYOUTS_MASKS` and create a new list of masks for striped below?

tests/jax/test_distributed_fused_attn.py
```diff
@@ -486,6 +489,72 @@ def test_context_parallel_allgather_attn_shardy(
...
+        "qkv_layout, attn_mask_type",
+        [DISTRIBUTED_CONTEXT_SELF_ATTN_LAYOUTS_MASKS[-1]],
+    )
+    @pytest.mark.parametrize(
+        "load_balanced",
+        [pytest.param(True, id="BALANCED")],
+    )
+    @pytest.mark.parametrize(
+        "stripe_size",
+        [pytest.param(64, id="STRIPE-64"), pytest.param(128, id="STRIPE-128")],
```
jberchtold-nvidia, line 511: Doesn't have to be in this PR since I'm guessing this applies to more tests than updated here, but TE/JAX has a pytest util called `pytest_parametrize_wrapper` that automatically converts common types into string representations like `stripe_64` so you don't need to list these manually

tests/jax/test_distributed_fused_attn.py
```diff
@@ -486,6 +489,72 @@ def test_context_parallel_allgather_attn_shardy(
...
+        kv_groups,
+        attn_mask_type,
+        dtype,
+        qkv_layout,
+        load_balanced,
+        window_size,
+        stripe_size,
+        num_segments_per_seq,
+    ):
+        if window_size != (-1, -1) and not qkv_layout.is_thd():
```
jberchtold-nvidia: So BSHD with window_size = (-1, -1) is supported? Does that mean window_size (-1, -1) means don't do any striping?

tests/jax/test_distributed_fused_attn.py
```diff
@@ -495,7 +564,7 @@ def test_context_parallel_allgather_attn_shardy(
     @pytest.mark.parametrize("dtype", [pytest.param(jnp.bfloat16, id="BF16")])
     @pytest.mark.parametrize(
         "qkv_layout, attn_mask_type",
-        DISTRIBUTED_CONTEXT_SELF_ATTN_LAYOUTS_MASKS,
+        DISTRIBUTED_CONTEXT_SELF_ATTN_LAYOUTS_MASKS[:-1],
```
jberchtold-nvidia: Same comment as above with this [:-1] slicing of the masks. If this array of masks [:-1] is for one feature and [-1:] is for this new CP striped attention feature, let's split them into two constant lists.

tests/jax/test_fused_attn.py
```diff
@@ -230,6 +230,7 @@ def make_mask(
 @jax.jit
 def get_seqlens_and_offsets(segment_ids):
     batch, max_seqlen = segment_ids.shape
+    # TODO: should this be max_seqlen + 1 ?
```
jberchtold-nvidia: reminder about TODO

tests/jax/test_fused_attn.py
```diff
@@ -577,7 +580,9 @@ def generate_random_segment_ids(
             return segment_ids, segment_pos, segment_pad
 
         if self.qkv_layout.is_thd():
-            self.num_segments_per_seq = 2
+            # If using default num segments of 0, set to 2
+            if self.num_segments_per_seq == 0:
```
jberchtold-nvidia: Why is the default overridden here? Can we make the default `None` instead of `0` to indicate it's not yet populated?

tests/jax/test_fused_attn.py
```diff
@@ -771,7 +778,7 @@ def to_dp_shardings(x):
 
     def test_forward(self):
         """
-        Test forward without JIT
+        Test forward with JITted primitive and unJITted reference
```
jberchtold-nvidia, line 784: Is there a reason the reference JAX impl is unjitted? It should be equivalent JIT'd and could speed up our tests. Ack, this is unrelated to this PR's focus, just mentioning it so we can discuss if this could be improved in a separate PR

tests/jax/test_distributed_fused_attn.py
```diff
@@ -327,7 +327,7 @@ def test_cross_attn(
 ]
```
jberchtold-nvidia, line 328: Thanks for measuring the test runtime! What is our total L1 test time at currently? We were intermittently failing previously but reduced attention tests by ~15mins iirc. This is now increasing by 11mins, so I'm concerned we may hit timeouts again.

```
================================================================================
TEST RUNTIME SUMMARY (grouped by function)
================================================================================
test_context_parallel_allgather_striped_attn                 | 128x |  719.65s | avg:   5.62s
================================================================================
TOTAL RUNTIME                                                |      |  719.65s |
================================================================================
```

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -1272,26 +1274,44 @@ def check_supported(self):
...
+        if self.config.qkv_layout.is_thd():
+            allowed_masks.append(AttnMaskType.PADDING_CAUSAL_MASK)
         if self.config.attn_mask_type not in allowed_masks:
             raise ValueError(
                 f"{header} only supports masking types: "
                 f" {','.join(map(str, allowed_masks))} got: {self.config.attn_mask_type}"
             )
+        # Do not allow CP + AG + THD + Striped with NO_MASK
+        if self.config.attn_mask_type is AttnMaskType.NO_MASK and self.config.qkv_layout.is_thd():
+            raise ValueError(f"{header} only supports CAUSAL_MASK for THD types")
```
jberchtold-nvidia: nit: We also support PADDING_CAUSAL_MASK for THD in addition to CAUSUAL_MASK, based on lines 1301-1302, right?

tests/jax/test_distributed_fused_attn.py
KshitijLakhani: I think with the merging of this PR we will be around the 80-85 mins mark for the L1 tests, so I do not expect to hit timeouts (closer to 120 mins)

I believe after your changes to reduce the L1 test timing we had stopped hitting the limit anyways. My reduction to attention tests was more of an additional effort and we reduced total time to closer to 70 mins if I remember right.

Now with sink attention and this PR, we should have a total increase of about ~15 mins so I expect to hit ~85 mins. Nonethless, I will report the findings from the last CI pipeline I run.

I do not think we are alarmingly close but a clean up in the future would only help.

tests/jax/test_distributed_fused_attn.py
KshitijLakhani: I will update this check to :  `if window_size != (-1, -1) and not qkv_layout.is_thd():`
Any BSHD layouts should be skipped
Only THD with or without SWA should be allowed
Thanks for pointing it out

I believe my parametrization of the inputs:  [DISTRIBUTED_CONTEXT_SELF_ATTN_LAYOUTS_MASKS[-1]], did not trigger a BSHD mask anyways so this check never got triggerred.

tests/jax/test_distributed_fused_attn.py
KshitijLakhani: I will revert this back as it makes no functional difference
The THD data layout (which we are trying to explicitly skip with this change) is anyways being skipped as part of the tests, so it isn't fully needed

tests/jax/test_distributed_fused_attn.py
KshitijLakhani: I will revert this back as it makes no functional difference
The THD data layout (which we are trying to explicitly skip with this change) is anyways being skipped as part of the tests, so it isn't fully needed

cyanguwa:
I think it's `Striped>=1` that we support, right? We can probably mention that `>=128` is recommended as well somewhere, for performance reasons.

Did you check the CP path in the CP vs non-CP comparison in the unit tests really uses the cuDNN backend? Just trying to make sure it's not using the unfused path.

tests/jax/test_distributed_fused_attn.py
KshitijLakhani: However, the logic to skip this for THD data layout in impl_test_context_parallel_attn() has changed and so just reverting the change will not be enough to skip this for THD data types, so I will add an additional check in test_context_parallel_allgather_attn_shardy() and test_context_parallel_allgather_attn() to skip for THD layouts

KshitijLakhani: In either case it is just a different way to put things - the older change was kipping THD layouts by filtering them via DISTRIBUTED_CONTEXT_SELF_ATTN_LAYOUTS_MASKS[:-1] but no we will pass DISTRIBUTED_CONTEXT_SELF_ATTN_LAYOUTS_MASKS and instead filter within test_context_parallel_allgather_attn_shardy() and test_context_parallel_allgather_attn() explicitly

tests/jax/test_distributed_fused_attn.py
KshitijLakhani: https://github.com/NVIDIA/TransformerEngine/pull/2379/files#r2566226953
All the comments in the earlier thread apply here

tests/jax/test_fused_attn.py
KshitijLakhani: I am removing this for now as it is not related to this PR.
Rather a general note for me - I'll take a look at this outside the scope of this PR

transformer_engine/jax/attention.py
```diff
@@ -1169,6 +1179,7 @@ def fused_attn(
     context_parallel_axis: str = "",
     context_checkpoint_name: str = "context",
     softmax_offset: Optional[jnp.ndarray] = None,
+    stripe_size: int = 0,
```
cyanguwa: Should the default be 1 or 0?

tests/jax/test_fused_attn.py
KshitijLakhani: Yes I will make this change. It was in my TODO but missed it.
Better that ways

tests/jax/test_fused_attn.py
KshitijLakhani: Unsure tbh. I was expecting it to be JItted as well but when I noticed it wasn't I thought it should at least be explicit in the doc string to address later. Agree with you

transformer_engine/jax/cpp_extensions/attention.py
cyanguwa: We support PADDING_MASK as well, right?

KshitijLakhani: This should change to :         

if self.config.attn_mask_type is not AttnMaskType.PADDING_CAUSAL_MASK and self.config.qkv_layout.is_thd():
            raise ValueError(f"{header} only supports PADDING_CAUSAL_MASK for THD types")

This function is in the AG helper class which was originally only supporting BSHD + AG + Dual Chunk.
I believe the AG BSHD supports `allowed_masks = [AttnMaskType.NO_MASK, AttnMaskType.CAUSAL_MASK]`

This PR only adds support for `AttnMaskType.PADDING_CAUSAL_MASK` as it would be the larger case but extending support to `AttnMaskType.CAUSAL_MASK` in the future should not be too hard

To summarize, the CP + AG has two supported cases:
1. BSHD + AG + Dual Chunk + (no mask + causal)
2. THD + AG + Striped>1 + (padding causal) - this PR

KshitijLakhani:
> I think it's `Striped>=1` that we support, right? We can probably mention that `>=128` is recommended as well somewhere, for performance reasons.

Sure let me mention that in one of the comments

> Did you check the CP path in the CP vs non-CP comparison in the unit tests really uses the cuDNN backend? Just trying to make sure it's not using the unfused path.

Yes it does use the fused cudnn backend for sure.
The entire debugging process relied on checking seqlens and offsets being passes to the fused cuDNN API in TE common

transformer_engine/jax/attention.py
KshitijLakhani: `stripe_size` default is set to 0 or 1 in different function calls based on context:
1. `stripe_size=1` is set as default in the load balancing API (`reorder_causal_striped()` / `reorder_causal_load_balance()`) - this is done on purpose so that existing users of this API do not have a breaking change. They can continue using these API calls where the default striping of 1 will be used - the P2P ring case
2. `stripe_size=0` is set as default in all other functions whereby `stripe_size=0` is loosely used as a way to indicate a BSHD usage, because, a THD usage (with P2P or AG) will implicitly set the `stripe_size` to non-zero via the reorder functions mentioned above or explicitly by setting it. 
An argument can be made for using `None` instead of `0` for BSHD cases - so I will make that change to reduce the confusion. However, we will still have two set of values for default - None and 1

tests/jax/test_distributed_fused_attn.py
```diff
@@ -327,9 +327,9 @@ def test_cross_attn(
...
```
jberchtold-nvidia, line 327: There is an empty file `texts/jax/test`. I can't seem to comment on the file directly since it has 0 lines, so commenting here

transformer_engine/jax/attention.py
```diff
@@ -386,23 +386,31 @@ def _obtain_batch_and_max_seqlen(qkv, qkv_layout):
...
+    tensor, strategy: ReorderStrategy, cp_size: int, seq_dim: int, stripe_size: int = 1
+):
     """Reorders a tensor for load balancing the compute of causal attention."""
     if strategy == ReorderStrategy.DualChunkSwap:
         return tex.attention.reorder_causal_dual_chunk_swap(tensor, cp_size, seq_dim, False)
     if strategy == ReorderStrategy.Striped:
-        return tex.attention.reorder_causal_striped(tensor, cp_size, seq_dim, False)
+        # stripe_size > 1 is only supported for CP+THD+AG+Striped>1+SWA
+        # stripe_size = 128 is recommended for CP+THD+AG+Striped>1+SWA
+        if stripe_size == 0:
```
mgoldfarb-nvidia: nit: `<= 0`

transformer_engine/jax/attention.py
```diff
@@ -386,23 +386,31 @@ def _obtain_batch_and_max_seqlen(qkv, qkv_layout):
...
 def inverse_reorder_causal_load_balancing(
-    tensor, strategy: ReorderStrategy, cp_size: int, seq_dim: int
+    tensor, strategy: ReorderStrategy, cp_size: int, seq_dim: int, stripe_size: int = 1
 ):
     """Inverse operation of `reorder_causal_load_balancing`."""
     if strategy == ReorderStrategy.DualChunkSwap:
         return tex.attention.reorder_causal_dual_chunk_swap(tensor, cp_size, seq_dim, True)
     if strategy == ReorderStrategy.Striped:
-        return tex.attention.reorder_causal_striped(tensor, cp_size, seq_dim, True)
+        # stripe_size > 1 is only supported for CP+THD+AG+Striped>1+SWA
```
mgoldfarb-nvidia, line 428: nit: add same check to `stripe_size` as in `reorder_causal_load_balancing`

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -1234,31 +1234,35 @@ def reorder_causal_dual_chunk_swap(tensor, cp_size: int, seq_dim: int, to_contig
...
 
 
-def reorder_causal_striped(tensor, cp_size: int, seq_dim: int, is_inverse: bool):
+def reorder_causal_striped(
+    tensor, cp_size: int, seq_dim: int, is_inverse: bool, stripe_size: int = 1
+):
     """Reorders a tensor for load balancing with striped pattern"""
     origin_shape = tensor.shape
-    if origin_shape[seq_dim] % cp_size != 0:
+    if stripe_size == 0:
```
mgoldfarb-nvidia: nit `<=`

tests/jax/test_distributed_fused_attn.py
KshitijLakhani: Thanks for catching it. Removed

docs/examples/attention/cp_ag_thd_dpa_jax_deep_dive.ipynb
cyanguwa, line 1: Could you move this notebook and its related pictures to a sub-folder, possibly "deep_dive_Jax_CP_AllGather_THD"?

cyanguwa (dismissed):

mgoldfarb-nvidia (approved):

docs/examples/attention/cp_ag_thd_example.png
jberchtold-nvidia, line 1: Do we want a png here or would it be better to embed the diagram as text in the notebook itself? With the text directly, it may be easier for users to read (light mode / dark mode, font size, etc.), but lmk if the notebook was introducing some unintended affects like line wrapping that broke the diagram

KshitijLakhani: I didn't try it actually, because I'd have to generate some of those again, but I see your point. Let me give it a shot.
I do not really expect users of TE to look at this document, but, contributors to TE might. Either ways, let me try it out !
You're suggesting to type out the text in the file in a markdown code block, right ? Or something else ?

jberchtold-nvidia: Yep, exactly, just putting it as a markdown. Maybe a markdown block inside the notebook then a triple backtick codeblock inside that, so it shows up in a monospaced font but doesn't show up as runnable code.

If for some reason it doesn't work due to formatting, then we don't need to block the PR on it. But if we can avoid code as a screenshot, I think it'd be nice. Thanks!

KshitijLakhani: Sounds reasonable. I agree !
Let me push the change
Thanks !

cyanguwa (approved):

jberchtold-nvidia (approved):

PR #2388, opened 2025-11-14
[Common] NVTEGroupedTensor class and helpers

phu0ngng:
# Description

NVTEGroupedTensor class and helpers 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/common/transformer_engine.cpp
+++ b/transformer_engine/common/transformer_engine.cpp
@@ -526,7 +526,6 @@ class GroupedTensorAllocator {
       NVTEGroupedTensor ret = reinterpret_cast<NVTEGroupedTensor>(index);
       free_list.pop_back();
       // 1-based indexing - fully reinitialize the tensor to avoid stale data
-      memory[index - 1].clear();
       memory[index - 1].scaling_mode = mode;
       memory[index - 1].num_tensors = num_tensors;
       memory[index - 1].logical_shape = logical_shape;


transformer_engine/common/common.h
```diff
@@ -281,6 +281,185 @@ struct Tensor {
   }
 };
 
+struct SimpleGroupedTensor {
```
ptrendx: If possible I would like to avoid another SimpleTensor-like struct. The main reason for this one seems to be the ability to support both "multiple tensors in a single allocation" and "multiple tensors in multiple allocations" cases. Do we really want to support the multiple allocations case though? The GEMM will work best when there is only 1 allocation anyway, so I don't think we need to overgeneralize this API - we can always add that multiallocation mode later.

transformer_engine/common/common.h
```diff
@@ -281,6 +281,185 @@ struct Tensor {
...
+ private:
+  std::vector<void *> dptr_list{};
+  void *base_dptr{nullptr};
+
+ public:
+  DType dtype;
+  size_t num_tensors;  // Only need for error checking
+  bool contiguous;
```
ptrendx: What exactly is this for?

zhongbozhu:
LGTM as discussed offline. 

Will be better if we can have some example usage of the new API. Otherwise, the new checkGroupOutputTensor seems complicated and I am not sure if it's too strict.

transformer_engine/common/common.h
```diff
@@ -281,6 +281,185 @@ struct Tensor {
...
+  // TODO: Discuss if we can still get num_elements for each field so that we can perform shape/size validation for each field.
+  size_t
+      sum_first_dims;  // Need for block scaling scale error checking and GroupedGEMM hints (only the data shapes)
+  size_t sum_second_dims;
```
ptrendx: I don't like this. First of all, you don't actually know those values (or at least one of them) if the shapes are on the GPU. The only thing you know is the shape of the allocation. It is also something that the columnwise tensor will duplicate with the rowwise (but with different meaning).
You also can't really use that for error checking (if both dimensions are variable then those sums do not really give you much without at least the size of the allocation, which is missing in this class).

transformer_engine/common/common.h
```diff
@@ -281,6 +281,185 @@ struct Tensor {
...
 };
 
+struct SimpleGroupedTensor {
+ private:
+  std::vector<void *> dptr_list{};
+  void *base_dptr{nullptr};
+
+ public:
+  DType dtype;
+  size_t num_tensors;  // Only need for error checking
```
ptrendx: I don't believe this is just for error checking and it should be in the GroupedTensor so that we could use regular SimpleTensor here.

transformer_engine/common/common.h
```diff
@@ -281,6 +281,185 @@ struct Tensor {
...
+  SimpleGroupedTensor columnwise_scale_inv;
+  SimpleGroupedTensor amax;
+  SimpleGroupedTensor columnwise_amax;
+  SimpleGroupedTensor scale;  // for FP8-DS only
+
+  // [TODO] Discuss whether the first_dims and second_dims should be according to layout N
+  // Shape information: first_dims[i] and second_dims[i] define the shape of the i-th tensor
+  // For 2D tensors: shape[i] = (first_dims[i], second_dims[i])
+  SimpleTensor first_dims;   // Device pointer to size_t array of length num_tensors
+  SimpleTensor second_dims;  // Device pointer to size_t array of length num_tensors
```
ptrendx: The way I think about it is:
 - we need to standardize first on which direction is used to say what is "first" and what is "second" (I prefer "last" BTW) -> I vote for rowwise
 - then my thinking is that if the rowwise allocation has shape [m, k], then existence (or not) of those shapes would tell us which of the dimension is constant (e.g. second_dim being noninitialized would mean that all tensors are of shape [m_i, k]), which could be used for additional optimizations (e.g. via specialized kernel choice).

Alternatively the "reference" shape could be a property of the GroupedTensor itself just to avoid setting the shape on otherwise uninitialized rowwise tensor.

transformer_engine/common/common.h
```diff
@@ -281,6 +281,185 @@ struct Tensor {
...
+        scaling_mode(scaling_mode),
+        nvte_tensor(0) {}
+
+  explicit operator NVTEGroupedTensor() const noexcept { return nvte_tensor; }
+
+  bool has_data() const noexcept { return data.has_data(); }
+  bool has_columnwise_data() const noexcept { return columnwise_data.has_data(); }
+  bool has_cumulative_tensor_sizes() const noexcept {
+    return cumulative_tensor_sizes.dptr != nullptr;
+  }
```
ptrendx: Without those the tensor would be kind of unusable though - it should always have them.

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -393,6 +393,105 @@ int nvte_is_non_tn_fp8_gemm_supported();
...
+  kNVTEGroupedAmax = 3,               /*!< Amax tensor */
+  kNVTEGroupedRowwiseScaleInv = 4,    /*!< Scale inverse tensor for decoding Rowwise Data */
+  kNVTEGroupedColumnwiseScaleInv = 5, /*!< Scale inverse tensor for decoding Columnwise Data */
+  kNVTEGroupedColumnwiseAmax = 6,     /*!< Columnwise Amax tensor */
+  kNVTEGroupedFirstDims = 7,          /*!< First dimension sizes (device pointer to size_t array) */
+  kNVTEGroupedSecondDims = 8,      /*!< Second dimension sizes (device pointer to size_t array) */
+  kNVTEGroupedCumulativeSizes = 9, /*!< Cumulative tensor sizes (device pointer to size_t array) */
+  kNVTENumGroupedTensorParams
+};
+
```
ptrendx, line 420: I would prefer to avoid completely forking off the API for the grouped tensor from the regular tensor - most of those parameters are the same between the two and have conceptually the same meaning. Also the comment is confusing - we do not want the user to provide a literal size_t* value so we should not tell them so.

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -393,6 +393,105 @@ int nvte_is_non_tn_fp8_gemm_supported();
...
+struct NVTEGroupedTensorInfo {
+  void *base_dptr;        /*!< Base pointer for contiguous layout, or nullptr */
+  void **dptr_list;       /*!< Array of pointers for non-contiguous layout, or nullptr */
+  NVTEDType dtype;        /*!< Data type */
+  size_t num_tensors;     /*!< Number of tensors in the group */
+  bool contiguous;        /*!< Whether data is contiguous */
+  size_t sum_first_dims;  /*!< Sum of first dimensions across all tensors (for validation) */
+  size_t sum_second_dims; /*!< Sum of second dimensions across all tensors (for validation) */
+};
+
```
ptrendx: Hopefully we can get rid of the SimpleGroupedTensor type and then this would not be needed either.

transformer_engine/common/transformer_engine.cpp
```diff
@@ -273,6 +273,284 @@ void CheckOutputTensor(const Tensor &t, const std::string &name, bool allow_empt
...
+    NVTE_CHECK(t.scale.num_tensors == expected_num_tensors, "Grouped tensor ", name,
+               " scale.num_tensors (", t.scale.num_tensors,
+               ") doesn't match expected num_tensors (", expected_num_tensors, ")");
+  }
+
+  // Check that shape information is provided and consistent
+  NVTE_CHECK(t.first_dims.dptr != nullptr, "Grouped tensor ", name,
+             " must have first_dims allocated");
+  NVTE_CHECK(t.second_dims.dptr != nullptr, "Grouped tensor ", name,
+             " must have second_dims allocated");
```
ptrendx: I don't believe we need both of them.

transformer_engine/common/transformer_engine.cpp
```diff
@@ -273,6 +273,284 @@ void CheckOutputTensor(const Tensor &t, const std::string &name, bool allow_empt
...
+
+  // Ensure shape arrays are 1D
+  NVTE_CHECK(t.first_dims.shape.size() == 1, "Grouped tensor ", name,
+             " first_dims must be 1D (got ", t.first_dims.shape.size(), "D)");
+  NVTE_CHECK(t.second_dims.shape.size() == 1, "Grouped tensor ", name,
+             " second_dims must be 1D (got ", t.second_dims.shape.size(), "D)");
+
+  // Ensure shape arrays have correct dtype (size_t = Int64)
```
ptrendx: nit: Well, to be precise Int64 is not quite size_t :-).

transformer_engine/common/transformer_engine.cpp
```diff
@@ -273,6 +273,284 @@ void CheckOutputTensor(const Tensor &t, const std::string &name, bool allow_empt
...
+  NVTE_CHECK(!t.second_dims.shape.empty() && t.second_dims.shape[0] > 0, "Grouped tensor ", name,
+             " second_dims has invalid shape");
+  NVTE_CHECK(t.first_dims.shape[0] == t.second_dims.shape[0], "Grouped tensor ", name,
+             " first_dims and second_dims must have the same size (got ", t.first_dims.shape[0],
+             " vs ", t.second_dims.shape[0], ")");
+  NVTE_CHECK(t.first_dims.shape[0] == t.num_tensors(), "Grouped tensor ", name,
+             " shape arrays size (", t.first_dims.shape[0], ") doesn't match num_tensors (",
+             t.num_tensors(), ")");
+
+  // Check cumulative_tensor_sizes if it's set (optional but must be consistent)
```
ptrendx: Why would it be optional? We need to know the offsets to be able to do anything interesting and for that we need the allocation.

transformer_engine/common/transformer_engine.cpp
```diff
@@ -273,6 +273,284 @@ void CheckOutputTensor(const Tensor &t, const std::string &name, bool allow_empt
...
+               " amax.num_elements (", t.amax.num_elements,
+               ") doesn't match expected num_tensors (", expected_num_tensors, ")");
+  } else if (t.scaling_mode == NVTE_NVFP4_1D_SCALING || t.scaling_mode == NVTE_MXFP8_1D_SCALING) {
+    if (t.has_data()) {
+      NVTE_CHECK(
+          !t.scaling_mode == NVTE_NVFP4_1D_SCALING || t.amax.num_elements == expected_num_tensors,
+          "Grouped tensor ", name, " amax.num_elements (", t.amax.num_elements,
+          ") doesn't match expected num_tensors (", expected_num_tensors, ")");
+      auto scale_shape = get_block_scale_shape(t.scaling_mode, t.data.sum_first_dims,
+                                               t.data.sum_second_dims, false);
```
ptrendx: This is not true - you can't just pad the sum of the dimensions. The proper size is $`128*4*\sum_i \lceil \frac{m_i}{128}\rceil\lceil\frac{k_i}{128}\rceil`$ (note that I said size and not shape since the shape does not really have meaning here) which is more than that. In practice we do not really have the means (and frankly, I don't think we need that) to properly check that the allocations here are correct and should rely on the caller, who has additional information (like the fact that some dimensions are constant or that they just downloaded the dimensions to the CPU and checked), to do the right thing.

transformer_engine/common/transformer_engine.cpp
```diff
@@ -387,6 +665,86 @@ Tensor *convertNVTETensorCheck(const NVTETensor t) {
...
+
+ private:
+  GroupedTensorAllocator() {
+    std::lock_guard<std::mutex> lock(mutex);
+    memory.reserve(MAX_GROUPED_TENSOR_NUM);
+  }
+
+  std::mutex mutex;
+  std::atomic<size_t> size;
+  // Allocate at most 5 MB for grouped tensors
```
ptrendx: Why 5? The regular one has 20.

transformer_engine/common/common.h
```diff
@@ -281,6 +281,81 @@ struct Tensor {
...
+  3. Shape information stored in first_dims and second_dims arrays
+  */
+
+  SimpleTensor data;
+  SimpleTensor columnwise_data;
+  SimpleTensor scale_inv;
+  SimpleTensor columnwise_scale_inv;
+  SimpleTensor amax;
+  SimpleTensor columnwise_amax;
+  SimpleTensor scale;  // for FP8-DS only
```
timmoon10, line 318: Having a giant pile of variables is fine since we want to make progress on this quickly, but in the future we should consider refactoring to handle polymorphism more gracefully:
```c++
// Visitor pattern
struct GroupedTensor {
 public:
  struct FP8Data {
    std::optional<SimpleTensor> data;
    std::optional<SimpleTensor> transpose;
    SimpleTensor scale_inv;
    std::optional<SimpleTensor> amax;
    std::optional<SimpleTensor> scale;
  };
  struct MXFP8Data {
    std::optional<std::tuple<SimpleTensor, SimpleTensor>> rowwise_data_and_scale;
    std::optional<std::tuple<SimpleTensor, SimpleTensor>> columnwise_data_and_scale;
  }
  std::variant<FP8Data, MXFP8Data, ...> data;
};

// Inheritance pattern
struct GroupedTensor { ... };
struct FP8GroupedTensor : public GroupedTensor {
    std::optional<SimpleTensor> data;
    std::optional<SimpleTensor> transpose;
    SimpleTensor scale_inv;
    std::optional<SimpleTensor> amax;
    std::optional<SimpleTensor> scale;
};
struct MXFP8GroupedTensor : public GroupedTensor {
    std::optional<std::tuple<SimpleTensor, SimpleTensor>> rowwise_data_and_scale;
    std::optional<std::tuple<SimpleTensor, SimpleTensor>> columnwise_data_and_scale;
};
```

transformer_engine/common/common.h
timmoon10: - I'm interpreting these dims as the logical tensor dims, which matches the row-wise data dims. Logical dims are completely independent on the data format, regardless of whether the column-wise data is transposed or not.
- I like the idea of the grouped tensor holding the "reference" shape and using it depending on whether `first_dim`/`second_dim` are empty. I don't think we can rely on the shape of the data tensors since they might need to be flattened to 1D, e.g. FP8 transpose when splitting along the first dim.
- Being able to split along 2 dims makes this very general. However, for MoE we always split along the first _logical_ dim (PyTorch has column-major weights and JAX has row-major, so we need to swap the usage of row-wise and column-wise data. However, we are still splitting along the first logical dim). We should decide whether future-proofing is worth the extra complexity.

transformer_engine/common/transformer_engine.cpp
timmoon10: I'd prefer if we changed all the comments and docs that reference a "size_t array" to say "int64 array".

transformer_engine/common/common.h
timmoon10: Actually, on second thought I think my last bullet point is still true. For MoE, we are _always_ splitting along the first logical dim. Internally we might be splitting a transposed matrix along the last dim, but that is a detail within the group tensor class, and shouldn't be exposed in the public API.

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -393,6 +393,106 @@ int nvte_is_non_tn_fp8_gemm_supported();
...
+void nvte_destroy_grouped_tensor(NVTEGroupedTensor tensor);
+
+/*! \brief Set a parameter of the grouped tensor.
+ *
+ *  \param[in/out] tensor Grouped tensor.
+ *  \param[in] param_name The parameter to be set.
+ *  \param[in] param The value to be set (NVTEBasicTensor).
+ */
+void nvte_set_grouped_tensor_param(NVTEGroupedTensor *tensor, NVTEGroupedTensorParam param_name,
+                                   const NVTEBasicTensor *param);
```
timmoon10, line 455: This works when we're setting basic tensors, but doesn't generalize to other types (`bool`/`float`/etc). Consider using a more general API like how we handle `NVTEQuantizationConfig`:
https://github.com/NVIDIA/TransformerEngine/blob/f8cb598c1c9b6228d39bf77cec3ea27c7ab6a1d7/transformer_engine/common/include/transformer_engine/transformer_engine.h#L369-L371
This is completely general, but also more cumbersome.

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -393,6 +393,106 @@ int nvte_is_non_tn_fp8_gemm_supported();
...
+ * own memory.
+ *
+ *  \param[in] scaling_mode    Scaling mode of the grouped tensor.
+ *  \param[in] num_tensors     Number of tensors in the group (must be > 0).
+ *  \param[in] logical_shape   Logical 2D shape of the grouped data.
+ *
+ *  \return A new TE grouped tensor.
+ */
+NVTEGroupedTensor nvte_create_grouped_tensor(NVTEScalingMode scaling_mode, size_t num_tensors,
+                                             NVTEShape logical_shape);
```
timmoon10, line 435: It's not a deal-breaker, but it's not ideal that these arguments have to be specified at construction time and can't be changed afterwards. This is an argument for implementing more general accessors.

transformer_engine/common/common.h
```diff
@@ -101,6 +101,7 @@ struct SimpleTensor {
     }
     return acc;
   }
+  bool has_data() const noexcept { return dptr != nullptr; }
```
timmoon10: When doing a group GEMM with a split of size zero, we can encounter an edge case where a tensor has a non-null dptr but dptr is not safe to access. See https://github.com/NVIDIA/TransformerEngine/pull/2330 for an attempt to make the behavior of `has_data` consistent. 
```suggestion
  bool has_data() const noexcept { return dptr != nullptr && numel() > 0; }
```

transformer_engine/common/common.h
```diff
@@ -281,6 +282,132 @@ struct Tensor {
...
+ public:
+  /*
+  Grouped tensor is a collection of tensors with different shapes but the same dtype and scaling mode
+
+  Shape Representation:
+  - logical_shape: 2D shape representing the conceptual layout (REQUIRED)
+    + When all_same_shape(): [num_tensors * M, N] where each tensor is (M, N)
+    + When varying_first_dim(): [~sum_of_first_dims, N] where N is common
+    + When varying_last_dim(): [M, ~sum_of_last_dims] where M is common
+    + When varying_both_dims(): [1, total_elements] (fully flattened)
```
timmoon10: I don't think "logical shape" is quite the right name for this concept. When I hear "logical shape", I think of the mathematical tensor that we are computing, which is completely independent of the physical data layout. An activation tensor might have logical shape `[batch_size, seqlen, num_heads, hidden_dim]`, even if the wgrad GEMM is actually using an FP8 transpose with shape `[hidden_dim, batch_size*seqlen*num_heads]`.

Really, what we're saying is that we have a full tensor with shape `[m, n]`, and we are slicing this tensor. So something like `full_tensor_logical_shape` might be the most accurate name, although it's too cumbersome. How about `full_tensor_shape`?

transformer_engine/common/common.h
```diff
@@ -281,6 +282,132 @@ struct Tensor {
...
+  bool has_columnwise_data() const noexcept { return columnwise_data.dptr != nullptr; }
+
+  bool all_same_first_dim() const noexcept { return first_dims.shape.empty(); }
+  bool all_same_last_dim() const noexcept { return last_dims.shape.empty(); }
+  bool all_same_shape() const noexcept {
+    return first_dims.shape.empty() && last_dims.shape.empty();
+  }
+  bool varying_both_dims() const noexcept {
+    return !first_dims.shape.empty() && !last_dims.shape.empty();
+  }
```
timmoon10: https://github.com/NVIDIA/TransformerEngine/pull/2330 is going to change the default tensor shape to be `[0]`, not `[]`. This is because 0-D tensors are perfectly valid and have one entry (for the same reason 2^0=1 and not 2^0=0). All standard tensor libraries follow this convention:
```
> import numpy as np
> np.zeros([]).size
1
> np.zeros([0]).size
0
```

transformer_engine/common/include/transformer_engine/transformer_engine.h
phu0ngng: In which case, do we need to change those attributes afterwards? 
I actually intended to make those fields initialized at construction time and unchange afterward.

transformer_engine/common/common.h
phu0ngng: I think the name `full_tensor_shape` is quite confusing. As we say the grouped tensor has `num_tensors` tensors, the `full_tensor` could also indicate the tensor in the grouped tensor. 

Besides, when we have the varying last dim or varying both dims, we are not slicing from a 2D tensor as the actual grouped tensor buffer would be a 1D.  In these cases, the 2D shape is no longer the full grouped tensor shape.

`grouped_tensor_logical_shape` could be a clear name. At the same time, the current `logical_shape` is a field of NVTEGroupedTensor which already serves the purpose.

transformer_engine/common/common.h
phu0ngng: Fixed. These queries now use `has_data()` instead of empty checks.

transformer_engine/common/common.h
timmoon10: Ok, maybe we can just make the comments a little clearer:
```suggestion
  - logical_shape: 2D shape representing the conceptual layout, i.e. the shape when the grouped tensors are flattened to 2D and concatenated together (REQUIRED)
    + When all_same_shape(): [num_tensors * M, N] where each tensor is (M, N)
    + When varying_first_dim(): [~sum_of_first_dims, N] where N is common
    + When varying_last_dim(): [M, ~sum_of_last_dims] where M is common
    + When varying_both_dims(): [1, total_elements] (fully flattened)
```

transformer_engine/common/common.h
```diff
@@ -154,12 +155,10 @@ struct Tensor {
...
   }
 
-  bool has_data() const noexcept { return data.dptr != nullptr; }
+  bool has_data() const noexcept { return data.has_data(); }
 
   // Check for size (not just pointer) for 0-dim or no token cases.
-  bool has_columnwise_data() const noexcept {
-    return columnwise_data.dptr != nullptr || columnwise_data.shape.size() != 0;
-  }
+  bool has_columnwise_data() const noexcept { return columnwise_data.has_data(); }
```
timmoon10: This is a nice approach. If this causes issues or test failures, I propose we merge https://github.com/NVIDIA/TransformerEngine/pull/2330 first rather than trying to fix them in this PR.

transformer_engine/common/include/transformer_engine/transformer_engine.h
timmoon10: I'm trying to future-proof. My worry is that we'll need to implement a general accessor in the future, and then `nvte_set_grouped_tensor_param`/`nvte_get_grouped_tensor_param` will be redundant. We've already gone through this before with `NVTEQuantizationConfig` and `NVTEMatmulConfig`.

I'm less worried about `nvte_create_grouped_tensor` since we could add an overloaded definition if we don't need these fields at construction time.

transformer_engine/common/common.h
phu0ngng: Done, but replace `concatenated` with `stacked`.

transformer_engine/common/include/transformer_engine/transformer_engine.h
phu0ngng: Hi, I don't think we can go with a similar API, i.e., using just `void* buf` and `size_in_bytes` as we do need different dtype for different fields.

timmoon10: Yet.

transformer_engine/common/transformer_engine.cpp
```diff
@@ -387,6 +509,90 @@ Tensor *convertNVTETensorCheck(const NVTETensor t) {
...
+  ~GroupedTensorAllocator() {}
+
+  NVTEGroupedTensor Allocate(NVTEScalingMode mode, size_t num_tensors, NVTEShape logical_shape) {
+    std::lock_guard<std::mutex> lock(mutex);
+    if (!free_list.empty()) {
+      uintptr_t index = free_list.back();
+      NVTEGroupedTensor ret = reinterpret_cast<NVTEGroupedTensor>(index);
+      free_list.pop_back();
+      // 1-based indexing - fully reinitialize the tensor to avoid stale data
+      memory[index - 1].clear();
```
ptrendx: Why do we clear it again?

timmoon10 (approved):
LGTM

transformer_engine/common/transformer_engine.cpp
phu0ngng: I see. It is already clear in the `Free()`. Updated

ptrendx (approved):

PR #2400, opened 2025-11-19
Minor improvements to CPU overhead

ksivaman:
# Description

Follow-up to #2377 to incorporate 2 more small changes to improve cpu overhead.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring
- [x] Performance

## Changes

- Cache `nvte_is_non_tn_fp8_gemm_supported` (+ @zhongbozhu)
- Address @vthumbe1503's review comment from #2377.
- Change `get_tensor_device` to reduce cpu overheads.

# Checklist:

### DIFF
--- a/tests/pytorch/distributed/test_comm_gemm_overlap.py
+++ b/tests/pytorch/distributed/test_comm_gemm_overlap.py
@@ -120,12 +120,14 @@ def _run_layer_with_overlap(
     os.environ["PYTORCH_JIT"] = "0"
     os.environ["NVTE_TORCH_COMPILE"] = "0"
     os.environ["NVTE_ALLOW_NONDETERMINISTIC_ALGO"] = "0"
+    os.environ["NVTE_FLASH_ATTN"] = "0"
 
     result = subprocess.run(test_cmd, env=os.environ, capture_output=True, check=False)
 
     os.unsetenv("PYTORCH_JIT")
     os.unsetenv("NVTE_TORCH_COMPILE")
     os.unsetenv("NVTE_ALLOW_NONDETERMINISTIC_ALGO")
+    os.unsetenv("NVTE_FLASH_ATTN")
 
     if (
         result.returncode != 0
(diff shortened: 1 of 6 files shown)


yaox12 (approved):

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -69,29 +68,25 @@ def validate_gemm_scale(scale: Optional[float], required: bool) -> float:
...
-            if tensor._data is not None
-            else tensor._transpose.device.index
-        )
-    try:
-        return (
-            tensor._data.device.index if tensor._data is not None else tensor._data_t.device.index
-        )
-    except AttributeError:
-        return torch.cuda.current_device()
+    if hasattr(tensor, "_rowwise_data") and tensor._rowwise_data is not None:
```
vthumbe1503, line 81: I am wondering where doing something like this would be better?

```
if isinstance(tensor, QuantizedTensorStorage):
  inner_tensors = tensor.get_data_tensors()
  for inner_tensor in inner_tensors:
    if inner_tensor is not None:
      return inner_tensor.device.index
elif hasattr(tensor, "device"):
  return tensor.device.index
else:
  return torch.cuda.current_device()
```
I remember hasattr taking quite a bit of time for some of the FSDP classes when I was working with FSDP2. But I guess it depends on the number of attributes present in a class which would be lesser for our classes, but still using get_data_tensors might be cleaner and potentially faster?

Other idea which might be even more cleaner, would be to implement device as a property member of the Quantized tensor class and this logic can go there. Hopefully torch wont add any function call overhead of its own when implementing a property like that.

vthumbe1503 (commented):
Left a minor comment. LGTM otherwise.

transformer_engine/pytorch/cpp_extensions/gemm.py
ksivaman: I tried these, and several other approaches, all of which were worse. `hasattr` is not the best option but it is better than using loops or other functions. The device method/property would have to be added to the storage classes, which also had larger overhead.

PR #2411, opened 2025-11-21
[PyTorch][NVFP4][MOE] NVFP4 Grouped Quantize with Hadamard Transform

zhongbozhu:
# Description

- This PR depends on the merge of this PR: https://github.com/NVIDIA/TransformerEngine/pull/2351 which is a grouped kernel for generating NVFP4 global amax values. 
- This PR is trying to build grouped-quantize kernels. There are two grouped quantize kernels needed for NVFP4: the regular quantize without Hadamard Transform, as well as a Hadamard Transform FP4 Cast fusion. This PR will prioritize the one with Hadamard and maybe support the other one (or in another future PR)

Status:

- [x] Grouped Rowwise-only quantize without Hadamard (@Oleg-Goncharov for adding transpose mode to it in case RHT get disabled)
- [x] Grouped Columnwise RHT-cast fusion: numerically correct
- [x] Grouped Columnwise RHT-cast fusion: optimization of the kernel for better memory SOL (current memory SOL is low). It has been root caused to be related with running high precision math in the epilogue of the RHT gemm. Running fast math will boost the perf but comes with some minor numerical impact due to the difference between x / y != x * (1/y) even for FP32, but divides are slower. Use ~NVTE_RHT_CAST_FUSION_USE_FAST_MATH~ `NVTE_USE_FAST_MATH` env var to control this behavior. 
- [x] Further fuse Rowwise quantize & Colwise RHT transform & Colwise cast transpose into one kernel. This new kernel targeting SOL perf will further assume 128 padding in the token dimension. Current padding to NVFP4 is 64 multiple, but we might have to further increase it to 128 and then try other methods to remove the padding overhead to achieve overall SOL performance. We can eliminate padding overhead using the following tricks in the next section. Megatron [PR ](https://github.com/NVIDIA/Megatron-LM/pull/2654)to also bump up this padding. 
- [x] Make Row-quant and Col-quant use two different random numbers like this PR [2487 ](https://github.com/NVIDIA/TransformerEngine/pull/2487)already did in unfused path, @negvet to review. 

Notes:

- Why is zero padding bad?
Zero padding means full tensor memory read and write, and often stems from the scaling factor layout of quantized recipes like mxfp8. 

- How to remove the padding overhead? 
1. Router padding: let the router outputs x multiple of tokens directly so that zero padding becomes an no-op. This method has been integrated to Megatron-core [here](https://github.com/NVIDIA/Megatron-LM/blob/93da800461fd4f6a8262c3bab227868b966aaa69/megatron/training/arguments.py#L3276). 
3. Router padding to 128 multiple sounds scary and will raise concerns about if this will incur numerical issues. A better solution is to fuse the zero padding into the token permute kernel. This is still an on-going work, but Megatron-core with the [HybridEP](https://github.com/deepseek-ai/DeepEP/tree/hybrid-ep) token dispatcher backend already supports this feature to pad to a multiple [here](https://github.com/NVIDIA/Megatron-LM/blob/93da800461fd4f6a8262c3bab227868b966aaa69/megatron/core/transformer/moe/token_dispatcher.py#L1012). 

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/common/hadamard_transform/group_hadamard_transform_cast_fusion.cu
+++ b/transformer_engine/common/hadamard_transform/group_hadamard_transform_cast_fusion.cu
@@ -36,18 +36,19 @@ namespace detail {
 namespace {
 
 using namespace cute;
-using cute::Tensor;  // Ensure unqualified Tensor refers to cute::Tensor, not transformer_engine::Tensor
+using cute::
+    Tensor;  // Ensure unqualified Tensor refers to cute::Tensor, not transformer_engine::Tensor
 
 using Stride2D = cute::Stride<int, cute::Int<1>>;
 
 constexpr int kMaxTensorsPerKernel = 64;  // Args must be <4 KB, expand 64 if needed
 struct MultiAmaxHadamardCastFusionArgs {
   // (output) Amax buffer for pre-RHT amax buffer
-  void* global_amax_list[kMaxTensorsPerKernel];
+  void *global_amax_list[kMaxTensorsPerKernel];
   // output C pointers for each tensor
-  void* output_colwise_list[kMaxTensorsPerKernel];
+  void *output_colwise_list[kMaxTensorsPerKernel];
   // output scale inverse pointers for each tensor
-  void* output_colwise_scale_inv_list[kMaxTensorsPerKernel];
+  void *output_colwise_scale_inv_list[kMaxTensorsPerKernel];
   // split sections of each tensor of input
   int split_sections[kMaxTensorsPerKernel];
   // Prefix sum (with leading zero) of split_sections of each tensor of input
@@ -58,16 +59,17 @@ struct MultiAmaxHadamardCastFusionArgs {
   int num_tensors;
 };
 
-
-__device__ __forceinline__ float* GetGlobalAmaxPtrByTensorId(MultiAmaxHadamardCastFusionArgs* kernel_args_ptr, int tensor_id){
+__device__ __forceinline__ float *GetGlobalAmaxPtrByTensorId(
+    MultiAmaxHadamardCastFusionArgs *kernel_args_ptr, int tensor_id) {
   // directly returns the global amax pointer by tensor id
   if (tensor_id < 0 || tensor_id >= kernel_args_ptr->num_tensors) {
     return nullptr;
   }
-  return reinterpret_cast<float*>(kernel_args_ptr->global_amax_list[tensor_id]);
+  return reinterpret_cast<float *>(kernel_args_ptr->global_amax_list[tensor_id]);
 }
 
-__device__ __forceinline__ int GetTensorId(MultiAmaxHadamardCastFusionArgs* kernel_args_ptr, int offset){
+__device__ __forceinline__ int GetTensorId(MultiAmaxHadamardCastFusionArgs *kernel_args_ptr,
+                                           int offset) {
   // Check the kernel args and get the corresponding id
   const int num_tensors = kernel_args_ptr->num_tensors;
   if (offset >= kernel_args_ptr->split_sections_range[num_tensors]) {
@@ -86,27 +88,23 @@ __device__ __forceinline__ float ComputeGlobalEncodeScaleFP4(const float global_
   constexpr float kFP4E2M1Max = 6.0f;
   // If scale is infinity, return max value of float32
   float global_encode_scale = cutlass::minimum_with_nan_propagation<float>{}(
-    kFP8E4M3Max * kFP4E2M1Max / global_amax, cutlass::platform::numeric_limits<float>::max());
+      kFP8E4M3Max * kFP4E2M1Max / global_amax, cutlass::platform::numeric_limits<float>::max());
   // If global amax is 0 or infinity, return 1
   return (global_amax == 0.f || global_encode_scale == 0.f) ? 1.f : global_encode_scale;
 }
 
-template <class ElementA,
-          class ElementB,
-          class ASmemLayout,
-          class BSmemLayout>
+template <class ElementA, class ElementB, class ASmemLayout, class BSmemLayout>
 struct SharedStorage {
   static constexpr int AccumulatorPipelineStageCount = 16;
   using AtomThrShapeMNK = cute::Shape<_1, _1, _1>;
 
-  using AccumulatorPipeline = cutlass::PipelineUmmaAsync<AccumulatorPipelineStageCount / 4, AtomThrShapeMNK>;
+  using AccumulatorPipeline =
+      cutlass::PipelineUmmaAsync<AccumulatorPipelineStageCount / 4, AtomThrShapeMNK>;
   using AccumulatorPipelineStorage = typename AccumulatorPipeline::SharedStorage;
 
   static constexpr int MainloopPipelineStageCount = size<3>(ASmemLayout{});
-  using MainloopPipeline = cutlass::PipelineTmaUmmaAsync<
-                             MainloopPipelineStageCount,
-                             Shape<_1,_1,_1>,
-                             AtomThrShapeMNK>;
+  using MainloopPipeline =
+      cutlass::PipelineTmaUmmaAsync<MainloopPipelineStageCount, Shape<_1, _1, _1>, AtomThrShapeMNK>;
   using MainloopPipelineStorage = typename MainloopPipeline::SharedStorage;
 
   alignas(16) AccumulatorPipelineStorage accumulator;
@@ -119,42 +117,43 @@ struct SharedStorage {
     cute::array_aligned<ElementA, cute::cosize_v<ASmemLayout>> smem_A;
     cute::array_aligned<ElementB, cute::cosize_v<BSmemLayout>> smem_B;
   } tensors;
-
 };
 
 CUTLASS_DEVICE
-cutlass::Array<cutlass::float_e2m1_t, 8>
-StochasticNumericConverterBase(cutlass::Array<float, 8> const &input, cutlass::Array<uint32_t, 2> const &rbits) {
+cutlass::Array<cutlass::float_e2m1_t, 8> StochasticNumericConverterBase(
+    cutlass::Array<float, 8> const &input, cutlass::Array<uint32_t, 2> const &rbits) {
   using result_type = cutlass::Array<cutlass::float_e2m1_t, 8>;
   result_type output;
   constexpr bool has_rs = ARCH_HAS_STOCHASTIC_ROUNDING;
   if constexpr (has_rs) {
     auto output_ptr = reinterpret_cast<uint16_t *>(&output);
-    asm volatile( \
-        "{\n" \
-        "cvt.rs.satfinite.e2m1x4.f32   %0, {%5, %4, %3, %2}, %10;\n" \
-        "cvt.rs.satfinite.e2m1x4.f32   %1, {%9, %8, %7, %6}, %11;\n" \
-        "}" \
-        : "=h"(output_ptr[0]),
-        "=h"(output_ptr[1])
-        : "f"(input[0]), "f"(input[1]), "f"(input[2]), "f"(input[3]),
-        "f"(input[4]), "f"(input[5]), "f"(input[6]), "f"(input[7]),
-        "r"(rbits[0]), "r"(rbits[1]));
+    asm volatile(
+        "{\n"
+        "cvt.rs.satfinite.e2m1x4.f32   %0, {%5, %4, %3, %2}, %10;\n"
+        "cvt.rs.satfinite.e2m1x4.f32   %1, {%9, %8, %7, %6}, %11;\n"
+        "}"
+        : "=h"(output_ptr[0]), "=h"(output_ptr[1])
+        : "f"(input[0]), "f"(input[1]), "f"(input[2]), "f"(input[3]), "f"(input[4]), "f"(input[5]),
+          "f"(input[6]), "f"(input[7]), "r"(rbits[0]), "r"(rbits[1]));
   } else {
-    NVTE_DEVICE_ERROR("FP4 cvt PTX instructions are architecture-specific. "
+    NVTE_DEVICE_ERROR(
+        "FP4 cvt PTX instructions are architecture-specific. "
         "Try recompiling with sm_XXXa instead of sm_XXX.");
   }
   return output;
 }
 
 CUTLASS_DEVICE
-cutlass::Array<cutlass::float_e2m1_t, 16>
-StochasticNumericConverter(cutlass::Array<float, 16> const &input, cutlass::Array<uint32_t, 4> const *rbits) {
+cutlass::Array<cutlass::float_e2m1_t, 16> StochasticNumericConverter(
+    cutlass::Array<float, 16> const &input, cutlass::Array<uint32_t, 4> const *rbits) {
   using result_type = cutlass::Array<cutlass::float_e2m1_t, 16>;
   result_type output;
-  cutlass::Array<cutlass::float_e2m1_t, 8> *result_ptr = reinterpret_cast<cutlass::Array<cutlass::float_e2m1_t, 8> *>(&output);
-  cutlass::Array<float, 8> const *source_ptr = reinterpret_cast<cutlass::Array<float, 8> const *>(&input);
-  cutlass::Array<uint32_t, 2> const *rbits_ptr = reinterpret_cast<cutlass::Array<uint32_t, 2> const *>(rbits);
+  cutlass::Array<cutlass::float_e2m1_t, 8> *result_ptr =
+      reinterpret_cast<cutlass::Array<cutlass::float_e2m1_t, 8> *>(&output);
+  cutlass::Array<float, 8> const *source_ptr =
+      reinterpret_cast<cutlass::Array<float, 8> const *>(&input);
+  cutlass::Array<uint32_t, 2> const *rbits_ptr =
+      reinterpret_cast<cutlass::Array<uint32_t, 2> const *>(rbits);
   CUTLASS_PRAGMA_UNROLL
   for (int i = 0; i < 2; i++) {
     result_ptr[i] = StochasticNumericConverterBase(source_ptr[i], rbits_ptr[i]);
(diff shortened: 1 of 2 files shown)


transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -100,10 +100,239 @@ py::object dequantize(const py::handle &input, transformer_engine::DType otype)
...
+             "NVFP4 multi_quantize: amax reduction is not supported");
+
+  size_t num_tensors = split_sections.size();
+
+  size_t rows = 1;
+  for (size_t i = 0; i < input.ndim() - 1; ++i) {
+    rows *= input.size(i);
+  }
+  size_t cols = input.size(input.ndim() - 1);
+
```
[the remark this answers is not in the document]
zhongbozhu: rows and cols are swapped in the context of RHT colwise kernel, closing this conversation

transformer_engine/common/hadamard_transform/group_hadamard_transform.cu
```diff
@@ -0,0 +1,603 @@
...
+  // Call the group tensor Hadamard transform amax implementation.
+  group_hadamard_transform_amax(*input_tensor, output_list, split_sections, num_tensors,
+                                static_cast<uint16_t>(random_sign_mask),
+                                static_cast<uint16_t>(random_sign_mask_t), false, stream);
+}
+
+// Grouped-tensor amax without doing hadamard transform
+void nvte_group_tensor_amax(const NVTETensor input, NVTETensor* outputs, const int* split_sections,
+                            const size_t num_tensors, cudaStream_t stream) {
+  NVTE_API_CALL(nvte_group_hadamard_transform_amax);
```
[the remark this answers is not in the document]
zhongbozhu: resolved

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -320,6 +321,70 @@ void quantize_bwd_helper(const NVTETensor grad, const NVTETensor input, NVTETens
...
+  }
+
+  // Check for unsupported options
+  if (quant_config_cpp.stochastic_rounding) {
+    NVTE_CHECK(output_tensors[0]->scaling_mode == NVTE_NVFP4_1D_SCALING,
+               "Stochastic rounding is only supported for NVFP4 quantization.");
+  }
+
+  // Take the scaling mode of the first output tensor
+  auto scaling_mode = output_tensors[0]->scaling_mode;
```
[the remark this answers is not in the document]
zhongbozhu: This is already validated in the torch csrc.

transformer_engine/pytorch/custom_recipes/quantization_nvfp4.py
```diff
@@ -612,6 +639,14 @@ def _quantize(self, tensor: torch.Tensor) -> Tuple[
                 if self.columnwise_usage
                 else global_amax_row
             )
+            # TODO: this is a horrible hack to pass zero tolerance test
+            # currently the amax of RHT transform still has a fp32 -> bf16 -> fp32 round-trip
+            # this is to simulate that behaviour and we will remove this once the amax of RHT transform is fixed
+            if self.columnwise_usage and with_rht_cast_fusion:
+                # global_amax_col = global_amax_col.to(torch.bfloat16).float()
+                global_amax_col = (
+                    torch.max(torch.abs(col_input.bfloat16())).to(torch.float32).view(1)
```
ptrendx: I don't think we want to do that. Instead, we should modify the amax calculation kernel to output the amax from the fp32 rht output and then here just do the max of the abs(col_input) without casting to bf16.

ptrendx: Ok, I see that you basically said the same in the comment. Is there any blocker for that amax change?

zhongbozhu: Blocked by some hardcoded logic in amax reduction assuming 2xBF16 and compare 4bytes at a time: https://github.com/NVIDIA/TransformerEngine/blob/f0572aa5c3acd9043a7c687751cb46c0218b923c/transformer_engine/common/hadamard_transform/hadamard_transform_utils.cuh#L114

transformer_engine/common/hadamard_transform/group_hadamard_transform_cast_fusion.cu
```diff
@@ -661,21 +670,27 @@ group_rht_gemm_device(MShape M, NShape N, KShape K, ClusterTileShape cluster_til
...
-        // (optional) path for faster math, use multiply to repalce div
-        // pvscales = cutlass::multiplies<cutlass::Array<ElementAccumulator, NumVecs>>{}(vec_maxs, global_encode_scale_multiplier);
+        if constexpr (kEnableFastMath) {
+          // path for faster math, use multiply to repalce div
+          pvscales = cutlass::multiplies<cutlass::Array<ElementAccumulator, NumVecs>>{}(vec_maxs, global_encode_scale_multiplier);
+        } else {
+          // regular path for slower math, use divide
+          pvscales = cutlass::divides<cutlass::Array<ElementAccumulator, NumVecs>>{}(vec_maxs, fp4_max);
+          pvscales = cutlass::multiplies<cutlass::Array<ElementAccumulator, NumVecs>>{}(pvscales, global_encode_scale);
+        }
```
timmoon10: Comment nits:
```suggestion
        if constexpr (kEnableFastMath) {
          // path for fast math: multiply inverse instead of divide
          pvscales = cutlass::multiplies<cutlass::Array<ElementAccumulator, NumVecs>>{}(vec_maxs, global_encode_scale_multiplier);
        } else {
          // path for accurate math: divide
          pvscales = cutlass::divides<cutlass::Array<ElementAccumulator, NumVecs>>{}(vec_maxs, fp4_max);
          pvscales = cutlass::multiplies<cutlass::Array<ElementAccumulator, NumVecs>>{}(pvscales, global_encode_scale);
        }
```

transformer_engine/common/hadamard_transform/group_hadamard_transform_cast_fusion.cu
```diff
@@ -1011,13 +1027,21 @@ void group_hadamard_transform_cast_fusion_columnwise(
     k_tile_size = 512;
   }
 
+  // TODO: haven't decided whether to expose this as a API option or not
+  // use fast math if there is a ENV var NVTE_RHT_CAST_FUSION_USE_FAST_MATH, default to false
+  static const bool use_fast_math =
+      transformer_engine::getenv<bool>("NVTE_RHT_CAST_FUSION_USE_FAST_MATH", false);
```
timmoon10: Comparing approaches:
- Check envvar (current): simple to implement, extremely obscure for users
- Add fast math to quantization config: vaguely similar to stochastic rounding (only affect NVFP4 quantization, no effect on dequantization), could be applied to other places we quantize to NVFP4, somewhat annoying to implement

transformer_engine/common/CMakeLists.txt
```diff
@@ -351,6 +353,7 @@ endforeach()
 
 set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} --expt-relaxed-constexpr")
 set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -O3")
+set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -lineinfo")
```
timmoon10: Debugging code:
```suggestion
```

transformer_engine/common/include/transformer_engine/cast.h
```diff
@@ -270,6 +270,20 @@ void nvte_multi_tensor_quantize(const NVTETensor *inputs, NVTETensor *outputs,
...
+ *  \param[in,out]  outputs          Output quantized tensors.
+ *  \param[in]      split_sections   Split sections of the input tensor.
+ *  \param[in]      num_tensors      Number of output tensors.
+ *  \param[in]      quant_config    (Optional) Quantization configurations.
+ *  \param[in]      stream           CUDA stream used for the operation.
+ */
+void nvte_group_nvfp4_quantize_with_amax(const NVTETensor input, NVTETensor *outputs,
+                                         const size_t *split_sections, size_t num_tensors,
+                                         const NVTEQuantizationConfig quant_config,
+                                         cudaStream_t stream);
```
timmoon10, line 285: - This API should be general and not specific to NVFP4.
- We should use the `GroupedTensor` API instead of CPU split points.

transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
```diff
@@ -470,9 +480,9 @@ rht_gemm_device(MShape M, NShape N, KShape K, ClusterTileShape cluster_tile,
         ++accumulator_pipe_consumer_state;
 
         // Cast data from FP32 to BF16 to FP32.
-        auto convert_accum_to_bf16 = cutlass::NumericArrayConverter<cutlass::bfloat16_t, ElementAccumulator, FragmentSize>{};
-        auto convert_bf16_to_accum = cutlass::NumericArrayConverter<ElementAccumulator, cutlass::bfloat16_t, FragmentSize>{};
-        tTR_rAcc_frag(_0{}) = convert_bf16_to_accum(convert_accum_to_bf16(tTR_rAcc_frag(_0{})));
+        // auto convert_accum_to_bf16 = cutlass::NumericArrayConverter<cutlass::bfloat16_t, ElementAccumulator, FragmentSize>{};
+        // auto convert_bf16_to_accum = cutlass::NumericArrayConverter<ElementAccumulator, cutlass::bfloat16_t, FragmentSize>{};
+        // tTR_rAcc_frag(_0{}) = convert_bf16_to_accum(convert_accum_to_bf16(tTR_rAcc_frag(_0{})));
```
timmoon10: Skipping the BF16 downcasting feels nicer, but are we okay with changing the numerics?

Also, we should delete this rather than just commenting it out.

transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
```diff
@@ -481,14 +491,27 @@ rht_gemm_device(MShape M, NShape N, KShape K, ClusterTileShape cluster_tile,
...
-        pvscales = cutlass::divides<cutlass::Array<ElementAccumulator, NumVecs>>{}(vec_maxs, fp4_max);
-        pvscales = cutlass::multiplies<cutlass::Array<ElementAccumulator, NumVecs>>{}(pvscales, global_encode_scale);
+        if constexpr (kEnableFastMath) {
+          // path for faster math, use multiply to repalce div
+          pvscales = cutlass::multiplies<cutlass::Array<ElementAccumulator, NumVecs>>{}(vec_maxs, global_encode_scale_multiplier);
+        } else {
+          // regular path for slower math, use divide
+          pvscales = cutlass::divides<cutlass::Array<ElementAccumulator, NumVecs>>{}(vec_maxs, fp4_max);
+          pvscales = cutlass::multiplies<cutlass::Array<ElementAccumulator, NumVecs>>{}(pvscales, global_encode_scale);
+        }
```
timmoon10: Nits: fast/accurate is the right tradeoff. If it was fast/slow, there would be no reason to keep the slow path.
```suggestion
        if constexpr (kEnableFastMath) {
          // Fast path: multiply by scale/fp4_max
          pvscales = cutlass::multiplies<cutlass::Array<ElementAccumulator, NumVecs>>{}(vec_maxs, global_encode_scale_multiplier);
        } else {
          // Accurate path: divide by fp4_max, multiply by scale
          pvscales = cutlass::divides<cutlass::Array<ElementAccumulator, NumVecs>>{}(vec_maxs, fp4_max);
          pvscales = cutlass::multiplies<cutlass::Array<ElementAccumulator, NumVecs>>{}(pvscales, global_encode_scale);
        }
```

transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
```diff
@@ -481,14 +491,27 @@ rht_gemm_device(MShape M, NShape N, KShape K, ClusterTileShape cluster_tile,
...
         auto qpvscale_scaled = cutlass::multiplies<cutlass::Array<ElementAccumulator, NumVecs>>{}(qpvscale_ups, global_decode_scale);
-        auto acc_scales = cutlass::divides<cutlass::Array<ElementAccumulator, NumVecs>>{}(1.0, qpvscale_scaled);
+        cutlass::Array<ElementAccumulator, NumVecs> acc_scales;
+        if constexpr (kEnableFastMath) {
+          // fast math: use reciprocal approximate to replace div
+          acc_scales = cutlass::reciprocal_approximate_ftz<decltype(qpvscale_scaled)>{}(qpvscale_scaled);
+        } else {
+          // regular path for slower math, use divide to replace div
+          acc_scales = cutlass::divides<cutlass::Array<ElementAccumulator, NumVecs>>{}(1.0, qpvscale_scaled);
+        }
```
timmoon10: Nit:
```suggestion
        if constexpr (kEnableFastMath) {
          // Fast path: compute scale with approximate reciprocal
          acc_scales = cutlass::reciprocal_approximate_ftz<decltype(qpvscale_scaled)>{}(qpvscale_scaled);
        } else {
          // Accurate path: compute scale with division
          acc_scales = cutlass::divides<cutlass::Array<ElementAccumulator, NumVecs>>{}(1.0, qpvscale_scaled);
        }
```

transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
```diff
@@ -800,20 +824,28 @@ void hadamard_transform_cast_fusion_columnwise(const Tensor &input_, Tensor &out
   } else if (m < 1024 || n < 1024) {
     k_tile_size = 512;
   }
+
+  // TODO: haven't decided whether to expose this as a API option or not
+  // use fast math if there is a ENV var NVTE_RHT_CAST_FUSION_USE_FAST_MATH, default to false
+  static const bool use_fast_math =
+      transformer_engine::getenv<bool>("NVTE_RHT_CAST_FUSION_USE_FAST_MATH", false);
```
timmoon10: I think this belongs in `QuantizationConfig` since it's relevant for all NVFP4 quantization.

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -320,6 +321,70 @@ void quantize_bwd_helper(const NVTETensor grad, const NVTETensor input, NVTETens
   }
 }
 
+template <bool IS_ACT, typename ParamOP, float (*OP)(float, const ParamOP &)>
+void group_quantize_fwd_helper(const NVTETensor input, NVTETensor *outputs,
```
timmoon10, line 325: We're copy-pasting the templated infrastructure for single-tensor quantization, and inheriting all of its complexity to handle unnecessary features.

transformer_engine/common/cast/nvfp4/group_quantize_transpose_nvfp4.cuh
```diff
@@ -0,0 +1,902 @@
...
+                              cudaStream_t stream) {
+#if FP4_TYPE_SUPPORTED
+  using namespace group_quantize_transpose_kernel;
+  using namespace ptx;
+  bool use_stochastic_rounding = quant_config ? quant_config->stochastic_rounding : false;
+
+  NVTE_CHECK(num_tensors == output_list.size(),
+             "Number of output tensors should match number of tensors.");
+  NVTE_CHECK(num_tensors <= kMaxTensorsPerKernel,
+             "Number of tensors should be less than or equal to ", kMaxTensorsPerKernel);
```
timmoon10, line 753: We'll want to handle this case in the future.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -709,6 +709,166 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>, bool> bulk_alloc
...
+          out_identity.set_amax(out_identity_amax.data_ptr,
+                                static_cast<DType>(out_identity_amax.dtype),
+                                out_identity_amax.shape);
+        }
+        out_identity_list.emplace_back(std::move(out_identity));
+        nvte_tensor_out_identity_list.push_back(out_identity_list.back().data());
+      }
+      nvte_group_nvfp4_quantize_with_amax(input.data(), nvte_tensor_out_identity_list.data(),
+                                          split_sections.data(), num_tensors, quant_config_list[0],
+                                          stream);
```
timmoon10, line 907: Do we need to compute a high-precision RHT before casting to NVFP4?

zhongbozhu: for rowwise: no, because it's not part of the recipe
for colwise: yes, but the RHT will be recomputed in the RHT-cast fusion

transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
zhongbozhu: right, so far we haven't plumbed through it because we don't know if such thing is worthy of a Megatron flag, plus that numerical folks often seem to be paranoid about fast math

transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
zhongbozhu: I think this change is welcomed since it will improve numerics. The fastmath needs a specific toggle for it and default to off.

transformer_engine/common/CMakeLists.txt
zhongbozhu: oops, didn't know it get slipped into the PR

transformer_engine/common/hadamard_transform/hadamard_transform_utils.cuh
```diff
@@ -0,0 +1,198 @@
...
+      : "=r"(a0)
+      : "r"(a0));
+}
+
+__device__ __forceinline__ void unpack_max_of_packed_bf16(uint32_t& packed_bf16, float& float_dst) {
+  __nv_bfloat162 bf16x2 = *reinterpret_cast<__nv_bfloat162*>(&packed_bf16);
+  float f_a = __bfloat162float(bf16x2.x);
+  float f_b = __bfloat162float(bf16x2.y);
+  asm volatile("max.xorsign.abs.f32 %0, %1, %2;\n\t" : "=f"(float_dst) : "f"(f_a), "f"(f_b));
+  float_dst = fabsf(float_dst);
```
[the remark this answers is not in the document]
timmoon10: The sign from `max.xorsign.abs.f32` is computed via XOR, so we need the `fabsf` to force it to be positive: https://docs.nvidia.com/cuda/parallel-thread-execution/#floating-point-instructions-max

transformer_engine/common/hadamard_transform/group_row_cast_col_hadamard_transform_cast_fusion.cu
```diff
@@ -0,0 +1,1483 @@
...
+          auto output_frgs =
+              reinterpret_cast<cutlass::Array<TQA, VectorSize> *>(raw_pointer_cast(tQArQA.data()));
+          Tensor amax =
+              make_tensor<ElementAccumulator>(prepend(take<1, rank(tQArA)>(tQArA.shape()), _1{}));
+          Tensor pvscales = make_tensor_like<ElementAccumulator>(amax);
+          transformer_engine::curanddx::detail::philox4x32_native_state<10> rng;
+          if constexpr (kEnableStochasticRounding) {
+            const size_t rng_sequence = global_thread_idx + k_tile * 512 +
+                                        scheduler.get_linear_tile_idx() * K_TILE_MAX * 512;
+            rng.init(rng_seed, rng_sequence, rng_offset);
```
timmoon10, line 1054: Have we confirmed that the row-wise and col-wise quantization are actually using different random numbers? If I am reading correctly, this is generating random numbers for the row-wise quantization. However, the RNG initialization looks the same as for col-wise quantization: https://github.com/NVIDIA/TransformerEngine/blob/c14b1563070bd1818f9040fceab34e9fd1d54325/transformer_engine/common/hadamard_transform/group_row_cast_col_hadamard_transform_cast_fusion.cu#L880-L882

Do we need to increase the offset so that the row-wise and col-wise RNGs are seeded differently?
```suggestion
            const size_t rng_sequence = global_thread_idx + k_tile * 512 +
                                        scheduler.get_linear_tile_idx() * K_TILE_MAX * 512 +
                                        tiles_in_m * tiles_in_n * K_TILE_MAX * 512;
            rng.init(rng_seed, rng_sequence, rng_offset);
```

Relevant Greptile comment: https://github.com/NVIDIA/TransformerEngine/pull/2411#pullrequestreview-3585406800

rich-junwang:
wow, this is nice. any plan to have grouped quantization for mxfp8? thanks!

zhongbozhu:
> wow, this is nice. any plan to have grouped quantization for mxfp8? thanks!

@rich-junwang There is actually no new kernel for mxfp8 group quant, because the regular mxfp8 quant kernel can just be used to quantize the input as a whole thanks to the two following properties:

- TN, NT, NN layout support means that there is no need for transpose in mxfp8 (which is not the case for NVFP4)
- There is no global per-tensor amax for MXFP8

Therefore, as long as we make sure each expert's input number of tokens is a 128 multiple, MXFP8 quant for dense is automatically a group quant by itself. Removing padding overhead is another story: like router padding or fuse zero padding to permute/unpermute [here](https://github.com/NVIDIA/Megatron-LM/pull/2763). 

There is already a prototype for MXFP8 [here](https://github.com/NVIDIA/TransformerEngine/pull/2360), but merging it is kind of a headache for now :)

PR #2412, opened 2025-11-21
[Common] Tuned NVFP4 cast kernel

Oleg-Goncharov:
# Description

This PR introduces a specialized CUDA kernel optimized for NVFP4 quantization of BF16 inputs on Blackwell architecture (sm100f family). The implementation achieves performance improvements by leveraging architecture-specific features:

### Rowwise + Colwise (transpose)
NVFP4 kernel performance 3

#### a) round-to-nearest
NVFP4 cast rowwise

#### b) stochastic rounding
NVFP4 cast rowwise + SR

Below are the performance measurements for quantizing tensors using dimensions representative of DSv3 **[8192×8, 7168]** on internal Cluster (B300).

NVFP4 cast DSv3

### Threads to data mapping (colwise case)
To reduce shared memory bank conflicts, the following mapping is use when reading from and writing to shmem buffers:
- Singe thread processes 16x2 elements (2x NVFP4 blocks).
- Cells of the same color belong to the same warp
- Indices of threads and their offsets are computed as:
```
const int tid_Y_colwise = (thread_lane % 4 + warp) % 4;
const int tid_X_colwise = thread_lane;

const int thread_offset_Y_colwise = tid_Y_colwise * SCALE_DIM;
const int thread_offset_X_colwise = tid_X_colwise * 2;
```
where `SCALE_DIM=16`.
The arrows in the figure below illustrate how thread indices increment, forming a zigzag pattern.

#### a) Reads from SHMEM Input Buffer
Colwise reads NVFP4

#### b) Writes to SHMEM Output Transpose Buffer
Colwise writes NVFP4

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Added a specialized kernel
- Added the logic to use it when the conditions are met

# Checklist:


transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -100,7 +100,8 @@ void quantize_fwd_helper(const NVTETensor input, NVTETensor output,
       int32_t cols = input_tensor->flat_last_dim();
       auto dtype = input_tensor->dtype();
       bool use_optimized_kernel = (dtype == DType::kBFloat16) && (rows % 32 == 0) &&
-                                  (cols % 32 == 0) && output_tensor->has_data();
+                                  (cols % 32 == 0) && output_tensor->has_data() &&
+                                  is_supported_by_CC_100();
```
ptrendx: Why do we need that additional check? NVFP4 is only supported on SM100+ anyway.

transformer_engine/common/util/ptx.cuh
```diff
@@ -867,19 +1078,19 @@ __device__ __forceinline__ void fma_f32_bf16(float &out, uint16_t const &a, uint
...
-  asm volatile(
-      "{\n\t"
-      ".reg.b32 val;\n"
-      "abs.f32 val, %1;\n"
-      "redux.sync.max.u32 %0, val, 0xFFFFFFFF;\n"
-      "}\n\t"
-      : "=r"(reinterpret_cast<uint32_t &>(out))
-      : "f"(in));
-#endif
+  constexpr bool has_redux_f32 = ARCH_HAS_REDUX_F32;
```
ptrendx: I'm changing this in #2482.

transformer_engine/common/util/ptx.cuh
```diff
@@ -1505,6 +1716,57 @@ __device__ __forceinline__ floatx4 up_cast(const bf16x4 &in) {
...
+  const uint32_t src_smem_ptr = __cvta_generic_to_shared(src_smem);
+  bf16x2 dst;
+  asm volatile("ld.shared.b32 %0, [%1];"
+               : "=r"(reinterpret_cast<uint32_t &>(dst))
+               : "r"(src_smem_ptr));
+  return dst;
+}
+
+// Loads 8x BF16 values from shared memory state space
+__device__ __forceinline__ void ld_shared_b128(uint64_t &elts03, uint64_t &elts47,
```
ptrendx: Can we have a consistent API between those functions? Some return value, some output things in the arguments.

transformer_engine/common/cast/nvfp4/specialized/quantize_transpose_nvfp4_tuned_1D.cuh
```diff
@@ -0,0 +1,765 @@
...
+  const bool return_transpose = output->has_columnwise_data();
+
+  checkCuDriverContext(stream);
+  CheckNoopTensor(*noop, "cast_noop");
+  CheckInputTensor(input, "input");
+  CheckOutputTensor(*output, "output", false);
+
+  NVTE_CHECK(input.has_data(), "Cannot quantize tensor without rowwise data.");
+  NVTE_CHECK(output->has_data(), "NVFP4 output tensor must be allocated.");
+  NVTE_CHECK(is_fp4_dtype(output->data.dtype), "Output must have FP4 type.");
```
tdophung, line 679: I see in `quantize_transpose_nvfp4.cuh:1188`, there is a check for  `NVTE_CHECK(output->scale_inv.dptr != nullptr)`, not sure if this is  also needed here

transformer_engine/common/cast/nvfp4/specialized/quantize_transpose_nvfp4_tuned_1D.cuh
```diff
@@ -0,0 +1,765 @@
...
+#endif  // (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
+}
+
+#endif  // FP4_TYPE_SUPPORTED
+}  // namespace quantize_transpose_tuned_kernel
+
+inline void quantize_transpose_tuned_1D(const Tensor &input, const Tensor *noop, Tensor *output,
+                                        const QuantizationConfig *quant_config,
+                                        cudaStream_t stream) {
+#if FP4_TYPE_SUPPORTED
```
tdophung, line 661: This whole first section of `quantize_transpose_tuned_1D()` looks like boiler plate code from `quantize_transpose()` in `quantize_transpose_nvfp4.cuh`, aside from a few things, like:
- blocks_X, blocks_Y being calculated from tunable configs
- the missing output.scale_inv check
- the involvement of 2D quantize (not needed in this case)

 Is there a way we can merge the logic between this part, and `quantize_transpose` in quantize_transpose_nvfp4.cuh so that it looks cleaner? I think we could also make the CHUNK_DIM_X, CHUNK_DIM_Y in `quantize_transpose_nvfp4.cuh` be TunableConfig, and `quantize _transpose_nvfp4_tuned_1D.cuh`  can import it

transformer_engine/common/util/ptx.cuh
```diff
@@ -153,6 +153,24 @@ __device__ __forceinline__ void mbarrier_arrive(uint64_t *mbar) {
...
+__device__ __forceinline__ void mbarrier_arrive_relaxed_cta_shared_cta(uint64_t *mbar) {
+#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
+  uint32_t mbar_ptr = __cvta_generic_to_shared(mbar);
+  asm volatile("mbarrier.arrive.relaxed.cta.shared::cta.b64 _, [%0], 1;" ::"r"(mbar_ptr));
+#else
+  NVTE_DEVICE_ERROR("mbarrier_arrive_relaxed_cta_shared_cta is only supported on SM 10.0+.");
+#endif  // #if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
+}
+
+__device__ __forceinline__ void mbarrier_arrive_release_cta_shared_cta(uint64_t *mbar) {
```
tdophung: for using the .release, from what I understand, it means we are telling the hardware to ensure visibility into the data from previous writes that also arrive on this mbarrier (as opposed to .relaxed where we do not guarantee this).

So why do we ommit the "memory" clobber for this .released instruction? Wouldn't we want the compiler to not order the mem write instructions before and after this barrier when we are using .released ? I did not see any use of this in `quantize_transpose_nvfp4_tuned_1D.cuh` or anywhere else so I was not sure of the intention.

Furthermore, if adding the "memory" clobber for the .release instruction, wouldn't it just become the baseline one in line 147 `mbarrier_arrive`

transformer_engine/common/util/ptx.cuh
```diff
@@ -243,6 +273,97 @@ __device__ __forceinline__ void mbarrier_wait_parity(uint64_t *mbar, const uint3
...
+      "bra WAIT; \n\t"                // just a time-out, try again
+      "DONE: \n\t"
+      "}\n\t"
+      :
+      : "r"(mbar_ptr), "r"(phase_parity)
+      : "memory");
+#else
+  NVTE_DEVICE_ERROR("mbarrier_wait_parity_relaxed_cta_shared_cta is only supported on SM 10.0+.");
+#endif  // #if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
+}
```
tdophung: why are we defining these when not using (yet)?

transformer_engine/common/cast/nvfp4/specialized/quantize_transpose_nvfp4_tuned_1D.cuh
```diff
@@ -0,0 +1,765 @@
...
+
+static_assert((TILE_DIM_Y % SCALE_DIM == 0) &&
+              "Tile size Y must be evenly divisible by the scale dim\0");
+static_assert((TILE_DIM_X % SCALE_DIM == 0) &&
+              "Tile size X must be evenly divisible by the scale dim\0");
+
+constexpr int TILES_Y = TunableConfig::CHUNK_DIM_Y / TILE_DIM_Y;
+constexpr int TILES_X = TunableConfig::CHUNK_DIM_X / TILE_DIM_X;
+
+constexpr int THREADS_PER_SCALE_ROWWISE = SCALE_DIM / ELTS_PER_THREAD;
```
tdophung, line 68: there is already assert on line 50 for ELTS_PER_THREAD == SCALE_DIM. should this just be 1?

tdophung (commented):
I will leave more comments tomorrow

tests/cpp/operator/test_cast_nvfp4_transpose.cu
```diff
@@ -114,14 +114,16 @@ void quantize_nvfp4_1d(float (*OP)(const float),
...
             // Compute "correct" per-block encoding scaling factor
-            const float S_enc_b_fp8 = S_dec_b_fp8 == 0 ? 0.f : S_enc / S_dec_b_fp8;
+            const float S_enc_b_fp8 = S_dec_b_fp32 == 0.f ? 0.f : S_enc / S_dec_b_fp32;
 
             const size_t scale_idx = i * scales_stride + block_X;
-            scales[scale_idx] = static_cast<fp8e4m3>(S_dec_b_fp8);
-            const float scale_reciprocal = S_enc_b_fp8;
+            scales[scale_idx] = S_dec_b_fp8;
+            // Numerical truncation to match GPU implementation, which uses mixed precision FMA instruction
+            const float scale_reciprocal = static_cast<float>(static_cast<bf16>(S_enc_b_fp8));
```
ptrendx: Let's have that behavior as an option behind the fast math flag.

ptrendx: In this test let's test both approaches, but in the pyTorch exact test let's test only the non-fast math one.

transformer_engine/common/cast/nvfp4/specialized/quantize_transpose_nvfp4_tuned_1D.cuh
ptrendx: I think that does not hurt to have this properly spelled out here (and I think it is actually the right choice here) - it would be useful if somebody changes this kernel and removes the assertion for those values to be equal, but doesn't notice that this value also depends on ELTS_PER_THREAD.

transformer_engine/common/cast/nvfp4/specialized/quantize_transpose_nvfp4_tuned_1D.cuh
Oleg-Goncharov: Right, good catch. This check is actually there, but moved two lines below by mistake during refactoring.

transformer_engine/common/cast/nvfp4/specialized/quantize_transpose_nvfp4_tuned_1D.cuh
Oleg-Goncharov: Although merging it into the generic kernel reduces the code duplication, I'd suggest to have it separate to avoid tight coupling. Because we may need to optimize it further (probably just tuning some parameters) in the future.

transformer_engine/common/util/ptx.cuh
Oleg-Goncharov: I’ll remove those unused functions, they were for testing warp-specialized communication schemes. I needed sub-CTA synchronization (a barrier for only a subset of threads) with weaker ordering than **__syncthreads()**, which is a full-CTA barrier and provides block-scoped memory visibility.

transformer_engine/common/util/ptx.cuh
Oleg-Goncharov: Same as above, I'll remove it too.

transformer_engine/common/cast/nvfp4/quantize_transpose_nvfp4.cuh
```diff
@@ -1159,13 +1160,19 @@ void quantize_transpose(const Tensor &input, const Tensor *noop, Tensor *output,
...
 
   // If transposed output is allocated, return the transposed data. Otherwise, it's not necesary to
   // return the transposed data.
   // TODO(Frank): Is there a better way to do this?
   bool return_transpose = output->has_columnwise_data();
 
+  if (!use_2d_quantization && (input.dtype() == DType::kBFloat16)) {
+    quantize_transpose_tuned_1D(input, noop, output, quant_config, stream);
+    return;
+  }
```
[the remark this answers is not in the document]
Oleg-Goncharov: It is guarded by macros FP4_TYPE_SUPPORTED which triggers the runtime NVTE error

tests/cpp/operator/test_cast_nvfp4_transpose.cu
```diff
@@ -619,8 +630,8 @@ void performTest(float (*OP)(const float),
     }
     ASSERT_EQ(err, cudaSuccess) << cudaGetErrorString(err);
 
-    const double atol = 0.05;
-    const double rtol = 0.1;
+    const double atol = 1.0E-6;
+    const double rtol = 1.0E-6;
```
tdophung, line 634: why is this changing? Looks like it got stricter so I think if it passes, it is better. But I am not understanding why it was lower before

transformer_engine/common/cast/nvfp4/specialized/quantize_transpose_nvfp4_tuned_1D.cuh
```diff
@@ -0,0 +1,789 @@
...
+constexpr int BUFF_OUT_T_SIZE = BUFF_OUT_T_DIM_Y * BUFF_OUT_T_DIM_X;
+
+// Manual swizzling parameters to reduce SHMEM bank conflicts
+constexpr int PACK_SIZE = 8;
+constexpr int WAVES = ELTS_PER_THREAD / PACK_SIZE;
+
+constexpr int THREADS_X_ROWWISE = TILE_DIM_X / ELTS_PER_THREAD;
+constexpr int THREADS_Y_ROWWISE = THREADS_NUM / THREADS_X_ROWWISE;
+
+constexpr int THREADS_X_TRANSP = TILE_DIM_X / 2;
```
tdophung: Can we use 1 convention for macros for transpose? Like you named THREADS_X_T or BUFF_OUT_T_DIM_Y before, so "_T", or "_TRANSP", or "_TR" (like the BUFFS_NUM_OUT_TR one. But just choose 1

transformer_engine/common/cast/nvfp4/specialized/quantize_transpose_nvfp4_tuned_1D.cuh
tdophung: But shouldn't the TunableConfig struct is for that purpose? To make sure that these values are tuned separately from the common patrt of the 2 kernels, and we keep the common part in one place only. Then it should not be tight coupling

Oleg-Goncharov:
`TunableConfig`/coupling: we already aligned on this with Przemek, we’re not moving these values into TunableConfig for this change. Let’s keep the scope as agreed.

transformer_engine/common/cast/nvfp4/specialized/quantize_transpose_nvfp4_tuned_1D.cuh
Oleg-Goncharov: Thanks, I’m fine to standardize. Please consolidate any remaining feedback into a single final pass by EOD today so we can merge without another round of incremental comments. This PR has been open since late November and was originally requested as time-sensitive.

tdophung (approved):
I have no other issue with this PR. We should merge. Thanks for the change!

PR #2437, opened 2025-12-01
[JAX] Triton binding

phu0ngng:
# Description

This PR adds utilities to lower the JAX custom call to the Triton kernel.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/tests/jax/test_triton_custom_calls.py
+++ b/tests/jax/test_triton_custom_calls.py
@@ -81,8 +81,7 @@ class TestTritonBinding:
             # For autotuned kernels, use the minimum BLOCK_SIZE from configs
             # to ensure all elements are processed by all configs
             block_size = min(
-                config.kwargs.get('BLOCK_SIZE')
-                for config in TestTritonBinding.amax_kernel.configs
+                config.kwargs.get("BLOCK_SIZE") for config in TestTritonBinding.amax_kernel.configs
             )
             grid = (triton.cdiv(n_elements, block_size),)
 


tests/jax/test_triton_custom_calls.py
```diff
@@ -0,0 +1,121 @@
...
+        def lowering(ctx, x, *, block_size):
+            """MLIR lowering using Triton kernel."""
+            n_elements = 1
+            for dim in ctx.avals_in[0].shape:
+                n_elements *= dim
+
+            # For autotuned kernels, use the minimum BLOCK_SIZE from configs
+            # to ensure all elements are processed by all configs
+            min_block_size = min(
+                config.kwargs.get("BLOCK_SIZE", block_size)
```
jberchtold-nvidia: Can we remove this `block_size` as an arg to the primitive, since we know for this triton kernel all autotune configs provide a block size?

tests/jax/test_triton_custom_calls.py
```diff
@@ -0,0 +1,121 @@
...
+                for config in TestTritonBinding.amax_kernel.configs
+            )
+            grid = (triton.cdiv(n_elements, min_block_size),)
```
jberchtold-nvidia: Will this make the grid incorrect for autotune configs that are setup in utils.py like this? 
```
config_call = gpu_triton.TritonKernelCall(
                config_kernel,
                grid_tuple[0],
                grid_tuple[1],
                grid_tuple[2],
                config_params,
            )
```
 
because the grid will be based on the smallest block size but we'll have bigger blocks in other configs so would potentially get an out of bounds read/write, right? Or is this handled somewhere in this PR or in JAX's triton_kernel_call that I missed. Thanks!

tests/jax/test_triton_custom_calls.py
phu0ngng: We should.

tests/jax/test_triton_custom_calls.py
phu0ngng: No, it won't hit an out-of-bounds issue.
We launch more thread blocks, but the `mask = offsets < n_elements` will be the guard.

jberchtold-nvidia: Oh I see, thanks!

jberchtold-nvidia (approved):

PR #2440, opened 2025-12-02
[JAX] Better error message when Q, K, V are sharded differently

jberchtold-nvidia:
# Description

When using TE/JAX's DPA and Q, K, or V are sharded differently, the downstream error message only occurs in the inner abstract and reports a shape issue.

For example, when sharded along the batch dim, the previous error message was:
```
AssertionError: Mismatched qkv batch size for q_batch_shape: [1], k_batch_shape: [1] and v_batch_shape: [4]
```
The new error message is:
```
AssertionError: Q, K, and V sharding specs must be identical but received q_spec=PartitionSpec('fsdp', None, None, None), k_spec=PartitionSpec('fsdp', None, None, None), v_spec=PartitionSpec(None, None, None, None)
```

The newly added test `test_mismatching_qkv_sharding_separate_qkv` only takes ~1 second.
```
================================================================================
TEST RUNTIME SUMMARY (grouped by function)
================================================================================
test                                                         |  12x |    2.14s | avg:   0.18s
test_context_parallel_allgather_attn                         | 320x | 1134.35s | avg:   3.54s
test_context_parallel_allgather_attn_shardy                  |  40x |  155.82s | avg:   3.90s
test_context_parallel_ring_attn                              | 1280x | 1753.03s | avg:   1.37s
test_context_parallel_ring_attn_shardy                       |  40x |   50.99s | avg:   1.27s
test_cross_attn                                              |  18x |   30.76s | avg:   1.71s
test_mismatching_qkv_sharding_separate_qkv                   |   1x |    1.16s | avg:   1.16s
test_self_attn                                               |  54x |  123.97s | avg:   2.30s
test_self_attn_shardy                                        |  18x |   16.04s | avg:   0.89s
================================================================================
TOTAL RUNTIME                                                |      | 3268.24s |
================================================================================
```

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add assertion that Q, K, and V shardings match when `qkv_layout.is_separate()`
- Add test to run with mismatched sharding (Q and K are sharded fsdp in batch dim, V is replicated in batch dim) and check that the new error message is reported as expected

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/attention.py
+++ b/transformer_engine/jax/cpp_extensions/attention.py
@@ -687,6 +687,16 @@ class FusedAttnFwdPrimitive(BasePrimitive):
         arg_shardings[-2] = arg_shardings[-4]
         arg_shardings = tuple(arg_shardings)
         out_shardings = (out_sharding, softmax_aux_sharding, rng_state_sharding)
+
+        if config.qkv_layout.is_separate():
+            q_spec = get_padded_spec(arg_infos[0])
+            k_spec = get_padded_spec(arg_infos[1])
+            v_spec = get_padded_spec(arg_infos[2])
+            assert q_spec == k_spec == v_spec, (
+                f"Q, K, and V sharding specs must be identical but received {q_spec=}, {k_spec=},"
+                f" {v_spec=}"
+            )
+
         impl = partial(FusedAttnFwdPrimitive.impl, config=config)
         return mesh, impl, out_shardings, arg_shardings
 
--- a/tests/jax/test_distributed_fused_attn.py
+++ b/tests/jax/test_distributed_fused_attn.py
@@ -708,13 +708,13 @@ class TestMismatchingQKVSharding:
         kv_segment_ids_spec,
     ):
         query = jax.random.uniform(
-            jax.random.PRNGKey(0), (batch_size, num_heads, seq_len, head_dim), dtype=jnp.bfloat16
+            jax.random.PRNGKey(0), (batch_size, seq_len, num_heads, head_dim), dtype=jnp.bfloat16
         )
         key = jax.random.uniform(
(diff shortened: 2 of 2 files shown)


transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -687,6 +687,16 @@ def partition(config, mesh, arg_infos, result_infos):
         arg_shardings[-2] = arg_shardings[-4]
         arg_shardings = tuple(arg_shardings)
         out_shardings = (out_sharding, softmax_aux_sharding, rng_state_sharding)
+
+        if config.qkv_layout.is_separate():
+            q_spec = get_padded_spec(arg_infos[0])
+            k_spec = get_padded_spec(arg_infos[1])
+            v_spec = get_padded_spec(arg_infos[2])
```
jberchtold-nvidia, line 694: @KshitijLakhani does this assertion make sense? For the case I'm testing above, where the batch dimension should always have the same sharding, I think it does make sense. But I'm not yet familiar enough with other parallelism techniques like CP to know if this assertion is valid on the non-batch axes

KshitijLakhani: I can think of [MQA/GQA](https://arxiv.org/pdf/2305.13245) needing different sharding for Q and K along the head dimension .
For e.g. if we have 8 query heads and 1 k/v head then we could have a different PartitionSpec for query (let's say split across 4 devices so that 2 query heads per device) and a different PartitionSpec for key (repeating across devices, so None) so I would not be as restrictive.

I think along the batch might be fine and probably even along seq dim as I do not think any of the CP strategies may require a different PartitionSpec for QKV.  Should be okay along the hidden dimension, too

cc: @mgoldfarb-nvidia @huanghua1994 @mingxu1067 to chime in

jberchtold-nvidia:
Deprioritized for now, unsure on best assertions to enforce user shardings without being overly restrictive and breaking valid use cases

PR #2443, opened 2025-12-02
[Common] Comm+GEMM overlap API updated to support cuBlasMp backend (incl. framework API)

denera:
# Description

This PR adds support for the NVTE cuBlasMp bindings in the Comm+GEMM overlap API.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -69,6 +77,30 @@ CommOverlapCore::CommOverlapCore(int myrank, int numranks, int mylocal, int numl
              num_comm_sm, set_sm_margin, use_ce, atomic_gemm);
 }
 
+CommOverlapCore::CommOverlapCore(int64_t nccl_comm_ptr, int tp_rank, int tp_size, int num_comm_sm,
+                                 bool is_p2p, bool atomic_gemm) {
+  _with_cublasmp = true;
```
ptrendx, line 88: This should probably check if TE was built with cublasmp and error out otherwise.

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
```diff
@@ -17,6 +18,12 @@
 
 #define NVTE_COMM_OVERLAP_MAX_STREAMS 3
 
+/* \brief Check if TE is built with cuBlasMp.
```
ptrendx: nit: cuBLASMp

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -526,6 +514,11 @@ class CommOverlapHelper : torch::CustomClassHolder {
                     ExtComm comm);
 
   void ub_barrier(ExtComm comm);
+
+  int64_t get_nccl_comm_ptr(std::string comm_name) {
+    NVTE_CHECK(backend_is_nccl, "Cannot get nccComm_t ptr if backend is not NCCL.");
```
ptrendx: This error message could be more descriptive - e.g. something like "chosen backend for the communication-computation overlap (cuBLASMp) requires NCCL communicator, but the passed ProcessGroup uses a different backend."

transformer_engine/common/CMakeLists.txt
```diff
@@ -271,7 +271,7 @@ if (NVTE_WITH_CUBLASMP)
     target_compile_definitions(transformer_engine PRIVATE NVTE_WITH_CUBLASMP)
     target_include_directories(transformer_engine PRIVATE ${CUBLASMP_DIR}/include ${NVSHMEM_DIR}/include)
     find_library(CUBLASMP_LIB
-                 NAMES cublasmp libcublasmp
+                 NAMES cublasmp libcublasmp.so.0
```
ptrendx: Do we need to be this specific?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -281,6 +283,8 @@ def collective_gemm_bootstrap(
         f" num_devices_per_process={num_devices_per_process}"
     )
     assert 0 <= process_id < num_total_devices, f"Invalid process_id={process_id}"
+    global collective_gemm_with_cublasmp
+    collective_gemm_with_cublasmp = use_cublasmp
```
ptrendx: Hmm, so that means that one either uses cublasMp for everything or for nothing?

build_tools/pytorch.py
```diff
@@ -96,6 +96,12 @@ def setup_pytorch_extension(
         libraries.append("nvshmem_host")
         cxx_flags.append("-DNVTE_ENABLE_NVSHMEM")
 
+    if bool(int(os.getenv("NVTE_WITH_CUBLASMP", 0))):
+        # Creating a cuBlasMp context requires direct access to the underlying NCCL
+        # communicator in a tensor-parallel process group. The header for ProcessGroupNCCL
+        # needs this CPP directive to be included properly.
+        cxx_flags.append("-DUSE_C10D_NCCL")
```
jberchtold-nvidia: To confirm my understanding, this additional build flag is only needed for the PyTorch extension, not the JAX extension?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -70,6 +70,7 @@
 
 
 num_cublas_streams = get_num_compute_streams()
+collective_gemm_with_cublasmp = False
```
jberchtold-nvidia: Why is this a global flag and not per-GEMM? If the restriction is truly global, can you add some error checking in the workspace size and actual gemm call that we are consistently using all cublasMP GEMM or no cublasMP GEMMs?

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -192,14 +193,22 @@ CommOverlapCore *CollectiveGemmPlanRegistry::get_executor(std::vector<size_t> bu
...
+        comm_handler.get_local_device_id_within_tp_domain(), comm_handler.tp_size,
+        comm_handler.get_tp_domain_id(), comm_handler.get_tp_num_domains(), comm_handler.tp_size,
+        comm_handler.allgather_func, comm_handler.barrier_func,
+        get_nvte_collective_op(collective_op), cgemm_config.num_max_streams, 1 /*comm_cga_size*/,
+        cgemm_config.gemm_priority, cgemm_config.comm_priority, cgemm_config.num_comm_sm,
+        true /*set_sm_margin*/, cgemm_config.use_ce, false /*atomic_gemm*/,
+        cgemm_config.aggregate_ag);
+  }
 
   CommOverlapCore *executor_ptr = executor.get();
```
jberchtold-nvidia, line 217: If we have multiple references to a unique_ptr, please instead use `shared_ptr`. I think technically this case won't use-after-free because the `plan_map`'s scope is larger, but to avoid potential future issues we can just make this a shared_ptr instead of `std:move`ing the unique_ptr but keeping a dangling reference to the underlying pointer.

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
```diff
@@ -17,6 +18,12 @@
 
 #define NVTE_COMM_OVERLAP_MAX_STREAMS 3
 
+/* \brief Check if TE is built with cuBLASMp.
+ *
+ * \return True if TE is built with cuBLASMp.
+ */
+bool nvte_built_with_cublasmp();
```
jberchtold-nvidia, line 26: Is this `nvte_built_with_cublasmp` declared with `extern "C"`? If we're aiming for a C-API here, I'm not sure if this is properly being exported or if it will contain C++ symbol name mangling

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -151,15 +149,17 @@ XLA_FFI_DEFINE_HANDLER_SYMBOL(CollectiveGemmInitHandler, CollectiveGemmInitFFI,
                                   .Attr<bool>("fuse_gelu")
                                   .Attr<bool>("grad")
                                   .Attr<bool>("use_split_accumulator")
-                                  .Attr<JAXX_Collective_Op>("collective_op"));
+                                  .Attr<JAXX_Collective_Op>("collective_op")
+                                  .Attr<bool>("use_cublasmp"));
```
jberchtold-nvidia: We need to be careful about breaking FFI interfaces. The JAX team stores HLO for some E2E testing and making breaking changes to the FFI interface has caused there tests to break in the past. If you'd like, I can share more info internally.

Can you make `use_cublasmp` default to false here if not specified in the FFI interface to avoid breaking previously generated HLO?

jberchtold-nvidia (dismissed):

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
ptrendx: Hmm, I don't think I agree with this comment (at least not without some additional consideration) - it basically all depends on the expected lifetimes and the ownership. Do we want to pass the ownership of the executor to the caller of this function? What can they do with it? If they can store it and use it a later date that is not tied to the plan_map lifetime, then I agree that we should instead pass a shared pointer. But if we expect the caller to have the lifetime that is for sure shorter than the plan_map then passing the raw pointer and relying on the unique pointer just for destruction is ok. Otherwise we could end up in a situation where it is not clear if we properly destroyed this executor (e.g. in the case of a GPU reset).

jberchtold-nvidia: Ah good point, I guess a crash due to use-after-free (won't happen currently but if we have an issue with lifetimes in future) is more clear to debug and potentially better than accidentally leaking GPU memory or other resources in a more subtle way. I'll resolve this thread

build_tools/pytorch.py
denera: Yes, this is just for PyT. The Comm+GEMM bootstrapping in TE/PyTorch relies on `c10d:ProcessGroupNCCL` and compiling with `USE_C10D_NCCL` exposes some API for extracting the NCCL communicator underneath that we have to pass into cuBLASMp. The same bootstrapping that @phu0ngng wrote for this in TE/JAX already gets the NCCL communicator from JAX itself, so we have everything we need there.

transformer_engine/jax/cpp_extensions/gemm.py
denera: The restriction is global. I streamlined this a little bit so that it is only ever set during initial bootstrapping, and none of the subsequent GEMM calls need to be told about it so that we can preserve the same FFI API as before. The cublas workspace sizes are controlled based on this too in the primitive abstract (set to 0 when using cuBLASMp so that cuBLASMp can handle its own workspace and we don't create any extra for the primitive).

transformer_engine/jax/csrc/extensions/gemm.cpp
denera: This is handled now in the latest commit alongside the change to how the global cublasmp vs. user buffers choice is handled, so CollectiveGEMM keeps the old FFI signature.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -23,6 +23,14 @@
 
 using namespace std::placeholders;
 
+bool nvte_built_with_cublasmp() {
+#ifdef NVTE_WITH_CUBLASMP
```
jberchtold-nvidia, line 28: Why do we have an option to specify whether to use cuBLASMp when initializing JAX's CGemm helper, but here whether or not we can use cuBLASMp seems dependent on the compilation flags?

Is the support matrix the following?
Built without `NVTE_WITH_CUBLASMP`: Can use userbuffers. Cannot use cuBLASMp
Built with `NVTE_WITH_CUBLASMP`: Can use BOTH userbuffers or cuBLASMp, depending on flag at import/initialization

So we need the flag at module init because it's possible a user has TE built with cuBLASMp support, but they'd like to test userbuffers? Or is userbuffers not supported when TE was built with `NVTE_WITH_CUBLASMP`

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
```diff
@@ -81,6 +92,9 @@ class CommOverlapCore {
                   int comm_priority, int num_comm_sm, bool set_sm_margin, bool use_ce,
                   bool atomic_gemm);
 
+  CommOverlapCore(int64_t nccl_comm_ptr, int tp_rank, int tp_size, int num_comm_sm, bool is_p2p,
```
jberchtold-nvidia: Can we make these constructors private and instead create factory functions, something like `make_user_buffer_comm_overlap_core` or `make_cublasmp_comm_overlap_core`? It isn't clear from reading this header that this particular constructor is for cuBLASMp only and the previous constructor is only for user buffers.

Or alternatively, if possible split these into two separate classes, one for UB and one for cuBLASMp that have a shared interface/base class.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
ptrendx: That support matrix looks ok to me - yes, we would like to be able to use both when compiled with cublasMp support.

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
denera: This C++ API needs to be refactored into a pure C API in a separate PR. I can make this distinction clearer in the scope of that effort if that's alright.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
denera: This particular call is just to preserve build configuration information at runtime (i.e. avoid or error-check against cublasmp code paths in Python when TE is not built with cublasmp). From a functional perspective though, userbuffers is always available regardless of whether TE was built with or without cublasmp.

jberchtold-nvidia: Got it, thanks! Then I'm on the same page now and will resolve this thread

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
jberchtold-nvidia: For now, can you add some brief documentation comments saying which constructor is for which backend so it's a bit clearer? Then the rest can be refactored in a follow up

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
jberchtold-nvidia: Resolving as other thread discussion mentioned the C API will be refactored in a follow-up

almogsegal:
FYI, cuBLASMp 0.8.0 is live: https://docs.nvidia.com/cuda/cublasmp/release_notes/index.html#cublasmp-v0-8-0

tests/pytorch/distributed/test_comm_gemm_overlap.py
```diff
@@ -82,6 +82,14 @@ def _run_gemm_with_overlap(
         if aggregate:
             test_cmd.append("--aggregate")
         if use_cublasmp:
+            if quantization == "mxfp8":
```
almogsegal, line 87: @denera cublasMp does support MXFP8. Have you faced any issue?

denera: I don't believe the TE/common bindings for cuBLASMp have MXFP8 enabled yet, and it's out of scope for this PR that's already large enough just to hook it up to the frameworks. I will add the MXFP8 support on our end in a subsequent PR.

build_tools/pytorch.py
```diff
@@ -88,6 +89,15 @@ def setup_pytorch_extension(
...
         cxx_flags.append("-DNVTE_ENABLE_NVSHMEM")
 
+    if bool(int(os.getenv("NVTE_WITH_CUBLASMP", 0))):
+        # Creating a cuBlasMp context requires direct access to the underlying NCCL
+        # communicator in a tensor-parallel process group. The header for ProcessGroupNCCL
+        # needs this CPP directive to be included properly.
+        cxx_flags.append("-DNVTE_WITH_CUBLASMP")
+        torch_lib_path = metadata.distribution("torch").locate_file("torch/lib")
+        library_dirs.append(torch_lib_path)
+        libraries.append("torch_cuda")
```
ptrendx: Is this NCCL communicator literally stored by cuBLASMp or is it copied somehow? I'm worried about the case where the process group gets destroyed and the NCCL communicator that was used in it also gets destroyed from underneath us.

examples/jax/collective_gemm/common.py
```diff
@@ -4,11 +4,14 @@
...
 
 import argparse
+import glob
+import os
 
 import jax
 import jax.numpy as jnp
 import numpy as np
 from jax.experimental import mesh_utils
+from jax.experimental.multihost_utils import sync_global_devices
```
ptrendx, line 14: What is the JAX version requirement for this function?

build_tools/pytorch.py
denera: Actually we don't use NCCL comm from PyTorch anymore. We create our own NCCL comm that spans the same devices as the PyTorch PG with the same TP ranks, so there's no risk of the NCCL comm disappearing if the PyT PG gets destroyed.

The comment you quoted here is leftover from an earlier iteration when we did try to extract the NCCL comm from PyT PG. I will remove it to avoid confusion, and I believe we might also no longer need to directly link to the `libtorch_cuda.so` anymore either because of this. I will double-check that and remove it as well.

tests/cpp_distributed/CMakeLists.txt
```diff
@@ -30,28 +32,45 @@ if(NOT DEFINED TE_LIB_PATH)
...
                test_comm_gemm.cu
                ../cpp/test_common.cu)
 
+list(APPEND test_comm_gemm_INCLUDES
+     ${CMAKE_SOURCE_DIR}/../../transformer_engine/common/include
+     ${CMAKE_SOURCE_DIR}/../../transformer_engine/common
+     ${CMAKE_SOURCE_DIR}/../../transformer_engine
+     ${CMAKE_SOURCE_DIR}
+     ${MPI_CXX_INCLUDE_PATH}
+     $ENV{CUBLASMP_HOME}/include)
```
ptrendx, line 52: Why do we need to include cuBLASMp headers in the test file? Are we exposing any API from cuBLASMp directly rather than through nvte_* functions?

tests/cpp_distributed/test_comm_gemm.cu
```diff
@@ -104,6 +104,96 @@ std::vector<T> CopyMatrix(const std::vector<T>& data, size_t mstart, size_t nsta
...
+  for (int r = 0; r < nranks; ++r) {
+    counts[r] = static_cast<int>(rows_per_rank[r] * sizeof(T));
+    if (r > 0) displs[r] = displs[r - 1] + counts[r - 1];
+  }
+
+  std::vector<T> gathered(global_rows * cols);
+  for (int64_t col = 0; col < cols; ++col) {
+    const auto* send = reinterpret_cast<const uint8_t*>(local.data() + col * local_rows);
+    auto* recv = reinterpret_cast<uint8_t*>(gathered.data() + col * global_rows);
+    CHECK_MPI(MPI_Allgatherv(send, static_cast<int>(local_rows * sizeof(T)), MPI_BYTE, recv,
```
ptrendx, line 192: Do we actually support different number of rows on each GPU to use this scheme both here and in the column allgather?

tests/cpp_distributed/test_comm_gemm.cu
```diff
@@ -240,24 +330,123 @@ class CommGemmFixure : public ::testing::TestWithParam<Params> {
...
+      std::vector<DType> out_ref(dims.d_rows_num * dims.d_cols_num);
+      NVTE_CHECK_CUDA(cudaMemcpy(out_ref.data(), d_ref.rowwise_dptr(),
+                                 out_ref.size() * sizeof(out_ref[0]), cudaMemcpyDefault));
+      out_golden = CastToFloat(out_ref);
+    } else {
+      // RS/AR reference: local partial GEMM (with bias) then float reduce.
+      //
+      // Epilogue ordering in the fused cuBLASMp kernel:
+      // - AllReduce: BIAS applied per-rank before AR, output = sum_r(A_r@B_r + bias) = sum_r(A_r@B_r) + nranks*bias
+      //              -> no bias correction needed in ref
```
ptrendx: Huh? Isn't this wrong?

examples/jax/collective_gemm/common.py
denera: It was first added [here](https://github.com/jax-ml/jax/commit/7613d2a5df482dbadf6e46d2fc9506a4fb18897c) and released with v0.3.1 (Feb 18, 2022).

The TE package already has a baked in `jax>=0.5.0` requirement that supersedes this so we don't need to specifically version-guard this `sync_global_devices()` API.

tests/cpp_distributed/test_comm_gemm.cu
denera: I don't think we ever have a use-case in TE (on the framework side) where the distributed matrix dim is not divisible by TP size but I believe cuBLASMp supports arbitrary global matrix dims without assuming that it is neatly divisible. The original `DistributeTensors` fixture in the CPP test suite sets a row/cols start and row/cols num for each operand on each calling rank, and my modifications to the CPP tests respect that existing convention when making sure we do AG/RS/AR comparisons between overlapped vs. non-overlapped distributed GEMM instead of doing overlapped distributed GEMM vs. global GEMM (which was causing numerical discrepancies before because the reference output was missing the reduction).

tests/cpp_distributed/test_comm_gemm.cu
denera: It's not wrong at least for now. This is a known bug in cuBLASMp and the fix is rolling out with v0.9.

I can gate the cuBLASMp versions in the CPP test and apply the bias correction to AR for v0.9+ so the tests don't break when our CI container bases update.

ptrendx: Wait, but isn't this behavior wrong? So we should guard against using this with bias and instead do bias manually for cuBLASMp < 0.9, right?

denera: Yeah, the behavior **is** wrong, but the bug is only in the all-reduce overlap, which we don't actually use or expose in the framework bindings.

Reduce-scatter overlap in cuBLASMp correctly applies the bias only once after the reduction so there's no need to guard against it.

transformer_engine/pytorch/module/linear.py
```diff
@@ -1032,6 +1035,17 @@ def _linear_backward(args: LinearBwdArgs) -> Tuple[Union[torch.Tensor, None], ..
...
         # --------------------------------------------------
 
+        # cuBLASMp's AG+GEMM consumes the gathered grad_output inline and does
+        # not preserve it for wgrad. Userbuffers leaves the gathered tensor in
+        # its persistent buffer; cuBLASMp does not, so we gather here.
+        if (
+            ctx.requires_wgrad
+            and ctx.ub_overlap_ag
+            and ctx.ub_obj_gradout is not None
+            and ctx.ub_obj_gradout.with_cublasmp()
```
ptrendx: `ctx` is not used here anymore - it should reference bwd_args.

denera: Yeah, I just saw this on my end too and fixed it. Catching branch up with main again and running tests. Will push up the fixes when everything's coming back green.

transformer_engine/pytorch/csrc/extensions/comm_gemm_overlap.cpp
```diff
@@ -116,14 +181,33 @@ void CommOverlapHelper::ub_barrier(ExtComm group) {
...
   NVTE_ERROR("Internal TE error: CommOverlapHelper::ub_barrier is a no-op when TE is compiled ",
              "with NVTE_UB_WITH_MPI=1!");
 #endif
 }
 
+ncclComm_t CommOverlapHelper::get_nccl_comm(std::string comm_name) {
+#ifdef NVTE_WITH_CUBLASMP
+  NVTE_CHECK(initialized, "Internal TE error: tex.CommOverlapHelper() is not initialized ",
+             "with valid process groups!");
+  NVTE_CHECK(backend_is_nccl,
```
ptrendx: Codex raised an issue (which I think is correct) that in the if statements in base.py:238 we default to mpi or gloo if those are available and not NCCL if you do not set the backend explicitly. This check though requires backend to be NCCL - we should reconcile that.

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -128,8 +143,45 @@ Error_Type GemmInitV2FFI(Buffer_Type lhs, Buffer_Type lhs_scale_inv, Buffer_Type
...
+                          convert_ffi_datatype_to_te_dtype(output->element_type()), d_shape);
+      TensorWrapper bias_(get_nvte_scaling_mode(JAXX_Scaling_Mode::NO_SCALING));
+      if (bias.element_count() > 0) {
+        bias_.set_rowwise_data(bias.untyped_data(),
+                               convert_ffi_datatype_to_te_dtype(bias.element_type()),
+                               std::vector<size_t>{static_cast<size_t>(bias.element_count())});
+      }
+      TensorWrapper pre_gelu_out_(get_nvte_scaling_mode(JAXX_Scaling_Mode::NO_SCALING));
+      // Match GemmV2FFI's operand swap: rhs becomes A, lhs becomes B.
+      cudaStream_t prepare_stream = cudaStreamPerThread;
```
ptrendx, line 175: You should synchronize that stream after calling that warmup iteration.

transformer_engine/pytorch/csrc/extensions/comm_gemm_overlap.cpp
denera: Yeah, this is a valid catch, but the fix is removing the NCCL requirement entirely. The check was leftover from when the bootstrapping was attempting to get a NCCL communicator directly out of the PyTorch PG, but we don't do that anymore. We simply create a new NCCL comm that spans the same devices/ranks as the PyTorch PG instead, and the PyTorch PG itself can have any backend in that case.

I'll make the fix on my end so it pushes up with the rest. Thanks!

examples/jax/collective_gemm/common.py
```diff
@@ -96,6 +99,20 @@ def _initialize_distributed(args):
...
+    # inside the GEMM custom call. Add COLLECTIVES so XLA captures those
+    # ops alongside the custom call instead of invalidating the capture.
+    # Lower the min-graph-size to 1 so single-matmul modules also get
+    # captured -- otherwise small test cases skip the captured path.
+    # Userbuffers does not need either flag.
+    if args.use_cublasmp:
+        xla_flags = os.environ.get("XLA_FLAGS", "")
+        os.environ["XLA_FLAGS"] = (
+            xla_flags
+            + " --xla_gpu_enable_command_buffer=+COLLECTIVES"
```
phu0ngng: Here, we need to make sure that CUSTOM CALL is captured by default. 
Would be better to add check for that?

examples/jax/collective_gemm/conftest.py
```diff
@@ -5,19 +5,29 @@
...
 @pytest.fixture(autouse=True)
 def distributed_args(request):
     """Fixture for querying distributed initialization arguments"""
     if request.cls:
+        use_cublasmp = request.config.getoption("--use-cublasmp")
+        if use_cublasmp and not nvte_built_with_cublasmp():
+            pytest.skip(
+                "Collective GEMM cuBLASMp backend tests require Transformer Engine to be built "
+                "with NVTE_WITH_CUBLASMP=1."
+            )
```
phu0ngng, line 30: Are we going to build CUBLASMP in our CI images by default?

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -128,8 +143,45 @@ Error_Type GemmInitV2FFI(Buffer_Type lhs, Buffer_Type lhs_scale_inv, Buffer_Type
       buffer_shape[0] = out_shape[0];
       buffer_shape[1] = out_shape[1];
     }
-    [[maybe_unused]] auto _ = CollectiveGemmPlanRegistry::getInstance().get_executor(
+    auto *executor = CollectiveGemmPlanRegistry::getInstance().get_executor(
         buffer_shape, buffer_dtype, config.collective_op);
+
+    // Run a dummy cuBLASMp matmul in the prepare stage so its lazy NCCL
```
phu0ngng, line 149: Could we make sure not to do any real GEMM computation here, rather than init NCCL mems/windows?

denera: Ideally yes, but currently no. There's a known bug in `ncclPutSignal` and `ncclWaitSignal` that does lazy initialization of RMA CE signal buffers, which means they're not CUDA graph-safe on first invocation. That forces us to do a dummy cuBlasMp GEMM call in the prepare phase just to trigger that lazy init so that invocations in the execute phase are graph-safe.

NCCL is aware of the bug and they're working on it. We can remove this dummy matmul here as soon as that bug is fixed.

examples/jax/collective_gemm/common.py
denera: CUSTOM_CALL is already part of the defaults for command buffers, so I don't believe we need to add it ourselves here.

The issue is that cuBLASMp has NCCL calls in it for device-initiated communication, and XLA exempts NCCL collectives from graph capture, so when NCCL collectives appear inside a CUSTOM_CALL chunk, we end up with a CUDA_ERROR_STREAM_CAPTURE_INVALIDATED error.

Adding COLLECTIVES to the command buffers is the only way I'm aware that gets around this issue. I don't know of any XLA API or flag that selectively enables command buffers for NCCL collectives only inside a CUSTOM_CALL.

examples/jax/collective_gemm/conftest.py
denera: I would like to but only starting with 26.05.

We need NCCL 2.30+ for cuBLASMp to be graph-safe. I'm told JAX containers are going to satisfy that requirement starting with 26.05 so we can safely install cuBLASMp without breaking anything else.

tests/cpp_distributed/CMakeLists.txt
denera: This is necessary now in order to version-guard all-gather tests against `CUBLASMP_VERSION >= 900`.

transformer_engine/pytorch/csrc/extensions/comm_gemm_overlap.cpp
```diff
@@ -75,9 +130,18 @@ CommOverlapHelper::CommOverlapHelper(c10d::ProcessGroup *world_group,
...
-  for (auto &pg : pgs) pg.second = nullptr;
+  for (auto &pg : torch_pgs) {
+    pg.second = nullptr;
+  }
+  torch_pgs.clear();
   backend_is_nccl = false;
   initialized = false;
+#ifdef NVTE_WITH_CUBLASMP
+  for (auto &comm : nccl_comms) {
+    NVTE_CHECK_NCCL(ncclCommDestroy(comm.second));
```
ptrendx: Are we sure that those communicators will not be used after the initialize_ub call finishes? Since the helper is going to be destroyed then and so the communicator also is going to be destroyed.

examples/jax/collective_gemm/common.py
phu0ngng: But then don't you need this mark this custom op with "collective"? 
Something like this https://github.com/tensorflow/tensorflow/commit/104721c892b9d937ad5ac9eb37940692f4a90588

transformer_engine/pytorch/csrc/extensions/comm_gemm_overlap.cpp
denera: You're right, these communicators are absolutely used by cuBLASMp after initialize_ub() and keeping them in the helper risks losing them.

I haven't seen this actually happen in practice, but I suspect that's because PyBind increments the reference count when the CommOverlapHelper is passed into the initialization for CommOverlap or CommOverlapP2P. Garbage collection does not wipe the helper out as a result even when it goes out of scope at the end of initialize_ub().

It's still not a good design though. It means that we actually leak this helper object — we can no longer access it in Python once it goes out of scope in initialize_ub() but it also doesn't get garbage collected because the reference count does not reach zero.

I'll see about doing the NCCL communicators in the CommOverlap/CommOverlapP2P init so that they get destroyed only when the overlap object is destroyed.

examples/jax/collective_gemm/common.py
denera: Commit message for this says:
```
Currently all FFIs are treated as opaque kernels which are further treated as compute kernels in LHS.

...

It introduces a annotation to specific compute_on="gpu_stream:collective", then it can be scheduled in the same way as native collectives and stream assignment will give it the collective stream which is of high priority and wont overlap with other collectives.
```

This doesn't sound like what we want for CollectiveGemm. This is not a custom collective. It's a compute op that has some internal communication (strictly P2P in TE/JAX, not collective) with a specific order-of-operations dependencies with the compute. We want XLA to treat it opaquely, and we want it to be invoked with the low priority compute stream that's used with the GEMM chunks while the internal high priority collective streams in Userbuffers or cuBLASMp are used for the overlapped communication (XLA does not know about these internal streams and doesn't need to).

transformer_engine/pytorch/csrc/extensions/comm_gemm_overlap.cpp
ptrendx: Ok, so should this check then be relaxed? It is still here.

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -671,10 +680,16 @@ class CommOverlapHelper : torch::CustomClassHolder {
 
   void ub_barrier(ExtComm comm);
 
-  ncclComm_t get_nccl_comm(std::string comm_name);
+  NcclCommSharedPtr get_nccl_comm(std::string comm_name);
 };
 
 class CommOverlap : torch::CustomClassHolder, public transformer_engine::CommOverlapBase {
+ private:
+  void *_warmup_workspace{nullptr};
```
ptrendx: Why do we keep this pointer throughout the execution if it is only needed during warmup?

transformer_engine/pytorch/csrc/extensions/comm_gemm_overlap.cpp
```diff
@@ -264,12 +266,13 @@ void cublasmp_capture_warmup(te::CommOverlapCore *core, int tp_size, te::CommOve
   const size_t b_bytes = b_shape[0] * b_shape[1] * bf16_bytes;
   const size_t d_bytes = d_shape[0] * d_shape[1] * bf16_bytes;
 
-  void *a_ptr = nullptr, *b_ptr = nullptr, *d_ptr = nullptr;
-  NVTE_CHECK_CUDA(cudaMalloc(&a_ptr, a_bytes));
-  NVTE_CHECK_CUDA(cudaMalloc(&b_ptr, b_bytes));
-  NVTE_CHECK_CUDA(cudaMalloc(&d_ptr, d_bytes));
+  NVTE_CHECK_CUDA(cudaMalloc(&warmup_workspace, a_bytes + b_bytes + d_bytes));
```
ptrendx: Ok, this is not right (well, it is actually resulting in a ~ok execution, but the intent is not correctly coded).
This warmup_workspace here is a local variable, so the cudaMalloc is modifying it rather than the _warmup_workspace from the class -> that variable stays as nullptr. If you really need to keep this pointer around then it should be passed as a pointer to pointer in order to be writable.
Now, if you just did that then that would expose another problem - at the end of this function you do cudaFree without setting the pointer back to nullptr, and so in the commoverlap destructor you would have a double free issue.
I don't see why you need to keep the workspace pointer, so probably just getting rid of that and keeping the pointers here completely local is the right choice, but please let me know why you wanted to keep that pointer in the first place.

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -645,11 +647,18 @@ void newton_schulz(int64_t ctx_ptr, int64_t m, int64_t n, at::Tensor x, int64_t
  **************************************************************************************************/
 
 class CommOverlapHelper : torch::CustomClassHolder {
+ public:
+  // Shared ownership of an ncclComm_t. The deleter calls ncclCommDestroy when
+  // the last reference (held by the helper and/or any CommOverlap consumers)
+  // is released, so the communicator outlives whichever owner is destroyed
+  // first.
+  using NcclCommSharedPtr = std::shared_ptr<std::remove_pointer<ncclComm_t>::type>;
```
ptrendx, line 659: This file does not include nccl.h directly.

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -2020,7 +2020,7 @@ def __init__(
         self.ub_overlap_rs_dgrad = ub_overlap_rs_dgrad and self.sequence_parallel
         # Bulk overlaps require the Userbuffers backend; the cuBLASMp backend
         # falls back to async NCCL ops via torch.distributed.
-        bulk_available = not is_userbuffer_cublasmp_backend()
+        bulk_available = not using_cublasmp_backend()
```
ptrendx: This will fail since it requires the UB communicators to be initialized, but this is called in every case, including single gpu. It should be guarded via the UB usage options.

transformer_engine/pytorch/csrc/extensions/comm_gemm_overlap.cpp
denera: >I don't see why you need to keep the workspace pointer, so probably just getting rid of that and keeping the pointers here completely local is the right choice, but please let me know why you wanted to keep that pointer in the first place.

This is what I had originally but Greptile complained about it leaking if the warmup GEMM call throws an exception.

I agree with you that I'd rather just have it be purely local, allocated before warmup GEMM and freed after. If you're okay with ignoring Greptile's warning about it, I can revert back to what it was before.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -134,39 +173,43 @@ void CommOverlapCore::initialize(int tp_size, int num_splits, int num_max_stream
...
 CommOverlapCore::~CommOverlapCore() {
-  cudaEventDestroy(_stop_comm);
-  cudaEventDestroy(_start_comm);
-  cudaEventDestroy(_stop_compute);
-  cudaEventDestroy(_start_compute);
-  if (_comm_launch_event) {
-    cudaEventDestroy(_comm_launch_event);
-  }
+  if (_with_cublasmp) {
+    nvte_comm_gemm_ctx_destroy(_cublasmp_ctx);
```
ptrendx, line 182: Does this touch the nccl communicator? If so, then the shared pointer is not good enough since its destructor is going to be called before the destructor of the base class and so the communicator is going to be destroyed before this function gets called.

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -99,6 +99,21 @@ std::tuple<TensorWrapper, std::vector<size_t>> xla_buffer_to_nvte_gemm_operand(
   return std::make_tuple(std::move(input), input_shape);
 }
 
+// Build a TensorWrapper for the prepare stage. Operand contents are
+// uninitialized at this point and no kernels are launched.
+static TensorWrapper prepare_operand_tensor(Buffer_Type buf, Buffer_Type scale_inv,
```
ptrendx, line 104: From Codex: this might not work for Hopper FP8 since it only creates the rowwise tensor, whereas Hopper can pass columnwise tensors to cuBLASMp. Not sure if this is a valid concern - if we only need to run the warmup on anything in order to get everything ready and there is no input dependency then passing only rowwise is fine.

tests/pytorch/distributed/test_comm_gemm_overlap.py
```diff
@@ -141,24 +164,26 @@ def _run_layer_with_overlap(
         raise AssertionError(result.stderr.decode())
 
 
+@pytest.mark.parametrize("use_cublasmp", (False, True))
```
ptrendx, line 176: You should skip cublasmp if support for it was not built in.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
denera: This does not touch the communicator. See here:
```
void nvte_comm_gemm_ctx_destroy(NVTECommGemmCtx* ctx) {
  NVTE_API_CALL(nvte_comm_gemm_ctx_destroy);
  if (ctx->workspace) {
    NVTE_CHECK_CUBLASMP(cublasMpBufferDeregister(ctx->grid_row_major.get(), ctx->workspace));
    NVTE_CHECK_CUBLASMP(cublasMpFree(ctx->grid_col_major.get(), ctx->workspace));
  }
  delete ctx;
}
```
The structure is deleted but it never invokes `ncclCommDestroy`.

`cuBlasMpFree` also does not touch the communicator. cuBLASMp assumes the communicator is an external resource. So it should be okay to keep the shared pointed at the TE/PyTorch level and clean them up where they're created in the first place.

transformer_engine/jax/csrc/extensions/gemm.cpp
denera: I've been testing on Hopper and Blackwell, and it's been working on both, so I don't think there's a dependency there on input dtype. The missing allocation we're using the warmup GEMM for is a CopyEngine RMA signal buffer, and the signal dtype is independent of the operands.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
ptrendx: Ok, then that works.

transformer_engine/pytorch/csrc/extensions/comm_gemm_overlap.cpp
ptrendx: Yes, I'm fine with that.

tests/pytorch/distributed/test_comm_gemm_overlap.py
denera: This is getting skipped in the runner function, lines 84-86:
```python
if use_cublasmp:
    if not tex.nvte_built_with_cublasmp():
        pytest.skip("Transformer Engine not built with cuBLASMp (NVTE_WITH_CUBLASMP=0).")
```

transformer_engine/common/include/transformer_engine/comm_gemm_overlap.h
```diff
@@ -109,6 +126,20 @@ class CommOverlapCore {
...
+                        bool grad, bool accumulate, cudaStream_t stream_main);
+
+  void cublasmp_gemm_rs(const TensorWrapper &A, bool transa, const TensorWrapper &B, bool transb,
+                        TensorWrapper &D, TensorWrapper &bias, TensorWrapper &pre_gelu_out,
+                        bool grad, bool accumulate, cudaStream_t stream_main);
+
+  void cublasmp_gemm_ar(const TensorWrapper &A, bool transa, const TensorWrapper &B, bool transb,
+                        TensorWrapper &D, TensorWrapper &bias, TensorWrapper &pre_gelu_out,
+                        bool grad, bool accumulate, cudaStream_t stream_main);
+
```
ptrendx: Why do we need those functions if we already have the nvte_* calls? In general this is supposed to be C API so we should not add and rely on even more C++ things to it.

denera: These exist only because the Userbuffers backend is all C++ and it's just keeping everything together.

I agree with you about the C API issue, but I think that's going to need to be a separate refactor PR entirely.

ptrendx (approved):

PR #2444, opened 2025-12-02
Add ccache support to TE and use it in GitHub actions

ptrendx:
# Description

Added ccache support to TE to speed up the compilation and using it in the github actions for faster CI.

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -82,7 +82,6 @@ jobs:
         env:
           NVTE_FRAMEWORK: pytorch
           MAX_JOBS: 1
-          SCCACHE_GHA_ENABLED: "true"
       - name: 'Sanity check'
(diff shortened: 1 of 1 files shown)


ptrendx:
I tried to convert the pyTorch and All build steps to use CCache too, but they have the different problem of requiring more disk space than is available on the machine -> Docker is run manually on those machines -> GitHub action for setting up the cache directory does not work there. In order to properly fix this I think we need the support for the common part building with build isolation (as a large portion of the problem is that pyTorch from pip brings in a whole new CUDA toolkit with it) @ksivaman.
So in the end I think we should just go ahead with the changes even if they do not benefit our GH CI and revisit the CI part once the support for the compilation with  build isolation is there.

ksivaman (approved):

PR #2473, opened 2025-12-03
[JAX] primitives for permutation on single GPU

tdophung:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes
- Implement primitives for each permutation triton kernels, using triton_call_lowering introduced in the triton binders in PR : 
- Include top level wrapper for token dispatch and combine 
- Add new tests for top level wrappers compared to reference pure jax implementations

# Checklist:


transformer_engine/jax/triton_extensions/utils.py
```diff
@@ -203,9 +207,22 @@ def lowering(ctx, x, *, block_size):
...
+
+    tensor_arg_idx = 0
+    for arg_name in arg_names:
+        if arg_name in constexpr_names:
+            # Skip constexpr/scalar args - they're specialized at compile time
+            continue
+        elif tensor_arg_idx < len(all_avals):
+            # Tensor arg - add to signature
+            signature[arg_name] = get_triton_dtype(all_avals[tensor_arg_idx])
+            tensor_arg_idx += 1
```
phu0ngng:
```suggestion
constexpr_names = set(constexprs.keys()) if constexprs else set()
tensor_arg_names = [n for n in arg_names if n not in constexpr_names]
signature = {n: get_triton_dtype(a) for n, a in zip(tensor_arg_names, all_avals)}
```

transformer_engine/jax/permutation.py
```diff
@@ -0,0 +1,479 @@
...
+    """
+    Dispatch tokens to experts based on routing map.
+
+    This is the forward pass of the MoE permutation. Tokens are scattered
+    to their designated experts according to the routing map.
+
+    Parameters
+    ----------
+    inp : jnp.ndarray
+        Input tensor of shape [num_tokens, hidden_size].
```
phu0ngng: This is not true. We should expect the input with the shape of `B, S, H`.

transformer_engine/jax/permutation.py
```diff
@@ -0,0 +1,479 @@
...
+    routing_map : jnp.ndarray
+        Routing mask of shape [num_tokens, num_experts]. Values: 1 = routed, 0 = not routed.
+    row_id_map : Optional[jnp.ndarray]
+        Pre-computed row ID map. If None, will be computed from routing_map.
```
phu0ngng: I don't think users have row_id_map before calling this VJP. 
I would not expose this to the API.

transformer_engine/jax/permutation.py
```diff
@@ -0,0 +1,479 @@
...
+
+    This is the forward pass of the MoE permutation. Tokens are scattered
+    to their designated experts according to the routing map.
+
+    Parameters
+    ----------
+    inp : jnp.ndarray
+        Input tensor of shape [num_tokens, hidden_size].
+    routing_map : jnp.ndarray
+        Routing mask of shape [num_tokens, num_experts]. Values: 1 = routed, 0 = not routed.
```
phu0ngng: `B, S, E` here.

transformer_engine/jax/permutation.py
```diff
@@ -0,0 +1,479 @@
...
+    row_id_map : Optional[jnp.ndarray]
+        Pre-computed row ID map. If None, will be computed from routing_map.
+    num_tokens : int
+        Number of input tokens.
+    num_experts : int
+        Number of experts.
+    num_out_tokens : int
+        Number of output tokens (total routed tokens).
+    hidden_size : int
+        Hidden dimension size.
```
phu0ngng: Should be inferred from the inputs.

transformer_engine/jax/permutation.py
```diff
@@ -0,0 +1,479 @@
...
+
+    output, _ = permute_with_mask_map(
+        inp,
+        row_id_map,
+        None,  # No probs
+        num_tokens,
+        num_experts,
+        num_out_tokens,
+        hidden_size,
+    )
```
phu0ngng: This is not how you write a VJP.

transformer_engine/jax/triton_extensions/permutation.py
```diff
@@ -0,0 +1,936 @@
...
+        assert RowIdMapPass1Primitive.inner_primitive is not None
+        return RowIdMapPass1Primitive.inner_primitive.bind(
+            routing_map,
+            num_tokens=num_tokens,
+            num_experts=num_experts,
+            block_size=block_size,
+        )
+
+    @staticmethod
+    def lowering(ctx, routing_map, *, num_tokens, num_experts, block_size):
```
phu0ngng: Why is the blocksize passed as an arg?

transformer_engine/jax/triton_extensions/permutation.py
```diff
@@ -0,0 +1,936 @@
...
+                probs_stride_token = 1
+                probs_stride_expert = 1
+        else:
+            probs_stride_token = 0
+            probs_stride_expert = 0
+
+        # Grid function equivalent: (num_tokens, cdiv(hidden_size, BLOCK_SIZE))
+        # For autotuned kernel, we use a fixed reasonable block size for grid calculation
+        block_size = 128
+        grid = (num_tokens, triton.cdiv(hidden_size, block_size))
```
phu0ngng: Why 128?

transformer_engine/jax/triton_extensions/permutation.py
```diff
@@ -0,0 +1,904 @@
...
+            routing_stride_token,
+            routing_stride_expert,
+            row_id_stride_token,
+            row_id_stride_expert,
+            grid=grid,
+            constexprs={"BLOCK_SIZE": block_size},
+        )
+
+
+register_primitive(RowIdMapPass1Primitive)
```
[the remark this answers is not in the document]
tdophung: This PR meant to onnly serve execution on single GPU first. Later PR will address these functions

transformer_engine/jax/permutation.py
tdophung: both this input shape and the routing map in the triton kernels expect a 2D shape of [num_tokens, hidden_size], so in these cases, we should flatten the first 2 dims into num_tokens= B * S then?

tdophung: for the sake of completeness, we discussed this in a sync and figured out that triton kernel only take a pointer to the blob of mem and uses some passed in strides to access it, so shape of the input data does not matter as long as size and strides are correct

transformer_engine/jax/permutation.py
jberchtold-nvidia: After the PR updates the VJPs are correct, so resolving this thread

transformer_engine/jax/permutation.py
```diff
@@ -0,0 +1,486 @@
...
+    - Backward: Permute gradients (scatter to experts)
+"""
+
+from functools import partial
+from typing import Optional, Tuple
+
+import jax
+import jax.numpy as jnp
+
+from transformer_engine.jax.triton_extensions.permutation import (
```
jberchtold-nvidia, line 25: Wanted to check my understanding: We are okay to import triton_extensions extensions here (which will trigger loading of triton itself, which may not be installed), because this `permutation.py` is not included in an `__init__.py` in this directory. So a user would have to explicitly import this file, in which case we do need to depend on Triton. Checking to make sure we don't have any places where we depend on Triton but don't actually need it. Thanks!

transformer_engine/jax/triton_extensions/permutation.py
```diff
@@ -0,0 +1,936 @@
...
+
+        assert routing_map_aval.shape == (
+            num_tokens,
+            num_experts,
+        ), f"routing_map shape mismatch: expected ({num_tokens}, {num_experts})"
+
+        row_id_map_shape = (num_tokens, num_experts * 2 + 1)
+        workspace_shape = (
+            num_experts,
+            triton.cdiv(num_tokens, 1024),
```
jberchtold-nvidia: Where does `1024` come from?

transformer_engine/jax/triton_extensions/permutation.py
```diff
@@ -0,0 +1,936 @@
...
+        return RowIdMapPass3Primitive.inner_primitive.bind(
+            row_id_map,
+            num_tokens=num_tokens,
+            num_experts=num_experts,
+        )
+
+    @staticmethod
+    def lowering(ctx, row_id_map, *, num_tokens, num_experts):
+        """MLIR lowering using triton_call_lowering."""
+        row_id_stride_token = num_experts * 2 + 1
```
jberchtold-nvidia, line 220: I see this `num_experts * 2 + 1` and the constant `1024` a handful of times. Should we move these to named helper functions or constants for clarity? Or is this a common calculation assumed to be understood. Thanks!

tests/jax/test_permutation.py
```diff
@@ -0,0 +1,1195 @@
...
+        test_output, test_probs = sort_chunks_by_map(
+            inp, row_id_map, probs, total_tokens, hidden_size, is_forward
+        )
+
+        ref_output, ref_probs = reference_sort_chunks_by_map(
+            inp, row_id_map, probs, total_tokens, hidden_size, is_forward
+        )
+
+        tols = dtype_tols(dtype)
+        assert_allclose(test_output, ref_output, **tols)
```
jberchtold-nvidia: If test_output and ref_output already have the dtype `dtype`, you don't need to do `tols = dtype_tols(dtype)`. That will be handled automatically inside `assert_allclose`. You only need it if you are using a different dtype (in which case `assert_allclose` takes a `dtype` arg I believe) or you need `tols` as an arg if you're intentionally modifying the rtol or atol

tests/jax/test_permutation.py
```diff
@@ -0,0 +1,1195 @@
...
+        expected_loss = jnp.sum(ref_output**2)
+        assert_allclose(loss_val, expected_loss, **tols)
+
+        # Compute reference gradient manually:
+        # d_loss/d_inp = unpermute(d_loss/d_output)
+        # where d_loss/d_output = 2 * ref_output
+        output_grad = 2 * ref_output
+
+        # Reference backward: unpermute the gradient
+        ref_grad = reference_token_combine(output_grad, row_id_map, None)
```
jberchtold-nvidia: You shouldn't need to compute the gradient manually for the reference. Computing `value_and_grad` of the ref impl like you do for the TE primitive should be sufficient

tests/jax/test_permutation.py
```diff
@@ -0,0 +1,1195 @@
...
+
+        # Use more relaxed tolerance for gradient comparison due to compounding numerical differences
+        grad_tols = dtype_tols(dtype, rtol=tols["rtol"] * 10, atol=tols["atol"] * 10)
```
jberchtold-nvidia: I'm not that familiar with the operation, but 10x tolerances seems quite high. If you use the non-relaxed dtype tols, what sort of errors are you getting? How large and what percentage of values mismatch? If we see large errors like this, it could be something subtle in our pure-JAX reference that is implicitly type promoting and computing in fp32 that is causing the mismatch.

transformer_engine/jax/permutation.py
```diff
@@ -0,0 +1,486 @@
...
+
+
+def _token_dispatch_fwd_rule(
+    inp: jnp.ndarray,
+    routing_map: jnp.ndarray,
+    num_out_tokens: int,
+) -> Tuple[Tuple[jnp.ndarray, jnp.ndarray], Tuple[jnp.ndarray, int, int, int]]:
+    """Forward pass rule for token_dispatch."""
+    # Infer dimensions from input shapes
+    num_tokens = inp.shape[0] * inp.shape[1] if inp.ndim == 3 else inp.shape[0]
```
mingxu1067: Could we add an assertion like `assert input.ndim in [2, 3]`?

transformer_engine/jax/permutation.py
```diff
@@ -0,0 +1,486 @@
...
+
+_token_dispatch.defvjp(_token_dispatch_fwd_rule, _token_dispatch_bwd_rule)
+
+
+# =============================================================================
+# Token Dispatch with Probs with VJP
+# =============================================================================
+
+
+def token_dispatch_with_probs(
```
mingxu1067: Could we merge this into `token_dispatch` since JAX accepts None as inputs.

mingxu1067 (dismissed):
Overall, LGTM, some nitpicking. Approved.

transformer_engine/jax/permutation.py
tdophung: yes this is intentional because we would not want to havee triton as a dependency if permutation.py (and later ops that usees triton kernels) are not imported

transformer_engine/jax/triton_extensions/permutation.py
tdophung: it is the default block size specified in make_row_id_map.
Howver I will change this into a macro for ease of changing it if needed. Alternatively, in the future, we can use autotuning to figure out?

transformer_engine/jax/triton_extensions/permutation.py
tdophung: the 1024 is just the default block size which I will change into a macro. the num_experts * 2 + 1 is the number of columns in the row_id_map as the column 0 -> E-1 is for destination row indices (where token will go), E -> 2E-1 is for ccorresponding expert id, and 2E is for number of experts, a specific token is routed to.

Personally I think putting num_experts * 2 + 1 is clear in these scenarios

transformer_engine/jax/triton_extensions/permutation.py
```diff
@@ -0,0 +1,1136 @@
...
+    Returns
+    -------
+    row_id_map : jnp.ndarray
+        The row_id_map for the permutation of shape `[num_tokens, num_experts * 2 + 1]`.
+        For each token, the last item is the number of experts that are routed (n_routed).
+        The first n_routed items are the destination row indices in the permuted tokens.
+        The [num_experts, num_experts + n_routed) items are the indices of the experts corresponding
+        to the first n_routed row indices above.
+    """
+    block_size = DEFAULT_BLOCK_SIZE
```
jberchtold-nvidia: Since the block size can't be controlled by the user, let's remove this block size here as well as the block size argument to all the primitives. After that, all block sizes will be computed directly from the Triton autotuned kernel min block size and we won't need to pass block_size in as a param or even have a `DEFAULT_BLOCK_SIZE`, since all our kernels are autotuned

transformer_engine/jax/triton_extensions/permutation.py
```diff
@@ -0,0 +1,1136 @@
...
+    "make_chunk_sort_map",
+    "sort_chunks_by_map",
+]
+
+DEFAULT_BLOCK_SIZE = 1024
+
+
+def _get_min_block_size(kernel, default=128):
+    if hasattr(kernel, "configs"):
+        return min(config.kwargs.get("BLOCK_SIZE", default) for config in kernel.configs)
```
jberchtold-nvidia: I believe all our Triton kernels are autotuned with block size, right?  If so, let's replace
```
return default
```
with 
```
raise ValueError("Non-autotuned Triton kernels are not supported currently")
```
since we aren't testing our block size logic on non-autotuned kernels. Just in case someone wants to add a non-autotuned kernel in the future, I'd prefer it  clearly gives  them an  untested/unsupported error rather than  something more subtle  like mismatching block sizes somewhere  in our  code that leads to some  subtler error  like  uninitialized values

jberchtold-nvidia (approved):
LGTM, thanks!

PR #2476, opened 2025-12-04
fix ce loss calculation when some tokens are ignored

yashaswikarnati:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/triton/cross_entropy.py
+++ b/transformer_engine/pytorch/triton/cross_entropy.py
@@ -66,7 +66,7 @@ def cross_entropy_forward(
         rank=rank,
         n_cols=V,
         ignore_idx=ignore_idx,
-        n_non_ignore_ptr=n_non_ignore,
+        n_non_ignore=n_non_ignore,
         BLOCK_SIZE=BLOCK_SIZE,
         num_warps=32,
     )


tests/pytorch/test_parallel_cross_entropy.py
```diff
@@ -89,7 +89,7 @@ def one_iteration_test(
         # Check that loss and grad input match
         tols = dtype_tols(dtype)
         test_loss = test_loss.to(dtype=torch.float64, device="cpu")
-        ref_loss = test_loss.to(dtype=torch.float64, device="cpu")
+        ref_loss = ref_loss.to(dtype=torch.float64, device="cpu")  # FIX: was test_loss
```
timmoon10: Once this PR is merged, this comment would provide no value in helping you understand the code. The right place to explain line-by-line changes is in the PR description, not in the code itself.
```suggestion
        ref_loss = ref_loss.to(dtype=torch.float64, device="cpu")
```

tests/pytorch/test_parallel_cross_entropy.py
```diff
@@ -154,3 +154,16 @@ def test_ignore_idx(self):
                 reduce_loss=False,
                 ignore_idx=True,
             )
+
+    def test_ignore_idx_reduced_loss(self):
+        """Test ignore_idx with reduce_loss=True - previously missing test case."""
```
timmoon10:
```suggestion
        """Test ignore_idx with reduce_loss=True."""
```

transformer_engine/pytorch/triton/cross_entropy.py
```diff
@@ -36,6 +36,10 @@ def cross_entropy_forward(
     B, SQ, V = _input.shape
     n_rows = B * SQ
 
+    n_non_ignore = (target.view(-1) != ignore_idx).sum().item()
```
timmoon10: Issues:
- This will introduce a GPU synchronization, which hurts performance and breaks CUDA Graphs.
- For the distributed case, we need to reduce over the ranks.

This makes me think we need to modify `online_softmax_kernel` to count the number of non-ignored entries.

transformer_engine/pytorch/triton/cross_entropy.py
```diff
@@ -90,14 +94,16 @@ def cross_entropy_forward(
         world_size=world_size,
         ignore_idx=ignore_idx,
         n_cols=V,
-        n_non_ignore=n_rows,
+        n_non_ignore=n_non_ignore,
```
timmoon10, line 100: I see that the kernel uses `n_non_ignore` both to access an intermediate buffer and to scale the reduced loss:
https://github.com/NVIDIA/TransformerEngine/blob/8a37042e5cfa9fe0a55c1e6481c6c946bf23637c/transformer_engine/common/triton/cross_entropy.py#L143
https://github.com/NVIDIA/TransformerEngine/blob/8a37042e5cfa9fe0a55c1e6481c6c946bf23637c/transformer_engine/common/triton/cross_entropy.py#L209

Won't this change break accesses to the intermediate buffer? I think we should pass in a GPU buffer for the number of non-ignored entries and we should rename `n_non_ignore` to `n_rows` to be less deceptive.

timmoon10 (commented):
Thanks for catching the bug. Unfortunately, it may require non-trivial changes to the kernel.

CC @sanandaraj5597 @BestJuly

transformer_engine/pytorch/triton/cross_entropy.py
sanandaraj5597: We split over vocab and not sequence, so we might not do the distributed all gather. But I agree with Tim, we should not do a `.item()` for sure.

sanandaraj5597:
@yashaswikarnati I think we should make n_rows as a separate argument to the cross entropy forward kernel and replace non_ignore usages with n_rows.

To fix this issue, we need to add a separate non_ignore argument which should be a GPU tensor with one value on device. Then we should index and load this value inside the kernel and use it inside the kernel.

transformer_engine/common/triton/cross_entropy.py
```diff
@@ -140,7 +143,7 @@ def cross_entropy_kernel(
     ori_X_y = tl.load(m_d_X_y_ptr + (2 * m_d_X_y_stride))
 
     for i in range(1, world_size):
-        offset = i * 3 * n_non_ignore * m_d_X_y_stride
+        offset = i * 3 * n_rows * m_d_X_y_stride
```
sanandaraj5597, line 153: There are 2 other places where we use n_non_ignore, probably those should be replaced with n_rows too?

yashaswikarnati: sorry, may be I'm missing something. We only use n_non_ignore for normalization, is there a place we are incorrectly using n_non_ignore?

sanandaraj5597: We use here: https://github.com/NVIDIA/TransformerEngine/blob/fd0cd12e912a101173275b15e10c3906f20d4551/transformer_engine/common/triton/cross_entropy.py#L177
and here:
https://github.com/NVIDIA/TransformerEngine/blob/fd0cd12e912a101173275b15e10c3906f20d4551/transformer_engine/common/triton/cross_entropy.py#L209

sanandaraj5597:
LGTM, would appreciate a second eye from @timmoon10

transformer_engine/pytorch/triton/cross_entropy.py
```diff
@@ -63,10 +65,14 @@ def cross_entropy_forward(
         m_d_X_y_stride=m_d_X_y.stride(-1),
         rank=rank,
         n_cols=V,
+        ignore_idx=ignore_idx,
+        n_non_ignore_ptr=n_non_ignore,
         BLOCK_SIZE=BLOCK_SIZE,
         num_warps=32,
     )
 
+    n_non_ignore = torch.clamp(n_non_ignore, min=1)
```
timmoon10, line 74: To handle the distributed case, do we need to reduce the number of non-ignored entries? Something like:
```python
    world_size = 1 if dist_process_group is None else dist.get_world_size(dist_process_group)
    if world_size > 1:
        m_d_X_y_gathered = torch.zeros(
            n_rows * 3 * world_size, dtype=torch.float32, device=_input.device
        )
        dist.all_gather_into_tensor(m_d_X_y_gathered, m_d_X_y, group=dist_process_group)
        dist.all_reduce(n_non_ignore, group=dist_process_group)
    else:
        m_d_X_y_gathered = m_d_X_y

    n_non_ignore = torch.clamp(n_non_ignore, min=1)
```

yashaswikarnati: resolved offline, sequence dimension is never sharded here, so dont have to reduce.

PR #2485, opened 2025-12-05
[JAX] Remove unused TE DPA module dtype which fixes cuDNN backend detection to properly use input dtypes

jberchtold-nvidia:
# Description

Currently TE/JAX's `te_flax.DotProductionAttention` stores a module `dtype` for both the fused and unfused module backends. This is okay for modules that create parameters. However, TE's DPA does not create any parameters. All QKV and optional bias are passed as input arguments to the layer. Therefore, this dtype is unused and can cause us to incorrect fallback to the unfused backend even if the inputs are bf16 because the module dtype defaults to fp32, even though the dtype is never used.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Remove unnecessary `dtype` field from `DotProductAttention`, `_FusedDotProductAttention` and `_UnfusedDotProductAttention`
- In DPA, assert Q, K, V have the same dtype
- In DPA, cast bias to QKV dtype if it isn't already.
- In DPA, fix cuDNN fused backend check to use QKV dtype instead of unused module dtype

# Checklist:

### DIFF
--- a/tests/pytorch/test_quantized_tensor.py
+++ b/tests/pytorch/test_quantized_tensor.py
@@ -20,6 +20,7 @@ from transformer_engine.pytorch import (
     Float8Tensor,
     MXFP8Tensor,
     NVFP4Tensor,
+    QuantizedTensor,
 )
 
 from transformer_engine.pytorch.utils import is_non_tn_fp8_gemm_supported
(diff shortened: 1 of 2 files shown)


transformer_engine/jax/flax/transformer.py
```diff
@@ -638,6 +630,24 @@ def __post_init__(self):
...
         super().__post_init__()
 
+    def _assert_dtypes(self, query: Array, key: Array, value: Array, qkv_layout: QKVLayout):
+        """Asserts that the dtypes of query, key, and value dtypes are consistent."""
+        if qkv_layout.is_qkvpacked():
+            pass  # No need to check dtypes for key and value since it is packed
+        elif qkv_layout.is_kvpacked():
+            assert (
+                key.dtype == query.dtype
+            ), f"Expected kv dtype={key.dtype} to match query dtype={query.dtype}."
```
KshitijLakhani: nit: Consider replacing f"Expected kv dtype={key.dtype} to match query dtype={query.dtype}." with more concise string f"Expected {key.dtype=} to match {query.dtype=} for {qkv_layout=}"

KshitijLakhani (changes requested):
Overall the change makes sense to me. We should infer the type from Q,K,V and use it for checking the backend, rather than risk users possibly giving mismatched dtype and Q,K,V types thereby falling to the unfused path.

I am concerned about users of the `DotProductAttention()` API like Maxtext that explicitly set the dtype in DPA : https://github.com/AI-Hypercomputer/maxtext/blob/7ebcc9a39cec69463c24be0dfe5ef168caf6abd0/src/MaxText/layers/attention_op.py#L1388

This will be a breaking change for them and I think they may see a  "Unexpected keyword argument" error. We need to find a more graceful way to do this.
I am thinking that we add a deprecation warning in `DotProductAttention::__post_init()` which informs users that this param will be removed in a future version, let's say 2.13 for e.g., so that they can track this as well. And then maybe we add this change in 2.13 

Until then, maybe could add a check in DPA to ensure that the Q,K,V dtype matches self.dtype so that we are still able to achieve the objective of this PR.

Welcome any other suggestions as well.

Lastly, we should make a Maxtext PR to go along with any DPA API change we make as part of this PR or a subsequent PR. We can do it ourselves, or loop in Faysal for the same

KshitijLakhani:
cc: @ksivaman We are still discussing the best way to merge this change in, however, either now or in the future, whenever this API changes, it will be a breaking change and we would need to mention this in our docs. Just want to keep you in the loop.

Issue #2446, opened 2025-12-03
Support MXFP8/NVFP4 tensors with pre-swizzled scales


timmoon10:
**Is your feature request related to a problem? Please describe.**

Disentangle scale swizzling from GEMM.

**Describe the solution you'd like**

- Add logic in C++ tensor class whether row-wise/col-wise scales are swizzled or not
- Add logic in PyTorch quantized tensors whether row-wise/col-wise scales are swizzled
- Add hints in quantizer whether pre-swizzling is helpful

**Describe alternatives you've considered**

**Additional context**

PR #2019, opened 2025-08-01
[PyTorch] Multi-tensor swizzle scaling factors for MXFP8 and fuse padding zeros

yaox12:
# Description

Please include a brief summary of the changes, relevant motivation and context.

Fixes # (issue)

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Please list the changes introduced in this PR:

- Change A
- Change B

# Checklist:

- [ ] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [ ] The functionality is complete
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2486, opened 2025-12-06
Add logic for block-scaled tensors with GEMM swizzled scales

timmoon10:
# Description

All of the supported block-scaled tensor formats (MXFP8, NVFP4, DSv3 FP8) have two ways of ordering their scaling factors:
- "Compact" ordering for quantization, dequantization, and communication
- "Swizzled" ordering for GEMM

The core infrastructure handles this in an ad hoc way, blindly assuming that the "right" scale ordering is used for the different operations. The PyTorch infrastructure only supports MXFP8 and NVFP4 scales are in compact order, although DSv3 FP8 does have awareness of "compact" and "GEMM-ready" formats. This situation makes it hard to implement fused kernels that can bypass the swizzle kernel.

This PR adds a` with_gemm_swizzled_scales` field in the C++ tensor class so that the core infrastructure can distinguish between the different scale orderings. It also adds this field in the PyTorch quantized tensor classes, and exposes a `optimize_for_gemm` option in the quantizer so that we can create tensors that do not need communication or checkpointing. Finally, it rips out all the DSv3 FP8 infrastructure for the compact format, which is no longer necessary.

Progress
- [x] MXFP8
- [x] DSv3 FP8
- [x] NVFP4
- [ ] ~Add option to pre-swizzle weights~
- [x] Pre-swizzle activations
- [x] Fused MXFP8 quantize + swizzle

Closes https://github.com/NVIDIA/TransformerEngine/issues/2446.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Support GEMM swizzled scales in C++ tensor class
- Support GEMM swizzled scales in PyTorch quantized tensor classes
- Support `optimize_for_gemm` option in PyTorch quantizer
- Expose PyTorch function to swizzle scales
- Support MXFP8 quantization with pre-swizzled scales
- Enable fused quantize+swizzle kernels in linear module and related
- Remove DSv3 FP8 compact data format. It was used to avoid all-gather interleaving, which we can now fix with the swap-first-dims kernel.

# Checklist:

### DIFF
--- a/transformer_engine/common/cast/dispatch/quantize.cuh
+++ b/transformer_engine/common/cast/dispatch/quantize.cuh
@@ -291,15 +291,10 @@ void quantize_bwd_helper(const NVTETensor grad, const NVTETensor input, NVTETens
       FP8BlockwiseRowwiseOption rowwise_option = FP8BlockwiseRowwiseOption::NONE;
       FP8BlockwiseColumnwiseOption columnwise_option = FP8BlockwiseColumnwiseOption::NONE;
       if (output_tensor->has_data()) {
-        const bool rowwise_compact = !output_tensor->with_gemm_swizzled_scales;
-        rowwise_option = rowwise_compact ? FP8BlockwiseRowwiseOption::ROWWISE_COMPACT
-                                         : FP8BlockwiseRowwiseOption::ROWWISE_GEMM_READY;
+        rowwise_option = FP8BlockwiseRowwiseOption::ROWWISE_GEMM_READY;
       }
       if (output_tensor->has_columnwise_data()) {
-        const bool columnwise_compact = !output_tensor->with_gemm_swizzled_scales;
-        columnwise_option = columnwise_compact
-                                ? FP8BlockwiseColumnwiseOption::COLUMNWISE_COMPACT
-                                : FP8BlockwiseColumnwiseOption::COLUMNWISE_GEMM_READY;
+        columnwise_option = FP8BlockwiseColumnwiseOption::COLUMNWISE_GEMM_READY;
       }
       quantize_transpose_vector_blockwise(
           grad_tensor->data, output_tensor->scale_inv, output_tensor->columnwise_scale_inv,


transformer_engine/common/transformer_engine.cpp
```diff
@@ -803,7 +808,148 @@ NVTEBasicTensor nvte_get_tensor_param(const NVTETensor tensor, NVTETensorParam p
     case kNVTEColumnwiseAmax:
       return t.columnwise_amax;
     default:
-      NVTE_ERROR("Unknown tensor parameter!");
+      NVTE_ERROR("Unsupported tensor parameter (", static_cast<int>(param_name),
+                 "). Consider using nvte_set_tensor_param_v2 instead.");
+  }
+}
+
+void nvte_set_tensor_param_v2(NVTETensor tensor, NVTETensorParam param, const void *buf,
```
zhongbozhu, line 816: Why do we need a v2 here?

timmoon10: I don't want to break the existing APIs. That said, this PR isn't fully backward-compatible because the GEMM no longer secretly assumes that MXFP8 scales are swizzled.

tests/pytorch/test_float8_blockwise_scaling_exact.py
```diff
@@ -87,126 +87,6 @@ def initialize_for_many_scales(
...
-        (2048, 2000),
-        (2000, 1024),
-        (2048, 1024),
-    ],
-)
-@pytest.mark.parametrize("x_dtype", [torch.float32, torch.bfloat16], ids=str)
-@pytest.mark.parametrize("quant_dtype", [torch.float8_e4m3fn, torch.float8_e5m2], ids=str)
-@pytest.mark.parametrize("eps", [0], ids=["eps_0"])
-@pytest.mark.parametrize("pow_2_scales", [True], ids=["pow2scales"])
-def test_quantization_1D_block_tiling_with_compact_data_and_scales(
```
ptrendx, line 114: Why don't we need this test anymore?

tests/pytorch/test_fusible_ops.py
```diff
@@ -2764,25 +2767,18 @@ def test_layernorm_mlp(
...
         layernorm_epsilon: float = 1e-5,
     ) -> None:
-        """
-        LayerNorm/RMSNorm + Linear + GELU + Linear
-
-        Note that this test checks only if the module runs
-        as when chaining multiple modules it is hard to validate
-        numerical accuracy.
-        """
+        """LayerNorm/RMSNorm + Linear + SwiGLU + Linear"""
```
ptrendx, line 2781: Not sure how this change is relevant here to be honest.

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -150,17 +150,10 @@ void quantize_fwd_helper(const NVTETensor input, NVTETensor output,
       FP8BlockwiseRowwiseOption rowwise_option = FP8BlockwiseRowwiseOption::NONE;
       FP8BlockwiseColumnwiseOption columnwise_option = FP8BlockwiseColumnwiseOption::NONE;
       if (output_tensor->has_data()) {
-        bool rowwise_compact = (quant_config_cpp.float8_block_scale_tensor_format ==
-                                Float8BlockScaleTensorFormat::COMPACT);
-        rowwise_option = rowwise_compact ? FP8BlockwiseRowwiseOption::ROWWISE_COMPACT
-                                         : FP8BlockwiseRowwiseOption::ROWWISE_GEMM_READY;
+        rowwise_option = FP8BlockwiseRowwiseOption::ROWWISE_GEMM_READY;
```
ptrendx, line 153: Why are you always choosing the gemm ready version here?

ptrendx: Shouldn't it have a similar logic to check the with_gemm_swizzled_scales?

transformer_engine/common/cast/mxfp8/gated_mxfp8.cuh
```diff
@@ -722,113 +764,140 @@ void quantize_gated(const Tensor &gated_input, const Tensor &grad, Tensor *outpu
...
+
+              const size_t out_act_mem = buff_size_aligned_out;
+              const size_t out_gate_mem = (IS_BWD ? buff_size_aligned_out : 0);
+              size_t out_mem = out_act_mem + out_gate_mem;
+
+              if (USE_ROWWISE_SCALING && USE_COLWISE_SCALING) { out_mem *= 2; }
+
+              const size_t shmem_size = in_mem + out_mem + TMA_SHMEM_ALIGNMENT;
+
+              // Zero out swizzled scales if padding is needed
```
ptrendx, line 820: Don't we already do this though when we create the scaling factor tensors? This seems like a pessimization, as we will do this now every time instead of once - we should instead note the requirement for the scaling factors to be zeroed out before the quantization in the docs of the quantization call.

transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh
```diff
@@ -619,168 +649,195 @@ void quantize(const Tensor &input, const Tensor *act_input, const Tensor *noop,
...
+              constexpr size_t act_input_mem = (IS_DACT ? buff_size_aligned_in : 0);
+              constexpr size_t in_mem = elt_input_mem + act_input_mem;
+
+              const size_t out_rowwise_mem = (use_rowwise_scaling ? buff_size_aligned_out : 0);
+              const size_t out_colwise_mem = (use_colwise_scaling ? buff_size_aligned_out : 0);
+              const size_t out_mem = out_rowwise_mem + out_colwise_mem;
+
+              const size_t dshmem_size = in_mem + out_mem + TMA_SHMEM_ALIGNMENT;
+
+              // Zero out swizzled scales if padding is needed
```
ptrendx, line 759: Same comment as with the gated.

transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh
```diff
@@ -41,9 +41,24 @@ constexpr size_t TOTAL_BANKS_WIDTH = (32 * 4) / 1;  // 128
 // Number of threads (rowwise scaling) that span 32 banks (4-byte banks) of shared memory
 constexpr size_t THREADS_PER_BANK = TOTAL_BANKS_WIDTH / SCALE_DIM_X;  // 4 = 128 / 32
 
+// Convert compact scale indices into GEMM swizzled scale index
+__device__ __forceinline__ size_t gemm_swizzled_scale_idx(size_t i, size_t j, size_t num_tiles_X) {
```
ptrendx: This function seems ~same as the one in the gated file, should it be some mxfp8 utility instead? Or is this logic too tied to the kernel itself? If the issue is the tile dimension, it could be a template.

transformer_engine/common/cast/nvfp4/quantize_nvfp4.cuh
```diff
@@ -142,17 +142,17 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK)
   constexpr size_t buff_size_aligned_out_mxfp8 =
       DIVUP_TO_MULTIPLE(buff_elems_total * sizeof(OType), TMA_SHMEM_ALIGNMENT);
 
-  constexpr size_t buff_size_nvfp4_scales =
-      CHUNK_DIM_Y * (CHUNK_DIM_X / SCALE_DIM_X) * sizeof(fp8e4m3);
-  constexpr size_t buff_size_mxfp8_scales =
-      (CHUNK_DIM_Y / SCALE_DIM_Y) * CHUNK_DIM_X * sizeof(fp8e8m0);
+  // constexpr size_t buff_size_nvfp4_scales =
```
ptrendx: Commented out code - why not remove it if it is not needed?

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
```diff
@@ -172,7 +172,17 @@ CommOverlapCore::~CommOverlapCore() {
 
 TensorWrapper CommOverlapCore::get_tensor_chunk(const TensorWrapper &source, size_t chunk_offset,
                                                 const std::vector<size_t> &chunk_shape) {
+  // Check tensor format
   const auto scaling_mode = source.scaling_mode();
+  NVTE_CHECK(scaling_mode == NVTE_DELAYED_TENSOR_SCALING || scaling_mode == NVTE_MXFP8_1D_SCALING,
```
ptrendx, line 177: We don't support the blockwise tensor in UB? This is a gap then.

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -561,6 +575,13 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
       NVTE_CHECK(cublas_version() >= 120900,
                  "FP8 block scaling requires cuBLAS 12.9+, but run-time cuBLAS version is ",
                  cublas_version());
+
+      // Check that matrix formats are valid
+      NVTE_CHECK((!(inputA->scaling_mode == NVTE_BLOCK_SCALING_2D &&
+                    inputB->scaling_mode == NVTE_BLOCK_SCALING_2D)),
+                 "Only 1D by 1D, 1D by 2D, and 2D by 1D block scaling supported, but got 2D by 2D");
```
ptrendx:
```suggestion
                 "Only 1D by 1D, 1D by 2D, and 2D by 1D block scaling GEMM is supported, but got 2D by 2D");
```

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -57,24 +57,24 @@ NVTEMatmulConfig nvte_create_matmul_config();
...
 
 /*! \brief Set an option in matrix multiplication configuration.
  *
- *  \param[in] config Matrix multiplication configuration.
- *  \param[in] attr Option type.
- *  \param[out] buf Memory address to read option value.
- *  \param[in] size_in_bytes Size of buf.
+ *  \param[in/out] config        Matrix multiplication configuration.
+ *  \param[in]     attr          Option type.
+ *  \param[in]     buf           Memory address to read option value.
```
ptrendx:
```suggestion
 *  \param[in]     buf           Memory address to read option value from.
```

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -57,24 +57,24 @@ NVTEMatmulConfig nvte_create_matmul_config();
...
- *  \param[in] attr Option type.
- *  \param[out] buf Memory address to write option value. Ignored if
- *                  NULL.
- *  \param[in] size_in_bytes Size of buf.
- *  \param[out] size_written Number of bytes that have been written to
- *                           buf. If buf is NULL, then the number of
- *                           bytes that would have been written.
+ *  \param[in]  config        Matrix multiplication configuration.
+ *  \param[in]  attr          Option type.
+ *  \param[out] buf           Memory address to write option value.
```
ptrendx:
```suggestion
 *  \param[out] buf           Memory address to write option value to.
```

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -264,6 +265,9 @@ NVTEShape nvte_tensor_scale_inv_shape(const NVTETensor tensor);
 void nvte_zero_tensor(const NVTETensor tensor, cudaStream_t stream);
 
 /*! \brief Set a parameter of the tensor.
+ *
+ *  This only supports tensor parameters of type NVTEBasicTensor. Use
+ *  nvte_set_tensor_param_v2 for other parameter types.
```
ptrendx: If we intend to move to the v2 param handling, we should just deprecate this call and note this deprecation here - the fact that it supports only basic tensor is quite evident from the function call.

transformer_engine/common/transpose/cast_transpose.h
```diff
@@ -30,6 +30,7 @@ void quantize_transpose_square_blockwise(const SimpleTensor &input, SimpleTensor
                                          const bool return_transpose, const bool pow_2_scale,
                                          const SimpleTensor &noop_tensor, cudaStream_t stream);
 
+/// TODO Compact format is removed. Replace this enum with bool.
```
ptrendx: Is this todo for this PR still or do you want to do that refactoring in the followup?

transformer_engine/common/common.h
```diff
@@ -133,6 +133,23 @@ struct Tensor {
...
+  /*! Map from NVTETensorParam to parameter sizes */
+  static constexpr size_t attr_sizes[] = {
+      sizeof(NVTEBasicTensor),  // kNVTERowwiseData
+      sizeof(NVTEBasicTensor),  // kNVTEColumnwiseData
+      sizeof(NVTEBasicTensor),  // kNVTEScale
+      sizeof(NVTEBasicTensor),  // kNVTEAmax
+      sizeof(NVTEBasicTensor),  // kNVTERowwiseScaleInv
+      sizeof(NVTEBasicTensor),  // kNVTEColumnwiseScaleInv
+      sizeof(NVTEBasicTensor),  // kNVTEColumnwiseAmax
+      sizeof(bool)              // kNVTEWithGEMMSwizzledScales
```
ptrendx: This is implementation defined so we should not rely on this for our API - please use a type with a defined size for this.

transformer_engine/common/common.h
```diff
@@ -400,7 +416,7 @@ struct QuantizationConfig {
       sizeof(bool),                          // force_pow_2_scales
       sizeof(float),                         // amax_epsilon
       sizeof(NVTETensor),                    // noop_tensor
-      sizeof(Float8BlockScaleTensorFormat),  // float8_block_scale_tensor_format
+      sizeof(Float8BlockScaleTensorFormat),  // (deprecated)
```
ptrendx, line 419: Now I see we already do this sizeof(bool) in the other places - this is very unfortunate, could we change it?

transformer_engine/pytorch/csrc/extensions/swizzle.cpp
```diff
@@ -0,0 +1,397 @@
...
+#include "extensions.h"
+#include "pybind.h"
+#include "util.h"
+
+namespace transformer_engine {
+namespace pytorch {
+
+namespace {
+
+void reset_tensor_data(transformer_engine::TensorWrapper &tensor, bool rowwise, bool columnwise) {
```
ptrendx, line 23: Why do we need that? We cache the tensors anyway, so there should not be overhead of just creating a new TensorWrapper.

transformer_engine/pytorch/csrc/extensions/swizzle.cpp
```diff
@@ -0,0 +1,397 @@
...
+  }
+  if (columnwise) {
+    tensor.set_columnwise_data(nullptr, dtype, shape);
+    tensor.set_columnwise_scale_inv(nullptr, dtype, shape);
+  }
+}
+
+}  // namespace
+
+std::tuple<std::optional<at::Tensor>, std::optional<at::Tensor>> swizzle_scales_for_gemm(
```
ptrendx, line 40: We could just output a new TensorWrapper here with those new scaling factors and then we would not need that reset thing I think?

transformer_engine/pytorch/csrc/common.cpp
```diff
@@ -301,11 +301,13 @@ std::vector<size_t> convertShape(const NVTEShape& shape) {
   return std::vector<size_t>(shape.data, shape.data + shape.ndim);
 }
 
-size_t roundup(const size_t value, const size_t multiple) {
+size_t roundup(size_t value, size_t multiple) {
   assert(multiple > 0);
   return ((value + multiple - 1) / multiple) * multiple;
 }
 
+size_t ceildiv(size_t numer, size_t denom) { return (numer + denom - 1) / denom; }
```
ptrendx, line 309: Don't we have a divup function already?

tests/pytorch/test_float8_blockwise_scaling_exact.py
timmoon10: FP8 block-scaling doesn't require a compact format anymore. Now it's always GEMM-ready.

tests/pytorch/test_fusible_ops.py
timmoon10: Before these changes, we didn't have a test for the SwiGLU kernel with quantized, GEMM-ready output. We test the unquantized SwiGLU kernel here: https://github.com/NVIDIA/TransformerEngine/blob/69636a08171d162a223ae3a01e4b36902b64dba1/tests/pytorch/test_numerics.py#L1633
We test the quantized SwiGLU kernel with non-GEMM-ready output here: https://github.com/NVIDIA/TransformerEngine/blob/69636a08171d162a223ae3a01e4b36902b64dba1/tests/pytorch/test_fusible_ops.py#L1566

It's also a good sanity-check for te.Sequential.

transformer_engine/common/cast/dispatch/quantize.cuh
timmoon10: The way FP8 block-scaling was implemented, the only advantage of the compact format over the GEMM-ready format was support for all-gathers. However, all-gather support only required a simple change to use the swap-first-dims kernel. Instead of propagating this PR's changes throughout the FP8 block-scaling logic, I found it simpler to just remove the compact format entirely.

transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp
timmoon10: Nope. From what I understand, UB support for FP8 block-scaling is missing two things:
- We need logic for handling scales when constructing tensor subviews here: https://github.com/NVIDIA/TransformerEngine/blob/28d08a745dc220862ea06e8919e8eead9864429b/transformer_engine/common/comm_gemm_overlap/comm_gemm_overlap.cpp#L173
- We need to communicate scales separately from the FP8 data here: https://github.com/NVIDIA/TransformerEngine/blob/bd007993dd7403253c6fd9084ebc409f9ea7adad/transformer_engine/pytorch/module/base.py#L466

transformer_engine/pytorch/csrc/common.cpp
timmoon10: I prefer if the PyTorch extensions minimize their reliance on internal headers from the core lib (mainly `common.h`), but I don't have a strong opinion.

transformer_engine/common/cast/mxfp8/gated_mxfp8.cuh
timmoon10: From what I can see, we just allocate scales with `at::empty`:
https://github.com/NVIDIA/TransformerEngine/blob/28d08a745dc220862ea06e8919e8eead9864429b/transformer_engine/pytorch/csrc/quantizer.cpp#L886-L897
Requiring zeroing out in this case is unintuitive, especially since it's not needed in the unpadded case. I figure that if the model is small enough for the padding to be relevant, then it's probably too small to see the full perf benefit of MXFP8 anyways.

transformer_engine/common/transpose/cast_transpose.h
timmoon10: I think that refactoring can go in a followup

transformer_engine/common/common.h
timmoon10: Replaced `bool` with `uint8_t` and `int` with `int32_t`.

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -193,6 +193,7 @@ def copy(self) -> NVFP4Quantizer:
             stochastic_rounding=self.stochastic_rounding,
         )
         quantizer.internal = self.internal
+        quantizer.optimize_for_gemm = self.optimize_for_gemm
```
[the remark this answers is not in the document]
timmoon10: `optimize_for_gemm` is set in the base class: https://github.com/NVIDIA/TransformerEngine/blob/75c299f23ec7c702415d9995a3ac8f8e938a9d72/transformer_engine/pytorch/quantized_tensor.py#L216

transformer_engine/common/cast/dispatch/gated.cuh
```diff
@@ -53,6 +54,20 @@ void quantize_gated_fwd_helper(const NVTETensor nvte_input, NVTETensor nvte_outp
...
+        transpose_in.scaling_mode = NVTE_DELAYED_TENSOR_SCALING;
+        transpose_in.data.dptr = output->data.dptr;
+        transpose_in.data.shape = {output->flat_first_dim(), output->flat_last_dim()};
+        transpose_in.data.dtype = output->data.dtype;
+        transpose_out.scaling_mode = NVTE_DELAYED_TENSOR_SCALING;
+        transpose_out.data.dptr = output->columnwise_data.dptr;
+        transpose_out.data.shape = {output->flat_last_dim(), output->flat_first_dim()};
+        transpose_out.data.dtype = output->data.dtype;
+        detail::transpose(transpose_in, /*noop=*/dummy, &transpose_out, stream);
+      }
```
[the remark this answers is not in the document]
timmoon10: The extra layer of abstraction would hurt maintainability more than just duplicating <20 LoC.

transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh
```diff
@@ -106,7 +110,7 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK)
   const size_t scales_offset_Y_colwise = scales_block_offset_Y_colwise + tid_Y_colwise;
   const size_t scales_offset_X_colwise = scales_block_offset_X_colwise + tid_X_colwise;
 
-  const bool rowwise_scale_is_within_bounds = scales_offset_X_rowwise < cols;
+  const bool rowwise_scale_is_within_bounds = SCALE_DIM_X * scales_offset_X_rowwise < cols;
```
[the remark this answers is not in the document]
timmoon10: Changing the kernel implementation is outside of scope for this PR. The tests already pass, so this is probably already correct, albeit with unclear variable names.

jberchtold-nvidia (dismissed):
LGTM from a TE/JAX perspective, thanks! I can't see if TE/JAX CI passed since a new pipeline was recently launched, but looks like you're running JAX tests as part of your launch so all good

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -298,14 +291,12 @@ void quantize_bwd_helper(const NVTETensor grad, const NVTETensor input, NVTETens
       FP8BlockwiseRowwiseOption rowwise_option = FP8BlockwiseRowwiseOption::NONE;
       FP8BlockwiseColumnwiseOption columnwise_option = FP8BlockwiseColumnwiseOption::NONE;
       if (output_tensor->has_data()) {
-        bool rowwise_compact = (quant_config_cpp.float8_block_scale_tensor_format ==
-                                Float8BlockScaleTensorFormat::COMPACT);
+        const bool rowwise_compact = !output_tensor->with_gemm_swizzled_scales;
         rowwise_option = rowwise_compact ? FP8BlockwiseRowwiseOption::ROWWISE_COMPACT
                                          : FP8BlockwiseRowwiseOption::ROWWISE_GEMM_READY;
```
yaox12: Why not just
```cpp
rowwise_option = FP8BlockwiseRowwiseOption::ROWWISE_GEMM_READY;
```
like above?

transformer_engine/common/cast/mxfp8/gated_mxfp8.cuh
yaox12: When I implemented the [grouped swizzling kernel](https://github.com/NVIDIA/TransformerEngine/pull/2019), I fused the zeroing out into the swizzling kernel.

yaox12: If we required padding M to 128, it's probably the non-padding case, and the zeroing out can be avoided. So I think it's fine.

transformer_engine/pytorch/csrc/common.h
```diff
@@ -120,6 +120,7 @@ class Quantizer {
   bool rowwise_usage = true;
   bool columnwise_usage = true;
   bool internal = false;
+  bool optimize_for_gemm = false;
```
yaox12, line 123: `optimize_for_gemm` is not clear enough for it's real meaning. There're many factors that could be "optimized for gemm", such as shape, alignment, etc. But here we refer to swizzled scaling factors. 

I suggest we just use a more straightforward name.

timmoon10: I prefer leaving it general since this is an abstract base class. Each recipe can then do specific things like alignment, padding, transposing, etc, and not require changes to the quantizer API.

timmoon10: The main weakness I see with the `optimize_for_gemm` design is that it makes the "usages" less clear. `rowwise_usage` and `columnwise_usage` imply that we are using the quantized tensor for particular GEMMs, but `optimize_for_gemm=false` implies that we might actually be using them for something other than GEMM. Maybe a principled approach would be to add new usages like "row-wise GEMM", "row-wise general", etc, but that would significantly increase the scope of this PR.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -654,6 +642,7 @@ std::pair<TensorWrapper, py::object> Float8BlockQuantizer::convert_and_update_te
     py::object tensor) const {
   const DType dtype = tensor.attr("_fp8_dtype").cast<DType>();
   bool is_2D_scaled = tensor.attr("_is_2D_scaled").cast<bool>();
+  const bool with_gemm_swizzled_scales = true;
```
zhongbozhu, line 645: For FP8 blockwise recipe, it's actually hardcoded "optimize for gemm" right? since the quantization kernel will always choose to transpose its scaling factor for cublas requirement.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -798,6 +776,7 @@ std::pair<TensorWrapper, py::object> Float8BlockQuantizer::convert_and_update_te
     const auto scale_inv_colwise_shape = getTensorShape(scale_inv_colwise);
     ret.set_columnwise_scale_inv(scale_inv_colwise_dptr, DType::kFloat32, scale_inv_colwise_shape);
   }
+  ret.set_with_gemm_swizzled_scales(with_gemm_swizzled_scales);
```
zhongbozhu, line 779: Same as above, this config is not really doing anything? Since the quantizer will always transpose scaling factor for cublas? In this context, transpose a scaling factor is essentially a simpler version of "swizzle" for better access pattern.

transformer_engine/pytorch/csrc/quantizer.cpp
timmoon10: That's right, this shouldn't do anything. It is distracting though and should be removed in the future.

PR #2502, opened 2025-12-10
[common] Add support for cuBLASLt GEMM for GroupedTensor

### LINKED ISSUE #2452: GroupedGemm: FP8 per-tensor via cuBLAS
Add support for GroupedGemm with FP8 per-tensor quantization using cuBLAS. Ensure that grouped operations are efficiently batched and fully compatible with device-supplied data buffers.

pggPL:
## Description

Adds `nvte_grouped_gemm` API using cuBLASLt grouped matmul for batched GEMM on tensors with varying shapes. A GPU kernel (`setup_grouped_gemm_kernel`) converts `NVTEGroupedTensor` format (contiguous buffer + offsets) to cuBLAS requirements (pointer arrays + per-matrix M/N/K).

### New API

```c
void nvte_grouped_gemm(
                       int transa, int transb, const NVTETensor alpha, const NVTEGroupedTensor A,
                       const NVTEGroupedTensor B, const NVTETensor beta, const NVTEGroupedTensor C,
                       NVTEGroupedTensor D, NVTETensor workspace_setup, NVTETensor workspace_cublas,
                       NVTEGroupedMatmulConfig config, cudaStream_t stream) 
```

Computes `D = alpha * op(A) @ op(B) + beta * C` for groups of matrices with potentially different shapes.

## Type of change

## Changes

- GPU setup kernel computing pointers/dims from grouped tensor metadata
- FP8 support with scale_inv handling and TN layout selection on Hopper  
- `GroupedGemmSetupWorkspace` struct for cuBLAS workspace layout
- Tests in `test_grouped_gemm.cu` comparing against `nvte_multi_tensor_gemm` (FP8/BF16, various shapes and transpose layouts)

## Checklist:


transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -228,6 +230,49 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
...
+ * - A, B, C (if provided), D must have the same num_tensors
+ * - For each i: D[i] = alpha[i] * op(A[i]) @ op(B[i]) + beta[i] * C[i]
+ * - Shape compatibility: if transa=false, transb=false:
+ *   - A[i]: (M[i], K[i]), B[i]: (K[i], N[i]), D[i]: (M[i], N[i])
+ */
+void nvte_grouped_gemm(int transa, int transb, const NVTETensor alpha, const NVTEGroupedTensor A,
+                       const NVTEGroupedTensor B, const NVTETensor beta, const NVTEGroupedTensor C,
+                       NVTEGroupedTensor D, NVTETensor workspace_setup, NVTETensor workspace_cublas,
+                       NVTEMatmulConfig config, cudaStream_t stream, const int64_t *avg_m,
+                       const int64_t *avg_n, const int64_t *avg_k);
```
timmoon10: The average sizes seem like advanced configs that would be better to leave out of the top-level API. Can we move them inside `NVTEMatmulConfig`?

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -11,6 +11,8 @@
 #ifndef TRANSFORMER_ENGINE_GEMM_H_
 #define TRANSFORMER_ENGINE_GEMM_H_
 
+#include <cstdint>
```
timmoon10: For compatibility with C:
```suggestion
#include <stdint.h>
```
Alternatively, if we hide the average dims inside `NVTEMatmulConfig`, there's no reason to expose `int64_t` in this header.

transformer_engine/common/gemm/cublaslt_gemm.cu
timmoon10, line 1: Nit: If the plain GEMM and grouped GEMM do not share much implementation, it would be better to have separate `.cu` files to make it easier to read.

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -1104,3 +1104,549 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
...
+  const bool has_row = t->has_data();
+  const bool has_col = t->has_columnwise_data();
+  NVTE_CHECK(has_row || has_col,
+             "Grouped GEMM operand is missing both row-wise and column-wise data");
+
+  // Not yet supported in grouped GEMM: block scaling, MXFP8, NVFP4 specialized layouts.
+  const auto sm = t->scaling_mode;
+  NVTE_CHECK(sm != NVTE_BLOCK_SCALING_1D && sm != NVTE_BLOCK_SCALING_2D && !is_mxfp_scaling(sm) &&
+                 !is_nvfp_scaling(sm),
+             "Grouped GEMM does not yet support NVFP4/MXFP8/block scaling operand selection");
```
timmoon10: Nit: It generalizes better if we flip the logic. We don't support most data formats by default, and unquantized and FP8 data are the exceptions:
```suggestion
  NVTE_CHECK(sm == NVTE_DELAYED_TENSOR_SCALING,
             "Grouped GEMM is only supported with unquantized data and tensor-scaled FP8 data, "
             "but got ", to_string(sm), ".");
```

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -1104,3 +1104,549 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
...
+  NVTE_CHECK(inputB->has_data() || inputB->has_columnwise_data(),
+             "Grouped GEMM: B tensor is missing both row-wise and column-wise data");
+}
+
+// Select row-wise vs column-wise storage and adjust transpose flag for grouped GEMM.
+// Mirrors the non-grouped GEMM logic for FP8 layout handling (TN-only on Hopper) and
+// fallback to column-wise data when row-wise is absent.
+struct GroupedOperandSelection {
+  const transformer_engine::GroupedTensor *tensor = nullptr;
+  const char *base = nullptr;
```
timmoon10: This name isn't clear to me. How about:
```suggestion
  const char *dptr = nullptr;
```

transformer_engine/common/include/transformer_engine/gemm.h
pggPL: It was suggestion of @ptrendx - customer may potentially want to use them if they know something more about the shapes. For example if there are multiple tensors of with k dimension D and one tensor with K dimension equal to 1, then it is potentially true that telling cublas that avg dim = D will result in better performance.

transformer_engine/common/gemm/cublaslt_gemm.cu
pggPL: done

transformer_engine/common/gemm/cublaslt_gemm.cu
pggPL: done

transformer_engine/common/include/transformer_engine/gemm.h
pggPL: done

transformer_engine/common/include/transformer_engine/gemm.h
timmoon10: We should definitely expose these options, but it would be better to put them in `NVTEMatmulConfig` rather than the top-level function signature. If you look at `nvte_cublas_gemm`: https://github.com/NVIDIA/TransformerEngine/blob/97a09c29cffa36efd48cd57f765b0cdc02674eaf/transformer_engine/common/include/transformer_engine/gemm.h#L109-L112
You see advanced options like `pre_gelu_out ` and `use_split_accumulator` that are only needed for specialized cases. It's even worse for `use_split_accumulator`, since that's only relevant for FP8 on Hopper and now it's sitting uselessly in the API forever. Compare with the v2 API, which is much closer to the original [GEMM API](https://www.netlib.org/lapack/explore-html/dd/d09/group__gemm_ga8cad871c590600454d22564eff4fed6b.html#ga8cad871c590600454d22564eff4fed6b):
https://github.com/NVIDIA/TransformerEngine/blob/97a09c29cffa36efd48cd57f765b0cdc02674eaf/transformer_engine/common/include/transformer_engine/gemm.h#L131-L133

The idea of `NVTEMatmulConfig` is to hold these advanced, kernel-specific options in a way where we can add or deprecate them easily without breaking API changes.

pggPL: I added analogous `NVTEGroupedMatmulConfig`.

pggPL: It's worth noting that cublas does not support `pre_gelu_out` and `use_split_accumulator` for grouped gemm.

transformer_engine/common/gemm/config.h
```diff
@@ -31,6 +33,29 @@ struct MatmulConfig {
...
+  bool avg_n_set = false;
+  bool avg_k_set = false;
+
+  static constexpr size_t attr_sizes[] = {
+      sizeof(int64_t),  // avg_m
+      sizeof(int64_t),  // avg_n
+      sizeof(int64_t),  // avg_k
+      sizeof(int)       // sm_count
+  };
+};
```
ptrendx:
```suggestion
  static constexpr size_t attr_sizes[] = {
      sizeof(avg_m), 
      sizeof(avg_n),
      sizeof(avg_k),
      sizeof(sm_count)
  };
};
```

transformer_engine/common/gemm/config.h
```diff
@@ -31,6 +33,29 @@ struct MatmulConfig {
...
+  int64_t avg_n = 0;
+  int64_t avg_k = 0;
+
+  // Number of streaming multiprocessors to use in GEMM kernel
+  int sm_count = 0;
+
+  // Track which attributes have been explicitly set
+  bool avg_m_set = false;
+  bool avg_n_set = false;
+  bool avg_k_set = false;
```
ptrendx: We could use std::optional instead?

transformer_engine/common/gemm/cublaslt_grouped_gemm.cuh
```diff
@@ -0,0 +1,17 @@
...
+
+#ifndef TRANSFORMER_ENGINE_COMMON_GEMM_CUBLASLT_GROUPED_GEMM_CUH_
+#define TRANSFORMER_ENGINE_COMMON_GEMM_CUBLASLT_GROUPED_GEMM_CUH_
+
+#include <cublasLt.h>
+#include <cuda_runtime.h>
+#include <transformer_engine/gemm.h>
+
+// nvte_grouped_gemm is declared in transformer_engine/gemm.h
+// This header is for internal use only.
```
ptrendx: why do we need this header then? It does not seem to actually do anything (yet). We can always add it later if we need to.

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -228,6 +295,45 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
...
+ *  \param[in]  alpha            Scale multipliers for A @ B (NVTETensor with num_tensors elements).
+ *  \param[in]  A                Input grouped tensor A.
+ *  \param[in]  B                Input grouped tensor B.
+ *  \param[in]  beta             Scale multipliers for C (NVTETensor with num_tensors elements).
+ *  \param[in]  C                Input grouped tensor C (can be NULL for beta=0).
+ *  \param[out] D                Output grouped tensor D.
+ *  \param[in]  workspace_setup  Workspace tensor for pointer array setup.
+ *  \param[in]  workspace_cublas Workspace tensor for cuBLAS operations.
+ *  \param[in]  config           Additional configuration (can be NULL for defaults).
+ *  \param[in]  stream           CUDA stream for the operation.
```
ptrendx: Let's change the order of the arguments to be more consistent with the GEMM:
 - A
 - transa
 - B
 - transb
 - C
 - D
 - alpha
 - beta
 - rest

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -0,0 +1,617 @@
...
+    const int64_t uniform_last = has_last ? 0 : static_cast<int64_t>(t->get_common_last_dim());
+
+    return {first_ptr, last_ptr,
+            t->tensor_offsets.has_data() ? static_cast<const int64_t *>(t->tensor_offsets.dptr)
+                                         : nullptr,
+            uniform_first, uniform_last};
+  }
+
+  // Create for C tensor (uses D's dimensions, only has offsets)
+  static TensorShapeInfo for_C(const transformer_engine::GroupedTensor *C,
```
ptrendx: Maybe a slightly more descriptive name like:

```suggestion
  static TensorShapeInfo create_shape_info_for_C(const transformer_engine::GroupedTensor *C,
```

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -0,0 +1,617 @@
...
+inline void validate_grouped_gemm_inputs(const transformer_engine::GroupedTensor *inputA,
+                                         const transformer_engine::GroupedTensor *inputB,
+                                         const transformer_engine::GroupedTensor *inputC,
+                                         const transformer_engine::GroupedTensor *outputD,
+                                         const transformer_engine::Tensor *alpha_tensor,
+                                         const transformer_engine::Tensor *beta_tensor) {
+  const size_t num_tensors = inputA->num_tensors;
+  NVTE_CHECK(num_tensors >= 1, "Grouped GEMM: num_tensors must be at least 1");
+  NVTE_CHECK(inputB->num_tensors == num_tensors,
+             "Grouped GEMM: A and B must have the same num_tensors");
```
ptrendx:
```suggestion
             "Grouped GEMM: A and B must have the same number of tensors");
```

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -0,0 +1,617 @@
...
+// Helper routines to keep nvte_grouped_gemm readable
+// -----------------------------------------------------------------------------
+inline void validate_grouped_gemm_inputs(const transformer_engine::GroupedTensor *inputA,
+                                         const transformer_engine::GroupedTensor *inputB,
+                                         const transformer_engine::GroupedTensor *inputC,
+                                         const transformer_engine::GroupedTensor *outputD,
+                                         const transformer_engine::Tensor *alpha_tensor,
+                                         const transformer_engine::Tensor *beta_tensor) {
+  const size_t num_tensors = inputA->num_tensors;
+  NVTE_CHECK(num_tensors >= 1, "Grouped GEMM: num_tensors must be at least 1");
```
ptrendx:
```suggestion
  NVTE_CHECK(num_tensors >= 1, "Grouped GEMM: number of tensors must be at least 1");
```

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
ptrendx: And same in the following checks.

tests/cpp/operator/test_grouped_gemm.cu
```diff
@@ -0,0 +1,521 @@
...
+CudaPtr<T> cuda_alloc(size_t bytes) {
+  void* ptr = nullptr;
+  NVTE_CHECK_CUDA(cudaMalloc(&ptr, bytes));
+  return CudaPtr<T>(static_cast<T*>(ptr));
+}
+
+// Helper owning GPU buffers that back NVTEGroupedTensor.
+// NVTEGroupedTensor does not own memory; data/offsets/scales
+// must be allocated and freed by the test.
+struct GroupedBuffers {
```
ptrendx: I assume that this (or something similar) will be useful in testing the other functionality using Grouped Tensors, so probably should land in test_common?

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -0,0 +1,617 @@
...
+            uniform_first, uniform_last};
+  }
+};
+
+// Helper functions to compute average dimensions from logical_shape for heuristics
+// These are hints for cuBLASLt algorithm selection, don't need to be exact
+inline int64_t compute_avg_first_dim(const transformer_engine::GroupedTensor *t) {
+  // logical_shape[0] is either num_tensors*M (uniform) or sum_of_M (varying first)
+  // In both cases, dividing by num_tensors gives the average
+  return static_cast<int64_t>(t->logical_shape.data[0]) / static_cast<int64_t>(t->num_tensors);
```
zhongbozhu, line 92: what is logical_shape.data[0]? do we have access to this field if we want cuda graph (ie. without D2H copy)

pggPL: `logical_shape` represents total shape of all tensors in groupped tensor. For example we can have grouped tensor of logical shape [10, 10] and tensors of shape [3, 10] and [7, 10] inside it. The logical shape must be constant to use cuda graphs.

tests/cpp/operator/test_grouped_gemm.cu
pggPL: moved to common

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
zhongbozhu: I see, this could be an issue for future cuda graph support of dynamic shapes MOE. Basically, how it works is like for input buffer shape [10, 10], this [10, 10] shape is static, but there will be an on-device tensor that says the splits will be [[3, 10], [3, 10]], and the actual valid shape will be [6, 10]. The excessive shape should be the gemm performance wasted (?) 

That being said, this function is simply for getting a cublas heuristic, so it's nothing fundamental. Can we support injecting static values directly here? Assuming users know their model well, and they can manually set those values and avoid knowing these shapes on host, since that would mean D2H copy, therefore no cuda graph.

pggPL: The only problem may be that the wrong statistics are computed, not any cuda errors or incorrect memory read, am I right?

There is support for custom statistic injection [here.](https://github.com/NVIDIA/TransformerEngine/pull/2502/changes#diff-8065e62e598dd5a0cb1561324ac9a78f268582220bb405d61d4a1234e57ca716R40)

zhongbozhu: wrong statistics can be avoided if we do zero pad, but that's another story, need to check with cublas folks

btw, does it PR have BF16 grouped gemm? is grouped gemm optimization sm100 only for now?

pggPL: it supports bf16 and there are plans to support it on other devices also

transformer_engine/common/gemm/config.h
```diff
@@ -31,6 +34,20 @@ struct MatmulConfig {
...
+  // Average dimension hints for cuBLASLt algorithm selection heuristics.
+  // nullopt means "not set" - compute automatically from tensor shapes.
+  std::optional<int64_t> avg_m;
+  std::optional<int64_t> avg_n;
+  std::optional<int64_t> avg_k;
+
+  // Number of streaming multiprocessors to use in GEMM kernel
+  int sm_count = 0;
+
+  static constexpr size_t attr_sizes[] = {sizeof(avg_m), sizeof(avg_n), sizeof(avg_k),
```
ptrendx: Agree with greptile, after the change to optional this is wrong - you would need to do something like
```cpp
sizeof(decltype(avg_m)::value_type
```

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -0,0 +1,606 @@
...
+  int64_t b_offset =
+      B_meta.offsets ? B_meta.offsets[idx] : (idx * B_meta.uniform_first * B_meta.uniform_last);
+  int64_t c_offset =
+      C_meta.offsets ? C_meta.offsets[idx] : (idx * C_meta.uniform_first * C_meta.uniform_last);
+  int64_t d_offset =
+      D_meta.offsets ? D_meta.offsets[idx] : (idx * D_meta.uniform_first * D_meta.uniform_last);
+
+  // Compute data pointers
+  // Note: const_cast is safe here - cuBLAS requires void** but won't modify A/B/C data
+  A_ptrs[idx] = const_cast<char *>(a_base) + a_offset * a_elem_size;
```
ptrendx: You don't actually need this here - the fact that you got a const pointer to GroupedTensor does not mean that the pointer inside it is const pointer. So you can declare those a_base etc. pointers as regular char* and not have to const_cast them afterwards.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -0,0 +1,608 @@
...
+  B_ptrs[idx] = b_base + b_offset * b_elem_size;
+  C_ptrs[idx] = c_base + c_offset * c_elem_size;
+  D_ptrs[idx] = d_base + d_offset * d_elem_size;
+
+  // Compute M, N, K dimensions from tensor shapes
+  // Input A is stored as {K,M} when !transa, {M,K} when transa
+  // Input B is stored as {N,K} when !transb, {K,N} when transb
+  M[idx] = static_cast<int>(transa ? a_first : a_last);
+  K[idx] = static_cast<int>(transa ? a_last : a_first);
+  N[idx] = static_cast<int>(transb ? b_last : b_first);
```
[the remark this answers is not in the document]
pggPL: ahh, it's bigger problem. it does not result in any bugs, but i think the code needs to be slightly redesigned.

zhongbozhu:
About integration to the linear layers, what is the progress here to make contiguous weight buffers so that the grouped gemm API (I assume it's gonna be called something like **general_contiguous_grouped_gemm**?) now can have A and B both GroupedTensor?

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -0,0 +1,645 @@
...
+
+#include "../common.h"
+#include "../util/cuda_runtime.h"
+#include "../util/handle_manager.h"
+#include "../util/logging.h"
+#include "./config.h"
+
+namespace {
+
+inline void CreateCublasHandle(cublasLtHandle_t *handle) {
```
ptrendx, line 23: I'm fine with merging this PR as is, but this will make it so the handles for grouped gemm and regular gemm will be different (since the holder class will be different), which is a slight inefficiency since I believe they hold some GPU memory. Let's clean that in a follow up PR @pggPL .

ptrendx (approved):

PR #2521, opened 2025-12-16
Enable post-RHT amax estimation with separate amax scale kernel

negvet:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/csrc/extensions/cast.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/cast.cpp
@@ -870,8 +870,8 @@ void split_quantize_nvfp4_impl_with_rht_helper(const TensorWrapper &input,
       }
 
       // Compute pre-RHT amax of split input
-      nvte_compute_amax_with_config(input_list[i].data(), output_list[i].data(), quant_config_list[i],
-                                    stream);
+      nvte_compute_amax_with_config(input_list[i].data(), output_list[i].data(),
+                                    quant_config_list[i], stream);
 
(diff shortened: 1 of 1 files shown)


zhongbozhu:
Having a separate amax scale API automatically means that we need a grouped API for MOE and therefore not a good choice moving forward. Can we fuse the amax scaling as part of the FP4 quantize kernels via the quantization config struct? Whenever we change the numeric for Dense, we have to change it for MOE.

negvet:
> Having a separate amax scale API automatically means that we need a grouped API for MOE and therefore not a good choice moving forward. Can we fuse the amax scaling as part of the FP4 quantize kernels via the quantization config struct? Whenever we change the numeric for Dense, we have to change it for MOE.

`nvte_scale_amax` is intentionally a standalone kernel. This helps to avoid amax scale through every kernel that deals with amax (quantize, gemm, etc).

This is functional with commont entry point:

- dense: call `nvte_scale_amax`
- moe: call the same, per-split

This is single explicit step that applies everywhere and reduce the chance we forget to scale in the new op/refactor.

I agree that extra launch cost in the loop is an issue (now, `nvte_compute_amax_with_config` is done per split). This is your main concern right?

What about compute all per-split pre‑RHT amaxes in one grouped call instead of calling `nvte_compute_amax_with_config` in a loop? Can I reuse `nvte_group_amax`?
And then we can decide whether grouped scaling is worth to be grouped as well.
What do you think?

PR #2523, opened 2025-12-16
[JAX] Fix incorrect calculation of segment pos from segment ids in user-facing API

KshitijLakhani:
# Description

SequenceDescriptor's `from_segment_ids_and_pos()` accepts the `segment_ids` and an optional `segment_pos ` as input. This class is supposed to serve as a convenience method to do two things: 
1. Stuff the `segment_ids` and `segment_pos` in a SequenceDescriptor object for TE to use downstream
2. If the `segment_pos` is not passed, then calculate/extrapolate it

In it's current form, the second functionality gives incorrect results for THD + non-reordered and THD + reordered cases as it merely uses an `arange` to calculate the `segment_pos` naively. This could result in incorrect masking for these cases.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

This PR makes few changes: 

1. Passed two new args to this function `from_segment_ids_and_pos()` : `is_thd` and `is_segment_ids_reordered`- the only cases that this function can currently guarantee to support is BSHD with and without load balancing and, THD without load balancing. 
2. BSHD with load balancing is supported natively because the segment_ids and segment_pos are not reordered before passing onto `from_segment_ids_and_pos()`.  However, if the `segment_pos` are reordered and passed to `from_segment_ids_and_pos()` it will assert
3. If THD + reordered use cases calls the function `from_segment_ids_and_pos()`, it will assert
4. The fused attn tests were modified to account for these two new args
    * For THD fused attn non-CP tests, segment_pos=None is passed so as to exercise the newly added THD path in `from_segment_ids_and_pos()` 
    * For THD  fused attn CP tests, segment_pos is explicitly passed (as before, not a new change)
    * For BSHD fused attn CP tests, segment_pos=None is passed so as to exercise the default BSHD path to generate segment_pos (as before, not a new change)
  

Impact on user of the API: 
1. These two new args, `is_thd` and `is_segment_ids_reordered` are not Optional and hence they will cause a `TypeError` for current users of this API - a breaking change. However, this is needed to ensure correct usage of this API
2. The user is now expected to let the API know whether this is a THD or BSHD layout and whether the `segment_ids` are reordered or not. It is expected that the `segment_ids` passed will be reordered only for THD load balancing. For all other cases the `segment_ids` should not be reordered

# Checklist:

### DIFF
--- a/qa/L0_jax_distributed_unittest/test.sh
+++ b/qa/L0_jax_distributed_unittest/test.sh
@@ -1,6 +1,7 @@
 # Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # See LICENSE for license information.
+export TRITON_PTXAS_PATH=/usr/local/cuda/bin/ptxas
 
 function error_exit() {
     echo "Error: $1"
--- a/qa/L0_jax_unittest/test.sh
+++ b/qa/L0_jax_unittest/test.sh
@@ -1,6 +1,7 @@
 # Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # See LICENSE for license information.
+export TRITON_PTXAS_PATH=/usr/local/cuda/bin/ptxas
 
 set -x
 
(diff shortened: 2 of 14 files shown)


KshitijLakhani:
@ksivaman this PR has a breaking change (FYI for TE release notes)

KshitijLakhani:
@mgoldfarb-nvidia @huanghua1994 @mingxu1067 - even though this PR has been merged, it would be great to get your feedback on this, in case you spot something that needs to be reverted. Thanks !

PR #2539, opened 2025-12-22
Fix ptxas compilation on sm103 for triton kernels

tdophung:
Compilation failures happen because the system does not know where ptxas on *B300 systems. To resolve this, the user just needs to point it to the right place. This change is to do this automatically in QA cycles and CI (if CI were to run on GB110)

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

add env var to jax unittest scripts

# Checklist:

### DIFF
--- a/qa/L1_jax_distributed_unittest/test.sh
+++ b/qa/L1_jax_distributed_unittest/test.sh
@@ -1,6 +1,7 @@
 # Copyright (c) 2022-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
 # See LICENSE for license information.
+export TRITON_PTXAS_PATH=/usr/local/cuda/bin/ptxas
 
 function test_fail() {
     RET=1


jberchtold-nvidia:
@tdophung SR: can you also add this to `L1_jax_distributed_unittest`? We don't currently run Triton tests in L1 since L1 only has a multi-GPU config, but we will likely start when we add custom partitioning support to these triton wrappers.

KshitijLakhani:
> @tdophung SR: can you also add this to `L1_jax_distributed_unittest`? We don't currently run Triton tests in L1 since L1 only has a multi-GPU config, but we will likely start when we add custom partitioning support to these triton wrappers.

@tdophung when you do this please add the time it takes for your additional tests in the PR description using `NVTE_JAX_TEST_TIMING=1` so that we can track how much time we are adding to our tests>
Also, we'd like to report this for any tests we add in the future to TE JAX as well
Thanks !

tdophung:
I added the flag to L1 files, I will also put the test time increase down once it is enabled for L1

PR #2540, opened 2025-12-23
Solve pytorch-triton and triton package contention

tdophung:
# Description
`pytorch-triton` and `triton` packages install to the same location at site-packages/triton, and `triton` does not work for pytorch's torch.compile() call as there are a few things pytorch has added onto their version of triton (creating `pytorch-triton` to make it work and validated it with the release of torch). However `pytorch-triton` should in theory (and experimented) still be compatible with how jax uses it*.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes
 - Add new env var to control when to use pytorch-triton in jax
 - switch pytorch back to using/checking for pytorch-triton by default
- Add documentation (comments) on this contention of packages 

# Checklist:

### DIFF
--- a/transformer_engine/jax/triton_extensions/utils.py
+++ b/transformer_engine/jax/triton_extensions/utils.py
@@ -122,8 +122,7 @@ def _check_triton_compatibility():
             "If you don't need Triton, use transformer_engine.jax.cpp_extensions instead."
         )
 
-    use_pytorch_triton_env = os.environ.get("NVTE_USE_PYTORCH_TRITON", "0").lower()
-    use_pytorch_triton_explicit = use_pytorch_triton_env in ("1", "true", "yes")
+    use_pytorch_triton_explicit = bool(int(os.environ.get("NVTE_USE_PYTORCH_TRITON", "0")))
 
     if is_pytorch_triton:
         if use_pytorch_triton_explicit:
@@ -186,8 +185,7 @@ def get_triton_info():
         if info['is_pytorch_triton']:
              print("Using pytorch-triton - compatible with both PyTorch and JAX")
     """
-    use_pytorch_triton_env = os.environ.get("NVTE_USE_PYTORCH_TRITON", "0").lower()
-    env_acknowledged = use_pytorch_triton_env in ("1", "true", "yes")
+    env_acknowledged = bool(int(os.environ.get("NVTE_USE_PYTORCH_TRITON", "0")))
 
     return {
         "version": _TRITON_VERSION,


transformer_engine/jax/triton_extensions/utils.py
```diff
@@ -142,17 +304,31 @@ def compile_triton(
...
+    from packaging import version
+
+    if version.parse(jax.__version__) >= version.parse("0.8.2"):
+        kernel = gpu_triton.TritonKernel(
+            compiled.name,  # arg0: kernel_name (str)
+            num_warps,  # arg1: num_warps (int)
+            num_ctas,  # arg2: num_ctas (int)
+            compiled.metadata.shared,  # arg3: shared_mem_bytes (int)
+            compiled.asm["ptx"],  # arg4: ptx (str)
+            "",  # arg5: ttir (str) - empty
```
tdophung, line 315: This will soon be the same as main. as this change is made here in: https://github.com/NVIDIA/TransformerEngine/pull/1921, to be merged. it is just in this PR so I can test triton calls locally with the nitghtly jax container without running into errors because of jax 0.8.2+

build_tools/jax.py
```diff
@@ -19,8 +19,33 @@ def install_requirements() -> List[str]:
...
+            Returns 'pytorch-triton' - for mixed JAX+PyTorch environments.
+            Install with: pip install pytorch-triton --index-url https://download.pytorch.org/whl/cu121
+
+            Note: Do NOT install pytorch-triton from PyPI directly - that's a placeholder.
+    """
+    use_pytorch_triton = os.environ.get("NVTE_USE_PYTORCH_TRITON", "0").lower() in (
+        "1",
+        "true",
+        "yes",
+    )
```
ksivaman:
```suggestion
    use_pytorch_triton = bool(int(os.environ.get("NVTE_USE_PYTORCH_TRITON", "0")))
```

ksivaman: It's better to be specific with envvars and follow convention from rest of the codebase.

build_tools/pytorch.py
```diff
@@ -22,7 +32,7 @@ def install_requirements() -> List[str]:
         "packaging",
         "pydantic",
         "nvdlfw-inspect",
-        "triton",
+        "pytorch-triton",
```
ksivaman, line 35: If `pytorch-triton` from PyPI is actually a placeholder, then we shouldn't list it here as a dependency

tdophung: `pytorch-triton` should be the default for transformerengine-pytorch. Not just a placeholder. It should be used, all the time when pytorch framework is used.

`triton` is the default for jax, unless either in 2 scenarios happen:
- there is both jax and pytorch installed, and they are using TE pytorch to call the triton kernels
- The user specify NVTE_USE_PYTORCH_TRITON=1 while using TE jax, to make sure there is no performance diff between using different versions of triton, between pytorch and jax.

xwang233:
`pytorch-triton` from pypi is breaking things, the package is a psuedo placeholder , it unzipped a `triton/__init__.py` that overwrites the actual `triton` package file and cause failures

```
root@8664b557fc9f:/workspace# unzip pytorch_triton-0.0.1-py3-none-any.whl
Archive:  pytorch_triton-0.0.1-py3-none-any.whl
  inflating: triton/__init__.py
  inflating: pytorch_triton-0.0.1.dist-info/METADATA
  inflating: pytorch_triton-0.0.1.dist-info/WHEEL
  inflating: pytorch_triton-0.0.1.dist-info/top_level.txt
  inflating: pytorch_triton-0.0.1.dist-info/RECORD
root@8664b557fc9f:/workspace# cat triton/__init__.py
raise RuntimeError("Should never be installed")
```

cc @nweidia

build_tools/pytorch.py
tdophung: after the break I now understand what you mean here

PR #2544, opened 2025-12-26
[PyTorch]Add Casting-Free FP8-Flow-MoE Blockwise Optimizations

xiaoxi-wangfj:
# Description

This PR introduces **blockwise, scaling-aware FP8 transpose optimizations** for FP8 MoE that enable a **casting-free, FP8-centric MoE dataflow** in TransformerEngine by eliminating unnecessary cast and re-quantization steps, while maintaining numerical stability in existing FP8 training workflows.

This PR is designed to be used in conjunction with PR https://github.com/NVIDIA/Megatron-LM/pull/2764

Further optimizations are introduced via two additional PRs:
- https://github.com/NVIDIA/Megatron-LM/pull/2763
- https://github.com/NVIDIA/TransformerEngine/pull/1921

### Background / Motivation

The design and theoretical background of this PR are described in the paper:  
[**FP8-Flow-MoE: A Casting-Free FP8 Recipe without Double Quantization Error**](https://arxiv.org/abs/2511.02302)

The follow figure illustrates the optimized MoE dataflow and highlights the key optimization points (marked as ①–⑤).

FP8FLOW-MoE

#### 1. FP8 Quantization Before Dispatch (DeepEP → GroupedLinear)

Quantization is performed **before DeepEP dispatch**, and **row-wise FP8 tensors are directly fed into GroupedLinear**.

- Keeps `dispatch → permute → expert computation` entirely in FP8
- `Float8BlockwiseQTensor` is propagated with a COMPACT layout (for `_rowwise_scale_inv`) along the `dispatch → permute → GroupedLinear` path, avoiding layout-induced `.T.contiguous()` calls and reducing unnecessary memory copies.

(Shown as **marker ①** in the figure)

#### 2. Scaling-Aware FP8 Transpose for Wgrad

GroupedLinear requires:
- **row-wise FP8** for Fprop/Dgrad
- **column-wise FP8** for Wgrad

To avoid `dequantize → transpose → requantize` , this PR introduces **`scaling_aware_fp8_transpose`**, which:

- Converts row-wise FP8 to column-wise FP8 via exponent manipulation only
- Preserves scale consistency across layouts
- reduce cpu overhead

(Shown as **marker ④** in the figure)

#### 3. Fused Permute + Padding / Unpermute + Unpadding

We fuse two memory movement operators along the MoE path：

- `permute + pad` in the forward pass  
- `unpermute + unpad` in the backward pass  

For details of this optimization, please refer to PR https://github.com/NVIDIA/TransformerEngine/pull/1921

(Shown as **marker ②** in the figure)

#### 4. Fused Activation + Quantization

Activation and FP8 quantization are fused into a single kernel, Produces FP8 outputs directly, while enabling FP8 persistence 

(Shown as **marker ③** in the figure)

#### 5. Add fine-grained recompute `moe_expert`

Because the entire `dispatch → permute → GroupedLinear` path stays in FP8, we enable **fine-grained recomputation at the `moe_expert` level**:

- Saves ~50% peak activation memory and avoids recomputation of the router compared to recomputing the full module `moe` level

(Shown as **marker ⑤** in the figure)

### Performance Results

We evaluate FP8-Flow-MoE on **DeepSeek-V3 (671B)** to validate scalability and robustness under realistic large-scale training conditions.

#### Throughput

Measured throughput (TGS, tokens/GPU/s) under different expert parallelism (EP) on **DeepSeek-V3 (671B)** :

- **vs. BF16**  
  +6% (EP8), +8% (EP16), +16% (EP32)

- **vs. TransformerEngine blockwise FP8 recipe**  
  +3% (EP8), +8% (EP16), up to **+21% (EP32)**

#### Memory Efficiency

With **AC = selective checkpointing** and **recompute-modules = moe_expert**:

- At EP8:
  - ~8 GB lower peak memory vs. BF16
  - ~16.5 GB lower peak memory vs. blockwise FP8

#### Numerical Accuracy

We trained for **>200B tokens**. The loss deviation of FP8-Flow-MoE stays within **0.19%** compared to both BF16 baselines and TransformerEngine blockwise FP8 recipe, with no observed instability or divergence.

### Limitations

- Currently validated on **NVIDIA Hopper architecture** with **blockwise FP8 recipe**

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- **Megatron-LM**: Added fused FP8 kernels for activation + quantization in `fused_bias_swiglu.py` and `fused_weighted_swiglu_quant.py`
- **Megatron-LM**: Integrated FP8 dispatch and expert recomputation support in Megatron-LM `fused_a2a.py`
- **TransformerEngine**: Added support for `Float8BlockwiseQTensor` inputs in `grouped_linear.py`
- **TransformerEngine**: Added `scaling_aware_fp8_transpose` operator in `triton/blockwise_scaling_aware_fp8_transpose.py` 

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/grouped_linear.py
+++ b/transformer_engine/pytorch/module/grouped_linear.py
@@ -44,6 +44,7 @@ from ..jit import no_torch_dynamo
 from ..cpu_offload import is_cpu_offload_enabled, mark_not_offload, start_offload
 
 from ..tensor.float8_tensor import Float8CurrentScalingQuantizer, Float8Quantizer
+from ..tensor.float8_blockwise_tensor import Float8BlockwiseQTensor
 from ..quantized_tensor import (
     QuantizedTensorStorage,
     Quantizer,
@@ -143,7 +144,12 @@ class _GroupedLinear(torch.autograd.Function):
         inp_view = inp.reshape(-1, in_features)
         inputmats: list
         if fp8 and not debug:
-            inputmats = tex.split_quantize(inp_view, m_splits, input_quantizers)
+            if isinstance(inp_view, Float8BlockwiseQTensor):
+                inputmats = inp_view.split_scaling_aware_fp8_transpose(
+                    m_splits, input_quantizers
+                )
+            else:
+                inputmats = tex.split_quantize(inp_view, m_splits, input_quantizers)
         elif debug:
             inputmats = DebugQuantizer.multi_tensor_quantize(
                 inp_view, input_quantizers, m_splits, activation_dtype
@@ -343,18 +349,28 @@ class _GroupedLinear(torch.autograd.Function):
                         # Unfused bias grad and multi-tensor quantize
                         for i in range(ctx.num_gemms):
                             grad_biases[i] = grad_output_mats[i].sum(dim=0)
+                        if isinstance(grad_output_view, Float8BlockwiseQTensor):
+                            grad_output = grad_output_view.split_scaling_aware_fp8_transpose(
+                                ctx.m_splits, ctx.grad_output_quantizers
+                            )
+                        else:
+                            grad_output = tex.split_quantize(
+                                grad_output_view,
+                                ctx.m_splits,
+                                ctx.grad_output_quantizers,
+                            )
+                else:
+                    # Multi-tensor quantize
+                    if isinstance(grad_output_view, Float8BlockwiseQTensor):
+                        grad_output = grad_output_view.split_scaling_aware_fp8_transpose(
+                            ctx.m_splits, ctx.grad_output_quantizers
+                        )
+                    else:
                         grad_output = tex.split_quantize(
                             grad_output_view,
                             ctx.m_splits,
                             ctx.grad_output_quantizers,
                         )
-                else:
-                    # Multi-tensor quantize
-                    grad_output = tex.split_quantize(
-                        grad_output_view,
-                        ctx.m_splits,
-                        ctx.grad_output_quantizers,
-                    )
             elif ctx.debug:
                 grad_output_mats = torch.split(grad_output_view, ctx.m_splits)
                 for i in range(ctx.num_gemms):
@@ -781,9 +797,10 @@ class GroupedLinear(TransformerEngineBaseModule):
         """
         debug = self.is_debug_iter()
 
-        assert not isinstance(
-            inp, QuantizedTensorStorage
-        ), "GroupedLinear doesn't support input tensor in FP8."
+        if not isinstance(inp, Float8BlockwiseQTensor):
+            assert not isinstance(
+                inp, QuantizedTensorStorage
+            ), "GroupedLinear doesn't support input tensor in FP8."
         assert len(m_splits) == self.num_gemms, "Number of splits should match number of GEMMs."
 
         is_grad_enabled = torch.is_grad_enabled()
--- a/transformer_engine/pytorch/permutation.py
+++ b/transformer_engine/pytorch/permutation.py
@@ -224,7 +224,10 @@ class _moe_permute_mask_map(torch.autograd.Function):
             fake_dtype = inp.dtype
             # blockwise scaling
             if blockwise_recipe:
-                fp8_scale = inp._rowwise_scale_inv.T.contiguous()
+                if inp._rowwise_data.shape[0] == inp._rowwise_scale_inv.shape[0]:
+                    fp8_scale = inp._rowwise_scale_inv
+                else:
+                    fp8_scale = inp._rowwise_scale_inv.T.contiguous()
                 scale_hidden_dim = fp8_scale.shape[1]
                 assert num_tokens == fp8_scale.shape[0], "scale and input shape mismatch"
                 inp = inp._rowwise_data
@@ -275,7 +278,7 @@ class _moe_permute_mask_map(torch.autograd.Function):
                     shape=output.shape,
                     dtype=fake_dtype,
                     rowwise_data=output,
-                    rowwise_scale_inv=permuted_scale.T.contiguous(),
+                    rowwise_scale_inv=permuted_scale,
                     columnwise_data=None,
                     columnwise_scale_inv=None,
                     fp8_dtype=fp8_dtype,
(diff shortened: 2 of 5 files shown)


rich-junwang:
Quick question, does this work with mxfp8 or it only applies to fp8? Thanks.

xiaoxi-wangfj:
> Quick question, does this work with mxfp8 or it only applies to fp8? Thanks.

@rich-junwang , Thanks for asking. This PR primarily targets the standard FP8 blockwise recipe, and the current scaling-aware FP8 transpose implementation is specialized for blockwise=128 scaling.

Among the five optimizations described in this PR:
Items 1, 2, and 4 are specific to the FP8 blockwise recipe.
Items 3 and 5 are not tied to a specific FP8 recipe and can be applied more generally, independent of whether blockwise scaling is used.

zhongbozhu:
Similar ideas should apply to MXFP8 right? Except that MXFP8 doesn't need any explicit transpose in the first place, but still needs two directional quantization & naturally E8M0 scales. Thanks for sharing this technique though!

rich-junwang:
With this method, mxfp8 doesn't have to quantize for both rowwise and columnwise right? It would be easier for communication and save some memory.

xiaoxi-wangfj:
> Similar ideas should apply to MXFP8 right? Except that MXFP8 doesn't need any explicit transpose in the first place, but still needs two directional quantization & naturally E8M0 scales. Thanks for sharing this technique though!

@zhongbozhu 
The feasibility of the FP8 transpose operator depends on E8M0 scales. MXFP8 uses UE8M0 to record the scale, which naturally satisfies this requirement. In addition, compared with blockwise quantization, MXFP8 has finer quantization granularity, making it numerically safer; as a result, underflow is less likely to occur during the transpose operation.
The transpose algorithm does not provide additional precision benefits for MXFP8, but it is usable, and our fused implementation is expected to deliver performance gains. However, we do not currently have suitable hardware to validate this. We are happy to provide support on the algorithmic principles and code-level details to facilitate its implementation on MXFP8.

xiaoxi-wangfj:
> With this method, mxfp8 doesn't have to quantize for both rowwise and columnwise right? It would be easier for communication and save some memory.
@rich-junwang 
Thank you for your attention. Yes, in this method and in our implementation, MXFP8 does not need to quantize both rowwise and columnwise. Since MXFP8 uses UE8M0 to record scales, it naturally fits the FP8 transpose algorithm assumptions, allowing us to avoid maintaining dual quantization directions. In our MoE FP8 path, together with operator fusion and the FP8 transpose implementation, this reduces communication overhead and saves memory.

rich-junwang:
I'm interested in mxfp8 implementation. let me know if any help is needed here.

ptrendx:
Hi @xiaoxi-wangfj, thank you for your contribution :-).

I read the paper and have a few questions. The math for the zero error transpose works only in the case of no underflow/overflow, but there is no discussion in the paper on how many underflows happen in the real training. What is more, the strategy for choosing the transposed scaling factor (as the maximum of the rowwise scaling factors) makes the underflows more likely. Do you have some statistics on the typical underflow numbers and how that scaling factor choice impacts it? Is there a performance reason why you chose to take the maximum scaling factor over computing it properly - the problem generally should be pretty bandwidth bound so the overhead of computing the right scaling factor should be negligible I think?
I think that considering this, since it is not numerically exactly equivalent, to be conservative we should expose this setting behind an option (maybe in the recipe?) rather than the default. We should also then add the support of getting the number of mismatches compared to the default recipe to the debug tools (but that can happen in a separate PR). 
What do you think?

xiaoxi-wangfj:
> Hi @xiaoxi-wangfj, thank you for your contribution :-).
> 
> I read the paper and have a few questions. The math for the zero error transpose works only in the case of no underflow/overflow, but there is no discussion in the paper on how many underflows happen in the real training. What is more, the strategy for choosing the transposed scaling factor (as the maximum of the rowwise scaling factors) makes the underflows more likely. Do you have some statistics on the typical underflow numbers and how that scaling factor choice impacts it? Is there a performance reason why you chose to take the maximum scaling factor over computing it properly - the problem generally should be pretty bandwidth bound so the overhead of computing the right scaling factor should be negligible I think? I think that considering this, since it is not numerically exactly equivalent, to be conservative we should expose this setting behind an option (maybe in the recipe?) rather than the default. We should also then add the support of getting the number of mismatches compared to the default recipe to the debug tools (but that can happen in a separate PR). What do you think?

Hi @ptrendx sorry for the late reply, and thank you for the very thorough review! 

**1. On the scaling choice (numerical impact of the current approach)**
We collected underflow statistics from real training runs for the scale-aware transpose (columnwise) and evaluated the impact from two perspectives: element_ratio and energy_ratio.
Underflow element_ratio: the number of elements that underflow (FTZ) divided by the total number of elements (i.e., the fraction of 1’s in the underflow mask).
Underflow energy_ratio:  the energy fraction of the underflowed elements in the BF16 reference tensor:
With the current approach, we observed:
Forward activations: underflow element_ratio ≈ 0.002%, underflow energy_ratio ≈ 2.4e-12
Backward gradout: underflow element_ratio ≈ 2%, underflow energy_ratio ≈ 4.0e-08

These results indicate that underflow occurs only on the wgrad path and is dominated by very small-magnitude values that contribute negligibly to the overall gradient energy. With data-parallel gradient averaging and Adam’s momentum and variance normalization, such sparse, low-energy perturbations are strongly attenuated in practice, which is consistent with our observed stable loss curves and gradient-norm behavior. Therefore, considering both numerical behavior and performance, we believe the current approach represents a reasonable and acceptable design choice.

We also implemented an optional variant that adds an additional per-column scaling selection on top of the original scaling-aware transpose (e.g., a per-column “max-no-overflow” strategy), which further reduces the wgrad underflow energy_ratio to 1.2e−09 (with element_ratio ≈ 0.2%). In our experiments, this variant shows similar end-to-end convergence stability compared to the 4.0e−08 version. However, achieving this tighter scaling incurs a non-negligible computational overhead. Given the limited numerical benefit and the measurable performance cost, we do not enable this variant by default. We plan to continue optimizing its kernel implementation and expose this strategy as an optional setting.

**2. Default knob and debug**
We agree that, to be more conservative, this behavior could be gated behind a recipe/option rather than being the default. Just to confirm: would you prefer this knob to primarily control whether we take the scale-aware transpose path (vs. a dequant → split_quantize reference path)?
We can also integrate underflow/mismatch counters into the debug tools (I can submit a separate PR for that).

Thanks again for your feedback!

ptrendx:
Hi @xiaoxi-wangfj , sorry, did not see that there was a question to me in your message. Thank you for the explanation. On the knob - ideally we would have a knob in the recipe whether to enable this optimization, yes. So basically whenever you have the if statement whether to choose the split_quantize or the scaling aware split quantize, please check if the knob in the recipe is set to enable this.

PR #2555, opened 2026-01-03
[NVFP4][Dense/MoE] Integrate Cutlass NVFP4 Row-Cast-Col-RHT-Transpose-Cast Fusion Kernel

zhongbozhu:
# Description

Note: https://github.com/NVIDIA/TransformerEngine/issues/2558 reported a bug in https://github.com/NVIDIA/TransformerEngine/pull/2411. Fix is here https://github.com/NVIDIA/TransformerEngine/pull/2564: make sure you cherry-pick this one too before it's in main. 

Previously, similar optimization has been applied for MOE grouped quantize with RHT in https://github.com/NVIDIA/TransformerEngine/pull/2411. This PR targets the dense linear layers & shared experts when being quantized to NVFP4. Having this fusion means high precision input only needs to be read once while without this fusion, it needs to be read twice. 

Similarly, we have env var NVTE_USE_FAST_MATH to control the numerical behavior of RHT quant fusion kernel to accelerate it further. The fast math is only applied to the high precision math so it will have minimal impact of the training convergence.

What fast-math toggle controls: 
1. replace x / y by x * (1/y) ==> UPDATE: this is now DEFAULT to all NVFP4 kernels
2. replace 1 / x by reciporal_approximate_ftz(x)
3. when RHT cast fusion is available, fusion allows nvfp4 quantize to be performed directly on FP32 data in register files, this will essentially remove a round trip between FP32 to BF16 then FP32. 

**Therefore, I DO recommend turn it on since it will significantly improve the RHT kernel performance.** 

The only reason why it's still not default open is because we want ZERO TOLERNACE test between our CUDA quantize kernels and our pytorch-based emulated quantize references. With fast math toggle turned on, it's hard to pass test with zero tolerance without further investigation of how to relax the test conditions while still providing high confidence of the test case. 

TODO items:
- [x] Merge the bug fix PR 2564 first. 
- [x] Some cutlass deprecating APIs are being used, output many warnings. 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
+++ b/transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
@@ -144,10 +144,7 @@ rht_gemm_device(MShape M, NShape N, KShape K, ClusterTileShape cluster_tile,
   using namespace cute;
   constexpr bool is_blackwell_arch = ARCH_BLACKWELL_FAMILY;
   if constexpr (!is_blackwell_arch) {
-    NVTE_DEVICE_ERROR(
-        "rht_gemm_device is only supported on Blackwell "
-        "with architecture-specific compilation. "
-        "Try recompiling with sm_100a or similar.");
+    NVTE_DEVICE_ERROR("RHT fusion is only supported on Blackwell.");
     return;
   } else {
   using X = Underscore;
--- a/transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
+++ b/transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
@@ -190,10 +190,7 @@ __global__ static void row_col_rht_gemm_device(
   // Abort immediately if compilation is not supported
   constexpr bool is_blackwell_arch = ARCH_BLACKWELL_FAMILY;
   if constexpr (!is_blackwell_arch) {
-    NVTE_DEVICE_ERROR(
-        "row_col_rht_gemm_device is only supported on Blackwell "
-        "with architecture-specific compilation. "
-        "Try recompiling with sm_100a or similar.");
+    NVTE_DEVICE_ERROR("RHT fusion is only supported on Blackwell.");
     return;
   } else {
     static_assert(kEnableRHTColQuant_ || kEnableRowQuant_,
--- a/transformer_engine/common/hadamard_transform/graph_safe_group_row_cast_col_hadamard_transform_cast_fusion.cu
+++ b/transformer_engine/common/hadamard_transform/graph_safe_group_row_cast_col_hadamard_transform_cast_fusion.cu
@@ -193,10 +193,7 @@ __launch_bounds__(512, 1) __global__ static void group_row_col_rht_gemm_device_g
   // Abort immediately if compilation is not supported
   constexpr bool is_blackwell_arch = ARCH_BLACKWELL_FAMILY;
   if constexpr (!is_blackwell_arch) {
-    NVTE_DEVICE_ERROR(
-        "group_row_col_rht_gemm_device_graph_safe is only supported on Blackwell "
-        "with architecture-specific compilation. "
-        "Try recompiling with sm_100a or similar.");
+    NVTE_DEVICE_ERROR("RHT fusion is only supported on Blackwell.");
     return;
   } else {
     static_assert(kEnableRHTColQuant_ || kEnableRowQuant_,
--- a/transformer_engine/common/hadamard_transform/group_hadamard_transform_cast_fusion.cu
+++ b/transformer_engine/common/hadamard_transform/group_hadamard_transform_cast_fusion.cu
@@ -173,10 +173,7 @@ __global__ static void group_rht_gemm_device(
   using namespace cute;
   constexpr bool is_blackwell_arch = ARCH_BLACKWELL_FAMILY;
   if constexpr (!is_blackwell_arch) {
-    NVTE_DEVICE_ERROR(
-        "group_rht_gemm_device is only supported on Blackwell "
-        "with architecture-specific compilation. "
-        "Try recompiling with sm_100a or similar.");
+    NVTE_DEVICE_ERROR("RHT fusion is only supported on Blackwell.");
     return;
   } else {
     using X = Underscore;
--- a/transformer_engine/common/hadamard_transform/group_row_cast_col_hadamard_transform_cast_fusion.cu
+++ b/transformer_engine/common/hadamard_transform/group_row_cast_col_hadamard_transform_cast_fusion.cu
@@ -185,10 +185,7 @@ __launch_bounds__(512, 1) __global__ static void group_row_col_rht_gemm_device(
   // Abort immediately if compilation is not supported
   constexpr bool is_blackwell_arch = ARCH_BLACKWELL_FAMILY;
   if constexpr (!is_blackwell_arch) {
-    NVTE_DEVICE_ERROR(
-        "group_row_col_rht_gemm_device is only supported on Blackwell "
-        "with architecture-specific compilation. "
-        "Try recompiling with sm_100a or similar.");
+    NVTE_DEVICE_ERROR("RHT fusion is only supported on Blackwell.");
     return;
   } else {
     static_assert(kEnableRHTColQuant_ || kEnableRowQuant_,


transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1468,14 +1545,26 @@ void NVFP4Quantizer::quantize_impl(const TensorWrapper& input, TensorWrapper& ou
...
+  // Restriction for the RHT cast fusion kernel because we are using MMA hardware for computing RHT
+  bool eligible_for_rht_cast_fusion =
+      input.dtype() == DType::kBFloat16 && rows % 64 == 0 && cols % 128 == 0;
+
   // Stochastic rounding
   // When both rowwise and columnwise quantization are used with RHT,
   // we need separate RNG states for each to ensure they use different random numbers.
   TensorWrapper te_rng_state;
   TensorWrapper te_rng_state_columnwise;
   QuantizationConfigWrapper quant_config_columnwise;
```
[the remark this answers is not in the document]
zhongbozhu: It's a case that won't be triggered in real training because 2D quantization & noop is for weight, where we don't do stochastic rounding, but adding checks here is harmless too.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1468,14 +1547,25 @@ void NVFP4Quantizer::quantize_impl(const TensorWrapper& input, TensorWrapper& ou
...
+
+  // Only need a separate rng state when:
+  // 1. Stochastic rounding is enabled
+  // 2. RHT is enabled
+  // 3. Columnwise usage is enabled
+  // 4. Rowwise and columnwise quantization are not fused,
+  //    because within a single kernel we can generate two different random numbers for rowwise and columnwise
+  const bool need_separate_columnwise_rng = this->stochastic_rounding && this->with_rht &&
+                                            this->columnwise_usage &&
+                                            (!eligible_for_rht_cast_fusion);
```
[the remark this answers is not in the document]
zhongbozhu: The kernel will use one rng state to generate two random numbers if it's fused. So we only need to generate two rng states when it's the unfused path.

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
```diff
@@ -0,0 +1,1386 @@
...
+  auto tma_load_b = make_tma_copy_B_sm100(
+    SM90_TMA_LOAD{},
+    tensorB,
+    sB(_,_,_,0),
+    cluster_tile_shape,
+    mma);
+
+  // Assert checks problem size should be multiple of 64
+  assert(M % 64 == 0);
+  assert(N % 64 == 0);
```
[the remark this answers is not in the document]
zhongbozhu: will fix

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
```diff
@@ -0,0 +1,1386 @@
...
+      CUTLASS_NOT_IMPLEMENTED();
+    #endif
+    work_tile_info.is_valid_tile = (valid == 1);
+    return work_tile_info;
+  }
+  };
+
+
+
+  // Allocate SMEMork
```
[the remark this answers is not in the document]
zhongbozhu: will fix

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1468,14 +1547,25 @@ void NVFP4Quantizer::quantize_impl(const TensorWrapper& input, TensorWrapper& ou
   }
   size_t cols = input.size(input.ndim() - 1);
 
+  // Restriction for the RHT cast fusion kernel because we are using MMA hardware for computing RHT
+  bool eligible_for_rht_cast_fusion =
+      input.dtype() == DType::kBFloat16 && rows % 64 == 0 && cols % 128 == 0;
```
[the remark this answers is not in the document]
zhongbozhu: this is for dense kernel, split_sections is for MOE, they have different requirements

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1570,103 +1657,47 @@ void NVFP4Quantizer::quantize_impl(const TensorWrapper& input, TensorWrapper& ou
...
+    if (eligible_for_rht_cast_fusion) {
+      // fusion kernel requires passing in RHT matrix directly for maximum performance
+      auto rht_matrix_nvte = makeTransformerEngineTensor(this->rht_matrix);
+      // Fusion kernel that does the following:
+      // 1. Rowwise quantization
+      // 2. RHT followed by columnwise quantization & transpose
+      NVTE_SCOPED_GIL_RELEASE({
+        nvte_hadamard_transform_cast_fusion(input.data(), out.data(), rht_matrix_nvte.data(),
+                                            quant_config, stream);
+      });
```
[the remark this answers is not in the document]
zhongbozhu: it's already checked if eligible_for_rht_cast_fusion is true

transformer_engine/pytorch/csrc/quantizer.cpp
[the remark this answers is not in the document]
zhongbozhu: this is intentional

zhongbozhu:
@ptrendx can we merge this PR soon?

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -639,7 +639,7 @@ void nvte_grouped_gemm(const NVTEGroupedTensor A, int transa, const NVTEGroupedT
                        NVTETensor workspace_cublas, NVTEGroupedMatmulConfig config,
                        cudaStream_t stream) {
   NVTE_ERROR("nvte_grouped_gemm requires cuBLAS 13.2+, but compile-time cuBLAS version is ",
-             CUBLAS_VERSION, ". Please upgrade to CUDA 13.1 or newer.");
+             CUBLAS_VERSION, ". Please upgrade to CUDA 13.2 or newer.");
```
ptrendx: There is a mismatch between CUDA and cuBLAS versions so this was actually correct AFAIK. Also, this file seems unrelated to the rest of the PR, so is it maybe an unintended change?

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -299,7 +299,7 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
 /* EXPERIMENTAL FEATURE AND SUBJECT TO CHANGE. */
 /*! \brief Grouped matrix multiplication: D = alpha * op(A) @ op(B) + beta * C
  *
- * \note Requires cuBLAS 13.2+ (CUDA 13.1+) and Blackwell (SM100) or newer GPU architecture.
+ * \note Requires cuBLAS 13.2+ (CUDA 13.2+) and Blackwell (SM100) or newer GPU architecture.
```
ptrendx: Same comment as in cublaslt_grouped_gemm.cu.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -846,6 +846,11 @@ void split_quantize_nvfp4_impl_with_rht_helper(const TensorWrapper &input,
   // Enable NVFP4 kernels to use math operations that sacrifice
   // accuracy for performance. These optimizations are experimental
   // and inconsistently implemented.
+  // What math is accelerated? Only the high precision math, so numerical impact is minimal
+  // 1. replace x / y by x * (1/y)
+  // 2. replace 1 / x by reciporal_approximate_ftz(x)
```
ptrendx: Point 2 is scary. I've heard that in the past there were issues with the FTZ setting for numerics.

transformer_engine/common/include/transformer_engine/hadamard_transform.h
```diff
@@ -49,6 +49,7 @@ void nvte_hadamard_transform_amax(const NVTETensor input, NVTETensor output, int
 /*! \brief Perform the columnwise hadamard transform cast fusion.
  *
  *  This function is experimental and the API is not stable.
+ *  This function will later be deprecated and replaced by nvte_hadamard_transform_cast_fusion
```
ptrendx: Saying that this function will be deprecated doesn't make sense - if we know it will be superseded we should deprecate it right away (just not remove it for some time).

transformer_engine/common/include/transformer_engine/hadamard_transform.h
```diff
@@ -61,6 +62,21 @@ void nvte_hadamard_transform_cast_fusion_columnwise(const NVTETensor input, NVTE
...
+ *
+ *  This function is experimental and the API is not stable.
+ *
+ *  \param[in]      input           Input tensor to apply Hadamard transform.
+ *  \param[in,out]  output          Output tensor.
+ *  \param[in]      hadamard_matrix Hadamard matrix.
+ *  \param[in]      quant_config    Quantization configuration.
+ *  \param[in]      stream          CUDA stream used for the operation.
+ */
+void nvte_hadamard_transform_cast_fusion(const NVTETensor input, NVTETensor output,
```
ptrendx: Maybe "nvte_nvfp4_quantize_with_rht"?

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
```diff
@@ -35,6 +35,7 @@ def check_quantization_nvfp4_versus_reference(
     M: int,
     N: int,
     contiguous: bool,
+    return_identity: bool,
```
ptrendx: return_rowwise would be a better name.

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
```diff
@@ -78,9 +79,11 @@ def check_quantization_nvfp4_versus_reference(
         x_nvfp4_sut = nvfp4_quantizer.update_quantized(x, x_nvfp4_sut)
 
     # Extract data from NVFP4Tensor
-    assert x_nvfp4_sut._rowwise_data is not None
-    qx: torch.Tensor = x_nvfp4_sut._rowwise_data.view(dtype=torch.uint8)
-    assert x_nvfp4_sut._rowwise_scale_inv is not None
+    qx: torch.Tensor = (
+        x_nvfp4_sut._rowwise_data.view(dtype=torch.uint8)
```
ptrendx, line 83: Isn't it already in uint8?

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
```diff
@@ -221,7 +239,7 @@ def test_rht_with_quantization_block_tiling_versus_reference(
 )
 @pytest.mark.parametrize("x_dtype", [torch.bfloat16], ids=str)
 @pytest.mark.parametrize(
-    "return_transpose", [True, False], ids=["quantize_transpose", "skip_transpose"]
+    "quantize_mode", ["quantize", "quantize_transpose", "quantize_colwise_only"]
```
ptrendx: I think "rowwise_only", "columnwise_only", "both_directions" would be more while actually provide more info at a glance - if I'm looking at the failing test in some time and it says "quantize" then that does not really give me any information without understanding the test first.

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
```diff
@@ -0,0 +1,1386 @@
...
+#include "cutlass/gemm/collective/builders/sm100_common.inl"
+#include "cutlass/numeric_conversion.h"
+#include "cutlass/numeric_types.h"
+#include "cutlass/pipeline/pipeline.hpp"
+#include "cutlass/platform/platform.h"
+#include "cutlass/util/GPU_Clock.hpp"
+#include "cutlass/util/command_line.h"
+#include "cutlass/util/print_error.hpp"
+
+// clang-format off
```
ptrendx, line 43: Why?

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
```diff
@@ -0,0 +1,1386 @@
...
+
+namespace transformer_engine {
+namespace detail {
+namespace {
+
+using namespace cute;
+
+// Ensure Tensor refers to cute::Tensor, not transformer_engine::Tensor
+using cute::Tensor;
```
ptrendx: Can we actually have cute::Tensor explicitly used in this file? Otherwise it is super easy to confuse people.

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
```diff
@@ -0,0 +1,1386 @@
...
+
+struct CLCResponse { uint32_t data[4] = {0}; };
+
+
+CUTLASS_DEVICE
+cutlass::Array<cutlass::float_e2m1_t, 8> StochasticNumericConverterBase(
```
ptrendx: I don't like all of the hardcoded numbers here (not specifically in this function, but throughout the file) - if in the future we will need to tune this kernel it will be a nightmare. Could we instead have some smaller set of parameters and the rest of the numbers derived from that?

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
```diff
@@ -0,0 +1,1386 @@
...
+  // Abort immediately if compilation is not supported
+  constexpr bool is_blackwell_arch = ARCH_BLACKWELL_FAMILY;
+  if constexpr (!is_blackwell_arch) {
+    NVTE_DEVICE_ERROR(
+        "row_col_rht_gemm_device is only supported on Blackwell "
+        "with architecture-specific compilation. "
+        "Try recompiling with sm_100a or similar.");
+    return;
+  }
+  static_assert(kEnableRHTColQuant_ || kEnableRowQuant_,
```
ptrendx: Does that mean that the previous functionality of only columnwise RHT is supported by this kernel? If so, when would we use this vs the other one? If we want to only use this new kernel for everything then we should probably just remove the old kernel in the same PR.

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
```diff
@@ -0,0 +1,1386 @@
...
+  auto sm_count = transformer_engine::cuda::sm_count();
+
+  NVTE_CHECK(n % hadamard_dimension == 0, "row_length must be divisible by hadamard_dimension.");
+
+  NVTE_CHECK(m % hadamard_dimension == 0, "num_rows must be divisible by hadamard_dimension");
+
+  int k_tile_size = 1024;
+
+  // TODO: add support for swizzle sf output
+  const bool use_swizzle_sf_output = false;
```
ptrendx, line 1319: We then should check whether the output tensor requested swizzling and fail if so.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1435,14 +1514,25 @@ void NVFP4Quantizer::quantize_impl(const TensorWrapper& input, TensorWrapper& ou
...
-  QuantizationConfigWrapper quant_config_columnwise;
-  const bool need_separate_columnwise_rng =
-      this->stochastic_rounding && this->with_rht && this->columnwise_usage;
+
+  // Only need a separate rng state when:
+  // 1. Stochastic rounding is enabled
+  // 2. RHT is enabled
+  // 3. Columnwise usage is enabled
+  // 4. Rowwise and columnwise quantization are not fused,
+  //    because within a single kernel we can generate two different random numbers for rowwise and columnwise
```
ptrendx, line 2246: Did we make sure that when we request the RNG state that we tell it that we will generate twice as many numbers? I don't see any changes to that in the diff.

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
zhongbozhu: It will complain about the code being too long and verbose, which is hard to fix from the kernel level.

transformer_engine/pytorch/csrc/quantizer.cpp
zhongbozhu: The rng per-thread has been updated to be num_tensor * 1024.

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
zhongbozhu: This one has more complication, because of the requirement of triggering this fusion is M being 64 multiple instead of 128 multiple. However, the swizzle fusion does requires 128 multiple. For moe, we can zero pad, but we cannot zero pad for dense easily. Unless we lift up the requirement of triggering RHT fusion once for all, or we have some more dispatch logic about if we use this fusion, do we trigger swizzle fusion or not.

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
zhongbozhu: The old kernel (can only do colwise) won't be used, but I wasn't sure if we should just remove the old kernel with the old API.

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
zhongbozhu: Yes it is, just adding it here for clarity

transformer_engine/common/include/transformer_engine/hadamard_transform.h
zhongbozhu: If we deprecate it, the kernel and the API should both be removed. 

EDIT:
to deprecate it, this PR will have to involve JAX, which I don't think is a good idea.

transformer_engine/pytorch/csrc/extensions/cast.cpp
zhongbozhu: Yeah, however, this part of fp32 math is pretty important for the fp4 quantization cost to be hidden under RHT gemm. Users are expected to know the danger and enable it with NVTE_USE_FAST_MATH.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
zhongbozhu: I think it's because of the last rebase to main had some problems that @pggPL fixed later, I also fixed it locally in this PR to pass the test.

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
zhongbozhu: Actually, I was wrong, it's still used by jax. 

we should have a separate PR for the cleanup.

transformer_engine/common/include/transformer_engine/hadamard_transform.h
zhongbozhu: This name comes from the previous API nvte_hadamard_transform_cast_fusion_columnwise, dropping the columnwise. I feel like this will have more consistency? I'm open with both.

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
zhongbozhu: I agree, but all the nvfp4 test cases follow the same naming convention here, we can have a separate PR for updating them all and leave this PR to mainly focus on the integration of this fused kernel.

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
zhongbozhu: fixed

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
zhongbozhu: sure

transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu
ptrendx: Sure, that is something we can solve in the followup, but at least for now we should at least fail explicitly if the output tensor has that option set to make sure that there is no silent wrong answer happening.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
ptrendx: But this comment is not correct - it should be CUDA 13.1.

transformer_engine/common/include/transformer_engine/hadamard_transform.h
ptrendx: Deprecated does not mean removed - deprecated means that it is marked for removal in the future.

transformer_engine/common/include/transformer_engine/hadamard_transform.h
ptrendx: Right, but here we have a more general functionality, so moving to the more general API name feels justified. Please change it.

transformer_engine/pytorch/csrc/quantizer.cpp
ptrendx: And how many numbers do we actually need here?

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
zhongbozhu: fixed

transformer_engine/pytorch/csrc/quantizer.cpp
zhongbozhu: I shouldn't have said num_tensor * 1024, since num_tensor==1. 

Within the kernel, which is a persistent kernel, to generate RNG, it's already using the current data tile index, not the CTA index, so the rng number needed here only needs to be account for one CTA: https://github.com/zhongbozhu/TransformerEngine/blob/4f24746b81f011c826dca290f5cc9eb5732e4cf3/transformer_engine/common/hadamard_transform/row_cast_col_hadamard_transform_cast_fusion.cu#L962 

For this kernel, there are 4 warps handling colwise quantization, and 8 warps handling rowwise quantization. One tile to process a time is 128x128 elements. So in the worse case for colwise quantize, that's 128 * 128 / 4 / 32 = 128 elements touched per thread, 1024 is way above it.

tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py
zhongbozhu: fixed

transformer_engine/common/util/ptx.cuh
```diff
@@ -14,9 +14,11 @@
 #include <cuda.h>
 #include <cuda_runtime.h>
 
-#if CUDA_VERSION >= 12080
+#include "common/common.h"
```
zhongbozhu, line 17: This is for having FP4_TYPE_SUPPORTED defined in this header file, which is later referenced in this ptx.cuh

Expect this fix to solve some compile issues for older CUDA versions.

transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
```diff
@@ -142,6 +142,14 @@ rht_gemm_device(MShape M, NShape N, KShape K, ClusterTileShape cluster_tile,
             const size_t* rng_state)
 {
   using namespace cute;
+  constexpr bool is_blackwell_arch = ARCH_BLACKWELL_FAMILY;
+  if constexpr (!is_blackwell_arch) {
+    NVTE_DEVICE_ERROR(
+        "rht_gemm_device is only supported on Blackwell "
+        "with architecture-specific compilation. "
+        "Try recompiling with sm_100a or similar.");
```
ptrendx:
```suggestion
        "RHT fusion is only supported on Blackwell.");
```
We already check for the architecture-specific compilation in the macro itself.

transformer_engine/common/include/transformer_engine/hadamard_transform.h
```diff
@@ -48,7 +48,7 @@ void nvte_hadamard_transform_amax(const NVTETensor input, NVTETensor output, int
 
 /*! \brief Perform the columnwise hadamard transform cast fusion.
  *
- *  This function is experimental and the API is not stable.
+ *  This has been deprecated in favor of nvte_quantize_with_hadamard_transform.
```
ptrendx: Please use \deprecated (https://www.doxygen.nl/manual/commands.html#cmddeprecated)

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -997,6 +997,11 @@ void split_quantize_nvfp4_impl_with_rht_helper(const TensorWrapper &input,
   // Enable NVFP4 kernels to use math operations that sacrifice
   // accuracy for performance. These optimizations are experimental
   // and inconsistently implemented.
+  // What math is accelerated? Only the high precision math, so numerical impact is minimal
+  // 1. replace x / y by x * (1/y)
```
ptrendx: This should be updated.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -2246,103 +2333,49 @@ void NVFP4Quantizer::quantize_impl(const TensorWrapper& input, TensorWrapper& ou
...
-                                     colwise_data_shape_2d);
-      out_transpose.set_rowwise_scale_inv(out_columnwise_scale_inv.data_ptr,
-                                          static_cast<DType>(out_columnwise_scale_inv.dtype),
-                                          out_columnwise_scale_inv.shape);
-      out_transpose.set_amax(out_columnwise_amax.data_ptr,
-                             static_cast<DType>(out_columnwise_amax.dtype),
-                             out_columnwise_amax.shape);
+  // Fast math toggle: RHT transform can be accelerated
+  // What math is accelerated? Only the high precision math, so numerical impact is minimal
+  // 1. replace x / y by x * (1/y)
```
ptrendx: Should be updated.

transformer_engine/common/hadamard_transform/hadamard_transform_cast_fusion.cu
zhongbozhu: "Blackwell" also includes sm120 though. Needs to be more specific here.

zhongbozhu: Updated

Issue #2558, opened 2026-01-05
NVFP4 performance regression in TE main branch


cael-ling:
**Describe the bug**

TransformerEngine [PR-2411](https://github.com/NVIDIA/TransformerEngine/pull/2411) introduced optimizations to improve NVFP4 computation performance. Previously, using an image built with this PR delivered significant performance gains during the pre-training of the Moonlight-16B model. This PR was merged into the TE main branch two weeks ago. However, when building an image from the main branch and running the same NVFP4 training configuration, the throughput per GPU dropped dramatically and became highly unstable.

**Steps/Code to reproduce bug**

To reproduce the issue, perform the following comparative experiments:

Step1: Build two Slurm images:
Refer to the script at [link](https://gitlab-master.nvidia.com/caell/te_issues/-/blob/main/20260105/build_te_image_0104_main_c988548.sh?ref_type=heads) to build two Slurm images, one image is based on PR-2411, and another image is based on TE main (commit id c988548f72bbc271fe2ab7bad1046b91b577aa29）

Step2: Launch a training job:
Refer to the script at [link](https://gitlab-master.nvidia.com/caell/te_issues/-/blob/main/20260105/eval_nvfp4_train_moonlight.sh?ref_type=heads) to launch a training job. It doesn’t necessarily have to be the Moonlight model, other models are also likely to experience similar performance issues.

**Expected behavior**

The training performance based on these two images are expected to be similar, but in reality, the performance gap is huge (570 TFLOPS VS 240-400 TFLOPS), see [log1](https://gitlab-master.nvidia.com/caell/te_issues/-/blob/main/20260105/result1.log?ref_type=heads),[log2](https://gitlab-master.nvidia.com/caell/te_issues/-/blob/main/20260105/result2.log?ref_type=heads), which makes NVFP4 training solution unusable. 

**Environment overview (please complete the following information)**

 - Environment location: Slurm cluster
 - Method of Transformer Engine install: please refer to [link](https://gitlab-master.nvidia.com/caell/te_issues/-/blob/main/20260105/build_te_image_0104_main_c988548.sh?ref_type=heads) 
 - If method of install is [Docker], provide `docker pull` & `docker run` commands used
please refer to [link](https://gitlab-master.nvidia.com/caell/te_issues/-/blob/main/20260105/eval_nvfp4_train_moonlight.sh?ref_type=heads)

zhongbozhu:
[...]

PR #2564, opened 2026-01-06
[NVFP4][MOE] Bug Fix for NVFP4 Grouped Quant

zhongbozhu:
# Description

Fix this: https://github.com/NVIDIA/TransformerEngine/issues/2558 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


zhongbozhu:
@ksivaman I don't see CI failures from my side, maybe it's because github is not informed super well with the CI status?

transformer_engine/common/hadamard_transform/group_row_cast_col_hadamard_transform_cast_fusion.cu
```diff
@@ -1461,7 +1470,9 @@ void group_hadamard_transform_cast_fusion(const Tensor &input_, std::vector<Tens
                             /*QA=*/reinterpret_cast<TQA *>(rowwise_data_base_ptr),
                             /*SFA=*/reinterpret_cast<TSFA *>(rowwise_scale_inv_base_ptr),
                             /*args=*/kernel_args,
-                            /*rng_state=*/rng_state, /*sm_count=*/sm_count,
+                            /*rng_state=*/rng_state,
+                            /*tile_scheduler_workspace=*/tile_scheduler_workspace,
```
ptrendx, line 1470: I would prefer a more generic workspace name to be honest. Proper handling of this would also require having some function that would return size of the required workspace.

transformer_engine/common/common.h
```diff
@@ -395,6 +395,7 @@ struct QuantizationConfig {
   bool nvfp4_2d_quantization = false;
   bool stochastic_rounding = false;
   bool use_fast_math = false;
+  NVTETensor tile_scheduler_workspace = nullptr;
```
timmoon10: This workspace is specific to `nvte_group_hadamard_transform_cast_fusion`, so would be better for that function to take a workspace arg rather than putting it in the quantization config:
https://github.com/NVIDIA/TransformerEngine/blob/500113dd038f2bb01ef9371b2416f3bb644ac6e4/transformer_engine/common/hadamard_transform/group_row_cast_col_hadamard_transform_cast_fusion.cu#L1487

As a bonus, the workspace could be opaque. If we ever change the kernel to use more or less workspace, we could change it internally and the user doesn't need to make any changes.

timmoon10 (changes requested):

transformer_engine/common/hadamard_transform/group_row_cast_col_hadamard_transform_cast_fusion.cu
zhongbozhu: from API level, it's called quant_workspace now

transformer_engine/common/common.h
zhongbozhu: Fixed.

transformer_engine/common/include/transformer_engine/hadamard_transform.h
```diff
@@ -115,13 +115,14 @@ void nvte_group_hadamard_transform_cast_fusion_columnwise(
  *  \param[in]      split_sections    Array specifying splits in dimension 0 for each output tensor.
  *  \param[in]      num_tensors       Number of output tensors, must be > 0.
  *  \param[in]      quant_config      Quantization configuration.
+ *  \param[in]      quant_workspace   Quantization workspace, a device buffer that can be opaque
```
timmoon10:
```suggestion
 *  \param[in]      quant_workspace   Workspace buffer. Must be at least 4 bytes.
```

transformer_engine/common/hadamard_transform/group_row_cast_col_hadamard_transform_cast_fusion.cu
```diff
@@ -1399,6 +1398,12 @@ void group_hadamard_transform_cast_fusion(const Tensor &input_, std::vector<Tens
     rng_state = reinterpret_cast<const size_t *>(rng_state_tensor.data.dptr);
   }
 
+  uint32_t *tile_scheduler_workspace = nullptr;
+  NVTE_CHECK(quant_workspace.data.dptr != nullptr, "Quantization workspace must be provided.");
+  NVTE_CHECK(quant_workspace.data.buffer_size_bytes() >= sizeof(uint32_t),
+             "Quantization workspace must be at least 4 bytes.");
```
timmoon10, line 1404: If we wanted to be fancy, we could add an option to query the workspace size, similar to how we do it for LayerNorm. If the workspace is not provided, we set the `NVTETensor` with the required size. This way the caller doesn't need to know the details of the workspace size.

That said, I think this approach is fine for now.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -872,10 +872,17 @@ void split_quantize_nvfp4_impl_with_rht_helper(const TensorWrapper &input,
   auto rht_matrix_nvte = makeTransformerEngineTensor(quantizer.rht_matrix);
 
   if (all_aligned_token_dim) {
+    // allocate a tile scheduler workspace
+    auto tile_scheduler_workspace_torch =
+        at::empty({1}, at::device(at::kCUDA).dtype(torch::kInt32));
+    auto nvte_tile_scheduler_workspace =
+        makeTransformerEngineTensor(tile_scheduler_workspace_torch);
     // call the fully-fused grouped kernel for rowwise quantization & colwise RHT quantization transpose
+    // Note that the workspace field can be opaque, this will leave room for furture extensions
```
timmoon10:
```suggestion
```

PR #2596, opened 2026-01-14
 [PyTorch] Add `pad_between_seqs` support for non-CP and CP (A2A and P2P) with FA3 + THD (varlen)

### LINKED ISSUE #2391: Support FlashAttention with pad_between_seqs=True
Currently, the dot product attention backend FA3 in TE does not support padding between sequences. As noted [here](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L663-L670)).

I opened a related [issue in FlashAttention repo](https://github.com/Dao-AILab/flash-attention/issues/1990) and confirmed that this feature could be supported by using the parameters cu_seqlens_q/k and seqused_q/k. 

However, one thing to keep in mind is that we need to manually set the output of the padded positions to zero. I noticed that the fused attention implementation also handles this in a similar way ([link](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/csrc/extensions/attention.cpp#L141-L158)).

Given this, it seems like support for padding between sequences could be reasonably added to FA3 in TE.

sudhakarsingh27:
# Description
## TLDR
Enable `pad_between_seqs=True` for FlashAttention 3 with THD format — both for context parallelism (A2A and P2P comm types) and non-CP paths. Previously `pad_between_seqs` was only supported with FusedAttention.

## Problem
When using THD format with variable-length sequences, sequences are padded for divisibility across CP ranks. With `pad_between_seqs=True`, the attention kernel needs to know actual (unpadded) token counts so it doesn't compute attention over padding tokens. FusedAttention already handled this via `cu_seqlens_q_padded`, but FlashAttention (both FA2 and FA3) had `pad_between_seqs` hardcoded to `False` in the CP path, and FA2 was entirely disabled for `pad_between_seqs + thd`. FA3 can natively handle this via its `seqused_q`/`seqused_k` mechanism.

## Solution
Use FA3's `seqused_q`/`seqused_k` tensors to communicate actual token counts per batch element. Pass `cu_seqlens_q_padded` for tensor memory layout while deriving `seqused_q = cu_seqlens_q[1:] - cu_seqlens_q[:-1]` from the real `cu_seqlens`. This applies to both the CP path (A2A and P2P) and the non-CP path.

Fixes [#2399](https://github.com/NVIDIA/TransformerEngine/issues/2391)

## Type of change

## Changes

### context_parallel.py
- `get_fa_args()`: Add `seqused_q`/`seqused_k` parameters, pass through to FA3 forward and backward positional arg lists (replacing hardcoded `None`s).
- `cp_p2p_fwd_flash_attn()` / `cp_p2p_bwd_flash_attn()`: Accept `pad_between_seqs`, `cu_seqlens_q_padded`, `cu_seqlens_kv_padded`. When enabled, derive `seqused` tensors and override `cu_seqlens` to padded versions (with half-padding for lower-triangle/upper-triangle sections).
- `AttnFuncWithCPAndKVP2P`: Thread `pad_between_seqs` and padded cu_seqlens through all forward/backward `cp_p2p_fwd/bwd_flash_attn` call sites. Save `ctx.pad_between_seqs` for backward.
- `AttnFuncWithCPAndQKVOA2A.forward()`: Add `pad_between_seqs` parameter. When enabled with FA3+THD, derive `seqused` and swap `cu_seqlens` for padded versions before calling `get_fa_args()`.
- `AttnFuncWithCPAndQKVOA2A.backward()`: Same seqused/cu_seqlens override. Use `zeros_like` (not `empty_like`) for gradient init when `pad_between_seqs` since FA3 skips padding positions. Add extra `None` in return tuple for the new `pad_between_seqs` gradient slot.
- `attn_forward_func_with_cp()`: Pass `pad_between_seqs` in A2A args list.

### backends.py
- `FlashAttention.forward()`: Accept `cu_seqlens_q_padded`/`cu_seqlens_kv_padded`. Detect `pad_between_seqs` by comparing padded vs actual cu_seqlens. Pass padded cu_seqlens to CP path. For non-CP FA3 path, derive and pass `seqused_q`/`seqused_k`.

### dot_product_attention.py
- Pass `cu_seqlens_q_padded`/`cu_seqlens_kv_padded` through to `FlashAttention`.

### utils.py
- Only disable FA2 (not FA3) when `pad_between_seqs + thd`. FA3 handles this natively via `seqused`.

### test_attention_with_cp.py
- Add `@pytest.mark.parametrize("pad_between_seqs", [False, True])` to flash attention CP tests.
- Skip `pad_between_seqs=True` for non-THD formats, when FA3 is not installed, and for `a2a+p2p` comm type (not yet supported).

### run_attention_with_cp.py
- Thread `pad_between_seqs` through `generate_input_shapes()` and `run_dpa_with_cp()`.
- When `pad_between_seqs`, set `cu_seqlens_q` to actual lengths (not just for FusedAttention).
- Handle FA3 backward NaN at padding positions: `nan_to_num(nan=0.0)`.
- Zero padding positions explicitly before comparison (FA3 doesn't guarantee zeros at padding slots).
- Add tensor names to NaN/Inf assertion messages for debuggability.

### test_attention.py
- Group FlashAttention with FusedAttention for padded input/output handling in `_run_dot_product_attention()` (previously FlashAttention used original unpadded inputs).
- Pass `cu_seqlens_q_padded`/`cu_seqlens_kv_padded` and `pad_between_seqs` to DPA call for FlashAttention backend.
- Add `pad_between_seqs=True` to parametrize with skip for non-THD formats.

### CP tests (`test_attention_with_cp.py`)

Added `@pytest.mark.parametrize("pad_between_seqs", [False, True])` to `test_cp_with_flash_attention`. Skip conditions: non-THD formats, FA3 not installed, `a2a+p2p` comm type.

5 new tests that run (all `pad_between_seqs=True, thd, bf16`):

| Test | CP comm | Model config |
|------|---------|--------------|
| `True-p2p-thd-cp_1_0-bf16` | P2P | causal, 1 head |
| `True-p2p-thd-cp_2_1-bf16` | P2P | causal, 2 heads |
| `True-a2a-thd-cp_1_0-bf16` | A2A | causal, 1 head |
| `True-a2a-thd-cp_1_2-bf16` | A2A | causal, sliding window |
| `True-a2a-thd-cp_2_1-bf16` | A2A | causal, 2 heads |

### Non-CP tests (`test_attention.py`)

Added `True` to `@pytest.mark.parametrize("pad_between_seqs", [False, True])` on `test_dot_product_attention`, with skip for non-THD. Also changed `_run_dot_product_attention` so FlashAttention uses padded inputs/cu_seqlens and receives `pad_between_seqs=True`.

48 new test IDs collected, but all are skipped because the main parametrize uses `qkv_layout=None` (defaults to sbhd, not thd). The non-CP `pad_between_seqs` + FA3 code path is exercised indirectly when other test functions call `test_dot_product_attention` with `qkv_layout="thd_thd_thd"` (e.g., `test_dpa_softmax_thd`).

# Checklist:


transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -1186,10 +1209,19 @@ def cp_p2p_bwd_flash_attn(
...
     """Per-tile backward call of CP P2P with FlashAttention backend"""
-    dq, dk, dv = [torch.empty_like(x) for x in [q_part, k_part, v_part]]
-    if fa_utils.v2_3_plus and not fa_utils.v2_7_0_plus:
+    if pad_between_seqs:
+        dq, dk, dv = [torch.zeros_like(x) for x in [q_part, k_part, v_part]]
+    else:
+        dq, dk, dv = [torch.empty_like(x) for x in [q_part, k_part, v_part]]
+    if use_flash_attn_3:
+        fa_backward_kwargs["window_size_left"] = -1
+        fa_backward_kwargs["window_size_right"] = -1
```
cyanguwa: This logic is duplicate to L1226-L1228, right?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -1760,7 +1823,14 @@ def forward(
...
                         else:
                             out_per_step[i], softmax_lse_per_step[i], rng_states[i] = (
-                                cp_p2p_fwd_flash_attn(*flash_attn_inputs, *prepare_outputs, section)
+                                cp_p2p_fwd_flash_attn(
+                                    *flash_attn_inputs,
+                                    *prepare_outputs,
+                                    section,
+                                    pad_between_seqs=pad_between_seqs,
+                                    cu_seqlens_q_padded=cu_seqlens_q_padded,
+                                    cu_seqlens_kv_padded=cu_seqlens_kv_padded,
```
cyanguwa: We can add these three args to either flash_attn_inputs or prepare_outputs so we don't have to repeat them 4 times?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -919,6 +921,11 @@ def forward(
         use_flash_attn_3 = False
         if flash_attention_backend is not None and flash_attention_backend > PkgVersion("3.0.0b"):
             use_flash_attn_3 = True
+
+        pad_between_seqs = False
+        if qkv_format == "thd" and cu_seqlens_q_padded is not None:
+            pad_between_seqs = not torch.equal(cu_seqlens_q_padded, cu_seqlens_q)
```
cyanguwa: Can pad_between_seqs be decided ahead of time, passed by the user or something? This wouldn't be CUDA Graph-compatible right?

tests/pytorch/attention/run_attention_with_cp.py
```diff
@@ -502,11 +505,14 @@ def run_dpa_with_cp(
                 tensors_to_deq[i] = tensor.dequantize()
         if not fp8_bwd:
             tensors[0], tensors[5] = tensors_to_deq
-    for tensor in tensors:
+    if pad_between_seqs == "True":
+        # FA3 backward may produce NaN at padding positions; replace with 0
+        tensors = [t.detach().nan_to_num(nan=0.0) if t is not None else t for t in tensors]
```
cyanguwa: Why would there be NaNs? Shouldn't our merge-related arithmetic be able to make it a number?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
sudhakarsingh27: This pattern exists in [dpa.py](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py#L1330-L1341) as well. But yes, it's definitely redundant here

tests/pytorch/attention/run_attention_with_cp.py
sudhakarsingh27: Right, claude was overcautious, I removed it

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Done

tests/pytorch/attention/run_attention_with_cp.py
cyanguwa: It looks like `name` is generated but not used here? Maybe we can use in the assertion messages?

tests/pytorch/attention/run_attention_with_cp.py
```diff
@@ -559,47 +562,68 @@ def run_dpa_with_cp(
         if is_training:
             dq, out = [x.index_select(0, seq_idx_q).contiguous() for x in [dq, out]]
             dk, dv = [x.index_select(0, seq_idx_kv).contiguous() for x in [dk, dv]]
-            dq_, dk_, dv_, out_ = [dq_, dk_, dv_, out_]
+            dq_, dk_, dv_, out_ = [x.clone() for x in [dq_, dk_, dv_, out_]]
```
cyanguwa: I don't think we need to clone here - plus, cloning to the same name doesn't do anything :) We will only compare these tensors with their non-CP counterpart, so probably no need for the clone at all.

tests/pytorch/attention/run_attention_with_cp.py
```diff
@@ -559,47 +562,68 @@ def run_dpa_with_cp(
...
             num_pads_q = cu_pads_q[1:] - cu_pads_q[:-1]
             for x in [dq, out, dq_, out_]:
-                assert torch.count_nonzero(x[cu_seqlens_q_padded[-1] :]).item() == 0
-                for b in range(config.batch_size):
-                    assert (
-                        num_pads_q[b] == 0
-                        or torch.count_nonzero(
+                if pad_between_seqs == "True":
+                    # FA3 doesn't guarantee zeros at padding positions; zero them explicitly
+                    x[cu_seqlens_q_padded[-1] :] = 0.0
```
cyanguwa: Didn't we allocate dq/dk/dv with `torch.zeros_like()`? so the pad positions should all be 0s from the beginning, and then FA doesn't touch them during the computation either. I think we need the assert like in the `pad_between_seqs=False` branch below, instead of a zeroing op.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -935,8 +939,16 @@ def forward(
...
+                    (
+                        cu_seqlens_q_padded
+                        if pad_between_seqs
+                        else (cu_seqlens_q if qkv_format == "thd" else None)
+                    ),
+                    (
+                        cu_seqlens_kv_padded
+                        if pad_between_seqs
+                        else (cu_seqlens_kv if qkv_format == "thd" else None)
+                    ),
```
cyanguwa, line 1039: Should pass in `cu_seqlens_padded` for `thd` and `cu_seqlens` for other formats, regardless of `pad_between_seqs` values? Because in DPA or FlashAttention, we do set `cu_seqlens_padded=cu_seqlens` if `cu_seqlens_padded=None` right?

tests/pytorch/attention/run_attention_with_cp.py
sudhakarsingh27: Yes

tests/pytorch/attention/run_attention_with_cp.py
sudhakarsingh27: Looks like only out_ needed this special handling, so removed this piece of code

tests/pytorch/attention/run_attention_with_cp.py
sudhakarsingh27: This is apparently needed to zero out individual elements in a tensor (`out_` here) but agree that others don't need it, so removing

transformer_engine/pytorch/attention/dot_product_attention/backends.py
sudhakarsingh27: That's only set in fused attention. Leaving this as is for flash attn

qa/L3_pytorch_FA_versions_test/test.sh
```diff
@@ -30,14 +42,43 @@ do
...
+    echo "Running tests in parallel: test_attention.py on GPU 0, test_attention_with_cp.py on GPUs $CP_GPUS ($CP_NUM_GPUS GPUs)"
+
+    CUDA_VISIBLE_DEVICES=0 NVTE_TORCH_COMPILE=0 python3 -m pytest -v -s \
+      --junitxml=$XML_LOG_DIR/pytest.xml \
+      $TE_PATH/tests/pytorch/attention/test_attention.py &
+    PID_ATTN=$!
+
+    CUDA_VISIBLE_DEVICES=$CP_GPUS NVTE_TORCH_COMPILE=0 python3 -m pytest -v -s \
+      --junitxml=$XML_LOG_DIR/pytest_test_attention_with_cp.xml \
+      $TE_PATH/tests/pytorch/attention/test_attention_with_cp.py &
```
cyanguwa, line 70: The a2a+p2p tests need 4 GPUs right? so you might need to budget a bit more than `"$NUM_GPUS" -ge 3`?

tests/pytorch/attention/run_attention_with_cp.py
```diff
@@ -559,13 +562,24 @@ def run_dpa_with_cp(
         if is_training:
             dq, out = [x.index_select(0, seq_idx_q).contiguous() for x in [dq, out]]
             dk, dv = [x.index_select(0, seq_idx_kv).contiguous() for x in [dk, dv]]
-            dq_, dk_, dv_, out_ = [dq_, dk_, dv_, out_]
+            out_ = out_.clone()
```
cyanguwa, line 565: What's this clone and rename for again? I feel you explained it to me before.

tests/pytorch/attention/test_attention.py
```diff
@@ -195,18 +197,18 @@ def test_dot_product_attention(
...
+        cross_attn_causal = (
             config.max_seqlen_q != config.max_seqlen_kv
             and config.attn_mask_type in ["causal", "padding_causal"]
         )
-        and (config.window_size[0] == -1 or FlashAttentionUtils.v2_3_plus)
-    ):
-        flash_attn_supported = True
+        sm = get_device_compute_capability()
+        if not cross_attn_causal and FlashAttentionUtils.v3_is_installed and sm == (9, 0):
+            flash_attn_supported = True
```
cyanguwa, line 211: > get_available_attention_backends doesn't know about FA3's seqused support yet.
It probably knows it by now, as you've finished the PR?

Should the `not cross_attn_causal` logic also go to get_available_attention_backends in utils.py?

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -85,11 +85,20 @@ def get_bash_arguments(num_gpus_per_node, **kwargs):
...
     if num_gpus > torch.cuda.device_count():
         pytest.skip(f"Test requires {num_gpus} GPUs, but found {torch.cuda.device_count()}")
 
+    if pad_between_seqs:
+        if qkv_format != "thd":
+            pytest.skip("pad_between_seqs only applies to THD format!")
+        if not FlashAttentionUtils.v3_is_installed:
+            pytest.skip("pad_between_seqs with CP requires Flash Attention v3!")
+        if cp_comm_type == "a2a+p2p":
+            pytest.skip("pad_between_seqs is not yet supported with A2A+P2P CP comm type!")
```
cyanguwa, line 100: What about AG?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -1188,7 +1214,10 @@ def cp_p2p_bwd_flash_attn(
     section,
 ):
     """Per-tile backward call of CP P2P with FlashAttention backend"""
-    dq, dk, dv = [torch.empty_like(x) for x in [q_part, k_part, v_part]]
+    if pad_between_seqs:
+        dq, dk, dv = [torch.zeros_like(x) for x in [q_part, k_part, v_part]]
+    else:
+        dq, dk, dv = [torch.empty_like(x) for x in [q_part, k_part, v_part]]
```
cyanguwa, line 1220: Just to confirm, we can't do this for fwd, right? Because fwd output is not allocated by us.

tests/pytorch/attention/test_attention.py
sudhakarsingh27: I agree and this is one of the instances where the flags to switch FA on/off are scattered across utils.py and actual tests.

Should we move the entire block (for `pad_bw_seqs`) to utils.py?

qa/L3_pytorch_FA_versions_test/test.sh
```diff
@@ -41,6 +53,35 @@ do
   fi
 
   # Run tests
-  NVTE_TORCH_COMPILE=0 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest.xml $TE_PATH/tests/pytorch/attention/test_attention.py
+  NUM_GPUS=$(nvidia-smi -L | wc -l)
+  echo "Detected $NUM_GPUS GPU(s)"
+  if [ "$NUM_GPUS" -ge 4 ]; then
```
sudhakarsingh27: This should be 5 because we need minimum of 4 for CP and 1 for non CP in parallel

qa/L3_pytorch_FA_versions_test/test.sh
```diff
@@ -41,6 +53,35 @@ do
...
+
+    CUDA_VISIBLE_DEVICES=$CP_GPUS NVTE_TORCH_COMPILE=0 python3 -m pytest -v -s \
+      --junitxml=$XML_LOG_DIR/pytest_test_attention_with_cp.xml \
+      $TE_PATH/tests/pytorch/attention/test_attention_with_cp.py &
+    PID_CP=$!
 
+    wait $PID_ATTN || test_fail "test_attention.py"
+    wait $PID_CP || test_fail "test_attention_with_cp.py"
+  else
+    echo "Running tests sequentially: need >=3 GPUs for parallel execution (1 for test_attention + 2 for test_attention_with_cp)"
```
sudhakarsingh27: similarly, update this comment

tests/pytorch/attention/run_attention_with_cp.py
sudhakarsingh27: `out_` is a view inside the CP custom autograd Function, so in-place zeroing is blocked by PyTorch

qa/L3_pytorch_FA_versions_test/test.sh
sudhakarsingh27: corrected

qa/L3_pytorch_FA_versions_test/test.sh
sudhakarsingh27: done

qa/L3_pytorch_FA_versions_test/test.sh
sudhakarsingh27: done

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: It's a limitation in Flash Attention code - forward never mutates out (so pre-zeroing is overwritten), backward treats dq/dk/dv as in-place mutable (so pre-zeroing sticks). Also this zeroing out works only for CP code where we can provide the args. 

None of the zeroing works for non-CP path because we only have the forward call in TE. 

```
 FA3 / Hopper (hopper/flash_attn_interface.py)
- Forward: mutates_args=() _ namespace flash_attn_3::_flash_attn_forward 
- Backward: mutates_args=("dq", "dk", "dv") _ namespace flash_attn_3::_flash_attn_backward
```

tests/pytorch/attention/test_attention.py
sudhakarsingh27: The block is already in utils.py so removed from here

cyanguwa (dismissed):

qa/L3_pytorch_FA_versions_test/test.sh
```diff
@@ -35,12 +53,64 @@ do
   then
     pip3 install flash-attn-4==${fa_version} nvidia-cutlass-dsl[cu13]==4.4.2 --no-build-isolation
   else
-    git clone https://github.com/Dao-AILab/flash-attention.git
-    cd flash-attention/hopper && python setup.py install
-    cd ../../
+    # FA3 source build (~20 min). Skip if already pre-installed in the base image
+    # via Dockerfile.base INSTALL_FA3=1 (auto-set when RUN_L3_TESTS=1).
```
sudhakarsingh27: > auto-set when RUN_L3_TESTS=1
Not true anymore

qa/L3_pytorch_FA_versions_test/test.sh
```diff
@@ -35,12 +53,64 @@ do
...
+    if python3 -c "import flash_attn_interface" 2>/dev/null; then
```
sudhakarsingh27: why is `flash_attn_interface` an indicator for FA3? It seems like a generic api for any FA version

qa/L3_pytorch_FA_versions_test/test.sh
sudhakarsingh27: Dropped the `(auto-set when RUN_L3_TESTS=1)` annotation in c4b6e076.

qa/L3_pytorch_FA_versions_test/test.sh
sudhakarsingh27: Switched the check to `import flash_attn_3` in c4b6e076 — the `flash_attn_3` package is created only by the FA3 source-install (and matches the cache layout produced by `install_flash_attn.sh`), so the intent is unambiguous now.

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -423,6 +423,12 @@ def test_cp_with_fused_attention(
             )
         if config.attn_bias_type == "post_scale_bias" and is_training:
             pytest.skip("Deterministic mode does not support post_scale_bias with requires_grad")
+        if qkv_format == "thd" and config.num_heads >= 20 and get_device_compute_capability() == (9, 0):
+            pytest.skip(
+                "Deterministic FusedAttention backward with THD format OOMs on sm90"
+                " for this particular test config since cuDNN reserves memory"
+                " proportional to bHSS (known cuDNN issue)."
+            )
```
KshitijLakhani: The motivation for this makes sense to me but seems like the way we are skipping the test is viewing it from a slightly narrow lens. What I mean by that is the main issue is total memory (bhSS) but we seem to be guarding on head dims only

This skip guard would not be correct if tomorrow someone were to add a test with small b,S and H>20 (IIUC) - it almost makes it seem that the issue is the num_heads rather than the total memory

Is there a better way to do this ?

qa/L3_pytorch_FA_versions_test/test.sh
```diff
@@ -54,8 +54,8 @@ do
     pip3 install flash-attn-4==${fa_version} nvidia-cutlass-dsl[cu13]==4.4.2 --no-build-isolation
   else
     # FA3 source build (~20 min). Skip if already pre-installed in the base image
-    # via Dockerfile.base INSTALL_FA3=1 (auto-set when RUN_L3_TESTS=1).
-    if python3 -c "import flash_attn_interface" 2>/dev/null; then
+    # via Dockerfile.base INSTALL_FA3=1.
+    if python3 -c "import flash_attn_3" 2>/dev/null; then
```
KshitijLakhani: nit: I'm not sure if we want to mention the Dockerfile.base here though I understand the motivation

qa/L1_pytorch_distributed_unittest/test.sh
```diff
@@ -22,14 +22,31 @@ mkdir -p "$XML_LOG_DIR"
...
+# Run CP tests (deterministic + non-deterministic) first so they can be parallelized.
+# Each needs 4 GPUs, so >=8 GPUs allows them to run concurrently on disjoint GPU sets.
+NUM_GPUS=$(python3 -c "import torch; print(torch.cuda.device_count())")
+echo "Detected $NUM_GPUS GPU(s)"
+if [ "$NUM_GPUS" -ge 8 ]; then
+    echo "Running CP tests in parallel: non-deterministic on GPUs 0-3, deterministic on GPUs 4-7"
+    CUDA_VISIBLE_DEVICES=0,1,2,3 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_attention_with_cp.xml $TE_PATH/tests/pytorch/attention/test_attention_with_cp.py &
+    PID_CP_NONDET=$!
+    CUDA_VISIBLE_DEVICES=4,5,6,7 NVTE_ALLOW_NONDETERMINISTIC_ALGO=0 python3 -m pytest -v -s --junitxml=$XML_LOG_DIR/pytest_test_attention_deterministic_with_cp.xml $TE_PATH/tests/pytorch/attention/test_attention_with_cp.py &
+    PID_CP_DET=$!
```
KshitijLakhani, line 34: That's neat !
Nice !

KshitijLakhani (dismissed):

tests/pytorch/attention/test_attention_with_cp.py
sudhakarsingh27: Good catch — gated on the actual `b*H*S*S` product instead of `num_heads` in d3bd4e4a. Threshold of `1e9` empirically matches the existing 5-case skip set on the `test_essential` fused subset (cp_2_0, cp_2_2, cp_3_1, cp_4_2, cp_4_3 — bHSS 1.07B–4.29B) and lets the smaller configs (cp_1_0/cp_2_1/cp_2_4/cp_3_2/cp_3_4, all ~0.40B) keep running. Local det+nondet still 33/0 + 45/0 with 5 OOM skips fired by the new gate.

qa/L3_pytorch_FA_versions_test/test.sh
sudhakarsingh27: Fair — dropped the `Dockerfile.base INSTALL_FA3=1` specifics in d3bd4e4a; the `import flash_attn_3` check is the actual contract, no need to couple to an out-of-tree provisioning detail.

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -628,6 +639,20 @@ def test_cp_with_fused_attention(
         pytest.skip("Deterministic mode does not support non-vanilla softmax with FusedAttention")
     if _deterministic and config.attn_bias_type == "post_scale_bias" and is_training:
         pytest.skip("Deterministic mode does not support post_scale_bias with requires_grad")
+    # Det FusedAttention backward with THD on sm90 OOMs because cuDNN reserves
+    # workspace proportional to b*H*S*S. Gate on that product, not num_heads,
```
KshitijLakhani: nit: I think this seems to be adding a comment in a very "Claude"y way and is not necessary. I'd suggest to remove

nit: For magic numbers like 1_000_000_000 I'd suggest using a good constant name like `max_allowable_mem_for_cudnn_on_sm90` and/or add a short comment explaining the choice of the number (this is too long maybe but I guess it conveys the point), in case we ever need to expand this check to another arch, we'd know the reason OR if someone else were to look at this test, they would not be confused at the choice of the number

KshitijLakhani (dismissed):

tests/pytorch/attention/test_attention_with_cp.py
sudhakarsingh27: Renamed the magic number to `SM90_DET_FUSED_THD_BWD_MAX_BHSS = 1 << 30` and replaced the prefatory comment with a casual note tying the constant to what cuDNN actually asks for (~128 * bHSS bytes per a local sweep on cuDNN 9.21.0 sm90 — at 1<<30 thats 128 GiB, doesnt fit on H100s 80 GB) — 1563b105. Also flagged the b>=3 caveat: cuDNN rounds batch up internally so workspace grows super-linearly past b=2 (b=4 asks for 4x the b=2 workspace, not 2x), kept as a note for future readers since the current fused-essential matrix is all b=2.

KshitijLakhani (approved):
LGTM

PR #2600, opened 2026-01-15
[PyTorch] `GroupedTensor` integration

ksivaman:
# Description

#2388 introduced the `GroupedTensor` class in the core library. This PR partly integrates this functionality to the PyTorch bindings.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- [x] Expose a python `GroupedTensor` class.
- [x] Integrate `GroupedTensor` into `GroupedLinear` such that the parameters are contiguous.
- [ ] Expose a C++ `grouped_quantize` API to python similar to the `split_quantize` which returns a quantized `GroupedTensor` that can be directly consumed by the GEMMs (#2502).

# Checklist:

### DIFF
--- a/transformer_engine/common/fused_router/utils.h
+++ b/transformer_engine/common/fused_router/utils.h
@@ -47,7 +47,7 @@ __device__ inline T warp_reduce_on_shmem(T *data_ptr, int data_size, ReduceFuncT
   // Some value is hanlded in local thread
   // Thread 0 is responsible for the: 0-th, 32-th, 64-th, 96-th ...
   // Reduce the value in local thread
-  volatile double val = lane_id < data_size ? static_cast<double>(data_ptr[lane_id]) : default_val;
+  double val = lane_id < data_size ? static_cast<double>(data_ptr[lane_id]) : default_val;
   for (int i = lane_id + kThreadsPerWarp; i < data_size; i += kThreadsPerWarp) {
     val = reduce_func(val, data_ptr[i]);
   }
@@ -85,7 +85,7 @@ __device__ inline T masked_warp_reduce_on_shmem(T *data_ptr, bool *mask, int dat
   // Some value is hanlded in local thread
   // Thread 0 is responsible for the: 0-th, 32-th, 64-th, 96-th ...
   // Reduce the value in local thread
-  volatile double val =
+  double val =
       lane_id < data_size && mask[lane_id] ? static_cast<double>(data_ptr[lane_id]) : default_val;
   for (int i = lane_id + kThreadsPerWarp; i < data_size; i += kThreadsPerWarp) {
     if (mask[i]) {
@@ -183,16 +183,16 @@ __device__ inline void naive_topk_and_mask(T *scores, int data_size, int topk, i
   // After looping topk times, the topk_indices will be the topk indices
   for (int k = 0; k < topk; k++) {
     // Find the max value and its index
-    volatile double val = (lane_id < data_size && !is_masked(k, lane_id))
-                              ? static_cast<double>(scores[lane_id])
-                              : -std::numeric_limits<double>::infinity();
-    volatile int index = (lane_id < data_size) ? lane_id : 0;
+    double val = (lane_id < data_size && !is_masked(k, lane_id))
+                     ? static_cast<double>(scores[lane_id])
+                     : -std::numeric_limits<double>::infinity();
+    int index = (lane_id < data_size) ? lane_id : 0;
     // Some value is hanlded in local thread
     // Thread 0 is responsible for the: 0-th, 32-th, 64-th, 96-th ...
     // Reduce the value in local thread
     for (int i = lane_id + kThreadsPerWarp; i < data_size; i += kThreadsPerWarp) {
-      volatile double cur_val = (is_masked(k, i)) ? -std::numeric_limits<double>::infinity()
-                                                  : static_cast<double>(scores[i]);
+      double cur_val = (is_masked(k, i)) ? -std::numeric_limits<double>::infinity()
+                                         : static_cast<double>(scores[i]);
       if (cur_val > val) {
         val = cur_val;
         index = i;
@@ -200,8 +200,8 @@ __device__ inline void naive_topk_and_mask(T *scores, int data_size, int topk, i
     }
     // Warp shuffle between threads
     for (int s = 16; s > 0; s /= 2) {
-      volatile auto shuffled_val = __shfl_xor_sync(0xffffffff, val, s);
-      volatile auto shuffled_index = __shfl_xor_sync(0xffffffff, index, s);
+      auto shuffled_val = __shfl_xor_sync(0xffffffff, val, s);
+      auto shuffled_index = __shfl_xor_sync(0xffffffff, index, s);
       if (shuffled_val > val) {
         val = shuffled_val;
         index = shuffled_index;


tests/pytorch/test_grouped_tensor.py
```diff
@@ -0,0 +1,429 @@
...
+            assert tensor.data_ptr() == original_data_ptr + expected_offset
+
+    @pytest.mark.parametrize("quantization", _quantization_params)
+    def test_split_into_quantized_tensors_quantized(self, quantization: str) -> None:
+        """Test split_into_quantized_tensors for quantized tensors"""
+        num_tensors = 3
+        shape = [(512, 512) for _ in range(num_tensors)]
+        quantizers = make_quantizers(quantization, num_tensors, shape)
+
+        grouped_tensor = GroupedTensor.make_grouped_tensor(
```
zhongbozhu: This is not graph safe for MOE activation inputs. Maybe we should make a `make_grouped_tensor_graph_safe` API?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,918 @@
...
+    - logical_shape provides the conceptual 2D interpretation
+    - All data is stored on device in contiguous layout
+
+    Note: This structure is used only for combined storage of multiple tensors with the same dtype and scaling mode.
+    """
+
+    def __init__(
+        self,
+        num_tensors: int,
+        shape: List[Tuple[int, int]],
```
zhongbozhu: We cannot do this, shape is supposed to be a tensor on device and we should have a tiny kernel that computes the offsets
Edit:
It's okay for weights where shapes are static, but for moe activation, this needs to be changed.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,918 @@
...
+            N = last_dims_list[0]
+            logical_shape = (sum_first, N)
+        else:
+            # Varying both dims: [1, total_elements]
+            total_elements = sum(s[0] * s[1] for s in shape)
+            logical_shape = (1, total_elements)
+
+        no_quantization = quantizers is None or len(quantizers) == 0 or quantizers[0] is None
+
+        # TODO(ksivaman): (Do we need multiple quantizers?)
```
zhongbozhu: I think we shouldn't have multiple quantizers, multiple quantizers only make sense for recipes like delayed scaling, where quantizer holds amax values. For blockwise current scaling recipes like mxfp8 and nvfp4, quantizer only holds configurations like whether we are doing RHT, then only one quantizer is enough, and it's better for reducing CPU overhead since pybind don't need to convert multiple objects. 

Having separate quantizers also doesn't make sense for cuda graph, because having multiple quantizers implicitly assumes splitting the input, which then implicitly assumes that host knows about  the shape of input for each expert, which will then break cuda graph.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,918 @@
...
+                # For grouped tensors, we need to calculate scale_inv size for all tensors
+                total_scale_elements = 0
+                scale_inv_offsets = [0]
+                for i, s in enumerate(shape):
+                    scale_inv_shape = quantizers[i].get_scale_shape(s, False)
+                    scale_elements = math.prod(scale_inv_shape)
+                    total_scale_elements += scale_elements
+                    if i < num_tensors - 1:
+                        scale_inv_offsets.append(total_scale_elements)
+                scale_inv = torch.empty(total_scale_elements, dtype=torch.uint8, device=device)
```
zhongbozhu, line 457: the number of scale elements cannot be calculated by summing the per-expert split elements for activation inputs where the `m_splits` is now on-device, but instead we should assume that 128x128 aligned shape has been satisfied (otherwise no grouped quantize possible, fall back), and the directly get the number of bytes by M * N / 16 (nvfp4) where M is the total number of tokens for all experts, because you cannot know the per-expert M, you have to calculate the total elements this way and assume padding has been satisfied already.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -724,8 +743,49 @@ def set_meta_tensor(self, fwd: bool, recipe: Recipe) -> None:
...
 
+    def make_grouped_weights(self, defer_init=False) -> None:
+        """
+        Convert parameters into a GroupedTensor and re-register them as parameters.
+        """
+
+        if defer_init:
+            return
+
+        weights = [getattr(self, f"weight{i}") for i in range(self.num_gemms)]
```
zhongbozhu: How about the case where we do fp8 param gather, each `weight` in this case is already a Float8Tensor?

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -957,8 +957,219 @@ class TensorWrapper {
...
+  }
+
+  /*! \brief Wrapped NVTEGroupedTensor. */
+  NVTEGroupedTensor tensor_ = nullptr;
+};
+
+/*! \enum Float8BlockScaleTensorFormat
+ *  \brief Data format for an FP8 block-scaled tensor
+ */
+enum class Float8BlockScaleTensorFormat {
```
zhongbozhu, line 1189: this flag has been deprecated, plus that for grouped tensor, because there is no TP, the format is always GEMM_READY

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
yaox12: Agree. One customer has reviewed this PR and commented that having multiple quantizers would introduce CPU overheads when calling `quantizer.update_usage/set_usage/...` if the number of tensors is very large, for example, in FSDP.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+        tensors: List[torch.Tensor],
+        noop_flag: Optional[torch.Tensor] = None,
+    ) -> Tuple[QuantizedTensorStorage, ...]:
+        """
+        Quantize the GroupedTensor inplace.
+        """
+
+        quantized_tensors = self.split_into_quantized_tensors()
+        for i in range(self.num_tensors):
+            self.quantizer.update_quantized(tensors[i], quantized_tensors[i], noop_flag=noop_flag)
```
cyanguwa, line 959: Shouldn't this be calling the group_quantize from the C side? Why is it still in a loop of `num_tensors` iterations? Thanks.

transformer_engine/pytorch/csrc/type_converters.cpp
```diff
@@ -170,6 +170,121 @@ TensorWrapper NVTETensorFromNVFP4Tensor(py::handle tensor, Quantizer *quantizer)
...
+    ret.set_last_dims(last_dims.data_ptr(), GetTransformerEngineDType(last_dims.scalar_type()),
+                      getTensorShape(last_dims));
+  }
+  if (!tensor.attr("tensor_offsets").is_none()) {
+    const auto &tensor_offsets = tensor.attr("tensor_offsets").cast<at::Tensor>();
+    ret.set_tensor_offsets(tensor_offsets.data_ptr(),
+                           GetTransformerEngineDType(tensor_offsets.scalar_type()),
+                           getTensorShape(tensor_offsets));
+  }
+
```
cyanguwa, line 284: Could we add the `with_gemm_swizzled_scales` support here too please?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+            columnwise_scale_inv_offsets=columnwise_scale_inv_offsets,
+            logical_shape=logical_shape,
+        )
+
+        grouped_tensor.quantized_tensors = grouped_tensor.split_into_quantized_tensors()
+        return grouped_tensor
+
+    def split_into_quantized_tensors(
+        self,
+    ) -> List[Union[QuantizedTensorStorage, torch.Tensor]]:
```
cyanguwa, line 603: Could this function be extended to supporting any number `n` of tensors that satisfies "num_tensors is divisible by n"? This would help with reshaping efforts, for example, in attention.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+from ..nvfp4_tensor import NVFP4Tensor
+from ..float8_tensor import Float8Tensor
+from ..float8_blockwise_tensor import Float8BlockwiseQTensor
+from .float8_tensor_storage import Float8TensorStorage
+from .mxfp8_tensor_storage import MXFP8TensorStorage
+from .float8_blockwise_tensor_storage import Float8BlockwiseQTensorStorage
+from .nvfp4_tensor_storage import NVFP4TensorStorage
+
+
+class GroupedTensor:
```
cyanguwa, line 24: Should this be GroupedTensor or GroupedTensorStorage, like with the other low-precision tensor types?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+        self.scale_inv_offsets = None
+        self.columnwise_scale_inv_offsets = None
+
+    def __repr__(self) -> str:
+        """String representation of the GroupedTensor."""
+        return (
+            f"GroupedTensor(num_tensors={self.num_tensors}, "
+            f"shape={self.shape}, "
+            f"logical_shape={self.logical_shape}, "
+            f"dtype={self.get_dtype()})"
```
cyanguwa, line 273: Should we add the quantizer information here?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+        + None last_dims: all tensors have the same last dimension
+        + Both None: all tensors have identical shapes
+        + Both set: each tensor has unique shape (first_dims[i], last_dims[i])
+
+    Data Layout:
+    - ALL data fields are stored as 1D flattened arrays (data, columnwise_data, scale_inv, etc.)
+    - logical_shape provides the conceptual 2D interpretation
+    - All data is stored on device in contiguous layout
+
+    Note: This structure is used only for combined storage of multiple tensors with the same dtype and scaling mode.
```
cyanguwa, line 49: This seems to be duplicate to the first sentence in the docstring.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+            first_dims  # Device pointer to int64_t array of length num_tensors (or None)
+        )
+        self.last_dims = (
+            last_dims  # Device pointer to int64_t array of length num_tensors (or None)
+        )
+
+        # Offsets for indexing into contiguous 1D layout (OPTIONAL - not needed if all_same_shape())
+        # tensor_offsets[i] = element offset to start of tensor i (cumulative sum of numel for tensors 0..i-1)
+        # Usage: tensor_i_ptr = data.data_ptr() + tensor_offsets[i] * element_size
+        # If None and all_same_shape(): offset[i] = i * M * N (where M, N are common dimensions)
```
cyanguwa, line 126: At first glance, these two tensors are confusing. Maybe we could explain the difference as one being the offset on the element level, one at the byte level?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+            f"GroupedTensor with {self.num_tensors} tensors "
+            f"({', '.join(shape_info) if shape_info else 'uniform'}), "
+            f"logical_shape={self.logical_shape}, "
+            f"dtype={self.get_dtype()}"
+        )
+
+    @staticmethod
+    def make_grouped_tensor_with_shapes(
+        num_tensors: int,
+        shape: List[Tuple[int, int]],
```
cyanguwa, line 297: Could this be named "shapes" please?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+        first_dims: Optional[torch.Tensor],
+        last_dims: Optional[torch.tensor],
+        logical_first_dim: int,
+        logical_last_dim: int,
+        quantizer: Optional[Quantizer] = None,
+        device: Optional[torch.device] = None,
+        dtype: Optional[torch.dtype] = None,
+    ) -> GroupedTensor:
+        """
+        Create a GroupedTensor for storing multiple weight tensors of the same shape.
```
cyanguwa, line 353: Why "weight" tensors?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+                    if self.has_data():
+                        tensor_data = self.data[start_offset:end_offset].view(tensor_shape)
+                        result.append(tensor_data)
+                    elif self.has_columnwise_data():
+                        tensor_data = self.columnwise_data[start_offset:end_offset].view(
+                            tensor_shape
+                        )
+                        result.append(tensor_data)
+                    else:
+                        raise RuntimeError("GroupedTensor has no data to split")
```
cyanguwa, line 681: The if/elseif/else code seems to be the same for both "offsets being None" or "not being None" cases, so maybe we can combine the two? Similarly, can the "data" part of the split for both "no_quantization" and "quantization" paths be combined, so there's only one "for i in range(self.num_tensors)" loop?

tests/pytorch/test_grouped_tensor.py
```diff
@@ -0,0 +1,441 @@
...
+        marks=pytest.mark.skipif(not mxfp8_available, reason=reason_for_no_mxfp8),
+    ),
+    pytest.param(
+        "nvfp4",
+        marks=pytest.mark.skipif(not nvfp4_available, reason=reason_for_no_nvfp4),
+    ),
+]
+
+
+def make_quantizer(quantization: str, num_tensors: int, shape: List[Tuple[int, int]]) -> Quantizer:
```
cyanguwa, line 57: We don't need "num_tensors" as an argument here anymore, I think, because we assume all tensors in the group use the same kind of quantizer.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+    def quantize(
+        self,
+        tensors: List[torch.Tensor],
+        noop_flag: Optional[torch.Tensor] = None,
+    ) -> Tuple[QuantizedTensorStorage, ...]:
+        """
+        Quantize the GroupedTensor inplace.
+        """
+
+        quantized_tensors = self.split_into_quantized_tensors()
```
cyanguwa, line 957: Should "quantized_tensors" here be "self.quantized_tensors"?

transformer_engine/pytorch/csrc/type_converters.cpp
zhongbozhu: I think for grouped tensor, since there is no TP, we can always assume with_gemm_swizzled_scales = True because current multi-tensor-swizzle doesn't work with grouped tensor anyway. 

Although I agree that having a toggle would be good.

tests/pytorch/test_grouped_tensor.py
```diff
@@ -0,0 +1,478 @@
...
+        original_data_ptr = grouped_tensor.data.data_ptr()
+        for i, qtensor in enumerate(grouped_tensor.quantized_tensors):
+            rowwise_data = _get_rowwise_data_tensor(qtensor, quantization)
+            numel = shape[i][0] * shape[i][1]
+            expected_offset = _rowwise_offset_bytes(i * numel, quantization)
+            assert rowwise_data.data_ptr() == original_data_ptr + expected_offset
+
+    @pytest.mark.parametrize(
+        "shape",
+        [[(256, 512), (512, 512), (768, 512)], [(512, 512), (512, 512), (512, 512)]],
```
zhongbozhu, line 366: please add edge cases to the test:
1. zero tokens at all 
2. 0 tokens in the beginning 
3. 0 tokens in the end
4. 0 tokens in the middle

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -80,6 +80,42 @@ py::object quantize(const at::Tensor &tensor, py::handle quantizer, const py::ob
...
+  grouped_input_tensor.set_rowwise_data(
+      tensor.data_ptr(), GetTransformerEngineDType(tensor.scalar_type()), getTensorShape(tensor));
+
+  // Create output GroupedTensor.
+  auto [grouped_output_tensor_cpp, grouped_output_py] = quantizer_cpp->create_grouped_tensor(
+      num_tensors, logical_shape, GetTransformerEngineDType(tensor.scalar_type()),
+      py::reinterpret_borrow<py::object>(quantizer), first_dims, logical_first_dim,
+      logical_last_dim);
+
+  NVTE_SCOPED_GIL_RELEASE({
```
zhongbozhu: let's build a switch here for recipe like `split_quantize`

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -121,7 +133,8 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
         py::arg("output") = py::none(), py::arg("noop") = py::none());
   m.def("dequantize", &transformer_engine::pytorch::dequantize, "Dequantize", py::arg("input"),
         py::arg("otype"));
-
+  m.def("group_quantize", transformer_engine::pytorch::group_quantize, py::arg("tensor"),
```
zhongbozhu, line 136: for future grouped quantization for weights, let's add a slot for an optional `noop` tensor for cuda graph

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -638,6 +857,77 @@ std::pair<TensorWrapper, py::object> Float8BlockQuantizer::create_tensor(
...
+
+  if (rowwise_usage) {
+    rowwise_data = at::empty({total_elements}, uint8_opts);
+    const auto scale_shape = get_scale_shape(logical_shape_vec, false);
+    const int64_t total_scale_elements = static_cast<int64_t>(product(scale_shape));
+    rowwise_scale_inv = at::empty({total_scale_elements}, float_opts);
+  }
+
+  if (columnwise_usage) {
+    columnwise_data = at::empty({total_elements}, uint8_opts);
```
vthumbe1503, line 888: A lot of the code is duplicated throughout for each and e very quantizer. Maybe we should have common code in the base Quantizer implementation

vthumbe1503: Or maybe call the utility function from type_conerters.cpp.

zhongbozhu:
Integrate NVFP4 Graph Safe Grouped Quantize: https://github.com/ksivaman/TransformerEngine-1/pull/14

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -42,6 +42,35 @@ std::vector<T> convert_shape_for_fp4(const std::vector<T>& shape) {
   return ret;
 }
 
+std::optional<at::Tensor> build_grouped_tensor_offsets(const size_t num_tensors,
```
zhongbozhu, line 45: Leave a TODO for the future, use a fused kernel for this function to avoid multiple torch calls.

transformer_engine/pytorch/module/grouped_linear.py
zhongbozhu: this has been discussed offline

tests/pytorch/test_grouped_tensor.py
zhongbozhu: added

Issue #2585, opened 2026-01-12
[JAX] Integrate Permutation to MaxText


tdophung:
After permutation primitive is merged, needs to integrate to MaxText for full model testing

Issue #2536, opened 2025-12-19
[JAX] Extend permutation primitive to multi-GPU


tdophung:
Current JAX primitive for permutation does not have any partitioning or shardy rule. Needs to define them to divide up the work to multiple GPUs.

The approach likely to be taken is to shard this along the B (batch axis) and potentially M (embedding/token hidden size) axis, though less likely. Needs to consider how the data should be sharded in previous operations like router and next operations like FC1, FC2 to avoid re-sharding or too many communications.

tdophung:
Integrated to maxtext with EP + FSDP in #2585 and https://github.com/nvjax-svc-0/maxtext/pull/30/. Closing

PR #2613, opened 2026-01-21
 Permutation to always return group_size/tokens_per_expert

tdophung:
# Description

To enable expert parallelism in MOE, we need to perform a ragged all-to-all after permutation to rearrange the permuted tokens (grouped by experts) onto all GPUs such that each GPU only store a subset of the experts. Ragged all-to-all is used because the number of tokens per expert is most often times, not the same between experts. To do this ragged all-to-all operation, we need to provide it with arguments specifying the number of tokens_per_expert, or often called group sizes in maxText.

In this PR, we compute these group sizes as part of the permutation operation and return them by default

https://github.com/NVIDIA/TransformerEngine/issues/2585
https://github.com/NVIDIA/TransformerEngine/issues/2536

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

return tokens_per_expert by default from summing up expert columns in routing map.

# Checklist:

### DIFF
--- a/transformer_engine/jax/permutation.py
+++ b/transformer_engine/jax/permutation.py
@@ -52,7 +52,7 @@ def token_dispatch(
     Optional[jnp.ndarray],
     jnp.ndarray,
     Optional[jnp.ndarray],
-    Optional[jnp.ndarray],
+    jnp.ndarray,
 ]:
     """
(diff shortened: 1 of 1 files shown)


jberchtold-nvidia (approved):
LGTM, thanks!

PR #2620, opened 2026-01-23
[PyTorch] Introduce QuantizerRole

negvet:
# Description

Introducing `QuantizerRole`

```
@dataclasses.dataclass(frozen=True)
class QuantizerRole:
    module_type: str = ""   # e.g. "linear", "grouped_linear", "dpa"
    tensor_type: str = ""   # e.g. "input", "weight", "grad_output", "qkv", "s"
    name: str = ""          # instance name, e.g. "qkv", "proj", "fc1", "fc2"
```

This is an API that allows to go down to "set this `LayerNormLinear` in this transformer layer to be less aggressively quantized." (fine-grained, per-module/per-tensor quantization control mechanism)
See `test_custom_recipe.py::test_custom_recipe_quantization_targets()`.

Quantizer factory uses roles to dispatch according to its needs.

TE module/op emits a list of `QuantizerRole`:

- `Linear`, `LayerNormLinear`, `LayerNormMLP` emit `module_type="linear"` with `tensor_type` in `{"input", "weight", "grad_output"}`.
- `GroupedLinear` emits `module_type="grouped_linear"`.

`CustomRecipe` accepts a `qfactory` callable that receives `QuantizerRole` and returns a quantizer.

Factories can be composed - e.g., dispatch (to different sub-factories as an option) based on `module_type` (`dpa` vs `linear`) and then refine based on `tensor_type`.

Summary:

- Modules implement `get_quantizer_roles()` that returns a list of QuantizerRole objects.
- During `set_meta_tensor()`, modules call `get_quantizer_roles()` and pass roles to `RecipeState.create()`.
- `RecipeState.create()` assigns roles to the state (e.g., `CustomRecipeState.roles`).
- `CustomRecipeState.make_quantizers()` calls `qfactory(role)` for each role to create quantizers.
- The factory can inspect `role.module_type`, `role.tensor_type`, and `role.name` to dispatch to different quantizers.

This PR enables granular control over the recipe. Which might have some limitations and edge cases.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/pytorch/quantization.py
```diff
@@ -41,9 +42,50 @@
...
+        `"qkv"`, `"proj"`, `"fc1"`, `"fc2"`, `"linear_39"`.
+        Empty string when not provided.
+    position : str
+        Module-internal sub-slot. For modules that fuse multiple sequential operations,
+        e.g. `LayerNormMLP` has `"fc1"` and `"fc2"` sub-slots.
+        Empty string for simple modules.
+    """
+
+    module_type: str
+    tensor_type: str
```
timmoon10: It would be useful to have a default role so users can play with quantizers without needing to use our modules. It's also a reasonable role for the `Quantize` op.
```suggestion
    module_type: str = ""
    tensor_type: str = ""
```

transformer_engine/pytorch/quantization.py
```diff
@@ -1383,24 +1431,17 @@ def make_quantizers(self) -> list:
...
-            roles = [
-                ("linear_grad_output", "linear_grad_input")[i % 2]
-                for i in range(self.num_quantizers)
-            ]
-        else:
-            roles = ["unknown"] * self.num_quantizers
+        roles: List[QuantizerRole]
+        if getattr(self, "roles", None) is None:
+            raise ValueError("CustomRecipeState requires roles to be set.")
+        roles = self.roles
```
timmoon10: Do we need to error out if there are no roles? If we have a default role, we could do something like:
```suggestion
        roles: List[QuantizerRole]
        if getattr(self, "roles", None) is None:
            roles = [QuantizerRole()] * self.num_quantizers
        else:
            roles = self.roles
```

transformer_engine/pytorch/quantization.py
```diff
@@ -41,9 +42,50 @@
...
+        DPA: `"qkv"`, `"s"`, etc.
+    name : str
+        Caller-provided module instance name (e.g. set by the training
+        framework), e.g.
+        `"qkv"`, `"proj"`, `"fc1"`, `"fc2"`, `"linear_39"`.
+        Empty string when not provided.
+    position : str
+        Module-internal sub-slot. For modules that fuse multiple sequential operations,
+        e.g. `LayerNormMLP` has `"fc1"` and `"fc2"` sub-slots.
+        Empty string for simple modules.
```
timmoon10: I feel `name` and `position` are redundant. I see how `position` is basically just there to accommodate `LayerNormMLP`, but I'm uneasy about designing just for that (especially since it's not used publicly in [Megatron-LM](https://github.com/search?q=repo%3ANVIDIA%2FMegatron-LM+LayerNormMLP&type=code) or [Megatron-Bridge](https://github.com/search?q=repo%3ANVIDIA-NeMo%2FMegatron-Bridge+LayerNormMLP&type=code)).

Instead of contorting `QuantizerRole` to work with `LayerNormMLP`, how about we contort `LayerNormMLP`? Instead of the module having a single `name`, it could have `fc1_name` and `fc2_name`.

transformer_engine/pytorch/custom_recipes/quantization_nvfp4.py
timmoon10: We want custom recipes to have a good default:
```suggestion
    if (role.module_type, role.tensor_type) in (("linear", "weight"), ("grouped_linear", "weight")):
        return NVFP4QuantizerRef(
            dtype=utils.Fp4Formats.E2M1,
            quant_tile_shape=(16, 16),
            pow_2_scales=False,
            with_rht=False,
        )
    return NVFP4QuantizerRef(
        dtype=utils.Fp4Formats.E2M1,
        quant_tile_shape=(1, 16),
        pow_2_scales=False,
        with_rht=True,
    )
```

transformer_engine/pytorch/module/linear.py
```diff
@@ -1308,6 +1308,27 @@ def __init__(
...
+            base = [
+                QuantizerRole(module_type="linear", tensor_type="input", name=name),
+                QuantizerRole(module_type="linear", tensor_type="weight", name=name),
+                QuantizerRole(module_type="linear", tensor_type="output", name=name),
+            ]
+        else:
+            base = [
+                QuantizerRole(module_type="linear", tensor_type="grad_output", name=name),
+                QuantizerRole(module_type="linear", tensor_type="grad_input", name=name),
+            ]
```
timmoon10, line 1536: "output" and "grad_input" roles don't make sense. In reality, we are implicitly assuming that the tensor will be consumed by another linear-like layer.
```suggestion
            base = [
                QuantizerRole(module_type="linear", tensor_type="input", name=name),
                QuantizerRole(module_type="linear", tensor_type="weight", name=name),
                QuantizerRole(module_type="linear", tensor_type="input", name=name),
            ]
        else:
            base = [
                QuantizerRole(module_type="linear", tensor_type="grad_output", name=name),
                QuantizerRole(module_type="linear", tensor_type="grad_output", name=name),
            ]
```
Alternatively, if we want to use the output in FP8 DPA, the right role would be `module_type="dpa"` and `module_type="input"`. We should probably make this configurable. I kind of like that this design is exposing the hidden assumptions we've been making.

tests/pytorch/test_custom_recipe.py
```diff
@@ -304,11 +307,11 @@ def quantizer_factory(role):
...
-    assert counts["linear_input"] == 1
-    assert counts["linear_weight"] == 1
-    assert counts["linear_output"] == 1
-    assert counts["linear_grad_output"] == 1
-    assert counts["linear_grad_input"] == 1
+    assert counts["input"] == 1
+    assert counts["weight"] == 1
+    assert counts["output"] == 1
+    assert counts["grad_output"] == 1
+    assert counts["grad_input"] == 1
```
timmoon10:
```suggestion
    assert counts["input"] == 2
    assert counts["weight"] == 1
    assert counts["output"] == 0
    assert counts["grad_output"] == 2
    assert counts["grad_input"] == 0
```

transformer_engine/pytorch/custom_recipes/quantization_nvfp4.py
timmoon10: Also, `tensor_type` doesn't have much meaning without `module_type`. There could be non-GEMM quantized operations with inputs that are different from linear inputs.

timmoon10 (commented):
Overall this design is quite clean and generalizable.

transformer_engine/pytorch/quantization.py
```diff
@@ -41,9 +42,61 @@
...
+    GEMM_MODULE_TYPES = frozenset({"linear", "grouped_linear"})
+
+    module_type: str = ""
+    tensor_type: str = ""
+    name: str = ""
+
+    def is_gemm(self) -> bool:
+        """Whether this role belongs to a GEMM-based module."""
+        return self.module_type in self.GEMM_MODULE_TYPES
+
```
timmoon10: I think this is baking in assumptions about what formats are similar (our recent experiences with grouped tensors makes me wonder if the requirements for "linear" and "grouped_linear" will diverge in the future), and it's also not giving us that much convenience.
```suggestion
```

transformer_engine/pytorch/quantization.py
negvet: Initially did not want to change the name since it is coming from outside, but we do that anyway for nested modules.

transformer_engine/pytorch/quantization.py
negvet: Done

transformer_engine/pytorch/module/linear.py
negvet: I agree about "output" and "grad_input" roles. Setting roles for those slots to `None` (the safest) and enabling the configuration. Also configured it in MHA.

transformer_engine/pytorch/quantization.py
negvet: Sure, removed

negvet:
About https://github.com/NVIDIA/TransformerEngine/pull/2620/commits/1d630848e9d558dd14903f554fbc90851d04ea3e

Custom recipe factories can now return stateful quantizer requests (not just stateless quantizers) — TE detects these request dataclasses, allocates the required shared state (scale/amax buffers), and replaces them with real quantizer instances wired into existing infrastructure. Factories cannot create stateful quantizers directly because the shared buffers must be allocated across all slots simultaneously and registered with TE's global state manager for cross-module distributed reduction, recompute, and checkpointing — a lifecycle that only TE can orchestrate. Delayed scaling is supported via `DelayedScalingRequest`; the composed DelayedScalingRecipeState integrates with distributed amax reduction, activation recompute, and checkpointing. Factories can mix stateful requests and stateless quantizers per-slot within the same `CustomRecipe`.

negvet:
@cyanguwa @ptrendx, please review, enabled custom recipe for attention with current scaling, current envvar-based routing is still functional.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -802,6 +809,80 @@ def set_meta_tensor(self, fwd: bool, recipe: Union[Recipe, List[Recipe]]) -> Non
...
+
+        Forward  (3 GEMMs x 3 = 9 slots):
+            GEMM1 -> QKV (at ``GEMM1_OUTPUT``),
+            GEMM2 -> O   (at ``GEMM2_INPUT``),
+            GEMM3 -> S   (at ``GEMM3_OUTPUT``).
+
+        Backward (3 GEMMs x 2 = 6 slots):
+            GEMM1 -> dQKV (at ``GRAD_OUTPUT1``),
+            GEMM2 -> dO   (at ``GRAD_INPUT2``),
+            GEMM3 -> dP   (at ``GRAD_INPUT3``).
```
ptrendx: This is not intuitive to me - O is created by the last gemm, whereas S is created by the first gemm inside DPA (so should be gemm 2?).

transformer_engine/pytorch/module/base.py
```diff
@@ -726,21 +797,42 @@ def set_meta_tensor(self, fwd: bool, recipe: Recipe) -> None:
...
             recipe,
             mode=("forward" if fwd else "backward"),
             num_quantizers=num_fp8_tensors,
+            roles=roles,
         )
 
         self.fp8_meta[fp8_meta_tensor_key] = recipe_state
         self.quantizers[fp8_meta_tensor_key] = recipe_state.make_quantizers()
 
+    def get_quantizer_roles(
```
ptrendx, line 1018: I don't really understand how to properly create this function. We should have some documentation here on what is expected - let's say you have some number of gemms, you want to assign potentially different quantizers to their inputs/weights and you would like to name those roles X/Y etc. - this is what you would need to have in it.

transformer_engine/pytorch/custom_recipes/quantization_factory_examples.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+
+
+def nvfp4_linear_fp8_dpa_factory(
+    role: Optional[QuantizerRole],
+):
+    """Quantizer factory: NVFP4 for ``Linear``, mixed FP8 for ``DotProductAttention``.
+
+    This factory demonstrates how to use ``CustomRecipe`` with ``fp8_dpa=True``
+    to combine NVFP4 quantization for linear layers with FP8 attention.
```
ptrendx, line 128: To be honest I was hoping that something like this would enable getting rid of fp8_dpa/fp8_mha toggles or at least make them irrelevant for the custom recipe.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
negvet: This is not intuitive to me either, this is a legacy of the cudnn naming convention, trying to fix in https://github.com/NVIDIA/TransformerEngine/pull/2620/commits/f21ce2f077a24f1612fca858dc2e50fbd4411ac6

transformer_engine/pytorch/module/base.py
negvet: Improved in https://github.com/NVIDIA/TransformerEngine/pull/2620/commits/0c1ec9bc693c161e4f0b7ecd41cb1255c5a3481c

transformer_engine/pytorch/custom_recipes/quantization_factory_examples.py
negvet: `fp8_dpa/fp8_mha` is a kernel-dispatch gate. The factory says how to quantize; `fp8_dpa/fp8_mha` says whether to run the FP8 attention kernel path at all. It is possible to infer the gate from the quantizer type produced by the factory. This is a refactor, but nothing dramatic. Still, I would propose to plan it as a follow up, not in this PR. 

cc @cyanguwa

negvet:
Complimentary Megatron-LM PR https://github.com/NVIDIA/Megatron-LM/pull/3755 that enables fp8 attention flags for custom recipe.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -802,6 +810,96 @@ def set_meta_tensor(self, fwd: bool, recipe: Union[Recipe, List[Recipe]]) -> Non
...
+        if fwd:
+            qkv_role = QuantizerRole(module_type="dpa", tensor_type="qkv", name=name)
+            o_role = self._output_quantizer_role
+            if o_role is None:
+                o_role = QuantizerRole(name=f"{name}.dpa_output" if name else "dpa_output")
+            s_role = QuantizerRole(module_type="dpa", tensor_type="s", name=name)
+            base = [
+                qkv_role,
+                qkv_role,
+                qkv_role,  # Group 1: QKV (inputs to Q·K^T)
```
cyanguwa, line 965: If I were to separate the Q, K, V tensors, would I just need to pass in q_role, k_role, v_role here?

negvet: Right, this is a valid extension for the future refactor.

transformer_engine/pytorch/quantization.py
```diff
@@ -1071,12 +1181,15 @@ def create(
...
             raise ValueError(f"{recipe.__class__.__name__} is not supported")
-        return cls(
+        state = cls(
             recipe,
             mode=mode,
             num_quantizers=num_quantizers,
             device=device,
         )
+        # Optional QuantizerRole objects
+        state.roles = roles
```
timmoon10: Injecting state like this is dangerous. Better to change the constructors to take in an optional `roles` kwarg.

transformer_engine/pytorch/module/linear.py
timmoon10: Making the output and grad input roles configurable is a really satisfying solution. It keeps our support for quantized MHA, avoids non-local quantizers, and generalizes to non-MHA use-cases.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -2279,6 +2279,24 @@ def get_attention_quantizers(fp8, quantizers):
     dP_quantizer.set_usage(rowwise=True, columnwise=False)
     dP_quantizer.interal = True
 
+    _fp8_types = (Float8Quantizer, Float8CurrentScalingQuantizer)
```
timmoon10: We should make sure not to break MXFP8 attention:
```suggestion
    _supported_quantizer_types = (Float8Quantizer, Float8CurrentScalingQuantizer, MXFP8Quantizer)
```

transformer_engine/pytorch/quantization.py
timmoon10, line 1: We should update `Float8BlockScalingRecipeState` and `NVFP4BlockScalingRecipeState` so that their `make_quantizers` respects the quantizer roles.

cyanguwa (approved):

cyanguwa (dismissed):
The changes look good to me. Would the new quantizer setup be compatible with the existing one? Would TE tests (test_attention.py) or attention users need to update their code immediately?

negvet:
> Would the new quantizer setup be compatible with the existing one? Would TE tests (test_attention.py) or attention users need to update their code immediately?

This new setup is fully compatible with the current one - no need to immediately update anything. Legacy users will not notice it.
This new approach can be fully tested (and updated as required) before deprecating the current one.

transformer_engine/pytorch/quantization.py
negvet: Fixed, with a fallback to the legacy positional indexing when `get_quantizer_roles()` is not defined by a module/op

timmoon10 (dismissed):
LGTM, pending CI

timmoon10 (dismissed):

negvet:
`QuantizerRole` change triggers a recipe state rebuild (e.g. via `output_quantizer_role` setter).
This leads to the silent loss of the state of the stateful recipes, e.g. `scale` / `amax_history` for delayed scaling.

https://github.com/NVIDIA/TransformerEngine/pull/2620/commits/83c405bc4a411d68380dbc9dfb9c84cdcbbce219 introduces `RecipeState.inherit_state_from()`, making buffers persistent. `DelayedScalingRecipeState` declares its buffers, stateless recipes have empty buffers (no op).

timmoon10 (dismissed):

ptrendx:
@negvet Could you fix the issues hit in the CI runs?

timmoon10 (approved):
Test failures are unrelated

negvet:
The same failures on the main

PR #2634, opened 2026-01-29
Add 2d quant for mxfp8

kunlunl:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh
```diff
@@ -774,11 +819,14 @@ void quantize(const Tensor &input, const Tensor *act_input, const Tensor *noop,
                 }
               }
 
+              if (use_2d_quantization) { scaling_type = ScalingType::BIDIMENSIONAL; }
```
ptrendx: Is this intentional? I think it will fail if the output does not have one of the rowwise/columnwise pieces allocated. Also the 2D quantization should not care about that?

kunlunl: There are two reasons for this:
* First, the current implementation of this quantize kernel makes it easy for me to obtain the 32x32 block amax from the columnwise code path, but it's more difficult from the rowwise code path. Therefore, I cannot implement 2D quant when using rowwise alone.
* Second, this 2D quant is currently only used for weights, so in practice it's only needed when using BIDIMENSIONAL mode.

However, I think one of your concerns is valid—I shouldn't directly modify the scaling type here; I should check it instead. I've now changed it so that when using 2D quant, it asserts that scaling type == BIDIMENSIONAL.

ptrendx: In the inference case the weights would only ask for the rowwise quantization, so we need to be able to support that.

kunlunl: I get it. You are right. I'll fix it after holiday.

Oleg-Goncharov:
Hi @kunlunl, could you please share performance benchmarks for this kernel?

kunlunl:
Benchmarked the direct MXFP8 quantization kernel path on a B200 node with preallocated output tensors, timing only `MXFP8Quantizer.update_quantized(x, out)`.

- Existing 1D MXFP8 is unchanged within noise: rowwise 1D median delta after vs before is `-0.03%`, bidirectional 1D median delta is `+0.09%` across the full BF16 shape grid.
- For large shapes (`rows * cols >= 16M`), 2D rowwise-only quantization has `+21.13%` median overhead vs rowwise 1D, with range `+13.87%` to `+40.46%`.
- For large shapes, 2D bidirectional quantization has `+34.84%` median overhead vs bidirectional 1D, with range `+23.44%` to `+43.08%`.
- Small 512/1024-row shapes are launch-bound at roughly 8-10 us, so their overhead percentages are noisier.

Raw data:

| dtype | shape | [before PR] rowwise-only 1D (ms) | [after PR] rowwise-only 1D (ms) | rowwise-only 1D delta | rowwise-only 2D (ms) | rowwise-only 2D overhead | (before PR) bidirectional 1D (ms) | [after PR] bidirectional 1D (ms) | bidirectional 1D delta | bidirectional 2D (ms) | bidirectional 2D overhead |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| bf16 | 512x4096 | 0.008495 | 0.008579 | +0.99% | 0.008508 | -0.83% | 0.009059 | 0.009260 | +2.22% | 0.009092 | -1.81% |
| bf16 | 512x8192 | 0.008423 | 0.008436 | +0.15% | 0.010898 | +29.18% | 0.009011 | 0.009208 | +2.19% | 0.009281 | +0.79% |
| bf16 | 512x14336 | 0.008415 | 0.008266 | -1.77% | 0.010321 | +24.86% | 0.009019 | 0.009003 | -0.18% | 0.008732 | -3.01% |
| bf16 | 1024x4096 | 0.008345 | 0.008319 | -0.31% | 0.008281 | -0.46% | 0.008971 | 0.009027 | +0.62% | 0.008605 | -4.67% |
| bf16 | 1024x8192 | 0.008388 | 0.008389 | +0.01% | 0.008289 | -1.19% | 0.008954 | 0.009027 | +0.82% | 0.009780 | +8.34% |
| bf16 | 1024x14336 | 0.008502 | 0.008407 | -1.12% | 0.008791 | +4.57% | 0.010690 | 0.010708 | +0.17% | 0.014229 | +32.88% |
| bf16 | 2048x4096 | 0.008506 | 0.008348 | -1.86% | 0.008343 | -0.06% | 0.009010 | 0.008923 | -0.97% | 0.009845 | +10.33% |
| bf16 | 2048x8192 | 0.008643 | 0.008326 | -3.67% | 0.009929 | +19.25% | 0.011830 | 0.011859 | +0.25% | 0.016414 | +38.41% |
| bf16 | 2048x14336 | 0.011797 | 0.011786 | -0.09% | 0.015681 | +33.05% | 0.020977 | 0.020964 | -0.06% | 0.027227 | +29.88% |
| bf16 | 4096x4096 | 0.008481 | 0.008275 | -2.43% | 0.009962 | +20.39% | 0.011860 | 0.011834 | -0.22% | 0.016379 | +38.41% |
| bf16 | 4096x8192 | 0.013639 | 0.013624 | -0.11% | 0.018682 | +37.13% | 0.024959 | 0.024925 | -0.14% | 0.030895 | +23.95% |
| bf16 | 4096x14336 | 0.026210 | 0.026269 | +0.23% | 0.029964 | +14.07% | 0.041282 | 0.041316 | +0.08% | 0.051001 | +23.44% |
| bf16 | 8192x4096 | 0.013846 | 0.013847 | +0.01% | 0.019450 | +40.46% | 0.024972 | 0.025005 | +0.13% | 0.031574 | +26.27% |
| bf16 | 8192x8192 | 0.030222 | 0.030212 | -0.03% | 0.034402 | +13.87% | 0.046742 | 0.046652 | -0.19% | 0.058996 | +26.46% |
| bf16 | 8192x14336 | 0.052623 | 0.052612 | -0.02% | 0.060849 | +15.66% | 0.077843 | 0.078558 | +0.92% | 0.104298 | +32.77% |
| bf16 | 8192x28672 | 0.101789 | 0.102477 | +0.68% | 0.124094 | +21.09% | 0.156356 | 0.153673 | -1.72% | 0.216968 | +41.19% |
| bf16 | 14336x4096 | 0.027565 | 0.026853 | -2.58% | 0.034055 | +26.82% | 0.042927 | 0.041762 | -2.71% | 0.055737 | +33.46% |
| bf16 | 16384x4096 | 0.030578 | 0.030799 | +0.72% | 0.038679 | +25.59% | 0.047245 | 0.047752 | +1.07% | 0.065045 | +36.21% |
| bf16 | 16384x8192 | 0.059549 | 0.059547 | -0.00% | 0.072155 | +21.17% | 0.089288 | 0.089659 | +0.42% | 0.125423 | +39.89% |
| bf16 | 16384x14336 | 0.102436 | 0.102350 | -0.08% | 0.121594 | +18.80% | 0.153783 | 0.153917 | +0.09% | 0.215644 | +40.10% |
| bf16 | 28672x8192 | 0.101802 | 0.102307 | +0.50% | 0.126185 | +23.34% | 0.157015 | 0.154281 | -1.74% | 0.220748 | +43.08% |

tests/pytorch/test_mxfp8_2d_quantize.py
```diff
@@ -0,0 +1,488 @@
...
+    return torch.pow(2.0, e8m0.to(torch.float32) - 127)
+
+
+def _mxfp8_2d_quantize_reference(
+    x: torch.Tensor,
+) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
+    """Reference MXFP8 2D quantization using one scale per 32x32 block."""
+    rows, cols = x.shape
+    assert rows % MXFP8_BLOCK_SIZE == 0
+    assert cols % MXFP8_BLOCK_SIZE == 0
```
ptrendx, line 71: Considering this is the only test of this functionality (and it would be better if we also had C++ tests covering it, since it is a change in common), we should test also the cases where the shapes are less nice, and so should not impose those requirements. The MXFP8 quantization kernel does not actually need them - the only real restriction if the 16B alignment from TMA, so the cols % 16 should be 0.

transformer_engine/common/recipe/__init__.py
```diff
@@ -359,8 +359,11 @@ class MXFP8BlockScaling(Recipe):
             `high_precision` keeps original high-precision operands for backward,
             and `dequantized` dequantizes saved operands to the active high-precision
             compute dtype (e.g. BF16/FP16/FP32) for backward.
+    enable_2d_quantization : bool, default = False
+                If set to `True`, 2D block scaling is used for weight tensors.
     """
 
+    enable_2d_quantization: bool = os.getenv("NVTE_MXFP8_ENABLE_2D_QUANTIZATION", "0") == "1"
```
ptrendx: This should be the last argument, otherwise it is a breaking API change.

transformer_engine/pytorch/quantization.py
```diff
@@ -1506,7 +1506,23 @@ def make_quantizers(self) -> list:
...
-        return [MXFP8Quantizer(self.dtype) for i in range(self.num_quantizers)]
+        if self.mode not in ("forward", "backward"):
+            raise RuntimeError(f"Unexpected recipe mode ({self.mode})")
+
+        if self.mode == "backward" or not self.recipe.enable_2d_quantization:
+            return [MXFP8Quantizer(self.dtype) for i in range(self.num_quantizers)]
+
+        def _use_2d_quantization(idx: int) -> bool:
+            role = self._slot_role(idx)
+            return role.module_type in ("linear", "grouped_linear") and role.tensor_type == "weight"
```
ptrendx, line 1541: If we want to use it in grouped linear then we should also enable the grouped quantization kernel with this feature.

tests/pytorch/test_mxfp8_2d_quantize.py
```diff
@@ -0,0 +1,488 @@
...
+    (128, 128),
+    (256, 256),
+    (256, 1024),
+    (1024, 256),
+    (256, 288),
+    (320, 320),
+    (352, 256),
+    (2048, 2048),
+    (1024, 2048),
+    (2048, 1024),
```
ptrendx, line 31: I don't think we need all these shapes to be honest and that increases the CI workload. Could we be more thoughtful here?

transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh
```diff
@@ -264,6 +264,13 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK)
         }
       }
 
+      if constexpr (kIs2DBlockScaling) {
+#pragma unroll
+        for (int i = 16; i > 0; i /= 2) {
+          thread_amax = fmaxf(thread_amax, __shfl_xor_sync(0xffffffff, thread_amax, i));
+        }
+      }
```
ptrendx: I'm pretty sure we already have the warp-level reduction somewhere (I think in utils.cuh). If that one is good, please use that instead. If not, please improve it rather than adding another reduction code here.

transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh
```diff
@@ -415,8 +422,26 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK)
...
+      if constexpr (kIs2DBlockScaling) {
+        __shared__ e8m0_t block_scales_2d[THREADS_X];
+        __shared__ float block_amax_2d[THREADS_X * THREADS_Y];
+        block_amax_2d[tid_X_rowwise * THREADS_Y + tid_Y_rowwise] = thread_amax;
+        __syncthreads();
+        if (tid_Y_rowwise == 0) {
+          float amax_2d = 0.0f;
+#pragma unroll
+          for (int i = 0; i < THREADS_Y; ++i) {
+            amax_2d = fmaxf(amax_2d, block_amax_2d[tid_X_rowwise * THREADS_Y + i]);
```
ptrendx: Do we need to use FP32 here? Maybe we could use BF16 version for the BF16 input data.

transformer_engine/common/recipe/__init__.py
kunlunl: Checked and addressed.

tests/pytorch/test_mxfp8_2d_quantize.py
kunlunl: Partially addressed in Python and fully covered at the common C++ level. I added a C++ operator test in `tests/cpp/operator/test_cast_mxfp8.cu` with an independent 32x32-block CPU reference, exact E8M0 scale comparison, and FP8 data comparison. It covers rowwise-only, colwise-only, and bidirectional modes over a broader shape matrix including less-nice shapes such as `{1,16}`, `{16,48}`, `{65,80}`, `{127,400}`, and `{993,512}`.

For the Python tests, I kept the current shape list 32-aligned. The Python `MXFP8Quantizer` public allocation path still enforces the existing 32-aligned MXFP8 storage contract, and relaxing it to `cols % 16 == 0` would broaden Python allocator behavior beyond the intended scope of this PR. The less-nice shape coverage is therefore in the common C++ test, which exercises the kernel/API path directly.

transformer_engine/pytorch/quantization.py
kunlunl: Addressed. I restricted the recipe policy to regular `linear` weight roles only. `grouped_linear` and other unsupported or unknown roles stay on the existing 1D MXFP8 path. There is also a test that verifies grouped-linear weight roles do not enable 2D quantization.

tests/pytorch/test_mxfp8_2d_quantize.py
kunlunl: Addressed with a smaller Python shape set. The Python test matrix now keeps representative 32-aligned shapes rather than the full original list. I did keep `(320, 320)` and `(352, 256)` because they were "less nice".

transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh
kunlunl: Addressed. The colwise 2D reduction now uses the existing warp-level max reduction helper.

transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh
kunlunl: Addressed. The rowwise 2D amax staging type now uses `IType` for the cast-only BF16/FP16 path, reducing shared-memory pressure and avoiding unnecessary FP32 staging there. It still uses `float` for FP32 input and for activation/dbias paths where keeping FP32 intermediate precision is safer.

transformer_engine/common/recipe/__init__.py
```diff
@@ -361,13 +361,16 @@ class MXFP8BlockScaling(Recipe):
...
+    enable_2d_quantization : bool, default = False
+                If set to `True`, 2D block scaling is used for weight tensors.
     """
 
     margin: int = 0
     fp8_format: Format = Format.E4M3
     fp8_dpa: bool = False
     fp8_mha: bool = False
     backward_override: Optional[str] = os.getenv("NVTE_BACKWARD_OVERRIDE", None)
+    enable_2d_quantization: bool = os.getenv("NVTE_MXFP8_ENABLE_2D_QUANTIZATION", "0") == "1"
```
ptrendx: If the env variable is not strictly needed I would prefer to not introduce one, but rather rely on the user just specifying the option in their MXFP8 recipe.

transformer_engine/pytorch/quantization.py
```diff
@@ -1520,7 +1520,23 @@ def make_quantizers(self) -> list:
...
-        return [MXFP8Quantizer(self.dtype) for i in range(self.num_quantizers)]
+        if self.mode not in ("forward", "backward"):
+            raise RuntimeError(f"Unexpected recipe mode ({self.mode})")
+
+        if self.mode == "backward" or not self.recipe.enable_2d_quantization:
+            return [MXFP8Quantizer(self.dtype) for i in range(self.num_quantizers)]
+
+        def _use_2d_quantization(idx: int) -> bool:
+            role = self._slot_role(idx)
+            return role.module_type == "linear" and role.tensor_type == "weight"
```
ptrendx: There is an issue here with the GroupedLinear - it uses the grouped quantization, which was not updated to be 2D-quant aware, so will break. We should add the 2D awareness to the grouped quantize kernel too.

transformer_engine/common/recipe/__init__.py
kunlunl: Done.

transformer_engine/pytorch/quantization.py
kunlunl: Done.

ptrendx (dismissed):

ptrendx (approved):

PR #2637, opened 2026-01-31
Fix FP8 block scaling with sequence parallel

cuichenx:
# Description
### Problem
Using Float8BlockQuantizer with sequence parallel fails with ``AssertionError: All-gather requires quantizable tensor for quantizer Float8BlockQuantizer`` when local tensor dimensions aren't divisible by 128.

### Solution
Skip the assert_dim_for_all_gather check for Float8BlockQuantizer since gather_along_first_dim already has a fallback path
Fix the fallback in _start_all_gather_fp8_blockwise to handle already-quantized inputs by dequantizing before high-precision all-gather

###Note
The fallback path (high-precision all-gather → quantize) may increase the communication overhead. 

### Verification
The code change does not alter convergence behavior
image

When SP is True, the previous code did not run. When SP is False, this change doesn't affect anything. 
image

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/pytorch/utils.py
```diff
@@ -452,6 +452,13 @@ def assert_dim_for_all_gather(
 ) -> None:
     """Assert that tensor dimensions are supported for all-gather"""
     if with_all_gather:
+        # Float8BlockQuantizer has a fallback path in gather_along_first_dim
+        # that handles non-quantizable local tensors by all-gathering in
+        # high precision first, then quantizing the result.
+        from .tensor import Float8BlockQuantizer
+
+        if isinstance(quantizer, Float8BlockQuantizer):
+            return
```
timmoon10: On second thought, this is a kludge. `gather_along_first_dim` is already able to handle the case where the local data is too small to be quantized, and there should be no special treatment for FP8 block scaling.

I think the right fix is to modify `Linear`/`LayerNormLinear`/`LayerNormMLP` so that they don't call `assert_dim_for_all_gather`, or at least we should add an option so they raise a warning rather than an error if the local data is not quantizable.

timmoon10 (changes requested):

PR #2651, opened 2026-02-04
[JAX] Debugging inspect utility

jberchtold-nvidia:
# Description

Given `jax.debug.print/callback` is currently broken (issue), this PR introduces an experimental alternative for use for our own internal debugging. This new debugging API allows us to inspect tensors but will be experimental and may have breaking changes without a deprecation process.

Usage:
```
     x = <some logic to compute x>

      from transformer_engine.jax.debug.experimental import inspect_array as te_inspect_array
      x= te_inspect_array(x, "some_name")

     <something that consumes  x>
```

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Introduces a new debugging tool for dumping binary blobs of tensor values for multi-GPU inspection.

# Checklist:

### DIFF
--- a/transformer_engine/jax/debug/experimental/inspect.py
+++ b/transformer_engine/jax/debug/experimental/inspect.py
@@ -101,7 +101,10 @@ register_primitive(InspectPrimitive)
 
 
 def _inspect_array_inner(x: jnp.ndarray) -> jnp.ndarray:
-    assert InspectPrimitive.outer_primitive is not None, "InspectPrimitive FFI is not registered. Please ensure the C++ extension is properly built and registered."
+    assert InspectPrimitive.outer_primitive is not None, (
+        "InspectPrimitive FFI is not registered. Please ensure the C++ extension is properly built"
+        " and registered."
+    )
     return InspectPrimitive.outer_primitive.bind(
         x,
         jnp.min(x).astype(jnp.float32),
--- a/tests/jax/test_custom_call_compute.py
+++ b/tests/jax/test_custom_call_compute.py
@@ -1934,7 +1934,8 @@ class TestDebugInspectFFI:
             jnp.float16,
             # Note: fp4 currently doesn't work
             # jnp.float4_e2m1fn
-        ] + ([jnp.float8_e4m3fn, jnp.float8_e5m2] if is_fp8_supported else [])
+        ]
+        + ([jnp.float8_e4m3fn, jnp.float8_e5m2] if is_fp8_supported else []),
     )
     def test_debug_inspect_ffi(self, shape, dtype):
         from transformer_engine.jax.debug.experimental import inspect_array, load_array_dump


transformer_engine/jax/csrc/extensions/inspect.cpp
```diff
@@ -0,0 +1,98 @@
...
+  // Log the tensor metadata to the console
+  printf("Tensor data written to %s (shape: [", filename.c_str());
+  for (size_t i = 0; i < input_buf.dimensions().size(); ++i) {
+    printf("%zu", static_cast<size_t>(input_buf.dimensions()[i]));
+    if (i < input_buf.dimensions().size() - 1) {
+      printf(", ");
+    }
+  }
+  printf("], dtype: %d", static_cast<int>(input_buf.element_type()));
+  printf(", min: %f, max: %f, mean: %f, std: %f)\n", min_val, max_val, mean_val, std_val);
```
tdophung, line 82: If you want to keep this print, it would also be nice to print out where the stat is coming from (which GPU/shard)

transformer_engine/jax/debug/experimental/inspect.py
```diff
@@ -0,0 +1,170 @@
...
+        filename (str): The path to the binary file containing the array data.
+        shape (tuple): The shape of the array to be loaded.
+        dtype (jnp.dtype): The data type of the array to be loaded.
+
+    Returns:
+        jnp.ndarray: The loaded JAX array.
+    """
+    with open(filename, "rb") as f:
+        data = f.read()
+    array = jnp.frombuffer(data, dtype=dtype).reshape(shape)
```
tdophung, line 173: Does this handle 4-bit data as well? Will the data printed be legible for e2m1 or e0m3?
This might be useful for debugging quantization outputs

transformer_engine/jax/csrc/extensions/inspect.cpp
```diff
@@ -0,0 +1,98 @@
...
+  std::vector<uint8_t> input_data(input_buf.size_bytes());
+  cudaMemcpyAsync(input_data.data(), input_buf.untyped_data(), input_buf.size_bytes(),
+                  cudaMemcpyDeviceToHost, stream);
+
+  float min_val{}, max_val{}, mean_val{}, std_val{};
+  cudaMemcpyAsync(&min_val, min_buf.untyped_data(), sizeof(float), cudaMemcpyDeviceToHost, stream);
+  cudaMemcpyAsync(&max_val, max_buf.untyped_data(), sizeof(float), cudaMemcpyDeviceToHost, stream);
+  cudaMemcpyAsync(&mean_val, mean_buf.untyped_data(), sizeof(float), cudaMemcpyDeviceToHost,
+                  stream);
+  cudaMemcpyAsync(&std_val, std_buf.untyped_data(), sizeof(float), cudaMemcpyDeviceToHost, stream);
```
tdophung: Should do error checking here. I think it is NVTE_CHECK_CUDA?

transformer_engine/jax/debug/experimental/inspect.py
```diff
@@ -0,0 +1,170 @@
...
+    )
+
+
+@partial(jax.custom_vjp, nondiff_argnums=())
+def _inspect(
+    x,
+):
+    """ """
+    output, _ = _inspect_fwd_rule(
+        x,
```
tdophung, line 123: is this for eventually inspecting grads? Maybe we can put in a TODO here for inspecting grads?

transformer_engine/jax/debug/experimental/inspect.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def inspect_array(x: jnp.ndarray, name: str) -> jnp.ndarray:
+    """Utility function to inspect JAX arrays by printing their name, shape, dtype, and statistics.
+
+    Args:
+        x (jnp.ndarray): The JAX array to inspect.
+        name (str): The name of the array for identification in the output.
+    """
+    # TODO: Handle the name of the tensor in the primitive and output files
```
tdophung: This does not need to be in this PR, but I did ask claude about this and it suggested 2 options:
1. We can hash the string name into a number, to pass to XLA, and maintain a hash map on the python side to attach to the filename. 
2. We do not pass the name to _inspect (like what you are doing here), but we "mv" the file in the filesystem to a different name depending on the argument `name`
Wdyt?

jberchtold-nvidia: I'm not sure if option 2 will work, since we are running the Python "inspect" function during trace time but the file doesn't get written until the JAX program is jit'd and actually running. So I'm not sure how we'd move the file unless we did something like in option 1 with a dictionary.

I do like option 1, that is a good idea and may be our best option. Another alternative I was considering but haven't looked into yet is a static buffer with a fixed maximum string capacity. On the attention side we pass static args thru a buffer that is a struct that can support decoding fields on the C++ side. There may be a way to pack the  string on the Python side and read on the  C++ side assuming the string doesn't exceed some fixed char array maxsize. But that idea still needs some investigation, so your option 1 sounds like the most promising.

transformer_engine/jax/debug/experimental/inspect.py
jberchtold-nvidia: This isn't directly for inspecting gradients. If we insert a custom primitive into a JAX graph, JAX errors since it doesn't understand that our inspect FFI is really a no-op and therefore it doesn't know how to compute the gradient of it. So this VJP rule just helps JAX know how to handle this in the backward and pass gradients thru.

This current inspect API wraps a single tensor, which makes sense for the forward. We just wrap some intermediate result in `inspect` and we can look at the results. Usually this would be immediately after some custom operation we added, e.g. comparing MaxText vs. TE permute. However, since we're just wrapping the tensor, the gradient in this VJP will actually be the incoming gradient from the following operation, not the gradient of our custom op.

So I think we do want support for inspecting gradients, but instead of wrapping individual tensors  on the forward, it may make more sense to wrap a given function, like  our permute, and let us inspect the forward outputs and gradient outputs from our op (not the incoming gradients)

I have a somewhat hard-coded version of this idea working on my branch that for a GEMM logs stats on fwd output and dgrad and wgrad. If we make it generic enough, something similar could be used for permute and other ops too

Wdyt?

transformer_engine/jax/csrc/extensions/inspect.cpp
jberchtold-nvidia: Good point, updated

transformer_engine/jax/csrc/extensions/inspect.cpp
jberchtold-nvidia: Good idea, added "[gpu%d]:" as a prefix here

transformer_engine/jax/debug/experimental/inspect.py
jberchtold-nvidia: I tried and unfortunately it doesn't seem to work with fp4. I'm not sure if this is an issue with my FFI or if it's an issue with fp4 support in jnp.frombuffer. But I've added unit tests for the other dtypes with a note that fp4 doesn't work currently

tdophung: Sounds good, we can figure out the expansion to 4-bit later

transformer_engine/jax/debug/experimental/inspect.py
tdophung: option 1 would likely be easier implement IMO since you do not have to try decoding the passed static buffer into multiple fields (I assume the struct might have multiple field?) . Anyways I will leave this here without resolving the thread to re-visit

transformer_engine/jax/debug/experimental/inspect.py
tdophung: that seems like a cool idea, but if there are multiple outputs, we need a way to differentiate them. and potentially save them to different bins.

transformer_engine/jax/csrc/extensions/inspect.cpp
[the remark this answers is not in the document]
tdophung: This is fine, we want to see this whenenver we use inspect util

tdophung:
LGTM! Thanks for making this change

PR #2654, opened 2026-02-06
[PyTorch] Python `GroupedTensor`

ksivaman:
# Description

Extracts the python pieces of `GroupedTensor` infrastructure from #2600. Since this is mainly focused on creation of weights as a single `GroupedTensor` and exposing them as multiple `QuantizedTensor`s for PyTorch, this portion does not need to be graph capturable.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- [x] Expose a python `GroupedTensor` class.
- [x] Integrate `GroupedTensor` into `GroupedLinear` such that the parameters are contiguous.

# Checklist:

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -137,7 +137,7 @@ Flax
       for _ in range(10):
         loss, (param_grads, other_grads) = fwd_bwd_fn(params, other_variables, inp)
 
-For a more comprehensive tutorial, check out our `Quickstart Notebook <https://github.com/NVIDIA/TransformerEngine/blob/main/docs/examples/quickstart.ipynb>`_.
+For a more comprehensive tutorial, check out our `Getting Started Guide <https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/getting_started.html>`_.
 
 .. overview-end-marker-do-not-remove
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -23,8 +23,6 @@ Additionally, we offer [Jupyter notebook tutorials](https://github.com/NVIDIA/Tr
   - **FP8 Weight Caching**: Avoiding redundant FP8 casting during multiple gradient accumulation steps to improve efficiency.
 - [Introduction to FP8](https://github.com/NVIDIA/TransformerEngine/blob/main/docs/examples/fp8_primer.ipynb)
   - Overview of FP8 datatypes (E4M3, E5M2), mixed precision training, delayed scaling strategies, and code examples for FP8 configuration and usage.
-- [TE Quickstart](https://github.com/NVIDIA/TransformerEngine/blob/main/docs/examples/quickstart.ipynb)
-  - Introduction to TE, building a Transformer Layer using PyTorch, and instructions on integrating TE modules like Linear and LayerNorm.
 - [Basic MNIST Example](https://github.com/NVIDIA/TransformerEngine/tree/main/examples/pytorch/mnist)
 
 # JAX
@@ -34,7 +32,9 @@ Additionally, we offer [Jupyter notebook tutorials](https://github.com/NVIDIA/Tr
   - Model Parallelism: Divide a model across multiple GPUs for parallel training.
   - Multiprocessing with Model Parallelism: Multiprocessing for model parallelism, including multi-node support and hardware affinity setup.
 - [Basic MNIST Example](https://github.com/NVIDIA/TransformerEngine/tree/main/examples/jax/mnist)
- 
+- [TE JAX Integration Tutorial](https://github.com/NVIDIA/TransformerEngine/blob/main/docs/examples/te_jax_integration.ipynb)
+  - Introduction to integrating TE into an existing JAX model framework, building a Transformer Layer, and instructions on integrating TE modules like Linear and LayerNorm.
+
 # Third party
 - [Hugging Face Accelerate + TE](https://github.com/huggingface/accelerate/tree/main/benchmarks/fp8/transformer_engine)
   - Scripts for training with Accelerate and TE. Supports single GPU, and multi-GPU via DDP, FSDP, and DeepSpeed ZeRO 1-3.
--- a/qa/L0_pytorch_unittest/test.sh
+++ b/qa/L0_pytorch_unittest/test.sh
@@ -48,7 +48,11 @@ python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_attention.xml $T
 NVTE_ALLOW_NONDETERMINISTIC_ALGO=0 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_attention_deterministic.xml $TE_PATH/tests/pytorch/attention/test_attention.py || test_fail "NVTE_ALLOW_NONDETERMINISTIC_ALGO=0 test_attention.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_kv_cache.xml $TE_PATH/tests/pytorch/attention/test_kv_cache.py || test_fail "test_kv_cache.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_hf_integration.xml $TE_PATH/tests/pytorch/test_hf_integration.py || test_fail "test_hf_integration.py"
-NVTE_TEST_CHECKPOINT_ARTIFACT_PATH=$TE_PATH/artifacts/tests/pytorch/test_checkpoint python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_checkpoint.xml $TE_PATH/tests/pytorch/test_checkpoint.py || test_fail "test_checkpoint.py"
+export NVTE_TEST_CHECKPOINT_ARTIFACT_PATH=$TE_PATH/artifacts/tests/pytorch/test_checkpoint
+if [ ! -d "$NVTE_TEST_CHECKPOINT_ARTIFACT_PATH" ]; then
+    python3 $TE_PATH/tests/pytorch/test_checkpoint.py --save-checkpoint all || error_exit "Failed to generate checkpoint files"
+fi
+python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_checkpoint.xml $TE_PATH/tests/pytorch/test_checkpoint.py || test_fail "test_checkpoint.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_fused_router.xml $TE_PATH/tests/pytorch/test_fused_router.py || test_fail "test_fused_router.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_partial_cast.xml $TE_PATH/tests/pytorch/test_partial_cast.py || test_fail "test_partial_cast.py"
 
--- a/tests/pytorch/test_checkpoint.py
+++ b/tests/pytorch/test_checkpoint.py
@@ -101,7 +101,7 @@ class TestLoadCheckpoint:
         # Path to save checkpoint
         if checkpoint_dir is None:
             checkpoint_dir = TestLoadCheckpoint._checkpoint_dir()
-        checkpoint_dir.mkdir(exist_ok=True)
+        checkpoint_dir.mkdir(parents=True, exist_ok=True)
         checkpoint_file = checkpoint_dir / f"{name}.pt"
 
         # Create module and save checkpoint
--- a/transformer_engine/pytorch/graph.py
+++ b/transformer_engine/pytorch/graph.py
@@ -451,11 +451,12 @@ def _make_graphed_callables(
                 if is_training:
                     inputs = tuple(i for i in static_input_surface if i.requires_grad)
                     with _none_grad_context_wrapper(inputs):
+                        outputs_requiring_grad = tuple(
+                            o for o in outputs if o is not None and o.requires_grad
+                        )
                         torch.autograd.backward(
-                            tuple(o for o in outputs if o.requires_grad),
-                            grad_tensors=tuple(
-                                torch.empty_like(o) for o in outputs if o.requires_grad
-                            ),
+                            outputs_requiring_grad,
+                            grad_tensors=tuple(torch.empty_like(o) for o in outputs_requiring_grad),
                         )
                         grad_inputs = tuple(input.grad for input in inputs)
 
@@ -616,19 +617,22 @@ def _make_graphed_callables(
                         # Note for _reuse_graph_input_output_buffers: grad output is only used
                         # within backward, so we can reuse the same static buffers every time.
                         static_grad_outputs_keys = tuple(
-                            (o.shape, o.dtype, o.layout) for o in static_outputs if o.requires_grad
+                            (o.shape, o.dtype, o.layout)
+                            for o in static_outputs
+                            if o is not None and o.requires_grad
                         )
                         if static_grad_outputs_keys in static_grad_outputs_dict:
                             static_grad_outputs = static_grad_outputs_dict[static_grad_outputs_keys]
                         else:
                             static_grad_outputs = tuple(
-                                torch.empty_like(o) if o.requires_grad else None
+                                torch.empty_like(o) if o is not None and o.requires_grad else None
                                 for o in static_outputs
                             )
                             static_grad_outputs_dict[static_grad_outputs_keys] = static_grad_outputs
                     else:
                         static_grad_outputs = tuple(
-                            torch.empty_like(o) if o.requires_grad else None for o in static_outputs
+                            torch.empty_like(o) if o is not None and o.requires_grad else None
+                            for o in static_outputs
                         )
                     if is_training:
                         inputs = tuple(i for i in static_input_surface if i.requires_grad)
@@ -636,7 +640,9 @@ def _make_graphed_callables(
                             bwd_graph, pool=mempool
                         ):
                             torch.autograd.backward(
-                                tuple(o for o in static_outputs if o.requires_grad),
+                                tuple(
+                                    o for o in static_outputs if o is not None and o.requires_grad
+                                ),
                                 grad_tensors=tuple(o for o in static_grad_outputs if o is not None),
                                 retain_graph=retain_graph_in_backward,
                             )
@@ -719,7 +725,8 @@ def _make_graphed_callables(
         ):
             # For now, assumes all static_outputs require grad
             static_grad_outputs = tuple(
-                torch.empty_like(o) if o.requires_grad else None for o in static_outputs
+                torch.empty_like(o) if o is not None and o.requires_grad else None
+                for o in static_outputs
             )
             if is_training:
                 inputs = tuple(i for i in static_input_surface if i.requires_grad)
@@ -727,7 +734,7 @@ def _make_graphed_callables(
                     bwd_graph, pool=mempool
                 ):
                     torch.autograd.backward(
-                        tuple(o for o in static_outputs if o.requires_grad),
+                        tuple(o for o in static_outputs if o is not None and o.requires_grad),
                         grad_tensors=tuple(o for o in static_grad_outputs if o is not None),
                         retain_graph=retain_graph_in_backward,
                     )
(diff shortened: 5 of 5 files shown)


transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+        first_dims: Optional[torch.Tensor],
+        last_dims: Optional[torch.tensor],
+        logical_first_dim: int,
+        logical_last_dim: int,
+        quantizer: Optional[Quantizer] = None,
+        device: Optional[torch.device] = None,
+        dtype: Optional[torch.dtype] = None,
+    ) -> GroupedTensor:
+        """
+        Create a GroupedTensor for storing multiple weight tensors of the same shape.
```
vthumbe1503, line 352: Minor comment: Intent of this API is to create grouped tensor with variable first_dims/last_dims, so we can write that in the comment, since this is not going to be used to create weights.

Also the API can be named to make_grouped_tensor_graph_safe? So, people know this API is safe to use within a forward/backward of a module which we need to be cuda graphable

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+            # Offsets are based on number of elements and not pointers.
+            # Kernels need to calculate precise pointers based on size of elements.
+
+            # TODO(ksivaman): Single kernel + remove the host offset calculation.
+            tensor_offsets = torch.cat(
+                [
+                    torch.zeros(1, device=first_dims.device, dtype=first_dims.dtype),
+                    torch.cumsum(first_dims * logical_last_dim, dim=0),
+                ]
+            )
```
vthumbe1503, line 401: I see the above comment to have single kernel and am not sure what your plan is to implement that.
But with torch op you can avoid one memory op using

```
tensor_offsets = torch.empty(num_tensors + 1, device=first_dims.device, dtype=first_dims.dtype)
torch.cumsum(first_dims * logical_last_dim, dim=0, out=tensor_offsets[1:])
```

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+from ..nvfp4_tensor import NVFP4Tensor
+from ..float8_tensor import Float8Tensor
+from ..float8_blockwise_tensor import Float8BlockwiseQTensor
+from .float8_tensor_storage import Float8TensorStorage
+from .mxfp8_tensor_storage import MXFP8TensorStorage
+from .float8_blockwise_tensor_storage import Float8BlockwiseQTensorStorage
+from .nvfp4_tensor_storage import NVFP4TensorStorage
+
+
+class GroupedTensor:
```
timmoon10, line 24: I don't think it's a good idea to put everything within a single class. We should have an abstract base class (`GroupedTensorBase`) and concrete classes like `GroupedTensor` (or `UnquantizedGroupTensor`?), `MXFP8GroupedTensor`, `NVFP4GroupedTensor`. The giant-pile-of-attrs design results in ugly implemenations (like the if-else blocks in `make_grouped_tensor`) and it generalizes poorly (`columnwise_data` is treated very differently between FP8 and MXFP8, enough that giving them the same name is questionable). We do use this design in the C++ grouped tensor class, but that should be viewed as a short-term expedient and not a long-term design (https://github.com/NVIDIA/TransformerEngine/pull/2388#discussion_r2543614556).

ptrendx: This ultimately depends on what we want to optimize for. If we believe that the majority of things we are going to write is going to be here is about "grouped" functionality that does not really care about the underlying format (or stuff where we could delegate that decision to C++ which has the full knowledge of the quantizer type and could implement things without huge if/else blocks) then it makes sense to have a single class here. If we believe that the majority of the functionality will be dependent on the quantization format then I agree that we should split this into multiple classes.
@ksivaman Can you comment on that?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+                for i, s in enumerate(shape):
+                    scale_inv_shape = quantizer.get_scale_shape(s, False)
+                    columnwise_scale_elements = math.prod(scale_inv_shape)
+                    total_columnwise_scale_elements += columnwise_scale_elements
+                    if i < num_tensors - 1:
+                        columnwise_scale_inv_offsets.append(total_columnwise_scale_elements)
+                columnwise_scale_inv = torch.empty(
+                    total_columnwise_scale_elements, dtype=torch.uint8, device=device
+                )
+        elif quantizer._get_compatible_recipe().delayed():
```
zhongbozhu, line 474: Since we have gone for single quantizer, we should remove delayed scaling recipe & per-tensor current scaling for now since their quantizers are not stateless.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+                columnwise_scale_inv_offsets = [0]
+                for i, s in enumerate(shape):
+                    columnwise_scale_inv_shape = quantizer.get_scale_shape(s, True)
+                    total_columnwise_scale_elements += math.prod(columnwise_scale_inv_shape)
+                    if i < num_tensors - 1:
+                        columnwise_scale_inv_offsets.append(total_columnwise_scale_elements)
+                columnwise_scale_inv = torch.empty(
+                    total_columnwise_scale_elements, dtype=torch.float32, device=device
+                )
+        elif quantizer._get_compatible_recipe().float8_current_scaling():
```
zhongbozhu, line 558: float8_current_scaling can work with GroupedTensor once we refactored its implementation to remove the amax tensor out of its quantizer. Then it will be safe to put a single quantizer into the grouped tensor.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+                    columnwise_data=columnwise_data,
+                    columnwise_scale_inv=columnwise_scale_inv,
+                    fp8_dtype=self.quantizer.dtype,
+                    quantizer=self.quantizer,
+                    with_gemm_swizzled_scales=self.quantizer.optimize_for_gemm,
+                )
+                result.append(tensor)
+
+            # Delayed scaling or current scaling (both use Float8TensorStorage)
+            elif recipe.delayed() or recipe.float8_current_scaling():
```
zhongbozhu, line 761: let's assert an error for this case?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
zhongbozhu: I think GroupedTensor in python should be a truthful copy of the C++ grouped tensor, so I do think it's okay to have a single class.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
ksivaman: `torch.empty` would do garbage initialization whereas we need `tensor_offsets[0]` to be explicitly 0, so either way we'd have to do multiple kernels if using pytorch ops. That's why the plan is to later add a small cuda kernel so that we can call it from the C++ extensions and also for Jax as well.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
ksivaman: This API is actually used to create the weights and is not graph safe (for now), which is fine as it's used 1 time during creation.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
ksivaman: This is what we discussed offline, but actually now that I think about it, this is used for FP8 parameters creation, so we cannot simply un-support recipes here. The correct method is to probably use multiple quantizers, or at least have a way for the user to supply multiple quantizers.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
ksivaman: Same comment as above^

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
ksivaman: Same comment as above^

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
vthumbe1503: Oh my mistake using torch.zeros instead of torch.empty should do the trick. Sure cuda kernel later sounds good.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
vthumbe1503: I thought make_grouped_tensor_with_shapes is used create weights. Since weight's shapes are going to be constant. Whats the intent of make_grouped_tensor_with_shapes then?

And whats the API we are going to be using to create inputs? Dont we need graph safe for that one?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
ksivaman: This `GroupedTensor` in python is used mostly as a storage class and is basically a copy of C `GroupedTensor` with some additional functionality for weight initialization and convenience. I think it's best to keep it simple and avoid over engineering at this stage. In the next steps when we implement a `QuantizedGroupedTensor`, say for FSDP2 + quantized parameter support, we could revisit if a small refactor would be helpful.

timmoon10: I agree matching the C++ class is reasonable for now so we can meet deadlines. In the long term, we should refactor both the C++ and Python class to have proper polymorphism. This will be a painful process.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+                shape_info.append("varying first dim")
+            if not self.all_same_last_dim():
+                shape_info.append("varying last dim")
+
+        return (
+            f"GroupedTensor with {self.num_tensors} tensors "
+            f"({', '.join(shape_info) if shape_info else 'uniform'}), "
+            f"logical_shape={self.logical_shape}, "
+            f"dtype={self.get_dtype()}"
+        )
```
ptrendx, line 291: Why are those different?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+            last_dims=None,
+            logical_first_dim=logical_first_dim,
+            logical_last_dim=logical_last_dim,
+            quantizer=quantizer,
+            device=device,
+            dtype=dtype,
+        )
+
+    @staticmethod
+    def make_grouped_tensor(
```
ptrendx, line 341: Is the fact that this is on the Python side and not C++ is intentional or a TODO? What is the actual usage of this call? Is it just a helper for the weights creation and meant to be changed later?

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+            offsets=offsets,
+            scale_inv_offsets=scale_inv_offsets,
+            columnwise_scale_inv_offsets=columnwise_scale_inv_offsets,
+            logical_shape=logical_shape,
+        )
+
+        grouped_tensor.quantized_tensors = grouped_tensor.split_into_quantized_tensors()
+        return grouped_tensor
+
+    def split_into_quantized_tensors(
```
ptrendx, line 606: Is this mostly for debugging? In general it would be good to revisit the docs for the functions and indicate which ones we expect to be used in the typical case (and e.g. are graph safe) and which ones are for debug/not performant/not graph safe.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+            f"GroupedTensor with {self.num_tensors} tensors "
+            f"({', '.join(shape_info) if shape_info else 'uniform'}), "
+            f"logical_shape={self.logical_shape}, "
+            f"dtype={self.get_dtype()}"
+        )
+
+    @staticmethod
+    def make_grouped_tensor_with_shapes(
+        num_tensors: int,
+        shape: List[Tuple[int, int]],
```
zhongbozhu, line 296: let's call it shapes

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
```diff
@@ -0,0 +1,942 @@
...
+            tensor_offsets = torch.cat(
+                [
+                    torch.zeros(1, device=first_dims.device, dtype=first_dims.dtype),
+                    torch.cumsum(first_dims * logical_last_dim, dim=0),
+                ]
+            )
+            offsets = tensor_offsets.tolist()
+            first_dims_list = first_dims.tolist()
+            for i in range(num_tensors):
+                shape.append((first_dims_list[i], logical_last_dim))
```
zhongbozhu, line 405: We should make this part consistent with the C++ grouped tensor creation. ie. making them both graph safe.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
zhongbozhu: having a single quantizer without disabling unsupported recipes will make it numerically incorrect for delayed scaling & per-tensor current scaling right? since they do need multiple quantizer instances with hold multiple stateful tensors

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -724,8 +742,49 @@ def set_meta_tensor(self, fwd: bool, recipe: Recipe) -> None:
...
+        """
+
+        if defer_init:
+            return
+
+        weights = [getattr(self, f"weight{i}") for i in range(self.num_gemms)]
+        weight_quantizers = self._get_weight_quantizers()
+
+        # Create the weight storage.
+        grouped_weights = GroupedTensor.make_grouped_tensor_with_shapes(
```
zhongbozhu, line 766: how can grouped linear reference this grouped_weights as a whole and trigger the grouped gemm?
maybe we need self.grouped_weights = grouped_weights here and then pass the reference of this variable to the grouped_linear_forward.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -724,8 +742,49 @@ def set_meta_tensor(self, fwd: bool, recipe: Recipe) -> None:
...
+        )
+
+        # Copy existing params into storage.
+        # TODO(ksivamani): Verify correctness of copy for all recipes.
+        with torch.no_grad():
+            for i in range(self.num_gemms):
+                grouped_weights.quantized_tensors[i].copy_(weights[i])
+
+        # Re-register the grouped weights as parameters.
+        for i in range(self.num_gemms):
```
zhongbozhu, line 783: one TODO item for the future reference: we might need to support a toggle to support registering the grouped parameters as a whole

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -724,8 +742,49 @@ def set_meta_tensor(self, fwd: bool, recipe: Recipe) -> None:
...
+                f"weight{i}",
+                torch.nn.Parameter(grouped_weights.quantized_tensors[i]),
+                init_fn=self.init_method,
+                get_rng_state_tracker=self.get_rng_state_tracker,
+                fp8_meta_index=self._offsets["weight"] + i * self._num_fp8_tensors_per_gemm["fwd"],
+            )
+
+        self.set_tensor_parallel_attributes(defer_init=defer_init)
+
     def reset_parameters(self, defer_init=False):
```
zhongbozhu, line 794: can we adjust the order of the functions here? like put helper functions on top and caller in the bottom

transformer_engine/pytorch/module/grouped_linear.py
vthumbe1503: Those changes will be part of this PR @zhongbozhu 
https://github.com/NVIDIA/TransformerEngine/pull/2669

transformer_engine/pytorch/module/grouped_linear.py
vthumbe1503: I think keeping self.set_tensor_parallel_attributes(defer_init=defer_init) at the end of reset_parameters is ideal. Since otherwise these attributes would be set again during meta device init after reset_parameters is called

ksivaman: Yes, this needs to be the last step such that those TP specific attributes are retained

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
ksivaman: Yes about the docs. But this is not only a debugging function but also used for creating different parameters out of the Grouped Tensor to be used as weights.

transformer_engine/pytorch/tensor/storage/float8_tensor_storage.py
```diff
@@ -104,6 +104,24 @@ def clear(self):
...
+            raise TypeError("copy_from_storage expects Float8TensorStorage")
+        if self._fp8_dtype != src._fp8_dtype:
+            raise RuntimeError("FP8 dtype mismatch in copy_from_storage")
+
+        def _copy_optional(
+            dst: Optional[torch.Tensor],
+            src_tensor: Optional[torch.Tensor],
+        ):
+            if dst is not None and src_tensor is not None:
+                dst.copy_(src_tensor)
```
timmoon10, line 119: We should handle the edge case that the tensor buffer don't match, or else we won't have confidence that the tensor is safe to read after this function:
```suggestion
            if dst is not None:
                if src_tensor is None:
                    raise RuntimeError(
                        f"Attempting to copy from {src.__class__.__name__} "
                        f"to {self.__class__.__name__}, but data buffers are not compatible."
                    )
                dst.copy_(src_tensor)
```
Alternatively, we could manually transpose the missing data or dequantize-requantize.

We should also do something similar for the other tensor classes.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -724,8 +742,51 @@ def set_meta_tensor(self, fwd: bool, recipe: Recipe) -> None:
...
+            return
+
+        weights = [getattr(self, f"weight{i}") for i in range(self.num_gemms)]
+        weight_quantizers = self._get_weight_quantizers()
+
+        # Create the weight storage.
+        grouped_weights = GroupedTensor.make_grouped_tensor_with_shapes(
+            num_tensors=self.num_gemms,
+            shape=[(self.out_features, self.in_features)] * self.num_gemms,
+            quantizer=weight_quantizers[0],
```
timmoon10, line 769: This will subtly break compatibility with FP8 DS. If we're not willing to fully deprecate FP8 DS, maybe we should raise a warning that we are prototyping an experimental design and that results may not be correct.

That sucks, so maybe we should add an option for contiguous weights and set it false by default. That way we can maintain backward compatibility.

zhongbozhu: This will also break FP8 CS too which is still being actively used. We should just skip creating grouped weight for these two recipes.

transformer_engine/pytorch/tensor/storage/grouped_tensor.py
timmoon10: Two approaches (also see https://github.com/NVIDIA/TransformerEngine/pull/2654#discussion_r2790926471):
- Make contiguous weights an opt-in feature. The default behavior should have discrete weights for backward compatibility.
- Raise a warning if running with FP8 DS so that users are aware there may be correctness issues.

transformer_engine/pytorch/module/grouped_linear.py
ksivaman: For now I have disabled the 2 recipe with Grouped weight. But for the next PR, I will make this feature accessible only via an envvar.

PR #2662, opened 2026-02-09
Add multi-precision training support to FSDP script

aagallo:
# Description

This PR adds comprehensive precision parameter support to the FSDP training script, enabling users to configure training with multiple precision formats (FP32, FP16, FP8, MXFP8, NVFP4) via command-line argument. The implementation includes automatic configuration of appropriate dtypes and format-specific recipes for each precision type.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Added precision() type validator function supporting fp32, fp16, fp8, mxfp8, and nvfp4 formats
- Added --precision command-line argument to parse_fsdp_args() with default value "fp8"
- Implemented match statement in train() function to configure precision-based training parameters
- Configured format-specific recipes for each precision type:
  - FP32/FP16: Uses standard PyTorch dtypes with FP8 disabled
  - FP8: Uses DelayedScaling recipe with HYBRID format
  - MXFP8: Uses MXFP8BlockScaling recipe with E4M3 format
  - NVFP4: Uses NVFP4BlockScaling recipe with bfloat16 dtype
- Set appropriate no_fp8 flags based on precision selection
- Updated layer_kwargs["params_dtype"] to use precision-determined dtype
- Imported required recipe classes: MXFP8BlockScaling and NVFP4BlockScaling

# Checklist:

Please reach out to Santosh Bhavani (sbhavani@nvidia.com) for additional context on the work

### DIFF
--- a/tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
+++ b/tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
@@ -1101,10 +1101,6 @@ def _test_nvfp4_partial_cast_matches_full(dp_group) -> None:
     WORLD_RANK = dist.get_rank(dp_group)
     WORLD_SIZE = dist.get_world_size(dp_group)
 
-    available, reason = is_nvfp4_available(return_reason=True)
-    if not available:
-        pytest.skip(reason)
-
     torch.manual_seed(1234)
     device = torch.device("cuda")
     # Shape must be divisible by WORLD_SIZE for even splitting
@@ -1196,6 +1192,11 @@ def _test_nvfp4_partial_cast_matches_full(dp_group) -> None:
 @pytest.mark.parametrize("world_size", [2])
 def test_nvfp4_partial_cast_matches_full(world_size: int) -> None:
     """Launch a distributed job for NVFP4 partial-cast equivalence test."""
+
+    available, reason = is_nvfp4_available(return_reason=True)
+    if not available:
+        pytest.skip(reason)
+
     python_exe = pathlib.Path(sys.executable).resolve()
     current_file = pathlib.Path(__file__).resolve()
     command = [


examples/pytorch/fsdp/fsdp.py
```diff
@@ -112,6 +124,38 @@ def get_layer_args(opts):
...
+            try:
+                values = self.type_converter(values)
+            except (ValueError, TypeError) as e:
+                raise argparse.ArgumentTypeError(f"invalid {self.dest} value: {values}") from e
+
+        setattr(namespace, self.dest, values)
+        setattr(namespace, f"{self.dest}_explicitly_set", True)
+
+
+class StoreTrueExplicitAction(argparse.Action):
```
vthumbe1503: Could you please explaing the point of this class if both no_fp8 and no_fp8_explicitly_set need to be set to true anyway? Why not just use action="store_true" below while parsing no-fp8 argument? 

The implementation above for StoreExplicitAction makes sense for dtype since you need to figure out if you need to set it up explicitly rather than figuring it out based on precision argument.

But i am not sure if I understand the need to set no_fp8_explicitly_set

vthumbe1503 (commented):
Thanks for generalizing fsdp script to support multiple recipes. Have. left a minor comment. LGTM after thats addressed.

examples/pytorch/fsdp/fsdp.py
```diff
@@ -200,15 +284,122 @@ def dist_print(text, all_ranks=False, no_new_line=False):
...
+
+    Returns:
+        Recipe object for FP8 training
+    """
+    match precision_value:
+        case "mxfp8":
+            return MXFP8BlockScaling(fp8_format=Format.E4M3)
+        case "nvfp4":
+            return NVFP4BlockScaling()
+        case _:
```
vthumbe1503: I guess this should be in case precision_value is fp8 and It would make sense to throw "NotImplemented" error if nothing matches

examples/pytorch/fsdp/fsdp.py
aagallo: The main reason is to detect user intention, and distinguish if the user has explicitly passed --no-fp8 in order to detect a conflict with --precision.

Without no_fp8_explicitly_set, we'd incorrectly raise a ValueError even when the user never typed --no-fp8 — just because the default happened to be True. This is why action='store_true' alone isn't sufficient here: it sets the value but gives no way to know how it got set.

aagallo:
@vthumbe1503 Thanks for the feedback!
- for "StoreTrueExplicitAction", I replied in the comment with an explanation. Let me know if you need more information
- for the "precision_value is fp8 recipes", I have updated the code accordingly

I also went through the different feedback cycles. Let me know if there is anything else pending on my side. Thanks again!

examples/pytorch/fsdp/fsdp.py
vthumbe1503: no_fp8_explicitly_set and no_fp8 is always same though right? Is there any case where one will be true and other will be false?

Also why would the default value of no_fp8 be true? It would be false right?

aagallo: You're actually right — with default=False, no_fp8 and no_fp8_explicitly_set are always identical, so StoreTrueExplicitAction is technically redundant for the current default. It was added defensively to guard against a future default=True change and to stay consistent with StoreExplicitAction for --dtype. If the team prefers simplicity, we can drop StoreTrueExplicitAction and use opts.no_fp8 directly in the conflict check — happy to simplify if that's the preference.

aagallo:
@vthumbe1503 I saw that the build is actually failing. Is there anything I need to change on my end? I haven't touched anything else other than the fsdp.py file

examples/pytorch/fsdp/fsdp.py
vthumbe1503: Yes, simplicity would be ideal. Apart from this, the changes look good. Lets get it merged. And the build failure shouldnt be related to your changes. Thanks!

examples/pytorch/fsdp/fsdp.py
```diff
@@ -200,15 +262,132 @@ def dist_print(text, all_ranks=False, no_new_line=False):
...
+            return torch.bfloat16, False, recipe
+        case _:
+            # Fail loudly if validation is bypassed or new preset added without updating this function
+            raise ValueError(
+                f"Invalid precision preset: {precision_value}. "
+                "Supported values: fp32, fp16, fp8, mxfp8, nvfp4"
+            )
+
+
+def get_recipe_for_precision(precision_value):
```
vthumbe1503: Function seems to be unused. Can be removed

examples/pytorch/fsdp/fsdp.py
```diff
@@ -200,15 +262,132 @@ def dist_print(text, all_ranks=False, no_new_line=False):
...
+        )
+
     # Initialize torch.distributed global process group
     dist.init_process_group(backend="nccl")
     torch.cuda.set_device(LOCAL_RANK)
     dist_print(f"WORLD_SIZE = {WORLD_SIZE}")
     torch.manual_seed(opts.seed)
 
+    # Determine final configuration based on precedence rules
+    if opts.precision is None:
```
vthumbe1503: So backward compatibility in this script is not that big of a deal. Having default value of precision as DelayedScaling should be ok and this if condition can be removed to make the code simpler.

vthumbe1503:
@aagallo changes look good to me. Are there any other changes you need to make before merging the PR?

aagallo:
@vthumbe1503 I am mostly battling with the feedback from greptile. If you are good, I am good as well to merge! Thanks for the support

vthumbe1503 (approved):

Issue #2560, opened 2026-01-05
[PyTorch] Support grouped linear op in te.Sequential


timmoon10:
**Is your feature request related to a problem? Please describe.**

Support a grouped linear fusible op, similar to the grouped linear module.

**Describe the solution you'd like**

API should resember grouped linear module.

**Describe alternatives you've considered**

**Additional context**

timmoon10:
This will be closed with https://github.com/NVIDIA/TransformerEngine/pull/2605 or https://github.com/NVIDIA/TransformerEngine/pull/2622.

PR #2605, opened 2026-01-15
[PyTorch] Add fusible ops for MoE

timmoon10:
# Description

This PR adds a grouped linear op, which can be used in the grouped MLP block in Mixture-of-Experts models.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add grouped linear operation
- Add multiply operation

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [ ] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2664, opened 2026-02-09
[PyTorch] Add ops for MoE grouped MLP

timmoon10:
# Description

This PR adds ops needed for the grouped MLP block in Mixture-of-Experts models. In particular, it adds a grouped linear op (similar to the `GroupedLinear` module) and a `ScaledSwiGLU` op. It is the same as https://github.com/NVIDIA/TransformerEngine/pull/2622, but doesn't include the fused ops with experimental kernels. Closes https://github.com/NVIDIA/TransformerEngine/issues/2560.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add grouped linear op
- Add scaled SwiGLU op
- Handle edge cases in `noop_cat` function

# Checklist:

### DIFF
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -1755,6 +1755,7 @@ class TestBasicOps:
         assert_close_grads(x_test, x_ref, **tols)
 
     def test_interleaved_swiglu(self):
+        """SwiGLU with block interleaved input format"""
         self.test_swiglu(
             out_shape=(32, 192),
             dtype=torch.float32,
@@ -1772,6 +1773,7 @@ class TestBasicOps:
         self,
         *,
         out_shape: Iterable[int] = (32, 32),
+        glu_interleave_size: Optional[int] = None,
         dtype: torch.dtype,
         device: torch.device = "cuda",
         quantization: Optional[str],
@@ -1780,7 +1782,7 @@ class TestBasicOps:
         limit: float = 0.75,
         alpha: float = 1.702,
     ):
-        # Test SwiGLU variant used in GPT OSS.
+        """SwiGLU variant used in GPT-OSS"""
         # Tensor dimensions
         in_shape = list(out_shape)
         in_shape[-1] *= 2
@@ -1805,7 +1807,17 @@ class TestBasicOps:
         )
 
         # Plain PyTorch implementation
-        x_glu, x_linear = x_ref.chunk(2, dim=-1)
+        x = x_ref
+        if glu_interleave_size is not None:
+            x = x.reshape(
+                *in_shape[:-1],
+                in_shape[-1] // (2 * glu_interleave_size),
+                2,
+                glu_interleave_size,
+            )
+            x = x.transpose(-3, -2)
+            x = x.reshape(in_shape)
+        x_glu, x_linear = x.chunk(2, dim=-1)
         x_glu = x_glu.clamp(min=None, max=limit)
         x_linear = x_linear.clamp(min=-limit, max=limit)
         out_glu = x_glu * torch.sigmoid(alpha * x_glu)
@@ -1817,7 +1829,11 @@ class TestBasicOps:
 
         forward = te_ops.Sequential(
             te_ops.Quantize(forward=False, backward=quantize_backward),
-            te_ops.ClampedSwiGLU(limit=limit, alpha=alpha),
+            te_ops.ClampedSwiGLU(
+                limit=limit,
+                alpha=alpha,
+                glu_interleave_size=glu_interleave_size,
+            ),
             te_ops.Quantize(forward=quantize_forward, backward=False),
         )
         with te.autocast(enabled=quantized_compute, recipe=recipe):
@@ -1833,10 +1849,19 @@ class TestBasicOps:
             tols = dtype_tols(tex.DType.kFloat8E4M3)
 
         # Check results
-        y_test = y_test.to(dtype=torch.float64, device="cpu")
-        dx_test = x_test.grad.to(dtype=torch.float64, device="cpu")
-        torch.testing.assert_close(y_test, y_ref, **tols)
-        torch.testing.assert_close(dx_test, x_ref.grad, **tols)
+        assert_close(y_test, y_ref, **tols)
+        assert_close_grads(x_test, x_ref, **tols)
+
+    def test_interleaved_clamped_swiglu(self):
+        """GPT-OSS SwiGLU with block interleaved input format"""
+        self.test_clamped_swiglu(
+            out_shape=(32, 192),
+            dtype=torch.float32,
+            quantization=None,
+            quantize_forward=False,
+            quantize_backward=False,
+            glu_interleave_size=32,
+        )
 
     @pytest.mark.parametrize("scale", (1, 0, -2.5, 3.5))
     @pytest.mark.parametrize("shape", ((), (1, 13), (4, 4, 2)))
@@ -2104,20 +2129,19 @@ class TestBasicOps:
                     assert b_test.grad is None
 
     @pytest.mark.parametrize("in_shape", ((71, 192), (5, 7, 128)))
-    @pytest.mark.parametrize("glu_interleave_size", (None, 32))
     @pytest.mark.parametrize("input_requires_grad", (False, True))
     @pytest.mark.parametrize("scales_requires_grad", (False, True))
     def test_scaled_swiglu(
         self,
         *,
         in_shape: Iterable[int],
-        glu_interleave_size: Optional[int],
+        glu_interleave_size: Optional[int] = None,
         dtype: torch.dtype = torch.float32,
         device: torch.device = "cuda",
         input_requires_grad: bool,
         scales_requires_grad: bool,
     ) -> None:
-        """Multiply two tensors"""
+        """SwiGLU with post-scale"""
 
         # Tensor dims
         out_shape = list(in_shape)
@@ -2169,17 +2193,18 @@ class TestBasicOps:
         # Check results
         tols = dtype_tols(dtype)
         y_test = y_test.to(dtype=torch.float64, device="cpu")
-        torch.testing.assert_close(y_test, y_ref, **tols)
-        if input_requires_grad:
-            dx_test = x_test.grad.to(dtype=torch.float64, device="cpu")
-            torch.testing.assert_close(dx_test, x_ref.grad, **tols)
-        else:
-            assert x_test.grad is None
-        if scales_requires_grad:
-            ds_test = scales_test.grad.to(dtype=torch.float64, device="cpu")
-            torch.testing.assert_close(ds_test, scales_ref.grad, **tols)
-        else:
-            assert scales_test.grad is None
+        assert_close(y_test, y_ref, **tols)
+        assert_close_grads(x_test, x_ref, **tols)
+        assert_close_grads(scales_test, scales_ref, **tols)
+
+    def test_interleaved_scaled_swiglu(self):
+        """SwiGLU with post-scale and block interleaved input format"""
+        self.test_scaled_swiglu(
+            in_shape=(32, 192),
+            glu_interleave_size=32,
+            input_requires_grad=True,
+            scales_requires_grad=True,
+        )
 
 
 class TestFusedOps:
--- a/transformer_engine/pytorch/ops/basic/swiglu.py
+++ b/transformer_engine/pytorch/ops/basic/swiglu.py
@@ -58,15 +58,22 @@ class SwiGLU(BasicOperation):
     glu_interleave_size : int, optional
         When set, the GLU activations will use a block interleaved
         format. Instead of interpreting the input tensor as a
-        concatenation of gates and linear units, it will be
-        interpreted as alternating blocks of gates and linear units.
-        This data format is experimental and primarily intended to
-        enable advanced fused kernels.
+        concatenation of gates and linear units (e.g.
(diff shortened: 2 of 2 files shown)


transformer_engine/pytorch/ops/basic/swiglu.py
```diff
@@ -0,0 +1,415 @@
...
+                grad_input = grad_input.transpose(1, 2).contiguous()
+                grad_input = grad_input.view(shape)
+
+        # Compute scales grad by recomputing SwiGLU
+        grad_extra_input = None
+        if ctx.extra_input_requires_grad:
+            swiglu_out = tex.swiglu(swiglu_in, None)
+            grad_extra_input = torch.linalg.vecdot(swiglu_out, grad_output)
+
+        # Clear input tensor if possible
```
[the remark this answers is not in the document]
timmoon10: [`torch.linalg.vecdot`](https://docs.pytorch.org/docs/stable/generated/torch.linalg.vecdot.html) computes the dot product over the last tensor dim and leaves all other dims unchanged. The existing code is correct and the suggested code exposes an unnecessary memory-bound kernel. See https://github.com/NVIDIA/TransformerEngine/pull/2622#discussion_r2784590137.

tests/pytorch/test_fusible_ops.py
```diff
@@ -1924,6 +1967,231 @@ def test_dropout(
...
+        input_requires_grad: bool,
+        weight_requires_grad: bool,
+    ) -> None:
+        """Grouped GEMM"""
+
+        # Split sizes
+        split_sizes = [split_alignment * i for i in range(group_size)]
+        random.shuffle(split_sizes)
+        split_sizes = torch.tensor(split_sizes, dtype=torch.int, device=device)
+
```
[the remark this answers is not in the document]
timmoon10: These split points are chosen this way precisely to test the edge case you mention. You've been making a lot of mistakes and nits while reviewing this PR, please be more thoughtful and substantive in the future.

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -0,0 +1,700 @@
...
+            dtype = torch.get_autocast_dtype("cuda")
+        else:
+            dtype = self.weight0.dtype
+
+        # Extract split sizes from extra input
+        split_sizes = basic_op_extra_inputs[0][0]
+        split_sizes_int = [int(s) for s in split_sizes.tolist()]
+        if len(split_sizes_int) != num_groups:
+            raise ValueError(f"Expected {num_groups} splits, but got {len(split_sizes_int)}.")
+
```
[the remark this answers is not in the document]
timmoon10: This proposed check requires a GPU-CPU synchronization, and therefore breaks compatibility with CUDA Graphs. We currently copy the split points to CPU, but that will be removed once we support the grouped tensor API from https://github.com/NVIDIA/TransformerEngine/pull/2654.

tests/pytorch/test_fusible_ops.py
```diff
@@ -172,6 +173,29 @@ def make_reference_and_test_tensors(
     return ref, test
 
 
+def assert_close(
```
ptrendx: Shouldn't this be in some more generic place?

tests/pytorch/test_fusible_ops.py
```diff
@@ -1727,10 +1762,18 @@ def test_swiglu(
...
 
         # Check results
-        y_test = y_test.to(dtype=torch.float64, device="cpu")
-        dx_test = x_test.grad.to(dtype=torch.float64, device="cpu")
-        torch.testing.assert_close(y_test, y_ref, **tols)
-        torch.testing.assert_close(dx_test, x_ref.grad, **tols)
+        assert_close(y_test, y_ref, **tols)
+        assert_close(x_test.grad, x_ref.grad, **tols)
+
+    def test_interleaved_swiglu(self):
```
ptrendx, line 1757: Why a separate test rather than a parameter in the other test? Does this only work for specific scenarios?

transformer_engine/pytorch/module/_common.py
```diff
@@ -125,14 +125,19 @@ def forward(
...
+        # Out-of-place concatenation when view tensors have different storage
+        # Note: This works around an edge case with the split_quantize
+        # function, which might allocate a buffer and construct
+        # subviews. However, in order to reduce CPU overheads, these
+        # views are configured manually outside of PyTorch. PyTorch
+        # doesn't know these views share the same memory, and it
+        # blocks us from reconstructing the full tensor because it
+        # thinks we are accessing out-of-bounds memory.
+        if tensors[0].untyped_storage().nbytes() < out_shape[dim] * data_ptr_stride:
+            return torch.cat(tensors, dim=dim)
```
ptrendx, line 129: Should this be done before this big for loop to short circuit it?

transformer_engine/pytorch/ops/basic/activation.py
```diff
@@ -27,8 +27,6 @@
     "SReLU",
     "SReGLU",
     "SiLU",
-    "SwiGLU",
-    "ClampedSwiGLU",
 ]
```
ptrendx, line 30: This looks like a breaking change in the API? Or we assume that people would always use those ops from the top level?

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -0,0 +1,700 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""Fusible operation for bias."""
```
ptrendx: Not really :-).

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -0,0 +1,700 @@
...
+            f"and {self.num_extra_outputs} extra tensor outputs. "
+            "It overrides `fuser_backward` instead of `op_backward`."
+        )
+
+    def fuser_forward(
+        self,
+        basic_op_ctxs: list[OperationContext],
+        input_: torch.Tensor,
+        *,
+        basic_op_extra_inputs: list[tuple[torch.Tensor, ...]],
```
ptrendx, line 468: To be honest I don't know how I feel about the decision to deal with splits as the extra input. This basically makes putting a router/permutation into the same sequential as the grouped linear impossible, right?

ptrendx: The way I thought about it would be to handle tuple of Tensors instead. If we want to support attention in sequential then we would need to go that route anyway (in order to be able to express what is currently the fp8_mha nicely).

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -0,0 +1,700 @@
...
+    canonicalize_dtype,
+    clear_tensor_data,
+    devices_match,
+    round_up_to_nearest_multiple,
+)
+from .._common import is_quantized_tensor, maybe_dequantize
+from ..op import BasicOperation, OperationContext
+
+
+class GroupedLinear(BasicOperation):
```
ptrendx, line 37: Please mark it experimental for now. I think we will need to iterate on this design.

transformer_engine/pytorch/ops/basic/swiglu.py
```diff
@@ -0,0 +1,415 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""Fusible operation for multiplying with extra input tensor."""
```
ptrendx: Not really.

transformer_engine/pytorch/ops/basic/swiglu.py
```diff
@@ -0,0 +1,415 @@
...
+from ...tensor import Float8CurrentScalingQuantizer, Quantizer
+from ...utils import clear_tensor_data
+from ..op import BasicOperation, OperationContext
+from .._common import maybe_dequantize
+
+__all__ = ["SwiGLU", "ClampedSwiGLU", "ScaledSwiGLU"]
+
+
+class SwiGLU(BasicOperation):
+    r"""Swish gated linear unit
```
ptrendx, line 24: The doc should include the information about the interleaving parameter and what it means.

transformer_engine/pytorch/ops/basic/swiglu.py
```diff
@@ -0,0 +1,415 @@
...
+    from GPT OSS implementation where the gates/pre-activations are assumed to be interleaved in the input tensor.
+
+    Parameters
+    ----------
+    limit : float
+        The clamp limit.
+    alpha : float
+        The scaling factor for the sigmoid function used in the activation.
+    cache_quantized_input : bool, default = ``False``
+        Quantize input tensor when caching for use in the backward pass.
```
ptrendx, line 208: We will need similar interleaving support here pretty soon.

transformer_engine/pytorch/ops/basic/swiglu.py
```diff
@@ -0,0 +1,415 @@
...
+        return dx, ()
+
+
+class ScaledSwiGLU(BasicOperation):
+    r"""SwiGLU with post-scaling
+
+    If the SwiGLU output has shape ``(d_1, ..., d_n)``, it is
+    multiplied with an extra input tensor of shape
+    ``(d_1, ..., d_{n-1})``.
+
```
ptrendx, line 352: Also please add the docs for the interleaving. BTW - is the "cache quantized input" parameter only for Lora or something that it is only in the regular swiglu and nowhere else?

transformer_engine/pytorch/ops/basic/swiglu.py
```diff
@@ -0,0 +1,415 @@
...
+                shape[-1] // (2 * self.glu_interleave_size),
+                2,
+                self.glu_interleave_size,
+            )
+            swiglu_in = swiglu_in.transpose(1, 2).contiguous()
+            swiglu_in = swiglu_in.view(shape)
+
+        # Compute scaled SwiGLU
+        swiglu_out = tex.swiglu(swiglu_in, None)
+        out = swiglu_out * scales.unsqueeze(-1)
```
ptrendx, line 424: Considering it is implemented with 2 kernels anyway, what is the benefit of having this operation here? I would prefer to have the ScaleWithExtraInput basic op or something like that instead.

tests/pytorch/test_fusible_ops.py
timmoon10: None of the other activations support interleaving since it's so hacky and specific to the fused GEMM + SwiGLU kernel.

transformer_engine/pytorch/ops/basic/activation.py
timmoon10: Accessing via `te.ops` or `te.ops.basic` is definitely the best practice. I don't see any downstream code accessing these via `te.ops.basic.activation`: https://github.com/search?q=ops.basic.activation&type=code
Megatron-LM accesses `SwiGLU` via `te.ops`:
https://github.com/NVIDIA/Megatron-LM/blob/24515084a6395bc4c30e20368d47d073868e95e6/megatron/core/extensions/transformer_engine.py#L2076
I don't see `ClampedSwiGLU` being accessed anywhere.

transformer_engine/pytorch/ops/basic/grouped_linear.py
timmoon10: It's hard to make this work well with a sequential structure.
- Supporting ops with tuple inputs/outputs is possible but messy. We'll need to add more complicated logic to handle grads properly. If the number of tensors doesn't match between two ops, maybe we could use extra inputs/outputs to make up the difference.
- Passing a tuple will allow us to pass split points to FC1, but not FC2. I suppose we could make this work if we make grouped-ness a persistent property, and make sure the other ops support grouped implementations. I worry that we would be at risk of overfitting for MoEs.

transformer_engine/pytorch/ops/basic/swiglu.py
timmoon10: It was an experimental feature for LoRA that we never really ended up using.

transformer_engine/pytorch/ops/basic/swiglu.py
timmoon10: This was the approach I used in my initial implementation (https://github.com/NVIDIA/TransformerEngine/pull/2605), but it's not compatible with the fused GEMM + SwiGLU kernel (https://github.com/NVIDIA/cudnn-frontend/blob/main/python/cudnn/grouped_gemm/grouped_gemm_swiglu/api.py). If we have a standalone scale op, then we need to cache its input for the backward pass. However, the fused kernel assumes you are doing activation recompute and it only outputs the SwiGLU input and scale output. Rather than intertwining the implementations of the SwiGLU and scale to support activation recompute, I just implemented a new op that does it explicitly

transformer_engine/pytorch/ops/basic/swiglu.py
timmoon10: Hopefully the fast kernels team has mercy on us.

tests/pytorch/test_fusible_ops.py
timmoon10: Moved to `utils.py`.

tests/pytorch/test_fusible_ops.py
timmoon10: Also, separating the test like this is good practice. Just adding more parameters to the SwiGLU test results in exponential growth in the number of cases, while creating standalone tests results in linear growth.

PR #2666, opened 2026-02-10
[pyTorch] Replace the make_empty implementation to use C++ implementation

ptrendx:
# Description

This PR unifies the implementation of the QuantizedTensor creation by using the C++ implementation of the create_tensor.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Replaced the Python implementations of the make_empty with the calls to C++ create_tensor

# Checklist:


transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -272,13 +274,26 @@ def make_empty(
...
         dtype: torch.dtype = torch.float32,
-        device: Optional[torch.device] = None,
+        device: Optional[Union[torch.device, str]] = None,
+        requires_grad: bool = False,
+        pin_memory: bool = False,
     ) -> QuantizedTensor:
         """Construct quantized tensor with uninitialized data"""
-        raise NotImplementedError(
-            f"{self.__class__.__name__} class does not implement make_empty function, "
-            "required for construction of unintialized quantized tensor"
```
negvet, line 319: This clear `NotImplementedError` is beneficial for custom quantizers that do not override `make_empty()`.
Now, if custom quantizer does not have `make_empty()`, then it will fail at the C++ `convert_quantizer`, because there is no registered C++ converter. C++ failure with `NVTE_ERROR("Unexpected type for quantizer")` is not as clear as `NotImplementedError`.

What about making C++ error more clear or even better add a check at base `Quantizer.make_empty`:
```
def make_empty(...):
    if getattr(self, "custom", False):
        raise NotImplementedError(
            f"{self.__class__.__name__} does not implement make_empty"
        )
    # ... existing C++ path ...
```

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -284,13 +286,26 @@ def make_empty(
...
+        device = torch.device(device)
+        result = tex.create_empty_quantized_tensor(
+            self,
+            list(shape),
+            dtype,
+            device,
+            pin_memory,
         )
+        if requires_grad:
+            result.requires_grad_(True)
```
vthumbe1503, line 342: Doing this in C++ itslef might be faster, since we are anyway going to call the QuantizedTensor.__new__ method with requires_grad argument. Calling this from python for custom quantized tensor has severe python overheads

vthumbe1503: But I see it can get quite complicated since we might have to change the create_tensor API to accept the requires_grad argument.

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -311,13 +313,34 @@ def make_empty(
...
```
vthumbe1503, line 342: It might make sense for the tex API to accept requires_grad as an argument as well considering CPU overheads. Given that we are passing things like pin_memory, device which are torch::TensorOptions, we might as well add requires_grad in the picture which is also an attribute of torch::TensorOptions.

vthumbe1503 (dismissed):
LGTM

vthumbe1503 (approved):

PR #2672, opened 2026-02-11
[JAX] TE Permutation integration to Maxtext

tdophung:
# Description

Changes needed on TE side to make maxtext integration works

Issue # 2585

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Masking out padding tokens on each local EP (will be used in local permute step)
- Pass along split_sizes and sorted_indices in residual of sort_chunks_by_index (local_permute) to avoid mismatch in size issue during tracing when EP>1

# Checklist:


tdophung:
This PR contain changes cherry-picked from #2651 . I can wait until this gets merged and then merge mine, but if my PR is needed more urgently, happy to remove the cherry picked change

transformer_engine/common/triton/permutation.py
```diff
@@ -578,6 +585,11 @@ def _make_chunk_sort_map_kernel(
     ).to(tl.int32)
     output_pre_split_sizes = tl.where(load_split_offset < output_chunk_idx, output_split_sizes, 0)
     dst_row = tl.sum(output_pre_split_sizes) + in_chunk_offset
+
+    # For tokens beyond the valid range (pid >= total_valid_tokens),
+    # use identity mapping to avoid corrupting valid data
+    dst_row = tl.where(pid < total_valid_tokens, dst_row, pid)
```
jberchtold-nvidia, line 591: Do we need to copy the tokens/overallocated space beyond the  valid range to the output? Or can we skip those and leave the output uninitialized since they are beyond the valid range and won't be used downstream?

Maybe this could be achieved by using a `mask` arg to load/store that is dependent  on `tl.sum(input_split_sizes)`?
https://triton-lang.org/main/python-api/generated/triton.language.store.html#triton.language.store

transformer_engine/jax/permutation.py
```diff
@@ -642,7 +641,13 @@ def _sort_chunks_by_index_bwd_rule(
         is_forward=False,
     )
 
-    return (inp_grad,)
+    # Return gradients for all inputs: (inp, split_sizes, sorted_indices)
```
jberchtold-nvidia, line 644: I'm not following why the split_sizes and sorted_indices can't be non-differentiable args, can you let me know  the reasoning. Thanks!

transformer_engine/common/triton/permutation.py
tdophung: your approach would mean we use the mask in make_chunk_sort_map to:

```
    dst_row = tl.sum(output_pre_split_sizes) + in_chunk_offset
    tl.store(dst_rows_ptr + pid, dst_row, mask=pid < total_valid_tokens)
```

then in sort_chunks_by_index, we guard the pid to exit early if it is not in the valid range:
```
    pid_t = tl.program_id(0)
    if pid_t >= total_valid_tokens:
        return
```
(`total_valid_tokens = tl.sum(input_split_sizes)`)

Is the above what you mean?

transformer_engine/jax/permutation.py
tdophung: `_te_local_permute` (in moe.py in maxtext) computes `split_sizes` from `all_shards_tokens_per_expert` via `jax.lax.dynamic_slice` (to support traced shard_index from jax.lax.axis_index for the raffed a2a). --->  split_sizes is a traced array when the function is jitted. 
There fore it  error out, so I have to make it NOT a nondiff arg

tdophung: previously I did not catch this as split size is likely not computed in jit in test_permutation.py in TE

transformer_engine/common/triton/permutation.py
tdophung: The approach above would also mean we need to somehow pass total_valid_tokens to sort_chunks_by_index to be able to guard it.

jberchtold-nvidia: Yes, that it was I had in mind to reduce read/writes, but I hadn't realized this change would affect the multiple kernel passes here and complicate the logic. Let's stick with your current approach

transformer_engine/jax/permutation.py
jberchtold-nvidia: Sounds good, thanks for the explanation!

transformer_engine/common/triton/permutation.py
tdophung: I think it could be a good change to use the mask because it would reduce the number of accesses to global mem to reach the identity-mapped phantom tokens. Performance wise it might be better to use the mask + piid guard approach
The draw back would be that it couples these 2 functions together + having to change a lot of the code to plumb thr total_valid_tokens through to sort_chunks_by_index, including pytorch side.

So I will leave it as is, and we can revisit this if it turns out to be a perf bottleneck

PR #2676, opened 2026-02-12
[PyTorch] Add dtype information to QuantizedTensorStorage class

ptrendx:
# Description

This PR adds the fake dtype information to the QuantizedTensorStorage class. This eliminates the need to guess the correct type for dequantize, as was the case in the distributed.py, and it eliminates the unintentional dequantization to FP32 when calling dequantize() on the Storage class with no dtype argument. 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Added the _dtype field to the QuantizedTensorStorage class
- Modified the dequantize call to use that new field when calling dequantize with no arguments
- Removed guessing of the dtype from distributed.py

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/quantized_tensor.py
+++ b/transformer_engine/pytorch/quantized_tensor.py
@@ -675,6 +675,7 @@ class QuantizedTensor(torch.Tensor):
         shape = shape if shape is not None else tensor.shape
         dtype = dtype if dtype is not None else tensor.dtype
         kwargs = tensor.get_metadata()
+        kwargs["fake_dtype"] = dtype
         return cls(shape=shape, dtype=dtype, requires_grad=requires_grad, **kwargs)
 
     def to_dtype(self, dtype: torch.dtype) -> QuantizedTensor:


transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -367,9 +356,12 @@ def __new__(
         shape: Iterable[int],
         dtype: torch.dtype,
         *,
+        fake_dtype: Optional[torch.dtype] = None,
```
timmoon10, line 371: Isn't this redundant with the `dtype` kwarg?

transformer_engine/pytorch/tensor/storage/float8_tensor_storage.py
```diff
@@ -75,14 +75,16 @@ def __new__(
         data: Optional[torch.Tensor],
         fp8_scale_inv: torch.Tensor,
         fp8_dtype: TE_DType,
+        fake_dtype: Optional[torch.dtype] = None,
```
timmoon10, line 78: I'd prefer to just name it `dtype` since `QuantizedTensor` is already using that name in its constructor.
```suggestion
        dtype: Optional[torch.dtype] = None,
```

transformer_engine/pytorch/tensor/storage/nvfp4_tensor_storage.py
```diff
@@ -106,10 +106,14 @@ def __new__(
...
         with_gemm_swizzled_scales: bool,
         *args,
+        fake_dtype: Optional[torch.dtype] = None,
         **kwargs,
     ):
-
-        instance = super().__new__(cls, *args, **kwargs)
+        if cls is NVFP4TensorStorage:
+            instance = object.__new__(cls)
+            instance._dtype = fake_dtype if fake_dtype is not None else torch.float32
```
timmoon10, line 114: NVFP4 compatibility with FP32 is not very good, so maybe it would be worth changing the default to BF16? The downside is that this default is different from standard PyTorch and from the other quantized tensors.
```suggestion
            instance._dtype = fake_dtype if fake_dtype is not None else torch.bfloat16
```

timmoon10 (approved):
Overall this is a big improvement. I have some naming nits.

transformer_engine/pytorch/quantized_tensor.py
ptrendx: This is mostly to avoid issues with MRO and still have fairly straightforward constructors for the Storage classes.

ptrendx: Also just noticed that the make_like call would be problematic there otherwise - we want to include the fake_dtype in get_metadata call, but if it was named dtype it would clash with the dtype that we pass directly in make_like.

transformer_engine/pytorch/tensor/storage/nvfp4_tensor_storage.py
ptrendx: Hmmm, I don't think we should do that - sure, it could result in some crashes when the tensor is quantized and then dequantized without the proper type set, but I think that is better than a silent unexpected dtype being set.

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -37,6 +37,7 @@ class QuantizedTensorStorage:
     XTensor should only implement the functionality needed
     to behave like regular torch.Tensor (like __torch_dispatch__)."""
 
+    _dtype: torch.dtype
     _quantizer: Optional[Quantizer]
```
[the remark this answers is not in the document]
ptrendx: Storage classes are internal classes and cannot be used in the user facing pieces, like the checkpoint, so there is no possibility of the break here.

PR #2678, opened 2026-02-13
[PyTorch] Make grouped weights opt-in

ksivaman:
# Description

Follow-up to #2654 that makes grouped weights opt-in via an envvar. The variable name is inaccurate for now but impending work will convert this opt-in option to use only a single-parameter.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Make grouped weights opt-in.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/grouped_linear.py
+++ b/transformer_engine/pytorch/module/grouped_linear.py
@@ -795,7 +795,7 @@ class GroupedLinear(TransformerEngineBaseModule):
     def reset_parameters(self, defer_init=False):
         super().reset_parameters(defer_init=defer_init)
         # Grouped tensor weights is an opt-in feature.
-        if bool(int(os.getenv("NVTE_GROUPED_LINEAR_SINGLE_PARAM", "0"))):
+        if bool(int(os.getenv("NVTE_ALLOC_CONTIGUOUS_GROUPED_LINEAR_WEIGHTS", "0"))):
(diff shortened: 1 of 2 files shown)


transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -793,7 +794,9 @@ def make_grouped_weights(self, defer_init=False) -> None:
 
     def reset_parameters(self, defer_init=False):
         super().reset_parameters(defer_init=defer_init)
-        self.make_grouped_weights(defer_init=defer_init)
+        # Grouped tensor weights is an opt-in feature.
+        if bool(int(os.getenv("NVTE_GROUPED_LINEAR_SINGLE_PARAM", "0"))):
+            self.make_grouped_weights(defer_init=defer_init)
```
timmoon10: We're not actually exposing a single param, but rather we're sneakily allocating the weights to be contiguous.
```suggestion
        if bool(int(os.getenv("NVTE_ALLOC_CONTIGUOUS_GROUPED_LINEAR_WEIGHTS", "0"))):
            self.make_grouped_weights(defer_init=defer_init)
```
Constructor kwargs are also much more Pythonic than these hidden envvars. It's not terrible for now since this is an internal implementation detail, but we should expose a proper kwarg when we actually add the option for a single param.

timmoon10 (approved):

PR #2680, opened 2026-02-13
[JAX] Integrate BF16 Grouped GEMM with on-device group sizes

### LINKED ISSUE #2659: [Core] TE common nvte_grouped_gemm treats output layout as column-wise instead of rowwise
**Describe the bug**

Output layout is assumed to be column-wise instead of row-wise in the nvte_grouped_gemm API

Please list *minimal* steps or code snippet for us to be able to reproduce the bug.

A  helpful guide on on how to craft a minimal bug report  http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports. 

**Expected behavior**

**Environment overview (please complete the following information)**

 - Environment location: [Bare-metal, Docker, Cloud(specify cloud provider - AWS, Azure, GCP, Collab)]
 - Method of Transformer Engine install: [pip install or from source]. Please specify exact commands you used to install.
 - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**

Otherwise, please provide:
- OS version
- PyTorch version
- Python version
- Transformer Engine version
- CUDA version
- CUDNN version

**Device details**
- GPU model

**Additional context**

jberchtold-nvidia:
# Description

Integrate new BF16 grouped GEMM from TE common/cuBLASLt that supports on-device group sizes without a D2H memcpy and stream sync. This grouped GEMM is faster and CUDA-graph safe.

Also fixes #2659

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Added a new cuda-graph-safe grouped GEMM to TE/JAX that is automatically used as the backend when the input data is bf16 (no scaling recipe) and not bias is required.
- Exposed a new `make_ragged_dot_cls` for easy integration into existing models. This will be most useful when quantization is supported and storing recipe state is required

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/gemm.py
+++ b/transformer_engine/jax/cpp_extensions/gemm.py
@@ -2205,9 +2205,7 @@ def grouped_gemm(
         " and padded with zeros to not affect the result of the MoE block."
     )
 
-    use_v2_ffi = _can_use_v2_grouped_gemm(
-        scaling_mode, lhs_data.dtype, has_bias
-    )
+    use_v2_ffi = _can_use_v2_grouped_gemm(scaling_mode, lhs_data.dtype, has_bias)
     if use_v2_ffi:
         num_gemms = group_sizes.shape[0]
         additional_arg_0 = jnp.ones((num_gemms,), jnp.float32)  # alpha
--- a/transformer_engine/jax/csrc/extensions/gemm.cpp
+++ b/transformer_engine/jax/csrc/extensions/gemm.cpp
@@ -560,12 +560,13 @@ JAXX_GroupedTensorWrapper make_grouped_tensor(Buffer_Type const &data,
 }
 
 // This FFI is EXPERIMENTAL and subject to change without deprecation, intended for use in JAX's internal implementation of grouped GEMM.
-Error_Type GroupedGemmV2FFI(
-    cudaStream_t stream, Buffer_Type lhs_data, Buffer_Type lhs_sinv, Buffer_Type rhs_data,
-    Buffer_Type rhs_sinv, Buffer_Type bias, Buffer_Type group_sizes, Buffer_Type alpha,
-    Buffer_Type beta, Result_Type output, Result_Type cublas_workspace, Result_Type setup_workspace,
-    Result_Type int64_workspace, size_t m, size_t n, size_t k, bool lhs_is_trans, bool rhs_is_trans,
-    JAXX_Scaling_Mode scaling_mode, bool is_grouped_dense_wgrad) {
+Error_Type GroupedGemmV2FFI(cudaStream_t stream, Buffer_Type lhs_data, Buffer_Type lhs_sinv,
+                            Buffer_Type rhs_data, Buffer_Type rhs_sinv, Buffer_Type bias,
+                            Buffer_Type group_sizes, Buffer_Type alpha, Buffer_Type beta,
+                            Result_Type output, Result_Type cublas_workspace,
+                            Result_Type setup_workspace, Result_Type int64_workspace, size_t m,
+                            size_t n, size_t k, bool lhs_is_trans, bool rhs_is_trans,
+                            JAXX_Scaling_Mode scaling_mode, bool is_grouped_dense_wgrad) {
   // Notes on matrix layouts and transpose:
   // Jax uses row-major data_layout, on entering this function, each input matrix pair:
   //   A: row-major [m, k] for N - [k, m] for T


transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -409,6 +409,397 @@ XLA_FFI_DEFINE_HANDLER_SYMBOL(GroupedGemmD2HGroupSizesHandler, GroupedGemmD2HGro
                                   .Ret<Buffer_Type>()      // dummy_output
                                   .Attr<int64_t>("num_gemms"));
 
+class JAXX_GroupedTensorWrapper {
```
jberchtold-nvidia, line 412: There is a PyTorch refactor PR that also adds a similar GroupedTensorWrapper struct to TE/common. I will move to that in the future once it is merged, but I don't want to block this PR on that currently and the logic is pretty straightforward and should be a small change to switch.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -535,6 +553,10 @@ inline size_t grouped_gemm_setup_workspace_size(size_t num_tensors) {
 
 }  // namespace
 
+size_t nvte_grouped_gemm_setup_workspace_size(size_t num_tensors) {
+  return grouped_gemm_setup_workspace_size(num_tensors);
```
phu0ngng: Register `nvte_` API with `NVTE_API_CALL`.

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -409,6 +409,383 @@ XLA_FFI_DEFINE_HANDLER_SYMBOL(GroupedGemmD2HGroupSizesHandler, GroupedGemmD2HGro
...
+                                  .Ret<Buffer_Type>()      // int64_workspace
+                                  .Attr<int64_t>("M")
+                                  .Attr<int64_t>("N")
+                                  .Attr<int64_t>("K")
+                                  .Attr<bool>("lhs_is_trans")
+                                  .Attr<bool>("rhs_is_trans")
+                                  .Attr<JAXX_Scaling_Mode>("scaling_mode")
+                                  .Attr<bool>("has_bias")
+                                  .Attr<bool>("is_grouped_dense_wgrad")
+                                  .Attr<bool>("use_async_d2h_group_sizes"),
```
phu0ngng: Remove these three attributes as they are not needed.

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -409,6 +409,383 @@ XLA_FFI_DEFINE_HANDLER_SYMBOL(GroupedGemmD2HGroupSizesHandler, GroupedGemmD2HGro
...
+  nvte_grouped_gemm(rhs_tensor, rhs_is_trans, lhs_tensor, lhs_is_trans, nullptr, out_tensor,
+                    alpha_tensor.data(), beta_tensor.data(), workspace_setup.data(),
+                    workspace_cublas.data(),
+                    nullptr,  // config (use defaults)
+                    stream);
+
+  return ffi_with_cuda_error_check();
+}
+
+XLA_FFI_DEFINE_HANDLER_SYMBOL(GroupedGemmCudaGraphableHandler, GroupedGemmCudaGraphableFFI,
```
phu0ngng: Should we simply call it v2 for now and eventually rename it back to GroupedGEMM? 

Besides, if it is possible, could we mark it as an experimental API and subject to changes?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1910,6 +1983,28 @@ def grouped_gemm_copy_group_sizes(
...
+    # feature-compatible with the main branch.
+    # Bias can be supported in a kernel or in pure-JAX in the future.
+
+    try:
+        get_grouped_gemm_setup_workspace_size(1)
+    except RuntimeError as e:
+        if "cublas" in str(e).lower():
+            # If the workspace size function is not available, it means the cuda-graphable implementation is not available.
+            return False
+        raise e
```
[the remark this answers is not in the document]
phu0ngng: I'm not sure about this helper. 

Why don't we simply check for NO_SCALING, then go with the v2?

Also, we should be able to check it inside primitive methods instead of passing it as an argument so that when we eventually remove the multi-stream approach, we don't need to change the primitive inputs.

transformer_engine/jax/flax/module.py
```diff
@@ -1438,3 +1441,26 @@ def te_dot_general(generate_quantizer_set, x, kernel, dims, **kwargs):
...
+            kernel,
+            group_sizes=group_sizes,
+            contracting_dims=((1,), (1,)),
+            quantizer_set=quantizer_set,
+        )
+        return out
+
+    return wrap_function_in_te_state_module(
+        te_grouped_dot_general, quantization_recipe, "ragged_dot"
+    )()
```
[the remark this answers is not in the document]
phu0ngng: Should we call it grouped_dense here to be consistent?

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -329,6 +329,31 @@ void nvte_multi_tensor_gemm(const NVTETensor *A, const NVTETensor *B, NVTETensor
...
+ * The setup workspace stores pointer arrays and per-matrix dimension arrays used
+ * by the grouped GEMM kernel. Its size depends only on the number of tensors (GEMMs)
+ * in the group and is independent of matrix dimensions.
+ *
+ * Pass the result as the size of the workspace_setup tensor in nvte_grouped_gemm.
+ *
+ *  \param[in] num_tensors  Number of tensors (GEMMs) in the group.
+ *  \return Required size in bytes for workspace_setup.
+ */
+size_t nvte_grouped_gemm_setup_workspace_size(size_t num_tensors);
```
phu0ngng: Nit: could we add `get` i.e., `nvte_get_grouped_gemm_setup_workspace_size`? :D

transformer_engine/jax/cpp_extensions/gemm.py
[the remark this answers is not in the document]
jberchtold-nvidia: > Why don't we simply check for NO_SCALING, then go with the v2?

In addition to NO_SCALING, we need to check the cuBLAS version. In fact, more than that, we need to check the cuBLAS version TE was built with, which may not be the same depending on how the user installed TE. Otherwise, we will dispatch to v2 but it will fail during lowering because the C++ functions raise an error in TE core based on the cuBLAS version it was compiled with. In this case, instead of erroring, I'd rather keep the old behavior and fall back to the slower implementation.

> Also, we should be able to check it inside primitive methods instead of passing it as an argument so that when we eventually remove the multi-stream approach, we don't need to change the primitive inputs.

I was planning to do this originally but ran into two issues:
1. I needed to check which backend we're using (original or V2) in multiple places in the primitive. The abstract requires it and the lowering requires it. In the lowering, we don't have easy access to dtype as it's not a jnp.dtype it's some mlir dtype that isn't as easy to convert.
2. The MLIR `ctx` bakes in information about the inputs. You can't simply slice `args` in lowering and pick and choose which primitive to dispatch to. If you do, it'll give an MLIR error that the `ctx` and `args` don't match. As a result, I'd have to update the original FFI to accept `alpha` and `beta` which would break FFI compatibility. Which is why I've gone with the solution of passing additional_arg_0 and additional_arg_1 which needs to happen outside the primitive. This isn't ideal, but was required to get around this fixed arg limitation. Additionally, the static arg indices are fixed at primitive registration and currently the original FFI and the V2 FFI take a different number of args, so for the original FFI I have to use an unused addtional_arg_1 in the outer impl.

As a result of both of these, it's simpler to just check once outside the primitive whether we should use the original or V2, then pass that as a bool flag. Given these restrictions of 1. and 2., I'm not sure this consolidated primitive is a cleaner solution than the split primitives (unless I can break FFI compatibility to unify the FFIs to make the original FFI take alpha and beta).

transformer_engine/common/include/transformer_engine/gemm.h
jberchtold-nvidia: Updated

transformer_engine/jax/csrc/extensions/gemm.cpp
jberchtold-nvidia: Sounds good, renamed to V2

transformer_engine/jax/csrc/extensions/gemm.cpp
jberchtold-nvidia: Removed has_bias and use_async_d2h_group_sizes. The `is_grouped_dense_wgrad` is still required currently but I agree in future we should refactor this logic

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
jberchtold-nvidia: Done

transformer_engine/jax/flax/module.py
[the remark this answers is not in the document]
jberchtold-nvidia: Updated to rename to grouped_dense

Issue #2590, opened 2026-01-13
[PyTorch] Torch.compile support


pggPL:
Make TE layers compatible with torch.compile.

sbhavani:
Related:

- https://github.com/NVIDIA/TransformerEngine/issues/1241
- https://github.com/NVIDIA/TransformerEngine/issues/705

cspades:
A few months ago, I also had trouble compiling TE fused RoPE in TorchTitan nightly:
```python
        # Apply RoPE.
        # xq, xk = apply_rotary_emb(xq, xk, freqs_cis=freqs_cis)
        xq = apply_rotary_pos_emb_te(
            xq,
            # Needs to be real-valued for TE fused RoPE.
            freqs_cis.unsqueeze(1).unsqueeze(1),
            tensor_format="bshd",
            interleaved=False,
            fused=True,
        )
        xk = apply_rotary_pos_emb_te(
            xk,
            freqs_cis.unsqueeze(1).unsqueeze(1),
            tensor_format="bshd",
            interleaved=False,
            fused=True,
        )
```
@sbhavani I think I may or may not have talked to you about this in our many conversations around that time, but this blocked me from getting Llama 405B metrics with FSDP2 + Torch Compile with this fused RoPE (for parity with Megatron-LM's fused RoPE) so we had to math out the best-case improvement instead.

I tried this adding it as a custom op, but didn't have time to dig further.

pstjohn:
Here's a simple repro

```python
import torch
import transformer_engine.pytorch
from transformer_engine.pytorch.attention.rope import RotaryPositionEmbedding

layer = transformer_engine.pytorch.TransformerLayer(128, 128, 8)
rope_layer = RotaryPositionEmbedding(128 // 8)

inputs = torch.randn(8, 128).to("cuda")
freqs = rope_layer.forward(128)

inputs.to("cuda")
freqs.to("cuda")
layer.to("cuda")

layer = torch.compile(layer, fullgraph=True)
output = layer(inputs, rotary_pos_emb=freqs)
```

ntadimeti:
related issue: https://github.com/NVIDIA/TransformerEngine/issues/1241

pggPL:
There is draft PR to add torch.compile support for te.Sequential #2608 - te.sequential will be the first TE layers will torch.compile support.

ben-da6:
[...]

PR #2686, opened 2026-02-17
[PyTorch] torch.compile support for permutation functions

pggPL:
# Description

This PR adds `torch.compile(fullgraph=True)` support for MoE permutation operations (`moe_permute`, `moe_unpermute`, `moe_sort_chunks_by_index`) by converting all `torch.autograd.Function` implementations to PyTorch custom operators using `torch.library.custom_op`.

**Note that this PR does not add torch.compile support for QuantizedTensor as an input.**

Related to #2590 

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:


tests/pytorch/test_permutation.py
```diff
@@ -218,6 +218,17 @@ def backward_wrapper(
     return act.backward(backward_input, retain_graph=retain_graph)
 
 
+def _maybe_compile(fn, use_torch_compile):
+    """Wrap fn with torch.compile(fullgraph=True) if requested."""
+    if use_torch_compile:
+        torch._dynamo.reset()
+        import torch._functorch.config as functorch_config
+
+        functorch_config.donated_buffer = False
```
ptrendx, line 227: What does it do and why do we need to do that? Could we add a comment here, especially since we would be using the internal function here (and so it will most probably break at some point).

tests/pytorch/test_permutation.py
```diff
@@ -1647,14 +1678,24 @@ def perf_test_cuda_kernel(cuda_kernel_fn):
 @pytest.mark.parametrize("hidden_size", [4096])
 @pytest.mark.parametrize("topK", [2, 5])
 @pytest.mark.parametrize("num_out_tokens", [None, 2039])
+@pytest.mark.parametrize("use_torch_compile", [False, True])
+@pytest.mark.skipif(
+    torch.__version__ < "2",
+    reason="torch.compile not available - skipping torch.compile tests",
+)
```
ptrendx: Isn't this redundant with the check inside the test?

tests/pytorch/test_permutation.py
```diff
@@ -1868,19 +1934,30 @@ def test_permutation_mask_map_fp8(
...
 
 @pytest.mark.parametrize("te_dtype", _te_dtypes)
 @pytest.mark.parametrize("num_tokens", [4096])
 @pytest.mark.parametrize("num_expert", [7, 16])
 @pytest.mark.parametrize("hidden_size", [4096])
+@pytest.mark.parametrize("use_torch_compile", [False, True])
+@pytest.mark.skipif(
+    torch.__version__ < "2",
+    reason="torch.compile not available - skipping torch.compile tests",
+)
```
ptrendx: Same.

transformer_engine/pytorch/permutation.py
```diff
@@ -6,511 +6,805 @@
...
+from transformer_engine.pytorch.quantized_tensor import (
+    QuantizedTensor,
+    _quantized_tensor_passthrough_ops,
+)
 from transformer_engine.pytorch.tensor.float8_tensor import Float8Tensor
 from transformer_engine.pytorch.tensor.float8_blockwise_tensor import Float8BlockwiseQTensor
 from transformer_engine.pytorch.tensor.mxfp8_tensor import MXFP8Tensor
 
+# Allow warnings.warn inside torch.compile without graph breaks
+torch._dynamo.config.reorderable_logging_functions.add(warnings.warn)
```
ptrendx: Why do we need that?

transformer_engine/pytorch/permutation.py
```diff
@@ -6,511 +6,805 @@
...
-        if index.dtype != torch.int32:
-            warnings.warn(
-                f"The data type of the input `index` of Permute is {index.dtype}! "
-                "The recommended type is torch.int32."
-            )
-            index = index.to(torch.int32)
+# ===================== _moe_permute_index_map custom ops =====================
 
-        topK = index.size(1)
+# Workspace state for moe_permute_index_map
```
ptrendx, line 29: I don't like it (although I realize this is not really the problem with this PR, but rather the original implementation).

ptrendx: If we can figure out how to change that however, that would be great. Maybe we could make moe_compute a functor (struct MoECompute with `__call__` methods and the workspaces, then moe_compute would just be a object of that class that we would create at the very beginning).

transformer_engine/pytorch/permutation.py
pggPL: In fact we do not need it, removed.

tests/pytorch/test_permutation.py
pggPL: This is optimization of torch.compile which is not compatible with retain_graph=True used in tests.

pggPL: I added some comment.

transformer_engine/pytorch/permutation.py
pggPL: why? I mean what you don't like about it

ptrendx: Well, the main thing is the fact that we implicitly rely on the fact that there is only one permutation happening at a time (and that problem would not be solved by my proposal BTW - this would need a change of this to be actual nn.Module but that has its own problems by effectively being an API break, we should still do it for TE 3.0 though). If you run 2 permutations in 2 streams then that has a chance of silent data corruption since both of those kernels would be using the same underlying workspace. This is something that the user has no way of knowing about without consulting the code. And with torch.compile the chance of this happening may be even bigger - we are at the whim of the compiler optimizations at this point.

tests/pytorch/test_permutation.py
ptrendx: Where?

transformer_engine/pytorch/permutation.py
pggPL: Can we change it in TE 3.0 then? I can indeed change it to functor, but as you said this will not solve a problem.

pggPL: Reply to offline discussion: 

1. there is no support for autograd for ops which mutate args, 
2. torch.compile does not put thing in different streams

ptrendx (approved):

PR #2691, opened 2026-02-19
NVFP4 primary weight support

WanZzzzzz:
# Description

This PR adds NVFP4 partial cast support for distributed training with ZeRO/FSDP optimizers. It enables efficient casting of FP32 master weight shards to NVFP4 model weights with coordinated scaling across data parallel ranks, while minimizing CPU overhead in large-scale training.

## Type of change

- [x] Infra/Build change
- [x] Code refactoring

## Changes

This PR introduces NVFP4 partial cast infrastructure and optimizations for distributed training:

### NVFP4 Partial Cast Kernel (`nvfp4_2d_partial_cast`)
- Implements nibble-accurate partial updates for NVFP4 tensors in distributed settings
- Supports two-level NVFP4 scaling: global FP32 scale + per-block FP8 E4M3 scale

### NVFP4 Transpose Kernel (`nvfp4_transpose`)
- Custom transpose kernel for nibble-packed NVFP4 data with shared memory optimization
- Uses vectorized `uint2` loads/stores with 64×64 tiles for efficient memory access
- Handles nibble repacking during transpose (unlike FP8 byte transpose)
- Enables columnwise data generation for GEMM operations after rowwise AllGather

### Fused Scale Kernel (`nvfp4_fused_scale`)
- Fuses per-block scale computation, global amax copy, and FP8 scale expansion into a single kernel
- Eliminates multiple kernel launches and avoids D2H transfers by accepting tensor pointers
- Reduces kernel launch overhead in the critical path

### Multi-Tensor Dispatch Pattern
- C++-side loop dispatch for NVFP4 multi-tensor operations
- Reduces Python–C++ transition overhead compared to per-tensor Python loops
- Collects metadata in Python and executes batched operations in C++ wrappers

### CPU Overhead Optimizations
- Batched dtype conversion via `torch.cat` / `torch.split`
- Replaced `torch.zeros()` with `torch.empty()` for immediately written buffers
- Consolidated metadata collection and allocation phases
- Optimized bucket partitioning for expert parallel buffers

### Scale Computation Improvements
- Fixed floating-point precision mismatch between Python and CUDA
- Uses FP32 constants consistent with CUDA arithmetic
- Ensures bitwise-identical results between partial and full quantization paths

### New Public API

`cast_master_weights_to_nvfp4()`

- Casts FP32 master weights to NVFP4 model weights
- Handles global and per-block amax reduction across data parallel groups
- Designed for low CPU overhead in distributed training loops

### Testing

| Test | Description |
|------|-------------|
| `test_nvfp4_transpose_kernel` | Verifies correctness for nibble-packed transpose |
| `test_nvfp4_partial_cast_matches_full` | Multi-GPU: partial cast + all-gather equals full cast |
| `test_single_gpu_partial_cast_vs_full` | Single-GPU: offset=0 partial cast matches reference quantizer |
| `_test_cast_master_weights_to_nvfp4` | 500-iteration training loop with bitwise-identical loss |

This feature also passed numeric validation in GPT-3 training on the corresponding Megatron-Core branch:

https://gitlab-master.nvidia.com/qiyuw/megatron-lm-all/-/tree/fp4_primary_opt?ref_type=heads

# Checklist:

### DIFF
--- a/tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
+++ b/tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
@@ -5,6 +5,7 @@
 import argparse
 import datetime
 import os
+import tempfile
 import subprocess
 import sys
 import pathlib
@@ -1241,17 +1242,37 @@ def test_single_gpu_partial_cast_vs_full():
     if test_tensor._amax_rowwise is not None:
         test_tensor._amax_rowwise.zero_()
 
-    # Create a mock distributed group for single GPU
+    # Create a local single-rank process group when running under plain pytest.
+    initialized_here = False
+    rendezvous_file = None
     if not dist.is_initialized():
-        dist.init_process_group(backend="nccl", init_method="env://", rank=0, world_size=1)
-    mock_group = dist.new_group(ranks=[0])
+        torch.cuda.set_device(0)
+        with tempfile.NamedTemporaryFile(delete=False) as f:
+            rendezvous_file = pathlib.Path(f.name)
+        dist.init_process_group(
+            backend="nccl",
+            init_method=rendezvous_file.resolve().as_uri(),
+            rank=0,
+            world_size=1,
+        )
+        initialized_here = True
 
-    quantize_master_weights(
-        [test_tensor],
-        [master_weight.view(-1)],  # Flatten as expected
-        [0],  # offset=0 means full tensor
-        mock_group,
-    )
+    if dist.get_world_size() != 1:
+        pytest.skip("test_single_gpu_partial_cast_vs_full requires world_size == 1")
+
+    mock_group = dist.new_group(ranks=[0], backend="nccl")
+    try:
+        quantize_master_weights(
+            [test_tensor],
+            [master_weight.view(-1)],  # Flatten as expected
+            [0],  # offset=0 means full tensor
+            mock_group,
+        )
+    finally:
+        if initialized_here:
+            dist.destroy_process_group()
+        if rendezvous_file is not None:
+            rendezvous_file.unlink(missing_ok=True)
 
     # Compare amax
     amax_match = torch.equal(test_tensor._amax_rowwise, ref_amax)


transformer_engine/pytorch/tensor/__init__.py
```diff
@@ -42,6 +46,9 @@
     "NVFP4Tensor",
     "prepare_for_saving",
     "restore_from_saved",
+    "cast_master_weights_to_fp8",
+    "cast_master_weights_to_nvfp4",
+    "replace_raw_data",
```
timmoon10: I don't think `te.pytorch.tensor` is a logical place for these functions.
```suggestion
```
We don't need to do it now, but we should move these functions into a better location (maybe `te.pytorch.tensors.distributed_optimizer_utils` or `te.pytorch.optimizers.distributed_optimizer`?). We'll need to do some import tricks to ensure backward compatibility though.

transformer_engine/pytorch/tensor/utils.py
```diff
@@ -155,6 +164,77 @@ def cast_master_weights_to_fp8(
         _cast_master_weights_to_fp8_mxfp8_scaling(mxfp8_scaling_params, *extra_args)
 
 
+def cast_master_weights_to_nvfp4(
+    model_weights,
+    master_weights,
+    start_offsets,
+    group,
+    fsdp_shard_model_weights=None,
+    manual_post_all_gather_processing=False,
```
timmoon10, line 223: We added this kwarg to the FP8 functions for backward compatibility, but there's no point keeping them for these brand-new NVFP4 APIs:
```suggestion
```

transformer_engine/pytorch/tensor/storage/nvfp4_tensor_storage.py
```diff
@@ -316,6 +316,15 @@ def update_usage(
         if columnwise_usage is None:
             columnwise_usage = self._columnwise_data is not None
 
+        # If both rowwise and columnwise are requested, create columnwise from rowwise if needed
+        if rowwise_usage and columnwise_usage:
+            assert (
+                self._rowwise_data is not None and self._rowwise_scale_inv is not None
```
timmoon10: We should also check that the row-wise amax exists.

transformer_engine/pytorch/tensor/storage/nvfp4_tensor_storage.py
```diff
@@ -356,3 +365,49 @@ def update_usage(
             self._columnwise_data = None
             self._columnwise_scale_inv = None
             self._amax_columnwise = None
+
+    def _create_columnwise(self):
+        """
+        Update columnwise data and columnwise scale inv. Can only be used when using 2D scaling.
```
timmoon10, line 382: We should enforce this assumption. We'll have silent data corruption if a user calls `update_usage` on a 1D-scaled tensor.

transformer_engine/pytorch/tensor/storage/nvfp4_tensor_storage.py
```diff
@@ -356,3 +365,49 @@ def update_usage(
...
+            self._columnwise_scale_inv,
+            M_tiles,
+            K_tiles,
+        )
+
+        # Also set columnwise amax (same as rowwise since it's just transposed data)
+        if self._amax_columnwise is None and self._amax_rowwise is not None:
+            self._amax_columnwise = self._amax_rowwise.clone()
+        elif self._amax_rowwise is not None:
+            self._amax_columnwise.copy_(self._amax_rowwise)
```
timmoon10: We require that NVFP4 data has an amax, so we don't need to handle the case where it is missing. I think we can simplify this:
```suggestion
        # Also set columnwise amax (same as rowwise since it's just transposed data)
        if self._amax_columnwise is None:
            self._amax_columnwise = torch.empty_like(self._amax_rowwise)
        self._amax_columnwise.copy_(self._amax_rowwise)
```

transformer_engine/common/include/transformer_engine/recipe.h
```diff
@@ -309,6 +309,144 @@ void nvte_nvfp4_compute_per_tensor_scale(const NVTETensor inpA, const bool use_r
...
+/*! \brief Transpose NVFP4 packed data.
+ *
+ *  Unlike FP8, NVFP4 packs two 4-bit values per byte. This function correctly
+ *  handles the nibble repacking during transpose.
+ *
+ *  \param[in]  input   Input tensor with packed FP4 data. Shape: [M, K/2] bytes.
+ *  \param[out] output  Output tensor with transposed packed data. Shape: [K, M/2] bytes.
+ *  \param[in]  stream  CUDA stream.
+ */
+void nvte_nvfp4_transpose(const NVTETensor input, NVTETensor output, cudaStream_t stream);
```
timmoon10: I think users will run into a few landmines:
- It's unintuitive that the function is in `recipe.h` rather than `transpose.h`.
- It doesn't take in NVFP4 tensors, but rather UINT8 tensors.
- It assumes the NVFP4 data has been quantized with 2D scaling.

I propose two approaches:
1. (Better) Make a single `nvte_nvfp4_2d_transpose` function that operations on NVFP4 tensors. It should internally perform the data transpose, scale transpose-expand, and amax copy. It should live in `transpose.h`.
2. (Worse) Rename this `nvte_nvfp4_data_transpose` or `nvte_4bit_transpose`. Move `nvte_nvfp4_data_transpose` and `nvte_nvfp4_scale_transpose` to `transpose.h`.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -260,6 +260,44 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
   m.def("fp8_transpose", &transformer_engine::pytorch::fp8_transpose, "Transpose with FP8 I/O",
         py::arg("input"), py::arg("dtype"), py::kw_only(), py::arg("out"),
         py::call_guard<py::gil_scoped_release>());
+  m.def("nvfp4_transpose", &transformer_engine::pytorch::nvfp4_transpose,
```
timmoon10: Ideally this could take in `NVFP4Tensor`s and perform both the data and scale transposes internally. Alternatively, we should name this something like `nvfp4_data_transpose` or `4bit_transpose`.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -260,6 +260,44 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
...
+        "Transpose NVFP4 packed data with nibble repacking", py::arg("input"), py::kw_only(),
+        py::arg("out"), py::call_guard<py::gil_scoped_release>());
+  m.def(
+      "nvfp4_scale_transpose", &transformer_engine::pytorch::nvfp4_scale_transpose,
```
timmoon10: We should be more explicit that this assumes 2D quantization:
```suggestion
      "nvfp4_2d_scale_transpose", &transformer_engine::pytorch::nvfp4_scale_transpose,
```

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -260,6 +260,44 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
...
+        py::arg("rows_padded"), py::arg("block_len"), py::call_guard<py::gil_scoped_release>());
+  m.def("nvfp4_multi_tensor_fused_scale",
+        &transformer_engine::pytorch::nvfp4_multi_tensor_fused_scale,
+        "Batched fused scale: compute per-block decode scale, copy global amax, expand to FP8 for "
+        "multiple tensors",
+        py::arg("block_amax_list"), py::arg("global_amax_list"), py::arg("per_block_scale_list"),
+        py::arg("target_scale_list"), py::arg("target_amax_list"), py::arg("tile_rows_list"),
+        py::arg("tile_cols_list"), py::arg("rows_padded_list"), py::arg("block_len"),
+        py::call_guard<py::gil_scoped_release>());
+  m.def("nvfp4_multi_tensor_create_columnwise",
```
timmoon10: We should be more explicit that this assumes 2D quantization. Also, "create" is not quite right since it requires that the output tensors are provided:
```suggestion
  m.def("nvfp4_2d_multi_tensor_transpose",
```

transformer_engine/common/recipe/nvfp4.cu
```diff
@@ -24,9 +75,804 @@ __global__ void compute_nvfp4_per_tensor_scale_kernel(float alpha_in, const floa
...
+    }
+
+    if (!any_valid) {
+      continue;
+    }
+
+    const float2 in01 = make_float2(vals[0], vals[1]);
+    const float2 in23 = make_float2(vals[2], vals[3]);
+    const auto packed =
+        transformer_engine::ptx::mul_cvt_fp32_to_fp4_4x<false>(in01, in23, scale_vec, 0);
```
timmoon10, line 226: This function is only supported when compiling for sm_100a or sm_103a. We should make sure that this source file compiles with support for architecture-specific Blackwell instructions:
https://github.com/NVIDIA/TransformerEngine/blob/63defeadaab055a389fac08e7266ab43860f7d30/transformer_engine/common/CMakeLists.txt#L169

tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
```diff
@@ -187,12 +244,23 @@ def __init__(self, weights, lr, dp_group, manual_post_all_gather_processing=Fals
...
+                buffer_rank_end = storage_rank_end
+            else:
+                weight_buffer_length = self.offsets[-1]
+                buffer_rank_start = rank_start
+                buffer_rank_end = rank_end
         else:
             weight_buffer_dtype = weights[0].dtype
+            weight_buffer_length = self.offsets[-1]
+            buffer_rank_start = rank_start
+            buffer_rank_end = rank_end
```
timmoon10: Nit: It's a bit convoluted, isn't it? It would be much nicer to disentangle the quantization logic from the buffer allocation by computing storage offsets in all cases (even if it's trivial for non-NVFP4 cases) and then using that blindly here.

tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
```diff
@@ -464,8 +567,23 @@ def step(self):
...
+                    continue
+                local_weights.append(local_weight)
+            cast_master_weights_to_nvfp4(
+                self.weights,
+                self.master_weights,
+                [idx[0] for idx in self.weight_indices],
+                self.dp_group,
+                local_weights,
+            )
+        elif isinstance(first_weight, QuantizedTensor):
```
timmoon10, line 565: Nit: It's not logically correct to assume that `QuantizedTensor`s are the same as FP8.
```suggestion
        elif isinstance(first_weight, (Float8Tensor, Float8BlockwiseQTensor, MXFP8Tensor)):
```
This pattern shows up a few other places in this test, and it comes from before this PR.

tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
```diff
@@ -803,5 +1009,249 @@ def main() -> None:
...
+
+    # Compare data
+    data_match = torch.equal(test_tensor._rowwise_data, ref_data)
+
+
 if __name__ == "__main__":
     main()
+    # test_nvfp4_transpose_kernel()
+    # test_single_gpu_partial_cast_vs_full()
+    # test_nvfp4_partial_cast_matches_full()
```
timmoon10: Debugging code:
```suggestion
```

transformer_engine/pytorch/tensor/utils.py
```diff
@@ -155,6 +164,77 @@ def cast_master_weights_to_fp8(
         _cast_master_weights_to_fp8_mxfp8_scaling(mxfp8_scaling_params, *extra_args)
 
 
+def cast_master_weights_to_nvfp4(
```
timmoon10: Do we need a separate API for NVFP4? `cast_master_weights_to_fp8` is already a general function that can handle multiple recipes. If it can handle both FP8 DS and MXFP8, there's nothing special about NVFP4 that makes it more different. I agree having "fp8" in the name is weird, so if we want to add a new API we might as well make it a general function with a name like `quantize_master_weights`:
```python
def quantize_master_weights(...):
    ...

def cast_master_weights_to_fp8(...):  # For backward compatibility
    quantize_master_weights(...)
```

transformer_engine/pytorch/tensor/utils.py
WanZzzzzz: fsdp_shard_model_weights=None is for future FSDP support. It's in the plan.
manual_post_all_gather_processing is also needed for the same reason as FP8 blockwise scaling:
https://github.com/WanZzzzzz/TransformerEngine/blob/38b92b1a168dcfaa6242fea50f03e5a1b873e3a0/transformer_engine/pytorch/tensor/utils.py#L535

transformer_engine/pytorch/tensor/utils.py
```diff
@@ -45,6 +48,12 @@ def replace_raw_data(tensor: QuantizedTensor, new_raw_data: torch.Tensor):
         new_raw_data.detach().copy_(old_raw_data)
         tensor._rowwise_data = new_raw_data
         del old_raw_data
+    elif isinstance(tensor, NVFP4Tensor):
+        old_rowwise = tensor._rowwise_data
+        assert old_rowwise.dtype == new_raw_data.dtype, "The data types of raw data don't match"
+        new_rowwise_data.detach().copy_(old_rowwise)
+        tensor._rowwise_data = new_rowwise_data
```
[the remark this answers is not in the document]
WanZzzzzz: Done

transformer_engine/pytorch/tensor/__init__.py
WanZzzzzz: I removed them. In MCore these functions are imported from transformer_engine.pytorch.tensor.utils

transformer_engine/pytorch/tensor/storage/nvfp4_tensor_storage.py
WanZzzzzz: Done.

transformer_engine/pytorch/tensor/storage/nvfp4_tensor_storage.py
WanZzzzzz: Done.

transformer_engine/pytorch/tensor/storage/nvfp4_tensor_storage.py
WanZzzzzz: Done.

tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
WanZzzzzz: Done.

tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
WanZzzzzz: Done.

transformer_engine/pytorch/tensor/utils.py
WanZzzzzz: Done.

transformer_engine/common/recipe/nvfp4.cu
WanZzzzzz: Moved nvfp4.cu to `transformer_engine_cuda_arch_specific_sources`

transformer_engine/pytorch/tensor/utils.py
timmoon10: I see, that makes sense for now then. Let's change the default to `True` though since that's preferred.

I want to flag a potential future problem with `manual_post_all_gather_processing=False`: it assumes that the quantized tensor has some way to handle the post-processing automatically. For FP8 on Hopper:
```python
cast_master_weights_to_fp8(..., manual_post_all_gather_processing=False)
torch.all_gather(...)

y = model(x)  # Float8Tensor internally performs FP8 transpose
```
This is not something TE will guarantee for future data formats. Maybe the next recipe has some interleaved format:
```python
cast_master_weights_to_futureformat(...)
torch.all_gather(...)
fix_futureformat_interleaving(...)

y = model(x)  # FutureFormatTensor assumes data is interleaved
```
In this case, we should throw an error with the user passes `manual_post_all_gather_processing=False` and it should be Mcore's responsibility to perform the post-processing in a way that's friendly to overlapping.

tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
```diff
@@ -26,13 +27,22 @@
...
+    QuantizedTensor,
+    Float8Tensor,
+    Float8BlockwiseQTensor,
+    NVFP4Tensor,
     is_mxfp8_available,
     QuantizedTensor,
     Float8Tensor,
     Float8BlockwiseQTensor,
     MXFP8Tensor,
 )
```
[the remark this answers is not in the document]
WanZzzzzz: removed

transformer_engine/pytorch/tensor/utils.py
WanZzzzzz: Ok, note it down.

tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
WanZzzzzz: Done.

tests/pytorch/distributed/test_cast_master_weights_to_fp8.py
```diff
@@ -803,5 +988,246 @@ def main() -> None:
...
+    )
+
+    # Compare amax
+    amax_match = torch.equal(test_tensor._amax_rowwise, ref_amax)
+
+    # Compare scale
+    scale_match = torch.equal(test_tensor._rowwise_scale_inv, ref_scale)
+
+    # Compare data
+    data_match = torch.equal(test_tensor._rowwise_data, ref_data)
```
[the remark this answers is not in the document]
WanZzzzzz: Done

PR #2694, opened 2026-02-20
[PyTorch] Zero-initialize learnable softmax_offset in DotProductAttention

fjosw:
# Description

The PyTorch implementation of DotProductAttention initializes the learnable `softmax_offset` parameter with `torch.empty()`, which leaves it containing uninitialized memory. Unlike all other `TransformerEngineBaseModule` subclasses (Linear, LayerNormLinear, LayerNormMLP, GroupedLinear), DotProductAttention does not call `self.reset_parameters(`) in its `__init__`, so the deferred initialization system that would normally overwrite the torch.empty() contents is never invoked. The JAX implementation explicitly uses nn.initializers.zeros for this parameter. This fix aligns the PyTorch behavior by using torch.zeros() directly. In Megatron-LM this is not a problem because the paramter is initialised explicitly but when used in isolation this can lead to problems.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change torch.empty() to torch.zeros() when creating the learnable softmax_offset parameter in DotProductAttention, ensuring it is zero-initialized rather than containing uninitialized memory.

# Checklist:


ptrendx:
LGTM, thank you for the fix. @cyanguwa Could you also take a look?

PR #2696, opened 2026-02-20
remove deprecated qkv/kv_packed apis

sudhakarsingh27:
# Description

Remove the depracated APIs - `nvte_fused_attn_fwd_qkvpacked`, `nvte_fused_attn_bwd_qkvpacked`, `nvte_fused_attn_fwd_kvpacked` and `nvte_fused_attn_bwd_kvpacked`. 

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


ptrendx:
Why do we need to do this now?

sudhakarsingh27:
> Why do we need to do this now?

@ptrendx, it's not a must but these APIs have been "deprecated" for too long so it was decided to remove these with the next release. 

> We should not do this before changing the major version.

So, can we only do this in the next TE 3.0 release?

ptrendx:
For better or worse they are part of our public API - we should not remove them in the minor release if we don't have a good reason for that.

sudhakarsingh27:
Does it mean we have to keep supporting the deprecated APIs as well? If we don't, then it's basically non-working code and I'm not sure if we should keep it in there.

PR #2698, opened 2026-02-22
Add fused_adam, quantized_model_init, and fsdp2 example

### LINKED ISSUE #2682: Example of `quantized_model_init` for low-precision compute weights, fp32 main weights using fused_adam with fsdp2
This is likely the optimal way to run FSDP2 with TE layers, but it would be great to have an example of using FusedAdam with master_weights=True (https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/optimizers/fused_adam.py#L75-L82), `quantized_model_init`, and `fully_shard` together.

pstjohn:
## Summary

- Fix `FusedAdam` to work with PyTorch-native FSDP2 (`fully_shard`) when parameters are `DTensor`-wrapped `Float8Tensor`/`QuantizedTensor`
- Fix `fuse_wgrad_accumulation` guard to avoid crashing with vanilla FSDP2 (previously assumed Megatron-Core FSDP exclusively)
- Add examples for `quantized_model_init` on single-GPU (`main.py`) and multi-GPU FSDP2 (`fully_shard.py`)

### Note: `fuse_wgrad_accumulation` remains incompatible with vanilla FSDP2

`fuse_wgrad_accumulation` still cannot be used with vanilla FSDP2. The feature writes weight gradients directly into `main_grad` and returns `None` to autograd, bypassing FSDP2's reduce-scatter. Each rank ends up with an unreduced gradient. Megatron-Core FSDP solves this by wiring `get_main_grad()` into its own reduce-scatter infrastructure. Vanilla FSDP2 does not yet expose an equivalent hook.

Fixes #2682

### DIFF
--- a/qa/L0_pytorch_unittest/test.sh
+++ b/qa/L0_pytorch_unittest/test.sh
@@ -37,6 +37,7 @@ python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_quantized_tensor
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_float8blockwisetensor.xml $TE_PATH/tests/pytorch/test_float8blockwisetensor.py || test_fail "test_float8blockwisetensor.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_float8_blockwise_scaling_exact.xml $TE_PATH/tests/pytorch/test_float8_blockwise_scaling_exact.py || test_fail "test_float8_blockwise_scaling_exact.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_float8_blockwise_gemm_exact.xml $TE_PATH/tests/pytorch/test_float8_blockwise_gemm_exact.py || test_fail "test_float8_blockwise_gemm_exact.py"
+python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/test_grouped_tensor.xml $TE_PATH/tests/pytorch/test_grouped_tensor.py || test_fail "test_grouped_tensor.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_gqa.xml $TE_PATH/tests/pytorch/test_gqa.py || test_fail "test_gqa.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_fused_optimizer.xml $TE_PATH/tests/pytorch/test_fused_optimizer.py || test_fail "test_fused_optimizer.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_multi_tensor.xml $TE_PATH/tests/pytorch/test_multi_tensor.py || test_fail "test_multi_tensor.py"
--- a/tests/pytorch/nvfp4/test_nvfp4_group_quantize_graph_safe.py
+++ b/tests/pytorch/nvfp4/test_nvfp4_group_quantize_graph_safe.py
@@ -10,7 +10,7 @@ from transformer_engine.pytorch.custom_recipes.quantization_nvfp4 import NVFP4Qu
 from transformer_engine.pytorch.custom_recipes import utils
 from transformer_engine.pytorch.constants import TE_DType
 from transformer_engine.common.recipe import NVFP4BlockScaling
-from transformer_engine.pytorch.tensor.storage.grouped_tensor import GroupedTensor
+from transformer_engine.pytorch.tensor.grouped_tensor import GroupedTensor
 
 import pytest
 import torch
--- a/tests/pytorch/test_grouped_tensor.py
+++ b/tests/pytorch/test_grouped_tensor.py
@@ -8,7 +8,7 @@ from typing import List, Tuple
 import pytest
 import torch
 import transformer_engine.pytorch as te
-from transformer_engine.pytorch.tensor.storage.grouped_tensor import GroupedTensor
+from transformer_engine.pytorch.tensor.grouped_tensor import GroupedTensor
 from transformer_engine.pytorch import (
     Quantizer,
     Float8Quantizer,
@@ -125,7 +125,7 @@ class TestGroupedTensor:
 
         grouped_tensor = GroupedTensor.make_grouped_tensor_with_shapes(
             num_tensors=num_tensors,
-            shape=shape,
+            shapes=shape,
             quantizer=None,
             device="cuda",
             dtype=torch.float32,
@@ -147,7 +147,7 @@ class TestGroupedTensor:
 
         grouped_tensor = GroupedTensor.make_grouped_tensor_with_shapes(
             num_tensors=num_tensors,
-            shape=shape,
+            shapes=shape,
             quantizer=None,
             device="cuda",
             dtype=torch.float32,
@@ -170,14 +170,18 @@ class TestGroupedTensor:
 
         grouped_tensor = GroupedTensor.make_grouped_tensor_with_shapes(
             num_tensors=num_tensors,
-            shape=shape,
+            shapes=shape,
             quantizer=None,
             device="cuda",
             dtype=torch.float32,
         )
 
-        # Get the original data pointer
-        original_data_ptr = grouped_tensor.data.data_ptr()
+        # GroupedTensor is a wrapper; use backing storage buffer pointer.
+        storage = grouped_tensor.rowwise_data
+        if storage is None:
+            storage = grouped_tensor.columnwise_data
+        assert storage is not None
+        original_data_ptr = storage.data_ptr()
 
         # Split into tensors
         tensors = grouped_tensor.split_into_quantized_tensors()
@@ -207,13 +211,18 @@ class TestGroupedTensor:
 
         grouped_tensor = GroupedTensor.make_grouped_tensor_with_shapes(
             num_tensors=num_tensors,
-            shape=shape,
+            shapes=shape,
             quantizer=quantizer,
             device="cuda",
+            dtype=torch.float32,
         )
 
-        # Get the original data pointer
-        original_data_ptr = grouped_tensor.data.data_ptr()
+        # GroupedTensor is a wrapper; use backing storage buffer pointer.
+        storage = grouped_tensor.rowwise_data
+        if storage is None:
+            storage = grouped_tensor.columnwise_data
+        assert storage is not None
+        original_data_ptr = storage.data_ptr()
 
         # Split into tensors
         tensors = grouped_tensor.split_into_quantized_tensors()
@@ -236,13 +245,17 @@ class TestGroupedTensor:
 
         grouped_tensor = GroupedTensor.make_grouped_tensor_with_shapes(
             num_tensors=num_tensors,
-            shape=shape,
+            shapes=shape,
             quantizer=None,
             device="cuda",
             dtype=torch.float32,
         )
 
-        original_data_ptr = grouped_tensor.data.data_ptr()
+        storage = grouped_tensor.rowwise_data
+        if storage is None:
+            storage = grouped_tensor.columnwise_data
+        assert storage is not None
+        original_data_ptr = storage.data_ptr()
         tensors = grouped_tensor.split_into_quantized_tensors()
 
         assert len(tensors) == num_tensors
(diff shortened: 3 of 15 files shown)


transformer_engine/pytorch/optimizers/fused_adam.py
```diff
@@ -419,7 +425,14 @@ def initialize_state(self, param, store_param_remainders):
                 store_param_remainders=store_param_remainders,
             )
             if not store_param_remainders:
-                self.set_scaled_state(param, "master_param", param.clone().detach().float())
+                # Extract local tensor from DTensor and dequantize QuantizedTensor
+                # to get a plain float32 copy for the master weight.
+                local_param = param._local_tensor if isinstance(param, DTensor) else param
+                if isinstance(local_param, QuantizedTensor):
+                    master = local_param.dequantize().clone().detach().float()
```
cspades: Should we use `dequantize(dtype=torch.float32)`, to fuse the cast into the [de-quantization's output buffer](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/csrc/extensions/cast.cpp#L94)? (Likely not a big deal since I don't think this will change anything numerically, and you only call this function during init and whenever you save and load DCP checkpoints.)

XueSongTap:
@pstjohn Hi, thanks for the great work! Does this PR plan to also handle the BF16 path? I noticed the BF16 branch still operates on the original p/p_grad without unwrapping when they're DTensors. In my experiments with FSDP2 + BF16, I'm seeing non-trivial overhead during the optimizer step from repeated DTensor dispatch. Curious if that's intentional or a planned follow-up.

transformer_engine/pytorch/optimizers/fused_adam.py
pstjohn: changed, thanks!

tests/pytorch/distributed/test_torch_fsdp2.py
```diff
@@ -32,28 +61,131 @@ def _run_test(fp_init, sharding_dims, recipe, layer_type):
...
+@pytest.mark.skipif(NUM_PROCS < 2, reason="Requires 2+ GPUs")
+def test_fsdp2_fused_adam_fp8_master_weights(fp_recipe):
+    """FusedAdam(master_weights=True) + FSDP2 + quantized_model_init."""
+    if fp_recipe in ("Float8BlockScaling", "MXFP8BlockScaling", "NVFP4BlockScaling"):
+        pytest.xfail(
+            f"{fp_recipe}: QuantizedTensor local tensors are not preserved "
+            "after optimizer step -- the fused CUDA kernel replaces them "
+            "with plain tensors"
+        )
+    _run_fused_adam_test("fused_adam_fp8_master_weights", fp_recipe)
```
pstjohn, line 124: this comment is wrong, I'm assuming these are being reverted to regular tensors after `fully_shard` because we need to call `.view()` on a block tensor that doesn't support reshaping, and we end up dequantizing...

tests/pytorch/distributed/test_torch_fsdp2.py
```diff
@@ -32,28 +61,130 @@ def _run_test(fp_init, sharding_dims, recipe, layer_type):
...
+
+
+@pytest.mark.skipif(NUM_PROCS < 2, reason="Requires 2+ GPUs")
+def test_fsdp2_safetensors_fp32_export(fp_recipe):
+    """Export FP32 model from optimizer master weights to safetensors."""
+    if fp_recipe == "MXFP8BlockScaling":
+        pytest.xfail(
+            "MXFP8BlockScaling: FusedAdam CUDA kernel does not support "
+            "MXFP8 quantized tensors, causing illegal memory access"
+        )
```
pstjohn, line 211: claude's analysis:

Root cause: MXFP8Tensor inherits from QuantizedTensor but NOT from Float8Tensor. In fused_adam.py step():                                                                                                 
  1. Line 617: isinstance(p, Float8Tensor) → False for MXFP8Tensor
  2. Line 629: p.dtype in [torch.float16, torch.bfloat16] → True (nominal dtype is bfloat16)          
  3. So p.data (which is still an MXFP8Tensor wrapper with MXFP8-layout storage) gets added to p_f16_model
  4. The multi_tensor_adam CUDA kernel treats this as plain bf16 memory → illegal memory access

transformer_engine/pytorch/optimizers/fused_adam.py
```diff
@@ -575,6 +588,10 @@ def step(self, closure=None, grad_scaler=None):
 
                 if p_grad is None:
                     continue
+                # Extract local tensors from DTensors (e.g. from FSDP2)
+                # so that multi_tensor kernels receive plain CUDA tensors.
+                if isinstance(p_grad, DTensor):
```
vthumbe1503: Is this really needed? Since p_grad is high precision, all Dtensor's op should get translated to local tensor's op without needing to extract it.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -365,6 +365,21 @@ def reshape(self, *shape: Tuple[int]) -> MXFP8Tensor:
...
+        """Return the underlying UntypedStorage of the FP8 data.
+
+        Note that MXFP8 tensor may involve multiple buffers: row-wise
+        FP8 data, row-wise scales, column-wise FP8 data, column-wise
+        scales. The UntypedStorage of the row-wise FP8 data is returned
+        if it exists, and otherwise the UntypedStorage of the
+        column-wise FP8 data.
+
+        """
+        data = self._rowwise_data if self._rowwise_data is not None else self._columnwise_data
```
vthumbe1503: Can you please provide reason(maybe in the PR details) as to why this is needed? And where is this used? Is this needed for DCP checkpointing. And if so why?

vthumbe1503 (approved):
Thanks for the PR. Left a few mino comments. LGTM post CI success.

tests/pytorch/distributed/run_fsdp2_fused_adam.py
```diff
@@ -0,0 +1,649 @@
...
+        with te.autocast(enabled=True, recipe=recipe):
+            ref_output = model(x).clone()
+
+    # ── Save checkpoint ──────────────────────────────────────────────
+    checkpoint_dir = "/tmp/te_test_fsdp2_dcp_parity"
+
+    if isinstance(recipe, transformer_engine.common.recipe.DelayedScaling):
+        model_state = {
+            k: v for k, v in model.state_dict().items() if not k.endswith("_extra_state")
+        }
```
vthumbe1503, line 505: Can you also please comment why we should avoid saving _extra_state. As in what error we get with dcp if we dont do so?

pstjohn: This one I remember, the others I'll need to comment out and run the test suite again 😅

But this is a known hassle where torch DCP needs the sizes of these tensors to remain consistent during saving & loading, and since this is pickled data, it changes when there's data in that field.

The alternative is a detailed load_planner for DelayedScaling that reads and allocates the extra state data tensor

transformer_engine/pytorch/optimizers/fused_adam.py
pstjohn: hmm, yeah I guess not -- reverted!

transformer_engine/pytorch/tensor/mxfp8_tensor.py
pstjohn: This would be in `tests/pytorch/distributed/test_torch_fsdp2.py::test_fsdp2_dcp_output_parity_async[MXFP8BlockScaling]`,

```
[rank1]:   File "/usr/local/lib/python3.12/dist-packages/torch/distributed/fsdp/_fully_shard/_fsdp_state.py", line 64, in fsdp_hook_wrapper
[rank1]:     return torch._dynamo.disable(
[rank1]:            ^^^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/usr/local/lib/python3.12/dist-packages/torch/_dynamo/eval_frame.py", line 1252, in _fn
[rank1]:     return fn(*args, **kwargs)
[rank1]:            ^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/usr/local/lib/python3.12/dist-packages/torch/distributed/fsdp/_fully_shard/_fsdp_state.py", line 257, in _pre_forward
[rank1]:     args, kwargs = self._root_pre_forward(module, args, kwargs)
[rank1]:                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank1]:   File "/usr/local/lib/python3.12/dist-packages/torch/distributed/fsdp/_fully_shard/_fsdp_state.py", line 133, in _root_pre_forward
[rank1]:     self._lazy_init()
[rank1]:   File "/usr/local/lib/python3.12/dist-packages/torch/distributed/fsdp/_fully_shard/_fsdp_state.py", line 203, in _lazy_init
[rank1]:     state._fsdp_param_group.lazy_init()
[rank1]:   File "/usr/local/lib/python3.12/dist-packages/torch/distributed/fsdp/_fully_shard/_fsdp_param_group.py", line 261, in lazy_init
[rank1]:     fsdp_param.reset_sharded_param()
[rank1]:   File "/usr/local/lib/python3.12/dist-packages/torch/distributed/fsdp/_fully_shard/_fsdp_param.py", line 890, in reset_sharded_param
[rank1]:     == local_tensor.untyped_storage().data_ptr()
```

 but we're currently xfailing that test anyways because of the error in multi_tensor_apply,
```
[rank0]:   File "/workspace/transformer_engine/pytorch/optimizers/multi_tensor_apply.py", line 21, in __call__
[rank0]:     return op(self.chunk_size, noop_flag_buffer, tensor_lists, *args)
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: RuntimeError: /workspace/transformer_engine/common/multi_tensor/multi_tensor_apply.cuh:92 in function multi_tensor_apply: CUDA Error: an illegal memory access was encountered
```

So i suppose we could revert this as well. I just added an additional description to the docstring as to why it's here

tests/pytorch/distributed/test_torch_fsdp2.py
```diff
@@ -32,28 +61,132 @@ def _run_test(fp_init, sharding_dims, recipe, layer_type):
...
+
+@pytest.mark.skipif(NUM_PROCS < 2, reason="Requires 2+ GPUs")
+def test_fsdp2_dcp_output_parity_async(fp_recipe):
+    """DCP save/load round-trip into a fresh model produces identical outputs."""
+    if fp_recipe in ("DelayedScaling", "Float8CurrentScaling"):
+        pytest.xfail(
+            f"async DCP save/load with {fp_recipe} produces different outputs: "
+            "the async staging may capture stale tensor state for FP8 scaling "
+            "factors, causing numerical divergence after reload"
+        )
```
pstjohn, line 182: this isn't right -- it's actually NaNs when we load back. I'm digging on this

pstjohn: * Sync save (dcp.save): Uses _SerialCpuLoader.values() (line 131) which calls tensor.detach().cpu(). For Float8Tensor, .cpu() calls self.dequantize().cpu(), properly converting FP8 data to a plain bfloat16 tensor.
                                                                                                                                                                                                                                               
* Async save (dcp.async_save): Uses StateDictStager._offload_tensor() which tries to deep-copy the tensor's underlying storage. Float8Tensor is a wrapper subclass (_make_wrapper_subclass) with data_ptr() == 0 (empty storage). The staging code at line 215 skips the storage copy for wrapper subclasses, creating a plain tensor with uninitialized garbage data. The actual FP8 data (in _data, _scale_inv attributes) is deep-copied but ignored by DCP when writing.

pstjohn: i'll put a proposed fix for this in #2721

transformer_engine/pytorch/tensor/mxfp8_tensor.py
vthumbe1503: So I guess we wont be needing this if multi tensor adam is implemented for mxfp8 right?

vthumbe1503:
Merging the PR since all tests pass except the one that is expected to fail on the main branch

PR #2706, opened 2026-02-25
Newton-Schulz via cuSOLVERMp

vcherepanov-nv:
# Description

Adds an API to call Newton-Schulz method on a distributed tensor.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Integrate cuSOLVERMp as a new dependency
- Add corresponding API to TE/common
- Add PyTorch binding and tests

# Checklist:


transformer_engine/pytorch/newton_schulz.py
cyanguwa, line 1: I think we agreed on putting this file in [transformer_engine/pytorch/optimizers](https://github.com/NVIDIA/TransformerEngine/tree/main/transformer_engine/pytorch/optimizers)?

Also, we want to fully replicate the functionality of [newton_schulz_tp](https://github.com/NVIDIA/Megatron-LM/blob/28c586e91506631835d8c5f29bf325a4e5aefddd/megatron/core/optimizer/muon.py#L80), so may need to take a look at supporting the `partition_dim` and `mode` parameters? Looking at Emerging-Optimizers' [implementation](https://github.com/NVIDIA-NeMo/Emerging-Optimizers/blob/39692907bdb6bbdc5c82c2318a5f5255622287cb/emerging_optimizers/orthogonalized_optimizers/muon_utils.py#L166), I think it's mostly wrapper code that we need to add in Python.

tests/pytorch/distributed/run_newton_schulz.py
```diff
@@ -0,0 +1,90 @@
...
+        newton_schulz(x_local, ctx, args.num_iterations)
+    finally:
+        ctx.destroy()
+
+    # Gather results
+    gathered = [torch.empty_like(x_local) for _ in range(world_size)]
+    dist.all_gather(gathered, x_local)
+    X = torch.cat(gathered, dim=0)
+
+    # Check: the resulting matrix should be orthogonal
```
cyanguwa: Checking whether the result is orthogonal might not be enough - I think we said we'd compare the actual values of the matrix with the result of a non-distributed version?

tests/pytorch/distributed/test_newton_schulz.py
```diff
@@ -0,0 +1,47 @@
...
+
+TEST_ROOT = Path(__file__).parent.resolve()
+NUM_PROCS = torch.cuda.device_count()
+LAUNCH_CMD = ["torchrun", f"--nproc_per_node={NUM_PROCS}"]
+
+
+@pytest.mark.parametrize("dtype", ["float32", "bfloat16"])
+@pytest.mark.parametrize("matrix_size", [256])
+@pytest.mark.parametrize("num_iterations", [5, 15])
+def test_newton_schulz(dtype, matrix_size, num_iterations):
```
cyanguwa: Could we expand the tests a bit to cover more sizes, shapes, etc?

https://github.com/NVIDIA-NeMo/Emerging-Optimizers/blob/main/tests/test_muon_utils.py

transformer_engine/pytorch/newton_schulz.py
```diff
@@ -0,0 +1,144 @@
...
+import torch.distributed as dist
+
+import transformer_engine_torch as tex
+
+_CUSOLVERMP_REQUIRED = (
+    "Newton-Schulz requires Transformer Engine to be built with NVTE_WITH_CUSOLVERMP=1"
+)
+
+
+class CusolverMpCtx:
```
cyanguwa, line 125: Is this the "normal" way of calling a distributed solver? Feels a bit clunky - but I can't come up with any good ideas :) @timmoon10 any thoughts?

timmoon10: I agree, it does seem cumbersome for the user to maintain the cuSolverMp context and pass it everywhere. Do we expect the user to need to maintain multiple contexts and switch between them? If not, do we need to expose the context to the user? We could implement as a singleton:
```python
class _CusolverMpCtx:
    def __init__(self, ...):
        self._ptr = tex.cusolvermp_ctx_create(...)

    def __del__(self):
        # Called when the context is manually destroyed or during Python teardown
        tex.cusolvermp_ctx_destroy(self._ptr)

_CUSOLVER_CTX = None

def init_cusolvermp(...):
    global _CUSOLVER_CTX
    if _CUSOLVER_CTX is None:
        _CUSOLVER_CTX = _CusolverMpCtx(...)

def destroy_cusolvermp():
    global _CUSOLVER_CTX
    _CUSOLVER_CTX = None

def newton_schulz(...):
    ctx = _CUSOLVER_CTX

    # Alternately, accept ctx as a kwarg that defaults to None
    # if ctx is None:
    #     ctx = _CUSOLVER_CTX

     ...
```
I'm loosely basing this on how PyTorch initializes NCCL with torch.distributed.init_process_group.

transformer_engine/common/newton_schulz/newton_schulz.cpp
timmoon10, line 1: This file depends on cuSolverMp being available, so `CMakeLists.txt` skips compilation if cuSolverMp is missing. However, this means that the C API functions may or may not be defined. I expect this will cause linker errors between `libtransformer_engine.so` and `transformer_engine_torch`.

Instead, how about we make sure the C API functions are always defined, even if they're missing implementations:
```c++
#include "transformer_engine/newton_schulz.h"

#ifdef NVTE_WITH_CUSOLVERMP
#include <cusolverMp.h>

namespace {
...
}  // namespace

#endif  // NVTE_WITH_CUSOLVERMP

NVTECusolverMpCtx* nvte_cusolvermp_ctx_create(...) {
#ifdef NVTE_WITH_CUSOLVERMP
  ...
#else
  NVTE_ERROR("Transformer Engine has not been built with cuSolverMp support.");
#endif  // NVTE_WITH_CUSOLVERMP
}
```

transformer_engine/common/include/transformer_engine/newton_schulz.h
```diff
@@ -0,0 +1,68 @@
...
+
+#include <nccl.h>
+#include <stdint.h>
+
+#include "transformer_engine.h"
+
+#ifdef __cplusplus
+extern "C" {
+#else
+#include <stdbool.h>
```
timmoon10: It doesn't seem that this header is being used.
```suggestion
```

transformer_engine/common/newton_schulz/newton_schulz.cpp
```diff
@@ -0,0 +1,230 @@
...
+  void operator()(cusolverMpHandle_t handle) const { cusolverMpDestroy(handle); }
+};
+using CusolverMpHandle =
+    std::unique_ptr<std::remove_pointer_t<cusolverMpHandle_t>, CusolverMpHandleDeleter>;
+
+struct CusolverMpGridDeleter {
+  void operator()(cusolverMpGrid_t grid) const { cusolverMpDestroyGrid(grid); }
+};
+using CusolverMpGrid =
+    std::unique_ptr<std::remove_pointer_t<cusolverMpGrid_t>, CusolverMpGridDeleter>;
```
timmoon10: Why do we need these in global scope? `NVTECusolverMpCtx` is declared in `newton_schulz.h`, but it is defined and used only in this file. It seems nicer to keep these helper structs in an anonymous namespace to avoid polluting the global namespace.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -545,6 +443,19 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
         &transformer_engine::pytorch::multi_tensor_compute_scale_inv_e8m0_cuda,
         "Fused compute E8M0 scale_inv from amax", py::call_guard<py::gil_scoped_release>());
 
+#ifdef NVTE_WITH_CUSOLVERMP
```
timmoon10: I don't think it's necessary for the PyTorch build to be aware of cuSolverMp availability. If we made sure the C API functions are always available, even if they just error out internally, then PyTorch doesn't need to know anything.

transformer_engine/pytorch/newton_schulz.py
timmoon10: ~Yea, I'd suggest something like`transformer_engine/pytorch/optimizers/muon`.~

build_tools/pytorch.py
```diff
@@ -87,6 +87,13 @@ def setup_pytorch_extension(
         libraries.append("nvshmem_host")
         cxx_flags.append("-DNVTE_ENABLE_NVSHMEM")
 
+    if bool(int(os.getenv("NVTE_WITH_CUSOLVERMP", "0"))):
+        cusolvermp_home = Path(os.getenv("CUSOLVERMP_HOME", "/usr"))
+        include_dirs.append(cusolvermp_home / "include")
+        library_dirs.append(cusolvermp_home / "lib")
+        libraries.append("cusolverMp")
+        cxx_flags.append("-DNVTE_WITH_CUSOLVERMP")
+
```
timmoon10: This can be removed if the C API functions are always available.
```suggestion
```

transformer_engine/common/CMakeLists.txt
```diff
@@ -227,6 +227,11 @@ list(APPEND transformer_engine_SOURCES
      comm_gemm/comm_gemm.cpp)
 endif()
 
+if (NVTE_WITH_CUSOLVERMP)
+list(APPEND transformer_engine_SOURCES
+     newton_schulz/newton_schulz.cpp)
+endif()
```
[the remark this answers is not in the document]
timmoon10: Good point.

transformer_engine/common/newton_schulz/newton_schulz.cpp
```diff
@@ -0,0 +1,230 @@
...
+      .rank = rank,
+      .stream = stream,
+      .in_ready = in_ready,
+      .out_ready = out_ready,
+      .handle = std::move(handle),
+      .grid = std::move(grid),
+      .workspace = nullptr,
+      .workspace_size = 0,
+      .workspace_registered = false,
+  };
```
timmoon10, line 158: Unfortunately we are still stuck on C++17.
```suggestion
  return new NVTECusolverMpCtx(
      nranks,
      rank,
      stream,
      in_ready,
      out_ready,
      std::move(handle),
      std::move(grid),
      nullptr,
      0,
      false,
  );
```

transformer_engine/common/newton_schulz/newton_schulz.cpp
```diff
@@ -0,0 +1,194 @@
...
+  const int64_t nb = n;
+
+  // Compute local leading dimension
+  const int64_t local_rows = cusolverMpNUMROC(m, mb, ctx->rank, 0, ctx->nranks);
+  const int64_t lld = std::max(local_rows, static_cast<int64_t>(1));
+
+  const cudaDataType_t cuda_dtype = get_cuda_dtype(t->dtype());
+
+  // Create matrix descriptor
+  auto mat_desc = MakeCusolverMpMatrixDesc(ctx->grid.get(), cuda_dtype, m, n, mb, nb, 0, 0, lld);
```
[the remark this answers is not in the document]
timmoon10: I agree this is a bug. TE should use row-major ordering in all of its external-facing APIs and it should internally reinterpret as column-major ordering when calling cuSolverMp.

timmoon10 (changes requested):

transformer_engine/pytorch/newton_schulz.py
timmoon10: Actually, this depends on how we're planning on exposing this. If TE owns an implementation of Muon, then I agree we can treat this as internal utilities. If the Muon implementation lives in Megatron-LM or somewhere else, then we should treat this as a general API in something like `transformer_engine/pytorch/cusolvermp`.

transformer_engine/common/include/transformer_engine/newton_schulz.h
```diff
@@ -0,0 +1,68 @@
...
+ *  \brief Functions for distributed Newton-Schulz matrix orthogonalization.
+ *
+ *  This API is a TE-native binding to the cuSolverMp library.
+ *  It computes an iterative Newton-Schulz matrix orthogonalization on a distributed matrix.
+ */
+
+#ifndef TRANSFORMER_ENGINE_COMMON_NEWTON_SCHULZ_H_
+#define TRANSFORMER_ENGINE_COMMON_NEWTON_SCHULZ_H_
+
+#include <nccl.h>
```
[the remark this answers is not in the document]
timmoon10: I'm also a little uneasy at exposing NCCL as a required dependency, but I see we already import the NCCL header elsewhere in the codebase: https://github.com/NVIDIA/TransformerEngine/blob/580e7aa28bebb83489271107ce50861cf84f3170/transformer_engine/common/util/logging.h#L15

transformer_engine/pytorch/newton_schulz.py
```diff
@@ -0,0 +1,141 @@
...
+            te.newton_schulz(x, ctx)
+    """
+
+    def __init__(self, ptr: int, nranks: int) -> None:
+        self._ptr = ptr
+        self.nranks = nranks
+
+    def destroy(self) -> None:
+        """Destroy the underlying cuSolverMp context."""
+        tex.cusolvermp_ctx_destroy(self._ptr)
```
timmoon10: Could we make this class automatically cleanup whenever the refcount goes to zero?
```suggestion
    def destroy(self) -> None:
        """Destroy the underlying cuSolverMp context."""
        if self._ptr is not None:
            tex.cusolvermp_ctx_destroy(self._ptr)
            self._ptr = None

    def __del__(self) -> Non:
        # Called when the context is manually destroyed or during Python teardown
        self.destroy()
```

transformer_engine/pytorch/newton_schulz.py
```diff
@@ -0,0 +1,141 @@
...
+def _get_nccl_comm_ptr(group: dist.ProcessGroup) -> int:
+    """Extract the raw NCCL communicator pointer from a PyTorch process group."""
+    backend = dist.get_backend(group)
+    if backend != "nccl":
+        raise RuntimeError(f"Newton-Schulz requires NCCL backend, got '{backend}'")
+    nccl_backend = group._get_backend(torch.device("cuda"))
+    return nccl_backend._comm_ptr()
+
+
+def cusolvermp_ctx_create(group: dist.ProcessGroup) -> CusolverMpCtx:
```
timmoon10: Could we move this logic into `CusolverMpCtx`'s constructor? I'm not sure if we are getting any benefit from two-stage initialization.

timmoon10 (commented):
We could tweak the design of the Python wrapper for the cuSOLVERMp context, but otherwise this looks good to me.

transformer_engine/common/include/transformer_engine/newton_schulz.h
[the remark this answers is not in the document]
vcherepanov-nv: We use NCCL types (ncclComm_t) in nvte_ctx_create API, so if we're taking the route of always defining the functions - we need to pull NCCL headers unconditionally.

timmoon10 (dismissed):
LGTM

timmoon10 (approved):

PR #2709, opened 2026-02-26
[PyTorch] Fix L3 FA tests

cyanguwa:
# Description

This PR fixes the L3 tests for FP8 current scaling in `L3_pytorch_FA_versions_test`. The fix is only related to the selection logic in the test and not the backend support itself.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- L3 CI test fix.

# Checklist:

### DIFF
--- a/tests/pytorch/attention/test_attention.py
+++ b/tests/pytorch/attention/test_attention.py
@@ -1865,7 +1865,7 @@ def test_mha_fp8_vs_f16(
         )
     fp8_meta = {}
     fp8_meta["recipe"] = fp8_recipe
-    available_backends, _, fused_attn_backends = get_available_attention_backends(
+    available_backends, _, _ = get_available_attention_backends(
         config,
         qkv_dtype=torch.float8_e4m3fn,
         qkv_layout=qkv_format.replace("hd", "h3d"),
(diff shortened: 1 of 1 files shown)


tests/pytorch/attention/test_attention.py
```diff
@@ -2138,8 +2139,8 @@ def test_dpa_fp8_vs_f16(dtype, model, qkv_layout, fp8_dpa_bwd, is_training, scal
             is_training=is_training,
             deterministic=_deterministic,
         )
-        _, fused_attn_supported, _ = available_backends
-        if not fused_attn_supported:
+        _, fused_attn_supported_f16, _ = available_backends
```
KshitijLakhani: So if the value for  `fused_attn_supported_fp8=1/True` and `fp8_dpa_bwd=0/False` then `fused_attn_supported_f16` be 0/False or 1/True right ?
If `fused_attn_supported_f16` is 0/False, there's no fp16 to compare the fp8 and hence we skip, but
,if `fused_attn_supported_f16` is 1/True, then compare fp8 and fp16 fwd only.

Now, if `fp8_dpa_bwd=1/True`, then `fused_attn_supported_f16` will be `False/0` always 
In that case what does fp8 get compared to ? Maybe I missed it, but I see no logic for this comparison - could you point me to it ?

cyanguwa: Makes sense. I removed the `if not fp8_bwd` logic. Could you please take a look at [229b1ef](https://github.com/NVIDIA/TransformerEngine/pull/2709/commits/229b1effece54f35b23564e554babeb6d6fb16c7)?

cyanguwa:
Pipeline 44991844.

PR #2702, opened 2026-02-24
[JAX] Remove GSPMD tests + adding guards and warning msg for GSPMD rules

phu0ngng:
# Description

GSPMD sharding propagation is being deprecated in favour of Shardy, which is now the default JAX partitioner. 
This commit removes all GSPMD-related tests. The GSPMD sharding propagation rules will be kept for another 3 months until June 2026.

GSPMD rules with the existing primitives will work with older JAX versions (until 0.9.1) with a printed warning.
For the incoming primitives that do not have the GSPMD rules, if users attempt to use them with GSPMD, an error will be raised before it crashes.

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2385, opened 2025-11-14
[Draft] TopK Fusion to JAX

mingxu1067:
# Description

- Adding TopK fusion to JAX for both forward and backward.

Fixes # (issue)

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Please list the changes introduced in this PR:

- Adding TopK fusion to JAX for both forward and backward.

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2711, opened 2026-02-26
[JAX] Support calling MOE router kernels from JAX side

tdophung:
# Description

Current router kernels are present in common and callable from Pytorch side but not JAX. This PR support JAX router for either standalone use or later intergation to Maxtext moe layer.

Fixes # 2710

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

1. Add custom calls to router kernels: including 
    a. `fused_topk_with_score_function_kernel`: main routing kernel that outputs a sparsed probs matrix for chosen experts + a routing map to feed in permutation. This supports 2 scoring functions: softmax and sigmoid. Also support group_topk algorithm
    b. `fused_score_for_moe_aux_loss`: step 1 of the side path to calculate auxiliary loss for load balancing. This step calculate the binary routing map, and the dense probs matrix for every experts
    c. `fused_moe_aux_loss`: step 2 of calculating auxiliary loss. This calculates the loss $$L_{\text{aux}} = C \cdot \sum_{i=1}^{E} \left(\sum_{t=1}^{N} p_{t,i}\right) \cdot f_i$$ where:
- $p_{t,i}$ = probability that token $t$ assigns to expert $i$ (from `fused_score_for_moe_aux_loss`)
- $f_i$ = number of tokens routed to expert $i$ (`tokens_per_expert[i]`, derived from `routing_map.sum(dim=0)`)
- $C = \frac{E \cdot \text{coeff}}{K \cdot T^2}$ where $T$ = `total_num_tokens`, $K$ = topk, $E$ = `num_experts`

2. Add custom partitioning for each of the above kernels when possible (sharded on the num token dimensions on the first 2 kernels, and just pure repetition on last kernel)

3. Add tests for both single GPU and distributed case to verify sharding correctness

# Checklist:

### DIFF
--- a/transformer_engine/jax/csrc/extensions/router.cpp
+++ b/transformer_engine/jax/csrc/extensions/router.cpp
@@ -201,14 +201,14 @@ Error_Type FusedMoEAuxLossBackwardFFI(cudaStream_t stream,
   auto tpe_dtype = convert_ffi_datatype_to_te_dtype(tokens_per_expert_buf.element_type());
 
   auto grad_probs_dims = grad_probs_buf->dimensions();
-  auto num_rows = static_cast<int>(grad_probs_dims[0]);
-  auto num_cols = static_cast<int>(grad_probs_dims[1]);
+  auto num_tokens = static_cast<int>(grad_probs_dims[0]);
+  auto num_experts = static_cast<int>(grad_probs_dims[1]);
 
   auto scalar_shape = std::vector<size_t>{1};
   auto tpe_dims = tokens_per_expert_buf.dimensions();
   auto tpe_shape = std::vector<size_t>{static_cast<size_t>(tpe_dims[0])};
   auto grad_probs_shape =
-      std::vector<size_t>{static_cast<size_t>(num_rows), static_cast<size_t>(num_cols)};
+      std::vector<size_t>{static_cast<size_t>(num_tokens), static_cast<size_t>(num_experts)};
 
   auto const_buf_tensor = TensorWrapper(const_buf_in.untyped_data(), scalar_shape, DType::kFloat32);
   auto tpe_tensor = TensorWrapper(tokens_per_expert_buf.untyped_data(), tpe_shape, tpe_dtype);
@@ -217,8 +217,9 @@ Error_Type FusedMoEAuxLossBackwardFFI(cudaStream_t stream,
   auto grad_probs_tensor =
       TensorWrapper(grad_probs_buf->untyped_data(), grad_probs_shape, grad_dtype);
 
-  nvte_fused_moe_aux_loss_backward(const_buf_tensor.data(), tpe_tensor.data(), num_rows, num_cols,
-                                   grad_aux_loss_tensor.data(), grad_probs_tensor.data(), stream);
+  nvte_fused_moe_aux_loss_backward(const_buf_tensor.data(), tpe_tensor.data(), num_tokens,
+                                   num_experts, grad_aux_loss_tensor.data(),
+                                   grad_probs_tensor.data(), stream);
 
   return ffi_with_cuda_error_check();
 }


tdophung:
I created this PR without being aware of https://github.com/NVIDIA/TransformerEngine/pull/2385/changes

But after reviewing it, I see that both PR are doing very similar things, I will startt with addressing @phu0ngng  comments on that PR

tests/jax/test_distributed_router.py
```diff
@@ -0,0 +1,553 @@
...
+        mesh_shape,
+        mesh_axes,
+        mesh_resource,
+        num_tokens,
+        num_experts,
+        topk,
+        score_function,
+        use_shardy,
+    ):
+        jax.config.update("jax_use_shardy_partitioner", use_shardy)
```
phu0ngng: We will remove all the GSPMD rules + tests this month.  So please remove any tests without shardy and go with it as the default.

transformer_engine/jax/cpp_extensions/router.py
```diff
@@ -0,0 +1,889 @@
...
+                num_groups=num_groups,
+                group_topk=group_topk,
+                scaling_factor=scaling_factor,
+                score_function=score_function,
+            ),
+            (logits_bdim, logits_bdim, logits_bdim),
+        )
+
+    @staticmethod
+    def infer_sharding_from_operands(
```
phu0ngng: Remove

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,353 @@
...
+    logits : jnp.ndarray
+        Logits from the gating GEMM, shape [num_tokens, num_experts].
+    topk : int
+        Number of top experts to select per token.
+    use_pre_softmax : bool
+        If True, apply softmax before top-k (only for softmax score function). Else, apply post top-k
+    num_groups : int
+        Number of groups for grouped top-k. -1 to disable.
+    group_topk : int
+        Top-k at group level. -1 to disable.
```
phu0ngng: I would prefer the default value to be 1.

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,353 @@
...
+    if expert_bias is not None and score_function != "sigmoid":
+        raise ValueError(
+            "expert_bias is only supported with score_function='sigmoid'. "
+            f"Got score_function='{score_function}'."
+        )
+
+    # Flatten to 2D if shape is [B, S, H]
+    original_shape = logits.shape
+    if logits.ndim > 2:
+        logits = logits.reshape(-1, original_shape[-1])
```
phu0ngng: I think we should do the "reshape" in C++ instead. 
In general, we try to avoid any reshaping on the JAX level as the tensor may lose the sharding pattern that triggers unexpected AllGather.

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,353 @@
...
+):
+    routing_map, intermediate_output = residuals
+    grad_probs, _ = g  # routing_map gradient is None (boolean)
+
+    grad_logits = fused_topk_with_score_function_bwd(
+        routing_map, intermediate_output, grad_probs,
+        topk, use_pre_softmax, scaling_factor, score_function,
+    )
+    # Return gradients for (logits, expert_bias)
+    return grad_logits, None
```
phu0ngng, line 238: Why is expert_bias not computed?

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,353 @@
...
+
+
+def _fused_moe_aux_loss_fwd(
+    probs, tokens_per_expert, total_num_tokens, num_experts, topk, coeff,
+):
+    num_rows = probs.shape[0]
+    num_cols = probs.shape[1]
+    aux_loss, const_buf = fused_moe_aux_loss_fwd(
+        probs, tokens_per_expert, total_num_tokens, num_experts, topk, coeff,
+    )
```
phu0ngng: `num_rows` and `num_cols` should be inferable inside the extension/primitive.

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,353 @@
...
+    _fused_topk_with_score_function_bwd,
+)
+
+
+# =============================================================================
+# Fused Score for MoE Aux Loss
+# =============================================================================
+
+
+def fused_compute_score_for_moe_aux_loss(
```
phu0ngng: Do we need a separate API for this? We should be able to use the same `fused_topk_with_score_function` API, and switch the code path to call the right NVTE API in CPP.

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,353 @@
...
+def _fused_moe_aux_loss_fwd(
+    probs, tokens_per_expert, total_num_tokens, num_experts, topk, coeff,
+):
+    num_rows = probs.shape[0]
+    num_cols = probs.shape[1]
+    aux_loss, const_buf = fused_moe_aux_loss_fwd(
+        probs, tokens_per_expert, total_num_tokens, num_experts, topk, coeff,
+    )
+    residuals = (const_buf, tokens_per_expert, num_rows, num_cols)
+    return aux_loss.squeeze(), residuals
```
phu0ngng: I wonder if we can make the primitive to return a scalar abstract value output instead of 1D vector, then we don't need to squeeze/reshape after.

tdophung: Do you mean returning the value as a scalar from the C++, or returning the value as a 1D vector from the C++, then in the JAX primitive impl, we squeeze after bind, then return scalar from here?

transformer_engine/jax/router.py
tdophung: because bias is not differentiated, the comment is misleading, I will change it

tests/jax/test_distributed_router.py
tdophung: removed

transformer_engine/jax/cpp_extensions/router.py
tdophung: I cannot remove infer_sharding_from_operands and partition() function just yet, because the BasePrimitive still requires it. I'll leave them here and we can remove them when we remove it from the BasePrimitive

transformer_engine/jax/router.py
tdophung: done

transformer_engine/jax/router.py
tdophung: done

transformer_engine/jax/router.py
tdophung: done

transformer_engine/jax/router.py
tdophung: I fused them together, also fused the primtive together , with score_for_moe_aux_loss not using expert_bias and other things while still getting the arg

transformer_engine/jax/cpp_extensions/router.py
phu0ngng: We don't want to remove the `partition()`, just `infer_sharding_from_operands()`.

Depends on whether #2702 or this PR merges first; we can rebase the other PR.

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,354 @@
...
+    Parameters
+    ----------
+    probs : jnp.ndarray
+        Probability/score tensor, shape [num_tokens, num_experts].
+    tokens_per_expert : jnp.ndarray
+        Token counts per expert, shape [num_experts]. Integer tensor.
+    total_num_tokens : int
+        Total token count for normalization.
+    num_experts : int
+        Number of experts.
```
phu0ngng: Can we also infer these two arguments from the input arrays?

transformer_engine/jax/router.py
phu0ngng: No, I mean in the abstract of the primitive here https://github.com/tdophung/TransformerEngine/blob/3fdeeef253e5496ceb6ca6de6053d9f619b0c2ba/transformer_engine/jax/cpp_extensions/router.py#L425, could we simply allocate the right output shape so that we don't need to squeeze it later?

```
aux_loss_aval = probs_aval.update(shape=(), dtype=i_dtype)
```

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,354 @@
...
+    )
+    residuals = (const_buf, tokens_per_expert, probs.shape[0], probs.shape[1])
+    return aux_loss.squeeze(), residuals
+
+
+def _fused_moe_aux_loss_bwd(
+    total_num_tokens,  # pylint: disable=unused-argument
+    num_experts,  # pylint: disable=unused-argument
+    topk,  # pylint: disable=unused-argument
+    coeff,  # pylint: disable=unused-argument
```
phu0ngng: Minor suggestion: use `del args`  instead of disabling pylint.

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,354 @@
...
+        topk,
+        use_pre_softmax,
+        num_groups,
+        group_topk,
+        scaling_factor,
+        score_function,
+        expert_bias,
+        compute_aux_scores,
+    )
+    residuals = (routing_map, intermediate_output)
```
phu0ngng: Should we rename `intermediate_output` to something more meaningful?

transformer_engine/jax/csrc/extensions/router.cpp
```diff
@@ -0,0 +1,255 @@
...
+// ============================================================================
+// Fused MoE Aux Loss - Backward
+// ============================================================================
+
+Error_Type FusedMoEAuxLossBackwardFFI(cudaStream_t stream,
+                                      Buffer_Type const_buf_in,           // scalar float32
+                                      Buffer_Type tokens_per_expert_buf,  // [num_experts]
+                                      Buffer_Type grad_aux_loss_buf,      // scalar
+                                      Result_Type grad_probs_buf,         // [num_rows, num_cols]
+                                      int64_t num_rows, int64_t num_cols) {
```
phu0ngng: Should we infer num_rows, num_cols from `grad_probs_buf`?
Same for other places. 

In general, I would recommend passing the function args only when needed to keep the API clean.

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,354 @@
...
+    const_buf, tokens_per_expert, num_rows, num_cols = residuals
+    # g is a scalar matching the squeezed output of _fwd
+    grad_aux_loss = g.reshape(1)
+
+    grad_probs = fused_moe_aux_loss_bwd(
+        const_buf,
+        tokens_per_expert,
+        grad_aux_loss,
+        num_rows,
+        num_cols,
```
phu0ngng: If this `fused_moe_aux_loss_bwd` really needs to take the `num_rows` and `num_cols` arguments, then we should rename these into something else more meaningful given the context we have.

transformer_engine/jax/csrc/extensions/router.cpp
```diff
@@ -0,0 +1,255 @@
...
+    Buffer_Type logits_buf,        // [num_tokens, num_experts]
+    Buffer_Type expert_bias_buf,   // [num_experts] or empty
+    Result_Type probs_buf,         // [num_tokens, num_experts] (or scores when compute_aux_scores)
+    Result_Type routing_map_buf,   // [num_tokens, num_experts]
+    Result_Type intermediate_buf,  // [num_tokens, num_experts]
+    int64_t topk, int64_t use_pre_softmax, int64_t num_groups, int64_t group_topk,
+    double scaling_factor, int64_t score_function, int64_t compute_aux_scores) {
+  auto dtype = convert_ffi_datatype_to_te_dtype(logits_buf.element_type());
+  auto dims = logits_buf.dimensions();
+  auto num_tokens = compute_num_tokens(dims);
```
phu0ngng: Also, we have a `product` function for this https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/jax/csrc/extensions/ffi.h#L93-L98

transformer_engine/jax/cpp_extensions/router.py
```diff
@@ -0,0 +1,752 @@
...
+    "fused_topk_with_score_function_bwd",
+    "fused_moe_aux_loss_fwd",
+    "fused_moe_aux_loss_bwd",
+]
+
+
+class ScoreFunction(IntEnum):
+    """Score function enum for fused MoE router kernels."""
+
+    SIGMOID = 0
```
jberchtold-nvidia: Would be nice to sync this with the C++ enum in router.cpp. You can look at JAXX_Scaling_Mode for how we expose an enum via pybind from C/C++ and can utilize the values to make a wrapper in Python

For example, here: https://github.com/NVIDIA/TransformerEngine/blob/bba7bf6a4101e150d1aaf9278608385214d09684/transformer_engine/jax/quantize/scaling_modes.py#L784
```
# This is the Python-level Enum
class ScalingMode(Enum):
    # The values are from JAXX_Scaling_Mode which is the pybind enum exported from C/C++, so we keep the values in sync
    NO_SCALING = JAXX_Scaling_Mode.NO_SCALING
    DELAYED_TENSOR_SCALING = JAXX_Scaling_Mode.DELAYED_TENSOR_SCALING
    MXFP8_1D_SCALING = JAXX_Scaling_Mode.MXFP8_1D_SCALING
    CURRENT_TENSOR_SCALING = JAXX_Scaling_Mode.CURRENT_TENSOR_SCALING
    NVFP4_1D_SCALING = JAXX_Scaling_Mode.NVFP4_1D_SCALING
    NVFP4_2D_SCALING = JAXX_Scaling_Mode.NVFP4_2D_SCALING

File: transformer_engine/jax/csrc/extensions/router.cpp [hunk 16]
```diff
@@ -0,0 +1,255 @@
...
+enum class ScoreFunction : int64_t {
+  kSigmoid = 0,
+  kSoftmax = 1,
+};
+
+// Compute num_tokens as the product of all dimensions except the last (num_experts).
+// Supports arbitrary-rank inputs (e.g., [B, S, E] or [num_tokens, E]).
+template <typename Dims>
+static int compute_num_tokens(const Dims &dims) {
+  int num_tokens = 1;
```
jberchtold-nvidia: Should we use size_t or int64_t instead of 32-bit int for number of tokens?

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,354 @@
...
+# =============================================================================
+
+
+def fused_topk_with_score_function(
+    logits: jnp.ndarray,
+    topk: int,
+    use_pre_softmax: bool = False,
+    num_groups: int = 1,
+    group_topk: int = 1,
+    scaling_factor: float = 1.0,
```
jberchtold-nvidia, line 67: Will this code still work if `scaling_factor` is a scalar `jnp.ndarray` of dtype float32 or a JitTracer inside jit? We had an issue in attention recently when a user passed in a JAX array scaling instead of a host-side scalar, which was easy to do because they used `jnp.sqrt` instead of regular `sqrt`. This gave a very confusing error at a much lower level internally.

If both host-side `float` and `jnp.ndarray` are allowed, that's great, nothing needs to be updated except the type signature.

If we expect host-side `float`, can you add an assertion at the top of this function `assert isinstance(scaling_mode, float), f"Expected scaling_mode to be of type float but received {type(scaling_mode)}"`

transformer_engine/jax/router.py
tdophung: We don;'t really need it, it can be inferred from the cpp file

transformer_engine/jax/router.py
tdophung: I'll rename it to saved_scores

transformer_engine/jax/csrc/extensions/router.cpp
tdophung: I now eliminated this compute_num_tokens and used product() in ffi.h per Phuong's comment

transformer_engine/jax/router.py
tdophung: I do not think it will work because this is a just. a host-side python scalar, not `jnp.ndarray`. You're right I will put the assert in

transformer_engine/jax/router.py
tdophung: Actually I revised the CUDA kernel and see that thiis cannot be done because it will trigger the wrong code path in the CUDA kernel. Let's just keep it at -1

transformer_engine/jax/router.py
tdophung: oops, I thought  I was replying to a different comment. we need this in this function because none of const_buf, tokens_per_expert, and grad_aux_loss have a dimension of num_tokens and num_experts that we can infer it from (only probs from the fwd have these dimensions). Therefore I am just going to rename them. Cannot remove

phu0ngng: If we have it in the forward path, how about putting them in the ctx/residuals and passing them to the backward?

transformer_engine/jax/router.py
```diff
@@ -0,0 +1,318 @@
...
+    del topk, coeff
+    const_buf, tokens_per_expert, num_tokens = residuals
+    grad_aux_loss = g.reshape(1)
+
+    grad_probs = fused_moe_aux_loss_bwd(
+        const_buf,
+        tokens_per_expert,
+        grad_aux_loss,
+        num_tokens,
+    )
```
tdophung, line 311: @phu0ngng num_tokens is the probs.shape[0] that was passed back via the residuals already.

Do you mean something else, like passing probs in residuals and then pass probs.shape[0] in place of num_tokens in the call to fused_moe_aux_loss_bwd?

phu0ngng: Just double-check the code again. I think the above comment is just outdated.
Could you rename the remaining num_rows and num_cols?

transformer_engine/jax/cpp_extensions/router.py
tdophung: removed in latest push

transformer_engine/jax/router.py
tdophung: I see, this is done!

transformer_engine/jax/router.py
tdophung: I renamed the remaining num_rows and num_cols in router.cpp to more meaningful names. However, we should probably do this on pytorch side too as the kernels on pytorch side is using "num_rows", "num_cols". But that's for a different PR

PR #2713, opened 2026-02-26
Add DCP compatibility for FSDP2-TP sharding in TransformerEngine.

cspades:
## Summary

- Support `(H/F)SDP2 x TP` strided sharding, and `DTensor` FP8 parameters for Torch DCP checkpointing, across all `TransformerEngineBaseModule`(s).
  - Except `GroupedLinear`, pending FSDP2 standalone pipe-cleaning. All other modules under `transformer_engine.pytorch.modules` are supported.
  - `FusibleOperation` support is also a WIP, except for `LayerNorm` or `RMSNorm` which are TE modules.
- Associated with BioNeMo-Recipes Llama3 TP: https://github.com/NVIDIA/bionemo-framework/pull/1483
  - Notably, TransformerEngine TP can be easily mixed with [`DTensor`-based TP](https://docs.pytorch.org/docs/stable/distributed.tensor.parallel.html) when unified by Torch DCP! In the Llama3 recipe, we use `DTensor`-based TP on the `torch.nn.Embedding`, TransformerEngine-based TP on the LM head, and weight-tie the LM head to the `torch.nn.Embedding`, which is why we do not need to call `set_device_mesh` for the LM head!
- Credit to @pstjohn for coming up with this idea!  

## Usage / Documentation

(`tp_mesh` and `weight_mesh` can also be passed in `TEModule.__init__`.)
```
    def set_device_mesh(
        self,
        tp_mesh: Optional[DeviceMesh] = None,
        weight_mesh: Optional[DeviceMesh] = None,
    ) -> None:
        """
        Set DeviceMesh(s) used for sharding weights and convert main weights into DTensor
        depending on the TransformerEngine class to support FSDP-TP sharding with FSDP2.

        TransformerEngine manages tensor parallel mechanics, while DTensor offers seamless
        integration with Torch DCP checkpointing. This method should only be invoked when
        using DTensor parameters, e.g. when using FSDP2 or DCP.

        When FSDP2 fully_shard() encounters any DTensor Shard(s), it will automatically
        convert them into FSDP-TP strided or non-strided shards depending on the current
        sharding dimension and factor of the DTensor. When the sharding dimension of FSDP
        matches that of TP, FSDP uses a _StridedShard placement type instead of Shard.
        This experimental FSDP-TP logic presides in this FSDP2 initialization function:
        ``torch.distributed.fsdp._fully_shard._fsdp_param._init_sharded_param``

        Parameters
        ----------
        tp_mesh : Optional[DeviceMesh]
            A 1-D DeviceMesh containing a TP mesh dimension, e.g. device_mesh["tp"].
            Only required when using TP with DTensor parameters, e.g. for FSDP2 or DCP.
        weight_mesh : Optional[DeviceMesh]
            A 1-D DeviceMesh containing a weight-sharding mesh dimension. Only required
            when using the FP8 Current (per-tensor) Scaling recipe on sharded DTensor
            parameters and if the DTensor DeviceMesh includes dimensions that do not
            shard weights, such as in the case of HSDP (DP-Replicate x DP-Shard).
            For example:
                - device_mesh["dp"] for FSDP.
                - device_mesh["dp_cp"] if using CP ranks in FSDP.
                - device_mesh["dp_shard"] if using HSDP ("dp_replicate", "dp_shard").
                - device_mesh["tp"] if using TP.
                - device_mesh["dp_cp_tp"] if strided-sharding with FSDP-TP.
        """
```

### DTensor Lifecycle in TransformerEngine

- Initialization
  - `__init__`
    - TransformerEngine model parameters are initialized either on device or `meta` device with the appropriate `tp_size` and TP sharding strategy, e.g. `parallel_mode` and `sequence_parallel`.
  - `TransformerEngineModule.set_device_mesh(tp_mesh, weight_mesh)`
    - Converts parameters to `DTensor` with appropriate TP `placement`(s) based on the TP sharding strategy specified in `__init__`, using `transformer_engine.pytorch.distributed._convert_param_to_dtensor_param`.
      - `tp_mesh` is a 1-D `DeviceMesh` containing the TP `ProcessGroup` that will be registered with the TransformerEngine module.
      - `weight_mesh` is the 1-D `DeviceMesh` containing the `ProcessGroup` that shards TransformerEngine module weights, the flattened combination of groups such as FSDP and TP. Specifically, it excludes non-weight groups such as DP-Replicate when using HSDP or HSDP-TP and is mainly required for per-Tensor scaling recipes like `Float8CurrentScaling`.
    - Needs to be invoked prior to `fully_shard` (which responds to the TP placements) and prior to `reset_parameters(defer_init=False)`, which quantizes parameters.
    - Can also be directly invoked during `__init__(tp_mesh, weight_mesh)` for supported TransformerEngine modules.
  - `fully_shard` shards the TransformerEngine model with FSDP2.
    - When `fully_shard` encounters TP sharding on `dim=0`, it will use a `_StridedShard` for DP. Put simply, this "pre-shards" the data prior to sharding on the current placement, followed by concatenating the pre-shards to get strided shards that will be re-sharded by the next placement. This effectively reverses the sharding order when processing the placements from left-to-right, and distributes shards as if we sharded on TP first, then FSDP, as required, even though DP appears before TP in the `DeviceMesh` and `DTensor.placements`. (See `Appendix` for visualization of this sharding strategy.)
  - `reset_parameters` is called if using meta device initialization.
- Training
  - **_Pre-forward_**, FSDP2 all-gathers the sharded DTensor "main" weight that it registered during `fully_shard`. (Note that this essentially shares the same properties as the compute weight besides shape, and supporting tools such as `FusedAdam` must be used to properly handle high-precision main weights.)
    - When using FSDP2 x TP, the all-gathered `Tensor` is actually a TP-sharded `DTensor`, which deviates from the original FSDP2 paradigm where the all-gathered `Tensor` is fully-unsharded and the `DTensor` wrapping is discarded. To support these `DTensor` compute weights in TransformerEngine modules, we utilize `transformer_engine.pytorch.distributed._extract_trainable_tensor_from_dtensor` to localize the `DTensor` and also inherit `requires_grad` attribute from the `DTensor` parameter as the local `Tensor` has this un-set during `DTensor.from_local(Tensor)` for FP8 parameters specifically!
  - _**Post-backward**_, the `Tensor` gradient is converted to `DTensor` and attached to the `DTensor.grad` attribute. Handled by DTensor <> Tensor Autograd conversion functions, and in the case of `FusibleOperation`, casted during the backward implementation.

### `QuantizedTensor` Storage

- When both row and column data are `None`, we send `untyped_storage()` to a default 1-byte storage that unblocks DCP checkpoint loading assertions using this as a definition for "emptiness". This is because a storage of 0 bytes is a `data_ptr() = nullptr` and breaks DCP.
  - While `untyped_storage` is not used anywhere in TransformerEngine, it may break code that uses `Storage` to figure out if a Tensor is empty or not, as now `QuantizedTensor` storage will always be a 1-byte storage even when both row and column data are not set. Those checks would instead need to compare the storage size against 1 byte instead of 0 bytes.

## Bugs

- Fix bug where `"shard"` was the presumed weight sharding sub-mesh in the `DTensor.device_mesh`. Now, users can precisely specify their own custom weight-sharding `DeviceMesh` for per-tensor `amax_reduction_group` via the `set_device_mesh(weight_mesh)` API.
- `TransformerEngineBaseModule`: `self.quantizers = {"scaling_fwd": [], "scaling_bwd": []}`

## Testing

- Megatron CI/CD Testing via Dummy PR: https://github.com/NVIDIA/Megatron-LM/pull/3661
  - Current: Just a lint failure in MBridge, all tests pass. ✅ 
- Megatron-LM Llama 8B (TP=4) Parity Tests: `main` vs. `cspades:cye/fsdp2-tp-dcp` with Megatron-LM `main` on PyTorch `25.11`
```
# TransformerEngine Main
[Rank 0] (after 1 iterations) memory (MB) | allocated: 23511.65 | max allocated: 25189.68 | reserved: 25678.00 | max reserved: 25678.00
 [2026-03-02 09:55:17.189564] iteration       99/15258789 | consumed samples:        12672 | elapsed time per iteration (ms): 12715.7 | throughput per GPU (TFLOP/s/GPU): 530.6 | learning rate: 4.866046E-07 | global batch size:   128 | lm loss: 1.124915E+00 | loss scale: 1.0 | grad norm: 5.474 | number of skipped iterations:   0 | number of nan iterations:   0 |
 [2026-03-02 09:55:29.768521] iteration      100/15258789 | consumed samples:        12800 | elapsed time per iteration (ms): 12578.7 | throughput per GPU (TFLOP/s/GPU): 536.4 | learning rate: 4.915198E-07 | global batch size:   128 | lm loss: 1.143806E+00 | loss scale: 1.0 | grad norm: 5.366 | number of skipped iterations:   0 | number of nan iterations:   0 |

# Post-DCP Modifications (This PR)
[Rank 0] (after 2 iterations) memory (MB) | allocated: 23511.65 | max allocated: 29783.24 | reserved: 25678.00 | max reserved: 31510.00
 [2026-03-02 09:29:36.550070] iteration       99/15258789 | consumed samples:        12672 | elapsed time per iteration (ms): 12556.5 | throughput per GPU (TFLOP/s/GPU): 537.3 | learning rate: 4.866046E-07 | global batch size:   128 | lm loss: 1.124463E+00 | loss scale: 1.0 | grad norm: 5.471 | number of skipped iterations:   0 | number of nan iterations:   0 |
 [2026-03-02 09:29:49.216068] iteration      100/15258789 | consumed samples:        12800 | elapsed time per iteration (ms): 12665.7 | throughput per GPU (TFLOP/s/GPU): 532.7 | learning rate: 4.915198E-07 | global batch size:   128 | lm loss: 1.142863E+00 | loss scale: 1.0 | grad norm: 5.355 | number of skipped iterations:   0 | number of nan iterations:   0 |
```
- NOTE(@cspades): `DelayedScaling` has DCP save/load disparity issues, i.e. on the scale of `+/-1` to the `uint8` parameter checkpoint!

### `_StridedShard` - Using FSDP2 x TP Strided-Sharding

```
# (DP=4, TP=2)
(_StridedShard(dim=0, sf=2), Shard(dim=0))

┌───┬───┐
│ 0 │ 4 │ ← DP=0
├───┼───┤
│ 1 │ 5 │ ← DP=1
├───┼───┤          FSDP all-gather happens across the DP ranks,
│ 2 │ 6 │ ← DP=2   so we need to form the 0-3 and 4-7 TP shards!
├───┼───┤
│ 3 │ 7 │ ← DP=3
└───┴───┘
  ↑   ↑
TP=0 TP=1
```

When `redistribute`'ing a global DTensor to `(_StridedShard(dim=0, sf=2), Shard(dim=0))`, `DTensor` will perform the following steps:

- Pre-shard the Tensor data with respect to the stride / shard factor, which is defined as the product of the parallelism sizes of all `Shard` placements to the right of `_StridedShard`. (In the above example, since TP=2, the factor is 2.)
  - `[0 1 2 3 4 5 6 7] -> [0 1 2 3] and [4 5 6 7]`.
  - In the context of this PR and `fully_shard`, this has already been done via initializing the TransformerEngine module with TP and calling `_convert_param_to_dtensor_param`!
- Shard the pre-shards for `_StridedShard`.
  - `[0] [1] [2] [3]` and `[4] [5] [6] [7]`
- Concatenate the strided shards.
  - `[0 4] [1 5] [2 6] [3 7]`, which are assigned to the `_StridedShard` ranks.
  - Note that this is very different if we did left-to-right-sharding, which would have given us `[0 1] [2 3] [4 5] [6 7]`!
- Subsequently / finally, each strided shard is sharded on the `Shard` placement.
  - `[0] [4]` / `[1] [5]` / `[2] [6]` / `[3] [7]`, which are assigned to the `Shard` ranks.
  - Note that this is very different if we did left-to-right sharding, which would have given us `[0] [1]` / `[2] [3]` / `[4] [5]` / `[6] [7]`!

PyTorch also supports the inverse / un-sharding of this `redistribute`, which is literally the inverse of these simple operations! (Though things get a bit more complicated with un-even shards from odd-numbered dimension sizes.)

### 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/distributed/run_fsdp2_allgather.py
+++ b/tests/pytorch/distributed/run_fsdp2_allgather.py
@@ -32,9 +32,9 @@ LOCAL_RANK = None
 
 # Fixed model dimensions — this test focuses on allgather correctness, not model flexibility.
 _NUM_HEADS = 8
-_HEAD_DIM = 64
-_HIDDEN_SIZE = _NUM_HEADS * _HEAD_DIM  # 512
-_FFN_SIZE = _HIDDEN_SIZE * 4  # 2048
+_HEAD_DIM = 128
+_HIDDEN_SIZE = _NUM_HEADS * _HEAD_DIM
+_FFN_SIZE = _HIDDEN_SIZE * 4
 _NUM_LAYERS = 2
 _BATCH_SIZE = 4
 _SEQ_LEN = 32
--- a/tests/pytorch/distributed/run_fsdp2_allgather.py
+++ b/tests/pytorch/distributed/run_fsdp2_allgather.py
@@ -1,0 +1,240 @@
+#!/usr/bin/python3
+
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""
+Standalone test for FP8 FSDP2 all-gather correctness.
+
+Verifies that FSDP2's internal all-gather of FP8 parameters produces the same
+result as a manual all-gather of dequantized FP32 values.
+"""
+
+import argparse
+import os
+import sys
+from contextlib import nullcontext
+
+import transformer_engine.pytorch as te
+import transformer_engine.common.recipe
+from transformer_engine.pytorch import fp8_model_init
+import torch
+import torch.distributed as dist
+import torch.nn.functional as F
+from torch import optim
+from torch.distributed.tensor import DTensor
+from torch.distributed._composable.fsdp import fully_shard
+from torch.distributed.device_mesh import init_device_mesh
+from torch import nn
+
+LOCAL_RANK = None
+
+# Fixed model dimensions — this test focuses on allgather correctness, not model flexibility.
+_NUM_HEADS = 8
+_HEAD_DIM = 64
+_HIDDEN_SIZE = _NUM_HEADS * _HEAD_DIM  # 512
+_FFN_SIZE = _HIDDEN_SIZE * 4  # 2048
+_NUM_LAYERS = 2
+_BATCH_SIZE = 4
+_SEQ_LEN = 32
+
+
+def dist_print(msg):
+    if LOCAL_RANK == 0:
+        print(msg)
+
+
+def _parse_args():
+    parser = argparse.ArgumentParser(
+        description="Test FP8 FSDP2 all-gather correctness with TransformerLayer."
+    )
+    parser.add_argument(
+        "--recipe",
+        type=str,
+        default="DelayedScaling",
+        choices=[
+            "DelayedScaling",
+            "Float8CurrentScaling",
+            "Float8BlockScaling",
+            "MXFP8BlockScaling",
+            "NVFP4BlockScaling",
+        ],
+    )
+    parser.add_argument(
+        "--sharding-dims",
+        type=int,
+        nargs="+",
+        required=True,
+        help=(
+            'Sharding mesh dimensions: ("dp_shard",), ("dp_replicate", "dp_shard"), '
+            'or ("dp_replicate", "dp_shard", "tp")'
+        ),
+    )
+    parser.add_argument("--seed", type=int, default=42)
+    args = parser.parse_args()
+    assert len(args.sharding_dims) <= 3
+    args.tp_size = args.sharding_dims[2] if len(args.sharding_dims) >= 3 else 1
+    return args
+
+
+def _get_recipe(name):
+    return getattr(transformer_engine.common.recipe, name)()
+
+
+def _get_device_mesh(world_size, sharding_dims):
+    dist_print(f"sharding-dims: {sharding_dims}")
+    if len(sharding_dims) == 1:
+        assert sharding_dims[0] == world_size
+        return init_device_mesh("cuda", (world_size,), mesh_dim_names=("dp_shard",))
+    elif len(sharding_dims) == 2:
+        assert sharding_dims[0] * sharding_dims[1] == world_size
+        return init_device_mesh(
+            "cuda",
+            (sharding_dims[0], sharding_dims[1]),
+            mesh_dim_names=("dp_replicate", "dp_shard"),
+        )
+    else:
+        assert sharding_dims[0] * sharding_dims[1] * sharding_dims[2] == world_size
+        return init_device_mesh(
+            "cuda",
+            (sharding_dims[0], sharding_dims[1], sharding_dims[2]),
+            mesh_dim_names=("dp_replicate", "dp_shard", "tp"),
+        )
+
+
+def _build_model(args):
+    kwargs = {
+        "params_dtype": torch.float32,
+        "device": "meta",
+        "tp_size": args.tp_size,
+        "fuse_qkv_params": True,
+    }
+    if args.tp_size > 1:
+        kwargs["tp_mesh"] = args.mesh["tp"]
+        kwargs["weight_mesh"] = args.mesh["dp_shard", "tp"]._flatten("weight_mesh")
+        kwargs["set_parallel_mode"] = True
+    elif "dp_replicate" in args.mesh.mesh_dim_names:
+        kwargs["weight_mesh"] = args.mesh["dp_shard"]
+
+    model = nn.Sequential(
+        *[
+            te.TransformerLayer(_HIDDEN_SIZE, _FFN_SIZE, _NUM_HEADS, **kwargs)
+            for _ in range(_NUM_LAYERS)
+        ]
+    )
+    inp_shape = [_SEQ_LEN, _BATCH_SIZE, _HIDDEN_SIZE]
+    return model, inp_shape
+
+
+def _shard_model(model, mesh):
+    dp_dims = (
+        ("dp_replicate", "dp_shard") if "dp_replicate" in mesh.mesh_dim_names else ("dp_shard",)
+    )
+    for child in model.children():
+        fully_shard(child, mesh=mesh[dp_dims])
+    fully_shard(model, mesh=mesh[dp_dims])
+    return model
+
+
+@torch.no_grad()
+def _test_fp8_fsdp2_allgather(model):
+    """
+    Compare the result of the FP8 AG by FSDP2 with a manual AG in FP32
+    after dequantizing the FP8 values.
+    """
+    # FP32 manual weight allgather
+    fp32_allgathered_params = {}
+    for name, param in model.named_parameters():
+        assert isinstance(
+            param, DTensor
+        ), f"[test_fp8_fsdp2_allgather] {param} should be a DTensor."
+        local_tensor = param._local_tensor
+        device_mesh = param.device_mesh
+        dist_group = (
+            device_mesh.get_group(mesh_dim="dp_shard")
+            if device_mesh.ndim > 1
+            else device_mesh.get_group()
+        )
+        # Perform manual allgather on local_tensor. zeros_like will create hp tensor since torch_dispatch
+        # for local_tensor will go down the dequantization route.
+        gathered_tensor = [
+            torch.zeros_like(local_tensor) for _ in range(dist.get_world_size(group=dist_group))
+        ]
+        dist.all_gather(gathered_tensor, local_tensor.dequantize(), group=dist_group)
+        full_tensor = torch.cat(gathered_tensor, dim=0)
+        fp32_allgathered_params[name] = full_tensor
+    # FP8 allgather using FSDP2
+    for module in model.modules():
+        # Not all modules are wrapped/sharded with FSDP2.
+        if hasattr(module, "unshard"):
+            module.unshard()
+    # Make sure allgathered parameters match exactly
+    for name, param in model.named_parameters():
+        if isinstance(param, DTensor):
+            # Will still be a DTensor in the case of TP, even after FSDP2 AG,
+            # because we wrap our weights as DTensor shards over the TP group.
+            param = param._local_tensor
+        torch.testing.assert_close(param.dequantize(), fp32_allgathered_params[name])
+    # Revert model to original sharded state
+    for module in model.modules():
+        # Not all modules are wrapped/sharded with FSDP2.
+        if hasattr(module, "reshard"):
+            module.reshard()
+
+
+def _main(args):
+    global LOCAL_RANK
+    assert "TORCHELASTIC_RUN_ID" in os.environ
+    WORLD_RANK = int(os.getenv("RANK", "0"))
+    WORLD_SIZE = int(os.getenv("WORLD_SIZE", "1"))
+    LOCAL_RANK = int(os.getenv("LOCAL_RANK", "0"))
+
+    torch.cuda.set_device(WORLD_RANK)
+    torch.manual_seed(args.seed)
+    torch.cuda.manual_seed(args.seed)
+
+    dist.init_process_group(backend="nccl", rank=WORLD_RANK, world_size=WORLD_SIZE)
+    device = torch.device(f"cuda:{LOCAL_RANK}")
+
+    mesh = _get_device_mesh(WORLD_SIZE, args.sharding_dims)
+    args.mesh = mesh
+
+    fp8_recipe = _get_recipe(args.recipe)
+
+    with fp8_model_init(enabled=True, recipe=fp8_recipe):
+        model, inp_shape = _build_model(args)
+
+    model = _shard_model(model, mesh)
+
+    for module in model.modules():
+        if hasattr(module, "reset_parameters"):
+            module.reset_parameters()
+
+    # Run a training step to initialize FSDP2 lazy state and update quantization
+    # scales before testing the allgather. Block-scaling formats (Float8BlockScaling,
+    # NVFP4BlockScaling) only exhibit allgather inconsistencies after weight updates.
+    input_data = torch.randn(inp_shape, device=device)
+    target = torch.randn(inp_shape, device=device)
+    nvfp4_ctx = (
+        torch.autocast(device_type="cuda", dtype=torch.bfloat16)
+        if args.recipe == "NVFP4BlockScaling"
+        else nullcontext()
+    )
+    optimizer = optim.Adam(model.parameters(), lr=1e-3)
+    optimizer.zero_grad()
+    with nvfp4_ctx, te.autocast(enabled=True, recipe=fp8_recipe):
+        output = model(input_data)
+        loss = F.mse_loss(output, target)
+    loss.backward()
+    optimizer.step()
+
+    _test_fp8_fsdp2_allgather(model)
+    dist_print("test_fp8_fsdp2_allgather passed.")
+
+    dist.destroy_process_group()
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(_main(_parse_args()))
--- a/tests/pytorch/distributed/run_fsdp2_model.py
+++ b/tests/pytorch/distributed/run_fsdp2_model.py
@@ -596,9 +596,10 @@ def _train(args):
         with te.autocast(enabled=True, recipe=fp8_recipe):
             output = model(input_data)
             post_load_loss = F.mse_loss(output, target)
-    # Allow for 1% disparity due to _extra_state disparity.
+
+    # FIXME(@cspades): Investigate and improve 10% loss disparity from DCP.
     assert torch.allclose(
-        pre_save_loss, post_load_loss, rtol=5e-2
+        pre_save_loss, post_load_loss, rtol=0.1
     ), f"Pre-Save Loss: {pre_save_loss} != Post-Load Loss: {post_load_loss}"
 
     # Clean up temporary checkpoint directory.
--- a/tests/pytorch/distributed/run_fsdp2_model.py
+++ b/tests/pytorch/distributed/run_fsdp2_model.py
@@ -7,12 +7,20 @@
 import os
 import sys
 import argparse
+from dataclasses import dataclass
 
 import transformer_engine.pytorch as te
 import transformer_engine.common.recipe
 
 import torch
 import torch.distributed as dist
+from torch.distributed.checkpoint import save, load
+from torch.distributed.checkpoint.state_dict import (
+    StateDictOptions,
+    get_state_dict,
+    set_state_dict,
+)
+from torch.distributed.checkpoint.stateful import Stateful
 from torch.distributed.tensor import DTensor
 import torch.nn.functional as F
 from torch import nn, optim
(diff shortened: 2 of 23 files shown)


tests/pytorch/distributed/run_fsdp2_model.py
```diff
@@ -30,6 +38,61 @@
...
+            model_state_dict=state_dict["model"],
+            optim_state_dict=state_dict["optim"],
+            # Non-strict checkpoint loading ignores empty optimizer states,
+            # skips loading non-FP8 checkpoint weights (e.g. _extra_state).
+            options=StateDictOptions(strict=False),
+        )
+
+
 def dist_print(msg):
     if LOCAL_RANK == 0:
```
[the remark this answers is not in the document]
cspades: Working on it! + GroupedLinear test case.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -786,30 +795,129 @@ def make_grouped_weights(self, defer_init=False) -> None:
...
+            # Convert to DTensor with properties equivalent to the original DTensor.
+            dtensor_member_param = getattr(self, "weight0")
+            grouped_param = _convert_param_to_dtensor_param(
+                grouped_param,
+                device_mesh=dtensor_member_param.device_mesh,
+                placements=dtensor_member_param.placements,
+                # DTensor / DCP will view this as a TP-sharded 3-D Tensor.
+                shape=(self.num_gemms, self.out_features, self.in_features),
+                # Default Stride: (out*in, in, 1)
+                stride=None,
```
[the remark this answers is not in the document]
cspades: Okay, this is an impressive catch. I wrote this code too quickly. Will fix!

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -1997,20 +2019,93 @@ def reset_layer_norm_parameters(self) -> None:
...
+                self.fc1_bias, tp_mesh, placements=(Shard(dim=0),)
+            )
+            # FC2 Weight -> Row-Parallel -> Shard(dim=1)
+            self.fc2_weight = _convert_param_to_dtensor_param(
+                self.fc2_weight, tp_mesh, placements=(Shard(dim=1),)
+            )
+            # LN & FC2 Bias -> Replicate()
+            self.fc2_bias = _convert_param_to_dtensor_param(
+                self.fc2_bias, tp_mesh, placements=(Replicate(),)
+            )
```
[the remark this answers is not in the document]
cspades: Not going to change TE conventions in this PR. Can be done in a different PR.

tests/pytorch/distributed/run_fsdp2_model.py
```diff
@@ -346,6 +507,131 @@ def _train(args):
...
+    # Perform an extra training step to change the weights such that
+    # state parity tests will fail unless the checkpoint is loaded
+    # without any errors or incongruities vs. the saved model state.
+    model.train()
+    for iteration in range(args.iter):
+        with te.autocast(enabled=True, recipe=fp8_recipe):
+            output = model(torch.randn(inp_shape).to(device))
+        loss = F.mse_loss(output, torch.randn(out_shape).to(device))
+        loss.backward()
+        optimizer.step()
```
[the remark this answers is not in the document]
cspades: Intended, I want to apply some nasty gradients to change the model weights without re-initializing a new model. Much easier this way.

cspades:
/te-ci L1 pytorch

https://gitlab-master.nvidia.com/dl/transformerengine/transformerengine/-/pipelines/46350515

cspades:
/te-ci L1 pytorch

https://gitlab-master.nvidia.com/dl/transformerengine/transformerengine/-/pipelines/46358818

cspades:
For some reason after 2.3k training steps, I start to get NaNs: https://wandb.ai/nvidia/bionemo-recipes/runs/nmzugu0a?nw=nwusercye_nv

Restarting from this checkpoint and around 500 steps later same thing.

tests/pytorch/distributed/run_fsdp2_allgather.py
```diff
@@ -0,0 +1,240 @@
...
+        model, inp_shape = _build_model(args)
+
+    model = _shard_model(model, mesh)
+
+    for module in model.modules():
+        if hasattr(module, "reset_parameters"):
+            module.reset_parameters()
+
+    # Run a training step to initialize FSDP2 lazy state and update quantization
+    # scales before testing the allgather. Block-scaling formats (Float8BlockScaling,
```
vthumbe1503, line 215: I believe Float8Blockscaling allgather should work now right?

tests/pytorch/distributed/run_fsdp2_allgather.py
```diff
@@ -0,0 +1,240 @@
...
+    # NVFP4BlockScaling) only exhibit allgather inconsistencies after weight updates.
+    input_data = torch.randn(inp_shape, device=device)
+    target = torch.randn(inp_shape, device=device)
+    nvfp4_ctx = (
+        torch.autocast(device_type="cuda", dtype=torch.bfloat16)
```
vthumbe1503, line 220: Why seperate nvfp4 context? In general, adding multiple context manager adds CPU overheads in the training loop.

tests/pytorch/distributed/run_fsdp2_model.py
```diff
@@ -4,26 +4,95 @@
...
 
 LOCAL_RANK = None
 
+# Needed for `torch.distributed.checkpoint.{save,load}` because
+# multiple processes need to write to the same directory.
+SHARED_TMP_DIR = "/tmp/pytest-shared-tmp"
+
+
+@dataclass
+class AppState(Stateful):
```
vthumbe1503, line 43: This seems like a useful class. With some things like extra state specific to TE. Might make sense to move it TE distributed  module. Thoughts @cspades ?

tests/pytorch/distributed/run_fsdp2_model.py
```diff
@@ -4,26 +4,95 @@
...
+                model_param = model_param.to_local()
+            if model_param.numel() == 0 and fqn in optimizer_state_dict["state"]:
+                # Empty model parameter. Clear the associated optimizer state
+                # when initializing the optimizer state upon DCP load, because
+                # empty optimizer state DTensors are not checkpointed with DCP,
+                # yet get_state_dict / _init_optim_state produce empty Tensors.
+                # TransformerEngine uses empty Tensors for dummy Parameters.
+                optimizer_state_dict["state"][fqn] = {}
+            if fqn.endswith("_extra_state"):
+                # Evict `_extra_state` quantization data from model checkpoint.
```
vthumbe1503, line 73: If this is evicted, how do we make sure it is updated correctly after load from checkpoint?

tests/pytorch/distributed/test_torch_fsdp2.py
```diff
@@ -64,35 +65,86 @@ def fp_recipe(request):
...
+
+
+@pytest.mark.skipif(NUM_PROCS % 2 != 0, reason="Requires even number of GPUs.")
+@pytest.mark.skipif(not te.torch_version() >= (2, 4, 0), reason="Requires PyTorch 2.4.0+")
+@pytest.mark.parametrize(
+    "sharding_dims",
+    (
+        # FSDP
+        [NUM_PROCS],
+        # HSDP
```
vthumbe1503, line 128: Does this work ok if NUM_PROCS < 4? i.e lets say NUM_PROCS = 2. TP dimension will be 0. Curious what happens in that case?

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -477,6 +488,10 @@ def __init__(
             return_max_logit=self.return_max_logit,
         )
 
+        if tp_mesh is not None or weight_mesh is not None:
+            # Apply DeviceMesh and DTensor-related modifications.
+            self.set_device_mesh(tp_mesh=tp_mesh, weight_mesh=weight_mesh)
```
vthumbe1503, line 493: in set_device_mesh function, it says weight_mesh is not necessary, but we call it only if both tp_mesh and weight_mesh is not None. So it should not include the condition weight_mesh is not none right?

transformer_engine/pytorch/module/base.py
```diff
@@ -1326,14 +1332,15 @@ def reset_parameters(self, defer_init: Optional[bool] = False) -> None:
...
+                    # If the amax_reduction_group is not set by `set_device_mesh`,
+                    # then default to the DTensor's full DeviceMesh.
                     device_mesh = dtensor_param.device_mesh
-                    amax_reduction_group = (
-                        device_mesh.get_group(mesh_dim="shard")
-                        if device_mesh.ndim > 1
-                        else device_mesh.get_group()
-                    )
-                    quantizer.amax_reduction_group = amax_reduction_group
+                    quantizer.amax_reduction_group = device_mesh.get_group()
```
vthumbe1503, line 1343: Which group will it return in case of multiple dimensions? For instance if weight is both FSDP-TP sharded then will this give the FSDP dim or TP dim?

transformer_engine/pytorch/ops/basic/rmsnorm.py
```diff
@@ -216,13 +219,16 @@ def op_backward(
         x, rstdevs = ctx.saved_tensors
 
         # Tensor dims
-        weight_dims = self.weight.size()
+        weight = self.weight
+        if isinstance(weight, DTensor):
+            weight = weight.to_local()
```
vthumbe1503, line 224: Shouldnt we use _extract_trainable_tensor_from_dtensor here?

vthumbe1503: Also applicable to couple in other places in ops folder

transformer_engine/pytorch/tensor/storage/mxfp8_tensor_storage.py
```diff
@@ -99,6 +101,7 @@ def __new__(
         instance._quantizer = quantizer.copy() if quantizer is not None else None
         instance._fp8_dtype = fp8_dtype
         instance._with_gemm_swizzled_scales = with_gemm_swizzled_scales
+        instance._default_storage = torch.UntypedStorage(1, device=torch.cuda.current_device())
```
vthumbe1503, line 104: I am leaning towards creating this default_storage on CPU instead due to coupe of reasons
1. Since Idea of default_storage here is to show unique identity, keeping it on CPU/GPU shouldnt matter
2. Calling torch cuda current device on every single Tensor creation has python overheads)

transformer_engine/pytorch/module/rmsnorm.py
```diff
@@ -141,6 +142,47 @@ def reset_parameters(self, defer_init: Optional[bool] = None) -> None:
...
+    ) -> None:
+        """
+        Set DeviceMesh(s) used for sharding weights and convert main weights into DTensor
+        depending on the TransformerEngine class to support FSDP-TP sharding with FSDP2.
+
+        TransformerEngine manages tensor parallel mechanics, while DTensor offers seamless
+        integration with Torch DCP checkpointing. This method should only be invoked when
+        using DTensor parameters, e.g. when using FSDP2 or DCP.
+
+        When FSDP2 fully_shard() encounters any DTensor Shard(s), it will automatically
```
vthumbe1503, line 158: Why is it that we havent added tp_mesh and weight_mesh to the constrictors of rmsnorm and layer_norm? But for every other layer we have?

transformer_engine/pytorch/module/linear.py
```diff
@@ -1315,24 +1338,93 @@ def set_meta_tensor(self, fwd: bool, recipe: Recipe) -> None:
...
+                setattr(
+                    self,
+                    weight,
+                    _convert_param_to_dtensor_param(param, tp_mesh, placements=placements),
+                )
+            for bias in self.bias_names:
+                param = getattr(self, bias)
+                placements = (Replicate(),)
+                if self.parallel_mode == "column":
+                    placements = (Shard(dim=0),)
```
vthumbe1503, line 1408: I am wondering if we can make all the set_device_mesh function share a helper `set_tp_mesh`
 defined in base.py that takes in a dictionary of `{parameter name: parallel_mode}` and tp_mesh that converts the parameters to Dtensors and use that in set_device_mesh of each module?

Something like this
`def set_tp_mesh(self, param_mode_dict: dict, tp_mesh:  Optional[DeviceMesh]):`

vthumbe1503: And put the big docstring that we have over there in base.py
. The docstring seems to be repeated in all places.

vthumbe1503:
Generally LGTM @cspades. Lets get it merged after comments are addressed.

tests/pytorch/distributed/run_fsdp2_allgather.py
cspades: I think so, this comment might be out of date - I believe I only xFail the NVFP4 test instance.

tests/pytorch/distributed/run_fsdp2_allgather.py
cspades: Copypasta from @pstjohn 's FusedAdam PR - Peter can you remind me what was the need for this double-context? Thanks!

tests/pytorch/distributed/run_fsdp2_model.py
cspades: Yup, also add a TODO for both of those two hacks:

- The empty param hack is mainly for biases, since TE uses an empty Tensor for some layers. In general, I believe it would be much cleaner to just... set the `self.bias` or multiple biases all to `None` but I think there must have been a reason for empty Tensors in the first place. (And.... it's in a lot of people's checkpoints by now...)
- `_extra_state` is immensely confusing due to its serialization, I've seen it work and not work, or change shape if you load a brand new model vs. load into a pre-trained model (i.e. re-load after running a few training steps). I hope this is an acceptable throwaway!

tests/pytorch/distributed/run_fsdp2_model.py
cspades: We don't, there's a pretty nasty 10% loss difference for specific test cases and it scales poorly with the model size, bigger model, the bigger the disparity.

I'm assuming it is caused by not checkpointing extra state, but as mentioned in the above comment, I need some help understanding how to support it. 🙏🏻

tests/pytorch/distributed/test_torch_fsdp2.py
cspades: Right, I have this filter below:
```
    parallel_size = math.prod(x for x in sharding_dims if x != 0)
    if NUM_PROCS < parallel_size:
        pytest.skip(
            f"Insufficient devices ({NUM_PROCS}) to test sharding configuration: {sharding_dims}"
        )
```
If you want this test to run with 2 GPUs, maybe we can add a test case that tests pure TP, no FSDP? Maybe just have a separate test module for this! (It'll be easy to deal with without FSDP2 annoyances and order-of-operations.)

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
cspades: Actually OR logic:
```
tp_mesh is not None or weight_mesh is not None
```
Your intuition is correct - `set_device_mesh` handles both arguments completely separately, and if one or the other is `None`, it won't do anything for that. It should be a `switch / case` function for DTensors.

transformer_engine/pytorch/module/base.py
cspades: This will actually error out if your device mesh is not pure FSDP (or just 1 dimension).

This is the backwards-incompatibility I mentioned a long while back, we can maybe hack in a WAR to support the old API (like maybe, pick the first mesh dimension, but obviously it's a very bad hack) but the right way is to expose this publicly so the user can tell us what the weight sharding mesh is!

transformer_engine/pytorch/module/rmsnorm.py
cspades: Like the ops these modules call? Or that we don't TP shard norms so it's just `Replicate()` and there's no need to set the weight-sharding mesh.

Now that I think about it, maybe a loophole where you make DTensors but the amax reduction group logic will take the `Replicate()` mesh, but I think it won't matter if you reduce on a larger group if this norm weight is replicated across GPUs. (It will affect non-max / non-mean reductions though, like sum. Should be an obvious way to do this in TEBaseModule.)

transformer_engine/pytorch/tensor/storage/mxfp8_tensor_storage.py
cspades: Agreed. I also want us to confirm with @timmoon10 if maybe `nullptr` or empty storage "works". Or otherwise, if a 1-byte default TE Tensor storage size is acceptable or breaks our user's logic.

I believe Torch DCP checkpoint load/save do use `untyped_storage()` but technically if our row-wise and col-wise data is None, then maybe the `data_ptr() > 0` check _correctly_ does not checkpoint the weight? Or do we want to checkpoint the weight anyway?

cspades (commented):
Answered all comments, they were all good questions.

transformer_engine/pytorch/module/linear.py
cspades: Wait, every module's TP mesh logic is different. They all have different `Shard()` and `Replicate()` patterns, the placement implementations are not common to every TEBaseModule!

transformer_engine/pytorch/tensor/storage/mxfp8_tensor_storage.py
cspades: Wait, one reason I put it on the same device (GPU) is because I am not 100% sure if we want DCP to write QuantizedTensors' data to different devices depending on the state of QuantizedTensor.

Maybe that works, since it looks like TE quantizes on-the-fly, on the right devices, so we can just assume that this storage always belongs on CPU.

transformer_engine/pytorch/ops/basic/rmsnorm.py
cspades: Oh, this one just seems to work fine, but I think you are right. This should definitely use that utility, there is no reason not to.

transformer_engine/pytorch/tensor/storage/mxfp8_tensor_storage.py
timmoon10: I guess we can't get around the storage logic being ad hoc and delicate, but it seems putting the storage on CPU is too much. To reduce CPU overhead during initialization, how about we construct the storage lazily?
```python
def untyped_storage(self):
    ...

    # Return dummy storage if there is no local data
    if self._default_storage is None:
        self._default_storage = torch.UntypedStorage(1, device=self.device)
    return self._default_storage
```

transformer_engine/pytorch/module/linear.py
vthumbe1503: Isnt the logic of selecting Shard vs Replicate and dimension decidable based on parallel_mode and the type of parameter? It seemed like that to me.

cspades: Just saw your edits, yeah if you can find a way to generalize the TP setup for all modules go for it!

transformer_engine/pytorch/tensor/storage/mxfp8_tensor_storage.py
cspades: SGTM

nvMelissa:
chatted with @cspades . Expected to come back to push this in the next 2 weeks. ETA: 9/24/26

PR #2718, opened 2026-02-28
pass params_dtype to qk_norm creation

pstjohn:
Previously layers would fail with 
```
            assert (
>               query_layer.dtype == key_layer.dtype and query_layer.dtype == value_layer.dtype
            ), "Queries, keys and values must have the same data type!"
E           AssertionError: Queries, keys and values must have the same data type!

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py:1063: AssertionError
```

if you created a layer with dtype != float32. This ensures the dtype of the layernorm layers match those of the base attention layer.

### DIFF
--- a/examples/jax/collective_gemm/common.py
+++ b/examples/jax/collective_gemm/common.py
@@ -131,10 +131,6 @@ def _initialize_distributed(args):
     )
 
     _distributed_initialized = True
-    jax.clear_caches()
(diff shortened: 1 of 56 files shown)


ksivaman (approved):
LGTM

negvet (approved):

PR #2719, opened 2026-03-01
Add MXFP8 attention

cyanguwa:
# Description

1. Added MXFP8 support in `FusedAttention` backend (fwd+bwd, BSHD/SBHD, TE-PyTorch)
2. Decoupled input/output format in APIs by introducing `o_format`,  `do_format`, and `dqkv_layout`
3. Implemented `mxfp8_quantize_fast_path()` to quantize multiple tensors, pad/permute/swizzle the `scale_inv`s for a more efficient quantization pipeline; added `qkv_scale_inv_format`, `do_scale_inv_format` to indicate `scale_inv`s' format
4. Implemented `multi_tensor_transpose_to_bhsd()` to permute tensors from BSHD/SBHD to BHSD, with the TMA path optimized for FP16/BF16 dtype, D=192/128 cases, `fallback_vec_aligned` path for Byte, D=8/4, and a `fallback_non_vec_aligned` path for Byte, D=6
5. Implemented `multi_tensor_pad_last_dim()` to pad multiple tensors' D to %4 for rowwise and %128 for columnwise
6. Implemented `multi_tensor_swizzle_row_scaling_narrow_k_kernel` and `multi_tensor_swizzle_col_scaling_narrow_m_kernel ` to optimize for small K/M dimensions
7. Added `MXFP8PaddedSizes`, `pad_s_d_for_mxfp8`, `generateMatrixStridesWithFormat`, `generateMatrixStridesWithLayout` for size/stride computation
9. Fixed bug in O/dQKV shape logic for MLA; added utility `nvte_convert_qkv_shape`
10. Refactored CP and added support for MXFP8 in `cp_comm_type={'a2a', 'p2p', 'a2a+p2p', 'all_gather'}`, with `{'a2a', 'p2p', 'a2a+p2p', 'all_gather'}` for MHA/GQA/MQA/MLA, `{'a2a', 'all_gather'}` for SWA, and `{'a2a'}` for sink attention
11. Fixed `scale_inv_offsets` calculation in `GroupedTensor`
12. MXFP8 attention requires cudnn-frontend v1.22+ and cuDNN 9.21+

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Please see Description.

# Checklist:


transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -1083,11 +1167,14 @@ def cp_p2p_bwd_fused_attn(
 ):
     """Per-tile backward call of CP P2P with FusedAttention backend"""
     if fp8:
-        aux_tensors = [
-            softmax_lse,
-            softmax_lse,
-            rng_states[cp_size - step - 1],
-        ]
+        if qkv_layout == "t3hd":
```
cyanguwa, line 1173: @sudhakarsingh27 keep this in mind when you work on FP8 impl_v0 removal. Thanks!

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -188,6 +196,22 @@ NVTE_QKV_Format nvte_get_q_format(NVTE_QKV_Layout qkv_layout);
...
+ *  \param[out]    dst_shape           The destination shape.
+ *  \param[out]    b                   The batch size.
+ *  \param[out]    h                   The number of heads.
+ *  \param[out]    s                   The sequence length.
+ *  \param[out]    d                   The head dimension.
+ *  \param[out]    t                   The number of tokens.
+ */
+void nvte_convert_qkv_format(NVTE_QKV_Format src_format, std::vector<size_t> src_shape,
+                             NVTE_QKV_Format dst_format, std::vector<size_t> &dst_shape, size_t *b,
+                             size_t *h, size_t *s, size_t *d, size_t *t);
```
timmoon10: - We should have pure C APIs. C++ does not have a stable ABI.
- This function has a heap allocation every time it is called. It would be preferable to make `src_shape` a const-ref.

transformer_engine/common/include/transformer_engine/swizzle.h
```diff
@@ -43,7 +43,8 @@ void nvte_swizzle_scaling_factors(const NVTETensor input, NVTETensor output, cud
  *  - data is quantized along K-dimension, i.e. 1D-scaling block lies along the K-dimension.
  */
 void nvte_multi_tensor_swizzle_scaling_factors(const NVTETensor* inputs, NVTETensor* outputs,
-                                               const size_t num_tensors, cudaStream_t stream);
+                                               const size_t num_tensors, cudaStream_t stream,
+                                               bool check_scale_inv_shapes = true);
```
timmoon10: - C does not support default arguments.
- Why do we need to support `check_scale_inv_shapes=false`? If attention needs scales in a different format than GEMMs, then the tensors should know about it and be able to handle it correctly. Probably the right way would be to change the tensor class so that the `with_gemm_swizzled_scales` bool becomes a `scale_format` enum.
- Less good, we could just make a new function without the scale-inv checks.

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -283,19 +308,18 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
...
+    const NVTETensor SoftmaxOffset, NVTETensor S, NVTETensor O, NVTETensorPack *Aux_CTX_Tensors,
+    const NVTETensor cu_seqlens_q, const NVTETensor cu_seqlens_kv,
+    const NVTETensor cu_seqlens_q_padded, const NVTETensor cu_seqlens_kv_padded,
+    const NVTETensor page_table_k, const NVTETensor page_table_v, const NVTETensor rng_state,
+    size_t max_seqlen_q, size_t max_seqlen_kv, bool is_training, bool return_max_logit,
+    bool cuda_graph, float attn_scale, float dropout, NVTE_QKV_Layout qkv_layout,
+    NVTE_QKV_Format o_format, NVTE_Bias_Type bias_type, NVTE_Mask_Type attn_mask_type,
+    NVTE_Softmax_Type softmax_type, int64_t window_size_left, int64_t window_size_right,
+    bool bottom_right_diagonal, NVTETensor workspace, cudaStream_t stream,
+    NVTE_QKV_Format qkv_scale_inv_format = NVTE_QKV_Format_NOT_SET);
```
timmoon10: C does not support default arguments.

Not necessary for this PR, but we should also consider changing these functions to take in an opaque config object rather than having a million args. For example, `NVTEQuantizationConfig` allows us to easily add or deprecate options without breaking anything downstream.

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -358,19 +385,19 @@ void nvte_fused_attn_fwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
...
+    NVTETensor dQ, NVTETensor dK, NVTETensor dV, NVTETensor dBias, NVTETensor dSoftmaxOffset,
+    const NVTETensor cu_seqlens_q, const NVTETensor cu_seqlens_kv,
+    const NVTETensor cu_seqlens_q_padded, const NVTETensor cu_seqlens_kv_padded,
+    size_t max_seqlen_q, size_t max_seqlen_kv, float attn_scale, float dropout,
+    NVTE_QKV_Layout qkv_layout, NVTE_QKV_Format o_format, NVTE_QKV_Format do_format,
+    NVTE_QKV_Layout dqkv_layout, NVTE_Bias_Type bias_type, NVTE_Mask_Type attn_mask_type,
+    NVTE_Softmax_Type softmax_type, int64_t window_size_left, int64_t window_size_right,
+    bool bottom_right_diagonal, bool deterministic, bool cuda_graph, NVTETensor workspace,
+    cudaStream_t stream, NVTE_QKV_Format qkv_scale_inv_format = NVTE_QKV_Format_NOT_SET,
+    NVTE_QKV_Format do_scale_inv_format = NVTE_QKV_Format_NOT_SET);
```
timmoon10: Default args are not supported in C.

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -584,6 +611,53 @@ void nvte_prepare_flash_attn_fwd(NVTETensor qkvi, NVTETensor qkv, cudaStream_t s
...
+                                                     size_t num_tensors,
+                                                     NVTE_QKV_Format original_format,
+                                                     cudaStream_t stream);
+
+/*!  \brief Pad the last dimension of multiple 2D tensors with zeros in one kernel launch.
+ *
+ *  Each tensor copies a row-major (rows, in_cols) input to a (rows, out_cols) output,
+ *  zero-filling the region [in_cols, out_cols) in every row.
+ *  Outputs must be pre-allocated with out_cols >= in_cols and matching dtype.
+ *  Up to 16 tensors may be processed in a single call.
```
timmoon10: If there are >16 tensors, couldn't we call the kernel multiple times? Then the caller doesn't need to care about this limit at all.

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -584,6 +611,53 @@ void nvte_prepare_flash_attn_fwd(NVTETensor qkvi, NVTETensor qkv, cudaStream_t s
...
+ *  Each input tensor is 4D in BSHD or SBHD layout, and the corresponding output tensor
+ *  is 4D in BHSD layout. Output tensors may be pre-allocated and have a larger D dimension.
+ *
+ *  \param[in]     inputs           List of input tensors.
+ *  \param[out]    outputs          List of output tensors.
+ *  \param[in]     num_tensors      Number of tensors in the list.
+ *  \param[in]     original_format  Original QKV format (NVTE_BSHD or NVTE_SBHD).
+ *  \param[in]     stream           CUDA stream.
+ */
+void nvte_multi_tensor_permute_to_grouped_tensor_fwd(NVTETensor *inputs, NVTETensor *outputs,
```
timmoon10: When I hear "permute forward", I think of the MoE dispatcher. Also, this isn't creating a grouped tensor, just multiple normal tensors. A less misleading name would be something like:
```suggestion
void nvte_multi_tensor_convert_to_bhsd_fwd(NVTETensor *inputs, NVTETensor *outputs,
```

transformer_engine/pytorch/csrc/common.h
```diff
@@ -301,6 +301,14 @@ class MXFP8Quantizer : public Quantizer {
   std::pair<TensorWrapper, py::object> create_tensor(const std::vector<size_t>& shape,
                                                      DType dtype) const override;
 
+  /*! @brief Construct an unquantized tensor that shares the quantizer's amax pointer.
+   *
+   * The amax is zeroed out. Most TE kernels that output amax expect
+   * amax to be initialized to zero.
+  */
+  std::pair<TensorWrapper, py::object> create_unquantized_tensor_with_amax(
+      const std::vector<size_t>& shape, DType dtype, std::optional<at::Tensor> data = std::nullopt);
```
timmoon10: I'm confused. MXFP8 quantizers don't have an amax.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -399,6 +404,21 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
...
+        "Permute multiple tensors from BSHD/SBHD to BHSD.", py::arg("inputs"),
+        py::arg("original_format") = std::string("bshd"),
+        py::arg("outputs") = std::vector<std::optional<at::Tensor>>{},
+        py::call_guard<py::gil_scoped_release>());
+  m.def("multi_tensor_permute_to_grouped_tensor_bwd",
+        &transformer_engine::pytorch::multi_tensor_permute_to_grouped_tensor_bwd,
+        "Permute multiple tensors from BHSD back to BSHD/SBHD.", py::arg("inputs"),
+        py::arg("original_format") = std::string("bshd"),
+        py::arg("outputs") = std::vector<std::optional<at::Tensor>>{},
+        py::call_guard<py::gil_scoped_release>());
```
timmoon10: These names are confusing since most people would associate "permute" with the MoE dispatcher and because these don't involve grouped tensors. Something like `multi_tensor_convert_to_bhsd_fwd` would be better.

tests/pytorch/attention/run_attention_with_cp.py
```diff
@@ -188,11 +195,15 @@ def run_dpa_with_cp(
...
+    if deterministic == "True":
+        os.environ["NVTE_ALLOW_NONDETERMINISTIC_ALGO"] = "0"
+    else:
+        os.environ["NVTE_ALLOW_NONDETERMINISTIC_ALGO"] = "1"
     fp8_bwd = fp8_bwd == "True" and dtype == "fp8"
     os.environ["NVTE_FP8_DPA_BWD"] = "1" if fp8_bwd else "0"
     fp8_dpa = fp8_dpa == "True" and dtype == "fp8"
-    fp8_mha = fp8_mha == "True" and dtype == "fp8"
-    f16_O = dtype == "fp8" and scaling_mode == "current" and f16_O == "True"
+    fp8_mha = fp8_mha == "True" and dtype == "fp8" and scaling_mode != "mxfp8"
```
ptrendx, line 205: Since this is a test, shouldn't we just skip the test that has scaling mode set to mxfp8 and fp8_mha? Otherwise we would just run the same test (without fp8_mha) again. Also we should make sure that in the actual attention code we disable fp8_mha when mxfp8 recipe is chosen. We should also log that choice in the debug log.

tests/pytorch/attention/test_attention.py
```diff
@@ -1803,20 +1803,45 @@ def get_model(dtype, config):
...
+        attn_mask_type="causal",
+    ),
+    "fp8_11": ModelConfig(
+        2,
+        4096,
+        128,
+        192,
+        head_dim_v=128,
+        attn_mask_type="causal_bottom_right",
+    ),
```
ptrendx, line 1831: Is it precommit doing something strange? The previous form was way easier to read.

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -39,13 +47,11 @@
...
-    "cp_2_0": ModelConfig(2, 4096, 12, 128, num_gqa_groups=2, attn_mask_type="causal"),  # GQA
+    "cp_2_0": ModelConfig(2, 4096, 32, 128, num_gqa_groups=4, attn_mask_type="causal"),  # GQA
     "cp_2_1": ModelConfig(2, 4096, 12, 128, num_gqa_groups=2),  # GQA
-    "cp_2_2": ModelConfig(
-        2, 4096, 12, 128, num_gqa_groups=2, attn_mask_type="causal", window_size=(512, 0)
-    ),  # GQA
+    "cp_2_2": ModelConfig(2, 4096, 32, 128, attn_mask_type="causal", window_size=(128, 0)),  # GQA
     "cp_2_3": ModelConfig(2, 4096, 12, 128, num_gqa_groups=2, window_size=(512, 512)),  # GQA
-    "cp_3_0": ModelConfig(2, 4096, 12, 192, attn_mask_type="causal", head_dim_v=128),  # MLA
+    "cp_3_0": ModelConfig(2, 4096, 128, 192, attn_mask_type="causal", head_dim_v=128),  # MLA
```
ptrendx, line 54: That makes this config much bigger, right? How long is this test?

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -73,10 +79,10 @@ def get_bash_arguments(num_gpus_per_node, **kwargs):
 qkv_formats = ["bshd", "sbhd", "thd"]
 cp_comm_types = ["p2p", "all_gather", "a2a", "a2a+p2p"]
 if test_essential:
-    configs = ["cp_1_0", "cp_1_2", "cp_2_1", "cp_3_2", "cp_3_3"]
+    configs = ["cp_2_0", "cp_2_2", "cp_3_0"]  # ["cp_1_0", "cp_1_2", "cp_2_1", "cp_3_2", "cp_3_3"]
```
ptrendx: Does the comment mean that you wanted to revert that change?

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -73,10 +79,10 @@ def get_bash_arguments(num_gpus_per_node, **kwargs):
...
     model_configs_flash_attn = {k: model_configs_flash_attn[k] for k in configs}
     dtypes = ["bf16"]
-    qkv_formats = ["sbhd", "thd"]
+    # qkv_formats = ["sbhd", "thd"]
```
ptrendx: Should this be uncommented?

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -214,21 +246,24 @@ def test_cp_with_flash_attention(dtype, model, qkv_format, cp_comm_type):
...
-        "cp_3_4",
+        "cp_2_1",
+        # "cp_2_2",
+        # "cp_2_3",
+        # "cp_2_4",
+        "cp_3_1",
+        # "cp_3_2",
+        # "cp_3_4",
         "cp_4_2",
+        "cp_4_3",
```
ptrendx: Those should either be uncommented or just removed from here (I assume that we are testing the "nonessential" cases too, maybe less frequently).

transformer_engine/common/common.h
```diff
@@ -1003,7 +1003,7 @@ size_t typeToSize(const DType type);
 size_t typeToNumBits(const DType type);
 
 void CheckNoopTensor(const Tensor &t, const std::string &name);
-void CheckInputTensor(const Tensor &t, const std::string &name);
+void CheckInputTensor(const Tensor &t, const std::string &name, bool check_scale_inv_shapes = true);
```
timmoon10, line 1006: I'm wary of this option. Why do we want to allow a tensor to have invalid scales? If attention requires scales in a different format, then the tensor should know about it and handle it within the tensor checking function.

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -188,6 +196,22 @@ NVTE_QKV_Format nvte_get_q_format(NVTE_QKV_Layout qkv_layout);
...
+ *  \param[in]     src_shape           The source shape.
+ *  \param[in]     dst_format          The destination format.
+ *  \param[out]    dst_shape           The destination shape.
+ *  \param[out]    b                   The batch size.
+ *  \param[out]    h                   The number of heads.
+ *  \param[out]    s                   The sequence length.
+ *  \param[out]    d                   The head dimension.
+ *  \param[out]    t                   The number of tokens.
+ */
+void nvte_convert_qkv_format(NVTE_QKV_Format src_format, const size_t *src_shape,
```
timmoon10: It might be helpful to make clear that this is converting tensor shapes and is not actually reordering data:
```suggestion
void nvte_convert_qkv_shape(NVTE_QKV_Format src_format, const size_t *src_shape,
```

transformer_engine/pytorch/csrc/extensions/attention.cpp
```diff
@@ -91,14 +91,29 @@ std::pair<TensorWrapper, py::object> quantizer_helper(py::handle quantizer,
...
+  } else if (detail::IsMXFP8Quantizers(quantizer.ptr())) {
+    // MXFP8
+    auto *T_quantizer_fp8 = dynamic_cast<MXFP8Quantizer *>(T_quantizer.get());
+    if (create_hp_tensor_for_cs) {
+      if (data.has_value()) {
+        std::tie(te_T, py_T) =
+            T_quantizer_fp8->create_unquantized_tensor(shape, dtype, data.value());
+      } else {
+        std::tie(te_T, py_T) = T_quantizer_fp8->create_unquantized_tensor(shape, dtype);
+      }
```
timmoon10, line 101: If this the only place we call `create_unquantized_tensor`, then we could just do:
```suggestion
      std::tie(te_T, py_T) = NoneQuantizer(py::none()).create_tensor(shape, dtype);
```
This is more logical too. If you want a high-precision tensor, ask `NoneQuantizer` to create a tensor. If want a quantized tensor, ask `T_quantizer_fp8` to create a tensor. Each quantizer does the thing it's good at.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1445,6 +1445,15 @@ std::pair<TensorWrapper, py::object> MXFP8Quantizer::create_tensor(const std::ve
   return {std::move(out_cpp), std::move(out_py)};
 }
 
+std::pair<TensorWrapper, py::object> MXFP8Quantizer::create_unquantized_tensor(
```
timmoon10: This is a strange place for this function since it doesn't have anything to do with MXFP8 and doesn't touch the quantizer state at all. I think we could remove it entirely by making the change at https://github.com/NVIDIA/TransformerEngine/pull/2719/changes#r3083600727.

transformer_engine/pytorch/csrc/extensions/attention.cpp
```diff
@@ -614,6 +653,144 @@ at::Tensor fa_prepare_bwd(at::Tensor q, at::Tensor k, at::Tensor v) {
...
+
+  size_t ki = 0;
+  for (size_t i = 0; i < inputs.size(); ++i) {
+    const int64_t in_cols = inputs[i].size(1);
+    const int64_t padded_cols =
+        static_cast<int64_t>(DIVUP_TO_MULTIPLE(static_cast<size_t>(in_cols), align));
+    if (in_cols == padded_cols) continue;
+
+    te_in_wrappers.push_back(makeTransformerEngineTensor(inputs[i]));
+    te_out_wrappers.push_back(makeTransformerEngineTensor(outputs[kernel_indices[ki]]));
```
timmoon10: Don't `inputs` and `outputs` have matching entries? If padding isn't necessary, the same tensor is put in corresponding spots. This means that this code will result in mismatched input and output tensors.
```suggestion
    te_in_wrappers.push_back(makeTransformerEngineTensor(inputs[kernel_indices[ki]]));
    te_out_wrappers.push_back(makeTransformerEngineTensor(outputs[kernel_indices[ki]]));
```

tests/pytorch/attention/test_attention_with_cp.py
cyanguwa: Yes, and done.

transformer_engine/common/common.h
cyanguwa: For "mxfp8_quantize_fast_path()", we'd have data and scale_invs in different shapes. But I've removed this flag, and created a different path `xxx_unchecked` to handle this.

transformer_engine/common/include/transformer_engine/fused_attn.h
cyanguwa: Done.

transformer_engine/pytorch/csrc/extensions/attention.cpp
cyanguwa: Reimplemented this bit of code. Initially we thought we could use the amax for something, but it's moot now.

transformer_engine/pytorch/csrc/extensions/attention.cpp
cyanguwa: Done.

transformer_engine/pytorch/csrc/quantizer.cpp
cyanguwa: Removed. The `quantizer_helper()` can handle this by creating a tensor from a None quantizer.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -1873,15 +2000,15 @@ def forward(
                         softmax_lse_in_packed_format,
                     )
             else:
-                if qkv_format in ["bshd", "sbhd"]:
+                if o_format in ["bshd", "sbhd"]:
                     flash_attn_fwd_second_half_out_correction(
                         out,
                         out_per_step[i],
                         softmax_lse,
```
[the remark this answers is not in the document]
cyanguwa: What you're looking at is V's quantization. dV is produced in FP16/BF16, and its quantization is done in the next gemm after attention, so what's specified in combine_and_quantize is irrelevant.

transformer_engine/common/swizzle/swizzle.cu
```diff
@@ -21,6 +21,17 @@ namespace {
...
 constexpr int NVFP4_BLOCK_SIZE = 16;
 
+int get_max_dynamic_smem() {
+  static int max_smem = -1;
+  if (max_smem < 0) {
+    int device;
+    NVTE_CHECK_CUDA(cudaGetDevice(&device));
+    NVTE_CHECK_CUDA(
+        cudaDeviceGetAttribute(&max_smem, cudaDevAttrMaxSharedMemoryPerBlockOptin, device));
+  }
```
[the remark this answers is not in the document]
cyanguwa: It's community standard these days that training is done on homogeneous clusters, if not the entire cluster, at least the node. PyTorch training is usually one process per GPU, while Jax training is one process per node. The implementation here should be safe enough to cover both cases.

transformer_engine/common/include/transformer_engine/swizzle.h
```diff
@@ -45,6 +45,17 @@ void nvte_swizzle_scaling_factors(const NVTETensor input, NVTETensor output, cud
 void nvte_multi_tensor_swizzle_scaling_factors(const NVTETensor* inputs, NVTETensor* outputs,
                                                const size_t num_tensors, cudaStream_t stream);
 
+/*! \brief Same as nvte_multi_tensor_swizzle_scaling_factors, but skips
+ *         scale_inv shape/padding validation.
+ *
+ *  Use this variant when the data and scale_inv tensors intentionally have
+ *  different shapes, e.g. when scale_invs have been transposed for attention.
```
timmoon10, line 52: Having a separate function for dangerous use-cases is better, although I'm still wary since we're intending to put the tensor in an invalid state. It would be better if the tensor could be configured to know what format the scales should have, similar to how they know whether the scales are swizzled or not swizzled. The right design would be to change from a bool called `with_gemm_swizzled_scales` to an enum called `scale_format`. That way GEMM could error out if it gets transposed scales and attention could error out if gets non-transposed scales, rather than silently producing incorrect values. This change may be beyond the scope of this PR though.

timmoon10 (dismissed):
I haven't finished going through all the changes, especially within the attention functions, but I don't see any problems with the changes that touch non-attention code. I'll continue going through the changes, but at this point I won't block this PR (pending CI, etc).

tests/pytorch/test_fused_rope.py
```diff
@@ -495,3 +747,385 @@ def test_rotary_position_embedding_forward_with_autocast_gives_same_result_as_wi
...
+    torch.testing.assert_close(grad_fused, grad_ref)
+    assert output_fused.is_contiguous()
+
+    # --- Megatron-LM Triton reference (forward only, uses 2D tables) ---
+    if HAVE_TRITON and tensor_format == "sbhd":
+        output_triton = _triton_mla_rope_q(t.float(), cos_2d, sin_2d, qk_head_dim, emb_dim).to(
+            dtype
+        )
+        torch.testing.assert_close(output_fused, output_triton)
+        torch.testing.assert_close(output_ref, output_triton)
```
ptrendx: Why do we need to run this kernel too? Is the PyTorch reference not enough?

tests/pytorch/test_fused_rope.py
```diff
@@ -495,3 +747,385 @@ def test_rotary_position_embedding_forward_with_autocast_gives_same_result_as_wi
...
+    if HAVE_TRITON and tensor_format == "sbhd":
+        okey_triton, oval_triton = _triton_mla_rope_kv(
+            kv.float(), k_pos_emb.float(), cos_2d, sin_2d, k_dim, v_dim, emb_dim
+        )
+        okey_triton = okey_triton.to(dtype)
+        oval_triton = oval_triton.to(dtype)
+        torch.testing.assert_close(okey_fused, okey_triton)
+        torch.testing.assert_close(oval_fused, oval_triton)
+        torch.testing.assert_close(okey_ref, okey_triton)
+        torch.testing.assert_close(oval_ref, oval_triton)
```
ptrendx: Same question

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -212,12 +218,54 @@ NVTE_QKV_Format nvte_get_kv_format(NVTE_QKV_Layout qkv_layout) {
...
+
+  auto set = [](size_t *ptr, size_t val) {
+    if (ptr) *ptr = val;
+  };
+  set(b, canonical[0]);
+  set(h, canonical[1]);
+  set(s, canonical[2]);
+  set(d, canonical[3]);
+  set(t, canonical[4]);
+}
```
ptrendx: This looks to me that we should have an AttentionShape class that be able to take the format and the shape and provide you with the b, h, s, d and t values or a shape in any other format.

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -776,14 +849,29 @@ void nvte_fused_attn_bwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
 #endif
   } else if (fused_attention_backend == NVTE_Fused_Attn_Backend::NVTE_FP8) {
 #if (CUDNN_VERSION >= 8900)
```
ptrendx: Random thing: we don't support cuDNN 8.9 anymore.

transformer_engine/common/fused_attn/utils.h
```diff
@@ -27,11 +27,198 @@ enum NVTE_QKV_Matrix {
...
+  int64_t d_v_scale;
+  int64_t d_qk_scale_padded;
+  int64_t d_v_scale_padded;
+};
+
+// Pad s and d for MXFP8 quantization
+inline MXFP8PaddedSizes pad_s_d_for_mxfp8(int64_t s_q, int64_t s_kv, int64_t d_qk, int64_t d_v) {
+  constexpr int64_t block_size = 32;
+  MXFP8PaddedSizes p;
+  p.s_q_padded = ((s_q + 127) / 128) * 128;
```
ptrendx: We have the divup_to_multiple function that does this.

transformer_engine/common/fused_attn/utils.h
```diff
@@ -27,11 +27,198 @@ enum NVTE_QKV_Matrix {
...
+  p.s_kv_padded = ((s_kv + 127) / 128) * 128;
+  p.s_q_scale = (s_q + block_size - 1) / block_size;
```
ptrendx: We have the divup function that does this.

transformer_engine/common/swizzle/swizzle.cu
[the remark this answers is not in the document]
ptrendx: I have a different issue here - getting the device properties is expesnsive. Could we move this function to the utils/cuda_runtime.cpp and utilize similar design as the rest of the functions there (with caching).

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -2180,18 +2300,289 @@ def print_quantizers(
...
+            shape = _bhsd_shape(s, d_pad)
+            numel = 1
+            for dim in shape:
+                numel *= dim
+            entries.append((total, numel, shape))
+            total += numel
+        if total == 0:
+            return [None] * len(scale_list)
+        device = next(s for s in scale_list if s is not None).device
+        buf = torch.empty(total, dtype=torch.uint8, device=device)
```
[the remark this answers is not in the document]
cyanguwa: All three paths of multi_tensor_transpose_to_bhsd (TMA, fallback_non_vec_aligned, fallback_vec_aligned) zero out the elements between D and D_out, so torch.empty here is safe.

transformer_engine/common/util/ptx.cuh
```diff
@@ -34,11 +34,7 @@ struct ArchSpecific {
   template <int CurrentArch, int ArchSpecific, int FamilySpecific>
   constexpr static bool compatible() {
     if constexpr (CurrentArch == id) {
-      static_assert(ArchSpecific == CurrentArch,
-                    "Compiled for the generic architecture, while utilizing arch-specific "
-                    "features. Please compile for smXXXa architecture instead of smXXX "
-                    "architecture.");
-      return true;
+      return ArchSpecific == CurrentArch;
```
ptrendx: This shouldn't be changed in this PR.

transformer_engine/common/util/ptx.cuh
```diff
@@ -52,11 +48,7 @@ struct FamilySpecific {
   template <int CurrentArch, int ArchSpecific, int FamilySpecific>
   constexpr static bool compatible() {
     if constexpr ((CurrentArch / 100) == (id / 100)) {
-      static_assert(FamilySpecific == CurrentArch,
-                    "Compiled for the generic architecture, while utilizing family-specific "
-                    "features. Please compile for smXXXf architecture instead of smXXX "
-                    "architecture.");
-      return true;
+      return FamilySpecific == CurrentArch;
```
ptrendx: Same here.

transformer_engine/pytorch/csrc/extensions/attention.cpp
```diff
@@ -614,6 +659,136 @@ at::Tensor fa_prepare_bwd(at::Tensor q, at::Tensor k, at::Tensor v) {
...
+
+  for (size_t i = 0; i < inputs.size(); ++i) {
+    auto &input = inputs[i];
+
+    NVTE_CHECK(input.dim() == 2, "multi_tensor_pad_last_dim: expected 2D input at index ", i,
+               ", got ", input.dim(), "D.");
+    NVTE_CHECK(input.is_cuda(), "multi_tensor_pad_last_dim: input must be a CUDA tensor at index ",
+               i, ".");
+    NVTE_CHECK(input.is_contiguous(),
+               "multi_tensor_pad_last_dim: input must be contiguous at index ", i, ".");
```
ptrendx: Shouldn't you just call contiguous on input rather than erroring out?

transformer_engine/common/fused_attn/fused_attn.cpp
cyanguwa: Removed.

transformer_engine/common/fused_attn/utils.h
cyanguwa: Replaced.

transformer_engine/common/fused_attn/utils.h
cyanguwa: Replaced.

transformer_engine/common/util/ptx.cuh
cyanguwa: Reverted, and moved flash_attn.cu to special_arch compile list in CMakeFile.

tests/pytorch/test_fused_rope.py
cyanguwa: Removed MLA RoPE.

tests/pytorch/test_fused_rope.py
cyanguwa: Removed MLA RoPE.

transformer_engine/pytorch/csrc/extensions/attention.cpp
cyanguwa: Done, together with another is_contiguous() check.

transformer_engine/common/fused_attn/fused_attn.cpp
cyanguwa: Replaced with AttentionShape.

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -712,22 +728,20 @@ void nvte_fused_attn_bwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
...
   NVTE_QKV_Format q_format = nvte_get_q_format(qkv_layout);
   NVTE_QKV_Format kv_format = nvte_get_kv_format(qkv_layout);
+  auto *q_dims = input_Q->data.shape.data();
+  auto *k_dims = input_K->data.shape.data();
+  auto *v_dims = input_V->data.shape.data();
+  AttentionShape q_shape(q_format, q_dims);
+  AttentionShape k_shape(kv_format, k_dims);
+  AttentionShape v_shape(kv_format, v_dims);
+  size_t b = q_shape.b(), h_q = q_shape.h(), d_qk = q_shape.d(), t_q = q_shape.t();
+  size_t h_kv = k_shape.h(), t_kv = k_shape.t(), d_v = v_shape.d();
```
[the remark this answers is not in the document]
cyanguwa: In bwd, we always use rowwise for V.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1599,27 +1686,31 @@ def backward(ctx, d_out, *_args):
...
+                    #                               fp8_dtype = tex.DType.kFloat8E4M3
+                    #   dq_, dk_, dv_:              torch.Tensor; dtype = torch.float16 or torch.bfloat16
+                    # MXFP8BlockScaling:
+                    #   out_, dq_, dk_, dv_, d_out: torch.Tensor; dtype = torch.float16 or torch.bfloat16
+                    out_ = out_fp8
+                    if ctx.fp8_recipe.float8_current_scaling() and _dpa_fp8_cs_o_in_f16:
+                        out_ = out
+                    if ctx.fp8_recipe.mxfp8():
+                        out_ = out
+                        aux_ctx_tensors.append(d_out)
```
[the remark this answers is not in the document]
cyanguwa: We disabled the use case of `fp8_mha=True` so `d_out` shouldn't be in MXFP8.

ptrendx (approved):

PR #2731, opened 2026-03-04
[PyTorch] Support single parameter for `GroupedLinear`

ksivaman:
# Description

Support option to expose single parameter for `GroupedLinear` module.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Support option to expose single parameter for `GroupedLinear` module.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/grouped_linear.py
+++ b/transformer_engine/pytorch/module/grouped_linear.py
@@ -788,9 +788,8 @@ class GroupedLinear(TransformerEngineBaseModule):
 
         # Re-register as a single grouped weight parameter.
         # Re-register as a single grouped weight parameter.
-        assert (
-            isinstance(grouped_weights, torch.Tensor)
-            and (weight_quantizers[0] is None or not weight_quantizers[0].internal)
+        assert isinstance(grouped_weights, torch.Tensor) and (
+            weight_quantizers[0] is None or not weight_quantizers[0].internal
         ), "Found internal quantizer with `single_grouped_parameter=True`."
         self.register_parameter(
             "weight",


transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -172,18 +177,18 @@ std::pair<GroupedTensorWrapper, py::object> NoneQuantizer::create_grouped_tensor
                                getTensorShape(*tensor_offsets));
   }
 
-  py::handle GroupedTensorClass(reinterpret_cast<PyObject*>(GroupedTensorStoragePythonClass));
+  py::handle GroupedTensorClass = grouped_tensor_python_class(this->internal);
   py::object out_py = GroupedTensorClass(
-      "num_tensors"_a = num_tensors, "quantizer"_a = std::move(quantizer),
-      "dtype"_a = GetATenDType(dtype), "data"_a = maybe_tensor_to_py(rowwise_data),
```
vthumbe1503, line 178: There is a faster way to construct python class which is used in other classes in this file. We should use that

transformer_engine/pytorch/tensor/grouped_tensor.py
```diff
@@ -0,0 +1,205 @@
...
+            ]
+            return torch.stack(dequantized_members, dim=0)
+
+        def maybe_unwrap(arg):
+            if isinstance(arg, GroupedTensor):
+                return grouped_to_stacked_tensor(arg)
+            return arg
+
+        def update_grouped_tensor_inplace(grouped: GroupedTensor, updated: torch.Tensor):
+            if not grouped.all_same_shape():
```
vthumbe1503, line 154: Since this is also repeated in grouped_to_stacked_tensor. might make sense to move this to the start of function and do it once?

ksivaman: They are 2 separate checks but the execution only happens once, either for the inplace case, or the normal case.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -35,8 +35,9 @@ PyTypeObject *Float8BlockwiseQuantizerClass = nullptr;
 PyTypeObject *NVFP4TensorPythonClass = nullptr;
 PyTypeObject *NVFP4TensorStoragePythonClass = nullptr;
 PyTypeObject *NVFP4QuantizerClass = nullptr;
-std::once_flag extension_init_flag;
```
vthumbe1503, line 38: This flag is used below to initialize all extensions once

ksivaman: Yep, just moved 2 lines below

transformer_engine/pytorch/tensor/grouped_tensor.py
```diff
@@ -0,0 +1,205 @@
...
+        ):
+            if maybe_tensor is not None:
+                device = maybe_tensor.device
+                break
+        if device is None:
+            device = torch.device("cuda")
+
+        strides = [1] * len(wrapper_shape)
+        for i in range(len(wrapper_shape) - 2, -1, -1):
+            strides[i] = strides[i + 1] * wrapper_shape[i + 1]
```
vthumbe1503, line 104: In other classes, I have recently added stride as an argument to constructor. 

This allows to create class from C++ with lesser CPU overhead. Any python compute code in __new__ has a lot of CPU overhead.

This logic can be done only if stride is not provided (whih will be object creation in python)

transformer_engine/pytorch/csrc/quantizer.cpp
ksivaman: Right...done

transformer_engine/pytorch/tensor/grouped_tensor.py
vthumbe1503: Oh I was saying from code duplication perspective. Anyway this is minor. Will leave it upto you

transformer_engine/pytorch/tensor/grouped_tensor.py
```diff
@@ -0,0 +1,205 @@
...
+        def update_grouped_tensor_inplace(grouped: GroupedTensor, updated: torch.Tensor):
+            if not grouped.all_same_shape():
+                raise NotImplementedError(
+                    "GroupedTensor __torch_dispatch__ currently supports only uniform member shapes"
+                )
+            updated_members = list(updated.unbind(dim=0))
+            if grouped.quantizer is None:
+                grouped_members = grouped.quantized_tensors
+                if grouped_members is None:
+                    grouped_members = grouped.split_into_quantized_tensors()
```
zhongbozhu, line 162: add some fall back logic here:
- if grouped quantize is available
- if it's not available, split it and trigger quantize one by one

transformer_engine/pytorch/tensor/grouped_tensor.py
```diff
@@ -0,0 +1,205 @@
...
+            raise RuntimeError(f"{cls.__name__} forbids shape-manipulation op: {func} ")
+
+        def grouped_to_stacked_tensor(grouped: GroupedTensor) -> torch.Tensor:
+            if not grouped.all_same_shape():
+                raise NotImplementedError(
+                    "GroupedTensor __torch_dispatch__ currently supports only uniform member shapes"
+                )
+            grouped_members = grouped.quantized_tensors
+            if grouped_members is None:
+                grouped_members = grouped.split_into_quantized_tensors()
```
zhongbozhu, line 137: same comment, add some fallback check

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -983,7 +1002,14 @@ def _customize_quantizers_float8_current_scaling(self, fwd: bool, recipe: Recipe
 
     def _get_weight_tensors(self) -> List[Union[torch.Tensor, QuantizedTensorStorage]]:
         """Get the weight tensors of the module."""
-        weight_tensors = [getattr(self, f"weight{i}") for i in range(self.num_gemms)]
+        grouped_weight = getattr(self, "weight", None)
```
zhongbozhu, line 1005: there is another case:
- suppose fp8 primary weight is already supported for grouped single weight, then you also shouldn't split it because this means that you can directly call the gemm 
- suppose we didn't turn on fp8 primary weight, but still turn on single weight, then we should 1) if grouped quantize kernel for weight is spported, call it 2) if it's not, then call this API to split and quantize it, so it's actually better to call a split_quantize instead which is more performant. 

This basically means that when this _get_weight_tensors is getting called, it's not going to try to split it beforehand, but we should instead try to call grouped_quantize or split_quantize here: https://github.com/NVIDIA/TransformerEngine/blob/a3bc040682672ccfab8f79f63d099b06d9c3135a/transformer_engine/pytorch/module/base.py#L1345

PR #2738, opened 2026-03-05
[Common] Persistent Grouped MXFP8 quantization kernel

Oleg-Goncharov:
# Description
This PR adds a persistent grouped MXFP8 quantization kernel with static scheduling. 

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Added persistent kernel 
- Added TunableConfig structure to tune performance

# Checklist:

### DIFF
--- a/tests/jax/test_custom_call_compute.py
+++ b/tests/jax/test_custom_call_compute.py
@@ -36,6 +36,7 @@ from transformer_engine.jax.quantize import (
     ScaledTensor1x,
     ScaledTensor2x,
     GroupedScaledTensor1x,
+    GroupedNoScaleTensor,
     ScalingMode,
     QuantizerFactory,
     QuantizeLayout,
@@ -150,8 +151,13 @@ def assert_dequantized_grouped_scaled_tensor(
     a: Union[GroupedScaledTensor1x, ScaledTensor2x], b: jnp.ndarray
 ):
     if isinstance(a, GroupedScaledTensor1x):
-        assert a.group_sizes.sum() == b.shape[0]
-        b = jnp.split(b, jnp.cumulative_sum(a.group_sizes)[:-1], axis=0)
+        group_sizes = (
+            a.first_dims
+            if a.first_dims is not None
+            else jnp.ones(a.original_shape[0], dtype=jnp.int32)
+        )
+        assert group_sizes.sum() == b.shape[0]
+        b = jnp.split(b, jnp.cumulative_sum(group_sizes)[:-1], axis=0)
         dq_a = a.dequantize()
         for dq_a_i, b_i in zip(dq_a, b):
             if len(dq_a_i) == 0:
@@ -1787,13 +1793,18 @@ class TestGroupedDense:
         ref_out = self._ref_grouped_dense(lhs, rhs, None, group_sizes, contracting_dims)
 
         # jitting grouped_gemm
+        lhs_tensor = GroupedNoScaleTensor(
+            data=lhs, amax=None, first_dims=group_sizes, last_dims=None, original_shape=lhs.shape
+        )
+        rhs_tensor = GroupedNoScaleTensor(
+            data=rhs, amax=None, first_dims=None, last_dims=None, original_shape=rhs.shape
+        )
         prim_out = jax.jit(
             tex.grouped_gemm, static_argnames=("contracting_dims", "use_async_d2h_group_sizes")
         )(
-            lhs,
-            rhs,
-            group_sizes,
-            contracting_dims,
+            lhs_tensor,
+            rhs_tensor,
+            contracting_dims=contracting_dims,
             use_async_d2h_group_sizes=True,
         )
 
@@ -1825,8 +1836,17 @@ class TestGroupedDense:
         )
         ref_out = self._ref_grouped_dense(lhs, rhs, None, group_sizes, contracting_dims)
 
+        lhs_tensor = GroupedNoScaleTensor(
+            data=lhs, amax=None, first_dims=group_sizes, last_dims=None, original_shape=lhs.shape
+        )
+        rhs_tensor = GroupedNoScaleTensor(
+            data=rhs, amax=None, first_dims=None, last_dims=None, original_shape=rhs.shape
+        )
         prim_out = jax.jit(tex.grouped_gemm, static_argnames=("contracting_dims",))(
-            lhs, rhs, group_sizes, contracting_dims, quantizer_set=quantizer_set
+            lhs_tensor,
+            rhs_tensor,
+            contracting_dims=contracting_dims,
+            quantizer_set=quantizer_set,
         )
 
         allclose_dtype = jnp.float8_e4m3fn
--- a/transformer_engine/jax/cpp_extensions/gemm.py
+++ b/transformer_engine/jax/cpp_extensions/gemm.py
@@ -37,6 +37,7 @@ from ..quantize import (
     ScaledTensor1x,
     ScaledTensor2x,
     GroupedScaledTensor1x,
+    GroupedNoScaleTensor,
     ScalingMode,
     Quantizer,
     GroupedQuantizer,
@@ -73,12 +74,14 @@ num_cublas_streams = get_num_compute_streams()
 # Cache whether the CUDA-graphable grouped GEMM implementation is available at import time.
 # Calling get_grouped_gemm_setup_workspace_size raises a RuntimeError mentioning "cublas" when
 # compiled against cuBLAS < 13.2, in which case the cuda-graphable path is unavailable.
+_v2_grouped_gemm_available_reason = ""
 try:
     get_grouped_gemm_setup_workspace_size(1)
     _v2_grouped_gemm_available = True
 except RuntimeError as e:
     if "cublas" in str(e).lower():
         _v2_grouped_gemm_available = False
+        _v2_grouped_gemm_available_reason = str(e)
     else:
         raise
 
@@ -1392,17 +1395,47 @@ class GroupedGemmCopySizesPrimitive(BasePrimitive):
 register_primitive(GroupedGemmCopySizesPrimitive)
 
 
+def _assert_grouped_gemm_dims_shapes(
+    lhs_first_dims_aval,
+    lhs_last_dims_aval,
+    rhs_first_dims_aval,
+    rhs_last_dims_aval,
+    out_first_dims_aval,
+    out_last_dims_aval,
+    num_groups: int,
+) -> None:
+    """Assert that all non-empty *_dims arrays have exactly num_groups elements.
+
+    rhs_first_dims / rhs_last_dims describe the ragged contracting K dimension.
+    K totals need not fill the entire buffer (padding is allowed), so only the
+    array length is checked, not the per-group sum.
+    """
+    for name, aval in [
+        ("lhs_first_dims", lhs_first_dims_aval),
+        ("lhs_last_dims", lhs_last_dims_aval),
+        ("out_first_dims", out_first_dims_aval),
+        ("out_last_dims", out_last_dims_aval),
+        ("rhs_first_dims", rhs_first_dims_aval),
+        ("rhs_last_dims", rhs_last_dims_aval),
+    ]:
+        if aval.size > 0:
+            assert (
+                aval.size == num_groups
+            ), f"grouped GEMM {name} has size {aval.size}, expected num_groups={num_groups}"
+
+
 class GroupedGemmPrimitive(BasePrimitive):
     """
     Primitive for grouped GEMM using nvte_multi_tensor_gemm (supports all scaling modes) or nvte_grouped_gemm (supporting BF16).
     """
 
-    # args = lhs_data, lhs_scale_inv, rhs_data, rhs_scale_inv, bias, group_sizes, group_offset, unused_placeholder
     name = "te_grouped_gemm_ffi"
-    # args = lhs_data, lhs_scale_inv, rhs_data, rhs_scale_inv, bias, group_sizes, alpha, beta
+    # args = lhs_data, lhs_scale_inv, rhs_data, rhs_scale_inv, bias,
+    #        lhs_first_dims, lhs_last_dims, rhs_first_dims, rhs_last_dims,
+    #        out_first_dims, out_last_dims, alpha, beta
     name_graph_safe = "te_grouped_gemm_v2_ffi"
     multiple_results = True
-    impl_static_args = (8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18)
+    impl_static_args = (13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24, 25, 26)
     inner_primitive = None
     outer_primitive = None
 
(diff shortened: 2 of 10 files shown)


transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -124,11 +142,95 @@ __device__ __forceinline__ size_t get_tensor_cols_num(
...
+  job.block_global_offset = is_single_tensor
+                                ? (ctaid_Y * CHUNK_DIM_Y * last_logical_dim + ctaid_X * CHUNK_DIM_X)
+                                : (job.block_id * ELTS_PER_CHUNK);
+  job.tensor_id = get_current_tensor_id(shape_rep, num_tensors, job.block_global_offset, ctaid_Y,
+                                        first_logical_dim, last_logical_dim, offsets_ptr);
+  job.rows =
+      get_tensor_rows_num(job.tensor_id, shape_rep, first_logical_dim, first_dims_ptr, num_tensors);
+  job.cols = get_tensor_cols_num(job.tensor_id, shape_rep, last_logical_dim, last_dims_ptr);
+  return job;
+}
```
ptrendx: This should be a constructor of the JobDescriptor struct (you can make the constructor `__device__` too).

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -124,11 +142,95 @@ __device__ __forceinline__ size_t get_tensor_cols_num(
...
+  const size_t CHUNK_DIM_X_ = CHUNK_DIM_X;
+  const size_t blocks_X_num_in_current_tensor = DIVUP(job.cols, CHUNK_DIM_X_);
+  block.block_id_in_current_tensor =
+      is_single_tensor ? job.block_id : (job.block_id - block.tensor_base / ELTS_PER_CHUNK);
+  block.block_id_Y = block.block_id_in_current_tensor / blocks_X_num_in_current_tensor;
+  block.block_id_X = block.block_id_in_current_tensor % blocks_X_num_in_current_tensor;
+  block.block_offset_Y = block.block_id_Y * CHUNK_DIM_Y;
+  block.block_offset_X = block.block_id_X * CHUNK_DIM_X;
+  return block;
+}
```
ptrendx: Similarly this should be a constructor too.

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -228,125 +330,322 @@ __device__ __forceinline__ void fence_acquire_tensormap(const CUtensorMap *tenso
...
 
+// Issue TMA global->shared transfer for one stage of input (and optional activation input).
+template <typename IType, bool IS_DACT>
+__device__ __forceinline__ void prefetch_input_stage(
+    IType *in_sh, IType *act_in_sh, const CUtensorMap &tensor_map_input,
+    const CUtensorMap &tensor_map_act_input, const size_t global_offset_X,
+    const size_t global_offset_Y, const size_t buff_offset, const size_t shmem_buff_size,
+    uint64_t *barrier, const bool leading_thread) {
+  if (leading_thread) {
+    ptx::mbarrier_arrive_expect_tx(barrier, shmem_buff_size);
```
ptrendx: 2 questions - why is this done before the TMA call and why is it done only by the leading_thread? In the other parts of the code (e.g. in `ptx::copy_2d_to_shared`) we do transfer, then arrive_expect on the leading thread and just arrive on all the other threads.

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -390,374 +684,331 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK) group_quantize_mxfp8_kernel
...
+  int IN_buff_readable_parity[BUFFS_NUM] = {0};
+  int32_t ctaid_X = static_cast<int32_t>(blockIdx.x);
+  int32_t ctaid_Y = static_cast<int32_t>(blockIdx.y);
+  size_t static_next_block_id = 0;
+  size_t static_block_stride = 0;
+  // In persistent mode, physical CTAs iterate over a virtual work grid via grid-stride.
+  if constexpr (PERSISTENT) {
+    if (launch_block_id >= total_work_blocks) {
+      return;
+    }
```
ptrendx: Is this possible?

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -390,374 +684,331 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK) group_quantize_mxfp8_kernel
...
-          Vec<IType, PACK_SIZE> in;
-          Vec<IType, PACK_SIZE> act_in;
+      // Stage-(STAGES - PREFETCH_STAGES) prefetches stage-0 of the next job.
+      // Validate that job before issuing TMA reads to avoid OOB accesses on graph-safe tails.
+      if ((stage >= STAGES - PREFETCH_STAGES) && allow_next_job_prefetch && !job_finished) {
+        prefetch_job = decode_job(shape_rep, is_single_tensor, num_tensors, first_logical_dim,
+                                  last_logical_dim, work_blocks_X, ctaid_X, ctaid_Y, offsets_ptr,
+                                  first_dims_ptr, last_dims_ptr);
+        allow_next_job_prefetch =
+            is_job_valid(prefetch_job, shape_rep, total_work_blocks, offsets_ptr);
```
ptrendx: If we are prevalidating the next job here, then why do we need earlier the check if the job we are about to do is going to be valid and draining it if it is not?

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -390,374 +684,331 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK) group_quantize_mxfp8_kernel
...
+  while (!job_finished) {
+    // Decode CTA assignment into logical tensor coordinates and validate bounds.
+    const JobDescriptor current_job =
+        decode_job(shape_rep, is_single_tensor, num_tensors, first_logical_dim, last_logical_dim,
+                   work_blocks_X, ctaid_X, ctaid_Y, offsets_ptr, first_dims_ptr, last_dims_ptr);
+    const bool current_job_is_valid =
+        is_job_valid(current_job, shape_rep, total_work_blocks, offsets_ptr);
+    if (!current_job_is_valid) {
+      if (has_prefetched_current_job) {
+        // A stage-0 prefetch may already be in flight for this CTA. Drain it before exiting.
```
ptrendx: Why do you need to drain it

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
Oleg-Goncharov: `ptx::mbarrier_arrive_expect_tx` is also called by a single thread in `ptx::copy_2d_to_shared`. I initialized the barriers using a single thread, which is sufficient for it to work. But we can also keep the previous approach, where all threads in the block participate explicitly. And since the async copy and `expect_tx` are in the same phase, it’s also valid to issue `expect_tx` first.

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
Oleg-Goncharov: For example, for small input tensors where `total_work_blocks` is less than `SMs * K`, with `K = STATIC_PERSISTENT_BLOCKS_PER_SM`

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
Oleg-Goncharov: We prefetch the first stage of the next CTA at the end of processing the current CTA. This check is only to avoid copying data for null blocks. The main termination check, i.e., when to stop processing the current chunk and exit the loop is at line 770.

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
Oleg-Goncharov: We destroy the barriers after exiting the loop. But this invalidation can be done once the mbarrier objects are guaranteed to have completed their current phase (drained). Otherwise, the TMA engine may finish the copy and attempt to call `complete` on an already invalidated mbarrier

tests/cpp/operator/test_cast_mxfp8_grouped.cu
```diff
@@ -651,8 +674,11 @@ std::vector<std::vector<size_t>> input_config = {
     {VARYING_FIRST_DIM,     3,      1024,144,                   128,384,512},
     {VARYING_FIRST_DIM,     4,      1536,160,                   128,384,512,512},
     {VARYING_FIRST_DIM,     5,      4096,512,                   128,256,384,1024,2304},
+    {VARYING_FIRST_DIM,     5,      16 * 4096,512,              128,256,384,1024,2304},
     {VARYING_LAST_DIM,      3,      256,896,                    128,256,512},
     {VARYING_BOTH_DIMS,     2,      1,(128*128)+(256*256),      128,256,        128,256},
+    // Empty tensor in the middle of the group must not terminate the persistent work loop.
+    {VARYING_BOTH_DIMS,     3,      1,(128*128)+(128*128),      128,0,128,      128,0,128},
```
jberchtold-nvidia, line 670: Can you also add a test case for only varying first dim that contains as group of size zero

tests/cpp/operator/test_cast_mxfp8_grouped.cu
```diff
@@ -129,7 +129,7 @@ void compute_ref(const ProcessingMethod processing_method,
                     const fp8e8m0 biased_exponent = float_to_e8m0(block_amax * Quantized_Limits<OutputType>::max_reciprocal());
                     const size_t scale_idx = i * scales_stride_rowwise + tile_X;
                     output_scales_rowwise[scale_idx] = biased_exponent;
-                    const float scale_reciprocal = exp2f_rcp(biased_exponent);
+                    float scale_reciprocal = exp2f_rcp(biased_exponent);
```
ptrendx: Why not const?

tests/cpp/operator/test_cast_mxfp8_grouped.cu
```diff
@@ -272,9 +272,13 @@ void performTest(const ProcessingMethod processing_method,
         const size_t elts = M * K;
         elts_num += elts;
 
+        auto divide_round_up_blocks = [](const size_t N, const size_t M) -> size_t {
+            return (N == 0) ? 0 : 1 + (N - 1) / M;
+        };
```
ptrendx: We have divup function already which I believe does exactly that.

transformer_engine/common/include/transformer_engine/cast.h
```diff
@@ -124,6 +124,19 @@ void nvte_quantize_noop(const NVTETensor input, NVTETensor output, NVTETensor no
...
+ *         The type of quantized tensor in the output depends on the scaling mode of the output
+ *         tensor. See file level comments.
+ *         For grouped tensors with a varying last dimension, the last dimension must be a multiple of 128.
+ *
+ *  \param[in]     input            Input grouped tensor to be cast.
+ *  \param[in,out] output           Output grouped MXFP8 tensor.
+ *  \param[in]     quant_config     Quantization configuration.
+ *  \param[in]     stream           CUDA stream used for the operation.
+ */
+void nvte_group_quantize_v2(const NVTEGroupedTensor input, NVTEGroupedTensor output,
```
ptrendx: We introduced group quantize API last month - I think we should just change it rather than right away do v2. It is a breaking change, but the number of people affected should be very small.

transformer_engine/common/include/transformer_engine/cast.h
```diff
@@ -124,6 +124,19 @@ void nvte_quantize_noop(const NVTETensor input, NVTETensor output, NVTETensor no
 void nvte_quantize_v2(const NVTETensor input, NVTETensor output,
                       const NVTEQuantizationConfig quant_config, cudaStream_t stream);
 
+/*! \brief Casts input grouped tensor to MXFP8.
```
ptrendx: Also, we should not say that it is a cast just to MXFP8 even if MXFP8 is the only supported mode right now.

transformer_engine/common/common.h
```diff
@@ -943,6 +985,8 @@ constexpr size_t scale_tensor_alignment_Y_rowwise = 128;
 constexpr size_t scale_tensor_alignment_X_colwise = 128;
 constexpr size_t scale_tensor_alignment_Y_colwise = 4;
 
+constexpr size_t SCALING_FACTORS_SWIZZLE_ALIGNMENT = 128;
```
ptrendx, line 988: Why do we have that if we already have the other scale alignments above?

transformer_engine/common/cast/core/common.cuh
```diff
@@ -180,6 +186,402 @@ void grouped_reduce_dbias(const ShapeRepresentation shape_rep, const size_t num_
...
+                    const int64_t *const __restrict__ last_dims_ptr) {
+  size_t cols_num = 0;
+  if constexpr (SHAPE_REP == ShapeRepresentation::SAME_BOTH_DIMS ||
+                SHAPE_REP == ShapeRepresentation::VARYING_FIRST_DIM) {
+    cols_num = last_logical_dim;
+  } else {
+    cols_num = static_cast<size_t>(last_dims_ptr[tensor_id]);
+    if (cols_num % 128 != 0) {
+      NVTE_DEVICE_ERROR(
+          "For non-single tensors, the last dimension of each tensor in a group "
```
ptrendx: It should probably say "for varying last dimensions support" rather than "For non-single tensors" as the second one is internal thing that the end user  has no knowledge about.

transformer_engine/common/cast/core/common.cuh
```diff
@@ -180,6 +186,402 @@ void grouped_reduce_dbias(const ShapeRepresentation shape_rep, const size_t num_
...
+        "tensormap.replace.tile.global_stride.b1024.b64  [tensor_map_ptr], 0, %4; \n"
+        "}\n" ::"l"(reinterpret_cast<uintptr_t>(&shared_tensor_map)),
+        "l"(global_data_ptr), "r"(static_cast<uint32_t>(global_dim_Y)),
+        "r"(static_cast<uint32_t>(global_dim_X)), "l"(static_cast<uint64_t>(global_stride_bytes))
+        : "memory");
+    *global_tensor_map = shared_tensor_map;
+  } else {
+    NVTE_DEVICE_ERROR(
+        "tensormap.replace is architecture-specific. "
+        "Try recompiling with sm_XXXa instead of sm_XXX.");
```
ptrendx: This type of error message is misleading considering our current build system where we would always build for the specific architecture anyway - we should leave only the first line of this.

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -30,331 +31,449 @@ namespace group_quantize_kernel {
...
-  shared_tensor_map = base_tensor_map;  // Copy the base tensor map into shmem
-  constexpr bool is_blackwell = ARCH_BLACKWELL_FAMILY;
-  if constexpr (is_blackwell) {
-    const size_t global_stride_bytes = global_dim_X * data_type_size_bytes;
-    if (global_stride_bytes % TMA_GMEM_ALIGNMENT != 0) {
-      NVTE_DEVICE_ERROR("Shape not supported. Data stride must be 16B aligned.");
-    }
-    if (global_data_ptr % TMA_GMEM_ALIGNMENT != 0) {
-      NVTE_DEVICE_ERROR("Tensor data pointer must be 16B aligned");
+  if constexpr (BF16_CAST_ONLY) {
```
ptrendx: I would merge this with the previous if constexpr - move the scale_idx calculation before everything and duplicate the biased_exponent calculation. That way at least you will have everything close together rather than this current complicated flow.

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -30,331 +31,449 @@ namespace group_quantize_kernel {
...
+  IType rIn[BUFF_DIM_Y];
+  IType4 rIn4x[BUFF_DIM_Y / 4];
+
+  if constexpr (BF16_CAST_ONLY) {
+    IType2 thread_amax_2x = {static_cast<IType>(0.0f), static_cast<IType>(0.0f)};
+#pragma unroll
+    for (int i = 0; i < BUFF_DIM_Y; i += 4) {
+      const uint32_t src_smem_ptr = __cvta_generic_to_shared(&sIn[buff][i][j]);
+
+      // Load 4x elts S2R and find amax
```
ptrendx, line 135: I think this would belong to ptx.cuh, no?

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -387,386 +501,259 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK) group_quantize_mxfp8_kernel
...
+  const size_t total_work_blocks = work_blocks_X * work_blocks_Y;
+  const size_t launch_block_id = blockIdx.y * gridDim.x + blockIdx.x;
+
+  int IN_buff_readable_parity[BUFFS_NUM] = {0};
+  int32_t ctaid_X = static_cast<int32_t>(blockIdx.x);
+  int32_t ctaid_Y = static_cast<int32_t>(blockIdx.y);
+  size_t static_next_block_id = 0;
+  size_t static_block_stride = 0;
+  // In persistent mode, physical CTAs iterate over a virtual work grid via grid-stride.
+  if constexpr (PERSISTENT) {
```
ptrendx: Is persistent mode faster in general than the nonpersistent mode? If so, why don't we just move to the persistent way of doing things rather than making the code even more complicated by another template argument?

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -387,386 +501,259 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK) group_quantize_mxfp8_kernel
...
+            sOutRowwise_ptr, scales_rowwise, thread_dbias_rowwise);
+      }
 
-    // Wait for shared memory writes to be visible to TMA engine.
-    ptx::fence_proxy_async_shared_cta();
-    __syncthreads();
-    // After syncthreads, writes by all threads are visible to TMA engine.
+      __builtin_assume(block_amax >= 0);
+      __builtin_assume(thread_amax >= 0);
+      block_amax = fmaxf(block_amax, thread_amax);
```
ptrendx: Considering that this is MXFP8, why do we need this at all? amax_ptr is never going to be true in this kernel.

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
Oleg-Goncharov: Since this is fairly specialized PTX and is unlikely to be reused elsewhere, maybe we should keep it here?

ptrendx:
Internal pipeline 47570778 passed.

PR #2739, opened 2026-03-05
[JAX] GEMM tex and FFI cleanup

phu0ngng:
# Description

This PR removes unused, untested, and partially supported features from the public GEMM primitive: fused GeLU, bias gradient, and grad mode — these were dead code paths not exercised by any JAX-side caller.

Besides, the PR also removes all the `FP8_2X_ACC_XGRAD` from the QuantizeConfig as it is no longer inferable from the recipes. Users can set the precision via the new env variable `TE_FP8_GEMM_HIGH_PRECISION_ACCUMULATION` instead.

Change details: 
  - Introduce`GemmV2FFI` replacing the old GemmFFI, remove untested/unused boolean flags (fuse_bias, fuse_gelu, grad), and consolidate other individual attributes into a GemmConfig struct.
  - Bias fusion is now inferred from `bias.size` > 0 rather than a separate fuse_bias flag
  - Old GemmFFI is kept as a deprecated shim (warns once via std::call_once, scheduled for removal September 2026) to avoid breaking existing users
  - `TE_FP8_GEMM_HIGH_PRECISION_ACCUMULATION` env-var is introduced to set the precision of the accumulation in MatMul, i.e., whether to promote to high dtype for storing the intermediate accumulation result. 
  - Move `assert_cublas_requirements` checks from lowering to abstract, providing earlier shape validation
  - In the `test_distributed_dense.py`, output sharding constraint is added to ensure the correct sharding pattern for the input gradients in the bprop.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:

### DIFF
--- a/docs/examples/attention/attention.ipynb
+++ b/docs/examples/attention/attention.ipynb
@@ -151,6 +151,7 @@
     "- flash-attention does not support `post_scale_bias`, and cuDNN attention does.\n",
     "- flash-attention supports KV-caching and paged attention, and cuDNN attention does not.\n",
     "- flash-attention uses bottom right diagonal for `causal` mask in cross attention (see [change log](https://github.com/Dao-AILab/flash-attention?tab=readme-ov-file#21-change-behavior-of-causal-flag)), and cuDNN attention supports both top left and bottom right.\n",
+    "- **Sliding window attention (SWA):** flash-attention has SWA(left, right) support for all mask types except top-left causal masks, with or without dropout, and without bias. cuDNN attention supports SWA(left, 0) starting from 9.2 and SWA(left, right) starting from 9.6, without dropout, and with `bias_type=\"no_bias\"`.\n",
     "- flash-attention outperforms cuDNN attention on Ampere architectures, and cuDNN attention has 20-50% advantages on Hopper architectures, based on our benchmarks for a number of commonly-used model configurations.\n",
     "\n",
     "To compare cuDNN attention and flash-attention, users can modify the `model_configs` dictionary in [benchmarks/attention/benchmark_attention.py](https://github.com/NVIDIA/TransformerEngine/blob/main/benchmarks/attention/benchmark_attention.py) to collect performance numbers. The script runs each entry in `model_configs` for `num_iters` times, each time with one forward pass and one backward pass. Both backends are tried, and if one backend does not have support for the specific user input, the runtimes and speedups in the final table would be 0."
@@ -389,7 +390,7 @@
     "\n",
     "| Attention Backend | Precision | Architecture | Sliding Window Attention | MQA/GQA | Multi-Latent Attention | Context Parallelism | Determinism Possible |\n",
     "| :---------------- | :-------- | :----------- | :----------------------- | :------ | :--------------------- | :------------------ | :------------ |\n",
-    "| cuDNN attention (all frameworks) | BF16, FP16, FP8 (PyTorch only) |  sm80+ | No  | Yes | Yes | Yes (`bshd`,`sbhd`, `thd`) | Yes |\n",
+    "| cuDNN attention (all frameworks) | BF16, FP16, FP8 (PyTorch only) |  sm80+ | Yes (cuDNN 9.2+) | Yes | Yes | Yes (`bshd`,`sbhd`, `thd`) | Yes |\n",
     "| flash-attention (PyTorch)           | BF16, FP16      |  sm80+ | Yes | Yes | Yes | Yes (`bshd`,`thd`)  | Yes                                                                                    |\n",
     "| Framework-native attention | BF16, FP16, FP32 |  Any   | No, unless used as a mask  | Yes | Yes (PyTorch only) | No                                  | Yes |\n",
     "\n",
--- a/docs/examples/attention/cp_ag_thd_dpa_jax_deep_dive.ipynb
+++ b/docs/examples/attention/cp_ag_thd_dpa_jax_deep_dive.ipynb
@@ -28,6 +28,7 @@
    "source": [
     "### Question 1: Why choose Striped>1 ?\n",
     "\n",
+    "\n",
     "Prior to the addition of this feature, Transformer Engine JAX attention already supported load balancing via a striping pattern, i.e., `stripe_size=1` for `CP + THD + P2P(Ring) + Striped + SWA`. However, this reordering technique does not lend itself well to an all-gathered (post-AG) pattern. The following example illustrates this distinction. For this example, `cp_size=4`, `num_segments=4`, `window_size=(8,0)`, and the pattern is for a single rank after striped reordering has been performed: \n",
     "\n",
     "#### I. Striped (`stripe_size=1`)\n",
--- a/transformer_engine/common/triton/permutation.py
+++ b/transformer_engine/common/triton/permutation.py
@@ -599,6 +599,10 @@ def _sort_chunks_by_map_kernel(
     input_ptr,
     row_id_map_ptr,
     probs_ptr,
+    # Pre-allocated output buffer for JAX input_output_aliases.
+    # Aliased to output_ptr in JAX so they point to the same memory.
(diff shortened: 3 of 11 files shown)


transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -457,6 +457,8 @@ def _dims_are_consecutive(dims):
             "cuBLAS GEMM operands have incompatible contracting dimensions: "
             f"{lhs.shape} @ idx {lhs_contracting_dims} X {rhs.shape} @ idx {rhs_contracting_dims}."
         )
+        assert_cublas_requirements(scaling_mode, lhs_contracting_size, "LHS")
+        assert_cublas_requirements(scaling_mode, rhs_contracting_size, "RHS")
```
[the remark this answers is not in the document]
phu0ngng: It's fine. The current message has enough info for debugging already.

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
```diff
@@ -138,8 +138,8 @@ void CommunicatorHandler::init(int num_total_devices, int num_devices_per_proces
 
   // Bootstrap UB via creating a dummy CommOverlapP2PBase object
   std::vector<size_t> buffer_shape{1, 1};
-  auto _ = CollectiveGemmPlanRegistry::getInstance().get_executor(buffer_shape, DType::kFloat32,
-                                                                  JAXX_Collective_Op::ALL_GATHER);
+  [[maybe_unused]] auto _ = CollectiveGemmPlanRegistry::getInstance().get_executor(
```
jberchtold-nvidia, line 141: This wasn't introduced in this PR, but why do we need to have an `auto _ =`? Would it not work to just call `CollectiveGemmPlanRegistry::getInstance().get_executor`?

transformer_engine/jax/csrc/extensions.h
```diff
@@ -170,6 +182,16 @@ XLA_FFI_REGISTER_STRUCT_ATTR_DECODING(
     transformer_engine::jax::ActivationConfig,
     ::xla::ffi::StructMember<transformer_engine::jax::ClampedSwigluConfig>("clamped_swiglu"));
 
+XLA_FFI_REGISTER_STRUCT_ATTR_DECODING(
+    transformer_engine::jax::GemmConfig,
+    ::xla::ffi::StructMember<transformer_engine::jax::JAXX_Scaling_Mode>("scaling_mode"),
+    ::xla::ffi::StructMember<transformer_engine::jax::JAXX_Collective_Op>("collective_op"),
+    ::xla::ffi::StructMember<int64_t>("lhs_axis_boundary"),
```
jberchtold-nvidia, line 189: Good idea, this struct-based approach lets us better set default values for newly added fields, right? So it's easier to be backwards compatible with older HLO but still have the flexibility to add new attributes as long as the default value keeps the same behavior as before. I recall doing something similar for attention

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -380,6 +378,12 @@ def get_rhs_axis_boundary(rhs_cdims, is_transposed):
     return min(rhs_cdims) if is_transposed else max(rhs_cdims) + 1
 
 
+@cache
+def _get_high_precision_accumulation_from_env() -> bool:
+    """Read TE_FP8_GEMM_HIGH_PRECISION_ACCUMULATION once per process (cached)."""
+    return os.getenv("TE_FP8_GEMM_HIGH_PRECISION_ACCUMULATION", "0") == "1"
```
jberchtold-nvidia: Is this a new env var? If so, we should prefix with `NVTE_` rather than `TE_`. Additionally, if possible would be great to keep this on the recipe state rather than global, but given this feature may be more dependent on hardware arch than recipe a global flag may be okay.

Was this difficult to pass thru this info to the GEMMs because we mostly track recipe state on the quantizers and for the GEMM we already have ScaledTensors and are missing the full recipe context?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -996,16 +908,17 @@ def _parse_operand_output_specs(
             (lhs_non_cspecs, rhs_non_cspecs),
         )
 
-        # Bias and Pre-GeLU sharding is based on GEMM output before any scatter
-        bias_specs = tuple(list(rhs_non_cspecs).copy())
-        gelu_specs = tuple(list(out_specs).copy())
+        # Bias sharding is based on GEMM output before any scatter
+        bias_specs = (
+            tuple(list(rhs_non_cspecs).copy()) if arg_infos[4].size > 0 else (None,)
```
jberchtold-nvidia: nit SR: why do we need `.copy()` here? Won't `list(rhs_non_cspecs)` already copy the list. And wouldn't wrapping that in a `tuple(` then as `tuple(list(rhs_non_cspecs).copy())` copy again. I wouldn't think we'd have to worry about multiple references mutating the same tuple since tuples are immutable. Not a functionality or perf concern since these are super tiny tuples of strings, but curious for my own Python understanding if there's an edge case here that I'm not familiar with

transformer_engine/jax/cpp_extensions/gemm.py
jberchtold-nvidia: From reading below this is because the current implementation used the global recipe state instead of the direct recipe passing usage. I'm okay with this env var and I think decoupling from the recipe makes sense since this is more of a flag for GEMMs (even unquantized GEMMs) than a quantization-specific recipe setting.

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -98,46 +98,75 @@ std::tuple<TensorWrapper, std::vector<size_t>> xla_buffer_to_nvte_gemm_operand(
...
+                                 Result_Type output, Result_Type bias_grad,
+                                 Result_Type pre_gelu_out, Result_Type workspace,
+                                 JAXX_Scaling_Mode scaling_mode, int64_t lhs_axis_boundary,
+                                 int64_t rhs_axis_boundary, bool lhs_transposed,
+                                 bool rhs_transposed, bool fuse_bias, bool fuse_gelu, bool grad,
+                                 bool use_split_accumulator, JAXX_Collective_Op collective_op) {
+  static std::once_flag gemm_init_warned;
+  std::call_once(gemm_init_warned, []() {
+    std::cerr << "[CollectiveGemmInitFFI] Deprecation: This API is deprecated and will be removed "
+                 "in September 2026. Use GemmInitV2FFI instead."
```
jberchtold-nvidia, line 162: Clever! I hadn't thought of using the initialize phase as a way to add C++ FFI warnings without introducing runtime overhead at each non-cuda-graphed invocation

transformer_engine/jax/csrc/extensions/cgemm_helper.cpp
phu0ngng: It will work without an unused return in modern C++, but I think it is not a good practice.

transformer_engine/jax/csrc/extensions.h
phu0ngng: The one that we have in attention is slightly different. There, we don't define a struct, and XLA automatically decodes the attribute, but rather dynamically queries the attribute at runtime, which is harder to catch debug https://github.com/NVIDIA/TransformerEngine/blob/d40b9de312c3cacec73b10a71b8d8419a11824c8/transformer_engine/jax/csrc/extensions/attention.cpp#L338-L347. 

I think this struct approach is better, it's flexible enough so that we don't need to introduce a new API whenever we want to add an attribute, it should also allow an optional attribute, i.e. struct variable with a default value, and also less bug prone.

transformer_engine/jax/cpp_extensions/gemm.py
phu0ngng: Let's me rename it to NVTE_.

PR #2740, opened 2026-03-05
[JAX] Collective GEMM with FP8 and MXFP8 support

phu0ngng:
# Description

This PR extends the JAX Collective GEMM support with DelayedScalingFP8, CurrentScalingFP8, and MXFP8.  
Unit tests for those quantization recipes are added. In addition, this PR also cleans up the test infrastructure in the collective gemm tests.

Note that Collective GEMM + MXFP8 requires all dimensions of the GEMM operands to be divisible by 128. 
Besides, in the case of CGEMM + MXFP8 + AllGather, the block scales are still all-gathered in the critical path, unlike the quantized data, which is collectively gathered overlapping with the computation. 

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or new content)
- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/.github/workflows/trigger-ci.yml
+++ b/.github/workflows/trigger-ci.yml
@@ -58,6 +58,8 @@ jobs:
            || github.actor == 'vthumbe1503'
            || github.actor == 'shengfangd'
            || github.actor == 'kainzhong'
+           || github.actor == 'cspades'
+           || github.actor == 'jomitchellnv'
          )
     steps:
       - name: Check if comment is issued by authorized person
--- a/tests/cpp/operator/CMakeLists.txt
+++ b/tests/cpp/operator/CMakeLists.txt
@@ -25,6 +25,7 @@ add_executable(test_operator
                test_normalization.cu
                test_normalization_mxfp8.cu
                test_memset.cu
+               test_splits_to_offsets.cu
                test_multi_cast_transpose.cu
                test_multi_padding.cu
                test_multi_unpadding.cu
--- a/tests/cpp/operator/test_grouped_gemm.cu
+++ b/tests/cpp/operator/test_grouped_gemm.cu
@@ -20,6 +20,7 @@
 #include <transformer_engine/cast.h>
 #include <transformer_engine/gemm.h>
 #include <transformer_engine/recipe.h>
+#include <transformer_engine/swizzle.h>
 #include <transformer_engine/transformer_engine.h>
 
 #include "../test_common.h"
@@ -32,6 +33,7 @@ namespace {
 enum class InputCase {
   kFP8Current,
   kBF16,
+  kMXFP8,
 };
 
 enum class ShapeCase {
@@ -44,8 +46,8 @@ enum class ShapeCase {
 size_t grouped_setup_workspace_size(const size_t num_tensors) {
   const size_t ptr_bytes = num_tensors * sizeof(void*);
   const size_t int_bytes = num_tensors * sizeof(int);
-  // Layout: 6 pointer arrays (A, B, C, D, alpha, beta) + 6 int arrays (a_rows, a_cols, b_rows, b_cols, d_rows, d_cols)
-  size_t size = 6 * ptr_bytes + 6 * int_bytes;
+  // Layout: 8 pointer arrays (A, B, C, D, alpha, beta, a_scale, b_scale) + 6 int arrays
+  size_t size = 8 * ptr_bytes + 6 * int_bytes;
   const size_t alignment = 256;
   size = ((size + alignment - 1) / alignment) * alignment;
   return size;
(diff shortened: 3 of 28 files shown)


examples/jax/collective_gemm/test_layernorm_mlp_grad.py
```diff
@@ -149,11 +158,17 @@ def run_layernorm_mlp_grad_tests(args, mesh=None):
...
-        enabled=False,
-        recipe=None,
+        enabled=use_quantization,
+        recipe=recipe,
         mesh_resource=MeshResource(dp_resource=DP_AXIS, tpsp_resource=TPSP_AXIS),
     ):
+        # Build quantizer_set inside autocast so create_set() reads the global recipe
+        # for correct fwd/bwd dtypes. One set per dense layer (GEMM1=AG, GEMM2=RS).
+        quantizer_set = QuantizerFactory.create_set() if use_quantization else noop_quantizer_set
+        quantizer_sets = (quantizer_set, quantizer_set)
```
jberchtold-nvidia: For DelayedScaling we need two distinct quantizer sets instead of 1 duplicated set, otherwise they will share an amax_history, right?

I think here we only run one iteration, so in practice both scales will default to 1, so there isn't an issue. But probably best practice to create two separate quantizer sets

transformer_engine/jax/quantize/helper.py
```diff
@@ -173,6 +175,60 @@ def is_scaling_mode_supported(
     return _is_scaling_mode_supported[scaling_mode], _reason_for_no_scaling_mode[scaling_mode]
 
 
+_RECIPE_NAME_TO_SCALING_MODE = {
+    "DelayedScaling": ScalingMode.DELAYED_TENSOR_SCALING,
```
jberchtold-nvidia: This is not a 1:1 mapping because a recipe name could have multiple scaling modes like NVFP4's 1D and 2D block scalings

For the specific case of `is_quantize_recipe_supported`, both NVFP4 scaling modes have the same hardware constraints, so it's functionally the same to just pick one scaling mode.

But what do you think about removing this "Recipe Name -> Scaling Mode" mapping and directly using the "Recipe Name -> Recipe Object" mapping below, then getting the scaling modes for X, DGRAD, and WGRAD using the other config methods in `helper.py` for configs

Something like
```python
def is_quantize_recipe_supported(recipe_name: str, gpu_id=None):
    recipe = get_quantization_recipe(recipe_name)
    config = get_quantize_config_with_recipe(recipe)
    for tensor_source in TensorSource:
         scaling_mode = config.get_scaling_mode(tensor_source)
         is_supported, reason = is_scaling_mode_supported(scaling_mode)
         if not supported:
              return is_supported, reason
    return True, None
````

File: transformer_engine/jax/cpp_extensions/gemm.py [hunk 3]
```diff
@@ -396,6 +396,48 @@ def assert_cublas_requirements(scaling_mode, contracting_size, tensor_name):
...
+    """Reorder tensor so the tpsp axis is leading: reshape (dp, n, tpsp, m, ...), transpose (2, 0, 1, 3, ...)."""
+    assert original_shape[0] % dp_or_fsdp_axis_size() == 0 or original_shape[0] == 1, (
+        f"Original_shape[0]={original_shape[0]} is not divisible by"
+        f" dp_or_fsdp_axis_size()={dp_or_fsdp_axis_size()}"
+    )
+    assert original_shape[1] % tpsp_axis_size() == 0 or original_shape[1] == 1, (
+        f"Original_shape[1]={original_shape[1]} is not divisible by"
+        f" tpsp_axis_size()={tpsp_axis_size()}"
+    )
+    reshaped = tensor.reshape(
```
jberchtold-nvidia, line 420: Will these reshapes trigger breaking of sharding constraints and replication or is it okay since these are only called in the inner impl after custom partitioning has occurred?

phu0ngng: We only do this reorder/reshape in the inner impl, not the outer.

examples/jax/collective_gemm/test_layernorm_mlp_grad.py
phu0ngng: This is indeed an aliasing. Let me go/I should have gone with `n_quantizer_sets=2` instead.

jberchtold-nvidia (approved):
LGTM, thanks!

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -955,12 +990,25 @@ def _parse_operand_output_specs(
...
+        if scaling_mode.is_1d_block_scaling():
+            rhs_scale_specs = rhs_specs
+            # Set the seq spec to None to trigger AG the scales as TE/Common CGEMM does not handle
+            # scale collecting yet
+            if collective_op.is_all_gather:
+                lhs_scale_specs = tuple(
+                    None if i == sequence_dim else s for i, s in enumerate(lhs_specs)
+                )
+            else:
+                lhs_scale_specs = lhs_specs
```
denera, line 1004: This AG is only required for overlap with Userbuffers. We'll conditionally disable it whenever we're using the cuBLASMp backend instead.

No changes needed in this PR, just dropping a note for reference.

PR #2741, opened 2026-03-06
Add guard at lowest JAX version that still supports triton kernel calling

tdophung:
# Description

To provide backward compatibility with older jax versions, we need to have a safeguard in place for jax versions too old to work with triton kernel calling. Using Claude Code to automate bisecting through JAX toolbox nightly containers between Sep 1, 2025 and Oct 1, 2025 (*), I have found that the first passing version of the container starts on Sep 24th, 2025, corresponding to jax 0.8.0.dev20250924 hence the guard is put there.

(*) the date range is determined by having a data point that the officially released jax toolbox ([nvcr.io/nvidia/jax:25.10-py3](http://nvcr.io/nvidia/jax:25.10-py3) fails  while the nightly jax container on Oct 1st passed.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes
Handles jax < 0.8.0.dev20250924 segfault error when calling triton kernels frfom JAX side

# Checklist:

### DIFF
--- a/tests/jax/test_distributed_permutation.py
+++ b/tests/jax/test_distributed_permutation.py
@@ -48,6 +48,7 @@ def _inject_permutation(request):
         return
     import sys
     from transformer_engine.jax.permutation import token_dispatch, token_combine
+
     mod = sys.modules[__name__]
     mod.token_dispatch = token_dispatch
     mod.token_combine = token_combine
--- a/tests/jax/test_distributed_router.py
+++ b/tests/jax/test_distributed_router.py
@@ -50,6 +50,7 @@ def _inject_router(request):
         fused_topk_with_score_function,
         fused_moe_aux_loss,
     )
+
     mod = sys.modules[__name__]
     mod.fused_topk_with_score_function = fused_topk_with_score_function
     mod.fused_moe_aux_loss = fused_moe_aux_loss
--- a/tests/jax/test_fused_router.py
+++ b/tests/jax/test_fused_router.py
@@ -29,6 +29,7 @@ def _inject_router(request):
         fused_topk_with_score_function,
         fused_moe_aux_loss,
     )
+
     mod = sys.modules[__name__]
     mod.fused_topk_with_score_function = fused_topk_with_score_function
     mod.fused_moe_aux_loss = fused_moe_aux_loss
--- a/tests/jax/test_permutation.py
+++ b/tests/jax/test_permutation.py
@@ -32,6 +32,7 @@ def _inject_permutation(request):
         token_combine,
         sort_chunks_by_index,
     )
+
     mod = sys.modules[__name__]
     mod.token_dispatch = token_dispatch
     mod.token_combine = token_combine


transformer_engine/jax/triton_extensions/utils.py
```diff
@@ -162,6 +162,21 @@ def _check_triton_compatibility():
         "If you don't need Triton, use transformer_engine.jax.cpp_extensions instead."
     ) from e
 
+# Minimum jaxlib version required for Triton kernel dispatch to work correctly.
+# jaxlib < 0.8.0.dev20250924 segfaults in pxla.py during Triton kernel execution
```
jberchtold-nvidia: Great work! I think we should expose a function here, something like `is_triton_extension_supported()` so us or users can check as well. Then in our Triton tests we can add something like the following block at the top:

```python
def test_some_triton_extension(...):
    if not is_triton_extension_supported():
        pytest.skip("... same info message about triton jax compatibility in older versions ...")
       return
    main test code
```
Similar to how we guard tests based on compute arch

Without this change, I think the tests on old containers would still fail, though the new error message is much more informative then before!

transformer_engine/jax/triton_extensions/utils.py
```diff
@@ -162,6 +162,21 @@ def _check_triton_compatibility():
...
+# (bisected: last known segfault = jax-2025-09-23, first known pass = jax-2025-09-24).
+_JAXLIB_MIN_VERSION = "0.8.0.dev20250924"
+import jaxlib  # noqa: E402
```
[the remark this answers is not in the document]
phu0ngng: I was also about to comment about jax vs jaxlib.

Besides, we have an existing helper here as well https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/jax/quantize/helper.py#L72-L78, perhaps we should find a way to repurpose this helper, especially since we will need to have jax guard all over places in the future.

On the side note, should we check for a release version, i.e., >= 0.8.0 instead?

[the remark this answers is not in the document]
tdophung: @phu0ngng  I think greptile has analyzed it pretty thoroughly on checking for > dev_version or >release_version. So the decision is between supporting dev versions after 0924 or not. I think we should support these dev versions, just in case other teams have not updated their pipeline to use a newer/release version. The ugly code to have the dev version instead of release is not that ugly and does not need any maintenance anyways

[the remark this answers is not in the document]
tdophung: regarding the other poitns (jax vs jaxlib) and using the `_jax_version_meets_requirement` : I agree and will fix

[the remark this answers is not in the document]
phu0ngng: Well, for the older version, I would prefer to guard against release versions for the two reasons. First, it is very unlikely that users will update their JAX version to another old dev version, but rather another release version. Second, I would prefer not to have to keep track of dev versions in the support matrix. 

On the other side, when we want a guard but need the latest unreleased features, we could add a guard against a dev version, but move to a release version eventually when it is released.

@jberchtold-nvidia for additional opinion.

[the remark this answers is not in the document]
jberchtold-nvidia: I agree with versioning against an official release version. The official release versions are tested more thoroughly than a nightly dev version, so I'd rather have a user be overly cautious and upgrade JAX slightly further than have them only upgrade the minimal amount to a dev version and potentially run into other issues on the particular nightly dev version.

With versioning to an official release version, we will be missing out on a certain period between `[jax_dev_release_where_it_is_fixed, jax_official_release_when_it_is_fixed)`, but I think that is an okay tradeoff for the additional stability I'd expect in an official release compared to a nightly dev release.

transformer_engine/jax/triton_extensions/utils.py
tdophung: I was thinking more about this and I realize that we could also use pytest markers to mark certain tests as triton kernel tests where we can check for the cached jaxlib version (with that exported is_triton_extension_supported() function) without having to add the same boilerplate to each test. 

Another approach is to do it for each test file (by doing conftest.py pytest hook, or 1 call to the skip function with is_triton_extension_supported at the beginning). Currently there are only 4 files: test_permutation.py, test_fused_router.py, anfd the distributed versions of each. However, I think we might want to use either NCCL deepEP or cuTile in replacement of these triton kernels in the future, or keep a combination of them, so this whole file approach might not be ideal, and I would rather do the marker approach above.

tdophung: I was thinking more about this and I realize that we could also use pytest markers to mark certain tests as triton kernel tests where we can check for the cached jaxlib version (with that exported is_triton_extension_supported() function) without having to add the same boilerplate to each test. 

Another approach is to do it for each test file (by doing conftest.py pytest hook, or 1 call to the skip function with is_triton_extension_supported at the beginning). Currently there are only 4 files: test_permutation.py, test_fused_router.py, anfd the distributed versions of each. However, I think we might want to use either NCCL deepEP or cuTile in replacement of these triton kernels in the future, or keep a combination of them, so this whole file approach might not be ideal, and I would rather do the marker approach above.

jberchtold-nvidia: Good point about switching backends, the pytest marker approach sounds good to me!

transformer_engine/jax/version_utils.py
```diff
@@ -0,0 +1,43 @@
...
+    triton_extensions (which would raise immediately on old jax).
+    """
+    return _jax_version_meet_requirement(TRITON_EXTENSION_MIN_JAX_VERSION)
+
+
+__all__ = [
+    "_jax_version_meet_requirement",
+    "is_triton_extension_supported",
+    "TRITON_EXTENSION_MIN_JAX_VERSION",
+]
```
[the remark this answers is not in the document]
tdophung: this is outdated, resolving

transformer_engine/jax/quantize/helper.py
```diff
@@ -40,6 +38,7 @@
     get_all_mesh_axes,
     with_sharding_constraint,
 )
+from transformer_engine.jax.version_utils import _jax_version_meet_requirement
```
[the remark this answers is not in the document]
tdophung: done

tests/jax/utils.py
```diff
@@ -40,6 +44,16 @@
...
 
 
+def require_triton():
+    """Skip the current test module if JAX is too old for Triton kernel support."""
+    if not is_triton_extension_supported():
+        pytest.skip(
+            f"JAX >= {TRITON_EXTENSION_MIN_JAX_VERSION} required for Triton kernel support. "
+            "Triton kernel dispatch segfaults with older jaxlib. "
+            "Upgrade with: pip install --upgrade jax jaxlib"
+        )
```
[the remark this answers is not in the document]
tdophung: dine

transformer_engine/jax/triton_extensions/utils.py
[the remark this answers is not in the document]
tdophung: I changed check to release version 0.8.0

transformer_engine/jax/version_utils.py
```diff
@@ -0,0 +1,42 @@
...
+"""
+
+from functools import lru_cache
+from importlib.metadata import version as get_pkg_version
+
+from packaging.version import Version as PkgVersion
+
+
+@lru_cache(maxsize=None)
+def _jax_version_meet_requirement(version: str):
```
phu0ngng: Nit: remove the `_` since the util is used outside of `quantize` now.

tests/jax/test_distributed_permutation.py
```diff
@@ -34,11 +34,23 @@
...
+@pytest.fixture(autouse=True, scope="function")
+def _inject_permutation(request):
+    """Lazy-load permutation API only for tests marked 'triton'. Other tests run without importing."""
+    if not request.node.get_closest_marker("triton"):
+        yield
+        return
+    import sys
+    from transformer_engine.jax.permutation import token_dispatch, token_combine
+
+    mod = sys.modules[__name__]
```
jberchtold-nvidia, line 52: Why is this editing of `sys.modules` required? Does `from transformer_engine.jax.permutation import token_dispatch, token_combine` not already import `token_dispatch` and `token_combine` into the scope of the current file?

tests/jax/utils.py
```diff
@@ -40,6 +44,17 @@
 NVTE_DEBUG_NUMERICS = bool(int(os.getenv("NVTE_DEBUG_NUMERICS", 0)))
 
 
+def require_triton():
```
jberchtold-nvidia: nit SR: name this something like `require_triton_or_skip_test_file()` to indicate the behavior when Triton is not supported. With the current name `require_triton`, I'd think it'd raise an assertion rather than silently skipping

tests/jax/test_distributed_permutation.py
tdophung: Imports inside the fixture only create names in the fixture’s local scope. The test methods `TestHighLevelPermutationAPI.test*` uses `token_dispatch`, `token_combine`, ... as module level globals. Those names are looked up from the test module, not from inside the fixture. So it cannot "see" these imports.

So we have to put them on the module on line 48-49 for the test methods to see it.

tdophung: I'll also put a short explanation in the comments

jberchtold-nvidia: Got it, makes sense. An explanation in comments would be great, thanks!

PR #2746, opened 2026-03-09
[PyTorch] Fix cross_entropy_forward stride guard for non-contiguous input

### LINKED ISSUE #2734: [PyTorch] cross_entropy_forward: stride guard misses non-contiguous transposed input
**Describe the bug**

`cross_entropy_forward` guards against non-contiguous input at https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/triton/cross_entropy.py#L54

```python
if _input.stride(-1) != 1:
    _input = _input.contiguous()
```

This check is insufficient. If we pass a non-contiguous transposed tensor with `stride(-1) == 1` and `stride(-2) != shape[-1]`, the Triton kernel will use `stride(-2)` for row jumps and reads out of bounds, producing **silently wrong results** without a NaN or an error.

```python
import torch
from transformer_engine.pytorch.cross_entropy import parallel_cross_entropy

s, b, v = 4, 2, 8
torch.manual_seed(42)
logits = torch.randn(s, b, v, device="cuda")         # [S, B, V], contiguous
target = torch.randint(0, v, (b, s), device="cuda")  # [B, S]

logits_transposed = logits.transpose(0, 1)          # [B, S, V], non-contiguous
print(logits_transposed.stride())                   # (8, 16, 1), stride(-2)=16 != V=8

logits_contiguous = logits_transposed.contiguous()  # [B, S, V], contiguous

loss_buggy   = parallel_cross_entropy(logits_transposed, target, 0.0, False, None)
loss_correct = parallel_cross_entropy(logits_contiguous, target, 0.0, False, None)

print(loss_buggy)                              # batch=1: [2.08, 2.08, 2.08, 2.08]  ← wrong
print(loss_correct)                            # batch=1: [4.13, 3.80, 2.11, 2.57]  ← correct
print(torch.allclose(loss_buggy, loss_correct))  # False
```

**Expected behavior**

Both calls should return identical results. The wrong-stride case should either raise an error or be silently corrected (as the `stride(-1)` case already is).

**Suggested fix**

```python
# transformer_engine/pytorch/triton/cross_entropy.py
if _input.stride(-1) != 1 or _input.stride(-2) != _input.shape[-1]:
    _input = _input.contiguous()
```

**Environment details**

- OS: Linux
- Python version: 3.12.3
- PyTorch version: 2.10.0
- Transformer Engine version: 2.11.0 (bug also present in upstream `main` as of 2026-03)
- CUDA version: 13.1

**Device details**

- GPU model: NVIDIA A40
```

Bias92:
The stride guard in cross_entropy_forward only checks stride(-1) != 1,
which misses transposed tensors where stride(-1) == 1 but stride(-2) != shape[-1].
The Triton kernel then uses the wrong row stride and produces silently incorrect results.

Added stride(-2) check, same approach as the backward fix in #2402.

Fixes #2734

Before fix:
Non-contiguous: [2.0794, 2.0794, 2.0794, 2.0794]  ← wrong (same values)
Contiguous:     [4.1277, 3.7957, 2.1120, 2.5712]  ← correct
before error

After fix:
Both return [
4.1277, 3.7957, 2.1120, 2.5712] ✓
after

### DIFF
--- a/tests/pytorch/test_parallel_cross_entropy.py
+++ b/tests/pytorch/test_parallel_cross_entropy.py
@@ -167,6 +167,8 @@ class TestParallelCrossEntropy:
                 reduce_loss=True,
                 ignore_idx=True,
             )
+
+
 def test_non_contiguous_transposed_input():
     """Regression test: stride(-2) != shape[-1] should not produce wrong results."""
     s, b, v = 4, 2, 8
@@ -183,6 +185,6 @@ def test_non_contiguous_transposed_input():
     loss_t = parallel_cross_entropy(logits_transposed, target, 0.0, False, None)
     loss_c = parallel_cross_entropy(logits_contiguous, target, 0.0, False, None)
 
-    assert torch.allclose(loss_t, loss_c), (
-        f"Non-contiguous transposed input gave wrong results: {loss_t} vs {loss_c}"
-    )
+    assert torch.allclose(
+        loss_t, loss_c
+    ), f"Non-contiguous transposed input gave wrong results: {loss_t} vs {loss_c}"


Bias92:
Tested locally on RTX 4060 Ti (WSL2):
- Reproduced the bug with transposed input (Match: False)
- Verified fix produces correct results (Match: True)
- Checked no other incomplete stride guards in the file

Full test suite needs CI — local editable build had C++ extension issues on WSL2.

ptrendx:
Thank you for your contribution @Bias92. I will merge once the CI passes.

PR #2748, opened 2026-03-09
[Core] MXFP8 grouped GEMM + tensor-scaled FP8 fixes

jberchtold-nvidia:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


tests/cpp/test_common.cu
```diff
@@ -1217,6 +1234,66 @@ GroupedBuffers build_grouped_tensor(const std::vector<Tensor*>& tensors,
...
+      }
+
+      // Allocate and copy
+      CudaPtr<> buffer = cuda_alloc(total_bytes);
+      for (size_t i = 0; i < num_tensors; ++i) {
+        tensors[i]->to_cpu();
+        NVTE_CHECK_CUDA(cudaGetLastError());
+        void* dst = static_cast<char*>(buffer.get()) + scale_offsets[i];
+        const void* src = get_cpu_ptr_fn(tensors[i]);
+        NVTE_CHECK_CUDA(cudaMemcpy(dst, src, numels[i], cudaMemcpyHostToDevice));
```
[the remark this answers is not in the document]
vthumbe1503: removed padding for mxfp8

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -383,6 +451,11 @@ inline void init_matmul_desc(cublasLtMatmulDescOpaque_t &matmulDesc, cublasOpera
   NVTE_CHECK_CUBLAS(cublasLtMatmulDescSetAttribute(&matmulDesc,
                                                    CUBLASLT_MATMUL_DESC_BETA_BATCH_STRIDE,
                                                    &alphabeta_batch_stride, sizeof(int64_t)));
+
+  // Fast accumulation mode: 0 = split accumulator (more accurate), 1 = fast accumulator
+  int8_t fastAccuMode = 0;  // Use split accumulator for accuracy
+  NVTE_CHECK_CUBLAS(cublasLtMatmulDescSetAttribute(&matmulDesc, CUBLASLT_MATMUL_DESC_FAST_ACCUM,
+                                                   &fastAccuMode, sizeof(fastAccuMode)));
```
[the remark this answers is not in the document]
vthumbe1503: Going to be addressed in this PR https://github.com/NVIDIA/TransformerEngine/pull/2669

zhongbozhu:
why do we need this PR when we already have this? https://github.com/NVIDIA/TransformerEngine/blob/7fb10d33621c75f4071487232d824fc159a24da7/transformer_engine/common/transformer_engine.cpp#L1232

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -542,6 +542,23 @@ NVTEScalingMode nvte_grouped_tensor_scaling_mode(const NVTEGroupedTensor tensor)
  */
 NVTEShape nvte_get_grouped_tensor_logical_shape(const NVTEGroupedTensor tensor);
 
+/* EXPERIMENTAL FEATURE AND SUBJECT TO CHANGE. */
+/*! \brief Set whether the grouped tensor has GEMM-swizzled scales.
+ *
+ *  \param[in] tensor Grouped tensor.
+ *  \param[in] val    1 if scales are swizzled, 0 otherwise.
+ */
+void nvte_set_grouped_tensor_swizzled_scales(NVTEGroupedTensor tensor, uint8_t val);
```
zhongbozhu: let's remove these two APIs and use in favor of nvte_set_grouped_tensor_param API

transformer_engine/common/transformer_engine.cpp
```diff
@@ -1358,3 +1358,19 @@ NVTEShape nvte_get_grouped_tensor_logical_shape(const NVTEGroupedTensor tensor)
   const auto &t = *transformer_engine::convertNVTEGroupedTensorCheck(tensor);
   return t.logical_shape;
 }
+
+void nvte_set_grouped_tensor_swizzled_scales(NVTEGroupedTensor tensor, uint8_t val) {
```
zhongbozhu: same comment, we already have the API for setting swizzle scales in the general nvte_set_grouped_tensor_param  API

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -392,17 +465,62 @@ inline void set_fp8_scale_pointers(cublasLtMatmulDescOpaque_t &matmulDesc,
...
+
+#if CUBLAS_VERSION >= CUBLAS_MXFP8_GROUPED_GEMM_VERSION
+  // For MXFP8, verify scales are swizzled and set scale mode
+  if (mxfp8_a || mxfp8_b) {
+    NVTE_CHECK(transformer_engine::cuda::cublas_version() >= CUBLAS_MXFP8_GROUPED_GEMM_VERSION,
+               "MXFP8 grouped GEMM requires cuBLAS ", CUBLAS_MXFP8_GROUPED_GEMM_VERSION,
+               "+, but run-time cuBLAS version is ", transformer_engine::cuda::cublas_version());
+  }
+
+  if (mxfp8_a) {
```
zhongbozhu: are we supporting other mixes of A &  B using different precisions? Otherwise shouldn't we assert in the host side that A & B are both mxfp8?

vthumbe1503: You are right that we are not now. I guess this is some sort of safeguard. So we dont have to make any changes in case such a need arises. As far as I remember we do that logic in GEMM.cpp too.

Currently if both inputs are not same type, then CUBLAS would raise an error anyway

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -118,28 +126,47 @@ struct GroupedGemmSetupWorkspace {
...
+    constexpr size_t kPtrAlignment = 16;  // cuBLAS requires 16-byte alignment for pointer arrays
+
+    // Helper to align offset to kPtrAlignment
+    auto align_offset = [&]() {
+      offset = (offset + kPtrAlignment - 1) / kPtrAlignment * kPtrAlignment;
+    };
 
-    // Pointer arrays first (all 8-byte aligned)
+    // Pointer arrays first (all 16-byte aligned for cuBLAS grouped GEMM)
+    align_offset();
```
zhongbozhu, line 143: when will this alignment be not just an noop?

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -159,8 +186,12 @@ struct GroupedGemmSetupWorkspace {
...
     const size_t ptr_size = num_tensors * sizeof(void *);
     const size_t int_size = num_tensors * sizeof(int);
-    // Layout: 6 ptr arrays, then 6 int arrays
-    size_t size = 6 * ptr_size + 6 * int_size;
+    constexpr size_t kPtrAlignment = 16;  // Must match from_buffers
+
+    // Layout: 8 ptr arrays (each 16-byte aligned), then 6 int arrays
+    // Each ptr array takes ptr_size bytes but needs to start at 16-byte boundary
+    auto aligned_ptr_size = ((ptr_size + kPtrAlignment - 1) / kPtrAlignment) * kPtrAlignment;
+    size_t size = 8 * aligned_ptr_size + 6 * int_size;
```
zhongbozhu, line 194: I am not a huge fan of such opaque functions for calculating offsets by hand, can we define the struct using `alignas` and just extract offsets from it? 
Eg. https://en.cppreference.com/w/cpp/language/alignas.html

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -118,28 +126,47 @@ struct GroupedGemmSetupWorkspace {
...
+    auto align_offset = [&]() {
+      offset = (offset + kPtrAlignment - 1) / kPtrAlignment * kPtrAlignment;
+    };
 
-    // Pointer arrays first (all 8-byte aligned)
+    // Pointer arrays first (all 16-byte aligned for cuBLAS grouped GEMM)
+    align_offset();
     ws.A_ptrs = reinterpret_cast<void **>(setup_ws_ptr + offset);
     offset += ptr_size;
+    align_offset();
```
zhongbozhu, line 146: same comment, let's declare alignment in the struct definition and compiler can help us do the alignment

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -266,15 +300,19 @@ inline GroupedOperandSelection select_grouped_operand(const transformer_engine::
...
              "Grouped GEMM operand is missing both row-wise and column-wise data");
 
-  // Currently only unquantized data and tensor-scaled FP8 are supported.
   const auto sm = t->scaling_mode;
-  NVTE_CHECK(sm == NVTE_DELAYED_TENSOR_SCALING,
-             "Grouped GEMM is only supported with unquantized data and tensor-scaled FP8 data");
+  const bool mxfp8 = is_mxfp_scaling(sm);
+
+  // Validate scaling mode
+  NVTE_CHECK(sm == NVTE_DELAYED_TENSOR_SCALING || mxfp8,
```
zhongbozhu, line 324: How about BF16? 

Technically BF16 has been treated as a per-tensor scaling with scaling factor 1, so the scaling_mode is also NVTE_DELAYED_TENSOR_SCALING. But that's not super clear to most users, let's fix the comment.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -296,6 +334,36 @@ inline GroupedOperandSelection select_grouped_operand(const transformer_engine::
...
   };
 
+  // MXFP8: Row-wise and column-wise data are scaled along different dimensions.
+  if (mxfp8) {
+    if (is_A) {
+      if (trans) {
+        NVTE_CHECK(has_row, "Grouped GEMM: MXFP8 transposed A is missing row-wise data");
+        use_rowwise();
+      } else {
+        NVTE_CHECK(has_col, "Grouped GEMM: MXFP8 non-transposed A is missing column-wise data");
```
zhongbozhu, line 364: why not using a lambda use_columnwise here?

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -383,6 +451,11 @@ inline void init_matmul_desc(cublasLtMatmulDescOpaque_t &matmulDesc, cublasOpera
   NVTE_CHECK_CUBLAS(cublasLtMatmulDescSetAttribute(&matmulDesc,
                                                    CUBLASLT_MATMUL_DESC_BETA_BATCH_STRIDE,
                                                    &alphabeta_batch_stride, sizeof(int64_t)));
+
+  // Fast accumulation mode: 0 = split accumulator (more accurate), 1 = fast accumulator
+  int8_t fastAccuMode = 0;  // Use split accumulator for accuracy
```
zhongbozhu: there is no fastAccuMode for sm100, we should be able to remove it?

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -392,17 +465,62 @@ inline void set_fp8_scale_pointers(cublasLtMatmulDescOpaque_t &matmulDesc,
...
+      cublasLtMatmulMatrixScale_t scale_mode = CUBLASLT_MATMUL_MATRIX_SCALE_PER_BATCH_SCALAR_32F;
+      NVTE_CHECK_CUBLAS(cublasLtMatmulDescSetAttribute(
+          &matmulDesc, CUBLASLT_MATMUL_DESC_A_SCALE_MODE, &scale_mode, sizeof(scale_mode)));
+    }
+    void *a_scale_ptrs = A_sel.scale_inv_ptrs;
     NVTE_CHECK_CUBLAS(cublasLtMatmulDescSetAttribute(
-        &matmulDesc, CUBLASLT_MATMUL_DESC_A_SCALE_POINTER, &a_scale_inv, sizeof(a_scale_inv)));
+        &matmulDesc, CUBLASLT_MATMUL_DESC_A_SCALE_POINTER, &a_scale_ptrs, sizeof(a_scale_ptrs)));
   }
   if (is_fp8_b) {
```
zhongbozhu: This part of logic is a bit confusing, since you are putting mxfp8 related logic inside fp8 per-tensor, while in the code above you have an if branch specifically for mxfp8, setting CUBLASLT_MATMUL_MATRIX_SCALE_VEC32_UE8M0.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -471,11 +589,15 @@ __global__ void setup_grouped_gemm_kernel(
...
     int *b_rows, int *b_cols, int *d_rows, int *d_cols, float **alpha_ptrs, float **beta_ptrs,
+    void **a_scale_inv_ptrs, void **b_scale_inv_ptrs,
     // Inputs
     char *a_base, char *b_base, char *c_base, char *d_base, TensorShapeInfo A_meta,
     TensorShapeInfo B_meta, TensorShapeInfo C_meta, TensorShapeInfo D_meta, size_t a_elem_size,
     size_t b_elem_size, size_t c_elem_size, size_t d_elem_size, float *alpha_ptr, float *beta_ptr,
-    size_t num_tensors) {
+    // Scale inputs: for tensor scaling, pass float* and set mxfp8_base to nullptr
+    // For MXFP8, pass nullptr for tensor_scale and set mxfp8_base
+    float *a_tensor_scale, float *b_tensor_scale, char *a_mxfp8_scale_base,
```
zhongbozhu: I would prefer a void * pointer, and an enum field for the recipe so that we can scale to nvfp4 later

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -512,6 +634,22 @@ __global__ void setup_grouped_gemm_kernel(
   // Fill alpha/beta pointers (per-matrix)
   alpha_ptrs[idx] = alpha_ptr + idx;
   beta_ptrs[idx] = beta_ptr + idx;
+
+  // Fill FP8 scale pointers (per-matrix)
+  // For tensor scaling: one float per tensor, indexed by tensor index
+  // For MXFP8: E8M0 blocks, offset computed from data offset (1 scale byte per 32 elements)
+  if (a_tensor_scale) {
```
zhongbozhu: same comment, use void* and then reinterpret cast here based on recipe enum

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -613,6 +771,11 @@ void nvte_grouped_gemm(const NVTEGroupedTensor A, int transa, const NVTEGroupedT
 
   auto setup_workspace = GroupedGemmSetupWorkspace::from_buffers(
       static_cast<char *>(setup_workspace_ptr), num_tensors);
+
+  // Set scale_inv_ptrs from workspace (kernel will fill these arrays for both tensor scaling and MXFP8)
+  A_sel.scale_inv_ptrs = setup_workspace.a_scale_inv_ptrs;
+  B_sel.scale_inv_ptrs = setup_workspace.b_scale_inv_ptrs;
```
zhongbozhu: why only scales needs to be set after `from_buffers`, shouldn't data fields also need to be specified?

transformer_engine/common/include/transformer_engine/transformer_engine.h
pggPL: Yeah, that's leftover since mxfp8 grouped gemm had been prepared before swizzle PR was merged. I removed it.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
pggPL: Added the assertion

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
pggPL: Fixed this, fastAccuMode is taken from config passed in nvte_groupped_gemm, by default it is false. I will need this for hopper support. 

The comment was also bad, fixed it.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
pggPL: Yeah, there is code spaghetti here, i split this into 2 functions: 1 for fp8, 1 for mxfp8.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
pggPL: Done

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
pggPL: I set all other fields in select_grouped_operand. This is the result of weird code organization, I cleaned it a little.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
pggPL: For MXFP8 we do not swap dims, columwnise data has the same shape as rowwise one. I added swap_dims param to use_columnwise and I use it here.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
pggPL: That's problematic, because alignment depends on num_tensors which is known at runtime. I haven't figured out anything cleaner.

PR #2749, opened 2026-03-10
[JAX] Grouped GEMM Refactor to use first_dims and last_dims

jberchtold-nvidia:
# Description

 This PR refactors the grouped GEMM API in the JAX backend to support fully ragged (variable-size per group)       
  dimensions across all tensor axes, replacing the previous single group_sizes parameter with six per-tensor        
  dimension parameters. The motivation is to generalize the interface so that forward and backward (wgrad) passes   
  can be expressed uniformly without special-casing, and to eliminate the need for callers to manually compute and  
  pass matrix dimensions (M, N, K) — these are now derived automatically from XLA buffer descriptors in C++.        
                                                                                                                    
  Addresses issue: #2648                                                                                              
   
  ## Type of change                                                                                                    
                  
 - [ ] Infra/Build change                                                                                              
 - [x] Code refactoring                                                                                                
                                                                                                                    
  ## Changes         
                                                                                                                    
  - grouped_gemm API signature change: replaced the single group_sizes positional argument with six keyword         
  arguments — lhs_first_dims, lhs_last_dims, rhs_first_dims, rhs_last_dims, out_first_dims, out_last_dims — each an
  optional (G,) int32 array describing per-group sizes along that tensor axis (empty (0,) arrays indicate a         
  uniform/non-ragged dimension)
  - Removed explicit M/N/K parameters from C++ FFI: matrix dimensions are now derived automatically from XLA buffer
  shapes inside the C++ handler, eliminating manual dimension computation in Python                                 
  - Removed is_grouped_dense_wgrad flag: the wgrad vs. forward distinction is now inferred from which dimension
  arrays are non-empty (non-empty rhs_first_dims indicates a ragged K contraction dimension, producing a            
  (num_groups, M, N) output)
  - New C++ config struct GroupedGemmV2Config: consolidates lhs_is_trans, rhs_is_trans, and scaling_mode into a     
  single FFI attribute struct, replacing individual attribute bindings                                              
  - New C++ helper make_grouped_tensor() overload: accepts first_dims/last_dims buffers, converts int32 group-size
  arrays to int64 in partitioned int64_workspace slots, and returns updated workspace offset to avoid aliasing      
  - dense.py updated: _grouped_dense_fwd_rule and _grouped_dense_bwd_rule updated to pass group_sizes via the
  appropriate new per-tensor parameter (lhs_first_dims/out_first_dims for forward; rhs_first_dims for wgrad)        
  - Tests updated: TestGroupedDense test cases migrated to the new keyword-argument API with explicit empty_gs = 
  jnp.empty((0,), jnp.int32) sentinels for non-ragged axes
# Checklist:

### DIFF
--- a/transformer_engine/common/fused_attn/fused_attn.cpp
+++ b/transformer_engine/common/fused_attn/fused_attn.cpp
@@ -310,7 +310,7 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
         // architecture
         ((cudnn_runtime_version < 8903 && (sm_arch_ == 80 || sm_arch_ == 90)) ||
          (cudnn_runtime_version >= 8903 && sm_arch_ >= 80 && sm_arch_ < 100) ||
-         (cudnn_runtime_version >= 90700 && sm_arch_ >= 80)) &&
+         (cudnn_runtime_version >= 90700 && sm_arch_ >= 100)) &&
         // sequence length
         ((cudnn_runtime_version < 90000 && max_seqlen_q % 64 == 0 && max_seqlen_kv % 64 == 0) ||
          (cudnn_runtime_version >= 90000)) &&
@@ -534,6 +534,10 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
         std::cout << "Warning: Given combination of sm_arch_ == 120 and cudnn_runtime_version < "
                      "91801 is not supported. "
                   << " Please upgrade your cuDNN version if possible." << std::endl;
+      } else if (deterministic && is_training) {
+        backend = NVTE_Fused_Attn_Backend::NVTE_No_Backend;
+        std::cout << "Warning: Deterministic fused attention on SM120 is not supported."
+                  << std::endl;
       } else {
         // Known missing support for T3HD/TH3D layouts on SM120
         const bool is_t3hd_or_th3d =
--- a/transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
+++ b/transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
@@ -112,7 +112,7 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
   }
 
   const DType ragged_offset_type = cudnn_runtime_version >= 90500 ? DType::kInt64 : DType::kInt32;
-  bool generate_stats = !return_max_logit;
+  bool generate_stats = true;  // Always return stats
   try {
     FADescriptor_v1 descriptor{
         b,
@@ -343,7 +343,7 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
         sdpa_options.set_sink_token(softmax_offset);
       }
 
-      std::shared_ptr<fe::graph::Tensor_attributes> Max, Sum_Exp;
+      std::shared_ptr<fe::graph::Tensor_attributes> Max;
       if (use_ragged_stats) {
         offset_stats =
             mha_graph->tensor(fe::graph::Tensor_attributes()
@@ -357,19 +357,12 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
                                     .set_name("Max")
                                     .set_dim({b, h, s_q, 1})
                                     .set_data_type(fe::DataType_t::FLOAT));
-        Sum_Exp = mha_graph->tensor(fe::graph::Tensor_attributes()
-                                        .set_name("Sum_Exp")
-                                        .set_dim({b, h, s_q, 1})
-                                        .set_data_type(fe::DataType_t::FLOAT));
         if (use_ragged_stats) {
           Max->set_stride({h * s_q, 1, h, 1}).set_ragged_offset(offset_stats);
-          Sum_Exp->set_stride({h * s_q, 1, h, 1}).set_ragged_offset(offset_stats);
         } else {
           Max->set_stride({h * s_q, s_q, 1, 1});
-          Sum_Exp->set_stride({h * s_q, s_q, 1, 1});
         }
         sdpa_options.set_logit_max(Max);
-        sdpa_options.set_score_sum_exp(Sum_Exp);
       }
 
       auto [O, Stats] = mha_graph->sdpa(Q, K, V, std::move(sdpa_options));
@@ -387,13 +380,11 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
         O->set_ragged_offset(offset_o);
       }
 
-      if (!return_max_logit) {
-        Stats->set_output(true).set_data_type(fe::DataType_t::FLOAT).set_dim({b, h, s_q, 1});
-        if (use_ragged_stats) {
-          Stats->set_stride({h * s_q, 1, h, 1}).set_ragged_offset(offset_stats);
-        } else {
-          Stats->set_stride({h * s_q, s_q, 1, 1});
-        }
+      Stats->set_output(true).set_data_type(fe::DataType_t::FLOAT).set_dim({b, h, s_q, 1});
+      if (is_ragged_q && cudnn_runtime_version >= 90600) {
+        Stats->set_stride({h * s_q, 1, h, 1}).set_ragged_offset(offset_stats);
+      } else {
+        Stats->set_stride({h * s_q, s_q, 1, 1});
       }
 
       std::tuple<std::shared_ptr<fe::graph::Tensor_attributes>,  // Q
@@ -403,7 +394,7 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
                  std::shared_ptr<fe::graph::Tensor_attributes>>  // O
           key_tensors_tuple = std::make_tuple(Q, K, V, attn_scale, O);
       auto Stats_tuple =
-          generate_stats ? std::make_tuple(Stats, nullptr) : std::make_tuple(Max, Sum_Exp);
+          return_max_logit ? std::make_tuple(Stats, Max) : std::make_tuple(Stats, nullptr);
       auto bias_tuple = is_bias ? std::make_tuple(bias) : std::make_tuple(nullptr);
       auto softmax_offset_tuple =
           is_softmax_offset ? std::make_tuple(softmax_offset) : std::make_tuple(nullptr);
@@ -1137,6 +1128,16 @@ void fused_attn_arbitrary_seqlen_fwd(
   size_t i = 0;
   if (Aux_CTX_Tensors->size == 0) {
     const auto cudnn_runtime_version = cudnnGetVersion();
+
+    Tensor *output_S = convertNVTETensorCheck(Aux_CTX_Tensors->tensors[i++]);
+    output_S->data.dptr = nullptr;
+    if (q_format == NVTE_QKV_Format::NVTE_THD && cudnn_runtime_version >= 90600) {
+      output_S->data.shape = {num_tokens_q, num_attn_heads, 1};
+    } else {
+      output_S->data.shape = {batch, num_attn_heads, max_seqlen_q, 1};
+    }
+    output_S->data.dtype = DType::kFloat32;
+
     if (return_max_logit) {
       Tensor *output_Max = convertNVTETensorCheck(Aux_CTX_Tensors->tensors[i++]);
       output_Max->data.dptr = nullptr;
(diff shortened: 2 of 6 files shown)


transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -617,137 +559,203 @@ JAXX_GroupedTensorWrapper make_grouped_tensor(Buffer_Type const &data,
...
-  // If we call cuBLAS GEMM for A * B, the output will be:
-  //   C: column-major with size [m, n] --> row-major with size [n, m].
-  // To make the output compatible with JAX, we need to swap A and B in cuBLAS GEMM call.
+// V2 variant: derives data shape from the 2D XLA buffer directly, converts group_sizes
+// int32→int64 per-tensor into int64_workspace, and wires first_dims/last_dims.
+// Only NO_SCALING is supported.
+JAXX_GroupedTensorWrapper make_grouped_tensor(Buffer_Type const &data,
+                                              Buffer_Type const &first_dims,
+                                              Buffer_Type const &last_dims,
+                                              Result_Type int64_workspace,
```
jberchtold-nvidia: There is only 1 int64_workspace but each tensor may have their own variable first_dims or last_dims. Do I need this workspace to be expanded or will all use cases have identical group_sizes here

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1331,17 +1331,57 @@ def impl(
...
 
 
+def _grouped_gemm_lhs_M(lhs_shape_2d: Tuple[int, int], lhs_is_trans: bool) -> int:
+    """Non-contracting output size M from the 2-D LHS buffer."""
+    return lhs_shape_2d[1] if lhs_is_trans else lhs_shape_2d[0]
+
+
+def _grouped_gemm_rhs_N(rhs_shape_2d: Tuple[int, int], rhs_is_trans: bool, num_groups: int) -> int:
+    """Non-contracting output size N from the 2-D RHS buffer."""
+    return rhs_shape_2d[0] // num_groups if rhs_is_trans else rhs_shape_2d[1]
```
phu0ngng: Suggest calling it lhs_non_contracting_dims and rhs_non_contracting_dims as M and N are still ambiguous.

Besides, I think we should not assume that lhs and rhs are 2D but can be N-D.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1352,53 +1392,85 @@ def abstract(
...
-        is_grouped_dense_wgrad,
         use_async_d2h_group_sizes,
         use_v2_ffi,
     ):
         """
         Grouped GEMM operation.
 
         Args:
-            lhs_data: Left-hand side input matrix data, 1D flattened array
+            lhs_data: Left-hand side input matrix data, 2D array [rows, cols]
```
phu0ngng: When the LHS needs to be transposed, we won't be able to have a 2D shape. 

Also, I would prefer us not to reshape/merge any axes until C++. Looking into the future, especially when we have a solution to handle the EP part, we may not need to go with shard_map anymore.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1352,53 +1392,85 @@ def abstract(
         rhs_data_aval,
         rhs_scale_inv_aval,
         bias_aval,
-        group_sizes_aval,
+        lhs_first_dims_aval,
+        lhs_last_dims_aval,
+        rhs_first_dims_aval,
+        rhs_last_dims_aval,
+        out_first_dims_aval,
+        out_last_dims_aval,
```
phu0ngng, line 1454: Why does `out_xxx_dims_aval` need to be the inputs for the primitives? Can't the primitive come up with that after having other dims and contracting dims info?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1875,13 +1951,35 @@ def _can_use_v2_grouped_gemm(
...
 def grouped_gemm(
     lhs: Union[jnp.ndarray, GroupedScaledTensor1x],
     rhs: Union[jnp.ndarray, GroupedScaledTensor1x],
-    group_sizes: jnp.ndarray,
+    lhs_first_dims: jnp.ndarray = None,  # (G,) int32 if LHS squashed first dim varies, else None/(0,)
+    lhs_last_dims: jnp.ndarray = None,  # (G,) int32 if LHS squashed last dim varies, else None/(0,)
+    rhs_first_dims: jnp.ndarray = None,  # (G,) int32 if RHS squashed first dim varies, else None/(0,)
+    rhs_last_dims: jnp.ndarray = None,  # (G,) int32 if RHS squashed last dim varies, else None/(0,)
+    out_first_dims: jnp.ndarray = None,  # (G,) int32 if output first dim varies, else None/(0,)
+    out_last_dims: jnp.ndarray = None,  # (G,) int32 if output last dim varies, else None/(0,)
```
phu0ngng: Either the GroupedScaledTensor should carry this information, or one should be able to interpolate this from grouped_sizes + contracting_dims.

transformer_engine/jax/cpp_extensions/gemm.py
jberchtold-nvidia: Agreed that it doesn't need to be an input the the `grouped_gemm` API. To avoid differing inner/outer primitive signatures, I've kept this as an arg to the primitive but am now deriving out first and last dims from the inputs inside the `grouped_gemm` function instead of requiring the user to specify it.

transformer_engine/jax/quantize/tensor.py
```diff
@@ -388,9 +391,11 @@ def __init__(
         flatten_axis,
         original_shape,
         group_axis=0,
+        last_dims=None,
```
phu0ngng: Nit: I think last_dims and first_dims should be positioned together.

transformer_engine/jax/quantize/tensor.py
```diff
@@ -388,9 +391,11 @@ def __init__(
...
     ):
         self.flatten_axis = flatten_axis
-        self.group_sizes = group_sizes
+        self.first_dims = first_dims
+        self.last_dims = last_dims
         self.original_shape = original_shape
         self.group_axis = group_axis
```
phu0ngng: Since we store first_dims and last_dims now, I think we no longer need the `group_axis`.

transformer_engine/jax/quantize/tensor.py
```diff
@@ -455,6 +470,36 @@ def tree_flatten(self):
         )
         return (children, aux_data)
 
+    @classmethod
+    def tree_unflatten(cls, aux_data, children):
+        """Reconstructs the tensor from its flattened representation."""
```
phu0ngng: I think we should be able to reuse the base `tree_unflatten` as the order is still `cls(*children, *aux_data)`.

transformer_engine/jax/dense.py
```diff
@@ -490,7 +491,8 @@ def _grouped_dense_fwd_rule(
                 is_colwise=False,
                 data_layout="N",
                 flatten_axis=ctx_kernel.flatten_axis,
-                group_sizes=ctx_kernel.group_sizes,
+                first_dims=ctx_kernel.first_dims,
+                last_dims=ctx_kernel.last_dims,
```
phu0ngng: I think we don't need to pass dims here, as the tensors should already carry them.

transformer_engine/jax/dense.py
```diff
@@ -518,15 +521,29 @@ def _grouped_dense_fwd_rule(
...
+            group_axis=0,
+            original_shape=grouped_gemm_x.shape,
+        )
+        grouped_gemm_kernel = GroupedNoScaleTensor(
+            data=grouped_gemm_kernel,
+            first_dims=None,
+            last_dims=None,
+            group_axis=0,
+            original_shape=grouped_gemm_kernel.shape,
+        )
```
phu0ngng: How about making the `grouped_quantize` to return GroupedNoScaleTensor when the quantizer set is empty?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1352,53 +1383,103 @@ def abstract(
         rhs_data_aval,
         rhs_scale_inv_aval,
         bias_aval,
-        group_sizes_aval,
+        lhs_first_dims_aval,
+        lhs_last_dims_aval,
+        rhs_first_dims_aval,
+        rhs_last_dims_aval,
+        out_first_dims_aval,
+        out_last_dims_aval,
```
phu0ngng, line 1454: But the out_xxx_dims could be the return buffers. Why should it be input buffers?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1352,53 +1383,103 @@ def abstract(
...
+            rhs_first_dims_aval,
+            rhs_last_dims_aval,
+            out_first_dims_aval,
+            out_last_dims_aval,
+            num_groups,
+        )
+
+        # Derive output shape from N-D buffer shapes using axis_boundary.
+        lhs_shape = lhs_data_aval.shape
+        rhs_shape = rhs_data_aval.shape
```
phu0ngng: Can't do this as the input could be both 1D.

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -1185,31 +1179,41 @@ def grouped_quantize(
...
-        - The group_axis is currently fixed to 0
+          along axis 0
         - The quantizer's q_layout determines whether row-wise, column-wise, or both
           quantization is applied
     """
 
     if quantizer is None:
-        if isinstance(x, NoScaleTensor):
+        if isinstance(x, GroupedNoScaleTensor):
+            assert amax is None, (
```
phu0ngng: After cleaning the `FSDP`, I think `amax` is no longer relevant and should be cleaned as well.

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -829,33 +753,36 @@ Error_Type GroupedGemmV2FFI(cudaStream_t stream, Buffer_Type lhs_data, Buffer_Ty
 XLA_FFI_DEFINE_HANDLER_SYMBOL(GroupedGemmV2Handler, GroupedGemmV2FFI,
                               FFI::Bind()
                                   .Ctx<FFI_Stream_Type>()  // stream
-                                  .Arg<Buffer_Type>()      // lhs_data
+                                  .Arg<Buffer_Type>()      // lhs_data (2D)
```
phu0ngng: Why 2D?

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -619,137 +619,99 @@ JAXX_GroupedTensorWrapper make_grouped_tensor(Buffer_Type const &data,
...
-  auto bias_ptr = has_bias ? reinterpret_cast<uint8_t *>(bias.untyped_data()) : nullptr;
-  auto bias_dtype = convert_ffi_datatype_to_te_dtype(bias.element_type());
+// Returns num_gemms from the first non-empty per-tensor group_sizes buffer,
+// falling back to the element count of alpha for the uniform-batch case.
+size_t grouped_gemm_num_gemms(Buffer_Type const &lhs_first_dims, Buffer_Type const &lhs_last_dims,
+                              Buffer_Type const &rhs_first_dims, Buffer_Type const &rhs_last_dims,
+                              Buffer_Type const &out_first_dims, Buffer_Type const &out_last_dims,
+                              Buffer_Type const &alpha) {
+  if (lhs_first_dims.element_count() > 0) {
+    return lhs_first_dims.dimensions()[0];
```
phu0ngng: If the first_dims is intended to be with shape (G,), isn't `element_count()` and `dimensions()[0]` should return the same result?

transformer_engine/jax/csrc/extensions/gemm.cpp
jberchtold-nvidia: Stale comment, I've remove this now

transformer_engine/jax/cpp_extensions/quantization.py
jberchtold-nvidia: Good catch, I've removed it now

transformer_engine/jax/csrc/extensions/gemm.cpp
jberchtold-nvidia: Good point, I've switched to use element count in both the check and the return value for consistency

transformer_engine/jax/quantize/tensor.py
```diff
@@ -365,34 +366,34 @@ class GroupedScaledTensor1x(ScaledTensor1x):
     where elements are grouped along a specified axis.
 
     Attributes:
-        group_sizes: Array containing the size of each group
+        first_dims: Per-group sizes of the first (row) 2D dim, or None if not ragged
```
jberchtold-nvidia, line 369: TODO: Document these better that first dims and last dims refers to 2D shapes on the C++ side but we keep larger shapes on the Python/XLA shape to preserve user's input sharding

phu0ngng (approved):

PR #2750, opened 2026-03-10
[PyTorch] Fix fuser so it releases tensors properly

kainzhong:
# Description

When training llama3 405B we observe OOMs with passing `use_te_op_fuser=True` in Megatron. This is because `OperationFuser` fails to properly release tensors that are saved for backward.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

In `_OperationFuserAutogradFunction`'s backward, we use
```
saved_tensors = restore_from_saved(func_ctx.tensor_objects, func_ctx.saved_tensors)
```
to restore the saved tensor from backward. However, we didn't detach `tensor_objects` from `func_ctx`, where `func_ctx` will not be released until the next iteration due to pytorch's internal mechanism (pytorch would clear tensor saved by `save_for_backward`, but not for tensors attached to `ctx` directly). Therefore the attached tensors are not be released in time and caused higher memory usage. After manually set `func_ctx.tensor_objects` to None these tensors can be properly freed.

| Before this fix | After this fix |
| --- | --- |
| image | image |

~~In addition, in `BasicLinear`'s backward, I manually called `clear_tensor_data` on the weight tensor if it's allocated by the quantizer (`w is not self.weight`). This is not necessary since with the previous fix where the weight tensor will still be released after fuser's backward, so it's more of an optimization.~~ Removed this from the PR because it's not compatible with the mocked fuser backward in tests.

(Tested on Megatron-LM's llama3 8B example with manually setting `use_te_op_fuser=True` in `gpt_builders`)

# Checklist:

### DIFF
--- a/.github/workflows/deploy_nightly_docs.yml
+++ b/.github/workflows/deploy_nightly_docs.yml
@@ -7,6 +7,7 @@ name: Deploy nightly docs
 on:
   push:
     branches: [ "main" ]
+  workflow_dispatch:
 jobs:
   build:
     uses: ./.github/workflows/docs.yml
@@ -21,9 +22,8 @@ jobs:
             name: "te_docs"
             path: "html"
       - name: Prepare for pages
-        uses: actions/upload-pages-artifact@v1.0.7
+        uses: actions/upload-pages-artifact@v3
         with:
-          name: github-pages
           path: "html"
   deploy:
     needs: prepare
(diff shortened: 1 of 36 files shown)


ksivaman:
@kainzhong Could you take a look at the failed CI?

kainzhong:
OK I think the mocked `fuser_backward` method doesn't like me saving `w = None` in `op_forward` when it does `device = w.device`. I'll just get rid of this optimization to merge PR first and handle it later, since as long as we detach `tensor_objects` from `func_ctx` the memory should still be deallocated in time when fuser's backward finishes...

PR #2758, opened 2026-03-13
[Common] Improved fused MoE aux loss kernel for large # of experts

denera:
# Description

Eliminates expensive cluster management API and minimizes number of atomic ops to optimize perf for larger number of experts.

TODO: Perf testing on all archs.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


transformer_engine/common/fused_router/fused_moe_aux_loss_v2.cu
```diff
@@ -0,0 +1,145 @@
...
+#include "../util/logging.h"
+#include "../utils.cuh"
+#include "common/util/cuda_runtime.h"
+#include "utils.h"
+
+namespace transformer_engine {
+namespace fused_router {
+
+template <typename DataType, typename IndexType>
+__global__ void fused_moe_aux_loss_forward_kernel_v2(const DataType* probs,
```
ptrendx: Could we make a little better name for this kernel rather than just v2? E.g. "large" vs "small" or something like that would be more understandable.

transformer_engine/common/fused_router/fused_moe_aux_loss_v2.cu
```diff
@@ -0,0 +1,145 @@
...
+                                                     const IndexType* tokens_per_expert,
+                                                     int total_num_tokens, int num_experts,
+                                                     int num_rows, int num_cols, int topk,
+                                                     float coeff, float* Const_buf) {
+  // -----------------------------------------------------------------------
+  // 1) Compute the constant coefficient (identical for all threads)
+  // -----------------------------------------------------------------------
+  const float C_coeff = (num_experts * coeff) / topk / total_num_tokens / total_num_tokens;
```
ptrendx: If it is a constant for the entire kernel, why don't we just compute it on the CPU and pass the value to the kernel instead?

denera: That's going to make the atomicAdd at the end read the coefficient off global memory, no?

Computing it on-device means we're reading off the register to multiply with the block sum and accumulating.

transformer_engine/common/fused_router/fused_moe_aux_loss_v2.cu
```diff
@@ -0,0 +1,145 @@
...
+                                        int total_num_tokens, int num_experts, int num_rows,
+                                        int num_cols, int topk, float coeff, NVTETensor aux_loss,
+                                        NVTETensor Const_buf, cudaStream_t stream) {
+  NVTE_API_CALL(nvte_fused_moe_aux_loss_forward_v2);
+  using namespace transformer_engine;
+  fused_router::fused_moe_aux_loss_forward_v2(
+      *convertNVTETensorCheck(probs), *convertNVTETensorCheck(tokens_per_expert), total_num_tokens,
+      num_experts, num_rows, num_cols, topk, coeff, *convertNVTETensorCheck(aux_loss),
+      *convertNVTETensorCheck(Const_buf), stream);
+}
```
ptrendx: Why do we do it as a separate nvte_ function? We should route to the right kernel internally based on the arguments.

transformer_engine/common/fused_router/fused_moe_aux_loss_v2.cu
```diff
@@ -0,0 +1,145 @@
...
+  shmem_block[threadIdx.x] = thread_sum;
+  __syncthreads();
+
+  const int warp_id = threadIdx.x / kThreadsPerWarp;
+  const int lane_id = threadIdx.x % kThreadsPerWarp;
+  if (warp_id == 0) {
+    CompType block_sum = warp_reduce_on_shmem(shmem_block, static_cast<int>(blockDim.x),
+                                              ReduceFuncType::SUM, lane_id);
+    if (lane_id == 0) {
+      atomicAdd(&Const_buf[1], block_sum * Const_buf[0]);
```
ptrendx: You should not read from the const_buf[0] as this is a race condition between the 1st block writing this value and the other blocks using the value. Since it is the constant value pass it to the kernel as argument and just reuse that across the threads.

transformer_engine/common/fused_router/fused_moe_aux_loss_v2.cu
ptrendx: If you pass it as an argument to the kernel then it will be a read from a constant memory, which is very fast (and it will be promoted to a register anyway if needed).

denera: Actually, it looks like the Coeff_buf is already on device coming into the `nvte_` call. Both PyTorch and JAX interfaces expect the C_coeff to be on device here outside the kernel and reused in the backward pass.

We can still compute on the CPU before the kernel launch but there's no escaping the `Coeff_buf[0] = C_coeff;` write in the kernel.

denera:
@ptrendx Greptile's P2 issue seems to stem from the fact that `check_shared_memory_capacity_num_experts()` is built for checking shared memory size vs. `num_experts` but we actually compute shared memory size based on `num_cols`. In practice though, `num_experts == num_cols` from the way we invoke all the fused router APIs from the framework side. It's not clear to me why the `num_experts` and `num_cols` were ever set up as separate parameters for these fused router functions in the first place, but I didn't want to make that change in this PR. I'd like to talk to someone who is familiar with the E2E use cases and understand whether there is ever a possibility of these two being different values somehow before we streamline the function signatures in a separate PR.

ptrendx (approved):

PR #2759, opened 2026-03-13
[PyTorch] [torch.compile] transformer_engine.pytorch.autocast suport inside torch.compile

pggPL:
# Description

Enable `torch.compile(fullgraph=True)` for FP8 autocast by moving compile-visible mutable state off class attributes, avoiding tracing through support checks, and adding test.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Move mutable FP8 autocast state from direct `cls` attribute writes to a dataclass-backed singleton object, because `torch.compile` does not support writes directly to class attributes.
- Replace `lru_cache`-based support checks with explicit module-level caches and mark the wrapper functions with `@torch.compiler.assume_constant_result` so `torch.compile` does not trace into `check_*_support()`.
- Add `torch.compile` coverage for FP8 autocast using a custom test module; the test is more involved because there is currently no simple TE layer that supports both FP8 and `torch.compile`.
- Make `DelayedScaling` explicitly unsupported under `torch.compile` and raise a clear error for that case.

# Checklist:

# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


tests/pytorch/test_torch_compile.py
```diff
@@ -0,0 +1,329 @@
...
+
+import pytest
+import torch
+
+from torch._opaque_base import OpaqueBaseMeta
+from torch._library.opaque_object import (
+    get_opaque_type_name,
+    register_opaque_type,
+    MemberType,
+)
```
ptrendx: I'm less concerned with the private module, but more concerned about trying to load this test in the older pytorch environment, since the opaque object is very new. So I agree that we need to guard this import.

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -915,9 +920,21 @@ def _recompute(ctx):
             # backward is not in autocast context, so we set the state here
             # we also have to set the quantizer states to what they were before the forward pass (only relevant for DelayedScaling recipe)
             final_autocast_state = (
-                FP8GlobalStateManager.get_autocast_state()
+                FP8GlobalStateManager.quantization_state.fp8_enabled,
+                FP8GlobalStateManager.quantization_state.fp8_calibration,
+                FP8GlobalStateManager.quantization_state.fp8_recipe,
+                FP8GlobalStateManager.quantization_state.fp8_distributed_group,
+                FP8GlobalStateManager.quantization_state.is_first_fp8_module,
+                FP8GlobalStateManager.quantization_state.fp8_graph_capturing,
```
ptrendx: Why can't we use the helper function here?

transformer_engine/pytorch/quantization.py
```diff
@@ -87,8 +89,49 @@ def check_fp8_block_scaling_support() -> Tuple[bool, str]:
...
+        _FP8_BLOCK_SCALING_SUPPORT = _compute_fp8_block_scaling_support()
+    return _FP8_BLOCK_SCALING_SUPPORT
+
+
 def check_recipe_support(recipe: Recipe) -> None:
     """Check if the given recipe is supported."""
+    if torch.compiler.is_compiling() and isinstance(recipe, DelayedScaling):
+        raise RuntimeError(
+            "DelayedScaling is not supported under torch.compile yet. "
+            "Use Float8CurrentScaling, MXFP8BlockScaling, or Float8BlockScaling instead."
```
ptrendx: We should probably just say to use "other recipes", otherwise this line will become stale.

transformer_engine/pytorch/quantization.py
```diff
@@ -87,8 +89,49 @@ def check_fp8_block_scaling_support() -> Tuple[bool, str]:
...
+    """Return if fp8 block scaling support is available."""
+    global _FP8_BLOCK_SCALING_SUPPORT
+    if _FP8_BLOCK_SCALING_SUPPORT is None:
+        _FP8_BLOCK_SCALING_SUPPORT = _compute_fp8_block_scaling_support()
+    return _FP8_BLOCK_SCALING_SUPPORT
+
+
 def check_recipe_support(recipe: Recipe) -> None:
     """Check if the given recipe is supported."""
+    if torch.compiler.is_compiling() and isinstance(recipe, DelayedScaling):
```
ptrendx, line 130: How often do we call this check (and so how often will this get called)? What is the overhead of this line?

transformer_engine/pytorch/quantization.py
```diff
@@ -103,6 +146,10 @@ def check_recipe_support(recipe: Recipe) -> None:
 
 def get_default_fp8_recipe() -> Recipe:
     """FP8 recipe with default args."""
+    assert not torch.compiler.is_compiling(), (
+        "get_default_fp8_recipe() must not be called during torch.compile tracing. "
+        "Pass an explicit recipe to te.autocast() instead of relying on the default."
+    )
```
ptrendx, line 152: Why?

transformer_engine/pytorch/quantization.py
```diff
@@ -398,90 +418,81 @@ def add_fp8_tensors_to_global_buffer(
...
                     fp8_meta[fp8_meta_tensor_key].amax_history
                 )
-                cls.global_scale_buffer[key].append(fp8_meta[fp8_meta_tensor_key].scale)
-            fp8_meta[index_in_buffer].append(len(cls.global_amax_buffer[key]) - 1)
+                FP8GlobalStateManager.quantization_state.global_scale_buffer[key].append(
+                    fp8_meta[fp8_meta_tensor_key].scale
+                )
+            fp8_meta[index_in_buffer].append(
+                len(FP8GlobalStateManager.quantization_state.global_amax_buffer[key]) - 1
+            )
```
ptrendx: When we have this long chain where we get to quantization_state every time, could we just grab the reference to it at the beginning once and then just use that rather than going through the state manager every time?

transformer_engine/pytorch/quantization.py
pggPL: Once per autocast - I guess we need to check it every time because the recipe can change. This is_compiling() is simple flag lookup, so it should be fast.

transformer_engine/pytorch/quantization.py
pggPL: I changed the message to describe why - creation of recipe object is not traceable, so user should do it outside the compiled region. I don't see the benefit of changing that - and possibly we would like to have some state in recipe(), so it should be created beforehand.

ptrendx (approved):

PR #2763, opened 2026-03-14
[JAX] MXFP8 Grouped Quant+GEMM

jberchtold-nvidia:
# Description

TE/JAX integrations of the V2 MXFP8 grouped quantization kernel and the V2 MXFP8 grouped GEMM which are both cuda-graph-safe.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add new primitive and FFI for V2 grouped quantize that currently only supports MXFP8
- Extend V2 grouped GEMM to support MXFP8
- For both V1 and V2, move swizzling from grouped GEMM FFI to grouped quantize FFI. This is required because currently V2 can only do swizzling when fused with quantization; an independent swizzle kernel that supports ragged groups is not available.
    - This entails updating the tests and dequantization logic for Q->DQ tests to support preswizzled scales.
- Some small kernels added to TE common to handle int32 -> int64 and offset calculations due to JAX's int32 dtype limitation

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/gemm.py
+++ b/transformer_engine/jax/cpp_extensions/gemm.py
@@ -2477,8 +2477,10 @@ def grouped_gemm(
         # Both V1 and V2 quantize produce pre-swizzled scales (V1 via
         # set_with_gemm_swizzled_scales, V2 via nvte_group_quantize). Require that
         # grouped_quantize has set pre_swizzled=True on the input tensors.
-        assert lhs.pre_swizzled, "lhs must be pre-swizzled for MXFP8 1D scaling"
-        assert rhs.pre_swizzled, "rhs must be pre-swizzled for MXFP8 1D scaling"
+        if not lhs.pre_swizzled:
+            raise ValueError("lhs must be pre-swizzled for MXFP8 1D scaling")
+        if not rhs.pre_swizzled:
+            raise ValueError("rhs must be pre-swizzled for MXFP8 1D scaling")
 
     if use_v2_ffi:
         additional_arg_0 = jnp.ones((num_gemms,), jnp.float32)  # alpha


transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -52,6 +52,53 @@
 __all__ = ["quantize", "quantize_dbias", "grouped_quantize", "grouped_dbias"]
 
 
+def _build_scale_spec(x_spec, scale_shape, mesh):
```
jberchtold-nvidia: Is this required for grouped GEMM or can it be removed?

transformer_engine/jax/csrc/extensions/quantization.cpp
```diff
@@ -494,5 +501,171 @@ XLA_FFI_DEFINE_HANDLER_SYMBOL(GroupedQuantizeHandler, GroupedQuantizeFFI,
...
+  // scale_inv in GEMM-swizzled layout.  Enable the fused swizzle so the kernel
+  // writes scales in the layout the GEMM will consume directly.
+  uint8_t swizzle_flag = 1;
+  nvte_set_grouped_tensor_param(out_grouped, kNVTEGroupedWithGEMMSwizzledScales, &swizzle_flag,
+                                sizeof(swizzle_flag));
+
+  QuantizationConfigWrapper quant_config{};
+  nvte_group_quantize(in_grouped, out_grouped, quant_config, stream);
+
+  nvte_destroy_grouped_tensor(in_grouped);
```
jberchtold-nvidia: Use TE common's GroupedTensorWrapper. Leave GEMM as is for now, but add TODO to refactor JAXX_GroupedTensorWrapper out in a follow-up PR

transformer_engine/jax/permutation.py
```diff
@@ -497,8 +497,9 @@ def _token_combine_bwd_rule(
                 hidden_size,
             )
             # The backward kernel only writes to positions that tokens map to.
-            # Padded positions may contain uninitialized (NaN) values - replace with zeros.
-            inp_grad = jnp.where(jnp.isnan(inp_grad), 0.0, inp_grad)
+            # Padded positions may contain uninitialized values (NaN, inf, or garbage).
+            # Replace any non-finite values with zeros.
+            inp_grad = jnp.where(jnp.isfinite(inp_grad), inp_grad, 0.0)
```
jberchtold-nvidia: Is this change needed?

tests/jax/test_custom_call_compute.py
```diff
@@ -1115,6 +1119,61 @@ def test_grouped_qdq(
 
         assert_dequantized_grouped_scaled_tensor(scaled_tensor, x)
 
+    @pytest.mark.skipif(not is_mxfp8_supported, reason=mxfp8_unsupported_reason)
+    def test_grouped_quantize_v1_pre_swizzled(
```
jberchtold-nvidia: Can I consolidate the tests in this file to re-use existing tests and document which shapes are for V1 and which for V2 instead of adding a bunch of new similar tests

tests/jax/test_custom_call_compute.py
```diff
@@ -77,6 +78,9 @@
 supported_recipes = helper.get_supported_quantization_recipes()
 supported_recipes = [pytest.param(r, id=r.__class__.__name__) for r in supported_recipes]
 
+is_v2_grouped_gemm_supported = get_device_compute_capability(0) >= 100
+v2_grouped_gemm_unsupported_reason = "V2 grouped GEMM requires SM100+ (Blackwell or newer)"
```
tdophung: maybe we should wrap this into utils somewhere, and reuse to guard the all calls to V2 grouped GEMM, not just from test_custom_call_compute

tests/jax/test_custom_call_compute.py
```diff
@@ -1084,16 +1094,25 @@ def test_grouped_qdq(
         key = jax.random.PRNGKey(0)
         subkeys = jax.random.split(key, 2)
 
-        # *32 so that the input shapes works for MXFP8
-        input_shape = (m * 32, n)
+        # Use 128 multiplier for V2-eligible MXFP8 shapes (both M and K 128-aligned)
```
tdophung: make it clearer that the 128 aligned is a cuBLASLt thing while 32 multiplier is a scaling factor for MXFP8 applying to chunks of 32 elements thing

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -2069,13 +2072,86 @@ def _can_use_v2_grouped_gemm(
...
+                        "The TE V2 grouped GEMM for MXFP8 requires the product of the first"
+                        " dimensions (up to axis_boundary) of RHS to be divisible by 128, but got"
+                        f" {rhs_first_dim} with rhs_shape={rhs_shape} and"
+                        f" rhs_axis_boundary={rhs_axis_boundary}, and"
+                        " NVTE_JAX_ENFORCE_V2_GROUPED_GEMM is enabled."
+                    )
+                return False
+        # V2 MXFP8 also requires that the "last" dimension (after axis_boundary) of both
+        # operands is a multiple of 128.  The V2 GEMM setup kernel computes per-group
+        # scale pointers as ``data_offset / 32``, which equals ``K_blocks * last_dim``.
```
tdophung: this took me  a bit to understand, not sure if you should clarify what K_blocks is as it is not defined in this file. If after 2nd read and it still feels pretty trivial then feel free to SR

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -1048,7 +1095,16 @@ def abstract(
...
         rowwise_out_aval = jax.core.ShapedArray(shape=rowwise_out_shape, dtype=out_dtype)
 
-        amax_aval = jax.core.ShapedArray(shape=(group_sizes_aval.size,), dtype=jnp.float32)
+        use_v2 = GroupedQuantizePrimitive._use_v2_kernel(scaling_mode, x_aval.shape, flatten_axis)
+        if use_v2:
+            # V2 path: 5th output is int64_workspace laid out as:
+            #   [n_groups int64 group_sizes | n_groups+1 int64 offsets]
+            # = (2*n_groups + 1) * sizeof(int64_t) bytes stored as uint8.
+            n_groups = group_sizes_aval.size
+            fifth_out_aval = jax.core.ShapedArray(shape=((2 * n_groups + 1) * 8,), dtype=jnp.uint8)
```
tdophung: fifth output seems like a bad name for this. Maybe group_sizes_and_offsets?

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -1006,6 +1007,52 @@ class GroupedQuantizePrimitive(BasePrimitive):
...
+             divisible by 128, matching the V2 grouped GEMM constraint so that the
+             two always agree on V1 vs V2.
+
+        Falls back to V1 when constraints are not met. V1 supports arbitrary shapes
+        but performs a D2H copy of group_sizes (not CUDA-graph safe).
+        """
+        if ScalingMode(scaling_mode) != ScalingMode.MXFP8_1D_SCALING:
+            return False
+        # Require SM100+ so V2 quantize (fused swizzle) is only used alongside V2 GEMM.
+        if get_min_device_compute_capability() < 100:
```
tdophung, line 1039: in gemm.py, you check for `get_device_compute_capability` but here it is `get_min_device_capability`. These would be okay if all GPUs on the systemm is the same compute cap (which is most of our products, maybe minus Galaxy ones, I don't remember clearly). But for consistency, please use the same thing.

tdophung: Same for the test file too

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -1414,6 +1414,24 @@ __global__ void convert_int32_to_int64_kernel(const int32_t *src, int64_t *dst,
...
+// Used to convert per-expert slice counts to per-expert row counts for multi-dim tensors.
+__global__ void convert_int32_to_int64_with_multiplier_kernel(const int32_t *src, int64_t *dst,
+                                                              size_t n, int64_t multiplier) {
+  size_t idx = blockIdx.x * blockDim.x + threadIdx.x;
+  if (idx < n) dst[idx] = static_cast<int64_t>(src[idx]) * multiplier;
+}
+
+// Computes exclusive prefix sums: offsets[0]=0, offsets[i]=sum(first_dims[0..i-1]*last_dim).
+// Produces n_groups+1 values. Single-threaded sequential scan; n_groups is typically small.
+__global__ void compute_grouped_tensor_offsets_kernel(const int64_t *first_dims, int64_t *offsets,
```
tdophung, line 1427: I have an idea for this in case n_groups ever gets large: do 32 threads cumsum in blocks  then warp shfl to reduce local sums to 1 sum.

tests/jax/test_custom_call_compute.py
tdophung: probably would also make the def grouped_gemm in gemm.py shorter?

transformer_engine/jax/cpp_extensions/quantization.py
tdophung: oh I see that it is updated_amax for V1. Not sure what would be the best name here given that it is different purposes in the 2 versions

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -2036,48 +2037,303 @@ def _should_enforce_v2_grouped_gemm() -> bool:
...
+    )
+    empty_gs = jnp.empty((0,), jnp.int32)
+    active_group_sizes = next(
+        (
+            gs
+            for gs in [lhs.first_dims, lhs.last_dims, rhs.first_dims, rhs.last_dims]
+            if gs.size > 0
+        ),
+        empty_gs,
+    )
```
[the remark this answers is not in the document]
jberchtold-nvidia: Fixed in latest version

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -2036,48 +2037,303 @@ def _should_enforce_v2_grouped_gemm() -> bool:
...
+    if lhs_layout_is_T:
+        lhs_contract_dim = tuple((lhs_ndim - 1 - i) % lhs_ndim for i in lhs_contract_dim)
+    if rhs_layout_is_T:
+        # For rhs [G, K, N], need to exclude the G dim from contract_dim
+        if lhs_first_dims.size > 0 or lhs_last_dims.size > 0:  # fwd/dgrad: rhs has G as first dim
+            rhs_contract_dim = tuple(
+                (rhs_ndim - 1 - i) % (rhs_ndim - 1) + 1 for i in rhs_contract_dim
+            )
+        else:
+            rhs_contract_dim = tuple((rhs_ndim - 1 - i) % rhs_ndim for i in rhs_contract_dim)
```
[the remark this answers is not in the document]
jberchtold-nvidia: Fixed

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
jberchtold-nvidia: Sounds good! Currently the kernel runtime is pretty small relative to our other kernels and our n_groups per device is fairly small with EP, but good idea for future if n_groups per device gets bigger

transformer_engine/jax/cpp_extensions/quantization.py
jberchtold-nvidia: Good catch, thanks! I've updated to get the changes in this PR to use min device capability

transformer_engine/jax/cpp_extensions/quantization.py
jberchtold-nvidia: Good point, this is a bad name. Instead of this overloaded 5th output, I've instead made both FFIs use 6 outputs and left the workspace empty on V1 for consistency. For V2, if we ever want to support delayed scaling we would need this updated amax output anyways

transformer_engine/jax/cpp_extensions/gemm.py
jberchtold-nvidia: I've reworded this so it's clearer. I believe we could support cases where this dim is not divisible by 128, there is no inherent limitation in the GEMM afaik. But currently the grouped quantize and grouped GEMM setup kernels do not handle these offsets correctly except for when this dim is divisible by 128 for simplicity

tests/jax/test_custom_call_compute.py
jberchtold-nvidia: This isn't solely due to cuBLASLt. The grouped quantize kernel also has these alignment requirements. I've refactored this test code to be less coupled to the internal V1/V2 logic and instead tried to select a handful of test cases that should cover both V1 and V2, and whether V1 or V2 is selected is more of an implementation detail than visible at the test-level (except for some small notes next to the configs to show both V1 and V2 should be covered).

tests/jax/test_custom_call_compute.py
jberchtold-nvidia: Good idea, I've decided to simplify the test code to make it less coupled to V1/V2. I still have some comments to indicate which test cases should trigger V1/V2, but there is less V1/V2 logic in the tests themselves and it is left as more of an internal implementation detail.

jberchtold-nvidia: Separately, I've also simplified the `grouped_gemm` function as I agree that function body was too complex. It is now refactored into several helper functions. It could be cleaned up further, but it's at least better than it was previously. Thanks!

tests/jax/test_custom_call_compute.py
tdophung: Thanks for the refactoring. I still think that the comments should disttinguid between different alignmentt requirements

tdophung (dismissed):
LGTM

PR #2784, opened 2026-03-20
[Common][JAX] Add CUB TopK MaxPairs interface

huanghua1994:
# Description

This PR introduces the new CUB TopK API for large N and K values.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Added `3rdparty/cccl` as a dependency since the CTK on the machine might not be new enough
- Added `transformer_engine/common/util/cub.cu` as the enter point to the CUB TopK function
- Added JAX FFI interface and JAX tests

# Checklist:

### DIFF
--- a/transformer_engine/common/util/cub.cu
+++ b/transformer_engine/common/util/cub.cu
@@ -5,13 +5,15 @@
  ************************************************************************/
 
 #include <transformer_engine/cub.h>
-#include "../common.h"
-#include <cuda/std/execution>
+
 #include <cub/device/device_topk.cuh>
+#include <cuda/std/execution>
+
+#include "../common.h"
 
 void nvte_topk(cudaStream_t stream, const NVTETensor keys_in, const NVTETensor values_in,
-               NVTETensor keys_out, NVTETensor values_out, NVTETensor workspace,
-               int num_items, int k, size_t workspace_bytes) {
+               NVTETensor keys_out, NVTETensor values_out, NVTETensor workspace, int num_items,
+               int k, size_t workspace_bytes) {
   NVTE_API_CALL(nvte_topk);
   using namespace transformer_engine;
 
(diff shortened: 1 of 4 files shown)


transformer_engine/common/util/cub.cu
```diff
@@ -0,0 +1,54 @@
...
+ ************************************************************************/
+
+#include <transformer_engine/cub.h>
+
+#include <cub/device/device_topk.cuh>
+#include <cuda/std/execution>
+
+#include "../common.h"
+
+void nvte_cub_topk(cudaStream_t stream, const NVTETensor keys_in, const NVTETensor values_in,
```
ptrendx: Why would the user of this API care if it is using CUB or something else? The name should be just nvte_topk.

transformer_engine/common/CMakeLists.txt
```diff
@@ -262,8 +273,11 @@ target_link_libraries(transformer_engine PUBLIC
 
 target_include_directories(transformer_engine PRIVATE
                            ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES})
+# Use CCCL from 3rdparty instead of the one from CUDA Toolkit
 target_include_directories(transformer_engine SYSTEM PRIVATE
-                           ${CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES}/cccl)
+                           ${CCCL_INCLUDE_DIR}/thrust
```
ptrendx, line 278: Why do we need thrust there? Are we using it anywhere?

transformer_engine/common/util/cub.cu
ptrendx: In general I don't think we need to expose the fact that it is using CUB under the hood in the API - could you remove those references?

transformer_engine/common/CMakeLists.txt
huanghua1994: If we don't include the thrust here, we will get the following errors in compilation:

 /usr/local/cuda-13.1/targets/x86_64-linux/include/cccl/thrust/system/cuda/config.h:91:2: error: #error The version of CUB in your include path is not compatible with this release of Thrust. CUB is now included in the CUDA Toolkit, so you no longer need to use your own checkout of CUB. Define THRUST_IGNORE_CUB_VERSION_CHECK to ignore this.

I think it is better using thrust in CCCL instead of using thrust in CTK.

transformer_engine/common/util/cub.cu
huanghua1994: I have removed `cub_` in function names and class names.

ptrendx:
Closing in favor of #2890

PR #2785, opened 2026-03-20
add high precision init weights to fully_shard example

pstjohn:
Updates the `fully_shard` example to demonstrate how to use `preserve_high_precision_init_val` from the FusedAdam optimizer with FSDP2. It turns out this actually is broken on ToT, so this fixes a small issue where those init vals don't get attached to the correct local_tensor in the weight DTensor.

Adds a test to ensure this stays functional

### DIFF
--- a/examples/pytorch/quantized_model_init/fully_shard.py
+++ b/examples/pytorch/quantized_model_init/fully_shard.py
@@ -52,6 +52,11 @@ NUM_LAYERS = 3
 SEQ_LEN = 32
 BATCH_PER_RANK = 2
 NUM_STEPS = 5
+# DTYPE is used for both params_dtype and activation tensors in this example.
+# float32 is chosen for params_dtype so that the high-precision init values
+# (which seed the optimizer's FP32 master weights) avoid a lossy BF16→FP8→FP32
+# round-trip.  Using float32 for activations as well keeps the example simple;
+# in production you would typically use BF16 activations inside te.autocast().
 DTYPE = torch.float32
 
 
@@ -226,6 +231,7 @@ def main():
         save_file(fp32_state, save_path)
         dist_print(f"\nSaved FP32 model ({len(fp32_state)} params) to {save_path}")
 
+    dist.barrier()  # wait for rank 0 to finish file I/O
     dist.destroy_process_group()
 
 
--- a/tests/pytorch/distributed/fsdp2_tests/run_fsdp2_fused_adam.py
+++ b/tests/pytorch/distributed/fsdp2_tests/run_fsdp2_fused_adam.py
@@ -333,7 +333,9 @@ def test_fused_adam_fp8_high_precision_init(recipe_name):
         local = param._local_tensor if isinstance(param, DTensor) else param
         hp_val = getattr(local, "get_high_precision_init_val", lambda: None)()
         if hp_val is not None:
-            optimizer.set_scaled_state(param, "master_param", hp_val.to(device=device))
+            optimizer.set_scaled_state(
+                param, "master_param", hp_val.to(device=device, dtype=torch.float32)
+            )
             local.clear_high_precision_init_val()
 
     # Verify high-precision init values are cleared after seeding


examples/pytorch/quantized_model_init/fully_shard.py
```diff
@@ -93,52 +93,51 @@ def main():
...
         master_weight_dtype=torch.float32,
     )
-    dist_print("Using FusedAdam with master_weights=True.")
 
-    # ── 6. Training loop ─────────────────────────────────────────────
+    # ── 6. Seed master weights from high-precision init values ───────
+    # By default, FusedAdam initializes master weights by dequantizing
+    # the FP8 parameters, which introduces quantization noise.  Instead,
+    # we seed them from the original BF16 init values preserved in step 2.
+    for param in model.parameters():
```
jomitchellnv: Do we know if params_dtype=torch.float32 would fix this issue and not require us to manually retrieve BF16 weights and manually cast them to FP32?

examples/pytorch/quantized_model_init/fully_shard.py
```diff
@@ -153,56 +152,22 @@ def main():
...
     dcp.save(
         {"model": model.state_dict(), "optimizer": optimizer.state_dict()},
         checkpoint_id=checkpoint_dir,
     )
     dist_print("  Checkpoint saved (FP8 weights + optimizer state).")
 
-    # Load checkpoint back. Provide empty state dict containers with the
+    # Load checkpoint back.  Provide empty state dict containers with the
     # same structure; DCP fills them from the saved files.
     state_to_load = {"model": model.state_dict(), "optimizer": optimizer.state_dict()}
```
jomitchellnv, line 182: Is it a requirement to store the model weights? If the optimizer weights hold our FP32 master weights, and we simply quantize those to build our FP4/FP8 weights, then why do we need to store the model weights (quantized in FP4/FP8)?

examples/pytorch/quantized_model_init/fully_shard.py
```diff
@@ -238,10 +208,10 @@ def main():
 
         for key, value in full_model_state.items():
             if key in opt_param_states and "master_param" in opt_param_states[key]:
-                # Prefer optimizer's FP32 master weight (maintained throughout training).
+                # Prefer optimizer's FP32 master weight.
```
jomitchellnv, line 221: Can we put this into a utility function? So we dont have to manually rip out the parts from the optimizer?

examples/pytorch/quantized_model_init/fully_shard.py
pstjohn: yeah i think this is indeed what we want

examples/pytorch/quantized_model_init/fully_shard.py
pstjohn: Yeah we should do this. CC @sbhavani

tests/pytorch/distributed/fsdp2_tests/run_fsdp2_fused_adam.py
```diff
@@ -253,6 +265,129 @@ def test_fused_adam_fp8_master_weights_no_meta(recipe_name):
...
+        master_weights=True,
+        master_weight_dtype=torch.float32,
+    )
+
+    for name, param in model.named_parameters():
+        optimizer.initialize_state(param, store_param_remainders=False)
+        local = param._local_tensor if isinstance(param, DTensor) else param
+        hp_val = getattr(local, "get_high_precision_init_val", lambda: None)()
+        if hp_val is not None:
+            optimizer.set_scaled_state(param, "master_param", hp_val.to(device=device))
```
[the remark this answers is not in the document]
pstjohn: Good catch — added the explicit `dtype=torch.float32` cast.

examples/pytorch/quantized_model_init/fully_shard.py
```diff
@@ -93,52 +96,53 @@ def main():
...
+            optimizer.set_scaled_state(
+                param, "master_param", hp_val.to(device=device, dtype=torch.float32)
+            )
+            local.clear_high_precision_init_val()
+
+    dist_print("Optimizer master weights seeded from high-precision init values.")
+
+    # ── 7. Training loop ─────────────────────────────────────────────
     x = torch.randn(SEQ_LEN, BATCH_PER_RANK, HIDDEN_SIZE, dtype=DTYPE, device=device)
     target = torch.randn(SEQ_LEN, BATCH_PER_RANK, HIDDEN_SIZE, dtype=DTYPE, device=device)
```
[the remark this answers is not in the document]
pstjohn: Fair point. Added a comment explaining why `DTYPE = torch.float32` is used for both params and activations, and noting that production code would typically use BF16 activations inside `te.autocast()`.

examples/pytorch/quantized_model_init/fully_shard.py
```diff
@@ -251,14 +226,6 @@ def main():
...
 
-        # Quick verification: all saved tensors are float32.
-        from safetensors.torch import load_file
-
-        loaded = load_file(save_path)
-        for k, v in loaded.items():
-            assert v.dtype == torch.float32, f"{k}: expected float32, got {v.dtype}"
-        dist_print(f"  Verified: all {len(loaded)} tensors are float32.")
-
     dist.destroy_process_group()
```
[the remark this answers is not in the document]
pstjohn: Added `dist.barrier()` before `dist.destroy_process_group()`.

examples/pytorch/quantized_model_init/fully_shard.py
```diff
@@ -93,52 +101,53 @@ def main():
...
         master_weight_dtype=torch.float32,
     )
-    dist_print("Using FusedAdam with master_weights=True.")
 
-    # ── 6. Training loop ─────────────────────────────────────────────
+    # ── 6. Seed master weights from high-precision init values ───────
+    # By default, FusedAdam initializes master weights by dequantizing
+    # the FP8 parameters, which introduces quantization noise.  Instead,
+    # we seed them from the original BF16 init values preserved in step 2.
+    for name, param in model.named_parameters():
```
vthumbe1503, line 137: Given user would have to repeat this in each of their training code snippets. Would it make sense to move this function in fused adam optimizer itself?

vthumbe1503:
Mostly LGTM. I think we can move the logic to avoid quantization errors in optimizer state, within fused_adam.

PR #2796, opened 2026-03-24
[JAX] Add warning if using BSHD and max_segments_per_seq > 1

jberchtold-nvidia:
# Description

Adds a small warning if the user tries to use BSHD with `max_segments_per_seq > 1`

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Adds a warning if the user tries to use BSHD with `max_segments_per_seq > 1`
- Adds a new test to validate this warning is shown correctly

# Checklist:

### DIFF
--- a/transformer_engine/common/fused_attn/fused_attn.cpp
+++ b/transformer_engine/common/fused_attn/fused_attn.cpp
@@ -310,7 +310,7 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
         // architecture
         ((cudnn_runtime_version < 8903 && (sm_arch_ == 80 || sm_arch_ == 90)) ||
          (cudnn_runtime_version >= 8903 && sm_arch_ >= 80 && sm_arch_ < 100) ||
-         (cudnn_runtime_version >= 90700 && sm_arch_ >= 80)) &&
+         (cudnn_runtime_version >= 90700 && sm_arch_ >= 100)) &&
         // sequence length
         ((cudnn_runtime_version < 90000 && max_seqlen_q % 64 == 0 && max_seqlen_kv % 64 == 0) ||
          (cudnn_runtime_version >= 90000)) &&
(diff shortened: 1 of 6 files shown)


transformer_engine/jax/attention.py
```diff
@@ -1436,6 +1436,15 @@ def fused_attn(
...
         )
+    if max_segments_per_seq > 1 and not qkv_layout.is_thd():
+        warnings.warn(
+            f"max_segments_per_seq={max_segments_per_seq} is set but qkv_layout={qkv_layout} is "
+            "not a THD layout. max_segments_per_seq > 1 only applies when using THD layouts "
+            "(e.g. QKVLayout.T3HD, QKVLayout.THD_T2HD, QKVLayout.THD_THD_THD) for sequence "
+            "packing. With non-THD layouts this parameter has no effect.",
+            UserWarning,
+            stacklevel=2,
+        )
```
KshitijLakhani, line 1447:
```suggestion
    if max_segments_per_seq > 1 and not qkv_layout.is_thd():
        warnings.warn(
            f"max_segments_per_seq={max_segments_per_seq} is set but qkv_layout={qkv_layout} is "
            "not a THD layout. max_segments_per_seq > 1 only applies when using THD layouts "
            "(e.g. QKVLayout.T3HD, QKVLayout.THD_T2HD, QKVLayout.THD_THD_THD) for sequence "
            "packing.",
            UserWarning,
            stacklevel=2,
        )
```

tests/jax/test_fused_attn.py
```diff
@@ -1586,3 +1587,65 @@ def test_backward(
             swa,
             seq_desc_format,
         )
+
+
+class TestMaxSegmentsPerSeqWarning:
```
KshitijLakhani: Thanks for testing this warning out, Jeremy !
I think negative testing is crucial in such cases..
However, I'm thinking that merging a test case for just checking a warning might be overkill. 
We've got quite a few warnings in TE but we do not test them (from at least what I'm aware of) so I'd suggest we drop the test and just keep the warning change in fused attn.
Local negative testing to ensure that the warning gets triggered, followed by a passing CI for fused attn tests should be enough, I think . Thoughts ?

jberchtold-nvidia: That sounds good to me, I'll remove the new test

jberchtold-nvidia: Updated

PR #2753, opened 2026-03-11
[Pytorch] Add QuantizedTensor support in FusedAdam.step for MXFP8BlockScaling and Float8BlockScaling quantized model init.

jomitchellnv:
Summary                                                         

  - Add a QuantizedTensor code path in FusedAdam.step() that routes block-scaling quantized parameters (MXFP8, Float8Blockwise,
  NVFP4) through the FP32 Adam kernel on master weights, then writes back via quantize_() after the optimizer step
  - Add FSDP2 all-gather hooks (fsdp_pre_all_gather / fsdp_post_all_gather) to Float8BlockwiseQTensor, enabling end-to-end FSDP2
  training with Float8BlockScaling recipe
  - Add meta-device initialization pattern for FSDP2 + block-scaling quantized tensors (device="meta" + reset_parameters() after
  sharding), required because wrapper subclass tensors have data_ptr() == 0

  Details

  FusedAdam QuantizedTensor path (fused_adam.py):
  - After the existing Float8Tensor check (per-tensor FP8 → FP8 kernel), a new QuantizedTensor check catches block-scaling types
  and routes them to the FP32 master-weight Adam kernel
  - After all kernels run, a post-step loop calls local_p.quantize_(master_w.data) to write updated FP32 master weights back to
  both rowwise and columnwise quantized data
  - Raises RuntimeError if master_weights=False (required for block-scaling) or capturable=True (quantize writeback can't be
  CUDA-graph captured)

  Float8BlockwiseQTensor FSDP2 hooks (float8_blockwise_tensor.py):
  - fsdp_pre_all_gather: Transposes columnwise data/scales from (K, M) to (M, K) before all-gather so FSDP2's dim0 concatenation
  works correctly. Always sends both rowwise and columnwise (the GEMM kernel needs both forms since they have different shapes,
  unlike MXFP8 which stores both in the same shape)
  - fsdp_post_all_gather: Transposes columnwise back to (K, M), repads scale_inv to GEMM alignment (multiples of 4),
  constructs/updates the tensor
  - Adds aten.as_strided and aten.slice no-op handlers in __torch_dispatch__ to prevent FSDP2 from silently dequantizing the
  reconstructed tensor via QuantizedTensor's default dispatch
  - Only supports 2D block scaling (block_scaling_dim=2, the default for weights)

  Meta-device init for FSDP2 (run_fsdp2_fused_adam.py):
  - _build_model() accepts use_meta_device param; creates on device="meta" when fp8_init=True
[...]

PR #2797, opened 2026-03-24
adds NVFP4 Fused Adam support

jomitchellnv:
# Description

  Summary

  - Add FSDP2 all-gather hooks (fsdp_pre_all_gather / fsdp_post_all_gather) to NVFP4Tensor, enabling end-to-end
  FSDP2 training with NVFP4BlockScaling
  - Add aten.as_strided, aten.slice, and aten.record_stream dispatch handlers required by FSDP2's internal tensor
   operations
  - Remove NVFP4BlockScaling xfails from FSDP2 integration tests now that the hooks are in place

  Details

  Without FSDP2 hooks, FSDP2 attempts data_ptr() on the NVFP4Tensor wrapper subclass and crashes. This PR follows
   the Float8BlockwiseQTensor FSDP2 hooks pattern (the closest analog since NVFP4 also stores columnwise data
  transposed), with NVFP4-specific adjustments:

  - FP4 packing: Data last dim is K//2 (two FP4 values packed per uint8 byte)
  - Scale dtype: uint8 (vs float32 for Float8Blockwise)
  - Block size: 16 (NVFP4_BLOCK_SCALING_SIZE) vs 128 for Float8Blockwise
  - Scale padding: Rowwise scale dim0 padded to round_up(M, 128), columnwise scale dim1 padded to
  round_up(ceil(M/16), 4) — both unpadded before all-gather and repadded after
  - Amax tensors: _amax_rowwise and _amax_columnwise (shape (1,)) passed via metadata rather than all-gathered,
  since they're scalar and identical across ranks

  Test plan

  - pytest tests/pytorch/test_nvfp4_fsdp2_hooks.py -v — 19 single-GPU unit tests covering round-trip shapes, data
   integrity, dequantize correctness, in-place update path, swizzled-scale rejection, and dispatch handlers
  - PYTHONPATH=... torchrun --nproc_per_node=2 -m pytest
  tests/pytorch/distributed/fsdp2_tests/run_fsdp2_fused_adam.py -v -k "fp8_master_weights and NVFP4" — multi-GPU
  FSDP2 + NVFP4 integration
  - PYTHONPATH=... pytest tests/pytorch/distributed/test_torch_fsdp2.py::test_fsdp2_fused_adam_tests -v — full
  FSDP2 fused_adam regression

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -551,6 +551,150 @@ def get_usages(self) -> Dict[str, bool]:
...
+            columnwise_data = columnwise_data.t().contiguous()
+
+        if columnwise_scale_inv is not None:
+            # Unpad dim1 from round_up(ceil(shard_M/16), 4) to ceil(shard_M/16)
+            m_blocks = math.ceil(shard_M / NVFP4_BLOCK_SCALING_SIZE)
+            columnwise_scale_inv = columnwise_scale_inv[:, :m_blocks]
+            # Transpose to (m_blocks, round_up(K, 128)) so M-blocks are in dim0
+            columnwise_scale_inv = columnwise_scale_inv.t().contiguous()
+
+        # Always send both orientations (GEMM needs both for fwd/bwd)
```
vthumbe1503: @jomitchellnv we shouldnt need both orientations. For forward pass only rowwise is needed and for backward pass only columnwise is needed.

Can you refer to this PR for the optimizations
https://github.com/NVIDIA/TransformerEngine/pull/2789
Right now the columnwise allhgather implementation is transposing, allgathering in pre_allgather and then again transposing after post allgather which is expensive.
You can make it better one of the two ways for columnwise allgather
1. Always allgather rowwise data similar to above PR. And if columnwise usage is needed then you can just transpose it in post allgather
2. allgather only columnwise data if columnwise usage is set and then interleave the the columnwise stacked data.
1 is fine to have for now.
I would also add assertions for 2d Scaling similar to fp8 blockscaling implemntation.

tests/pytorch/test_nvfp4_fsdp2_hooks.py
```diff
@@ -0,0 +1,303 @@
...
+@pytest.mark.skipif(not nvfp4_available, reason=reason_for_no_nvfp4)
+class TestNVFP4FSDP2Hooks:
+    """Tests for fsdp_pre_all_gather / fsdp_post_all_gather round-trip."""
+
+    @staticmethod
+    def setup_class(cls) -> None:
+        torch.manual_seed(42)
+        torch.cuda.manual_seed(42)
+
+    @pytest.mark.parametrize("shape", _test_shapes)
```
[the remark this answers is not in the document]
jomitchellnv: ▎ Good catch — fixed. Changed to @classmethod so pytest actually invokes the seed initialization.

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -564,6 +708,44 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+
+        # slice — FSDP2 applies this for shard unpadding.
+        # When the slice covers the full dimension, return self.
+        if func == aten.slice.Tensor:
+            tensor = args[0]
+            dim = args[1]
+            start = args[2]
+            length = args[3]
+            if start == 0 and length == tensor.size(dim):
+                return NVFP4Tensor.make_like(tensor)
```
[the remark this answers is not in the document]
jomitchellnv: Fixed. Renamed length → end, added defensive len(args) checks for optional positional args, and added step ==   1 guard. Also handles the case where start/end are None (the ATen defaults

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -551,6 +551,150 @@ def get_usages(self) -> Dict[str, bool]:
...
+            columnwise_scale_inv = columnwise_scale_inv[:, :m_blocks]
+            # Transpose to (m_blocks, round_up(K, 128)) so M-blocks are in dim0
+            columnwise_scale_inv = columnwise_scale_inv.t().contiguous()
+
+        # Always send both orientations (GEMM needs both for fwd/bwd)
+        rowwise_usage = True
+        sharded_tensors = (rowwise_data, rowwise_scale_inv)
+        columnwise_usage = self._quantizer.columnwise_usage
+        if columnwise_usage:
+            sharded_tensors += (columnwise_data, columnwise_scale_inv)
```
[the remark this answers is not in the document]
jomitchellnv: Added an assertion to fail early if _rowwise_data is None:
  ▎ assert self._rowwise_data is not None, (
  ▎     "FSDP2 requires rowwise data, but _rowwise_data is None. "
  ▎     "Ensure the NVFP4Quantizer was created with rowwise=True."
  ▎ )

transformer_engine/pytorch/tensor/nvfp4_tensor.py
jomitchellnv: Thanks for the pointer to #2789 — that makes sense. I'll adopt option 1: only all-gather rowwise data +
  scales, and derive columnwise locally in post_all_gather if columnwise_usage is set. Will update.

On the 2D scaling assertion — NVFP4 doesn't have _is_2D_scaled like Float8Blockwise, but I can add a guar on with_2d_quantization. Could you clarify what layout constraint you're thinking of? The NVFP4 rowwise scale has M in dim0 (round_up(M, 128), ...) regardless of that flag, so it should be compatible with dim0 all-gather either way. Want to make sure I add the right check.

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -551,6 +551,154 @@ def get_usages(self) -> Dict[str, bool]:
...
+            current_m_blocks = columnwise_scale_inv.shape[1]
+            target_m_blocks = round_up_to_nearest_multiple(
+                math.ceil(full_M / NVFP4_BLOCK_SCALING_SIZE), 4
+            )
+            if current_m_blocks < target_m_blocks:
+                columnwise_scale_inv = torch.nn.functional.pad(
+                    columnwise_scale_inv, (0, target_m_blocks - current_m_blocks)
+                )
+            elif current_m_blocks > target_m_blocks:
+                columnwise_scale_inv = columnwise_scale_inv[:, :target_m_blocks]
```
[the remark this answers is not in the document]
jomitchellnv: Fixed. Added an assertion in fsdp_pre_all_gather that shard_M % NVFP4_BLOCK_SCALING_SIZE == 0, which
guarantees m-blocks align across ranks after all-gather. Converted the elif trim to an unreachable assertion —
if the shard alignment holds, current_m_blocks can never exceed target_m_blocks.

transformer_engine/pytorch/tensor/nvfp4_tensor.py
vthumbe1503: Yes, exactly with_2d_quantization should be asserted to be True. If with_2d_quantization is False then columnwise is not derivable from rowwise data/scales(Although this is not a common use-case)

tests/pytorch/distributed/fsdp2_tests/run_fsdp2_model.py
```diff
@@ -363,7 +363,7 @@ def _train(args):
 @pytest.mark.parametrize("fp8_init", [False, True])
 @pytest.mark.parametrize("layer_type", ["LayerNormLinear", "TransformerLayer"])
 def test_distributed(recipe_name, fp8_init, sharding_dims, layer_type):
-    if recipe_name in ("Float8BlockScaling", "NVFP4BlockScaling") and fp8_init:
+    if recipe_name == "Float8BlockScaling" and fp8_init:
```
vthumbe1503, line 373: Does fp8 block scaling still fail? You had fixed this in here https://github.com/NVIDIA/TransformerEngine/pull/2753
right?

tests/pytorch/test_nvfp4_fsdp2_hooks.py
```diff
@@ -0,0 +1,303 @@
...
+            out=result,
+        )
+        assert result2 is result  # same object
+        torch.testing.assert_close(result2.dequantize(), first_deq)
+
+    def test_swizzled_scales_rejected(self):
+        """Verify that GEMM-swizzled scales raise NotImplementedError."""
+        shape = (512, 256)
+        qt = _make_nvfp4_tensor(shape)
+        qt._with_gemm_swizzled_scales = True
```
vthumbe1503: The right way to do this is to set quantizer.optimize_for_gemm = True and the quantized tensor automatically gets _with_gemm_swizzled_scales attribute set to true with in fact the scales being actually swizzled.

Here we are just setting the attribute to be True without actually swizzling the scales.

tests/pytorch/test_nvfp4_fsdp2_hooks.py
```diff
@@ -0,0 +1,303 @@
...
+        # Columnwise data: each shard contributes (K, shard_M//2) -> repeated gives (K, M//2)
+        expected_col_data = torch.cat([orig_columnwise_data] * world_size, dim=1)
+        assert torch.equal(result._columnwise_data, expected_col_data)
+
+        # Amax values passed through metadata — should be preserved
+        assert torch.equal(result._amax_rowwise, orig_amax_row)
+        assert torch.equal(result._amax_columnwise, orig_amax_col)
+
+    @pytest.mark.parametrize("shape", _test_shapes)
+    def test_round_trip_dequantize(self, shape: Tuple[int, int]):
```
vthumbe1503: There is a lot of code duplication with the previous round trip test without the dequantize. Could we merge this test into that one?

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -564,6 +722,49 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+        if func == aten.as_strided.default:
+            tensor = args[0]
+            shape = args[1]
+            strides = args[2]
+            if (
+                len(shape) == len(strides) == 2
+                and tuple(strides) == (shape[-1], 1)
+                and tuple(shape) == tuple(tensor.size())
+            ):
+                return NVFP4Tensor.make_like(tensor)
```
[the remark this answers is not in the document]
jomitchellnv: Fixed — as_strided now validates storage_offset == 0 and raises NotImplementedError for any non-identity call
   instead of silently falling through

transformer_engine/pytorch/tensor/nvfp4_tensor.py
jomitchellnv: Added an assertion in fsdp_pre_all_gather that with_2d_quantization == True when columnwise data is included  in the all-gather. This ensures we fail early rather than silently passing None columnwise tensors

tests/pytorch/distributed/fsdp2_tests/run_fsdp2_model.py
jomitchellnv: removed it!

tests/pytorch/test_nvfp4_fsdp2_hooks.py
jomitchellnv: I replaced the direct qt._with_gemm_swizzled_scales = True hack with the proper API: creating an
 NVFP4Quantizer with quantizer.optimize_for_gemm = True and quantizing through it. Since NVFP4's C++ path  doesn't wire up optimize_for_gemm yet (hardcoded false with TODO at quantizer.cpp:1731 and :1993), the test  gracefully skips with pytest.skip() explaining why, and will auto-activate once the C++ TODO is addressed.

tests/pytorch/test_nvfp4_fsdp2_hooks.py
jomitchellnv: Merged test_round_trip_dequantize into test_round_trip_data_integrity. The dequantize check
  (orig_deq, result.dequantize(), assert_close) was added at the end of the data integrity test. The separate
  test_round_trip_dequantize method was removed entirely. This eliminates the duplicated setup (tensor creation,
  pre_all_gather, simulate, post_all_gather

vthumbe1503 (dismissed):
LGTM.

vthumbe1503 (approved):

PR #2803, opened 2026-03-25
[PyT][Test] Add xfailing FSDP2 memory leak detection tests

pstjohn:
## Summary

- Add xfailing tests that demonstrate two known FSDP2 + FP8 memory issues (#2681, #2717)
- Tests use layer-level forward hooks and backward memory deltas to detect FP8 temporary tensor accumulation, comparing against a bf16 baseline
- Includes bf16 control tests that validate the measurement methodology passes cleanly

### Issue #2681: FP8 weight copy accumulation during forward
FP8 weight copies created by `te.autocast()` accumulate across layers (~0.68 MiB/layer excess over bf16 baseline). Detected for all 5 recipes with `no_quant_init`.

### Issue #2717: Transpose cache retained after backward
`_create_transpose` tensors persist after backward until the next forward frees them (~3 MiB excess over bf16). Detected for `DelayedScaling` and `Float8CurrentScaling` with `quant_init`.

### New tests (in `run_fsdp2_mem_leak.py`)
| Test | Type | What it checks |
|------|------|---------------|
| `test_bf16_no_excess_forward_memory` | control (PASS) | bf16 per-layer increments are uniform |
| `test_bf16_no_excess_backward_memory` | control (PASS) | bf16 vs bf16 backward delta shows zero excess |
| `test_fp8_temp_accumulation_across_layers` | xfail | FP8 per-layer forward increment exceeds bf16 |
| `test_transpose_cache_retained_after_backward` | xfail | FP8 backward delta exceeds bf16 baseline |

All FP8 tests parametrized over 5 recipes × {no_quant_init, quant_init}.

## Test plan
- [x] `pytest tests/pytorch/distributed/test_torch_fsdp2.py` — all 4 outer tests pass (including existing model and fused_adam tests)
- [x] bf16 control tests PASS
- [x] FP8 accumulation tests XFAIL for affected configurations
- [x] Pre-commit hooks pass

PR #2805, opened 2026-03-26
[PyT] Fix FSDP2 memory leaks for FP8 weight workspaces and transpose caches

### LINKED ISSUE #2681: te.autocast() and fully_shard doesn't de-allocate quantized weights until the backwards pass
When running a model with fsdp2 and te.autocast(), quantized weights are created during the forward pass and not de-allocated until the backwards pass. This essentially undoes the memory savings of FSDP2, since we end up accumulating the entire model's worth of quantized weights on each rank.

https://github.com/NVIDIA/bionemo-framework/blob/main/bionemo-recipes/recipes/llama3_native_te/train_fsdp2_cp.py

Image

https://nvidia.slack.com/archives/C03V462SAMS/p1771004012335309

### LINKED ISSUE #2717: _create_transpose tensor accumulating during FSDP2 with quantized_model_init
with quantized_model_init (DelayedScaling), FusedAdam, and FSDP2, we allocate a `transformer_engine/pytorch/tensor/storage/float8_tensor_storage.py:204:_create_transpose` tensor during the backwards pass that accumulates over all the model layers until the subsequent forward layer

Image

https://nvidia.slack.com/archives/C038G319G6R/p1771868674747129?thread_ts=1771868398.818749&cid=C038G319G6R

pstjohn:
## Summary

Fixes memory leaks where FP8 quantized weight workspaces and transpose caches accumulate during FSDP2 training, defeating FSDP2's per-layer memory savings.

- **Issue #2681**: FP8 quantized weight copies created during forward pass were cached and saved for backward, accumulating across all layers simultaneously instead of being freed per-layer.
- **Issue #2717**: `_create_transpose` tensors allocated during backward persisted on FSDP2's reusable all-gather buffers until the next forward pass.

### Approach

When FSDP2 is detected (via `_get_module_fsdp_state`), the fix applies to `layernorm_mlp.py`, `layernorm_linear.py`, and `linear.py`:

1. **Skip columnwise/transpose creation during forward** — backward's FSDP2 all-gather recreates them
2. **Disable workspace caching** (`cache_name=None`) — prevents `_fp8_workspaces` from retaining per-layer copies
3. **Don't save workspace copies for backward** — re-quantize from FSDP2 all-gathered weight instead
4. **Clear `_transpose` after backward dgrad GEMMs** — prevents transpose data persisting on reusable buffers

Guarded to Float8Quantizer, Float8CurrentScalingQuantizer, and MXFP8Quantizer. Blockwise quantizers (Float8BlockScaling, NVFP4BlockScaling) have separate internal caching not yet addressed.

### Additional fixes

- **Pre-existing `isinstance` bug in `layernorm_mlp.py`**: The FC1 backward path had `isinstance(ctx.fc1_weight_quantizer, QuantizedTensorStorage)` — this was always `False` since a `Quantizer` is never a `QuantizedTensorStorage`, so `update_usage(columnwise_usage=True)` was never called. Fixed to check `ctx.fc1_weight` instead, which is the correct target.

### Test changes

Updates the memory leak detection tests from #2803:
- Removes `xfail` markers for fixed recipes
- Adds targeted `xfail` for blockwise recipes
- Increases backward test tolerance to 1 MiB (from 256 KiB) to account for temporary backward workspace re-creation

## Test plan

- [x] All FSDP2 memory leak tests pass (14 passed, 6 xfailed for blockwise, 2 skipped)
- [x] All existing FSDP2 functional tests pass (model tests, fused_adam tests)
- [x] No regressions in `test_torch_fsdp2.py` (4 passed)

Closes #2681
Closes #2717


transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -793,6 +833,7 @@ def _forward(
             ctx.fc2_weight_requires_grad = fc2_weight.requires_grad
             ctx.fc1_weight = fc1_weight
             ctx.fc2_weight = fc2_weight
+            ctx.is_fsdp2 = fsdp2_skip_columnwise
```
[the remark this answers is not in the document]
pstjohn: Good catch — renamed `ctx.is_fsdp2` to `ctx.fsdp2_skip_columnwise` throughout (forward save + all 4 backward usages) to match the local variable name and make the semantics self-documenting.

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -1361,9 +1422,19 @@ def fc2_wgrad_gemm(
...
+                    fc1_weight = origin_fc1_weight
+                elif ctx.fc1_weight_quantizer is not None:
+                    ctx.fc1_weight_quantizer.set_usage(rowwise=False, columnwise=True)
+                    fc1_weight = ctx.fc1_weight_quantizer(origin_fc1_weight)
+
             # Make sure required data is available
             if ctx.fc1_weight_quantizer is not None and isinstance(
-                ctx.fc1_weight_quantizer, QuantizedTensorStorage
+                ctx.fc1_weight, QuantizedTensorStorage
             ):
```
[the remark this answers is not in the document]
pstjohn: Good eye — this was indeed an incidental fix of a pre-existing bug where the `isinstance` check on the quantizer (not the weight) was always `False`. I'll add a note in the PR description calling this out as a separate bug fix for reviewability.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -978,10 +978,10 @@ std::pair<TensorWrapper, py::object> Float8BlockQuantizer::create_tensor(
   if (internal) {
     // Use direct C API call bypassing pybind11 overhead
     py::dict kwargs;
-    kwargs["rowwise_data"] = py::cast(data_rowwise);
-    kwargs["columnwise_data"] = py::cast(data_colwise);
-    kwargs["rowwise_scale_inv"] = py::cast(scale_inv_rowwise);
-    kwargs["columnwise_scale_inv"] = py::cast(scale_inv_colwise);
+    kwargs["rowwise_data"] = rowwise_usage ? py::cast(data_rowwise) : py::none();
+    kwargs["columnwise_data"] = columnwise_usage ? py::cast(data_colwise) : py::none();
+    kwargs["rowwise_scale_inv"] = rowwise_usage ? py::cast(scale_inv_rowwise) : py::none();
```
vthumbe1503: This change makes sense to me. But why is it affecting only blockwise and not other quantized tensors?

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -295,6 +295,13 @@ def forward(
         # ------------------------------------------------------
         weightmat = weight
         is_weight_param_quantized = False
+        try:
+            from ..distributed import _get_module_fsdp_state
+
+            _get_module_fsdp_state(module)
+            is_fsdp2 = True
+        except (RuntimeError, ImportError):
+            is_fsdp2 = False
```
vthumbe1503: try except can be expensive and can incur python overheads. Can we cache _is_fsdp2 in the module. So we dont have to do it in every forward pass?

transformer_engine/pytorch/csrc/quantizer.cpp
vthumbe1503: Actually better change might be to not allocate the data/scale itself if appropriate usages are not set. Are we not already doing that @pstjohn ?

transformer_engine/pytorch/module/layernorm_linear.py
```diff
@@ -788,6 +840,20 @@ def backward(
...
+                    weight._transpose_invalid = True
+            # FSDP2: Clear blockwise columnwise caches after dgrad GEMM.
+            # (Issues #2681, #2717)
+            if getattr(ctx, "is_fsdp2", False) and hasattr(weight, "_columnwise_data"):
+                if getattr(weight, "_columnwise_data", None) is not None:
+                    weight._columnwise_data = None
+                    weight._columnwise_scale_inv = None
+                    if hasattr(weight, "_amax_columnwise"):
+                        weight._amax_columnwise = None
+
```
vthumbe1503: Can we use weight.update_usage method here to reduce the lines of code? That should take care of eliminating usage buffers that need to be eliminated

vthumbe1503 (commented):
LGTM. Have left a few minor comments.

transformer_engine/pytorch/csrc/quantizer.cpp
vthumbe1503: https://github.com/pstjohn/TransformerEngine/blob/3cb076a7cf929b2087772cef62a49f25319a22da/transformer_engine/pytorch/csrc/quantizer.cpp#L1048-L1049

If rowwise_usage is not set, data_rowwise should be undefined and py::cast should technically already return py::none(). So this change might not be needed

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -1144,6 +1217,25 @@ def backward(
...
+            # (Issues #2681, #2717)
+            if getattr(ctx, "fsdp2_skip_columnwise", False) and hasattr(
+                fc2_weight, "_columnwise_data"
+            ):
+                if getattr(fc2_weight, "_columnwise_data", None) is not None:
+                    fc2_weight._columnwise_data = None
+                    fc2_weight._columnwise_scale_inv = None
+                    if hasattr(fc2_weight, "_amax_columnwise"):
+                        fc2_weight._amax_columnwise = None
+
```
vthumbe1503, line 1250: Same comment as in the layernorm_linear.py file

transformer_engine/pytorch/module/linear.py
```diff
@@ -776,6 +827,21 @@ def backward(ctx, grad_output: torch.Tensor) -> Tuple[Union[torch.Tensor, None],
...
+                        weight_fp8._transpose = None
+                        weight_fp8._transpose_invalid = True
+                # FSDP2: Clear blockwise columnwise caches after dgrad GEMM.
+                # (Issues #2681, #2717)
+                if getattr(ctx, "is_fsdp2", False) and hasattr(weight_fp8, "_columnwise_data"):
+                    if getattr(weight_fp8, "_columnwise_data", None) is not None:
+                        weight_fp8._columnwise_data = None
+                        weight_fp8._columnwise_scale_inv = None
+                        if hasattr(weight_fp8, "_amax_columnwise"):
+                            weight_fp8._amax_columnwise = None
```
vthumbe1503: Same comment as in the layernorm_linear.py file

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -1163,6 +1238,14 @@ def backward(
                 ub_type=tex.CommOverlapType.AG if ctx.ub_overlap_ag else None,
             )
 
+            # FSDP2: Clear columnwise/transpose caches after FC2 dgrad GEMM
+            # to prevent them from persisting on the all-gathered buffer.
+            # (Issues #2681, #2717)
+            if getattr(ctx, "fsdp2_skip_columnwise", False) and isinstance(
+                fc2_weight, QuantizedTensorStorage
+            ):
+                fc2_weight.update_usage(columnwise_usage=False)
```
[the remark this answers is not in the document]
pstjohn: Good catch — fixed both issues:

1. **Cache name**: switched from `fsdp2_skip_columnwise` to the raw `is_fsdp2` flag so workspace caching stays disabled during gradient-checkpoint recomputation.
2. **Backward cleanup**: now store `ctx.is_fsdp2 = is_fsdp2` (the raw flag, independent of `is_recomputation`) and use it for the post-dgrad cleanup guards. `fsdp2_skip_columnwise` still controls forward-time columnwise skipping, which correctly should NOT happen during recomputation.

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -484,18 +489,42 @@ def _forward(
...
+                )
 
+            # FSDP2: Don't cache workspaces — they would persist across
+            # layers, defeating FSDP2 memory savings. (Issue #2681)
+            fc1_cache = (
+                None if (is_first_microbatch is None or fsdp2_skip_columnwise) else "fc1_weight"
+            )
+            fc2_cache = (
+                None if (is_first_microbatch is None or fsdp2_skip_columnwise) else "fc2_weight"
+            )
```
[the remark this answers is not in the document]
pstjohn: Good catch — this is the same root cause as the sibling comment. During recomputation `fsdp2_skip_columnwise = is_fsdp2 and not True = False`, so both the cache name and the backward cleanup guards were incorrectly disabled.

Fixed by using the raw `is_fsdp2` flag for both:
- Cache name now uses `is_fsdp2` directly (consistent with `linear.py` / `layernorm_linear.py`)
- Backward cleanup now reads `ctx.is_fsdp2` (stored separately from `ctx.fsdp2_skip_columnwise`)

transformer_engine/pytorch/module/layernorm_linear.py
pstjohn: Done — added a lazy-cached `is_fsdp2` property on `TransformerEngineBaseModule` that calls `_get_module_fsdp_state()` once and caches the result. All three module files now use `module.is_fsdp2` instead of the try/except block.

transformer_engine/pytorch/module/layernorm_linear.py
pstjohn: Done — replaced the manual cache cleanup with `weight.update_usage(columnwise_usage=False)` across all three module files.

To make this work, I added a fast-path to each storage type's `update_usage()` that handles `columnwise_usage=False` without resolving `rowwise_usage` from the data state. Without this, the existing assertion (`"Must retain some data"`) would fire when FSDP2 resharding had already freed the rowwise data. The fast-path is backward-compatible — existing callers are unaffected.

Changed files: `float8_tensor_storage.py`, `float8_blockwise_tensor_storage.py`, `nvfp4_tensor_storage.py`, `mxfp8_tensor_storage.py`.

transformer_engine/pytorch/module/layernorm_mlp.py
pstjohn: Done — same `update_usage(columnwise_usage=False)` change applied here for both FC1 and FC2 cleanup.

transformer_engine/pytorch/module/linear.py
pstjohn: Done — same `update_usage(columnwise_usage=False)` change applied here.

transformer_engine/pytorch/csrc/quantizer.cpp
pstjohn: Agreed — reverted. The default-constructed `at::Tensor` does cast to a valid Python object, so the explicit check wasn't strictly necessary. Removed to keep the PR scope to Python changes only.

vthumbe1503 (dismissed):

vthumbe1503 (approved):

PR #2811, opened 2026-03-30
[PyTorch] [torch.compile] Split linear forward into forward and setup context.

pggPL:
# Refactor: Split `_Linear.forward` into forward computation and context setup

## Description

This PR splits `_Linear.forward` into three functions:
- **`_linear_forward_impl`** — forward computation (GEMM, quantization, communication)
- **`_linear_setup_ctx`** — saving state into autograd context for backward
- **`_linear_backward`** — backward pass (already existed as a monolith, extracted similarly)

This is done as a step in enabling `torch.compile` for `Linear`.

The design follows these principles:

1. **Simple ctx saves are moved from forward to `_linear_setup_ctx`.**
   For example, `ctx.activation_dtype = activation_dtype`, `ctx.fp8 = fp8`, `ctx.sequence_parallel = sequence_parallel`, and all other direct assignments from `non_tensor_args` to `ctx` now live in `_linear_setup_ctx`.

2. **Values needed in both forward and setup_ctx are computed once in `Linear.forward` (module level) and passed via `non_tensor_args`.**
   For example:
   - `backward_override = FP8GlobalStateManager.get_fp8_recipe().backward_override` — used in `_linear_forward_impl` (to decide `save_original_input`) and in `_linear_setup_ctx` (to set `ctx.backward_override` and apply backward overrides).
   - `custom = is_custom(input_quantizer) or is_custom(weight_quantizer)` — used in `_linear_forward_impl` (to gate quantization path) and saved as `ctx.custom`.
   - `backward_input_needs_gather` — used in `_linear_forward_impl` (to decide input tensor layout) and saved as `ctx.backward_input_needs_gather`.
   - UB debug flag zeroing — in debug mode, all userbuffer overlap flags (`ub_overlap_rs_fprop`, `ub_overlap_ag_fprop`, `ub_overlap_ag_dgrad`, `ub_overlap_rs_dgrad`, `ub_bulk_dgrad`, `ub_bulk_wgrad`) are zeroed once in `Linear.forward` before being passed down, instead of being zeroed separately in forward_impl and setup_ctx.

3. **Values that are costly to compute and deeply coupled with forward computation are computed in `_linear_forward_impl` and returned via `ctx_attrs` dict.**
   For example:
   - `weight_quantizer` — may be mutated during weight quantization (e.g., replaced with `weight._quantizer` for pre-quantized weights).
   - `fsdp_shapes` — returned by `_fsdp_scatter_tensors()` which also modifies tensors in-place.
   - `owns_input` — depends on `saved_inputmat` which is determined during the input quantization/gather logic.
   
   These cannot be easily moved outside the autograd function because they are tightly coupled with the computation. They also won't work directly with `torch.compile` (which can only return tensors from custom operators), but that will be addressed in a follow-up PR.

Most of the lines in this PR are indentation changes (moving code from class method into module-level functions), which makes the diff large but the actual logic changes minimal. Follow-up PRs for `Linear` + `torch.compile` should not have this property.

## Type of change

- [x] Code refactoring

## Checklist:


transformer_engine/pytorch/module/linear.py
```diff
@@ -80,962 +80,1078 @@
...
+            input_quantizer=ctx.input_quantizer,
+            weight_quantizer=ctx.weight_quantizer,
+            grad_input_quantizer=ctx.grad_input_quantizer,
+            grad_weight_quantizer=ctx.grad_weight_quantizer,
+            grad_output_quantizer=ctx.grad_output_quantizer,
+        )
         if ctx.reduce_and_update_bwd_fp8_tensors and not is_graph_capturing():
             nvtx_range_push(f"{nvtx_label}.reduce_and_update_fp8_tensors")
             FP8GlobalStateManager.reduce_and_update_fp8_tensors(forward=False)
             nvtx_range_pop(f"{nvtx_label}.reduce_and_update_fp8_tensors")
```
timmoon10, line 1153: Why does this logic happen outside `_linear_backward`? I see the corresponding forward pass logic also happens outside of `_linear_forward`. I assume it's because we want to avoid calling `is_first_fp8_module` in the forward pass and you want symmetry in the backward pass?

timmoon10 (approved):
LGTM, this is mostly straightforward code reshuffling.

transformer_engine/pytorch/module/linear.py
pggPL: This is very problematic for torch.compile - torch.compile likes if input tensors to the ops are passed explicitly to the OP as an arguments. Thus this part is probelmatic and I don't want them inside the op. By the way, we will not support DelayedScaling for torch.compile because of this reason, but maybe we will figure out it somehow in the future.-

PR #2814, opened 2026-03-31
Pass input_output_alias to TritonAutotunedKernelCall

tdophung:
# Description
https://nvbugspro.nvidia.com/bug/5810384
To remove the WAR that was put in place for this bug.

This should also serves as part 2 to WAR to the intermittent sort_chunks_by_index bug seen before in https://github.com/NVIDIA/TransformerEngine/pull/2730

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

if JAX version >= 0.9.3, which contains the fix to restore all aliased input buffers that was saved away during autotuning: we pass the input_output_alias tuples to TritonAutotunedKernelCall

If JAX version < 0.9.3, which does not contain the fix, we pass an empty dict to the call.

# Checklist:

### DIFF
--- a/transformer_engine/jax/triton_extensions/utils.py
+++ b/transformer_engine/jax/triton_extensions/utils.py
@@ -43,8 +43,10 @@ import jax
 import jax.numpy as jnp
 
 from ..version_utils import (
+    TRITON_AUTOTUNED_INPUT_OUTPUT_ALIAS_MIN_JAX_VERSION,
     TRITON_EXTENSION_MIN_JAX_VERSION,
     is_triton_extension_supported,
+    jax_version_meet_requirement,
 )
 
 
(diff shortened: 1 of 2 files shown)


transformer_engine/jax/triton_extensions/utils.py
```diff
@@ -474,23 +474,22 @@ def lowering(ctx, x, *, block_size):
...
-        #    output can reuse an input's buffer. XLA may or may not honor this.
-        # 2. TritonAutotunedKernelCall's input_output_aliases_with_sizes triggers
-        #    save/restore logic during autotuning (see jaxlib/gpu/triton_kernels.cc:630-701).
-        #
-        # The problem: The save phase (triton_kernels.cc:632) only saves if buffers[input_idx] == buffers[output_idx],
-        # but the restore phase (triton_kernels.cc:697-700) unconditionally iterates over all aliases and tries
-        # to access input_copies[input_idx]. If XLA didn't actually alias the buffers, input_copies[input_idx] doesn't exist, creating an empty vector whose .data() returns nullptr, causing CUDA_ERROR_INVALID_VALUE during the restore memcpy.
-        #
-        # WAR: Don't pass aliases to TritonAutotunedKernelCall.
+        input_output_aliases_with_sizes = ()
```
jberchtold-nvidia, line 479: We should guard this on the JAX version corresponding to the version that has the fixed XLA version as users may still be on older JAX/XLA. I think currently that will be a dev nightly release, so we can guard on that for now and then add a TODO(tdophung) to update to an official release in the future when it's included in the next release

tdophung: i will guard against an official release version so no follow up is needed

tdophung:
manually triggered CI passed. The current CI failure shown above is due to timeout while reserving prenyx in first CI pipeline

PR #2816, opened 2026-03-31
[PyTorch] [torch.compile] Remove internal tensor state from Float8CurrentScalingQuantizer

pggPL:
# Description

Removing tensor state from float8current scaling quantizer to get better torch.compile compatibility.

There are some different ways of dealing with that:
- having workspace in module and passing in into quantize,
- passing scale and amax separately from quantizers

but I have tried them and they seem to be worse. Current solution pays with some cpu overhead (2 at::empty per quantization vs 1 before).

Also, there is some weird behavior of fused adam which is fixed in this PR.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:


transformer_engine/common/multi_tensor/adam.cu
pggPL, line 1: There is a weird behavior in current code. FusedAdam reads scale, amax and scale_inv from quantizer, and uses them to quantize new weights. 

This code is here:
https://github.com/NVIDIA/TransformerEngine/blob/c9035a4854edb67dc896ed0e129e3e7ecbf52251/transformer_engine/common/multi_tensor/adam.cu#L155

So in case of current scaling, it silently assumes that there is correct scale in quantizer. What if for example qunatizers are not alive?

I changed that to computing scale from scale inv, which tensors own (i'm not 100% sure if this is numerically ok, but this is fp32 so i guess do). And in case of current scaling we do not need to store anything.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -897,15 +889,17 @@ void Float8CurrentScalingQuantizer::quantize_impl(const TensorWrapper& input, Te
 
 void Float8CurrentScalingQuantizer::quantize(const TensorWrapper& input, TensorWrapper& out,
                                              const std::optional<TensorWrapper>& noop_flag) {
-  this->quantize_impl(input, out, noop_flag, true);
+  const auto opts = at::TensorOptions().dtype(torch::kFloat32).device(torch::kCUDA);
+  at::Tensor amax_and_scale = at::empty({2}, opts);
```
pggPL, line 902: Here we alloc 2 floats: scale and amax and use them as buffers in qunatization.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -556,14 +551,7 @@ Float8CurrentScalingQuantizer::Float8CurrentScalingQuantizer(const py::handle& q
...
 
-void Float8CurrentScalingQuantizer::set_quantization_params(TensorWrapper* tensor) const {
-  // transfer amax and scale pointer from quantizer to output tensor (only as gpu buffer, no meaningful data in them)
-  tensor->set_scale(scale.data_ptr(), GetTransformerEngineDType(scale.scalar_type()),
-                    getTensorShape(scale));
-  at::TensorOptions opts = opts.dtype(torch::kFloat32).device(torch::kCUDA);
-  tensor->set_amax(amax.data_ptr(), GetTransformerEngineDType(amax.scalar_type()),
-                   getTensorShape(amax));
-}
+void Float8CurrentScalingQuantizer::set_quantization_params(TensorWrapper* tensor) const {}
```
pggPL, line 554: I'm not sure if this is ok, I'm not sure what is the philosophy of `set_quantization_params`, but I guess we are slightly breaking it.

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -273,14 +271,8 @@ def __init__(
         amax: Optional[torch.Tensor] = None,
     ) -> None:
         super().__init__(rowwise=rowwise, columnwise=columnwise)
-        if scale is None:
-            scale = torch.empty(1, dtype=torch.float32, device=device)
-        if amax is None:
-            amax = torch.empty(1, dtype=torch.float32, device=device)
-        self.scale = scale
-        self.amax = amax
+        del device, use_existing_amax, scale, amax  # Kept for backward compatibility
```
timmoon10, line 281: We should raise a warning if someone passes in these kwargs.

transformer_engine/pytorch/csrc/quantizer.cpp
timmoon10: I don't think `set_quantization_params` has a coherent meaning. Making it empty is better.

timmoon10 (dismissed):
I strongly agree with this change. It was kind of perverse for the quantizers to hold temporary workspaces in the quantizer. Two thoughts:

- We originally decided on storing these tensors to reduce the CPU overhead of constructing tensors and to help with debugging: https://github.com/NVIDIA/TransformerEngine/pull/1471#discussion_r1964410144
- The changes to FP8 CS attention look reasonable to me, but I'm a little wary given how complicated it is (see https://github.com/NVIDIA/TransformerEngine/pull/2012).

vthumbe1503 (approved):
FSDP2 related changes LGTM. It seems like we didnt face bitwise parity issues before for DCP load with current scaling, since we were dependent on quantizer for the scales. And now since the quantizer is stateless, small parity mismatch should be fine. We are also going to hit this with other block scaling quantizers which are xfailed currently and whenwe try to fix them.

@pstjohn @cspades @jomitchellnv FYI

PR #2820, opened 2026-04-01
Refactor Amax Kernel ldmatrix loads, TMA/compute barriers,  swizzle_idx

cael-ling:
# Description

To improve the efficiency of Amax kernel for nvFP4 quantization, the following code-refactoring are applied:

1. Transposed path
The transposed path uses ldmatrix_x4_m8n8_shared_b16 instead of row-major loads plus four in-register transposes before the same WMMA operand pattern, cutting instructions and warp-synchronous work on the hot path.

2. TMA + compute pipeline barrier
__syncthreads() barrier is moved to after the full nested (stage_y, stage_x) compute for the current stage pair: one __syncthreads(), then ptx::fence_proxy_async_shared_cta(), so generic shared visibility is established before the next async TMA reuses the buffer. Fewer barriers, lower overhead, same “no readers left before reuse” semantics.

3. swizzle_idx
GroupHadamardAmaxTmaKernel now computes swizzle_idx once per thread before the for (stage_y) loop and passes it into ComputeKernel, avoiding redundant work in the hot nested loops. Behavior unchanged; small micro-optimization and clearer split between loop-invariant mapping and per-tile pointer math.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/common/hadamard_transform/graph_safe_group_hadamard_transform.cu
+++ b/transformer_engine/common/hadamard_transform/graph_safe_group_hadamard_transform.cu
@@ -90,9 +90,8 @@ __device__ __forceinline__ void ComputeKernel(uint32_t b_frag_i[4], uint32_t b_f
   if (kReturnTransposedAmax) {
     if (!kReturnIdentityAmax) {
       ldmatrix_x4_m8n8_shared_b16<true>(a_frag[0], a_frag[1], a_frag[2], a_frag[3],
-                                         reinterpret_cast<uint4*>(in_sh_ptr) + swizzle_idx);
-    }
-    else {
+                                        reinterpret_cast<uint4*>(in_sh_ptr) + swizzle_idx);
+    } else {
       matrix_transpose_m8_n8_b16_inplace(a_frag[0]);
       matrix_transpose_m8_n8_b16_inplace(a_frag[1]);
       matrix_transpose_m8_n8_b16_inplace(a_frag[2]);
@@ -100,11 +99,11 @@ __device__ __forceinline__ void ComputeKernel(uint32_t b_frag_i[4], uint32_t b_f
     }
 
     mma_m16_n16_k16_b16_b16_b16_noacc<kReturnTransposedAmax>(
-      a_frag[0], a_frag[2], a_frag[1], a_frag[3], b_frag_t[0], b_frag_t[1], b_frag_t[2],
-      b_frag_t[3], c_frag[0], c_frag[1], c_frag[2], c_frag[3], temp_amax_t_reg);
-    asm volatile("max.xorsign.abs.bf16x2 %0, %1, %2;\n\t" 
-                : "=r"(local_amax_t_reg)
-                : "r"(local_amax_t_reg), "r"(temp_amax_t_reg));
+        a_frag[0], a_frag[2], a_frag[1], a_frag[3], b_frag_t[0], b_frag_t[1], b_frag_t[2],
+        b_frag_t[3], c_frag[0], c_frag[1], c_frag[2], c_frag[3], temp_amax_t_reg);
+    asm volatile("max.xorsign.abs.bf16x2 %0, %1, %2;\n\t"
+                 : "=r"(local_amax_t_reg)
+                 : "r"(local_amax_t_reg), "r"(temp_amax_t_reg));
   }
 
   if (kReturnPreRhtAmax) {
--- a/transformer_engine/common/hadamard_transform/hadamard_transform.cu
+++ b/transformer_engine/common/hadamard_transform/hadamard_transform.cu
@@ -51,9 +51,8 @@ __device__ __forceinline__ void ComputeKernel(uint32_t b_frag_i[4], uint32_t b_f
   if (kReturnTransposedAmax) {
     if (!kReturnIdentityAmax) {
       ldmatrix_x4_m8n8_shared_b16<true>(a_frag[0], a_frag[1], a_frag[2], a_frag[3],
-                                         reinterpret_cast<uint4*>(in_sh_ptr) + swizzle_idx);
-    }
-    else {
+                                        reinterpret_cast<uint4*>(in_sh_ptr) + swizzle_idx);
+    } else {
       matrix_transpose_m8_n8_b16_inplace(a_frag[0]);
       matrix_transpose_m8_n8_b16_inplace(a_frag[1]);
       matrix_transpose_m8_n8_b16_inplace(a_frag[2]);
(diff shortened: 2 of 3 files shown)


transformer_engine/common/hadamard_transform/graph_safe_group_hadamard_transform.cu
```diff
@@ -133,6 +106,18 @@ __device__ __forceinline__ void ComputeKernel(uint32_t b_frag_i[4], uint32_t b_f
                  : "=r"(local_pre_rht_amax_reg)
                  : "r"(a_frag[0]), "r"(local_pre_rht_amax_reg));
   }
+
+  if (kReturnIdentityAmax) {
+    ldmatrix_x4_m8n8_shared_b16<false>(a_frag[0], a_frag[1], a_frag[2], a_frag[3],
```
zhongbozhu: even kReturnIdentityAmax is not triggered in practice, but technically kReturnIdentityAmax and kReturnPreRhtAmax can share the same ldmatrix_x4_m8n8_shared_b16 instruction?

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -26,46 +26,19 @@ constexpr int kThreadsPerWarp = 32;
...
-  int local_rank = (threadIdx.x % kThreadsPerWarp);
-
-  int ld_row_idx = local_rank % kHadamardDimension;
-  int ld_col_idx = local_rank / kHadamardDimension + warp_id * 2;
-  int swizzle_idx = swizzle_128B_atom_32B(ld_row_idx, ld_col_idx);
-
   uint32_t temp_amax_reg;
   uint32_t temp_amax_t_reg;
 
-  if (kReturnIdentityAmax) {
```
frsun-nvda, line 45: I think the overall goal of this function is to read the data from SMEM once with any combination of (`kReturnIdentityAmax`, `kReturnTransposeAmax`, `kReturnTransposedAmax`). I think your approach change the code such that in the situation where  `kReturnIdentityAmax` and `kReturnTransposeAmax` is `true`, the `ComputeKernel` will need to read the data twice from SMEM. I don't think it will be faster than the existing approach.

transformer_engine/common/hadamard_transform/hadamard_transform.cu
```diff
@@ -26,46 +26,19 @@ constexpr int kThreadsPerWarp = 32;
...
-    if (!kReturnIdentityAmax) {
-      ldmatrix_x4_m8n8_shared_b16<false>(a_frag[0], a_frag[1], a_frag[2], a_frag[3],
-                                         reinterpret_cast<uint4*>(in_sh_ptr) + swizzle_idx);
-    }
-
-    matrix_transpose_m8_n8_b16_inplace(a_frag[0]);
-    matrix_transpose_m8_n8_b16_inplace(a_frag[1]);
-    matrix_transpose_m8_n8_b16_inplace(a_frag[2]);
-    matrix_transpose_m8_n8_b16_inplace(a_frag[3]);
+    ldmatrix_x4_m8n8_shared_b16<true>(a_frag[0], a_frag[1], a_frag[2], a_frag[3],
```
frsun-nvda: If we just want to read once, it might be possible to just 
```
if (kReturnTransposedAmax) {
    if (!kReturnIdentityAmax) {
          ldmatrix_x4_m8n8_shared_b16<true>(a_frag[0], a_frag[1], a_frag[2], a_frag[3], ....);
    } else {
          matrix_transpose_m8_n8_b16_inplace(a_frag[0]);
          matrix_transpose_m8_n8_b16_inplace(a_frag[1]);
          matrix_transpose_m8_n8_b16_inplace(a_frag[2]);
          matrix_transpose_m8_n8_b16_inplace(a_frag[3]);
    }
    ....
```
However, due to reasons I can't remember, the approach was abandoned. The above code could ensure that only 1 SMEM read occurs for `ComputeKernel`.

transformer_engine/common/hadamard_transform/graph_safe_group_hadamard_transform.cu
cael-ling: After a_frag is initialized by ldmatrix in kReturnPreRhtAmax, it is modified by the assembly instruction in its branch:
asm volatile("max.xorsign.abs.bf16x2 %0, %1, %2;\n\t"
                 : "=r"(a_frag[0])
                 : "r"(a_frag[0]), "r"(a_frag[1]));
so a_frag cannot be directly used in the following kReturnIdentityAmax.

transformer_engine/common/hadamard_transform/hadamard_transform.cu
cael-ling: Hi, the PR's code is fast because kReturnIdentityAmax is always false in the current logic. NCU testing shows a 19% performance improvement for the groupamaxkernel and 12% for amax kernel for DeepSeek V3/Moonlight 16B/Ling mini 2.0 model shapes.

cael-ling: Image

Image

That is a reasonable engineering trade: if production only uses I = false, prioritizing the path that removes 4× transpose is a clear win at the same ldmatrix count (1 in both tables); if I is re-enabled (I think it is only a experimental one? ), you accept +1 ldmatrix when T is on (still usually cheaper than 4 transposes?). Document the assumption in a comment or design note so the I + T case is not forgotten.

transformer_engine/common/hadamard_transform/hadamard_transform.cu
cael-ling: Correct, this results in 1× ldmatrix + fewer transpose in comparison to table 1 above.  **Functionally fully equivalent, and also ensuring a single SMEM read occur**; for some combinations, no additional transpose is required.

Image

frsun-nvda: As long as the CI passes with the unit tests, this LGTM. I vaguely remember for correctness reasons this approach was abandoned, but the code LGTM. @zhongbozhu Just FYI.

cael-ling: There are three changes in the PR. The remaining issue is in the ldmatrix logic. I refactored the ldmatrix logic in a new commit. Table 1 below shows the current code logic on the main branch, while Table 3 shows the logic in the new commit. 

image

image

Both tables have ldmatrix = 1. The key difference is that the new change reduces transpose operations. With the current logic where **identity** is false, the new commit results in zero transposes. Even if identity is enabled in the future, performance is guaranteed to remain unchanged.

And this is the latest performance comparisons:

Image

Thanks for the valuable suggestions @frsun-nvda @zhongbozhu

PR #2821, opened 2026-04-01
[Common] Fix fused router for large top-K and expert counts

harryzhou2000:
# Description

Fixed fused router support for large topk and num_expert. Now num_expert <=2304 and any topk is supported with reasonable performance.

Current benchmark shows fused topk forward kernel is faster than pytorch at topk=32, which would be around 8x faster than before optimization.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Fix fused MoE router kernels to support large top-K values and large numbers of experts (1024+) by adding a warp-level radix-selection top-K algorithm (O(E), independent of K) alongside the existing naive O(K^2*E) implementation, dispatched at `topk >= 16` boundary.
- Expand dynamic shared memory allocation via `cudaFuncSetAttribute` in both forward and backward
  kernel launchers to avoid silent failures when expert count exceeds the default 48 KB limit.
- Rewrite `apply_softmax_on_float` to use a numerically stable online max+sum accumulation
  (two-pass → single-pass) with NaN-safe warp reduction, eliminating shared-memory round-trips.

## Details

**Radix top-K selection** (`utils.h`):
Implements a 4-bit radix selection algorithm (8 passes over float32) that finds the K-th largest
value in O(E/32) per warp, independent of K. Phase 1 narrows the bit pattern of the K-th value
via histogram counting; Phase 2 gathers elements into output arrays with deterministic tie-breaking
(value DESC, index ASC) matching `torch.topk` behavior.
**Dispatch logic** (`fused_topk_with_score_function.cu`, `fused_score_for_moe_aux_loss.cu`):
Template parameter `TopkFuncType` (Naive/Radix) is selected at launch time based on
`topk < 16`. Both forward kernels and backward kernels now call `cudaFuncSetAttribute` to
request the required dynamic shared memory size before launch.
**Tests** (`test_fused_router.py`):
- Add `num_experts=1024` to all parametrized test cases.
- Add `_get_tolerances()` helper that scales `atol`/`rtol` with expert count to account for
  O(N * eps) accumulation divergence between fused and reference implementations.

# Checklist:

### DIFF
--- a/tests/pytorch/test_fused_router.py
+++ b/tests/pytorch/test_fused_router.py
@@ -29,12 +29,10 @@ def _get_tolerances(dtype: torch.dtype, num_experts: int):
     """
     # Default tolerances for torch.testing.assert_close
     base_atol, base_rtol = 1e-5, 1.3e-6
-
-    eps = {
-        torch.float32: 2e-7,
-        torch.float16: 1e-3,
-        torch.bfloat16: 4e-3,
-    }.get(dtype, 2e-7)
+    # TODO: account for fp16, bf16 as dtype
+    if dtype != torch.float32:
+        raise NotImplementedError("tolerances implemented for fp32 only")
+    eps = 2e-7
     # The worst-case rounding error from summing N values is O(N * eps).
     # Use 2 * num_experts * eps as the tolerance floor so tests pass for
     # large expert counts while remaining tight for small ones.


tests/pytorch/test_fused_router.py
```diff
@@ -227,13 +261,13 @@ def run_comparison(
...
     # Check the gradient
-    torch.testing.assert_close(logits.grad, logits_clone.grad)
+    torch.testing.assert_close(logits.grad, logits_clone.grad, atol=atol, rtol=rtol)
 
 
 @pytest.mark.parametrize("dtype", [torch.float32])
 @pytest.mark.parametrize("num_tokens", [2048, 7168, 8992])
-@pytest.mark.parametrize("num_experts", [128, 32])
-@pytest.mark.parametrize("topk", [4, 8])
+@pytest.mark.parametrize("num_experts", [1024, 128, 32])
```
tdophung, line 267: self resolve: just make sure this does not make the CI go OOM, otherwise it should be fine

transformer_engine/common/fused_router/fused_score_for_moe_aux_loss.cu
```diff
@@ -149,10 +149,21 @@ void fused_score_for_moe_aux_loss_forward_kernel_launcher(
   size_t shared_memory_size = num_experts * num_token_per_block * sizeof(CompType)  // logits
                               + topk * num_token_per_block * sizeof(CompType)       // topk_logits
                               + topk * num_token_per_block * sizeof(int);           // topk_indices
-  fused_score_for_moe_aux_loss_forward_kernel<DataType>
-      <<<grid_size, kThreadsPerBlock, shared_memory_size, stream>>>(
-          logits, num_tokens, num_experts, topk, score_function, scores, routing_map,
-          intermediate_output);
+  if (topk < 16) {
+    cudaFuncSetAttribute(fused_score_for_moe_aux_loss_forward_kernel<DataType, TopkFuncType::Naive>,
```
tdophung: use NVTE_CHECK_CUDA here

transformer_engine/common/fused_router/fused_topk_with_score_function.cu
```diff
@@ -254,10 +253,23 @@ void fused_topk_with_score_function_forward_kernel_launcher(
     shared_memory_size += num_groups * num_token_per_block * sizeof(CompType);   // group_scores
     shared_memory_size += num_experts * num_token_per_block * sizeof(CompType);  // maksed_scores
   }
-  fused_topk_with_score_function_forward_kernel<DataType, BiasType>
-      <<<grid_size, kThreadsPerBlock, shared_memory_size, stream>>>(
-          logits, num_tokens, num_experts, topk, use_pre_softmax, num_groups, group_topk,
-          scaling_factor, score_function, expert_bias, probs, routing_map, intermediate_output);
+  if (topk < 16) {
```
tdophung, line 259: maybe add a comment here explaining the break even point point would be helpful:
like
`// Radix selection is O(E), independent of K; switch at K=16 where naive O(K^2*E) starts to dominate`

transformer_engine/common/fused_router/fused_score_for_moe_aux_loss.cu
tdophung: I also wonder if we should have an error message saying that this might be due to the num experts exceeding 2304 (based on your PR description) so can't get enough shared mem. (and print out how large num_experts can be depending on the hardware's size of shared mem in the case it errors out would be helpful)

transformer_engine/common/fused_router/utils.h
```diff
@@ -187,22 +187,250 @@ __device__ inline void apply_softmax_bwd_on_float(float *grad, float *fwd_output
...
+  //   merged_sum = sum_a * exp(max_a - merged_max) + sum_b * exp(max_b - merged_max)
+  //
+  // NaN guard: when data_size < 32, some lanes have (max=-inf, sum=0).
+  // Merging two such lanes computes expf(-inf - (-inf)) = expf(NaN) = NaN,
+  // and 0.0 * NaN = NaN in IEEE 754, contaminating valid lanes.
+  // Fix: treat -inf max as "no data" and skip the expf computation.
+#pragma unroll
+  for (int offset = 16; offset > 0; offset >>= 1) {
+    float other_max = __shfl_xor_sync(0xffffffff, local_max, offset);
+    float other_sum = __shfl_xor_sync(0xffffffff, local_sum, offset);
```
tdophung: why not using warp_shuffle_xor(const T &x, uint32_t idx) from utils.cuh?

transformer_engine/common/fused_router/utils.h
```diff
@@ -187,22 +187,250 @@ __device__ inline void apply_softmax_bwd_on_float(float *grad, float *fwd_output
...
+// Convert float to an unsigned integer that preserves descending sort order.
+// After conversion, a numerically larger float maps to a larger uint32.
+// This allows bitwise radix selection to find top-K by searching from MSB.
+__device__ inline unsigned int float_to_ordered_uint(float f) {
+  unsigned int u = __float_as_uint(f);
+  // If sign bit is set (negative), flip all bits.
+  // If sign bit is clear (positive or +0), flip only the sign bit.
+  unsigned int mask = (u & 0x80000000u) ? 0xFFFFFFFFu : 0x80000000u;
+  return u ^ mask;
+}
```
tdophung: could probably be in utils.cuh

transformer_engine/common/fused_router/utils.h
```diff
@@ -187,22 +187,250 @@ __device__ inline void apply_softmax_bwd_on_float(float *grad, float *fwd_output
...
+    unsigned int total_counts[RADIX_SIZE];
+#pragma unroll
+    for (int b = 0; b < RADIX_SIZE; b++) {
+      unsigned int c = counts[b];
+      // Butterfly reduction
+      c += __shfl_xor_sync(0xffffffff, c, 16);
+      c += __shfl_xor_sync(0xffffffff, c, 8);
+      c += __shfl_xor_sync(0xffffffff, c, 4);
+      c += __shfl_xor_sync(0xffffffff, c, 2);
+      c += __shfl_xor_sync(0xffffffff, c, 1);
```
tdophung: could again use the warp_reduce utils in utils.cuh

harryzhou2000: utils.cuh does not seem to contain such a allreduce sum function, so I am adding a T warp_allreduce_sum(T x) in utils.cuh

tests/pytorch/test_fused_router.py
harryzhou2000: I ran this test locally and it seemed ok.

transformer_engine/common/fused_router/fused_score_for_moe_aux_loss.cu
harryzhou2000: Adding checks for shared_memory_size.

tdophung:
retriggered CI and all passed: https://gitlab-master.nvidia.com/dl/transformerengine/transformerengine/-/pipelines/48710028

PR #2829, opened 2026-04-03
[PyTorch][CP] Add THD format support for AllGather-based Context Parallelism

sudhakarsingh27:
# Description

Add THD (variable-length sequence) format support to `AttnFuncWithCPAndKVAllGather`. Previously, AllGather-based CP only supported fixed-length formats (`bshd`/`sbhd`). THD format packs variable-length sequences into a single `[t, h, d]` tensor tracked by `cu_seqlens`, which is needed for workloads with heterogeneous sequence lengths.

The key challenge is that AllGather CP splits Q across 2 steps (one per local chunk), but THD tensors cannot be naively sliced like fixed-length formats. This PR uses an **offset-based approach**: the full Q tensor is passed to the attention kernel each step, with per-step `cu_seqlens_q_padded` values directing the kernel to read the correct chunk. This avoids tensor slicing and follows the padded THD convention used by the backends.

## Type of change

## Changes

- **Offset-based Q chunking**: Per-step `cu_seqlens_q_padded` selects which chunk the kernel reads from the full Q tensor, instead of slicing Q per step.
- **Per-step KV sequence metadata**: Computes visible KV token counts, per-step `cu_seqlens_kv`, and window ranges for causal, full/no-window, and SWA cases.
- **THD AllGather helper kernels**: Adds sync-free THD reorder / valid-copy helper kernels for the AllGather CP THD path.
- **FlashAttention v3 AllGather THD support**: Enables FlashAttention v3 AG+THD coverage, including padded THD cases through `pad_between_seqs` and `seqused_k` handling.
- **max_logit masking fix**: Handles non-zero-starting `cu_seqlens_q_padded` in the valid-token mask without `.item()` D2H synchronizations.
- **Tests**: Adds/extends CP THD coverage for FusedAttention, FlashAttention v3, AllGather, SWA/full/causal masks, padded Flash THD, and helper-kernel unit coverage in `test_cp_utils.py`.

# Checklist:

## Latest validation update (2026-06-11)

CP THD AllGather coverage was validated across FusedAttention and FlashAttention v3.

| Backend | Coverage | Pad | Result | CI/local test |
|---|---|---|---|---|
| Flash v3 | AG+THD causal | `False/True` | PASS | CI 54405200 H100 L3 FA3 CP; local focused H100 run |
| Flash v3 | AG+THD full/no-window | `False/True` | PASS | Local focused H100 run |
| Flash v3 | AG+THD causal SWA `(128,0)` | `False/True` | PASS | CI 54405200 H100 L3 FA3 CP; local focused H100 run |
| Fused | AG+THD causal | n/a | PASS | CI 54405200 H100 L1 CP; local focused H100 run |
| Fused | AG+THD full/no-window | n/a | PASS | Local focused H100 run |
| Fused | AG+THD causal SWA `(128,0)` | n/a | PASS | CI 54405200 H100 L1 CP; local focused H100 run |
| Fused | AG+THD causal SWA `(512,512)` | n/a | PASS | Local focused H100 run |

`test_cp_utils.py` also passed 14/14 on both H100 and B200 L1 in CI pipeline 54405200. It covers the THD helper kernels, including reorder/copy-valid tests against the legacy Python reference paths.

FusedAttention does not expose the FlashAttention-specific `pad_between_seqs` axis; Fused THD padding semantics are covered through the THD/padding mask path.

Helper-kernel microbenchmarks, single H100, bf16, `batch=16`, `seqlen=4096`, `heads=16`, `dim=64`:

| Kernel path | cp sizes | Legacy wall ms | New kernel wall ms | Speedup |
|---|---|---:|---:|---:|
| `thd_cp_reorder_sequences` contiguous->rank | 2, 4, 8 | 11.6829-37.4912 | 0.0954-0.1090 | 122.42x-346.08x |
| `thd_cp_reorder_sequences` rank->contiguous | 2, 4, 8 | 11.0553-19.2267 | 0.0966-0.1094 | 114.43x-177.87x |
| `thd_cp_copy_valid_tokens` | 2, 4, 8 | 0.7070-9.4218 | 0.0911-0.1052 | 7.76x-89.59x |

Note: the helper-kernel numbers are fixed-shape microbenchmarks, not a sweep over sequence length or number of sequences.


transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2891,24 +2900,36 @@ def forward(
...
-        chunk_ids_for_kv_ag = get_seq_chunk_ids_for_reordering_before_attn(cp_size, k.device)
-        k_ag = torch.index_select(k_ag, dim=0, index=chunk_ids_for_kv_ag)
-        v_ag = torch.index_select(v_ag, dim=0, index=chunk_ids_for_kv_ag)
-        # [cp*2, s//2, b, h, d] -> [cp*s, b, h, d]
-        k_ag = k_ag.view(-1, *k.shape[1:])
-        v_ag = v_ag.view(-1, *v.shape[1:])
+        if qkv_format == "thd":
+            # [cp*t, h, d] -> reorder to contiguous per-sequence order -> [t_full, h, d]
+            chunk_ids_for_kv_ag = get_seq_chunk_ids_for_reordering_before_attn(cp_size, k.device)
+            k_ag = reorder_seq_chunks_after_a2a_before_attn_thd(
```
sudhakarsingh27: This reorder_seq_chunks_after_a2a_before_attn_thd and the other releated method are not "a2a" specific now, rename them to something like dualchunk_to_contiguous_order_thd and the other one contiguous_to_dualchunk_order_thd

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2834,9 +2836,14 @@ def forward(
...
-        assert not padding, f"{attn_mask_type} mask type is not supported!"
+        if qkv_format == "thd":
+            # THD always uses padding mask types; per-step masks set internally
+            assert padding, f"THD format requires padding mask type, got {attn_mask_type}!"
+        else:
+            assert not padding, f"{attn_mask_type} mask type is not supported!"
         if use_fused_attention and causal and "bottom_right" not in attn_mask_type:
-            attn_mask_type = attn_mask_type + "_bottom_right"
+            if qkv_format != "thd":
+                attn_mask_type = attn_mask_type + "_bottom_right"
```
sudhakarsingh27: Why only for non THD cases, we're creating `*_bottom_right` masks? Shouldn't THD cases also have it? 
Edit: for THD, there are per step masks but those masks are same for each step i.e. "padding_causal_bottom_right" for both steps if causal otherwise "padding" both steps for THD, so maybe there's no need to distinguish these two paths separately.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2921,39 +2946,130 @@ def forward(
...
+                    # Causal: visible KV covers chunks 0..chunk_id
+                    chunk_id = local_seq_chunk_ids[step_idx]
+                    visible_padded = padded_chunk_sizes * (chunk_id + 1)
+                    visible_actual = torch.minimum(actual_seqlens, visible_padded)
+                    cs = torch.zeros_like(cu_seqlens_q_original)
+                    cs[1:] = visible_actual.cumsum(0)
+                    thd_cu_seqlens_kv_per_step[step_idx] = cs
+                else:
+                    # Non-causal: all KV tokens visible
+                    thd_cu_seqlens_kv_per_step[step_idx] = cu_seqlens_q_original.clone()
```
sudhakarsingh27: should this be created from `cu_seqlens_kv` instead?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2921,39 +2946,130 @@ def forward(
...
+            thd_cu_seqlens_q_padded_per_step[1][:-1] += padded_chunk_sizes
+
+            # Per-step KV cu_seqlens (non-padded): how many actual KV tokens are
+            # visible for each sequence.
+            thd_cu_seqlens_kv_per_step = [None, None]
+            for step_idx in range(2):
+                if causal:
+                    # Causal: visible KV covers chunks 0..chunk_id
+                    chunk_id = local_seq_chunk_ids[step_idx]
+                    visible_padded = padded_chunk_sizes * (chunk_id + 1)
```
sudhakarsingh27: this `padded_chunk_sizes` is created from `cu_seqlens_q_padded`, it should be created from `cu_seqlens_kv_padded` which are same if cu_seqlens_q == cu_seqlens_kv but not always

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2921,39 +2946,130 @@ def forward(
...
+
+            if causal:
+                # Q is always the last chunk in the visible KV range,
+                # so bottom_right alignment is always correct.
+                thd_attn_mask_type_per_step = [
+                    "padding_causal_bottom_right",
+                    "padding_causal_bottom_right",
+                ]
+            else:
+                thd_attn_mask_type_per_step = ["padding", "padding"]
```
sudhakarsingh27: since this might be addressed before, could be removed

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: fixed now

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -100,7 +100,10 @@ def test_cp_with_flash_attention(dtype, model, qkv_format, cp_comm_type):
         pytest.skip("CP implementation with KV all-gather does not support bias yet!")
     if qkv_format == "thd":
         if cp_comm_type == "all_gather":
-            pytest.skip("CP implementation with KV all-gather does not support THD format yet!")
+            pytest.skip(
+                "FlashAttention does not support THD padding; use FusedAttention for"
+                " THD+all_gather CP."
```
cyanguwa, line 105: Maybe swap the words a little bit so it doesn't sounds like FlashAttention doesn't support THD, but just our CP implementation with it doesn't? (Also, THD implies padding in our terminology?)

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3025,6 +3157,18 @@ def forward(
...
+                        # Copy valid token ranges from this step's output.
+                        # Each step writes at different positions (no overlap, no correction needed).
+                        step_padded = thd_cu_seqlens_q_padded_per_step[i - 1]
+                        step_valid = thd_cu_seqlens_q_per_step[i - 1]
+                        batch_size = step_valid.shape[0] - 1
+                        for b in range(batch_size):
+                            s = step_padded[b].item()
+                            sz = (step_valid[b + 1] - step_valid[b]).item()
+                            if sz > 0:
+                                out[s : s + sz].copy_(out_per_step[i - 1][s : s + sz])
```
cyanguwa, line 3170: This "for" loop might be too costly. Could this logic be written in another way in Python, or simply in C++/CUDA? Do we have some thd kernels that already do this?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3236,44 +3442,72 @@ def backward(ctx, dout, *_args):
...
+                    if ctx.qkv_format == "thd":
+                        # dQ: copy valid token ranges from this step's dQ
+                        step_padded = thd_cu_seqlens_q_padded_per_step[i - 1]
+                        step_valid = thd_cu_seqlens_q_per_step[i - 1]
+                        batch_size = step_valid.shape[0] - 1
+                        for b in range(batch_size):
+                            s = step_padded[b].item()
+                            sz = (step_valid[b + 1] - step_valid[b]).item()
+                            if sz > 0:
+                                dq[s : s + sz].copy_(dq_per_step[i - 1][s : s + sz])
```
cyanguwa, line 3454: Same here, regarding the "for" loop.

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -267,8 +270,6 @@ def test_cp_with_fused_attention(
     if qkv_format == "thd" and config.attn_bias_type == "post_scale_bias":
         pytest.skip("THD format does not support post_scale_bias yet!")
     if qkv_format == "thd":
-        if cp_comm_type == "all_gather":
-            pytest.skip("CP implementation with KV all-gather does not support THD format yet!")
```
cyanguwa, line 271: A general comment - please run the CP file with "test_essential=False" offline because the essential tests may not cover everything.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3095,34 +3170,49 @@ def forward(
...
+            # [cp*t, h, d] -> reorder to contiguous per-sequence order -> [t_full, h, d]
+            # Use padded cu_seqlens since reorder computes slice boundaries via integer
+            # division by 2*cp_size, which requires divisible values.
+            chunk_ids_for_kv_ag = get_seq_chunk_ids_for_reordering_before_attn(cp_size, k.device)
+            k_ag = reorder_thd_sequences_to_contiguous(
+                k_ag, cu_seqlens_kv_padded, chunk_ids_for_kv_ag, cp_size
+            )
+            v_ag = reorder_thd_sequences_to_contiguous(
+                v_ag, cu_seqlens_kv_padded, chunk_ids_for_kv_ag, cp_size
+            )
```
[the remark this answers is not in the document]
sudhakarsingh27: Incorporated. Check again

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3082,6 +3250,13 @@ def forward(
             f" {q.shape[seq_dim_qkv]}, seq_len_kv = {k.shape[seq_dim_qkv]}."
         )
 
+        if qkv_format == "thd" and not use_fused_attention and not use_flash_attn_3:
+            assert False, (
```
sudhakarsingh27: the condition should be inside the assert

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3113,11 +3288,24 @@ def forward(
...
+
+        if qkv_format == "thd":
+            # Save original global cu_seqlens before division
+            cu_seqlens_q_original = cu_seqlens_q.clone()
+            cu_seqlens_kv_original = cu_seqlens_kv.clone()
+        else:
+            seq_dim = qkv_format.index("s")
+            assert (
+                q.shape[seq_dim] % 2 == 0 and k.shape[seq_dim] % 2 == 0
+            ), "Sequence length per GPU needs to be divisible by 2!"
```
sudhakarsingh27, line 3105: The divisibility check should include 2*cp_size but it's being checked elsewhere so we shouldn't add it here redundantly

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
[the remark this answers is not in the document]
sudhakarsingh27: Addressed on the current branch. The THD path now has the same dependency as the non-THD path after K/V preparation:

```python
cp_stream.wait_stream(torch.cuda.current_stream())
```

This sits after the THD `reorder_thd_sequences_to_contiguous` calls, so step 1 on `cp_stream` cannot read `k_ag` / `v_ag` before the all-gather + reorder writes complete. The separate FA3 internal-workspace race is handled by `628f73cc`, which serializes only consecutive FA3 per-step calls with GPU stream waits.

Validation: FusedAttention all_gather THD pytest passed (`4 passed, 476 skipped, 9824 deselected`), and FA3 bucket32k all_gather THD correctness runner exited 0.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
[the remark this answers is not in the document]
sudhakarsingh27: Addressed on the current branch. The `max_logit` merge now waits before the default-stream `torch.maximum` reads a per-step result produced on `cp_stream`:

```python
torch.cuda.current_stream().wait_stream(flash_attn_streams[i - 1])
max_logit = torch.maximum(max_logit, max_logit_per_step[i - 1])
```

That wait is intentionally before the merge; the later post-loop `wait_stream(cp_stream)` was too late for this read.

Validation: FusedAttention all_gather THD pytest passed (`4 passed, 476 skipped, 9824 deselected`).

transformer_engine/pytorch/cpp_extensions/fused_attn.py
```diff
@@ -366,23 +365,18 @@ def fused_attn_fwd(
...
-                    # Expand: T×3, F×1, T×3, F×1, T×2, F×2, T×7, F×1 → TTTF|TTTF|TTFF|TTTTTTTF
-                    valid = torch.repeat_interleave(values, counts)
-                    # Finally, replace invalid (F) positions with -inf
+                    tq = max_tensor.shape[0]
+                    valid = torch.zeros(tq, dtype=torch.bool, device=max_tensor.device)
+                    b = actual_seqlens.shape[0]
+                    for b_idx in range(b):
+                        start = cu_seqlens_q_padded[b_idx].item()
+                        n_valid = actual_seqlens[b_idx].item()
+                        valid[start : start + n_valid] = True
```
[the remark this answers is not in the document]
sudhakarsingh27: Fixed in `b897900d`. The THD max-logit path no longer builds the valid-token mask with per-batch `.item()` calls. It now builds the same interval mask on GPU using start/end deltas and `scatter_add_`, preserving non-zero `cu_seqlens_q_padded` offsets without D2H synchronization.

Validation: `python3 -m black --check transformer_engine/pytorch/cpp_extensions/fused_attn.py` passed, and FusedAttention all_gather THD pytest passed (`4 passed, 476 skipped, 9824 deselected`).

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Resolved on the current branch, with final cleanup in `0e926c42`. The THD reorder entry points are now `reorder_thd_sequences_to_rank_sharded` and `reorder_thd_sequences_to_contiguous`, and the stale Python permutation helpers were removed. Both wrappers call the fused `tex.thd_reorder` path, so this logic is no longer A2A-named or A2A-specific.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3236,44 +3423,72 @@ def backward(ctx, dout, *_args):
...
+                        for b in range(batch_size):
+                            s = step_padded[b].item()
+                            sz = (step_valid[b + 1] - step_valid[b]).item()
+                            if sz > 0:
+                                dq[s : s + sz].copy_(dq_per_step[i - 1][s : s + sz])
+                        # dK/dV: add full tensor (kernel zeros non-valid positions)
+                        if i > 1:
+                            flash_attn_streams[i - 1].wait_event(dkv_update_done)
+                        dk.add_(dk_per_step[i - 1])
+                        dv.add_(dv_per_step[i - 1])
```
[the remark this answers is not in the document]
sudhakarsingh27: Updated in `0e926c42`. The code no longer claims or depends on cuDNN zeroing non-valid positions. The dK/dV accumulation is treated row-wise: padding rows may accumulate, but valid rows are independent and only valid rows are copied/scattered after the per-step reductions. Revalidated today with the focused FusedAttention THD all_gather pytest (`4 passed, 476 skipped, 9824 deselected`) and the FA3 THD all_gather bucket32k correctness runner.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2921,39 +2946,131 @@ def forward(
...
+                # Causal: visible KV covers chunks 0..chunk_id
+                # chunk_id = local_seq_chunk_ids[step_idx]
+                visible_padded = [
+                    padded_chunk_sizes_kv * (chunk_id + 1) for chunk_id in local_seq_chunk_ids
+                ]
+                # [AG+THD] Is this needed?
+                visible_actual = [
+                    torch.minimum(actual_seqlens_kv, visible_padded_split)
+                    for visible_padded_split in visible_padded
+                ]
```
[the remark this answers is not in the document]
sudhakarsingh27: Resolved on the current branch. The open-question development comment is gone; the clamp remains because padded chunk endpoints can exceed actual sequence lengths, and the current comments distinguish padded offsets from valid/unpadded KV visibility.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3105,19 +3246,36 @@ def backward(ctx, dout, *_args):
...
-        chunk_ids_for_kv_ag = get_seq_chunk_ids_for_reordering_before_attn(cp_size, k.device)
-        k_ag = torch.index_select(k_ag, dim=0, index=chunk_ids_for_kv_ag)
-        v_ag = torch.index_select(v_ag, dim=0, index=chunk_ids_for_kv_ag)
-        # [cp*2, s//2, b, h, d] -> [cp*s, b, h, d]
-        k_ag = k_ag.view(-1, *k.shape[1:])
-        v_ag = v_ag.view(-1, *v.shape[1:])
+        if ctx.qkv_format == "thd":
+            cu_seqlens_kv_padded = ctx.cu_seqlens_kv_padded
+            thd_cu_seqlens_q_per_step = ctx.thd_cu_seqlens_q_per_step
+            cu_seqlens_q_padded_rank = cu_seqlens_q_padded * 2
```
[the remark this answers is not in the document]
sudhakarsingh27: Resolved in `0e926c42`; the unused backward `cu_seqlens_q_padded_rank` assignment was removed.

tests/pytorch/attention/test_attention_with_cp.py
sudhakarsingh27: Resolved in the current test wording. The skip message now says the CP implementation with QKVO A2A+P2P / hierarchical A2A does not support THD format, rather than implying FlashAttention itself does not support THD.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Resolved by moving the THD reorder path to the fused kernel. The current helpers call `tex.thd_reorder` directly, and `0e926c42` removed the remaining dead Python permutation builders.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Resolved by the same fused reorder path as the forward case. The current THD contiguous/rank-sharded reorder helpers call `tex.thd_reorder`; `0e926c42` removed the stale Python permutation builders.

tests/pytorch/attention/test_attention_with_cp.py
sudhakarsingh27: Status: not claiming this broad offline request as fully closed yet. Current validation covers the focused FusedAttention THD all_gather pytest, FA3 THD all_gather bucket32k correctness, and prior cp2/cp4/cp8 AG-vs-a2a sweeps. Plan is to run the full CP file with `test_essential=False` offline after the lint/review blockers are clear and report that result back here.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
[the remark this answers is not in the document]
sudhakarsingh27: Still resolved on the current branch. The earlier fix adds the needed stream waits before default-stream max-logit merging, and today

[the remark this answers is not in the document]
sudhakarsingh27: Complete validation note for the previous reply: after `0e926c42`, the focused FusedAttention THD all_gather pytest still passes (`4 passed, 476 skipped, 9824 deselected`) and the FA3 THD all_gather bucket32k correctness runner exits 0.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -272,10 +273,36 @@ def reorder_seq_chunks_for_a2a_after_attn(x, chunk_ids_for_a2a, seq_dim, cp_size
...
+        P = torch.cat(
+            [
+                tex.thd_get_partitioned_indices(cu_seqlens, total_tokens, cp_size, rank)
+                for rank in range(cp_size)
+            ]
+        )
+        P_inv = torch.empty_like(P)
+        P_inv[P.long()] = torch.arange(total_tokens, dtype=P.dtype, device=P.device)
+        _thd_reorder_perm_cache[key] = (P, P_inv)
+    return _thd_reorder_perm_cache[key]
```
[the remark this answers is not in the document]
sudhakarsingh27: Resolved in `0e926c42`. The stale THD reorder permutation/cache helpers are gone; the reorder wrappers now call fused `tex.thd_reorder` directly with device-resident `cu_seqlens`, so there is no Python CUDA-device cache key to maintain for this path.

transformer_engine/pytorch/cpp_extensions/fused_attn.py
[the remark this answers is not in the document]
sudhakarsingh27: Still resolved by `b897900d` and revalidated today after the review-fix commit. The THD max-logit mask is built on GPU with start/end deltas and `scatter_add_`, without the previous per-batch `.item()` loop. Focused FusedAttention THD all_gather pytest passed again after `0e926c42`.

transformer_engine/common/fused_attn/context_parallel.cu
```diff
@@ -96,12 +108,82 @@ __global__ void thd_partition_indices_kernel(int *output, int *cu_seqlens, int b
...
+
+  for (int token_id = warpid; token_id < total_tokens; token_id += num_warps) {
+    int seq_id = binary_search(token_id, padded_s, batch + 1);
+    int local = token_id - padded_s[seq_id];
+    int valid_len = valid_s[seq_id + 1] - valid_s[seq_id];
+    // local can be negative when a segment's padded start is shifted past earlier tokens (e.g.
+    // step-1 chunks: cu_seqlens_padded[:-1] += chunk_size). Those tokens are outside any valid
+    // run, so skip them -- otherwise the first chunk's already-written rows get clobbered.
+    if (local >= 0 && local < valid_len) {
+      float4 *src_tok = reinterpret_cast<float4 *>(
```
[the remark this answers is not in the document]
sudhakarsingh27: Refuted by the standalone oracle and current validation. The tested helper returns the containing padded segment for token offsets, and the kernel guard correctly copies zero rows for `valid_len == 0`. The oracle covers empty/zero-valid and `valid == padded` edge cases and is bit-identical to the old Python copy loop; the focused THD all_gather pytest also passed after rebuild. No code change needed for this one.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Fixed in `0e926c42`; the condition is now the assert predicate with the explanatory message, instead of `if ...: assert False`.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Agreed. I did not add a redundant local divisibility assert here; this path continues to rely on the existing padded divisibility setup/checks for `2 * cp_size`, while this assert stays focused on backend capability for THD all_gather.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3061,9 +3084,25 @@ def forward(
...
+        # Upgrade causal -> causal_bottom_right (and padding_causal ->
+        # padding_causal_bottom_right) for AG CP. This applies to all qkv_format
+        # values including "thd": after AG the per-step Q chunk is shorter than
+        # the visible KV slice, so Q[0] must align with KV[kv_len - q_len] rather
+        # than KV[0]. The bottom_right variant encodes that alignment for both
+        # bshd/sbhd (slice-based) and THD (cu_seqlens/seqused_k-based) paths.
+        if use_fused_attention and causal and "bottom_right" not in attn_mask_type:
+            attn_mask_type = attn_mask_type + "_bottom_right"
+        assert attn_bias_type == "no_bias", f"{attn_bias_type} bias type is not supported!"
+        assert q.shape[-1] % 8 == 0, "Hidden size per attention head should be multiple of 8!"
```
sudhakarsingh27: Why did we add these two additional checks?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3061,9 +3084,25 @@ def forward(
...
         assert (
-            "padding" not in attn_mask_type
+            qkv_format == "thd" or "padding" not in attn_mask_type
         ), f"No support for cp_comm_type='all_gather' and {attn_mask_type=}."
```
sudhakarsingh27, line 3046: this is a repeated check, from just a few lines above

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3214,55 +3286,190 @@ def forward(
         out_per_step = [None, None]
         softmax_lse_per_step = [None, None]
         rng_states = [None, None]
+        if qkv_format == "thd":
+            out = torch.zeros(o_shape, dtype=fwd_nominal_dtype, device=q.device)
+        else:
+            out = torch.empty_like(q)
```
sudhakarsingh27: is it better to just have torch.zeros for both paths?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3346,6 +3570,18 @@ def forward(
                         out_f16[:, i - 1].copy_(out_per_step[i - 1])
                     elif o_format == "sbhd":
                         out_f16[i - 1].copy_(out_per_step[i - 1])
+                    elif qkv_format == "thd":
+                        # Copy valid token ranges from this step's output.
+                        # Each step writes at different positions (no overlap, no correction needed).
+                        # Sync-free fused copy of every segment's valid rows in one launch (replaces
+                        # a per-batch .item() slice-copy loop that stalled comm/compute overlap).
+                        tex.thd_valid_copy(
```
sudhakarsingh27: Information from local experiments leaking into the comments. The comments should only tell what the code does and only in most important/pertinent cases share extra information. In this case, if the user hasn't seen the python version, they wouldn't know how to make sense of per-batch .item() or comm/compute overlap

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3214,55 +3286,190 @@ def forward(
...
+
         for i in range(len(local_seq_chunk_ids) + 1):
             if i < len(local_seq_chunk_ids):
+                # FA3 (hopper) forward allocates an internal persistent-scheduler workspace
+                # (tile_count_semaphore) per call and frees it on return; the caching allocator
+                # records that free only against the stream it was allocated on. With per-step
+                # calls overlapping across streams, the allocator can hand step i's FA3 the block
+                # step i-1's FA3 just freed while step i-1's scheduler kernel is still atomically
+                # incrementing it -> shared semaphore -> out-of-range tiles -> illegal memory access
+                # (FA3-only, only under overlap; root cause in mb_runs/FA3_AG_RECORDSTREAM_RACE.md
```
sudhakarsingh27: Again, leaking experiments data in the comments

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3166,34 +3224,48 @@ def forward(
...
-        chunk_ids_for_kv_ag = get_seq_chunk_ids_for_reordering_before_attn(cp_size, k.device)
-        k_ag = torch.index_select(k_ag, dim=0, index=chunk_ids_for_kv_ag)
-        v_ag = torch.index_select(v_ag, dim=0, index=chunk_ids_for_kv_ag)
-        # reshape/flatten: [cp*2, s//2, b, h, d] -> [cp*s, b, h, d]
-        k_ag = k_ag.view(-1, *k.shape[1:])
-        v_ag = v_ag.view(-1, *v.shape[1:])
+
+        if qkv_format == "thd":
+            # [cp*t, h, d] -> reorder to contiguous per-sequence order -> [t_full, h, d]
+            # Use padded cu_seqlens since reorder computes slice boundaries via integer
```
sudhakarsingh27: also note here that the padded cu_seqlens is global

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Addressed in the pending local patch: removed the two extra generic checks here. The all_gather path now keeps only the CP-specific support checks instead of adding the duplicate no-bias assert and head-dim `% 8` assert at this point.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Addressed in the pending local patch. The repeated check was removed; the only remaining `attn_bias_type == "no_bias"` assert in this block is the CP all_gather support check.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Addressed in the pending local patch. `out` is now allocated with `torch.zeros` for both THD and non-THD paths; the only branch left is choosing `o_shape` for THD versus `q.shape` for the packed non-THD layout.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Addressed in the pending local patch. The comment now only states the current behavior: copy each segment's valid token range from the step output, with each step writing distinct positions.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Addressed in the pending local patch. The FA3 stream comment was trimmed to describe only the required stream serialization and workspace-lifetime reason, without references to local experiment artifacts.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Addressed in the pending local patch. The reorder comment now explicitly says the padded cu_seqlens are global sequence offsets, and that reorder derives per-sequence chunk boundaries from those offsets.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3214,55 +3286,190 @@ def forward(
...
+            # (unpadded). This deliberately does NOT encode CP padding
+            # boundaries — those are conveyed to the kernel via
+            # ``cu_seqlens_kv_padded_`` (passed separately, see below), which
+            # describes the tensor layout. The kernel uses cu_seqlens_kv for
+            # visibility and cu_seqlens_kv_padded for offsets; conflating the
+            # two would incorrectly mask actual tokens.
+            actual_seqlens_kv = cu_seqlens_kv_original[1:] - cu_seqlens_kv_original[:-1]
+            padded_chunk_sizes_kv = (cu_seqlens_kv_padded[1:] - cu_seqlens_kv_padded[:-1]) // (
+                2 * cp_size
+            )
```
sudhakarsingh27: Addressed in the pending local patch. `visible_padded` and `visible_actual` are now computed inside the `if causal or sliding_window_attn:` block where they are used.

tests/pytorch/attention/test_attention_with_cp.py
sudhakarsingh27: Follow-up: this offline validation is now complete on the current worktree with `NVTE_TEST_ESSENTIAL=0`. Full `tests/pytorch/attention/test_attention_with_cp.py` passed: `426 passed, 50838 skipped, 3 warnings` in `18:30`. I also re-ran the focused THD all_gather slice: `50 passed, 4222 skipped, 46992 deselected`.

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -538,8 +555,12 @@ def test_cp_with_fused_attention(
     if config.attn_bias_type != "no_bias" and cp_comm_type in ["all_gather", "a2a", "a2a+p2p"]:
         pytest.skip("No support for bias with cp_comm_type={all_gather, a2a, a2a+p2p}!")
 
-    if qkv_format == "thd" and cp_comm_type in ["all_gather", "a2a+p2p"]:
-        pytest.skip("No support for THD format with cp_comm_type={all_gather, a2a+p2p}!")
+    if qkv_format == "thd":
+        if cp_comm_type == "a2a+p2p":
```
cyanguwa: These two "if"s can be combined?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3214,55 +3278,180 @@ def forward(
         out_per_step = [None, None]
         softmax_lse_per_step = [None, None]
         rng_states = [None, None]
+        out_shape = o_shape if qkv_format == "thd" else q.shape
```
cyanguwa: Shouldn't out_shape be always equal to o_shape? d_v might be different from d_q, right?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3641,30 +3882,73 @@ def backward(ctx, dout, *_args):
...
+                            aux_ctx_tensors = [
+                                softmax_lse_per_step[i],
+                                softmax_lse_per_step[i],
+                                rng_states[i],
+                            ]
+                        else:
+                            aux_ctx_tensors = [
+                                softmax_lse_per_step[i],
+                                rng_states[i],
+                            ]
```
cyanguwa: The t3hd-related logic can be removed now, right? Because of #2959.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3750,20 +4034,44 @@ def backward(ctx, dout, *_args):
                                 for x in [dq_per_step[i], dk_per_step[i], dv_per_step[i]]
                             ]
                     else:
-                        dq_per_step[i], dk_per_step[i], dv_per_step[i] = [
-                            torch.empty_like(x) for x in [q_part, k_part, v_part]
-                        ]
+                        if ctx.use_flash_attn_3 and ctx.qkv_format == "thd":
+                            dq_per_step[i], dk_per_step[i], dv_per_step[i] = [
+                                torch.zeros_like(x) for x in [q_part, k_part, v_part]
+                            ]
```
cyanguwa, line 3956: Do we have this problem in fwd? i.e. for FA3 + THD, do we need to zero out `out_per_step[i]`?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3790,55 +4098,83 @@ def backward(ctx, dout, *_args):
...
-                        kv_seq_range_per_step[i - 1][1],
-                    )
-                    # add to dk/dv: [cp*s, b, h, d]
-                    dk[seq_start_idx:seq_end_idx].add_(dk_per_step[i - 1])
-                    dv[seq_start_idx:seq_end_idx].add_(dv_per_step[i - 1])
-                    if i < len(local_seq_chunk_ids):
-                        flash_attn_streams[i - 1].record_event(dkv_update_done)
+                    if ctx.qkv_format == "thd":
+                        # dQ: copy every segment's valid token range from this step's dQ.
+                        tex.thd_valid_copy(
```
cyanguwa: Nit: would thd_copy_valid_tokens be a more descriptive name?

transformer_engine/pytorch/cpp_extensions/fused_attn.py
```diff
@@ -366,23 +365,19 @@ def fused_attn_fwd(
...
-                    valid = torch.repeat_interleave(values, counts)
-                    # Finally, replace invalid (F) positions with -inf
+                    tq = max_tensor.shape[0]
+                    starts = cu_seqlens_q_padded[:-1].to(device=max_tensor.device, dtype=torch.long)
+                    ends = (starts + actual_seqlens.to(dtype=torch.long)).clamp(max=tq)
+                    delta = torch.zeros(tq + 1, dtype=torch.int32, device=max_tensor.device)
+                    updates = torch.ones_like(starts, dtype=torch.int32)
+                    delta.scatter_add_(0, starts.clamp(max=tq), updates)
+                    delta.scatter_add_(0, ends, -updates)
+                    valid = delta[:-1].cumsum(0) > 0
```
cyanguwa, line 377: Does this change (L373-L380) improve performance? There still seems to be quite a few small kernels involved. 

Also, is converting to `torch.long` necessary here?

transformer_engine/pytorch/csrc/extensions/attention.cpp
```diff
@@ -975,6 +975,52 @@ at::Tensor thd_get_partitioned_indices(const at::Tensor &cu_seqlens, int total_t
...
+  auto te_cu_seqlens = makeTransformerEngineTensor(cu_seqlens);
+  auto te_out = makeTransformerEngineTensor(out);
+
+  nvte_cp_thd_reorder(te_inp.data(), te_cu_seqlens.data(), te_out.data(), cp_size, scatter ? 1 : 0,
+                      total_tokens, at::cuda::getCurrentCUDAStream());
+
+  return out;
+}
+
+void thd_valid_copy(at::Tensor out, const at::Tensor &inp, const at::Tensor &cu_seqlens_padded,
```
cyanguwa: Should `cp_` be in the names as well? i.e. `thd_cp_reorder_sequences`/`cp_thd_reorder_sequences` or `thd_cp_copy_valid_tokens`/`cp_thd_copy_valid_tokens`? (Just a suggestion)

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -550,6 +673,12 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
   m.def("thd_get_partitioned_indices", &transformer_engine::pytorch::thd_get_partitioned_indices,
         "Generate partitioned indices for inputs in THD format",
         py::call_guard<py::gil_scoped_release>());
+  m.def("thd_reorder", &transformer_engine::pytorch::thd_reorder,
+        "Fused dual-chunk THD reorder for context parallel (gather/scatter), inline index",
+        py::call_guard<py::gil_scoped_release>());
+  m.def("thd_valid_copy", &transformer_engine::pytorch::thd_valid_copy,
+        "Sync-free copy of valid THD token rows into an accumulator (CP AllGather fwd/bwd)",
```
cyanguwa: What's "token rows" here? Sequences? Ranks?

Also, what's "inline index" in thd_reorder? :)

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -168,6 +168,129 @@ void init_router_bindings(pybind11::module &m) {
...
+      .def("copy_into_buffer",
+           static_cast<void (CommOverlapP2P::*)(const at::Tensor &, bool)>(
+               &CommOverlapP2P::copy_into_buffer),
+           py::arg("input"), py::arg("local_chunk") = false)
+      .def("get_buffer", &CommOverlapP2P::get_buffer, py::arg("local_chunk") = false,
+           py::arg("shape") = std::nullopt)
+      .def("get_communication_stream", &CommOverlapP2P::get_communication_stream);
+}
+
+}  // namespace
```
cyanguwa: This change doesn't seem to be related to this PR. Is it necessary?

cyanguwa (changes requested):

cyanguwa:
I feel the code changes are complex enough that I have to rely on the tests. Please make sure the tests cover both Fused and Flash v3 backends, SWA/full attention/causal mask, pad_between_seqs=T/F combinations, and post the performance numbers if there's any (for the new reorder/valid_copy kernels vs the old PyTorch-based functions). Thanks!

cyanguwa:
Also, please fix the DCO, and make the comments a bit more succinct if you can (some comments are quite thorough but also very long :) ). Thanks!

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Pushed in `27d2b84a`, with pre-commit follow-up `af2bd1c3`: the duplicate check is gone. The only remaining `attn_bias_type == "no_bias"` assert in this AllGather support block is the CP all_gather support check.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Pushed in `27d2b84a`, with pre-commit follow-up `af2bd1c3`: `visible_padded` and `visible_actual` are now computed only inside the `if causal or sliding_window_attn:` block where they are used.

tests/pytorch/attention/test_attention_with_cp.py
sudhakarsingh27: Pushed in `27d2b84a`, with pre-commit follow-up `af2bd1c3`: combined the nested THD/a2a+p2p skip. I also fixed the adjacent FlashAttention THD+all_gather skip so FA3-required cases still skip when FA3 is absent; validated both before and after FA3 install.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Pushed in `27d2b84a`, with pre-commit follow-up `af2bd1c3`: `out` now always uses `o_shape = q.shape[:-1] + v.shape[-1:]`, so `d_v != d_q` is handled.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Pushed in `27d2b84a`, with pre-commit follow-up `af2bd1c3`: removed the `t3hd` duplicate aux tensor special case, matching #2959’s removal of the legacy FP8DS path.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: `out*` can't be zeroed because FA3 doesn't take in a tensor for out (it returns an output tensor). But backward does allow zeroing of `dq`, `dk`, `dv`.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Pushed in `27d2b84a`, with pre-commit follow-up `af2bd1c3`: renamed the Python-visible/C++ wrapper to `thd_cp_copy_valid_tokens`.

transformer_engine/pytorch/cpp_extensions/fused_attn.py
sudhakarsingh27: I think this was a correctness issue the former approach didn't account for AllGather's offset based cu_seqlens

transformer_engine/pytorch/csrc/extensions/attention.cpp
sudhakarsingh27: Pushed in `27d2b84a`, with pre-commit follow-up `af2bd1c3`: the Python-visible/C++ wrapper names are now `thd_cp_reorder_sequences` and `thd_cp_copy_valid_tokens`. The underlying C API names remain `nvte_cp_thd_*`.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
sudhakarsingh27: Pushed in `27d2b84a`, with pre-commit follow-up `af2bd1c3`: cleaned up the pybind docstrings. They no longer use “token rows” or “inline index”; they now describe reordering between CP rank-sharded dual-chunk order and contiguous per-sequence order, and copying valid THD sequence entries from a padded tensor.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
sudhakarsingh27: Pushed in `27d2b84a`, with pre-commit follow-up `af2bd1c3`: removed the unrelated `bind_comm_overlap_extensions` extraction. I checked NVIDIA `origin/main`; it does not have that extraction, and the effective pybind PR diff versus `origin/main` is now just the two THD CP bindings.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Still resolved at PR head `af2bd1c3`. The Python wrappers are not A2A-specific (`reorder_thd_sequences_to_contiguous` / `reorder_thd_sequences_to_rank_sharded`) and now call the renamed fused binding `tex.thd_cp_reorder_sequences`.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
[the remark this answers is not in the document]
sudhakarsingh27: Still resolved at PR head `af2bd1c3`. The old Python permutation/cache path is gone; the wrappers call fused `tex.thd_cp_reorder_sequences` with device-resident `cu_seqlens`, so there is no Python CUDA-device cache key for this path.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: You are right; my earlier replies misread this as the extra generic `attn_bias_type` / head-dim checks and missed that the non-THD padding assertion in the early branch duplicated the later AllGather support assert. Fixed in pushed commit `46b764e`: the early block now only enforces THD requires a padding mask, and the later assert handles the non-THD `padding` unsupported case. I also fixed the DCO failure by rewriting the two flagged commits with signoff trailers; DCO is now passing.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3214,55 +3276,181 @@ def forward(
...
+                visible_actual = [
+                    torch.minimum(actual_seqlens_kv, visible_padded_split)
+                    for visible_padded_split in visible_padded
+                ]
+                thd_cu_seqlens_kv_per_step = [
+                    torch.zeros_like(cu_seqlens_kv_original) for _ in range(2)
+                ]
+                # Adjust chunks for each step
+                thd_cu_seqlens_kv_per_step[0][1:] = visible_actual[0].cumsum(0)
+                thd_cu_seqlens_kv_per_step[1][1:] = visible_actual[1].cumsum(0)
```
cyanguwa, line 3299: L3286-L3370 could be cudafied if this becomes a performance bottleneck.

cyanguwa (dismissed):

cyanguwa (approved):

PR #2830, opened 2026-04-03
Port softmax ops to libtorch stable ABI

pstjohn:
Proof of concept for migrating pybind11 functions to the PyTorch stable ABI. Ports all 8 scaled softmax functions:

- Add stable_common.h with stable ABI helpers (tensor allocation, TensorWrapper construction, CUDA stream, dtype converters)
- Add registration.cpp with STABLE_TORCH_LIBRARY schema definitions
- Rewrite softmax.cpp: at::Tensor -> torch::stable::Tensor, use stable allocation and stream APIs, TORCH_BOX() for impl registration
- Remove softmax registrations from pybind.cpp
- Update Python callers to use torch.ops.transformer_engine_stable

The pattern is mechanical (API translation, no logic changes) and establishes the template for porting the remaining ~70 Category A functions that have no py::handle/py::object dependencies.

Breakdown of currently registered functions by category: https://docs.google.com/spreadsheets/d/1p9bgmas65M03-yak3zmPImMLXfb7mp2fG_VVcJPNFqo/edit?gid=1617092873#gid=1617092873

### DIFF
--- a/build_tools/pytorch.py
+++ b/build_tools/pytorch.py
@@ -14,7 +14,15 @@ from typing import List
 
 def install_requirements() -> List[str]:
     """Install dependencies for TE/PyTorch extensions."""
-    return ["torch>=2.1", "einops", "onnxscript", "onnx", "packaging", "pydantic", "nvdlfw-inspect"]
+    return [
+        "torch>=2.10",
+        "einops",
+        "onnxscript",
+        "onnx",
+        "packaging",
+        "pydantic",
+        "nvdlfw-inspect",
+    ]
 
 
 def test_requirements() -> List[str]:
--- a/transformer_engine/pytorch/attention/dot_product_attention/softmax.py
+++ b/transformer_engine/pytorch/attention/dot_product_attention/softmax.py
@@ -7,8 +7,9 @@ import os
 from typing import Callable, Tuple, Union, Optional
 import torch
 from torch import nn
-import transformer_engine_torch as tex
 from transformer_engine.pytorch.export import is_in_onnx_export_mode
+
+_ops = torch.ops.transformer_engine
 
 
 THREADS_PER_WARP = 32
@@ -47,7 +48,7 @@ class ScaledUpperTriangMaskedSoftmax(torch.autograd.Function):
     def forward(ctx, inputs: torch.Tensor, scale: float) -> torch.Tensor:
         """ScaledUpperTriangMaskedSoftmax fwd"""
         scale_t = torch.tensor([scale])
-        softmax_results = tex.scaled_upper_triang_masked_softmax_forward(inputs, scale_t[0])
+        softmax_results = _ops.scaled_upper_triang_masked_softmax_forward(inputs, scale_t[0])
 
         ctx.save_for_backward(softmax_results, scale_t)
         return softmax_results
@@ -56,7 +57,7 @@ class ScaledUpperTriangMaskedSoftmax(torch.autograd.Function):
     def backward(ctx, output_grads: torch.Tensor) -> Tuple[Union[torch.Tensor, None], ...]:
         """ScaledUpperTriangMaskedSoftmax bwd"""
         softmax_results, scale_t = ctx.saved_tensors
-        input_grads = tex.scaled_upper_triang_masked_softmax_backward(
+        input_grads = _ops.scaled_upper_triang_masked_softmax_backward(
             output_grads, softmax_results, scale_t[0]
         )
 
@@ -75,7 +76,7 @@ class ScaledAlignedCausalMaskedSoftmax(torch.autograd.Function):
     def forward(ctx, inputs: torch.Tensor, scale: float) -> torch.Tensor:
         """ScaledAlignedCausalMaskedSoftmax fwd"""
         scale_t = torch.tensor([scale])
-        softmax_results = tex.scaled_aligned_causal_masked_softmax_forward(inputs, scale_t[0])
+        softmax_results = _ops.scaled_aligned_causal_masked_softmax_forward(inputs, scale_t[0])
         ctx.save_for_backward(softmax_results, scale_t)
         return softmax_results
 
@@ -83,7 +84,7 @@ class ScaledAlignedCausalMaskedSoftmax(torch.autograd.Function):
     def backward(ctx, output_grads: torch.Tensor) -> Tuple[Union[torch.Tensor, None], ...]:
         """ScaledAlignedCausalMaskedSoftmax bwd"""
         softmax_results, scale_t = ctx.saved_tensors
-        input_grads = tex.scaled_aligned_causal_masked_softmax_backward(
+        input_grads = _ops.scaled_aligned_causal_masked_softmax_backward(
             output_grads, softmax_results, scale_t[0]
         )
 
@@ -103,7 +104,7 @@ class ScaledMaskedSoftmax(torch.autograd.Function):
         """ScaledMaskedSoftmax fwd"""
         scale_t = torch.tensor([scale])
 
-        softmax_results = tex.scaled_masked_softmax_forward(inputs, mask, scale_t[0])
+        softmax_results = _ops.scaled_masked_softmax_forward(inputs, mask, scale_t[0])
         ctx.save_for_backward(softmax_results, scale_t)
         return softmax_results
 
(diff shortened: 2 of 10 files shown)


transformer_engine/pytorch/csrc/extensions/softmax.cpp
```diff
@@ -4,234 +4,210 @@
...
   auto mask_cu = makeTransformerEngineTensor(mask);
   auto softmax_results_cu = makeTransformerEngineTensor(softmax_results);
 
   nvte_scaled_masked_softmax_forward(input_cu.data(), mask_cu.data(), softmax_results_cu.data(),
-                                     scale_factor, at::cuda::getCurrentCUDAStream());
+                                     static_cast<float>(scale_factor),
+                                     getCurrentCUDAStreamRaw(input.get_device_index()));
 
   return softmax_results;
 }
```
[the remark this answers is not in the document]
pstjohn: Fixed in 56111505 — restored all four mask shape guards (`pad_batches`, `mask.size(1) == 1`, query/key dim checks).

transformer_engine/pytorch/csrc/extensions/softmax.cpp
```diff
@@ -4,234 +4,210 @@
...
 
-  // Produce gradients in place.
   nvte_scaled_upper_triang_masked_softmax_backward(
-      output_grads_cu.data(), softmax_results_cu.data(), output_grads_cu.data(), scale_factor,
-      at::cuda::getCurrentCUDAStream());
+      output_grads_cu.data(), softmax_results_cu.data(), output_grads_cu.data(),
+      static_cast<float>(scale_factor), getCurrentCUDAStreamRaw(output_grads.get_device_index()));
 
   return output_grads;
 }
```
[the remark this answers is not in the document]
pstjohn: Fixed in 56111505 — restored the square-shape check (`output_grads.sizes()[1] == output_grads.sizes()[2]`).

transformer_engine/pytorch/csrc/extensions/softmax.cpp
```diff
@@ -4,234 +4,210 @@
...
   auto softmax_results_cu = makeTransformerEngineTensor(softmax_results);
 
-  nvte_scaled_softmax_forward(input_cu.data(), softmax_results_cu.data(), scale_factor,
-                              at::cuda::getCurrentCUDAStream());
+  nvte_scaled_softmax_forward(input_cu.data(), softmax_results_cu.data(),
+                              static_cast<float>(scale_factor),
+                              getCurrentCUDAStreamRaw(input.get_device_index()));
 
   return softmax_results;
 }
```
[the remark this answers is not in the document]
pstjohn: Fixed in 56111505 — added `check_fp16_bf16()` helper in `stable_common.h` and restored dtype validation on all 8 functions.

transformer_engine/pytorch/attention/dot_product_attention/softmax.py
```diff
@@ -7,9 +7,10 @@
 from typing import Callable, Tuple, Union, Optional
 import torch
 from torch import nn
-import transformer_engine_torch as tex
 from transformer_engine.pytorch.export import is_in_onnx_export_mode
 
+_ops = torch.ops.transformer_engine
```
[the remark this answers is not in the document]
pstjohn: `torch.ops` is a lazy namespace — `torch.ops.transformer_engine` succeeds even if the library is not loaded. The actual resolution happens at call time (e.g. `_ops.scaled_softmax_forward(...)`), which will raise a clear `AttributeError` if the op is not registered. Wrapping every call in try/except would add overhead without improving the error message, and failing at import time would be worse since it would prevent importing the module entirely even for code paths that don't use these ops.

transformer_engine/pytorch/csrc/stable_common.h
```diff
@@ -0,0 +1,135 @@
...
+#ifndef USE_CUDA
+#define USE_CUDA
+#endif
+
+// PyTorch Stable ABI headers
+#include <torch/csrc/stable/accelerator.h>
+#include <torch/csrc/stable/library.h>
+#include <torch/csrc/stable/ops.h>
+#include <torch/csrc/stable/tensor.h>
+#include <torch/headeronly/core/ScalarType.h>
```
timmoon10, line 20: What is the minimum PyTorch version that will be required? I see that these files were first added in PyTorch 2.9.0. At the moment, I believe we only enforce 2.1+:
https://github.com/NVIDIA/TransformerEngine/blob/a88fdc1b8139c86e9e03507a4feba652cee0aa5f/transformer_engine/pytorch/__init__.py#L16

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -232,31 +232,7 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
...
-        "Scaled Upper-Triangular Masked Softmax BWD", py::call_guard<py::gil_scoped_release>());
-  m.def("scaled_aligned_causal_masked_softmax_forward",
-        &transformer_engine::pytorch::scaled_aligned_causal_masked_softmax_forward,
-        "Scaled Bottom-Right Corner Aligned Masked Softmax FWD",
-        py::call_guard<py::gil_scoped_release>());
-  m.def("scaled_aligned_causal_masked_softmax_backward",
-        &transformer_engine::pytorch::scaled_aligned_causal_masked_softmax_backward,
-        "Scaled Bottom-Right Corner Aligned Masked Softmax BWD",
-        py::call_guard<py::gil_scoped_release>());
+  // Softmax functions: moved to stable ABI (registration.cpp + softmax.cpp)
```
timmoon10: Dead comment.

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -350,31 +350,9 @@ py::object dropout_bwd(const at::Tensor &grad_output, const at::Tensor &mask,
                        std::optional<at::Tensor> grad_input = std::nullopt);
 
 /***************************************************************************************************
- * Softmax
+ * Softmax: moved to stable ABI (stable_common.h + softmax.cpp + registration.cpp)
  **************************************************************************************************/
```
timmoon10: It's helpful how coding assistants will add notes so you can follow along with what it did, but keeping them in the code results in useless comments. We don't need the code to keep a historical log of how it's changed, that's what Git is for.
```suggestion
```

transformer_engine/pytorch/csrc/stable_common.h
pstjohn: Ah, yes. Probably 2.10 to be honest, [torch::stable::empty](https://docs.pytorch.org/cppdocs/api/function_namespacetorch_1_1stable_1a4630f53cacac471d5a37d0e110640835.html) was added there

transformer_engine/pytorch/csrc/extensions.h
pstjohn: I think the idea is that, 🤞, we'll eventually remove this file entirely 😄

But i can definitely remove

transformer_engine/pytorch/csrc/extensions/pybind.cpp
pstjohn: removed!

pstjohn:
Here's a demo PR for the next set of easy modules:
https://github.com/pstjohn/TransformerEngine/pull/2

build_tools/pytorch.py
```diff
@@ -14,7 +14,15 @@
 
 def install_requirements() -> List[str]:
     """Install dependencies for TE/PyTorch extensions."""
-    return ["torch>=2.1", "einops", "onnxscript", "onnx", "packaging", "pydantic", "nvdlfw-inspect"]
+    return [
+        "torch>=2.10",
```
ptrendx, line 18: Requiring the (close to) latest Pytorch is a problem which I don't think we will be able to do before TE 3.0 to be honest.

pstjohn: 😢  fair enough -- is there a timeline for TE 3.0?

The alternative, that might let us get away with a lower pytorch version at the cost of a more extensive refactor, is to make sure any work matrices get allocated on the python side and passed in as arguments

timmoon10: Unfortunately, we've found that allocating in Python has non-trivial CPU overhead compared to allocating in C++.

pstjohn: If we could validate this and get through all the ops, it would be cool if we could completely remove the pybind11 library in TE 3.0 🤷

PR #2435, opened 2025-12-01
support cuda graph capture offloading module

lhb8125:
# Description

This PR supports offloading modules captured by partial cuda graph in Megatron-LM.

Fixes # (issue)

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Please list the changes introduced in this PR:

- Megatron's offloading relys on cpu_offload_v1 but we need to reuse mark_not_offload() so that the weights won't be offloaded.
- Do not offload the output of core_attention
- Refine the allocation strategy of fp8&fp4 tensors because the previous impl allocates tensor by from_blob(), which is not compatitable with ```record_stream()```
- Add ```pre_warmup_hook``` and ```post_warmup_hook```
- Passing ```cuda_graph_stream``` and ```cuda_graph_event``` to ```user_kwargs``` so that the fwd&bwd replay runs at a side stream, where the ```cuda_graph_event``` records on current stream after finishing computing.

# Checklist:

- [ ] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [ ] The functionality is complete
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2831, opened 2026-04-03
Add capture_time_hooks to make_graphed_callables for non-capturable per-callable hooks

buptzyb:
# Description

## Summary

This PR adds `capture_time_hooks` to `make_graphed_callables` for per-callable work that must run during warmup and CUDA Graph construction but remain outside the CUDA Graph capture context. This is needed for non-capturable operations such as CPU-side state updates and FSDP parameter un-shard/re-shard work that may allocate memory.

These callbacks are graph-construction hooks, not PyTorch `Module` hooks: they receive only the original callable/module, must return `None`, are not recorded into the graph, and are not replayed.

## Changes

- Add one `capture_time_hooks` entry per callable, with support for `forward_pre_hooks`, `forward_hooks`, `backward_pre_hooks`, and `backward_hooks`.
- Run the hooks around each callable's forward and backward during warmup and graph construction, outside the corresponding CUDA Graph capture context.
- Support both capture paths: no-order warmup runs all forwards followed by backwards in reverse order, while ordered warmup follows `_order` across chunks and microbatches. Existing `backward_dw()` warmup behavior is unchanged.
- Call the existing `pre_warmup_hook` and `post_warmup_hook` once around the complete warmup phase rather than once per callable.
- Reject modules with registered backward pre-hooks consistently with the existing checks for other pre-registered `Module` hooks.
- Add parameterized CUDA coverage for both `_order=None` and `_order=[1, 2, -2, -1]`.

## Hook contract

`capture_time_hooks` is a list with one entry per callable. Each entry may be `None` or a dictionary containing any subset of the four supported hook groups. Each group is a `{hook_id: hook_fn}` dictionary, and every hook has signature `hook(module) -> None`.

The list length must match the number of callables. Missing hook groups are treated as empty, unsupported group names raise `ValueError`, and a non-`None` hook return raises `RuntimeError`. The hooks cannot replace inputs, outputs, `grad_inputs`, or `grad_outputs`.

Normal PyTorch `Module` hooks may still be registered after graph construction.

## Execution order

For each callable, the order is:

1. `forward_pre_hooks`
2. forward execution
3. `forward_hooks`
4. `backward_pre_hooks` (training only)
5. backward execution (training only)
6. `backward_hooks` (training only)

During graph construction, pre-hooks run before entering the corresponding capture context and post-hooks run after leaving it.

## Tests and validation

The CUDA test verifies:

- exact forward/backward hook order;
- original module identity;
- execution outside CUDA Graph capture;
- invocation during every warmup iteration and graph construction;
- no invocation during graph replay.

Validation performed:

- focused capture-time hook test on H100: `2 passed`;
- full `tests/pytorch/test_cuda_graphs.py` on H100: `412 passed, 423 skipped`;
- `pylint==3.3.1`, pre-commit, Python compilation, and diff checks passed.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist

### DIFF
--- a/transformer_engine/pytorch/graph.py
+++ b/transformer_engine/pytorch/graph.py
@@ -703,28 +703,34 @@ def _make_graphed_callables(
                     fwd_graph = fwd_graphs[per_callable_fwd_idx]
 
                     # Call pre_forward hooks before forward graph capture (outside capture context)
-                    if capture_time_hooks is not None and capture_time_hooks[callable_idx] is not None:
+                    if (
+                        capture_time_hooks is not None
+                        and capture_time_hooks[callable_idx] is not None
+                    ):
                         _hooks = capture_time_hooks[callable_idx]
                         _pre_fwd_with_kwargs = _hooks.get("forward_pre_hooks_with_kwargs", {})
                         for hook_id, hook in _hooks.get("forward_pre_hooks", {}).items():
                             if hook_id in _pre_fwd_with_kwargs:
                                 if hook(func, args, kwargs) is not None:
                                     raise RuntimeError(
-                                        "capture_time_hooks forward_pre_hooks must not return a value "
-                                        "(args/kwargs must not be modified via hook return)"
+                                        "capture_time_hooks forward_pre_hooks must not return a"
+                                        " value (args/kwargs must not be modified via hook return)"
                                     )
                             else:
                                 if hook(func, args) is not None:
                                     raise RuntimeError(
-                                        "capture_time_hooks forward_pre_hooks must not return a value "
-                                        "(args must not be modified via hook return)"
+                                        "capture_time_hooks forward_pre_hooks must not return a"
+                                        " value (args must not be modified via hook return)"
                                     )
 
                     with _graph_context_wrapper(fwd_graph, pool=mempool):
                         outputs = func(*args, **kwargs)
 
                     # Call forward hooks after forward graph capture (outside capture context)
-                    if capture_time_hooks is not None and capture_time_hooks[callable_idx] is not None:
+                    if (
+                        capture_time_hooks is not None
+                        and capture_time_hooks[callable_idx] is not None
+                    ):
                         _hooks = capture_time_hooks[callable_idx]
                         _fwd_with_kwargs = _hooks.get("forward_hooks_with_kwargs", {})
                         for hook_id, hook in _hooks.get("forward_hooks", {}).items():
@@ -851,8 +857,8 @@ def _make_graphed_callables(
                             ].values():
                                 if hook(callable_module, static_grad_outputs) is not None:
                                     raise RuntimeError(
-                                        "capture_time_hooks backward_pre_hooks must not return a value "
-                                        "(grad_output must not be modified via hook return)"
+                                        "capture_time_hooks backward_pre_hooks must not return a"
+                                        " value (grad_output must not be modified via hook return)"
                                     )
 
                         inputs = tuple(i for i in static_input_surface if i.requires_grad)
@@ -877,7 +883,10 @@ def _make_graphed_callables(
                             # Get the callable module for this backward index
                             callable_module = graph_callables[per_callable_bwd_idx]
                             for hook in capture_time_hooks[callable_idx]["backward_hooks"].values():
-                                if hook(callable_module, grad_inputs, static_grad_outputs) is not None:
+                                if (
+                                    hook(callable_module, grad_inputs, static_grad_outputs)
+                                    is not None
+                                ):
                                     raise RuntimeError(
                                         "capture_time_hooks backward_hooks must not return a value "
                                         "(grad_input must not be modified via hook return)"
--- a/transformer_engine/pytorch/graph.py
+++ b/transformer_engine/pytorch/graph.py
@@ -365,6 +365,12 @@ def _make_graphed_callables(
                 + "for each callable must contain only Tensors. Other types are not allowed."
             )
 
+    if capture_time_hooks is not None and len(capture_time_hooks) != len(callables):
+        raise ValueError(
+            f"capture_time_hooks has {len(capture_time_hooks)} entries but there are "
+            f"{len(callables)} callables"
+        )
+
     # If a callable is an nn.Module, its graph's full input surface is the args the user explicitly
     # passes to forward (ie, its sample_args) AND the module's parameter attributes.
     # Note: These per_callable_* variables are not actually
@@ -454,7 +460,7 @@ def _make_graphed_callables(
     visited_te_modules = {}
     need_bwd_dw_graph = {}
 
-    def _run_warmup_forward(func_idx, func):
+    def _run_warmup_forward(func_idx, func, callable_idx):
         """Run forward for one callable during warmup; returns flattened outputs."""
         args = sample_args[func_idx]
         kwargs = sample_kwargs[func_idx]
@@ -484,42 +490,44 @@ def _make_graphed_callables(
 
         if (
             capture_time_hooks is not None
-            and func_idx < len(capture_time_hooks)
-            and capture_time_hooks[func_idx] is not None
-            and "forward_pre" in capture_time_hooks[func_idx]
+            and capture_time_hooks[callable_idx] is not None
+            and "pre_forward" in capture_time_hooks[callable_idx]
         ):
-            for hook in capture_time_hooks[func_idx]["forward_pre"].values():
-                hook(func, args, kwargs)
+            for hook in capture_time_hooks[callable_idx]["pre_forward"].values():
+                result = hook(func, args, kwargs)
+                if result is not None:
+                    args, kwargs = result
 
         hooks = []
         for module in func.modules():
             hooks.append(module.register_forward_hook(hook_fn))
-        outputs, _ = _tree_flatten(func(*args, **kwargs))
+        outputs = func(*args, **kwargs)
         for hook in hooks:
             hook.remove()
 
         if (
             capture_time_hooks is not None
-            and func_idx < len(capture_time_hooks)
-            and capture_time_hooks[func_idx] is not None
-            and "forward" in capture_time_hooks[func_idx]
+            and capture_time_hooks[callable_idx] is not None
+            and "forward" in capture_time_hooks[callable_idx]
         ):
-            for hook in capture_time_hooks[func_idx]["forward"].values():
-                hook(func, args, outputs)
+            for hook in capture_time_hooks[callable_idx]["forward"].values():
+                result = hook(func, args, outputs)
+                if result is not None:
+                    outputs = result
 
+        outputs, _ = _tree_flatten(outputs)
         return outputs
 
-    def _run_warmup_backward(func_idx, func, outputs, warmup_iter):
+    def _run_warmup_backward(func_idx, func, outputs, warmup_iter, callable_idx):
         """Run dgrad backward for one callable during warmup."""
         static_input_surface = per_callable_static_input_surfaces[func_idx]
 
         if (
             capture_time_hooks is not None
-            and func_idx < len(capture_time_hooks)
-            and capture_time_hooks[func_idx] is not None
-            and "pre_backward" in capture_time_hooks[func_idx]
+            and capture_time_hooks[callable_idx] is not None
+            and "pre_backward" in capture_time_hooks[callable_idx]
         ):
-            for hook in capture_time_hooks[func_idx]["pre_backward"].values():
+            for hook in capture_time_hooks[callable_idx]["pre_backward"].values():
                 hook(func)
 
         inputs = tuple(i for i in static_input_surface if i.requires_grad)
@@ -533,11 +541,10 @@ def _make_graphed_callables(
 
         if (
             capture_time_hooks is not None
-            and func_idx < len(capture_time_hooks)
-            and capture_time_hooks[func_idx] is not None
-            and "backward" in capture_time_hooks[func_idx]
+            and capture_time_hooks[callable_idx] is not None
+            and "backward" in capture_time_hooks[callable_idx]
         ):
-            for hook in capture_time_hooks[func_idx]["backward"].values():
+            for hook in capture_time_hooks[callable_idx]["backward"].values():
                 hook(func)
 
         # Filter module params that get None grad from grad_inputs and remove them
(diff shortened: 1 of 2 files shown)


transformer_engine/pytorch/graph.py
```diff
@@ -1187,6 +1365,23 @@ def make_graphed_callables(
...
+                   Per-callable hooks invoked at capture time (during warmup iterations and
+                   graph capture), but intentionally executed **outside** the CUDA graph
+                   capture context so they are **not** recorded into the graph and will
+                   **not** be replayed. Use this for operations that are inherently
+                   non-capturable but essential for correct module execution, such as
+                   CPU-side state updates.
+                   Each element corresponds to one callable and is a dict with any subset
+                   of the following keys:
+                   - ``"forward_pre"``: dict of hooks called *before* the forward pass.
+                     Hook signature: ``hook(module, args, kwargs)``.
```
timmoon10: Doesn't this sound weird? Let's make it `"pre_forward"` to be consistent with `"pre_backward"`.

timmoon10: Why is `capture_time_hooks[callable_idx]["forward_pre"]` a `dict`? As far as I can tell, we just iterate through the hooks, so it would be more straightforward to make it a `list` (and similar for `"forward"`, `"pre_backward"`, "`backward`").

transformer_engine/pytorch/graph.py
```diff
@@ -448,111 +449,185 @@ def _make_graphed_callables(
...
-                    need_bwd_dw_graph[func_idx] = need_backward_dw
-                else:
-                    grad_inputs = None
-                del outputs, grad_inputs
-            if post_warmup_hook is not None:
-                post_warmup_hook()
+                            bwd_idx[m_chunk] += 1
+
+        if post_warmup_hook is not None:
+            post_warmup_hook()
```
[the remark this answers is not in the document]
timmoon10: We should confirm that this doesn't cause problems, and probably update the docs for `pre_warmup_hook`/`post_warmup_hook` to be more explicit about its behavior. That said, I don't think this function will break the use-cases from  https://github.com/NVIDIA/TransformerEngine/pull/2435/changes#r2850937378.

CC @lhb8125

transformer_engine/pytorch/graph.py
```diff
@@ -448,111 +449,185 @@ def _make_graphed_callables(
...
+                            modules.add(basic_op)
+            if modules:
+                if func_idx not in visited_te_modules:
+                    visited_te_modules[func_idx] = modules
+                else:
+                    visited_te_modules[func_idx].update(modules)
+
+        if (
+            capture_time_hooks is not None
+            and func_idx < len(capture_time_hooks)
```
timmoon10: I'd prefer if length mismatches triggered errors. This pattern is inviting silent bugs.

transformer_engine/pytorch/graph.py
```diff
@@ -1145,6 +1322,7 @@ def make_graphed_callables(
     _reuse_graph_input_output_buffers: bool = False,
     pre_warmup_hook: Optional[Callable] = None,
     post_warmup_hook: Optional[Callable] = None,
+    capture_time_hooks: Optional[List[Optional[Dict[str, Dict]]]] = None,
```
timmoon10, line 1307: Nit: It's kind of uncomfortable that the APIs for the warmup and capture hooks are inconsistent. Also, the `capture_time_hooks` is slightly misleading because they are applied during warmup, not just during capture.

transformer_engine/pytorch/graph.py
```diff
@@ -448,111 +449,185 @@ def _make_graphed_callables(
...
+            hook.remove()
+
+        if (
+            capture_time_hooks is not None
+            and func_idx < len(capture_time_hooks)
+            and capture_time_hooks[func_idx] is not None
+            and "forward" in capture_time_hooks[func_idx]
+        ):
+            for hook in capture_time_hooks[func_idx]["forward"].values():
+                hook(func, args, outputs)
```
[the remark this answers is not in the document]
timmoon10: This is a good point.

timmoon10 (changes requested):

transformer_engine/pytorch/graph.py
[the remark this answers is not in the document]
buptzyb: Smart agent!

transformer_engine/pytorch/graph.py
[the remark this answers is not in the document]
buptzyb: Yes, I believe lhb8125's idea was to call the two hooks only once, before and after all warmups. The original implementation is kind of a typo, actually.

I updated the docstring: "A hook function that will be called once before all warmup iterations (not once per callable)".

transformer_engine/pytorch/graph.py

buptzyb: Because they are stored as `dict` in the pytorch Module: see https://github.com/pytorch/pytorch/blob/a20b9f8ca14a08acf9b3ffe2dbfa650cf38cc8a6/torch/nn/modules/module.py#L459-L473. I don't know why pytorch do so, but we should keep align with it, so it can be very easy for the users to extract the hook: `capture_time_hooks[callable_idx]["pre_forward"] = callables[callable_idx]._forward_pre_hooks`

transformer_engine/pytorch/graph.py
buptzyb: Removed `and func_idx < len(capture_time_hooks)` so it triggers an error if `capture_time_hooks` is not long enough.

transformer_engine/pytorch/graph.py
buptzyb: Yes, I've thought about the naming for a long time, but still failed to think of a good name...

At first I want to call it `non_capturable_hooks` or `graph_exterior_hooks`, but this only indicates the hooks are not captured into the graph, not saying they won't be called around the graph replay. I also thought of `warmup_capture_hooks`, but this confuses the users when we have three hook arguments all named in `warmup`... Maybe `graph_construction_hooks` is better?

And `per_callable_hooks`? But this doesn't distinguish between capture and replay time...

So eventually I came to `capture_time_hooks`. I persuaded myself that warmup is also a part of the capture time. Do you have a better suggestion?

buptzyb: Unless we give it a very long name: `per_callable_non_capturable_hooks_at_warmup_and_capture_time` lol

buptzyb:
Hi @ksivaman could you help review? Thanks!

transformer_engine/pytorch/graph.py
```diff
@@ -453,111 +465,209 @@ def _make_graphed_callables(
...
+            torch.autograd.backward(outputs_requiring_grad, grad_tensors=grad_outputs)
+            grad_inputs = tuple(input.grad for input in inputs)
+
+        if (
+            capture_time_hooks is not None
+            and capture_time_hooks[callable_idx] is not None
+            and "backward_hooks" in capture_time_hooks[callable_idx]
+        ):
+            for hook in capture_time_hooks[callable_idx]["backward_hooks"].values():
+                if hook(func, grad_inputs, grad_outputs) is not None:
```
timmoon10: Our behavior with `grad_inputs` doesn't match the [`torch.nn.Module` backward hooks](https://docs.pytorch.org/docs/2.12/generated/torch.nn.Module.html#torch.nn.Module.register_full_backward_hook). Here we pass in the grads from both inputs and weights, ignoring non-tensor inputs and tensor inputs without grads. `torch.nn.Module` passes in the grads from the inputs, including non-tensor inputs and inputs without tensor inputs with grads.

transformer_engine/pytorch/graph.py
```diff
@@ -453,111 +465,209 @@ def _make_graphed_callables(
...
+                    if isinstance(module_group, OperationFuser):
+                        for basic_op in module_group._basic_ops:
+                            modules.add(basic_op)
+            if modules:
+                if func_idx not in visited_te_modules:
+                    visited_te_modules[func_idx] = modules
+                else:
+                    visited_te_modules[func_idx].update(modules)
+
+        if capture_time_hooks is not None and capture_time_hooks[callable_idx] is not None:
```
timmoon10: Having existence checks every time we access the hooks is quite burdensome. How about we preprocess `capture_time_hooks` so it has a fixed structure?

<details> <summary> Function to preprocess `capture_time_hooks` </summary>

```python
def _canonicalize_capture_time_hooks(
    num_callables: int,
    capture_time_hooks: Optional[List[Optional[Dict[str, Dict]]]],
) -> List[Dict[str, Dict]]:
    """Fill defaults in capture_time_hooks"""
    expected_keys = {
        "forward_pre_hooks",
        "forward_pre_hooks_with_kwargs",
        "forward_hooks",
        "forward_hooks_with_kwargs",
        "backward_pre_hooks",
        "backward_hooks",
    }

    # Make sure there is one entry per callable
    if capture_time_hooks is None:
        capture_time_hooks = [None] * num_callables
    if len(capture_time_hooks) != num_callables:
        raise ValueError(
            f"capture_time_hooks has {len(capture_time_hooks)} entries, "
            f"but there are {num_callables} callables."
        )

    # Make sure each callable has a dict with expected keys
    for callable_idx in range(num_callables):
        if capture_time_hooks[callable_idx] is None:
            capture_time_hooks[callable_idx] = {}
        hooks = capture_time_hooks[callable_idx]
        for key in expected_keys:
            if key not in callable_hooks:
                hooks[key] = {}
        if set(hooks.keys()) != expected_keys:
            raise ValueError(
                "Found unexpected keys in capture_time_hooks "
                f"({set(hooks.keys()) - expected_keys})."
            )

    return capture_time_hooks
```

</details>

<details> <summary> Triggering pre-forward hooks </summary>

```python
        # Call pre-forward capture-time hooks
        hooks = capture_time_hooks[callable_idx]["forward_pre_hooks"]
        hooks_with_kwargs = capture_time_hooks[callable_idx]["forward_pre_hooks_with_kwargs"]
        for hook_id, hook in hooks.items():
            hook_args = (
                (func, args, kwargs) if hook_id in hooks_with_kwargs else (func, args)
            )
            if hook(*hook_args) is not None:
                raise RuntimeError(
                    "capture_time_hooks forward_pre_hooks must not return a value "
                    "(args/kwargs must not be modified via hook return)"
                )
```

</details>

transformer_engine/pytorch/graph.py
```diff
@@ -688,6 +844,23 @@ def hook_fn(
...
                     if is_training:
+                        # Call pre_backward hooks before backward graph capture (outside capture context)
+                        if (
+                            capture_time_hooks is not None
+                            and capture_time_hooks[callable_idx] is not None
+                            and "backward_pre_hooks" in capture_time_hooks[callable_idx]
+                        ):
+                            # Get the callable module for this backward index
+                            callable_module = graph_callables[per_callable_bwd_idx]
+                            for hook in capture_time_hooks[callable_idx][
```
timmoon10: Shouldn't the hooks match the modules?
```suggestion
                            callable_module = callables[callable_idx]
                            for hook in capture_time_hooks[callable_idx][
```
The existing code is probably correct, but it's so indirect and complicated that I find it hard to check.

transformer_engine/pytorch/graph.py
```diff
@@ -453,111 +465,209 @@ def _make_graphed_callables(
...
-                    # registered to these params are not wrongly triggered.
-                    num_required_grad_sample_args = sum(
-                        arg.requires_grad for arg in flatten_sample_args[func_idx]
+        if (
+            capture_time_hooks is not None
+            and capture_time_hooks[callable_idx] is not None
+            and "backward_pre_hooks" in capture_time_hooks[callable_idx]
+        ):
+            for hook in capture_time_hooks[callable_idx]["backward_pre_hooks"].values():
+                if hook(func, grad_outputs) is not None:
```
timmoon10: We are treating `grad_outputs` inconsistently between warmup and capture. During warmup, we pass in only the required grads. During capture, we pass in grads for all of the outputs (matching [`torch.nn.Module` pre-backward hooks](https://docs.pytorch.org/docs/2.12/generated/torch.nn.Module.html#torch.nn.Module.register_full_backward_pre_hook)).

See @greptile-apps's review in https://github.com/NVIDIA/TransformerEngine/pull/2831#issuecomment-4183686878.

transformer_engine/pytorch/graph.py
```diff
@@ -1191,9 +1447,38 @@ def make_graphed_callables(
...
         inputs and outputs with the same dtype and shape.
     pre_warmup_hook: callable, default = None
-                      A hook function that will be called before the warmup iterations.
+                      A hook function that will be called once before all warmup iterations
+                      (not once per callable).
     post_warmup_hook: callable, default = None
-                      A hook function that will be called after the warmup iterations.
+                      A hook function that will be called once after all warmup iterations
+                      (not once per callable).
+    capture_time_hooks: list of dict, optional
```
timmoon10, line 1352: We put a lot of effort into matching the `torch.nn.Module` hook APIs, but they are causing correctness issues and we don't actually need them for our use-cases. Instead of making promises we can't fulfill, how about we simplify? We can just have `"forward_pre_hooks"`, `"forward_hooks"`, `"backward_pre_hooks"`, `"backward_hooks"`, each with signature `hook(module)`.

timmoon10 (commented):
We should add a test. `make_graphed_callables` has gotten too complicated for me to have any confidence of correctness just by looking at the code.

transformer_engine/pytorch/graph.py
buptzyb: Addressed by adopting your later simplification suggestion. Capture-time hooks now have signature `hook(module)` and no longer expose `grad_inputs` or `grad_outputs`, so we no longer claim `torch.nn.Module` backward-hook semantics.

transformer_engine/pytorch/graph.py
buptzyb: Implemented `_canonicalize_capture_time_hooks` based on this suggestion, with the reduced four-key hook contract. It validates one entry per callable, fills missing hook maps, and rejects unsupported keys. Invocation and return-value validation are centralized in `_run_capture_time_hooks`.

transformer_engine/pytorch/graph.py

transformer_engine/pytorch/graph.py
buptzyb: Now capture-time hooks no longer receive `grad_outputs` during either warmup or graph capture. So this problem disappears as well.

transformer_engine/pytorch/graph.py
buptzyb: Implemented as suggested. Thanks!

transformer_engine/pytorch/graph.py
```diff
@@ -453,111 +506,159 @@ def _make_graphed_callables(
...
-                        if hasattr(module, "need_backward_dw") and module.need_backward_dw():
-                            need_backward_dw = True
-                            module.backward_dw()
-                    need_bwd_dw_graph[func_idx] = need_backward_dw
-                else:
-                    grad_inputs = None
-                del outputs, grad_inputs
-            if post_warmup_hook is not None:
-                post_warmup_hook()
+                            bwd_idx[m_chunk] += 1
```
timmoon10, line 658: Noting that we are changing the warmup behavior when `_order is not None`. Instead of just doing a forward and backward for each entry in `callables`, we are performing things in the order specified by `_order`. I think this is an improvement since it makes the warmup and capture passes match more closely.

timmoon10 (approved):
LGTM, pending CI.

PR #2840, opened 2026-04-06
[JAX] Use avg m,n,k heuristics for Grouped GEMM

jberchtold-nvidia:
# Description

This PR computes estimates for the avg M, N, K values for the TE grouped GEMM API. This informs heuristics on which kernel to select on the backend.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Compute avg m, n, k and pass to `nvte_grouped_gemm`
- Small fix to `NVTE_JAX_ENFORCE_V2_GROUPED_GEMM` env var check to ensure the value is an int. Previously `NVTE_JAX_ENFORCE_V2_GROUPED_GEMM=true` would evaluate to false.

# Checklist:

### DIFF
--- a/qa/L0_pytorch_unittest/test.sh
+++ b/qa/L0_pytorch_unittest/test.sh
@@ -42,6 +42,7 @@ python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_gqa.xml $TE_PATH
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_fused_optimizer.xml $TE_PATH/tests/pytorch/test_fused_optimizer.py || test_fail "test_fused_optimizer.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_multi_tensor.xml $TE_PATH/tests/pytorch/test_multi_tensor.py || test_fail "test_multi_tensor.py"
 NVTE_CUTEDSL_FUSED_GROUPED_MLP=1 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_fusible_ops.xml $TE_PATH/tests/pytorch/test_fusible_ops.py || test_fail "test_fusible_ops.py"
+python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_backward_override.xml $TE_PATH/tests/pytorch/test_backward_override.py || test_fail "test_backward_override.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_permutation.xml $TE_PATH/tests/pytorch/test_permutation.py || test_fail "test_permutation.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_parallel_cross_entropy.xml $TE_PATH/tests/pytorch/test_parallel_cross_entropy.py || test_fail "test_parallel_cross_entropy.py"
 python3 -m pytest --tb=auto --junitxml=$XML_LOG_DIR/pytest_test_cpu_offloading.xml $TE_PATH/tests/pytorch/test_cpu_offloading.py || test_fail "test_cpu_offloading.py"
--- a/tests/cpp/operator/test_normalization.cu
+++ b/tests/cpp/operator/test_normalization.cu
@@ -46,6 +46,10 @@ void performTest(const size_t N, const size_t H, const bool zero_centered_gamma,
     GTEST_SKIP() << "cuDNN normalizations not supported on pre-Hopper GPUs yet!";
   }
 
+  if (fused_bwd_add && use_cudnn && (cudnnGetVersion() < 92100)) {
+    GTEST_SKIP() << "cuDNN < 9.21 does not support fused RMSNorm backward+add";
+  }
+
   using WeightType = InputType;
   DType itype = TypeInfo<InputType>::dtype;
   DType wtype = TypeInfo<WeightType>::dtype;
--- a/tests/cpp/operator/test_normalization.h
+++ b/tests/cpp/operator/test_normalization.h
@@ -15,6 +15,7 @@
 
 #include <cuda_bf16.h>
 #include <cuda_runtime.h>
+#include <cudnn.h>
 
 #include <transformer_engine/normalization.h>
 #include <transformer_engine/transformer_engine.h>
--- a/tests/pytorch/test_backward_override.py
+++ b/tests/pytorch/test_backward_override.py
@@ -1,0 +1,1848 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+from __future__ import annotations
+
+from contextlib import nullcontext
+import math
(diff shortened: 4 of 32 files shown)


transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -683,6 +683,37 @@ size_t grouped_gemm_num_gemms(Buffer_Type const &lhs_first_dims, Buffer_Type con
...
+  bool first_ragged = first_dims.element_count() > 0;
+  bool last_ragged = last_dims.element_count() > 0;
+  bool any_ragged = first_ragged || last_ragged;
+  // Per-group left: divide by num_gemms if first dim is ragged OR if tensor has no ragged dims
+  // (G is folded into left_size).  Keep as-is only when last dim is ragged but first is not.
+  size_t pg_left = (first_ragged || !any_ragged) ? left_size / num_gemms : left_size;
+  size_t pg_right = last_ragged ? right_size / num_gemms : right_size;
+  int64_t non_contract = static_cast<int64_t>(is_trans ? pg_right : pg_left);
+  int64_t contract = static_cast<int64_t>(is_trans ? pg_left : pg_right);
+  return {non_contract, contract};
```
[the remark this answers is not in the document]
jberchtold-nvidia: Resolved in latest version

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -683,6 +683,37 @@ size_t grouped_gemm_num_gemms(Buffer_Type const &lhs_first_dims, Buffer_Type con
...
+//   shape, so the static dim's size is already the per-group value.
+//
+// The transpose flag determines the mapping from per-group (left, right) to (m_or_non_contract,
+// k_or_contract):
+//   lhs represents [m, k] (not transposed) or [k, m] (transposed).
+//   rhs represents [k, n] (not transposed) or [n, k] (transposed).
+// Returns {non_contracting_avg, contracting_avg}, i.e. {avg_m, avg_k} for lhs or {avg_n, avg_k}
+// for rhs.
+std::pair<int64_t, int64_t> grouped_gemm_avg_dims(Buffer_Type const &first_dims,
+                                                  Buffer_Type const &last_dims, size_t left_size,
```
tdophung: I am not entirely sure if left_size and right_size are good variable names or not. I did not understand this at first before looking at the comments above, but I also cannot think of a better name for them

tdophung: self-resolve if you also cannot think of a better name

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -683,6 +683,37 @@ size_t grouped_gemm_num_gemms(Buffer_Type const &lhs_first_dims, Buffer_Type con
   }
 }
 
+// Compute per-group average m, n, k from the 2D operand shapes and ragged indicators.
+// For each operand, left_size and right_size are the total 2D dimensions of the entire buffer.
+// - If a dim is ragged (first_dims/last_dims non-empty), its total is the sum of per-group sizes,
+//   so the average is total / num_gemms.
+// - If a dim is static but the tensor has no ragged dims at all, the group batch dimension G is
+//   folded into left_size (since G is always dim 0 and axis_boundary >= 1), so we divide it out.
```
tdophung: should we have an assert for num_gemms == G? Or this is always true based on how this is called already?

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -683,6 +683,37 @@ size_t grouped_gemm_num_gemms(Buffer_Type const &lhs_first_dims, Buffer_Type con
...
+std::pair<int64_t, int64_t> grouped_gemm_avg_dims(Buffer_Type const &first_dims,
+                                                  Buffer_Type const &last_dims, size_t left_size,
+                                                  size_t right_size, size_t num_gemms,
+                                                  bool is_trans) {
+  bool first_ragged = first_dims.element_count() > 0;
+  bool last_ragged = last_dims.element_count() > 0;
+  bool any_ragged = first_ragged || last_ragged;
+  // Per-group left: divide by num_gemms if first dim is ragged OR if tensor has no ragged dims
+  // (G is folded into left_size).  Keep as-is only when last dim is ragged but first is not.
+  size_t pg_left = (first_ragged || !any_ragged) ? left_size / num_gemms : left_size;
```
tdophung: Do we worry about whether the average calculated is an int or not? As in, does left_size have to be divisible by num_gemms in this case. Similarly, for right_size also

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -741,11 +772,21 @@ Error_Type GroupedGemmV2FFI(cudaStream_t stream, Buffer_Type lhs_data, Buffer_Ty
   auto out_tensor = make_grouped_tensor(*output, out_first_dims, out_last_dims, int64_base,
                                         int64_capacity, int64_offset, num_gemms, stream);
 
+  auto [avg_m, avg_k_lhs] = grouped_gemm_avg_dims(lhs_first_dims, lhs_last_dims, lhs_left_size,
+                                                  lhs_right_size, num_gemms, lhs_is_trans);
+  auto [avg_n, avg_k_rhs] = grouped_gemm_avg_dims(rhs_first_dims, rhs_last_dims, rhs_left_size,
+                                                  rhs_right_size, num_gemms, !rhs_is_trans);
+  // Use k from lhs (both sides should agree for well-formed inputs).
+  (void)avg_k_rhs;
```
tdophung: This should in theory be correct but with the averages being rounded if left_size or right_size  is not divisible by num_gemm (depending on which case it falls to in line 710-711), then avg_k_lhs might not be the same as avg_k_rhs. So, maybe put this here?

```suggestion
  NVTE_CHECK(avg_k_lhs == avg_k_rhs,
  "Contracting dimension mismatch: lhs avg_k=", avg_k_lhs,
  " vs rhs avg_k=", avg_k_rhs);
```

transformer_engine/jax/csrc/extensions/gemm.cpp
tdophung: Actually I see left_size, right_size being used all over so this is actuallly consistent with previous code. resolving

transformer_engine/jax/csrc/extensions/gemm.cpp
jberchtold-nvidia: Good point, I've added the assert

transformer_engine/jax/csrc/extensions/gemm.cpp
jberchtold-nvidia: Good point, I've added float math and rounding here. For weights, (G, K, N), we are guaranteed to be an exact multiple of G. For the data, (M, K) where M is ragged, we can only assume M  >= sum(group_sizes), not necessarily that M is divisible by G since the buffer is overallocated.

transformer_engine/jax/csrc/extensions/gemm.cpp
jberchtold-nvidia: I've updated the documentation to clarify that G and "num_gemms" are the same value

PR #2848, opened 2026-04-08
Skip activation kernels when tensor size is zero

timmoon10:
# Description

We have encountered some obscure CUDA errors when calling SwiGLU on a tensor with no entries. This PR handles this case by skipping the kernel launch when the tensor has no entries. Since the kernel implementation is heavily templated, this also fixes this bug for some quantization cases, although I haven't handled them exhaustively.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Skip activation kernels and quantization kernels when tensor size is zero.
- Add empty-tensor test cases in activation unit tests.

# Checklist:

### DIFF
--- a/tests/cpp/operator/test_cast.cu
+++ b/tests/cpp/operator/test_cast.cu
@@ -64,7 +64,7 @@ void performTest(const std::vector<size_t>& shape) {
   cudaDeviceSynchronize();
   auto err = cudaGetLastError();
   ASSERT_EQ(err, cudaSuccess) << cudaGetErrorString(err);
-  if (isFp8Type(otype)) {
+  if (isFp8Type(otype) && full_size > 0) {
     auto [atol_amax, rtol_amax] = getTolerances(DType::kFloat32);
     compareResults("amax", output_c.amax(), ref_amax, atol_amax, rtol_amax);
     float ref_scale_inv = 1.f / output_c.scale();
@@ -91,6 +91,8 @@ std::vector<std::vector<size_t>> test_cases = {
   {5, 160},
   {5, 4, 3, 160},
   {217, 256},
+  {0, 128},
+  {128, 0},
 };
 }  // namespace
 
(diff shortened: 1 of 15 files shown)


tests/cpp/operator/test_cast.cu
```diff
@@ -64,7 +64,7 @@ void performTest(const std::vector<size_t>& shape) {
   cudaDeviceSynchronize();
   auto err = cudaGetLastError();
   ASSERT_EQ(err, cudaSuccess) << cudaGetErrorString(err);
-  if (isFp8Type(otype)) {
+  if (isFp8Type(otype) && full_size > 0) {
```
Oleg-Goncharov, line 67: So this problem shows up not only for activations, but also for a regular cast?

Oleg-Goncharov (dismissed):
LGTM

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -870,6 +870,15 @@ void group_quantize(const GroupedTensor *input, const GroupedTensor *activations
     }
   }
 
+  // Skip kernel if tensor size is zero
+  if (elts_total == 0) {
+    if constexpr (IS_DBIAS) {
+      NVTE_ERROR("Invalid grouped tensor shape for DBias computation (first_logical_dim=",
```
vthumbe1503, line 874: In this case we can output dbias = zero tensor also right instead of throwing an error?

tests/cpp/operator/test_cast.cu
timmoon10: Yep, many of our kernels are not robust to empty tensors. I still expect to see problems in the FP8 block-scale quantization kernels and transpose kernels.

transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh
```diff
@@ -579,6 +579,14 @@ void quantize(const Tensor &input, const Tensor *act_input, const Tensor *noop,
   const size_t rows = input.flat_first_dim();
   const size_t cols = input.flat_last_dim();
 
+  // Skip kernel if tensor size is zero
+  if (rows == 0 || cols == 0) {
+    if constexpr (IS_DBIAS) {
+      NVTE_ERROR("Invalid tensor shape for DBias computation (shape=", input.shape(), ").");
+    }
+    return;
+  }
```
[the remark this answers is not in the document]
timmoon10: Actually, that logic isn't quite right either. If the tensor is empty, then the workspace size will be zero. When the user allocates such a workspace, its data pointer will be null and when they try to run the function it will think it's another workspace size query.

PR #2865, opened 2026-04-10
Optimizations for MXFP8/NVFP4 dequantize kernels

YigongQin:
# Description

  - Handle empty tensors in dequantize for CUDA graph compatibility
  - Add swizzled scale support to the NVFP4 dequantize kernel, reusing the existing MXFP8 swizzle index computation
  - Add C++ unit tests for both NVFP4 and MXFP8 dequantization (including swizzled scale variants)
  - Fix to_cpu() and set_scale() in test infrastructure to correctly sync amax/scale for NVTE_NVFP4_1D_SCALING mode

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

-   Handle empty tensors in dequantize for CUDA graph compatibility — Early return when input has zero elements, avoiding kernel launches on empty tensors.
-   Add GEMM-swizzled scale support to NVFP4 dequantize kernel — Template the kernel with WITH_GEMM_SWIZZLED_SCALES to support reading scales from swizzled layout, reusing the MXFP8 swizzle index computation.
-   Add GEMM-swizzled scale support to MXFP8 dequantize kernel — Extend the MXFP8 dequantize kernel to handle swizzled scale inputs.
-   Add C++ unit tests for NVFP4 dequantization — 21 tests for compact scales + 21 tests for swizzled scales, covering multiple sizes and output dtypes (fp32, bf16, fp16).
-   Add C++ unit tests for MXFP8 dequantization with swizzled scales — New swizzled test suite for MXFP8.
-   Fix to_cpu() to sync amax/scale for NVFP4 tensors — Previously only synced for NVTE_DELAYED_TENSOR_SCALING, causing the CPU reference to use stale amax=0.
-   Fix set_scale() to work for NVFP4 tensors — Same condition fix, enabling the scale to be properly uploaded to GPU before quantization.
-   Fix swizzled test ordering — Move from_cpu() before the FP4 data copy to prevent from_cpu() from overwriting the copied data with zeros.

# Checklist:

### DIFF
--- a/tests/cpp/operator/test_dequantize_nvfp4.cu
+++ b/tests/cpp/operator/test_dequantize_nvfp4.cu
@@ -151,19 +151,19 @@ void performTest_dequantize_nvfp4_swizzled(const size_t rows, const size_t cols)
     setRandomScale(&quantized_swizzled);
     quantized_swizzled.set_with_gemm_swizzled_scales(true);
 
-    // Copy FP4 data
+    // Copy amax and scale from compact to swizzled before FP4 data,
+    // since from_cpu() uploads all CPU buffers (including zero-init data).
+    quantized_compact.to_cpu();
+    quantized_swizzled.set_amax(quantized_compact.amax());
+    quantized_swizzled.set_scale(quantized_compact.scale());
+    quantized_swizzled.from_cpu();
+
+    // Copy FP4 data after from_cpu() to avoid being overwritten
     const size_t data_bytes = rows * cols / 2;
     if (data_bytes > 0) {
         cudaMemcpy(quantized_swizzled.rowwise_dptr(), quantized_compact.rowwise_dptr(),
                    data_bytes, cudaMemcpyDeviceToDevice);
     }
-
-    // Copy amax from compact to swizzled (GPU-to-GPU)
-    // Read amax from compact tensor, set on swizzled tensor, upload
-    quantized_compact.to_cpu();
-    quantized_swizzled.set_amax(quantized_compact.amax());
-    quantized_swizzled.set_scale(quantized_compact.scale());
-    quantized_swizzled.from_cpu();
 
     // Swizzle scales
     if (data_bytes > 0) {
--- a/tests/cpp/operator/test_dequantize_nvfp4.cu
+++ b/tests/cpp/operator/test_dequantize_nvfp4.cu
@@ -1,0 +1,277 @@
+/*************************************************************************
+ * Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+
+#include <cstring>
+#include <memory>
+#include <random>
+#include <vector>
+
+#include <cuda_bf16.h>
+#include <cuda_fp8.h>
+#include <cuda_runtime.h>
+#include <gtest/gtest.h>
+
+#if FP4_TYPE_SUPPORTED
+#include <cuda_fp4.h>
+#endif
+
+#include <transformer_engine/cast.h>
+#include <transformer_engine/swizzle.h>
+#include "../test_common.h"
+#include "transformer_engine/transformer_engine.h"
+
+using namespace transformer_engine;
+using namespace test;
+
+#if FP4_TYPE_SUPPORTED
+
+namespace {
+
+float2 cvt_fp4x2_to_float2(fp4e2m1x2 fp4_pair) {
+    const __half2_raw raw =
+        __nv_cvt_fp4x2_to_halfraw2(
+            *reinterpret_cast<__nv_fp4x2_storage_t *>(&fp4_pair), __NV_E2M1);
+    const __half2 h2(raw);
+    return {static_cast<float>(h2.x), static_cast<float>(h2.y)};
+}
+
+template <typename OType>
+void compute_ref_dequantize_nvfp4(const uint8_t *packed_data,
+                                  const fp8e4m3 *scales,
+                                  float amax,
+                                  OType *output,
+                                  size_t rows,
+                                  size_t cols,
+                                  size_t scale_stride) {
+    constexpr float factor_inv = 1.0f / (6.0f * 448.0f);
+    constexpr size_t BLOCK_SIZE = 16;
+    const size_t Mread = cols / BLOCK_SIZE;
+    const size_t bytes_per_block = BLOCK_SIZE / 2;
+
+    for (size_t row = 0; row < rows; ++row) {
+        for (size_t block = 0; block < Mread; ++block) {
+            const fp8e4m3 scale = scales[row * scale_stride + block];
+            const float final_scale = static_cast<float>(scale) * amax * factor_inv;
+
+            for (size_t pair_idx = 0; pair_idx < bytes_per_block; ++pair_idx) {
+                const size_t byte_idx =
+                    (row * Mread + block) * bytes_per_block + pair_idx;
+                fp4e2m1x2 fp4_pair;
+                std::memcpy(&fp4_pair, &packed_data[byte_idx], 1);
+                const float2 values = cvt_fp4x2_to_float2(fp4_pair);
+
+                const size_t col0 = block * BLOCK_SIZE + pair_idx * 2;
+                output[row * cols + col0] =
+                    static_cast<OType>(values.x * final_scale);
+                output[row * cols + col0 + 1] =
+                    static_cast<OType>(values.y * final_scale);
+            }
+        }
+    }
+}
+
+// Quantize a high-precision input to NVFP4, then dequantize and compare
+// against a CPU reference computed from the quantized data.
+template <typename OutputType>
+void performTest_dequantize_nvfp4(const size_t rows, const size_t cols) {
+    using namespace test;
+    DType otype = TypeInfo<OutputType>::dtype;
+
+    Tensor input("input", std::vector<size_t>{rows, cols}, otype);
+    fillCase<fp32>(&input, InputsFillCase::uniform);
+
+    Tensor quantized("quantized", std::vector<size_t>{rows, cols},
+                     DType::kFloat4E2M1, true, false, NVTE_NVFP4_1D_SCALING);
+    quantized.set_amax(0.0f);
+    setRandomScale(&quantized);
+
+    if (rows > 0 && cols > 0) {
+        nvte_quantize(input.data(), quantized.data(), 0);
+        cudaDeviceSynchronize();
+    }
+
+    Tensor output("output", std::vector<size_t>{rows, cols}, otype, true, false);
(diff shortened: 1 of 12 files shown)


zianglih:
The following relevant unit tests passed on SM100 (with the drop `optimize_for_gemm = False` changes):
```
python3 -m pytest --tb=auto tests/pytorch/test_backward_override.py
python3 -m pytest --tb=auto tests/pytorch/test_sanity.py
python3 -m pytest --tb=auto tests/pytorch/test_cpu_offloading.py
PYTORCH_JIT=0 NVTE_TORCH_COMPILE=0 NVTE_ALLOW_NONDETERMINISTIC_ALGO=0 NVTE_FUSED_ATTN=0 python3 -m pytest --tb=auto tests/pytorch/test_cuda_graphs.py
NVTE_TEST_NVINSPECT_ENABLED=1 NVTE_TEST_NVINSPECT_CONFIG_FILE=tests/pytorch/debug/test_configs/dummy_feature.yaml NVTE_TEST_NVINSPECT_FEATURE_DIRS=transformer_engine/debug/features PYTORCH_JIT=0 NVTE_TORCH_COMPILE=0 NVTE_ALLOW_NONDETERMINISTIC_ALGO=0 python3 -m pytest --tb=auto tests/pytorch/test_sanity.py
```

zianglih:
After this PR, fwd is around 3%-4% faster for DeepSeek shape MoE:
```
# With the optimization
NVTE_BACKWARD_OVERRIDE=dequantized python benchmarks/linear/benchmark_grouped_linear.py --recipe mxfp8 --fwd-only
       m     k     n recipe  num_gemms  grouped_fwd_time_ms
0  16384  7168  2048  mxfp8          4             0.272829
1  32768  7168  2048  mxfp8          4             0.509788
2  65536  7168  2048  mxfp8          4             0.948633
3  98304  7168  2048  mxfp8          4             1.391146
0  16384  7168  2048  mxfp8          8             0.303238
1  32768  7168  2048  mxfp8          8             0.533896
2  65536  7168  2048  mxfp8          8             1.003446
3  98304  7168  2048  mxfp8          8             1.470030

# Without the optimization
git restore --source 77b8681de5cf -- transformer_engine/pytorch/module
NVTE_BACKWARD_OVERRIDE=dequantized python benchmarks/linear/benchmark_grouped_linear.py --recipe mxfp8 --fwd-only
       m     k     n recipe  num_gemms  grouped_fwd_time_ms
0  16384  7168  2048  mxfp8          4             0.282720
1  32768  7168  2048  mxfp8          4             0.526736
2  65536  7168  2048  mxfp8          4             0.982166
3  98304  7168  2048  mxfp8          4             1.451485
0  16384  7168  2048  mxfp8          8             0.313753
1  32768  7168  2048  mxfp8          8             0.551043
2  65536  7168  2048  mxfp8          8             1.040773
3  98304  7168  2048  mxfp8          8             1.527951
```

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    auto [atol, rtol] = getTolerances(otype);
+    if (num_elems > 0) {
+        compareResults("output_nvfp4_swizzled", output_swizzled,
+                       ref_output.get(), true, atol, rtol);
+    }
+}
+
+std::vector<std::pair<size_t, size_t>> nvfp4_tensor_dims = {
```
zhongbozhu, line 215: There is one edge case: 

For MXFP8, When the input shape is like 64x64, it will produce scaling factor shape 64x2, but then zero padded to 128x4. We should be able to inject some very large random values in the padded region during malloc (because we don't use torch.zeros to malloc but torch.empty), and detect whether dequantize results is affected. If things work as expected, this line will be triggered https://github.com/NVIDIA/TransformerEngine/blob/a073ad5b3ff5c1bc00d9e98669deabe901542aad/transformer_engine/common/cast/mxfp8/quantize_mxfp8.cuh#L759  and the dequantize numerics won't be affected.

For NVFP4, I think we optimize for GEMM (or swizzle fusion) is actually not enabled, same for the zero-out edge case handling logic? https://github.com/NVIDIA/TransformerEngine/blob/a073ad5b3ff5c1bc00d9e98669deabe901542aad/transformer_engine/common/cast/nvfp4/quantize_transpose_nvfp4.cuh#L1189 So there shouldn't be any unswizzle logic needed here?

zhongbozhu: For NVFP4, I believe currently only device-init grouped quantize with RHT has the swizzle fusion feature, so the scaling factor zero-out is the job of the dedicated swizzle kernel. So if we dequantize + unswizzle for NVFP4, the unswizzle logic might not be correct.

tests/cpp/operator/test_dequantize_nvfp4.cu
```diff
@@ -0,0 +1,277 @@
...
+    quantized_swizzled.set_amax(0.0f);
+    setRandomScale(&quantized_swizzled);
+    quantized_swizzled.set_with_gemm_swizzled_scales(true);
+
+    // Copy amax and scale from compact to swizzled before FP4 data,
+    // since from_cpu() uploads all CPU buffers (including zero-init data).
+    quantized_compact.to_cpu();
+    quantized_swizzled.set_amax(quantized_compact.amax());
+    quantized_swizzled.set_scale(quantized_compact.scale());
+    quantized_swizzled.from_cpu();
```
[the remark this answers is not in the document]
YigongQin: This redundant line has been removed

tests/cpp/operator/test_dequantize_nvfp4.cu
YigongQin: For both MXFP8 and NVFP4, the unit test logic is: 1. generate compact scales (or from quantization); 2. call `nvte_swizzle_scaling_factors` to swizzle compact scales; 3. compare results of `nvte_dequantize` with compact scales and swizzled scales. Quantize with swizzle fusion is never enabled for both MXFP8 and NVFP4

transformer_engine/pytorch/module/linear.py
```diff
@@ -1710,13 +1710,6 @@ def _get_quantizers(self, fp8_output, fp8_grad, is_grad_enabled):
                 grad_output_quantizer.optimize_for_gemm = True
             if fp8_grad:
                 grad_input_quantizer = self.quantizers["scaling_bwd"][FP8BwdTensorIdx.GRAD_INPUT1]
-        fp8_recipe = FP8GlobalStateManager.get_fp8_recipe()
-        if fp8_recipe.backward_override == "dequantized" and (
-            fp8_recipe.mxfp8() or fp8_recipe.nvfp4()
-        ):
-            input_quantizer.optimize_for_gemm = False
-            if grad_output_quantizer is not None:
-                grad_output_quantizer.optimize_for_gemm = False
```
timmoon10, line 1751: I'm of two minds about this:

- Logically, GEMM-optimized data is not guaranteed to support anything except GEMMs. Even if MXFP8 and NVFP4 dequant happens to support them, these are custom optimizations. Future recipes can not be expected to support dequantizing GEMM-optimzied data by default.
- It's a little pedantic to have edge-case logic that won't be triggered by any of our existing use-cases. Given how subtle this is, I worry about it becoming stale and distracting.

I think for now, this change is fine. However, if we encounter problems in a future recipe, we should reimplement it properly:
```python
# LOGICALLY WRONG!
# Fails if we add a new recipe
if recipe.backward_override == "dequantized" and recipe.future_recipe():
    input_quantizer.optimize_for_gemm = False

# LOGICALLY RIGHT!
# Automatically handles new recipes
if recipe.backward_override == "dequantized" and not (
    recipe.float8_per_tensor_scaling()
    or recipe.float8_block_scaling()
    or recipe.mxfp8()
    or recipe.nvfp4()
):
    input_quantizer.optimize_for_gemm = False
```

CC @ptrendx @ksivaman @zhongbozhu

tests/cpp/test_common.h
```diff
@@ -286,6 +286,16 @@ class Tensor {
     tensor_.set_amax(nullptr, DType::kFloat32, tensor_.defaultShape);
   }
 
+  void set_amax(float amax_val) {
+    if (!amax_cpu_data_) {
+      amax_cpu_data_ = std::make_shared<float>(0);
+      float* amax_gpu = nullptr;
+      cudaMalloc((void**)&amax_gpu, sizeof(float));  // NOLINT(*)
```
timmoon10: It's a code smell that amax needs this special allocation logic that different from all the other data fields (`data`, `columnwise_data`, `scale_inv`, etc).

The right fix is to allocate in the `Tensor` constructor, along with all the other GPU buffers:
https://github.com/NVIDIA/TransformerEngine/blob/0c2e7b09c6d33109803ea089fbf80421a326e0a4/tests/cpp/test_common.cu#L395

While checking this, I've also found that our current implementation leaks the scale and amax buffers. We should deallocate them in the `Tensor` destructor, along with all the other GPU buffers: https://github.com/NVIDIA/TransformerEngine/blob/0c2e7b09c6d33109803ea089fbf80421a326e0a4/tests/cpp/test_common.h#L155-L157

YigongQin: adedd amax allocation in constructor and amax/scale deallocation in destructor

tests/cpp/test_common.h
```diff
@@ -286,6 +294,11 @@ class Tensor {
     tensor_.set_amax(nullptr, DType::kFloat32, tensor_.defaultShape);
   }
 
+  void set_amax(float amax_val) {
+    NVTE_CHECK(amax_cpu_data_);
+    *amax_cpu_data_ = amax_val;
```
timmoon10: We should make sure that the GPU is also updated, similar to `set_scale` and `set_scale_inv`.
```suggestion
    *amax_cpu_data_ = amax_val;
    from_cpu();
```

tests/cpp/test_common.cu
```diff
@@ -396,6 +396,13 @@ Tensor::Tensor(const std::string& name,
         cudaMemset(scale, 0, sizeof(float));
         scale_cpu_data_ = std::make_shared<float>(0);
         tensor_.set_scale(scale, DType::kFloat32, std::vector<size_t>{1});
+
+        // Allocate amax alongside other GPU buffers
```
timmoon10: This is a very Claude-y comment. It's helpful for keeping track of its changes, but it provides no value in the final code. It's obvious we're allocating amax alongside scale, why are we mentioning it specially?
```suggestion
```

The problem with these comments is that the code just becomes littered with historical comments about why each line was added, and it no longer becomes a coherent thing that stands by itself. If I want to learn history I can look at Git, but the really important thing is how the code works _today_.

timmoon10 (dismissed):
LGTM, pending CI. These kernels will be very useful.

timmoon10 (approved):

PR #2871, opened 2026-04-13
[PyTorch] Avoid autograd's gradient accumulation in grouped MLP if possible

ksivaman:
# Description

If `.grad` field of a param is not `None` (many cases such as externally maintained grads for various optimizations, multiple microbatches etc.), pytorch by default accumulates gradients resulting in expensive copy and adds.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring
- [x] Performance

## Changes

- Return `None` gradient when possible in grouped MLP backward.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/ops/basic/grouped_linear.py
+++ b/transformer_engine/pytorch/ops/basic/grouped_linear.py
@@ -1045,7 +1045,7 @@ class GroupedLinear(BasicOperation):
                     grad_weight = torch.stack(grad_weights, dim=0)
             final_weight_grads = [grad_weight]
         else:
-            if delay_wgrad and ctx.weight_requires_grad:
+            if delay_wgrad and ctx.weight_requires_grad and not accumulate_into_main_grad:
                 final_weight_grads = [None] * num_groups
             else:
                 final_weight_grads = grad_weights
--- a/transformer_engine/pytorch/ops/fused/backward_grouped_mlp.py
+++ b/transformer_engine/pytorch/ops/fused/backward_grouped_mlp.py
@@ -155,8 +155,8 @@ def _compute_grad_params(
                 )
             w_list = [packed_wgrad]
         else:
-            if delay_wgrad:
-                w_list = list(w_list) if accumulate_into_main_grad else [None] * num_groups
+            if delay_wgrad or accumulate_into_main_grad:
+                w_list = [None] * num_groups
             if accumulate_into_main_grad:
                 for idx in range(num_groups):
                     wp = getattr(fc_op, f"weight{idx}")


transformer_engine/pytorch/ops/fused/backward_grouped_mlp.py
```diff
@@ -155,8 +155,8 @@ def _compute_grad_params(
                 )
             w_list = [packed_wgrad]
         else:
-            if delay_wgrad:
-                w_list = list(w_list) if accumulate_into_main_grad else [None] * num_groups
+            if delay_wgrad or accumulate_into_main_grad:
+                w_list = [None] * num_groups
```
timmoon10, line 246: Is this correct? I recall we needed to return a dummy wgrad because that affects how grad callbacks are applied, and we got race conditions if we returned `None`.

I'm also wary about only making this change in this particular fused op. If we are changing our contract with Mcore, we should apply it consistently in the linear modules and linear ops.

ksivaman: Yes that's correct and it doesn't change here. In the very next line we have the same logic as before: https://github.com/ksivaman/TransformerEngine-1/blob/f7a28865b1e9e263ae92c6dc97604280093ebc8f/transformer_engine/pytorch/ops/fused/backward_grouped_mlp.py/#L160-L169

timmoon10: ~Doesn't that logic make the `or accumulate_into_main_grad` here redundant?~

ksivaman: We still want to return a `None` gradient unless megatron explicitly sets `grad_added_to_main_grad` attr to the weight, in which case we supply a dummy grad.

timmoon10: It's a bit convoluted, but it seems that the `Linear` module also returns `None` if delayed wgrad is enabled and `fuse_wgrad_accumulation=False`: https://github.com/NVIDIA/TransformerEngine/blob/dc92b3968d0356ddc709e217f7644ecb7e7f752e/transformer_engine/pytorch/module/linear.py#L939-L950

The `GroupedLinear` op has similar behavior with the same combination:
https://github.com/NVIDIA/TransformerEngine/blob/dc92b3968d0356ddc709e217f7644ecb7e7f752e/transformer_engine/pytorch/ops/basic/grouped_linear.py#L1039-L1051

The implementations are all inconsistent when delayed wgrad is enabled and `main_grad` accumulation is enabled (`Linear` module returns dummy tensor, `GroupedLinear` op returns `None`, this fused op returns dummy tensor).

ksivaman: So it is not redundant

timmoon10: Ok, I think this logic matches the `Linear` module. There is an edge case where the unfused `GroupedLinear` op is inconsistent.

| Config | `Linear` module | `GroupedLinear` op | Fused grouped MLP |
|----|----|----|----|
| Vanilla | Return wgrad | Return wgrad | Return wgrad | Return wgrad |
| Delayed wgrad | Return `None` | Return `None` | Return `None` |
| Accumulate into `main_grad`, without `grad_added_to_main_grad` | Return `None` | Return `None` | Return `None` |
| Accumulate into `main_grad` with `grad_added_to_main_grad` | Return dummy | Return dummy | Return dummy |
| Delayed wgrad, accumulate into `main_grad`, without `grad_added_to_main_grad` | Return `None` | Return `None` | Return `None` |
| Delayed wgrad, accumulate into `main_grad` with `grad_added_to_main_grad` | Return dummy | Return `None` | Return dummy |

timmoon10 (dismissed):

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -1045,7 +1045,7 @@ def fuser_backward(
                     grad_weight = torch.stack(grad_weights, dim=0)
             final_weight_grads = [grad_weight]
         else:
-            if delay_wgrad and ctx.weight_requires_grad:
+            if delay_wgrad and ctx.weight_requires_grad and not accumulate_into_main_grad:
```
timmoon10, line 1048: I think this is techically correct behavior, but it's quite hard to understand what everything is doing. In the case where `single_grouped_weight=False` and `accumulate_into_main_grad=True`, then the weight grads have already been set in this if-block:
https://github.com/ksivaman/TransformerEngine-1/blob/e47c7ff6b14c7e24477f2ef84de22c1d463df70a/transformer_engine/pytorch/ops/basic/grouped_linear.py#L1006

It's confusing why we don't have corresponding logic for the `single_grouped_weight=True` case. However, in the previous if-statement, we return early if `accumulate_into_main_grad=True` and `single_grouped_weight=True`.

timmoon10 (approved):
This technically matches the spec and satisfies edge cases, but the logic is a ratnest. The next time we run into trouble, it would be easier to just rewrite it than to add another unmotivated if-statement.

PR #2877, opened 2026-04-14
[PyTorch] Split TE ops op_forward into op_forward and setup_context

pggPL:
## Description

Split the forward pass of fusible operations (`op_forward` / `fuser_forward`) into two separate methods — `compute` and `save_ctx` — to enable `torch.compile` compatibility. Under `torch.compile`, the forward graph is traced without side effects, so separating pure computation from context saving allows the compute portion to be compiled while context saving remains outside the compiled region.

**Type of change:** New feature (non-breaking change which adds functionality)

## Changes

- **Split API for `BasicOperation`**: Added `op_forward_compute` and `op_forward_save_ctx` as an alternative to `op_forward`. Subclasses implement either the legacy single-method API or the new split API (defining both is an error).
- **Split API for `FusibleOperation` / `FusedOperation`**: Added `fuser_forward_compute` and `fuser_forward_save_ctx` as an alternative to `fuser_forward`, with the same auto-detection via `__init_subclass__`.
- **Migrated all basic ops** (`BasicLinear`, `Bias`, `Activation`, `SwiGLU`, `LayerNorm`, `RMSNorm`, `Dropout`, `Quantize`, `AllGather`, `AllReduce`, `ReduceScatter`, `Reshape`, `Identity`, `ConstantScale`, `L2Normalization`) to the split API.
- **Migrated fused ops** (`ForwardLinearBiasAdd`, `ForwardLinearBiasActivation`, `ForwardLinearScaleAdd`, `ForwardGroupedMLP`, `UserbuffersForwardLinear`) to the split API.
- **Fuser dispatch**: Updated `OperationFuser` to detect `_use_split_forward` and call the appropriate path, sharing `fwd_kwargs` between `compute` and `save_ctx`.
- **Type annotation fix**: `tensors_to_save` now correctly typed as `Optional[Union[torch.Tensor, QuantizedTensorStorage]]` instead of `Optional[torch.Tensor]`.
- **Documentation**: Exposed `fuser_forward_compute`, `fuser_forward_save_ctx`, `op_forward_compute`, `op_forward_save_ctx` in `docs/api/pytorch.rst`. Added class-level docstrings documenting the two API choices.

## Checklist

### DIFF
--- a/transformer_engine/pytorch/ops/basic/activation.py
+++ b/transformer_engine/pytorch/ops/basic/activation.py
@@ -80,13 +80,14 @@ class _ActivationOperation(BasicOperation, metaclass=abc.ABCMeta):
 
         """
 
-    def op_forward(
+    def op_forward_compute(
         self,
-        ctx: OperationContext,
         input_: torch.Tensor,
-        prev_op_grad_output_quantizer: Optional[Quantizer],
-        next_op_input_quantizer: Optional[Quantizer],
-    ) -> torch.Tensor:
+        *,
+        requires_grad: bool,
+        prev_op_grad_output_quantizer: Optional[Quantizer] = None,
+        next_op_input_quantizer: Optional[Quantizer] = None,
+    ) -> tuple[torch.Tensor, tuple[Optional[torch.Tensor], ...]]:
 
         # Compute dtype
         dtype: torch.dtype
@@ -109,15 +110,31 @@ class _ActivationOperation(BasicOperation, metaclass=abc.ABCMeta):
             input_quantizer.set_usage(rowwise=True, columnwise=False)
             x = input_quantizer(x)
 
-        # Save state for backward pass
-        if ctx.requires_grad:
-            if is_cpu_offload_enabled():
-                mark_activation_offload(x)
-            ctx.save_for_backward(x)
-            ctx.dtype = dtype
-            ctx.prev_op_grad_output_quantizer = prev_op_grad_output_quantizer
+        if requires_grad:
+            return y, (x,)
+        return y, (None,)
 
-        return y
+    def op_forward_save_ctx(
+        self,
+        ctx: OperationContext,
+        input_: torch.Tensor,
+        tensors_to_save: tuple[Optional[torch.Tensor], ...],
+        *,
+        requires_grad: bool,
+        prev_op_grad_output_quantizer: Optional[Quantizer] = None,
+        next_op_input_quantizer: Optional[Quantizer] = None,
+    ) -> None:
+        if not requires_grad:
+            return
+        (x,) = tensors_to_save
+        if is_cpu_offload_enabled():
+            mark_activation_offload(x)
+        ctx.save_for_backward(x)
+        if torch.is_autocast_enabled():
+            ctx.dtype = torch.get_autocast_dtype("cuda")
+        else:
+            ctx.dtype = input_.dtype
+        ctx.prev_op_grad_output_quantizer = prev_op_grad_output_quantizer
 
     def op_backward(
         self,
--- a/transformer_engine/pytorch/ops/basic/activation.py
+++ b/transformer_engine/pytorch/ops/basic/activation.py
@@ -6,11 +6,12 @@
 
 from __future__ import annotations
 import abc
-from typing import Optional
+from typing import Optional, Union
 
 import torch
 
 import transformer_engine_torch as tex
+from ...quantized_tensor import QuantizedTensorStorage
 from ...cpu_offload import is_cpu_offload_enabled, mark_activation_offload
 from ...tensor.float8_tensor import Float8CurrentScalingQuantizer, Quantizer
 from ...utils import clear_tensor_data
@@ -87,7 +88,7 @@ class _ActivationOperation(BasicOperation, metaclass=abc.ABCMeta):
         requires_grad: bool,
         prev_op_grad_output_quantizer: Optional[Quantizer] = None,
         next_op_input_quantizer: Optional[Quantizer] = None,
-    ) -> tuple[torch.Tensor, tuple[Optional[torch.Tensor], ...]]:
+    ) -> tuple[torch.Tensor, tuple[Optional[Union[torch.Tensor, QuantizedTensorStorage]], ...]]:
 
         # Compute dtype
         dtype: torch.dtype
@@ -118,7 +119,7 @@ class _ActivationOperation(BasicOperation, metaclass=abc.ABCMeta):
         self,
         ctx: OperationContext,
         input_: torch.Tensor,
-        tensors_to_save: tuple[Optional[torch.Tensor], ...],
+        tensors_to_save: tuple[Optional[Union[torch.Tensor, QuantizedTensorStorage]], ...],
         *,
         requires_grad: bool,
         prev_op_grad_output_quantizer: Optional[Quantizer] = None,
--- a/transformer_engine/pytorch/ops/fused/forward_linear_bias_activation.py
+++ b/transformer_engine/pytorch/ops/fused/forward_linear_bias_activation.py
@@ -10,7 +10,6 @@ from typing import Any, Optional
 
 import torch
 
-from ...cpu_offload import is_cpu_offload_enabled, mark_activation_offload
 from ...quantization import FP8GlobalStateManager
 from ...tensor import Quantizer
 from ..basic import BasicLinear, Bias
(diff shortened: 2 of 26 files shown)


transformer_engine/pytorch/ops/basic/activation.py
```diff
@@ -82,11 +84,13 @@ def _activation_backward_impl(self, *args, **kwargs) -> torch.Tensor:
...
         self,
-        ctx: OperationContext,
         input_: torch.Tensor,
+        requires_grad: bool,
         prev_op_grad_output_quantizer: Optional[Quantizer],
         next_op_input_quantizer: Optional[Quantizer],
-    ) -> torch.Tensor:
+        *extra_tensor_inputs: torch.Tensor,
+    ) -> Any:
+        del prev_op_grad_output_quantizer, extra_tensor_inputs
```
timmoon10: I see this `del` idiom throughout this PR, but does it have an effect? The function caller owns a ref to these variables, so there's no hope of triggering early deallocation.

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -128,14 +128,29 @@ def forward(
...
+            op_input = x
+            forward_result = op.fuser_forward(
                 [basic_op_ctxs[idx] for idx in basic_op_idxs],
                 x,
                 basic_op_extra_inputs=extra_inputs,
                 prev_op_grad_output_quantizer=prev_op_grad_output_quantizer,
                 next_op_input_quantizer=next_op_input_quantizer,
                 basic_op_kwargs=[basic_op_kwargs[idx] for idx in basic_op_idxs],
             )
+            if getattr(type(op), "separate_fuser_setup_context", False):
```
timmoon10: Since this is defined in the base class, we could simplify and avoid some `getattr` overhead:

```suggestion
            if op.separate_fuser_setup_context:
```

transformer_engine/pytorch/ops/op.py
```diff
@@ -409,35 +465,36 @@ def _load_fp8_metas(self, fp8_metas: Optional[dict[str, Any]]) -> None:
                 self._fp8_metas[mode][fp8_meta_key].scale.copy_(scale)
                 self._fp8_metas[mode][fp8_meta_key].amax_history.copy_(amax_history)
 
-    @abc.abstractmethod
-    def op_forward(
+    def op_forward(self, *args: Any, **kwargs: Any) -> Any:
```
timmoon10: This is trying to use the same function name to support two different behaviors, with two different APIs. This will be hard to reason about and there will be many bugs from using the wrong API. It seems more straightforward to implement a new function:
```python
def op_forward(self, ctx, ...):
    """Forward pass"""

    # Some advanced ops may override this default
    self.op_forward_compute(...)
    self.op_forward_setup_context(ctx)

def op_forward_compute(self,  ...):
    """Forward pass compute"""
    ...

def op_forward_setup_context(self, ctx):
    """Setup context for backward pass"""
    ...

```

transformer_engine/pytorch/ops/op.py
```diff
@@ -707,6 +784,35 @@ def __init__(
...
         return True
 
+    #: When ``True``, :meth:`fuser_forward` must return
+    #: ``(pipeline_output, fused_op_extra_outputs, setup_tensors)``.
+    #: The operation fuser calls :meth:`fuser_setup_context` afterward
+    #: with the same ``basic_op_*`` arguments as ``fuser_forward`` plus
+    #: ``setup_tensors`` (a :obj:`FusedOpSetupTensors`, or ``None``).
+    separate_fuser_setup_context: ClassVar[bool] = False
+
+    def fuser_setup_context(
```
timmoon10: Right now this requires that the context state is a kwarg, lives in the op, or is a tensor. However, what if we want to save other context state? For example, maybe one of the saved tensors could have multiple possible formats and we need an enum or something to keep track (e.g. what if grouped linear op could save either split sizes or split offsets). My first thought is to allow `op_forward` to pass a dict or tuple to `setup_context`, although I'm not sure how well this would work with `torch.compile`.
```python
x, fused_op_extra_outputs, setup_tensors, context_state = op.fuser_forward(...)
op.fuser_setup_context(ctxs, ..., setup_tensors=setup_tensors, context_state=context_state)
```
If we're lucky, could we get rid of `setup_tensors` entirely and just store the tensors in a dict?

transformer_engine/pytorch/ops/op.py
```diff
@@ -48,12 +49,40 @@ class OperationContext:
...
+# Saved linear activation / weight between :meth:`FusedOperation.fuser_forward` and
+# :meth:`FusedOperation.fuser_setup_context`: ``torch.Tensor`` (including
+# :class:`~transformer_engine.pytorch.quantized_tensor.QuantizedTensor` subclasses) or
+# :class:`~transformer_engine.pytorch.quantized_tensor.QuantizedTensorStorage` carriers
+# (Float8, MXFP8, NVFP4, blockwise, …). Same objects as for
+# :meth:`OperationContext.save_for_backward`, not raw ``torch.Storage``.
+FusedOpSavedSlot: TypeAlias = Optional[Union[torch.Tensor, QuantizedTensorStorage]]
+
+# ``None`` if no linear backward state; otherwise ``(input-side, weight-side)`` saves.
+FusedOpSetupTensors: TypeAlias = Optional[tuple[FusedOpSavedSlot, FusedOpSavedSlot]]
```
timmoon10: This is specific to linear layers. Grouped linear layers are the important thing now and they may require caching multiple weight tensors. I wonder what will be the big thing in two years.

If we need type hints for `torch.compile`, could we do:
```python
FusedOpSetupTensors: TypeAlias = Sequence[FusedOpSavedSlot]
# Alternatively tuple[FusedOpSavedSlot, ...]
```

transformer_engine/pytorch/ops/fused/forward_linear_bias_activation.py
```diff
@@ -121,34 +126,58 @@ def fuser_forward(
...
+        saved_input, saved_weight = setup_tensors
+        with_quantized_compute = FP8GlobalStateManager.is_fp8_enabled()
+        if with_quantized_compute:
+            backward_override = FP8GlobalStateManager.get_fp8_recipe().backward_override
+        else:
+            backward_override = None
+        lin_idx = self._op_idxs["linear"]
+        linear_op = self.basic_ops[lin_idx]
+        linear_op_ctx = basic_op_ctxs[lin_idx]
+        linear_op.setup_context(
```
timmoon10: It is quite nice that the fused ops can reuse logic from the unfused ops.

transformer_engine/pytorch/ops/op.py
pggPL: That's good point.

transformer_engine/pytorch/ops/op.py
pggPL: This will not work with `torch.compile` and this is main motivation of that change. If we could do this, we could just return ctx and pass it into the setup_context.

There is a purpose in that. Setup context is run once with fake tensors (not once per iteration), so that the compiler can trace which tensors need to be saved and is able to link outputs from forward to inputs for backward. This helps compiler to decide which fusions to perform and where apply recomputation.

So basically if we are unable to split setup context and forward, we should use old API and it will result in graph break.

pggPL:
@timmoon10 I've cleaned the code (there was some ai slop i haven't looked into) and opened the PR, I think it is fine now.

transformer_engine/pytorch/ops/op.py
```diff
@@ -124,6 +162,97 @@ def fuser_forward(
...
+        *,
+        requires_grad: list[bool],
+        basic_op_extra_inputs: list[tuple[torch.Tensor, ...]],
+        prev_op_grad_output_quantizer: Optional[Quantizer],
+        next_op_input_quantizer: Optional[Quantizer],
+        basic_op_kwargs: list[dict[str, Any]],
+    ) -> tuple[
+        torch.Tensor,
+        Iterable[Iterable[torch.Tensor]],
+        list[tuple[Optional[Union[torch.Tensor, QuantizedTensorStorage]], ...]],
```
timmoon10, line 177: Nit: This module does `from __future__ import annotations`, so it's safe to replace `Union` with:
```suggestion
        list[tuple[Optional[torch.Tensor | QuantizedTensorStorage], ...]],
```

transformer_engine/pytorch/ops/op.py
```diff
@@ -490,6 +743,53 @@ def fuser_forward(
...
+        tensors_to_save: list[tuple[Optional[Union[torch.Tensor, QuantizedTensorStorage]], ...]],
+        *,
+        requires_grad: list[bool],
+        basic_op_extra_inputs: list[tuple[torch.Tensor, ...]],
+        prev_op_grad_output_quantizer: Optional[Quantizer],
+        next_op_input_quantizer: Optional[Quantizer],
+        basic_op_kwargs: list[dict[str, Any]],
+    ) -> None:
+        self._check_no_extra_io("fuser_forward_save_ctx", "op_forward_save_ctx")
+        self.op_forward_save_ctx(
```
timmoon10, line 783: Many of the ops have logic to skip if `requires_grad=False`. We could move that check here, so there's no need to pass `requires_grad` into `op_forward_save_ctx`.

transformer_engine/pytorch/ops/basic/activation.py
```diff
@@ -109,15 +111,31 @@ def op_forward(
...
-        # Save state for backward pass
-        if ctx.requires_grad:
-            if is_cpu_offload_enabled():
-                mark_activation_offload(x)
-            ctx.save_for_backward(x)
-            ctx.dtype = dtype
-            ctx.prev_op_grad_output_quantizer = prev_op_grad_output_quantizer
+        if requires_grad:
+            return y, (x,)
+        return y, (None,)
```
timmoon10, line 116: This is fine, but I want to make sure this check is optional. It should be fine to just return y, (x,) and to ignore x when it's not needed.

If it is a requirement to return `None`s when `requires_grad=False`, then we should consider handling that in `fuser_forward_compute`.

timmoon10 (commented):
Overall LGTM, with some minor questions about making the interfaces a little nicer.

pggPL:
Closing this PR — my other torch.compile PRs (especially the TensorProto mechanism in #3153) have forced a rethink of how the ops fuser should expose compute vs. context-saving, so this split needs to be redesigned on top of that work (and the recent fuser refactors #3080/#3117). I'll re-propose it as a fresh PR. @timmoon10 thanks for the review — I agree with the interface suggestions and will incorporate them in the new version.

PR #2881, opened 2026-04-15
Check numerics in MXFP8 C++ tests by dequantizing to FP32

timmoon10:
# Description

We have experienced some failures in the MXFP8 C++ tests because of expected numerical errors. In particular, FP32 numerical errors will sometimes cause the MXFP8 scales to differ by one. However, our current tests assume that MXFP8 scales are bit-wise exact between the TE implementation and reference implementation. This PR changes the MXFP8 tests so that we dequantize MXFP8 values and compare against an FP32 reference implementation.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add MXFP8 dequantization function to test utilities
- Change FP8 tolerances to match machine epsilon
- Change MXFP8 tests to use FP32 reference implementation
- Change MXFP8 tests to check that dequantized MXFP8 values and FP32 reference values are within FP8 tolerances

# Checklist:

### DIFF
--- a/tests/cpp/operator/test_cast_mxfp8.cu
+++ b/tests/cpp/operator/test_cast_mxfp8.cu
@@ -36,121 +36,50 @@ enum ActivationType {
     SReLU
 };
 
-template <typename InputType, typename OutputType>
+// Compute the pre-quantization float values that the GPU kernel processes before
+// casting to FP8. These are used as the reference for dequantized output comparison.
+template <typename InputType>
 void compute_ref(const ProcessingMethod processing_method,
                  float (*OP)(const float),
-                 const bool rowwise,
-                 const bool colwise,
                  const InputType* input,
                  const InputType* grad,
-                 OutputType* output_rowwise,
-                 OutputType* output_colwise,
-                 fp8e8m0* output_scales_rowwise,
-                 fp8e8m0* output_scales_colwise,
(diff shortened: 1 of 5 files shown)


ptrendx (changes requested):
Blocking before reviewing as I don't think I agree with the principle here  - will try to review today.

ptrendx:
Yeah, I don't agree with this approach. There are 2 problems here:
- this basically requires us to check everything against the output type tolerances, which in the case of FP8 are big
- this doesn't actually lets us test the exact logic of the quantization. For example, if we changed the way we compute the scaling factor as a result of a mistake, this test would not catch it.

In general I am in favor of trying to make the reference match closely with what the kernel is supposed to do and check that. This obviously has issues when the input value lands very close to the limit between the 2 scaling factors, resulting in the errors like this but I would prefer us taking the approach of choosing the input so that it behaves reasonably well rather than stopping testing the actual behavior of the function.

timmoon10:
Ok, makes sense. I agree that it's important to test that the direct MXFP8 casts are bit-wise exact. The next time we attempt this, we should only allow numerical errors in the fused activation kernels, and only when the scaling factor differs by one. We should also get rid of our existing approach of allowing a certain number of numerical failures, since that's also more permissive than necessary.

PR #2885, opened 2026-04-15
Optimized Bias/Dbias Support for GroupedLinear

vthumbe1503:
# Description

- Scaled Bias Addition to GroupedTensor needed after CUBLAS GGEMM is integrated in Pytorch.
- Triton Kernel for Dbias computation and its integration in GroupedLinear module

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Grouped Scaled Bias Add Kernel:
   - CUBLAS GGEMM doesnt support scaled bias addition at the end of grouped gemm. This PR introduces the scaled variant of bias-add kernel and optimizes it close to SOL.
   - Integrated at the Pytorch Level by adding optional arguments for bias_scale in grouped_gemm API.
   - Perf for Grouped Scaled Bias Add Kernel
image<br>

 - Triton Kernel for dbias_dscales refactored to also support dbias only use-case
   - We need grouped dbias computation for BF16. For Quantized tensors grouped dbias is already fused as part of the grouped quantize
   - This kernel is integrated in GroupedLinear module.

# Checklist:


transformer_engine/pytorch/csrc/extensions/gemm.cpp
```diff
@@ -610,8 +610,8 @@ std::optional<std::vector<at::Tensor>> te_general_grouped_gemm(
 
 py::object te_general_grouped_gemm_for_grouped_tensor(
     py::handle A, bool transa, py::handle B, bool transb, py::handle D, py::object bias,
-    at::Tensor alpha, at::Tensor beta, at::Tensor workspace_setup, at::Tensor workspace_cublas,
-    bool use_split_accumulator, int math_sm_count) {
+    at::Tensor bias_scale, at::Tensor alpha, at::Tensor beta, at::Tensor workspace_setup,
```
timmoon10, line 613: We should avoid the overhead of constructing a tensor when `bias_scale` isn't needed. `std::optional` also communicates the intent more clearly.
```suggestion
    std::optional<at::Tensor> bias_scale, at::Tensor alpha, at::Tensor beta, at::Tensor workspace_setup,
```

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -394,13 +398,17 @@ def general_grouped_gemm_for_grouped_tensor(
     sm_count = get_sm_count()
     sm_count = sm_count - int(os.getenv("NVTE_EXT_MARGIN_SM", str(sm_count)))
 
+    if bias_scale is None:
+        bias_scale = torch.empty(0, dtype=torch.float32, device=device)
+
```
timmoon10, line 403: We can avoid this overhead by making the `tex` function take an optional argument. 
```suggestion
```

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -429,12 +429,14 @@ void nvte_grouped_gemm_with_discrete_out(const NVTEGroupedTensor A, int transa,
...
-/*! \brief Grouped bias add for grouped GEMM outputs.
+/*! \brief Grouped Bias add for grouped GEMM outputs.
 *
+* When \p scale is a valid tensor: output[row,col] += bias[col] * scale[row],
+* When \p scale is empty/null: output[row,col] += bias[col].
 * Requires uniform last-dimension across all output tensors and bias tensors.
 */
 void nvte_grouped_bias_add(const NVTEGroupedTensor output, const NVTEGroupedTensor bias,
-                           cudaStream_t stream);
+                           const NVTETensor scale, cudaStream_t stream);
```
timmoon10, line 439: I think it makes more sense to create a separate API for `nvte_grouped_scaled_bias_add`. Grouped bias is a natural generalization of linear layer biases, but grouped scaled bias is less intuitive (especially that the biases are per-group, but the scales are per-token) and it should be treated as more exotic.

tests/pytorch/test_numerics.py
```diff
@@ -8,14 +8,14 @@
...
-from torch.nn import Parameter
-
 from transformer_engine.pytorch.quantization import (
     FP8GlobalStateManager,
     get_align_size_for_quantization,
 )
+import torch
+import torch.nn as nn
+from torch.nn import Parameter
+
```
timmoon10, line 18: Nit: Is there a reason we're reordering? If the import order causes problems, then that's a bug we need to fix. Otherwise, this ordering seems strangely unmotivated and haphazard. It's also considered good Python style to put third party imports before local imports ([PEP 8](https://peps.python.org/pep-0008/#imports)).

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -845,43 +846,102 @@ __forceinline__ __device__ int64_t compute_grouped_tensor_offset(const TensorSha
...
+                                        const float *__restrict__ scale_base,
+                                        TensorShapeInfo d_meta, int n, int total_rows,
+                                        int num_tensors) {
+  using VecStorage = transformer_engine::VectorizedStorage<T, kVec>;
+  using VecType = typename VecStorage::LType;
 
-  const int64_t m = d_meta.first_dims ? d_meta.first_dims[tensor_idx] : d_meta.uniform_first;
-  const int64_t n = d_meta.last_dims ? d_meta.last_dims[tensor_idx] : d_meta.uniform_last;
+  constexpr int kMaxTensors = 257;
+  __shared__ int cumsum[kMaxTensors];
```
timmoon10, line 863: The variable name is wrong.
```suggestion
  constexpr int kMaxTensors = 256;
  __shared__ int cumsum[kMaxTensors + 1];
```

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -845,43 +846,102 @@ __forceinline__ __device__ int64_t compute_grouped_tensor_offset(const TensorSha
...
+    while (lo < hi) {
+      int mid = (lo + hi) >> 1;
+      if (cumsum[mid + 1] <= row_start)
+        lo = mid + 1;
+      else
+        hi = mid;
+    }
+    tensor_idx = lo;
+  }
+  int bias_idx = tensor_idx * n;
```
timmoon10, line 906: Have we benchmarked whether this binary search is any better than just scanning through the tensors. Computing the cumsums is still O(n), so we're not improving the asymptotics. We're also introducing thread syncs and shared memory accesses.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -845,43 +846,102 @@ __forceinline__ __device__ int64_t compute_grouped_tensor_offset(const TensorSha
...
-  transformer_engine::VectorizedStorer<T, kVec, true> storer(d_ptr, elements);
-  const int64_t vec_id = static_cast<int64_t>(blockIdx.y) * blockDim.x + threadIdx.x;
-  if (vec_id >= vec_count) return;
-  const int64_t vec_start = vec_id * kVec;
-  const int64_t col = vec_start % n;
-  loader.load(vec_id, elements);
-  const auto *b_vec = reinterpret_cast<const VecType *>(bias_ptr + col);
   VecStorage b_in;
-  b_in.scratch_.aligned = *b_vec;
+  b_in.scratch_.aligned = *reinterpret_cast<const VecType *>(bias + bias_idx + col);
```
timmoon10, line 909: This value is immediately wiped out in the loop. I guess the compiler might be smart enough not to do an unnecessary memory access, but it makes the code harder to read.

timmoon10 (changes requested):
My big question is whether the kernel implementation changes are providing a perf benefit.

transformer_engine/pytorch/csrc/extensions/gemm.cpp
vthumbe1503: Yes, agreed. Done now.

transformer_engine/pytorch/cpp_extensions/gemm.py
vthumbe1503: Done

transformer_engine/common/include/transformer_engine/gemm.h
vthumbe1503: Agreed. done

tests/pytorch/test_numerics.py
vthumbe1503: It was a container problem. And I agree that its ant-PEP standards. Reverted the change

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
vthumbe1503: cumsum is not the bottleneck in this kernel inspite of the syncthreads since this is completely memory bound by the grouped tensor load.
As part performance improvement cycle, I was doing this. Since you can cumsum once in a grid strided loop. And reuse the cumsum for binary search as well as finding the valid rows in the grouped tensor.(When sum(first_dims) < total_rows of grouped tensor)

Through my performance testing I have found that it doesnt help perf much and so I have removed it and made it simpler.

I should have marked the PR as draft since I was still going through the performance improvement cycle when you reviewed the PR.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
vthumbe1503: One of the main optimization in this kernel is to load the bias once per group of rows in the input tensor. Since generally speaking contigous group of rows would ideally belong to same or atleast contigous group_ids. And input rows having same group_ids would share the bias tensor.

vthumbe1503: But the code is cleaned up now.

vthumbe1503:
> My big question is whether the kernel implementation changes are providing a perf benefit.

@timmoon10 there has been a quite significant improvement to the perf of this kernel. This kernel is going to be used as part of GroupedLinear integration which would need the cublas grouped gemm along with scaled bias integration.Fused Grouped MLP hadnt been using it since it only uses cublas for wgrad which doesnt need the bias.

Main Perf Limitations of the kernel before
- It dependent on an implementation that significantly overlaunches thread blocks(each group in the tensor was using all the rows across all groups to determine the number of blocks)
- Number of blocks was so high. It could even exceed 65536 limit of a cuda kernel for some problem sizes
- This implementation also didnt optimize for the case when group sizes can be widely different
- Also extremely suboptimal for the cases where sum(first_dims) << total_rows for the grouped tensor.

Current implementation optimizes across a lot of dimensions
- We determine the launch configuration through number of sms and have grid strided loop inside
- The implementation load balances the grouped tensor across CTAs in terms of first_dims. Each CTA can potentially handle multiple groups if some of the group sizes are too small.
- Implementation also makes sure we calculate the sum(first_dims) using a warp-shuffle based reduction. And uses that to dynamically create mapping between CTA and block of rows. Instead of depending on the total number of rows for CTA to block mapping.
- Bias Tensor is loaded once per group of rows if they share the same group_id. If group_id changes as we iterate along the rows. Then the new apt bias tensor is reloaded.


Without optimizations the perf would hit atmost 1TB/s and in the worst case was even 0.1TB/s. With this PR, we are hitting close to 6TB/s for all the scenarios(including worst-case scenarios like group imbalance) which is already 70-80% of the bandwidth of B100/GB200

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -845,43 +845,69 @@ __forceinline__ __device__ int64_t compute_grouped_tensor_offset(const TensorSha
...
+                                        const float *scale_base, TensorShapeInfo d_meta,
                                         TensorShapeInfo bias_meta, size_t num_tensors) {
-  const size_t tensor_idx = blockIdx.x;
+  using VecStorage = transformer_engine::VectorizedStorage<T, kVec>;
+  using VecType = typename VecStorage::LType;
+
+  const size_t tensor_idx = blockIdx.y;
   if (tensor_idx >= num_tensors) return;
 
+  const int64_t n = d_meta.last_dims ? d_meta.last_dims[0] : d_meta.uniform_last;
```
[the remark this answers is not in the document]
vthumbe1503: Not relevant anymore.

transformer_engine/pytorch/csrc/extensions/swizzle.cpp
```diff
@@ -341,7 +341,8 @@ at::Tensor convert_block_scaling_to_mxfp8_tensor(transformer_engine::TensorWrapp
 std::optional<SwizzledGroupedScales> maybe_swizzle_grouped_tensor(GroupedTensorWrapper &input,
                                                                   bool rowwise_usage,
                                                                   bool columnwise_usage) {
-  if (input.scaling_mode() != NVTE_MXFP8_1D_SCALING) {
+  if (input.scaling_mode() != NVTE_MXFP8_1D_SCALING &&
+      input.scaling_mode() != NVTE_NVFP4_1D_SCALING) {
     return std::nullopt;
   }
```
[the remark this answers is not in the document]
vthumbe1503: Removed this change. SInce it is out of scrope for this PR.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
vthumbe1503: Now we are using kMaxGroups common variable which is shared name used for both cublas grouped gemm helpers and the grouped_bias_add_kernel helpers

transformer_engine/pytorch/triton/grouped_dbias_dscales.py
```diff
@@ -2,16 +2,76 @@
...
+    specialization; otherwise runs the fused dbias+dscales specialization
+    (all three must be provided together).
+    """
+    if _is_deterministic_mode():
+        raise RuntimeError(
+            "grouped_dbias Triton kernel uses non-deterministic atomic adds "
+            "and cannot be used when deterministic execution is requested "
+            "(NVTE_ALLOW_NONDETERMINISTIC_ALGO=0). "
+            "Disable determinism or use a deterministic fallback."
+        )
```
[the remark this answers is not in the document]
vthumbe1503: But that would require us to break cuda graph. The intent of this triton kernel is to do dbias computation of when m_splits for Grouped GEMM is on the device. We cant simply do a torch.sum. So as of now it is fine to throw error in such a case.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -1397,6 +1542,14 @@ void nvte_grouped_bias_add(const NVTEGroupedTensor output, const NVTEGroupedTens
              CUBLAS_VERSION, ". Please upgrade to cuBLAS 13.3 (shipped with CUDA 13.2) or newer.");
 }
 
+void nvte_grouped_scaled_bias_add(const NVTEGroupedTensor output, const NVTEGroupedTensor bias,
+                                  const NVTETensor scale, cudaStream_t stream) {
+  NVTE_ERROR(
+      "nvte_grouped_scaled_bias_add requires cuBLAS 13.3+, but compile-time cuBLAS version "
+      "is ",
+      CUBLAS_VERSION, ". Please upgrade to cuBLAS 13.3 (shipped with CUDA 13.2) or newer.");
```
timmoon10, line 1550: Not blocking, but maybe something for a future refactor: this check doesn't make sense here (or in `nvte_grouped_bias_add`) since the kernel implementation doesn't rely on cuBLAS. In fact, it's a sign that we should move these kernels to a separate source file that isn't cuBLAS-specific.

timmoon10 (approved):
LGTM, those perf results are quite nice.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
vthumbe1503: Agreed makes sense.

PR #2893, opened 2026-04-16
[PyTorch] Add test to compare single vs multi-param fused GMLP

ksivaman:
# Description

Adds a new test that compares bitwise identical outputs and grads for fused grouped MLP for the contiguous and discrete parameter case.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring
- [x] Testing

## Changes

- Adds a new test that compares bitwise identical outputs and grads for fused grouped MLP for the contiguous and discrete parameter case.

# Checklist:

### DIFF
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -3669,6 +3669,158 @@ class TestSequentialModules:
             assert_close(fc1.weight.grad, fc1_w_ref_grad, **tols)
             assert_close(fc2.weight.grad, fc2_w_ref_grad, **tols)
 
+    @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


tests/pytorch/test_fusible_ops.py
timmoon10, line 1: If we wanted the test to be even more strict, instead of testing the single-param vs multi-param case against each other, we could test both against a CPU FP64 impl. This would involve using `make_reference_and_test_tensors` to generate values that are representable in MXFP8 (or NVFP4). This approach would catch cases where the single-param and multi-param impls are both wrong in the same way.

That said, this test will do the job.

timmoon10 (approved):
LGTM

PR #2894, opened 2026-04-16
Improve the dimension checks for the FP8 recipes

ptrendx:
# Description

Improved the dimension checks for the FP8 recipes. The old one was too restrictive and did not have any recipe-aware logic.

## Dimension check changes

  | Layer | Recipe | Before | After |
  |---|---|---|---|
  | Python `assert_dim_for_fp8_exec` (`utils.py`) | all FP8/FP4 | `M % 8 == 0` ∧ `last_dim % 16 == 0`, run before the recipe is known | **Removed** — guard was recipe-agnostic and stricter than any underlying kernel required |
  | C++ `MXFP8Quantizer::create_tensor` / `get_scale_shape` | MXFP8 | `first_dim % 32` ∧ `last_dim % 32` | `last_dim % 16` only (TMA 16 B alignment). Scale tensor uses `DIVUP` for partial trailing blocks |
  | C++ `NVFP4Quantizer::create_tensor` / `get_scale_shape` | NVFP4 | `first_dim % 32` ∧ `last_dim % 32` | `last_dim % 32` always; `first_dim % 16` only when RHT is enabled (Hadamard block size) |
  | C++ `swizzle_block_scaling.cu` kernel | Float8 1×128 block | `data_rows % 4 == 0` | Unchanged — real kernel requirement; error message clarified with actual `data_rows` |
  | C++ `cublaslt_gemm.cu` MXFP8 path | MXFP8 | No explicit check — cuBLAS returns opaque CUBLAS_STATUS_NOT_SUPPORTED` / "no algo found" | `lda % 16 == 0` ∧ `ldb % 16 == 0` via  `min_alignment_elements(dtype)` using `typeToNumBits` (16 B alignment for FP8) |
  | C++ `cublaslt_gemm.cu` NVFP4 path | NVFP4 | No explicit check — opaque cuBLAS error | `lda % 32 == 0` ∧ `ldb % 32 == 0` (16 B alignment for FP4 = 32 elements) |
  | C++ `cublaslt_gemm.cu` 1D block-scaled FP8 path | Float8BlockScaling 1×128 | No explicit check | `n % 8 == 0` for the native 1D block-scaling path |
  | Python `is_quantizable` → `supports_quantized_allgather` (`mxfp8_tensor.py`) | MXFP8 | Both-dim `% 32` | Renamed. `last_dim % 16`; also `first_dim % 32` when columnwise usage (block boundary across ranks) |
  | Python `is_quantizable` → `supports_quantized_allgather` (`nvfp4_tensor.py`) | NVFP4 | Both-dim `% 32` | Renamed. `last_dim % 32`; also `first_dim % 16` when columnwise usage |

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -363,11 +363,14 @@ def supports_only_rowwise_all_gather(self) -> bool:
...
+    # pylint: disable-next=unused-argument
+    def supports_quantized_allgather(self, inp: torch.Tensor) -> bool:
+        """Whether tensor shape supports quantized all-gather.
+
+        When False, the distributed all-gather falls back to gathering
+        in high precision and quantizing afterward. This is needed when
+        the local shard's shape would cause scaling factor blocks to
+        span across GPU boundaries.
         """
         return True
```
timmoon10, line 375: - We shouldn't assume that quantized tensors use a block-scale format.
- While we're touching this function, changing the default to `False` is more general. It will always work, even without a custom all-gather impl.
```suggestion
    def supports_quantized_allgather(self, inp: torch.Tensor) -> bool:
        """Whether tensor shape supports quantized all-gather.

        When False, the distributed all-gather falls back to gathering
        in high precision and quantizing afterward.
        """
        return False
```

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -303,16 +308,9 @@ def make_empty(
...
-        assert shape[-1] % NVFP4_BLOCK_SCALING_SIZE == 0, (
-            f"Incorrect shape {shape} for NVFP4. Tensor dims must be divisible by"
-            f" {NVFP4_BLOCK_SCALING_SIZE}"
-        )
-
-        flat_first_dim = math.prod(shape[:-1])
-        assert flat_first_dim % NVFP4_BLOCK_SCALING_SIZE == 0, (
-            f"Incorrect shape {shape} for NVFP4. Tensor dims must be divisible by"
-            f" {NVFP4_BLOCK_SCALING_SIZE}"
-        )
```
timmoon10, line 315: It would be safer to keep this check if allocating column-wise data.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -86,13 +86,18 @@ def quantize_impl(self, tensor: torch.Tensor) -> QuantizedTensor:
...
+
+        For distributed all-gather with columnwise scaling, the first
+        dimension must be aligned to the block size so that no scaling
+        block spans across GPU boundaries.
+        """
         if inp.ndim < 2:
             return False
-        if inp.shape[-1] % MXFP8_BLOCK_SCALING_SIZE != 0:
+        if inp.shape[-1] % 16 != 0:
             return False
```
timmoon10, line 99: Wouldn't partial blocks break all-gather? If we wanted to be as permissive as possible, we could reintroduce the check if row-wise usage is requested:
```suggestion
        if inp.shape[-1] % 16 != 0:
            return False
        if self.rowwise_usage and inp.shape[-1] % MXFP8_BLOCK_SCALING_SIZE != 0:
            return False
```

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -111,21 +116,19 @@ def make_empty(
...
+        # trailing blocks (allowed by the relaxed last_dim % 16 constraint) get
+        # one scale each, matching the C++ quantizer (DIVUP in quantizer.cpp).
         data = None
         scale_inv = None
         if self.rowwise_usage:
             data = torch.empty(shape, dtype=torch.uint8, device=device, pin_memory=pin_memory)
             scale_inv = torch.empty(
                 round_up_to_nearest_multiple(math.prod(shape[:-1]), 128),
-                round_up_to_nearest_multiple(shape[-1] // MXFP8_BLOCK_SCALING_SIZE, 4),
+                round_up_to_nearest_multiple(math.ceil(shape[-1] / MXFP8_BLOCK_SCALING_SIZE), 4),
```
timmoon10, line 131: Nit: I'd prefer going through a ceildiv utility function. It's not like FP64's 52 mantissa bits aren't enough, but I have residual superstition about floating point error.

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -92,6 +92,15 @@ struct GemmParam {
...
 };
 
+// Minimum number of elements for 16-byte alignment, given a data type.
+// cuBLAS requires (dim * typeSize) % 16 == 0 for FP8 tensor core usage,
+// i.e. dim % (128 / typeBits) == 0.
+constexpr size_t kAlignmentBytes = 16;
+
+size_t min_alignment_elements(transformer_engine::DType dtype) {
+  return kAlignmentBytes * 8 / transformer_engine::typeToNumBits(dtype);
+}
```
timmoon10, line 102: The comment is misplaced. Also, the `min_` is kind of redundant.
```suggestion
constexpr size_t kAlignmentBytes = 16;

// Number of elements for 16-byte alignment, given a data type.
// cuBLAS requires (dim * typeSize) % 16 == 0 for FP8 tensor core usage,
// i.e. dim % (128 / typeBits) == 0.
size_t alignment_elements(transformer_engine::DType dtype) {
  return kAlignmentBytes * 8 / transformer_engine::typeToNumBits(dtype);
}
```

tests/pytorch/test_numerics.py
```diff
@@ -1271,6 +1272,168 @@ def test_linear_accuracy(dtype, bs, model, return_bias, bias):
...
+    raise ValueError(f"Unknown recipe: {recipe_obj.__class__.__name__}")
+
+
+@pytest.mark.parametrize("recipe", fp8_recipes)
+@pytest.mark.parametrize("inference", [True, False], ids=["inference", "training"])
+@pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float32], ids=["bf16", "fp32"])
+def test_linear_tight_dims(recipe, inference, dtype):
+    """te.Linear with the tightest M/N/K per recipe, vs a pytorch baseline.
+
+    Previously the Python assert_dim_for_fp8_exec rejected any M not divisible
```
timmoon10, line 1305: These historical comments are good for checking what Claude did, but they're useless and distracting if they end up in the final code. `assert_dim_for_fp8_exec` is gone and mentioning it doesn't help you understand the code as it is.

tests/pytorch/test_numerics.py
```diff
@@ -1271,6 +1272,168 @@ def test_linear_accuracy(dtype, bs, model, return_bias, bias):
...
+    # tensor cores would drop 13 mantissa bits and diverge far beyond the FP32
+    # tolerance we want to check.
+    prev_tf32_matmul = torch.backends.cuda.matmul.allow_tf32
+    prev_tf32_cudnn = torch.backends.cudnn.allow_tf32
+    if dtype == torch.float32:
+        torch.backends.cuda.matmul.allow_tf32 = False
+        torch.backends.cudnn.allow_tf32 = False
+    try:
+        te_linear = Linear(K, N, bias=False, params_dtype=dtype, device=device)
+        torch_linear = torch.nn.Linear(K, N, bias=False, device=device, dtype=dtype)
```
timmoon10, line 1342: The real baseline isn't PyTorch, it's mathematical ground truth. FP64 compute helps us get as close as possible:
```suggestion
        torch_linear = torch.nn.Linear(K, N, bias=False, device="cpu", dtype=torch.float64)
```
This also lets us get rid of the TF32 logic.

tests/pytorch/test_numerics.py
```diff
@@ -1271,6 +1272,168 @@ def test_linear_accuracy(dtype, bs, model, return_bias, bias):
...
+        # Forward-path quantizers: [input, weight, output].
+        fwd_state = RecipeState.create(recipe, mode="forward", num_quantizers=3)
+        input_quantizer, weight_quantizer, _ = fwd_state.make_quantizers()
+
+        # Share weights: TE gets the raw weight (it quantizes internally); the
+        # baseline gets dequantize(quantize(W)) so both do the same matmul.
+        W = torch.randn(N, K, dtype=dtype, device=device)
+        with torch.no_grad():
+            te_linear.weight.copy_(W)
+            torch_linear.weight.copy_(weight_quantizer(W).dequantize())
```
timmoon10, line 1353: Nit: It's a bit odd that the TE and reference impls don't actually share the same weight values.
```suggestion
        # Share weights
        W = torch.randn(N, K, dtype=dtype, device=device)
        W = weight_quantizers(W).dequantize()
        with torch.no_grad():
            te_linear.weight.copy_(W)
            torch_linear.weight.copy_(W)
```

tests/pytorch/test_numerics.py
```diff
@@ -1271,6 +1272,168 @@ def test_linear_accuracy(dtype, bs, model, return_bias, bias):
...
+            torch.cuda.synchronize()
+
+            torch.testing.assert_close(x_te.grad, x_ref.grad, **tols)
+            torch.testing.assert_close(te_linear.weight.grad, torch_linear.weight.grad, **tols)
+    finally:
+        torch.backends.cuda.matmul.allow_tf32 = prev_tf32_matmul
+        torch.backends.cudnn.allow_tf32 = prev_tf32_cudnn
+
+
+def test_mxfp8_scale_shape_partial_block():
```
timmoon10, line 1412: This doesn't test numerics, and generally `test_quantized_tensor.py` is a more natural place for basic quantization tests.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
ptrendx: I don't think so - in the end the MXFP8 tensors are always rowwise in memory so the last dimension does not matter as it is not the dimension that is being allgathered. The dimension that is checked is the first dimension when the columnwise usage is true.

tests/pytorch/test_numerics.py
ptrendx: Missed that during reviewing of the generated code. Will clean.

tests/pytorch/test_numerics.py
ptrendx: I know you like that approach, but I generally disagree - while you are correct that the FP64 would give us the best estimate of the real values that "should" be there, it doesn't necessarily make the best reference. This is because in the case of the quantized execution we know that we deviate from the ground truth (due to various factors: quantization, TF32, etc.) and using the reference that is correct but closer to the actual computation being done helps with tightening the tolerances. And those ultimately provide better confidence in the result being correct.

transformer_engine/pytorch/tensor/nvfp4_tensor.py
ptrendx: Actually, not sure if this check is needed at all (even the 32 elements on the rowwise side) that was left there. This check comes from the GEMM and so it is checked there. There is also the check in the allgather support function. Here we should just check if the quantization functions support this shape - ultimately we don't know what is the goal of the person invoking this function, maybe they just need this as storage for their custom operations and so they don't care about the GEMM requirements?

pggPL:
Please update the docs also

kainzhong:
Merged main and resolved conflicts for @ptrendx

kainzhong (commented):
CC told me that in this branch `tests/pytorch/test_recipe.py:593-594` still uses the deleted `is_quantizable()` method. Other parts seem fine.

PR #2898, opened 2026-04-17
[PyT] [Common] add support for enabling cuda graph under thd format in megatron.

HaochenYuan:
# Description

This PR supports enabling cuda graph under thd format in megatron.  
Related Megatron [PR 4359](https://github.com/NVIDIA/Megatron-LM/pull/4359).

Two root causes block THD under CUDA Graph capture:

1. **GPU→CPU sync forbidden during capture** — `torch.equal()` on CUDA tensors and
   `cu_seqlens[-1]` as a Python int both trigger implicit device sync.
2. **Garbage at padded positions** — fused attention and CP ring kernels leave
   padded slots undefined; in MLA + CP these survive into gradient accumulation.

## Patches

| # | File | Change |
|---|---|---|
| 1 | `attention/dot_product_attention/dot_product_attention.py` | Drop `torch.equal()` check; force `pad_between_seqs=True` for THD (always safe). |
| 2 | `attention/dot_product_attention/context_parallel.py` (`AttnFuncWithCPAndKVP2P.forward`) | Zero-fill `out`/`out_ret` past `cu_seqlens_q_padded[-1]//cp_size` after CP assembly. `arange`-based mask, no sync. |
| 3 | `attention/dot_product_attention/context_parallel.py` (`AttnFuncWithCPAndKVP2P.backward`, existing zero-fill) | Under `is_current_stream_capturing()`, use `dq.shape[0]` instead of `cu_seqlens[-1]` for the slice end. |
| 4 | `attention/dot_product_attention/context_parallel.py` (`AttnFuncWithCPAndKVP2P.backward`, before return) | Zero-fill `dq`/`dk`/`dv` past `cu_seqlens_q_padded[-1]//cp_size` via `arange` mask. |
| 5 | `attention/dot_product_attention/backends.py` (`FusedAttnFunc.backward`) | `empty_like` → `zeros_like` for `dq`/`dk`/`dv` so MLA+CP path doesn't leak garbage. |
| 6 | `cpp_extensions/fused_attn.py` (`fused_attn_fwd`/`fused_attn_bwd`) | Zero-fill output tensors past `cu_seqlens_q[-1]` for `t3hd`/`th3d`/`thd_*` layouts. `arange` mask. |

All zero-fills use `torch.arange(N, device=...) >= bound` so the bound stays a CUDA
tensor — graph-safe.

## Validation

Both Moonlight-16B (MLA + MoE, CP=2) and Qwen3-8B (GQA, CP=2) produce noGraph vs
cudaGraph **bitwise identical** results in `--deterministic-mode` against this branch. See details in [test](https://github.com/NVIDIA/Megatron-LM/pull/4359/changes#diff-6aa776aac35e47a601540500829fb5c29ed8c5d1e94811279ff9f8a7071bfea0).

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1486,9 +1486,11 @@ def backward(ctx, d_out, *_args):
         rest = [None]
         if ctx.use_FAv2_bwd:
             softmax_lse, rng_state = aux_ctx_tensors
-            dq = torch.empty_like(q)
-            dk = torch.empty_like(k)
-            dv = torch.empty_like(v)
+            # THD + CP fix: zeros_like ensures padded positions start from safe values,
+            # preventing garbage from propagating through backward gradient accumulation.
```
timmoon10: Nit: Your comments should describe what the code is doing, not why you made a specific line change. When you read code, you don't care that there used to be a bug months ago, you care about what it does today. Agent-written code is especially bad about this since it's so short-sighted and task-focused.
```suggestion
            # Initialize with zeros to ensure padded positions start from safe values.
```
This bad commit style shows up throughout this PR.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2044,6 +2044,17 @@ def forward(
 
         nvtx_range_pop(f"{nvtx_label}")
 
+        # THD CUDA Graph: zero-fill output at padded positions after CP assembly.
+        # cu_seqlens_q_padded is GLOBAL; divide by cp_size to get local actual_T.
+        if qkv_format == "thd" and out_ret is not None and hasattr(out_ret, "shape"):
+            import torch as _torch
```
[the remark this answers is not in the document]
timmoon10: Good point, this is a strong code smell and I see this antipattern multiple times. If there is a good reason for this, we should add a comment.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2044,6 +2044,17 @@ def forward(
 
         nvtx_range_pop(f"{nvtx_label}")
 
+        # THD CUDA Graph: zero-fill output at padded positions after CP assembly.
+        # cu_seqlens_q_padded is GLOBAL; divide by cp_size to get local actual_T.
+        if qkv_format == "thd" and out_ret is not None and hasattr(out_ret, "shape"):
```
timmoon10: `hasattr(out_ret, "shape")` is obfuscating what you are doing. You're really checking that `out_ret` is a `torch.Tensor` so that you can access `.shape` and `.data`.
```python
        if qkv_format == "thd" and out_ret is not None and isinstance(out_ret, torch.Tensor):
```
This antipattern shows up multiple times in this file.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2680,10 +2691,17 @@ def backward(ctx, dout, *_args):
...
-            dk[cu_seqlens_kv_padded[-1] :].fill_(0)
-            dv[cu_seqlens_kv_padded[-1] :].fill_(0)
+            if torch.cuda.is_current_stream_capturing():
+                _q_end, _kv_end = dq.shape[0], dk.shape[0]
+            else:
+                _q_end = cu_seqlens_q_padded[-1]
+                _kv_end = cu_seqlens_kv_padded[-1]
+            dq[_q_end:].fill_(0)
+            dk[_kv_end:].fill_(0)
+            dv[_kv_end:].fill_(0)
```
timmoon10: We zero out the pads when not capturing a stream, but do a noop when capturing a stream. They're doing different behavior, so one path must be wrong (unless there's CUDA Graph logic elsewhere, which I don't see). If you are relying on the fact that the pads are zeroed out elsewhere, then the non-capture-path is redundant and we can remove this entirely.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1330,13 +1330,10 @@ def forward(
...
-                        cu_seqlens_q_padded is not None
-                        and not torch.equal(cu_seqlens_q_padded[:-1], cu_seqlens_q[:-1])
-                    ) or (
-                        cu_seqlens_kv_padded is not None
-                        and not torch.equal(cu_seqlens_kv_padded[:-1], cu_seqlens_kv[:-1])
-                    )
+                    # THD + CUDA Graph fix: torch.equal() triggers GPU->CPU sync,
+                    # which is forbidden during CUDA graph capture.
+                    # pad_between_seqs=True is always safe for THD with padded cu_seqlens.
+                    pad_between_seqs = True
```
timmoon10: Are there perf implications to blindly assuming that there is padding? If cuDNN has some alignment requirement and we always pad in real training workflows, then maybe this is fine.

@cyanguwa

cyanguwa: Our CP does have some seqlen requirement, `seqlen % (2 x cp_size) == 0`, so there's probably some padding between sequences for CP. Also, it doesn't affect the amount of compute flash/cuDNN attention does - both backends only touch the "meaningful" token positions.

transformer_engine/pytorch/cpp_extensions/fused_attn.py
```diff
@@ -351,6 +351,15 @@ def fused_attn_fwd(
...
     )
 
+    # THD CUDA Graph: zero-fill output at positions beyond cu_seqlens[-1].
+    # Uses pure CUDA ops (no CPU sync) for CUDA graph capture compatibility.
+    if qkv_layout in ("t3hd", "th3d", "thd_t2hd", "thd_th2d", "thd_thd_thd"):
+        _out = output_tensors[0]
+        _aT_fwd = cu_seqlens_q[-1]
+        if _out.shape[0] > 0:
+            _m_fwd = torch.arange(_out.shape[0], device=_out.device) >= _aT_fwd
+            _out[_m_fwd] = 0
```
cyanguwa: Do you only have pad tokens at the end of the batch? Would the pads between sequences in the batch also need to be zeroed out?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1486,9 +1486,11 @@ def backward(ctx, d_out, *_args):
...
+            dq = torch.zeros_like(q)
+            dk = torch.zeros_like(k)
+            dv = torch.zeros_like(v)
```
cyanguwa: Can you guard this change with `if torch.cuda.is_current_stream_capturing()`? I think when we're running without CUDA graphs, we'd still want to get tensors from the PyTorch cache allocator uninitialized, and not pay the cost of zeroing them out at every single call.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
cyanguwa: We do zero out for the per-step results (i.e. non-CP attention calls) for [forward](https://github.com/NVIDIA/TransformerEngine/blob/abdb40668ae0c28370ffb0d896bb4b4d4e896892/transformer_engine/pytorch/csrc/extensions/attention.cpp#L175) and [backward](https://github.com/NVIDIA/TransformerEngine/blob/abdb40668ae0c28370ffb0d896bb4b4d4e896892/transformer_engine/pytorch/csrc/extensions/attention.cpp#L492).

Also, for CP P2P fwd [out](https://github.com/NVIDIA/TransformerEngine/blob/abdb40668ae0c28370ffb0d896bb4b4d4e896892/transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py#L1990), but I don't think we zero out dq/dk/dv in bwd.

cyanguwa:
> Both Moonlight-16B (MLA + MoE, CP=2) and Qwen3-8B (GQA, CP=2) produce noGraph vs
cudaGraph bitwise identical results in --deterministic-mode against this branch. See details in [test](https://github.com/NVIDIA/Megatron-LM/pull/4359/changes#diff-6aa776aac35e47a601540500829fb5c29ed8c5d1e94811279ff9f8a7071bfea0).

Do you know if these tests were running with FlashAttention backend or FusedAttention backend?

HaochenYuan:
> > Both Moonlight-16B (MLA + MoE, CP=2) and Qwen3-8B (GQA, CP=2) produce noGraph vs
> > cudaGraph bitwise identical results in --deterministic-mode against this branch. See details in [test](https://github.com/NVIDIA/Megatron-LM/pull/4359/changes#diff-6aa776aac35e47a601540500829fb5c29ed8c5d1e94811279ff9f8a7071bfea0).
> 
> Do you know if these tests were running with FlashAttention backend or FusedAttention backend?

The Megatron E2E bitwise tests run with FusedAttention, not FlashAttention. In the Qwen3 THD config, padded cu_seqlens make TE infer `pad_between_seqs=True`, and FlashAttention is not available for THD with inter-sequence padding, so auto-selection picks FusedAttention. Moonlight is also FusedAttention-only because `head_dim_qk != head_dim_v`.

The unit tests cover the backends separately: `test_cp_with_flash_attention` exercises the FA path, and the CP FusedAttention tests exercise the Fused path. The THD FA unit test explicitly passes `pad_between_seqs=False` because that test has no inter-sequence padding.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
HaochenYuan: The comments now describe the invariant(zero-init so padded positions don't leak uninitialized values into accumulation) instead of tagging the change as a "fix" against a past bug.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
HaochenYuan: Done — `isinstance(x, torch.Tensor)` everywhere this PR uses the pattern.

transformer_engine/pytorch/cpp_extensions/fused_attn.py
HaochenYuan: Yes, Megatron PR #4359 only adds tail padding to the packed THD buffer. These zero-fills only clear positions `>= cu_seqlens_*_padded[-1]`; inter-sequence padding stays with the existing TE handling.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
HaochenYuan: Done. `zeros_like` is now gated on CUDA graph capture; outside capture uses `empty_like`. I also added the THD tail zero-fill in this FAv2-bwd branch since it returns `dQ/dK/dV` directly.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
HaochenYuan: The precise `torch.equal` check cannot stay because it syncs. The new default is conservative: no padded cu_seqlens => `False`, padded cu_seqlens supplied => `True`. Callers that know there is no inter-sequence padding can still pass `pad_between_seqs=False` explicitly; the THD FA unit-test path now does that.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
HaochenYuan: Fixed. Replaced the capture-specific branch with sync-free `arange` masks, so eager and captured runs execute the same zero-fill. The pre-CP and post-CP zero-fills are both kept because they target different CP-combine stages (per-rank pre-combine for FA's output vs assembled post-combine).

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2044,6 +2044,17 @@ def forward(
...
+        # THD CUDA Graph: zero-fill output at padded positions after CP assembly.
+        # cu_seqlens_q_padded is GLOBAL; divide by cp_size to get local actual_T.
+        if qkv_format == "thd" and out_ret is not None and hasattr(out_ret, "shape"):
+            import torch as _torch
+
+            _local_aT = cu_seqlens_q_padded[-1] // cp_size
+            if out_ret.shape[0] > 0:
+                _m = _torch.arange(out_ret.shape[0], device=out_ret.device) >= _local_aT
+                out_ret.data[_m] = 0
+                out.data[_m.view(-1, *([1] * (out.dim() - 1))).expand_as(out)] = 0
```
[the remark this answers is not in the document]
HaochenYuan: Fixed. The forward zero-fill now only runs when `cu_seqlens_q_padded` is provided, with the tensor/shape guards folded into the same `if`.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2731,6 +2749,16 @@ def backward(ctx, dout, *_args):
...
 
+        # THD CUDA Graph: zero-fill dQ/dK/dV at padded positions after CP backward.
+        if ctx.qkv_format == "thd":
+            import torch as _torch
+
+            _local_aT_bwd = cu_seqlens_q_padded[-1] // get_distributed_world_size(ctx.cp_group)
+            for _dg in [dq, dk, dv]:
+                if _dg is not None and hasattr(_dg, "shape") and _dg.shape[0] > 0:
+                    _mb = _torch.arange(_dg.shape[0], device=_dg.device) >= _local_aT_bwd
+                    _dg[_mb] = 0
```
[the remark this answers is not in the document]
HaochenYuan: Fixed. `dQ` is guarded by `cu_seqlens_q_padded`, and `dK/dV` by `cu_seqlens_kv_padded`; if the corresponding padded seqlens are absent, that zero-fill is skipped.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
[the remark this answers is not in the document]
HaochenYuan: Removed the local `_torch` aliases and use the module-level `torch`.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1679,9 +1679,17 @@ def backward(ctx, d_out, *_args):
         rest = [None]
         if ctx.use_FAv2_bwd:
             softmax_lse, rng_state = aux_ctx_tensors
-            dq = torch.empty_like(q)
-            dk = torch.empty_like(k)
-            dv = torch.empty_like(v)
+            # During CUDA graph capture, allocate with zeros so the memset is baked into
+            # the captured graph and replay buffers start clean. Outside capture, allocate
+            # with empty for perf and rely on the explicit tail zero-fill below.
```
timmoon10: I'm suspicious of this comment, do you know where this graph-unsafe zero-fill is taking place? Shouldn't we need a guard to protect us from calling it during graph capture? Otherwise, if the graph and non-graph path are calling the same kernels, then the memset is unnecessary.

HaochenYuan: Fixed, this comment was misleading. No graph-unsafe zero-fill here; both allocation paths are capturable. I updated the comment to describe the actual tradeoff: keep zero-initialized buffers during capture, use `empty_like` outside capture to avoid the extra memset, and rely on the explicit THD tail zero-fill in both modes.

timmoon10: I still don't understand the benefit zeroing is giving us. As far as I can tell, zeroing is the only difference between capture and non-capture, so we have three options:
1. The non-capture path is incorrect. But why haven't we seen this bug before?
2. The capture path is calling an unnecessary memset. We should just initialize with `torch.empty`.
3. I missed something and graph capture does affect the logic elsewhere. The comment needs to be improved, since even a TE developer couldn't understand.

HaochenYuan: The right option is 2. I removed the gated zeros_like and tested moonlight-16B e2e correctness, sure it is not needed after adding the **explicit THD tail zero-fill** in one previous commit. FA bwd writes the real-token outputs, and the tail zero fill clears the static padding before return. The comment is also deleted.

timmoon10 (dismissed):
LGTM, pending CI

HaochenYuan:
The remaining CI failure looks like an infra/runner issue？: GitHub reports "The hosted runner lost communication with the server" during the Build step.

transformer_engine/pytorch/cpp_extensions/fused_attn.py
cyanguwa: Tail padding is probably a very special case of `pad_between_seqs=True`. If we want to claim that THD + CUDA graph is working, could we please zero out all the inter-sequence padding as well (here and in a few other places)? Thanks!

cyanguwa:
> The remaining CI failure looks like an infra/runner issue？: GitHub reports "The hosted runner lost communication with the server" during the Build step.

The failed tests are due to a different issue which we're fixing, so I think you can ignore those. Thanks!

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2897,10 +2913,20 @@ def backward(ctx, dout, *_args):
             dim = ctx.qkv_format.index("s")
             dq, dk, dv = [x.view(*x.shape[:dim], -1, *x.shape[dim + 2 :]) for x in [dq, dk, dv]]
 
-        if ctx.qkv_format == "thd" and not ctx.use_fused_attention:
-            dq[cu_seqlens_q_padded[-1] :].fill_(0)
-            dk[cu_seqlens_kv_padded[-1] :].fill_(0)
-            dv[cu_seqlens_kv_padded[-1] :].fill_(0)
```
timmoon10, line 2852: Future reference: This is the critical change if I understand correctly. This pad zeroing implementation requires a D2H memcpy, so we are replacing it with a sync-free impl in order to support CUDA Graphs. We also add sync-free pad zeroing in a few other places. This is not required for correctness (otherwise the non-CUDA-Graph version would have failed), but it is safer. @cyanguwa should confirm that this extra safety is worth the cost of the extra kernel launches.

cyanguwa: I think it's worth it, otherwise we can't guarantee the results are correct. I do think this code should be guarded with "if cuda_graph" though, so that it doesn't happen for the non-cuda-graph path.

timmoon10: I don't think it makes sense to guard unless there's some other difference between the capture and non-capture code paths (similar logic to https://github.com/NVIDIA/TransformerEngine/pull/2898#discussion_r3375304411).

HaochenYuan: I remove some tail zero-fill code, only keep ones that cause loss/gradnorm to be Nan in megatron e2e test. Also, add graph guard code here for this zero-fill operation.

HaochenYuan:
Add one fuesd_moe_aux_loss kernel in this PR, the only change is allow the **total_num_tokens** to be a tensor type, so when we enable router fusion in megatron with cuda graph in thd format, the **total_num_tokens** can vary across different micro-batch. The previous int type couldn't allow value change when replay graph.

transformer_engine/pytorch/cpp_extensions/fused_attn.py
HaochenYuan: In current megatron thd e2e test, due to accuracy problem, we let cu_seqlens_* to be equal to cu_seqlens_*_padded,  and address the inter-sequence padding by loss mask. So here we want to only focus on tail padding.

transformer_engine/common/include/transformer_engine/fused_router.h
```diff
@@ -190,25 +190,45 @@ void nvte_fused_score_for_moe_aux_loss_backward(const NVTETensor intermediate_ou
...
+/*! \brief Forward pass for auxiliary loss. Device-tensor total_num_tokens path:
+ *  the coefficient is computed on device from a 0-dim int64 GPU tensor so its
+ *  value stays dynamic across CUDA Graph replays. Adds one extra single-thread
+ *  kernel launch versus the host-int path; prefer this only when the caller
+ *  needs CUDA-graph-safe semantics.
+ *
+ *  \param[in]     total_num_tokens   0-dim int64 GPU tensor with the total token count.
+ *  Other parameters as in :c:func:`nvte_fused_moe_aux_loss_forward`.
+ */
+void nvte_fused_moe_aux_loss_forward_tensor(const NVTETensor probs,
```
timmoon10: This function name isn't very descriptive. Maybe something like `nvte_fused_moe_aux_loss_forward_with_total_num_tokens_tensor` or `nvte_fused_moe_aux_loss_forward_graph_safe`? If we change the name, we should propagate it consistently.

transformer_engine/common/fused_router/fused_moe_aux_loss.cu
```diff
@@ -132,6 +132,115 @@ void fused_moe_aux_loss_forward(const Tensor& probs, const Tensor& tokens_per_ex
...
 }
 
+/* -------------------------------------------------------------------------
+ *  CUDA-graph-safe variant: total_num_tokens lives in a 0-dim int64 device
+ *  tensor whose value can change between graph replays. A single-thread
+ *  prelude kernel reads it on device and writes C_coeff into Coeff_buf[0];
+ *  the main reduction kernel then reads C_coeff from Coeff_buf[0] instead
+ *  of receiving it as a launch parameter, so the value stays dynamic.
+ * ------------------------------------------------------------------------- */
+__global__ void fused_moe_aux_loss_compute_coeff_kernel(const int64_t* total_num_tokens_ptr,
```
timmoon10: Instead of launching a separate kernel for the coeff compute, could we do the compute redundantly within the main kernel? Each thread will have the same memory accesses and a little extra math, and we'll avoid the overhead of a kernel launch:
```c++
const int64_t total_num_tokens = *total_num_tokens_ptr;
const float C_coeff = (static_cast<float>(num_experts) * coeff) / static_cast<float>(topk * total_num_tokens * total_num_tokens);
if (blockIdx.x == 0 && threadIdx.x == 0) {
  Coeff_buf[0] = C_coeff;
}
```
This change would also mean that the graph-safe kernel has almost no downside compared to the legacy kernel.

transformer_engine/common/include/transformer_engine/fused_router.h
HaochenYuan: good advice, i've changed the kernel name.

transformer_engine/common/fused_router/fused_moe_aux_loss.cu
HaochenYuan: Done

timmoon10 (dismissed):
LGTM, pending CI.

transformer_engine/common/fused_router/fused_moe_aux_loss.cu
```diff
@@ -132,6 +132,100 @@ void fused_moe_aux_loss_forward(const Tensor& probs, const Tensor& tokens_per_ex
...
+  extern __shared__ float shmem[];
+  CompType* shmem_block = reinterpret_cast<CompType*>(shmem);
+  shmem_block[threadIdx.x] = thread_sum;
+  __syncthreads();
+
+  const int warp_id = threadIdx.x / kThreadsPerWarp;
+  const int lane_id = threadIdx.x % kThreadsPerWarp;
+  if (warp_id == 0) {
+    CompType block_sum = warp_reduce_on_shmem(shmem_block, static_cast<int>(blockDim.x),
+                                              ReduceFuncType::SUM, lane_id);
```
timmoon10: The function signature was changed in https://github.com/NVIDIA/TransformerEngine/pull/3012.
```suggestion
    CompType block_sum = warp_reduce_on_shmem<CompType, ReduceFuncType::SUM>(shmem_block, static_cast<int>(blockDim.x), lane_id);
```

timmoon10 (approved):

Issue #2897, opened 2026-04-16
[PyTorch] Reduce CPU overhead in grouped MLP block


timmoon10:
**Is your feature request related to a problem? Please describe.**

The fused operations for the grouped MLP block (see `ForwardGroupedMLP_CuTeGEMMSwiGLU_MXFP8` and `BackwardGroupedMLP_CuTeGEMMDSwiGLU_MXFP8`, first added in https://github.com/NVIDIA/TransformerEngine/pull/2769) has significant CPU overhead. When I run a basic benchmark (on GB200 with 64 experts and 128 hidden size), I find the forward pass takes ~1.2 ms and the backward pass takes ~2.1 ms.

**Describe the solution you'd like**

Based on profiling, here are some rough estimates for some slow sections:
- [ ] wgrad tensor allocations: 350 us
- [ ] `tex.get_device_pointer_for_data_and_scales`: >100 us, with ~50 us before entering `nvte_multi_tensor_swizzle_scaling_factors`
- [ ] `clear_tensor_data`: 200 us
- [ ] Initializing weight quantizers in the forward pass: 90 us
- [ ] Tensor reshapes before and after cuDNN kernels: ~50 us
- [ ] cuDNN group GEMM kernels: ~150 us

As we make optimizations, we should also adapt them to the unfused grouped linear op, and generally consider cleanups and refactors.

**Describe alternatives you've considered**

**Additional context**

zhongbozhu:
Huge CPU overhead improvements is also needed here: https://github.com/NVIDIA/TransformerEngine/blob/76c2a9e90275a6856039fbba9fcb09bc98c48605/transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py#L268 
```
        fc1_x_data = grouped_fc1_x.rowwise_data.view(in_shape[0], in_shape[1])
        fc1_x_data = fc1_x_data.view(dtype=torch.float8_e4m3fn)
        fc1_x_data = fc1_x_data.unsqueeze(0).permute(1, 2, 0)
        fc1_x_scales = grouped_fc1_x.scale_inv
        fc1_x_scales = fc1_x_scales.view(dtype=torch.float8_e8m0fnu)
        fc1_x_scales = fc1_x_scales.view(
            1,
            (in_shape[0] + 127) // 128,
            (in_shape[1] + 127) // 128,
            MXFP8_BLOCK_SCALING_SIZE,
            4,
            4,
        )
[...]

PR #2900, opened 2026-04-18
[PyTorch] Expose function to bulk-allocate tensors backed by the same buffer

timmoon10:
# Description

Allocating PyTorch tensors carries non-trivial CPU overhead, which becomes especially painful when allocating per-expert tensors in the grouped linear layer. This PR generalizes the bulk allocation approach used in the split-quantize functions (see https://github.com/NVIDIA/TransformerEngine/pull/1793), which involves allocating a single large buffer and creating tensor subviews with `at::from_blob`. We expose a dedicated bulk-allocation function that is exposed to Python, refactor the split-quantize functions, and bulk-allocate wgrads in the grouped linear implementations.

This is incremental progress toward https://github.com/NVIDIA/TransformerEngine/issues/2897. When I run the grouped MLP benchmark discussed in that issue, I see a runtime reduction of 150 us (for reference the backward pass takes ~2.1 ms).

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Add dedicated function for bulk-allocating PyTorch tensors
- Bulk-allocate wgrad tensors in grouped linear implementations

# Checklist:


transformer_engine/pytorch/csrc/extensions/allocate.cpp
```diff
@@ -0,0 +1,64 @@
...
+  out.reserve(n);
+  std::vector<int64_t> shape_int64;
+  for (size_t i = 0; i < n; ++i) {
+    shape_int64.assign(shapes[i].begin(), shapes[i].end());
+    if (byte_sizes[i] == 0) {
+      // Work around problems with from_blob when constructing an
+      // empty tensor. Passing a null pointer fails because it checks
+      // that the pointer is on GPU. Passing a non-null pointer can
+      // cause bugs in TE kernels.
+      out.emplace_back(at::empty(shape_int64, at::device(at::kCUDA).dtype(dtypes[i])));
```
vthumbe1503: Minor comment: I am not sure what the CPU overheads of doing at::empty with 0 elements is. If it is not good, we can probably cache this tensor across the code base. Given that such a tensor wont be used for writing.

Also might not necessarily be on the scope of this PR.

transformer_engine/pytorch/csrc/extensions/allocate.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <vector>
+
+#include "../extensions.h"
+
+namespace transformer_engine {
+namespace pytorch {
+
+std::vector<at::Tensor> bulk_allocate(const std::vector<std::vector<size_t>> &shapes,
+                                      const std::vector<at::ScalarType> &dtypes,
+                                      const std::vector<size_t> &alignments) {
```
ptrendx: Why isn't this optional?

transformer_engine/pytorch/csrc/extensions/allocate.cpp
```diff
@@ -0,0 +1,64 @@
...
+             " alignments.");
+  if (n == 0) return {};
+
+  // Compute per-tensor sizes and offsets
+  size_t total_bytes = 0;
+  std::vector<size_t> byte_sizes(n);
+  std::vector<size_t> offsets(n);
+  for (size_t i = 0; i < n; ++i) {
+    if (alignments[i] > 0) {
+      total_bytes = roundup(total_bytes, alignments[i]);
```
ptrendx: This effectively means that alignments[0] is never actually used. If alignment[0] is more that what CUDA provides (256B I believe?) then we need to add the logic to pad the allocation from the beginning ourselves.

transformer_engine/pytorch/csrc/extensions/allocate.cpp
```diff
@@ -0,0 +1,64 @@
...
+    }
+    offsets[i] = total_bytes;
+    byte_sizes[i] = product(shapes[i]) * at::elementSize(dtypes[i]);
+    total_bytes += byte_sizes[i];
+  }
+
+  // Single backing allocation
+  auto buffer = std::make_shared<at::Tensor>(
+      at::empty({static_cast<int64_t>(total_bytes)}, at::device(at::kCUDA).dtype(torch::kUInt8)));
```
ptrendx: We should provide the device as the argument to the API.

transformer_engine/pytorch/csrc/extensions/allocate.cpp
ptrendx: This should also not hardcode the default device.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -849,62 +747,51 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>, bool> bulk_alloc
...
       rowwise_scale_shapes.emplace_back(
           quantizer_cpp_list[i]->get_scale_shape(shape_list[i], false));
     }
 
-    // Offsets in full buffer
-    size_t buffer_size = 0;
-    std::vector<size_t> data_offsets, scale_offsets, amax_offsets;
+    // Check whether data and scales can be packed in contiguous
+    // buffer. Amaxes are not contiguous since they are aligned to
+    // 16B.
```
ptrendx, line 758: I don't think I understand the logic here. What is the case where we would not be able to pack those buffers together if we can control the alignment requirements for the individual tensors in the allocation?

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -484,10 +484,12 @@ def backward(
...
-                    wgrad_list = [
-                        torch.empty(w.size(), dtype=ctx.activation_dtype, device=ctx.device)
-                        for w in weights
-                    ]
+                    weight_shape = list(weights[0].size())
+                    wgrad_list = tex.bulk_allocate(
+                        [weight_shape] * ctx.num_gemms,
+                        [ctx.activation_dtype] * ctx.num_gemms,
+                        [0] * ctx.num_gemms,  # alignment
+                    )
```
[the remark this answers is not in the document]
ptrendx: That is a general assumption for the groupedlinear module.

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -955,13 +955,12 @@ def fuser_backward(
...
-                    grad_weights[group_idx] = torch.empty(
-                        weight_shape,
-                        dtype=ctx.dtype,
-                        device=device,
-                    )
+                weight_shape = [self.out_features, self.in_features]
+                grad_weights = tex.bulk_allocate(
+                    [weight_shape] * num_groups,
+                    [ctx.dtype] * num_groups,
+                    [0] * num_groups,  # alignment
```
ptrendx: Why 0 and not e.g. 256B?

ptrendx: Proper alignment is important for cuBLAS heuristics.

transformer_engine/pytorch/ops/fused/backward_grouped_mlp.py
```diff
@@ -199,8 +199,11 @@ def _compute_grad_params(
                     w_list[idx] = wp.main_grad
                 accumulate_into_main_grad = not getattr(fc_op.weight0, "overwrite_main_grad", False)
             else:
-                for idx in range(num_groups):
-                    w_list[idx] = torch.empty(weight_shape, dtype=dtype, device=device)
+                w_list = tex.bulk_allocate(
+                    [list(weight_shape)] * num_groups,
+                    [dtype] * num_groups,
+                    [0] * num_groups,  # alignment
```
ptrendx: Same  - we should have 256 here.

transformer_engine/pytorch/csrc/extensions/allocate.cpp
timmoon10: Good point, I suppose the safest approach is to overallocate by `alignment[0]` so that we can pad if we get an unaligned buffer.

transformer_engine/pytorch/csrc/extensions/cast.cpp
timmoon10: I think there was some alignment requirement for the weight tensors, and some implementation difference depending on if the weights are contiguous. In any case, this is the logic in the existing implementation and I'm trying to avoid functional changes.

transformer_engine/pytorch/csrc/extensions/allocate.cpp
timmoon10: That might save some overhead, but would also reduce flexibility. For example, if a module outputs the tensors produced by `split_quantize`, then duplicating tensors might break autograd. I think we should prefer generality and safety until we have a good reason to microoptimize.

transformer_engine/pytorch/csrc/extensions/allocate.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <vector>
+
+#include "../extensions.h"
+
+namespace transformer_engine {
+namespace pytorch {
+
+std::vector<at::Tensor> bulk_allocate(const std::vector<std::vector<size_t>> &shapes,
+                                      const std::vector<at::ScalarType> &dtypes,
+                                      std::optional<at::Device> device,
```
vthumbe1503: Minor Comment, not blocking: I guess we should use c10::Device here as well to maintain consistency?

vthumbe1503 (dismissed):
LGTM

vthumbe1503 (approved):

PR #2906, opened 2026-04-21
Add head dim 256 support for SDPA on Blackwell

yaox12:
# Description

Add head dim 256 support for SDPA on Blackwell via cuDNN FE cuteDSL kernels.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

The existing C++ cuDNN SDPA path cannot handle head_dim=256 training on SM100+; cuDNN frontend ≥1.22.1 ships a Python-only (CuTe DSL) kernel for this shape. This PR wires that kernel into TE's FusedAttention dispatch.

- C++ filter — `transformer_engine/common/fused_attn/fused_attn.cpp`
- New Python wrapper — `transformer_engine/pytorch/attention/dot_product_attention/cudnn_fe_sdpa.py`
- Dispatcher — `transformer_engine/pytorch/cpp_extensions/fused_attn.py`
- Python-side filter — `transformer_engine/pytorch/attention/dot_product_attention/utils.py`
- Tests — `tests/pytorch/attention/test_headdim256_cudnn_fe.py`

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/attention/dot_product_attention/cudnn_fe_sdpa.py
+++ b/transformer_engine/pytorch/attention/dot_product_attention/cudnn_fe_sdpa.py
@@ -141,7 +141,7 @@ def _causal_and_window(
     attn_mask_type: str, window_size: Tuple[int, int]
 ) -> Tuple[bool, Tuple[int, int]]:
     is_causal = "causal" in attn_mask_type
-    left, right = window_size
+    left, _ = window_size
     if is_causal:
         return True, (left if left is not None else -1, 0)
     return False, (-1, -1)
--- a/transformer_engine/pytorch/attention/dot_product_attention/cudnn_fe_sdpa.py
+++ b/transformer_engine/pytorch/attention/dot_product_attention/cudnn_fe_sdpa.py
@@ -1,0 +1,277 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""Dispatch layer for the cuDNN frontend Python SDPA (CuTe DSL), head_dim=256.
+
+The kernels in ``cudnn.sdpa`` are Python-only (CuTe DSL), so invocation happens
+entirely on the Python side. This module adapts TransformerEngine's
+``fused_attn_fwd``/``fused_attn_bwd`` calling convention to the wrappers
+``sdpa_fwd_wrapper_sm100_d256``/``sdpa_bwd_wrapper_sm100_d256`` and hides the
+layout massaging required by the kernel.
+"""
+
+from __future__ import annotations
+
+import functools
+from typing import List, Optional, Tuple
+
+import torch
+
+
+@functools.lru_cache(maxsize=None)
+def _sdpa_fwd_wrapper():
+    from cudnn import sdpa_fwd_wrapper_sm100_d256  # pylint: disable=no-name-in-module
+
+    return sdpa_fwd_wrapper_sm100_d256
+
+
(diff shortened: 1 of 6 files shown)


transformer_engine/pytorch/attention/dot_product_attention/utils.py
timmoon10, line 1: I don't think we should use the `use_fused_attention=True` code path for the CuTe DSL kernels.
- The C++ and Python APIs in the cuDNN frontend don't really have anything to do with each other, and they're only put together for organizational reasons.
- This doesn't generalize well if we want to add more backends. I could imagine TE eventually having some DPA kernels written in Triton or maybe some new package will come out with a better attention implementation.
- Putting the C++ and Python cuDNN APIs together means we will need to do redundant feature checks, once to decide `use_fused_attn=True` and again to choose between the C++ or Python cuDNN APIs (and again in the C++ infrastructure to choose the `NVTE_Fused_Attn_Backend`). This extra layer is not so bad now since there's only one Python case, but it will scale poorly if we get closer to a 50/50 split between C++ and Python.

The right approach would be to make a new backend, but we're restricted by the current design of the backend selection logic. In particular, `get_attention_backend` is hard-coded to return `use_flash_attention`, `use_fused_attention`, `use_unfused_attention`, so we're kind of stuck. I propose changing this function to return a list of strings. Adding a new backend just involves adding logic to handle a new string.

CC @cyanguwa

transformer_engine/pytorch/cpp_extensions/fused_attn.py
```diff
@@ -98,12 +97,17 @@
     "F16_max512_seqlen": NVTE_Fused_Attn_Backend.NVTE_F16_max512_seqlen,
     "F16_arbitrary_seqlen": NVTE_Fused_Attn_Backend.NVTE_F16_arbitrary_seqlen,
     "FP8": NVTE_Fused_Attn_Backend.NVTE_FP8,
+    # Python-only sentinel (no C++ counterpart). Set by ``get_attention_backend``
+    # when the cuDNN frontend CuTe-DSL d=256 SDPA path is applicable; checked
+    # in ``fused_attn_fwd`` / ``fused_attn_bwd`` to route to the Python kernel.
+    "F16_cudnn_fe_sdpa": "F16_cudnn_fe_sdpa",
```
timmoon10, line 103: Since `FusedAttnBackend` is no longer guaranteed to return an `NVTE_Fused_Attn_Backend`, the safe thing to do would be to check the return every thing we access this dict. That's really painful, so it's a hint we're doing something wrong.

transformer_engine/pytorch/attention/dot_product_attention/cudnn_fe_sdpa.py
```diff
@@ -0,0 +1,277 @@
...
+    max_seqlen_q: int,
+    max_seqlen_kv: int,
+    is_training: bool,
+    deterministic: bool,
+    device_compute_capability: Tuple[int, int],
+    return_max_logit: bool = False,
+) -> bool:
+    """Whether the cuDNN-FE SDPA d=256 kernel can service this configuration."""
+    if device_compute_capability[0] < 10:
+        return False
```
timmoon10, line 77: Does the kernel work with SM 12 or beyond? To be safe:

```suggestion
    if device_compute_capability[0] == 10:
        return False
```

transformer_engine/pytorch/attention/dot_product_attention/utils.py
yaox12: Yes, I agree. Though that's not an easy refactor.

PR #2907, opened 2026-04-21
[Common][PyTorch] Fix int32 overflow and -1 sentinel handling in moe_permute

### LINKED ISSUE #2908: [Bug] moe_permute CUDA kernel: int32 overflow and incorrect -1 sentinel handling
Thanks for these efficient MoE kernels — we rely on them in training.

Two independent latent bugs in `transformer_engine/common/permutation/permutation.cu` and its PyTorch extension caller. Both reproduce on `main` (`264da2b`), release tag `v2.13` (`2877704`), and `v2.9`. The buggy code has been unchanged since the kernel was introduced in #936 (Aug 2024).

## Environment for repros

```
conda-forge:
  python              3.13.13
  pytorch             2.10.0   cuda129_mkl_py313_h623d66f_303
  transformer-engine-torch  2.13  py313hba49f57_1
  cuda-version        12.9
GPU: NVIDIA H20Z
```

---

## Bug 1 — `int32` overflow in `moe_unpermute_kernel` / `moe_permute_kernel`

`source_token * num_cols` and `source_row * num_cols` are computed in `int`:

```cpp
// moe_unpermute_kernel
const int source_token = blockIdx.x;
int source_row = row_id_map[source_token];
const T *source_row_ptr = input + source_row * num_cols;   // int * int → overflows
```

Once `num_out_tokens * num_cols ≥ 2³¹`, the pointer offset wraps and the kernel reads/writes at a bogus address. **We hit this on DeepSeek-V3 MoE training** (long-context runs with `hidden = 7168`, `topK = 8`; `num_out_tokens * hidden` crosses the 2³¹ boundary once per-rank token count grows past a few hundred thousand). The failure mode is either silent corruption — sudden training-loss spikes / NaNs that took us a while to trace back to the kernel — or an outright `CUDA error: an illegal memory access was encountered`.

### Repro

```python
# repro_int_overflow.py  (~20 GB GPU memory needed)
import torch
from transformer_engine.pytorch import moe_permute

num_tokens = 2 ** 18 + 1   # 262_145
num_cols   = 2 ** 13       # 8_192  →  (num_out_tokens - 1) * num_cols = 2**31 exactly
topk, num_experts = 1, 4

x       = torch.randn(num_tokens, num_cols, dtype=torch.bfloat16, device="cuda")
indices = (torch.arange(num_tokens, dtype=torch.int32, device="cuda")
           .remainder(num_experts).view(-1, topk))

# Case A: just below 2**31  -> 262144 tokens  -> OK, matches reference bit-for-bit
# Case B: at 2**31          -> 262145 tokens  -> BUG
te_permuted, _ = moe_permute(x, indices, num_out_tokens=num_tokens * topk, map_type="index")
torch.cuda.synchronize()
```

Observed (case B):

```
RuntimeError: .../transformer_engine/common/permutation/permutation.cu:252 in
function nvte_permute_launcher: CUDA Error: an illegal memory access was encountered
```

Case A (`num_tokens = 2**18`, product just under `2**31`) runs clean with diff `0.0` against `torch.argsort + index_select`.

---

## Bug 2 — Incorrect handling of `-1` sentinels in `routing_map`

Libraries such as **DeepEP** (and any expert-parallel mask that sets non-local `(token, slot)` pairs to `-1`) feed a `routing_map` with `-1` entries. `cub::DeviceRadixSort::SortPairs` is **signed ascending**, so those sentinels land at the **head** of `sorted_row_id`, not the tail.

`moe_permute_row_map`, however, assumes the opposite:

```cpp
if (idx >= num_rows * topK) return;

int source_row = sorted_row_id[idx];
int source_token_id = source_row / topK;
int source_topK_id = source_row % topK;

if (idx >= num_out_tokens) {
  // Set the indices of dropped tokens to -1
  row_id_map[source_topK_id * num_rows + source_token_id] = -1;
} else {
  // Create a row id map for subsequent unpermute operation
  row_id_map[source_topK_id * num_rows + source_token_id] = idx;
}
```

For `idx < num_minus_ones`, the kernel reads `sorted_row_id[idx] == -1`, computes `source_token_id = -1 / topK` / `source_topK_id = -1 % topK` (implementation-defined), and writes a valid `idx` into an unrelated slot of `row_id_map`. Conversely, valid tokens that should be kept lose their mapping. The subsequent `unpermute` then silently mixes sentinel rows into the output.

### Expert-parallel context (why `-1` appears)

```
4 global experts (E0/E1/E2/E3), 2 ranks.
  Rank 0 holds: E0, E1
  Rank 1 holds: E2, E3
Router assigns (replicated across ranks):
  token_a → [0, 3]
  token_b → [2, 1]
  token_c → [3, 2]

On Rank 0 we mask out non-local experts:
  a → [ 0, -1]    # E3 not on this rank
  b → [-1,  1]
  c → [-1, -1]
  num_out_tokens = 2
```

Each rank then calls `moe_permute` with its masked indices, expecting `-1` slots to be skipped.

### Repro

```python
# repro_minus_one.py  (trivial, any GPU)
import torch
from transformer_engine.pytorch import moe_permute

def reference_permute(x, indices, num_out_tokens):
    topk = indices.size(1)
    flat = indices.view(-1)
    sorted_idx = torch.argsort(flat, stable=True)
    sorted_idx = sorted_idx[flat[sorted_idx] != -1]
    assert sorted_idx.numel() == num_out_tokens
    return x.index_select(0, sorted_idx // topk), sorted_idx

torch.manual_seed(0)
N, topk, H = 8, 2, 16
tokens  = torch.randn(N, H, dtype=torch.bfloat16, device="cuda")
indices = torch.randint(0, 4, (N, topk), dtype=torch.int32, device="cuda")
# EP-style mask: half the (token, slot) pairs → -1
flat = indices.view(-1)
flat[torch.randperm(flat.numel(), device="cuda")[: flat.numel() // 2]] = -1
num_out_tokens = int((indices != -1).sum())

te_permuted, _  = moe_permute(tokens, indices,
                              num_out_tokens=num_out_tokens, map_type="index")
ref_permuted, _ = reference_permute(tokens, indices, num_out_tokens)

print("max |TE - ref| =", (te_permuted.float() - ref_permuted.float()).abs().max().item())
```

Observed (main / v2.13):

```
indices =
[[ 0 -1]
 [-1  2]
 [-1  2]
 [ 1  1]
 [ 3 -1]
 [ 3 -1]
 [-1  1]
 [-1 -1]]
num_out_tokens = 8

max |TE - ref| = 4.562e+00          # bf16 — full-scale garbage, not noise
TE permuted (first 3 rows):
[[-0.92578125 -0.42578125 -2.640625    0.14550781]
 [-0.12695312  1.21875     1.4375      1.0625    ]    ← wrong row
 [-0.59765625 -0.328125   -0.91015625 -0.8046875 ]]   ← wrong row
ref permuted (first 3 rows):
[[-0.92578125 -0.42578125 -2.640625    0.14550781]
 [-0.18359375  0.3828125   0.39257812 -0.08300781]
 [-0.18359375  0.3828125   0.39257812 -0.08300781]]
```

Expected: the first row (token 0 → expert 0) matches; rows 1 and 2 should both be token 3 (the only other survivor among indices with expert 1), but TE returns two different arbitrary rows instead.

---

## Related prior report

Issue #1336 (*"the max error of moe_permute/unpermute.grad could reach 3.6e+00"*, Nov 2024) is very likely the same root cause — the reporter observed full-scale errors that grew with `topK` and noted *"This has had some impact on training loss."* Neither `-1`-sentinel nor `int32`-overflow handling has been touched since that report.

## Fix

PR #2907 proposes a minimal fix:

1. Widen `source_token`, `source_row`, `dest_row` to `int64_t` inside the kernels.
2. In the caller, advance `sorted_row_id_ptr` past the `num_minus_ones` sentinel prefix and pre-fill `row_id_map` with `-1` via `torch::full`; switch the launcher grid to `num_out_tokens` blocks.

No public API / dtype changes. Happy-path workloads (no `-1`, offset within `int32`) are unchanged.

jing-4369:
Fixes #2908 — full description, repros, and DeepSeek-V3 context there.

## Changes

- **`permutation.cu`** — widen `source_token`, `source_row`, `dest_row` to `int64_t` inside `moe_unpermute_kernel` and `moe_permute_kernel` so `row * num_cols` stays 64-bit. Simplify `moe_permute_row_map` to only process the valid `[0, num_out_tokens)` range; launcher grid becomes `num_out_tokens` blocks.
- **`permutation.cpp`** — advance `sorted_row_id_ptr` past the `num_minus_ones` sentinel prefix left by `cub::DeviceRadixSort` (signed ascending), and pre-fill `row_id_map` with `-1` via `torch::full` so dropped slots are marked without the kernel ever dereferencing a sentinel.

No public API / dtype changes. `+17 / -18` lines across the two files.

## Test plan

- [x] Happy-path `routing_map` (no `-1`, offsets within int32) — unchanged.
- [x] `-1`-sentinel repro from #2908 → `max |TE - ref| = 0.0` on bf16 (was `4.56e0`).
- [x] `int32`-boundary repro from #2908 → no longer raises `illegal memory access`; matches reference.
- [ ] Full `tests/pytorch/test_permutation.py` via CI.

### DIFF
--- a/transformer_engine/common/permutation/permutation.cu
+++ b/transformer_engine/common/permutation/permutation.cu
@@ -19,19 +19,12 @@ static __global__ void moe_permute_row_map(const int *sorted_row_id, int *row_id
   const int tid = threadIdx.x;
   const int idx = bid * blockDim.x + tid;
 
-  if (idx >= num_rows * topK) return;
+  if (idx >= num_out_tokens) return;
 
   int source_row = sorted_row_id[idx];
   int source_token_id = source_row / topK;
   int source_topK_id = source_row % topK;
-
-  if (idx >= num_out_tokens) {
-    // Set the indices of dropped tokens to -1
-    row_id_map[source_topK_id * num_rows + source_token_id] = -1;
-  } else {
-    // Create a row id map for subsequent unpermute operation
-    row_id_map[source_topK_id * num_rows + source_token_id] = idx;
-  }
+  row_id_map[source_topK_id * num_rows + source_token_id] = idx;
 }
 
 template <typename T, typename TCompute, bool hasProb>
@@ -42,7 +35,7 @@ __global__ void moe_unpermute_kernel(const T *input, T *unpermuted_output, const
   TCompute *s_prob = reinterpret_cast<TCompute *>(s_mem);
 
   // Each block corresponds to one dest token
-  const int source_token = blockIdx.x;
+  const int64_t source_token = blockIdx.x;
   const int tid = threadIdx.x;
 
   if (hasProb) {
@@ -65,7 +58,7 @@ __global__ void moe_unpermute_kernel(const T *input, T *unpermuted_output, const
     TCompute frag_elem[kElementsPerAccess];
     TCompute frag_sum[kElementsPerAccess];
 
-    int source_row = row_id_map[source_token];
+    int64_t source_row = row_id_map[source_token];
 
     // source_row == -1 represents a dropped token
     if (source_row != -1) {
@@ -134,7 +127,7 @@ __global__ void moe_permute_kernel(const T *input_bwd, const T *input_fwd, T *ac
   TCompute *s_prob = reinterpret_cast<TCompute *>(s_mem);
 
   // Each block corresponds to one source token
-  const int source_token = blockIdx.x;
+  const int64_t source_token = blockIdx.x;
   const int tid = threadIdx.x;
 
   if (hasProb) {
@@ -172,7 +165,7 @@ __global__ void moe_permute_kernel(const T *input_bwd, const T *input_fwd, T *ac
     for (int k = 0; k < topKTile; k++) {
       if (k == topK) break;
 
-      int dest_row = row_id_map[index];
+      int64_t dest_row = row_id_map[index];
       index += num_rows;
 
       if (dest_row != -1) {
@@ -239,7 +232,7 @@ void nvte_permute_launcher(const T *input, T *output, const int *sorted_row_id, 
     // moe_permute_fwd
 
     int threads = 64;
-    int blocks = (num_rows * topK + threads - 1) / threads;
+    int blocks = (num_out_tokens + threads - 1) / threads;
 
     moe_permute_row_map<<<blocks, threads, 0, stream>>>(sorted_row_id, row_id_map, num_rows, topK,
                                                         num_out_tokens);
--- a/transformer_engine/common/permutation/permutation.cu
+++ b/transformer_engine/common/permutation/permutation.cu
@@ -19,7 +19,7 @@ static __global__ void moe_permute_row_map(const int *sorted_row_id, int *row_id
   const int tid = threadIdx.x;
   const int idx = bid * blockDim.x + tid;
 
-  if (idx >= num_rows * topK) return;
+  if (idx >= static_cast<int64_t>(num_rows) * topK) return;
 
   int source_row = sorted_row_id[idx];
   int source_token_id = source_row / topK;
@@ -27,10 +27,10 @@ static __global__ void moe_permute_row_map(const int *sorted_row_id, int *row_id
 
   if (idx >= num_out_tokens) {
     // Set the indices of dropped tokens to -1
-    row_id_map[source_topK_id * num_rows + source_token_id] = -1;
+    row_id_map[static_cast<int64_t>(source_topK_id) * num_rows + source_token_id] = -1;
   } else {
     // Create a row id map for subsequent unpermute operation
-    row_id_map[source_topK_id * num_rows + source_token_id] = idx;
+    row_id_map[static_cast<int64_t>(source_topK_id) * num_rows + source_token_id] = idx;
   }
 }
 
@@ -239,7 +239,7 @@ void nvte_permute_launcher(const T *input, T *output, const int *sorted_row_id, 
     // moe_permute_fwd
 
     int threads = 64;
-    int blocks = (num_rows * topK + threads - 1) / threads;
+    int blocks = (static_cast<int64_t>(num_rows) * topK + threads - 1) / threads;
 
     moe_permute_row_map<<<blocks, threads, 0, stream>>>(sorted_row_id, row_id_map, num_rows, topK,
                                                         num_out_tokens);
--- a/transformer_engine/common/permutation/permutation.cu
+++ b/transformer_engine/common/permutation/permutation.cu
@@ -19,12 +19,19 @@ static __global__ void moe_permute_row_map(const int *sorted_row_id, int *row_id
   const int tid = threadIdx.x;
   const int idx = bid * blockDim.x + tid;
 
-  if (idx >= num_out_tokens) return;
+  if (idx >= num_rows * topK) return;
 
   int source_row = sorted_row_id[idx];
   int source_token_id = source_row / topK;
   int source_topK_id = source_row % topK;
-  row_id_map[source_topK_id * num_rows + source_token_id] = idx;
+
+  if (idx >= num_out_tokens) {
+    // Set the indices of dropped tokens to -1
+    row_id_map[source_topK_id * num_rows + source_token_id] = -1;
+  } else {
+    // Create a row id map for subsequent unpermute operation
+    row_id_map[source_topK_id * num_rows + source_token_id] = idx;
+  }
 }
 
 template <typename T, typename TCompute, bool hasProb>
@@ -232,7 +239,7 @@ void nvte_permute_launcher(const T *input, T *output, const int *sorted_row_id, 
     // moe_permute_fwd
 
     int threads = 64;
-    int blocks = (num_out_tokens + threads - 1) / threads;
+    int blocks = (num_rows * topK + threads - 1) / threads;
 
     moe_permute_row_map<<<blocks, threads, 0, stream>>>(sorted_row_id, row_id_map, num_rows, topK,
                                                         num_out_tokens);
@@ -364,6 +371,13 @@ void nvte_device_radix_sort_pairs(void *temp_storage, size_t *temp_storage_bytes
                                   int *keys_out, int *values_in, int *values_out,
                                   size_t num_items) {
   NVTE_API_CALL(nvte_device_radix_sort_pairs);
-  cub::DeviceRadixSort::SortPairs(temp_storage, *temp_storage_bytes, keys_in, keys_out, values_in,
-                                  values_out, num_items);
+  // Sort keys as uint32_t so any negative-int sentinel (e.g. `-1` placed by an
+  // expert-parallel rank mask) becomes a large unsigned value and lands at the
+  // tail of the sorted output, matching the existing capacity-drop convention
+  // (drops encoded as a large positive expert id) and the
+  // `idx >= num_out_tokens` drop branch in moe_permute_row_map.
+  auto *u_keys_in = reinterpret_cast<uint32_t *>(keys_in);
+  auto *u_keys_out = reinterpret_cast<uint32_t *>(keys_out);
+  cub::DeviceRadixSort::SortPairs(temp_storage, *temp_storage_bytes, u_keys_in, u_keys_out,
+                                  values_in, values_out, num_items);
 }
--- a/transformer_engine/pytorch/csrc/extensions/permutation.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/permutation.cpp
@@ -51,13 +51,19 @@ std::tuple<at::Tensor, at::Tensor, std::vector<at::Tensor>> moe_permute_fwd(
       reinterpret_cast<int *>(sorted_indices_ptr), reinterpret_cast<int *>(row_id_ptr),
       reinterpret_cast<int *>(sorted_row_id_ptr), num_tokens * topK);
 
-  // Output buffer alloc
+  // Signed radix sort places -1 sentinel entries (e.g. expert-parallel rank mask)
+  // at the HEAD of sorted_row_id.  Skip that prefix so the kernel sees only the
+  // valid suffix, and pre-fill row_id_map with -1 so the dropped slots are marked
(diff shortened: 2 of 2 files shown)


transformer_engine/pytorch/csrc/extensions/permutation.cpp
```diff
@@ -51,13 +51,22 @@ std::tuple<at::Tensor, at::Tensor, std::vector<at::Tensor>> moe_permute_fwd(
...
+  // Signed radix sort places -1 sentinel entries (e.g. expert-parallel rank mask)
+  // at the HEAD of sorted_row_id.  Skip that prefix so the kernel sees only the
+  // valid suffix, and pre-fill row_id_map with -1 so the dropped slots are marked
+  // without the kernel ever dereferencing a sentinel.
   num_out_tokens = (num_out_tokens > 0) ? num_out_tokens : num_tokens * topK;
+  NVTE_CHECK(num_out_tokens <= num_tokens * topK, "num_out_tokens (", num_out_tokens,
+             ") must not exceed num_tokens*topK (", num_tokens * topK, ")");
+  const int num_minus_ones = num_tokens * topK - num_out_tokens;
+  sorted_row_id_ptr = reinterpret_cast<char *>(sorted_row_id_ptr) +
+                      static_cast<size_t>(num_minus_ones) * sizeof(int);
```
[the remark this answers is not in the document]
tdophung: This (the greptile command) looks correct. Can you please help cast num_tpkens to int64 before multiplication and - num_out_tokens?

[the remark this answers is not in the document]
tdophung: This is not a huge deal because even with topK=128, youwould need > 16M tokens per rank for the int product to overflow. But better to be consistent, and also, this casting of 1 value on the CPU side probably would not slow dow much

transformer_engine/common/permutation/permutation.cu
```diff
@@ -239,7 +232,7 @@ void nvte_permute_launcher(const T *input, T *output, const int *sorted_row_id,
     // moe_permute_fwd
 
     int threads = 64;
-    int blocks = (num_rows * topK + threads - 1) / threads;
+    int blocks = (num_out_tokens + threads - 1) / threads;
```
tdophung: this is correct here but has an implied prerequisite that host prefills the buffer with -1 and shift the ptr by num_minus_ones (what you did in the other file). Better make it more explicit with a comment so no regression will happen by someone accidentally changing this behavior and mess up the number of blocks here. Something like:

"// row_id_map MUST be pre-initialized to -1; sorted_row_id MUST point past the sentinel prefix"

transformer_engine/pytorch/csrc/extensions/permutation.cpp
```diff
@@ -51,13 +51,22 @@ std::tuple<at::Tensor, at::Tensor, std::vector<at::Tensor>> moe_permute_fwd(
...
 
-  // Output buffer alloc
+  // Signed radix sort places -1 sentinel entries (e.g. expert-parallel rank mask)
+  // at the HEAD of sorted_row_id.  Skip that prefix so the kernel sees only the
+  // valid suffix, and pre-fill row_id_map with -1 so the dropped slots are marked
+  // without the kernel ever dereferencing a sentinel.
   num_out_tokens = (num_out_tokens > 0) ? num_out_tokens : num_tokens * topK;
+  NVTE_CHECK(num_out_tokens <= num_tokens * topK, "num_out_tokens (", num_out_tokens,
+             ") must not exceed num_tokens*topK (", num_tokens * topK, ")");
+  const int num_minus_ones = num_tokens * topK - num_out_tokens;
```
tdophung: This is probably going to introduce a regression for the capacity-drop path. This shift assumes the dropped routes are -1 sentinels at the head of sorted_row_id (cub's signed radix sort), which is true for the EP-mask case this PR targets. But the pre-existing capacity-drop path encodes drops as a large positive expert id that sorts to the tail. For that case, the head is valid low-expert-id rows, and shifting past them drops the wrong tokens.(just fyi, capacity-dropping case means no -1 in indices, num_out_tokens < num_tokens * topK because some expert exceeded capacity))

See in this file `tests/pytorch/test_permutation.py`, in `pytorch_permute_index_map`, we have:

 `sorted_indices[:num_out_tokens]` (keeps the head),
so I'd expect test_permutation_index_map[..., num_out_tokens=2039, ...] to fail. We can run the te_ci to  confirm it.

tdophung: I think another solution to this without doing num_tokens * topk - num_out_tokens (or counting the number of -1 on host side) is to sort the keys as `uint32_t` instead of` int32_t`. So,  `-1` becomes `UINT_MAX` and sorts to the tail, unifying both capacity-dropping and dropless under the original `idx >= num_out_tokens -->  drop logic`. That removes the need for the prefix shift you did, and the row_id_map pre-fill. This just needs expert_id to be <= UINT_MAX, which I do not think we are reaching there anytime soon

tdophung (changes requested):

transformer_engine/pytorch/csrc/extensions/permutation.cpp
jing-4369: Thanks for the careful review. Acknowledging the capacity-drop regression concern and the unsigned-sort suggestion below — both make sense. Waiting on the te_ci result you triggered before I push any code change, so we have a concrete signal on what needs to move.

tdophung: Here is the CI pipeline: https://gitlab-master.nvidia.com/dl/transformerengine/transformerengine/-/pipelines/50478896

It failed in the expected tests

jing-4369:
@tdophung Pushed the unsigned-sort approach in 4f46dc25. Net diff is now +13/-6 in permutation.cu only; permutation.cpp is unchanged from upstream, and the earlier comments around NVTE_CHECK / launcher block-count no longer apply.

Could you re-trigger /te-ci pytorch when convenient?

tdophung:
> @tdophung Pushed the unsigned-sort approach in [4f46dc2](https://github.com/NVIDIA/TransformerEngine/commit/4f46dc25ba2647fc78848b480e1b60dfbe20568d). Net diff is now +13/-6 in permutation.cu only; permutation.cpp is unchanged from upstream, and the earlier comments around NVTE_CHECK / launcher block-count no longer apply.
> 
> Could you re-trigger /te-ci pytorch when convenient?

Hi, I can trigger it now. However, for future reference, you could also trigger it by commenting "/te-ci <framework> <level>" on the PR. <framework> is either pytorch or jax, and <level> can be L0, L1, L2. Alternatively, you can test it out locally also,, by running the related tests in tests/pytorch or tests/jax, depending on the change, or using the scripts in qa/L0_pytorch_unittest/test.sh

transformer_engine/common/permutation/permutation.cu
tdophung: +1 to greptile's P1 on moe_permute_row_map / its launcher: now that the kernel is back to walking all num_rows * topK entries, the int * int products are the same overflow class this PR is fixing in the other two kernels. Could you widen those to int64_t for consistency?

tdophung: source_topK_id * num_rows  =. int * int. same as comment on line 242

tdophung: same as comment on line 242

tdophung:
I can see that the CI finished and the failures are not related to your change: https://gitlab-master.nvidia.com/dl/transformerengine/transformerengine/-/pipelines/50584460

the only test failing was fused_qkv not permutation

jing-4369:
It seems my "/te-ci pytorch" does not work? @tdophung could you trigger it on your side once you have a moment? The branch is also rebased onto latest main now.

transformer_engine/common/permutation/permutation.cu
```diff
@@ -19,18 +19,18 @@ static __global__ void moe_permute_row_map(const int *sorted_row_id, int *row_id
   const int tid = threadIdx.x;
   const int idx = bid * blockDim.x + tid;
 
-  if (idx >= num_rows * topK) return;
+  if (idx >= static_cast<int64_t>(num_rows) * topK) return;
 
   int source_row = sorted_row_id[idx];
   int source_token_id = source_row / topK;
   int source_topK_id = source_row % topK;
```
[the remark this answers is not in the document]
jing-4369: I think this analysis swaps cub's keys for values. The call is

```cpp
nvte_device_radix_sort_pairs(
    d_temp_storage, &temp_storage_bytes,
    /* keys_in  */ reinterpret_cast<int *>(indices_ptr),       // routing_map, contains -1
    /* keys_out */ reinterpret_cast<int *>(sorted_indices_ptr),
    /* values_in*/ reinterpret_cast<int *>(row_id_ptr),        // [0, num_tokens*topK)
    /* values_out*/reinterpret_cast<int *>(sorted_row_id_ptr),
    num_tokens * topK);
```

so `sorted_row_id[idx]` is the **values_out** stream — always a linear position in `[0, num_tokens*topK)`, never the sentinel key. The uint32 cast only reorders the keys; the values that ride along stay valid linear positions. For idx ≥ num_out_tokens, `source_row` points to the (token, slot) pair whose routing-map entry was `-1`, and the existing `row_id_map[source_topK_id * num_rows + source_token_id] = -1` write correctly marks that slot as dropped for the downstream unpermute. Adding the early-exit would instead leave row_id_map's dropped slots uninitialized (`torch::empty`), which is what would actually corrupt the unpermute result.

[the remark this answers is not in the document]
tdophung: Agreeing with @jing-4369 . Greptile is wrong here as `sorted_row_id` is value output, while `sorted_indices`  is the key output, not the other way around according to the function signature. resolving

tdophung (approved):
LGTM CI also passed in this pipeline: [50938895](https://gitlab-master.nvidia.com/dl/transformerengine/transformerengine/-/pipelines/50938895)

PR #2912, opened 2026-04-21
[JAX] Add an MoE Block (Layer) that compound router, permutation, groupedGEMM and communication

### LINKED ISSUE #2895: [JAX] Create initial MoE Block
Lots of optimization on the MoE flow are tightly coupled to how Maxtext moe.py is written. Including:
1. Different strategies of EP (ring-of-experts/ragged-all-to-all)
2. Order of communication and computation with regards to different parallelization axis (like quantization before or after FSDP AG)
3. Layout mismatch between steps in MoE flow resulting in extra loop_transpose_fusion overhead with AG.

If we control the MoEBlock, we can experiment with weight layouts and sharding strategies and order to provide a full MoE solution to JAX customers, without being coupled to how Maxtext is written.

This initial MoEBlock would implement the ring-of-experts EP strategy only, as it is the fastest we get currently with the E2E team, before other strategies arrives. Then the Block can be adapted to also house them.

tdophung:
# Description

Most of MoE building blocks integration work has been deeply coupled with Maxtext development. Now creating this MoE block to isolate the work from Maxtext and provide more room for experimentation. `MoEBlock` is a self-contained Flax-Linen module that wires together TE's fused router, pluggable token-dispatch backends (pure-JAX argsort or Triton `sort_chunks_by_index`), `grouped_dense`-based expert FFN, and ragged-all-to-all (A2Av) expert parallelism via `shard_map`

This first iteration will start with ring-of-experts EP, sharding on batch dimention for FSDP, CUBLASLt groupedGEMM and 2 permutation backend: pure JAX or Triton kernels. The block also exposes pluggable knobs for: weight layout (`wi_kernel_axes`/ `wo_kernel_axes`), permutation backend, A2A vs no-EP (single GPU) path, data-parallelism axes for true FSDP (batch sharded across `(ep, fsdp)` simultaneously), top-K with optional grouped/sigmoid scoring (for DSv3 workload), and optional auxiliary load-balancing loss.

Fixes #2895

## Type of change

- [ ] Infra/Build change
- [X] Code refactoring

## Changes
- New `transformer_engine/jax/flax/moe.py` -- `MoEBlock` Linen module:
  gate -> fused topk -> global permute -> A2A EP shard_map (ragged_a2a fwd, local permute, 3x grouped GEMM SwiGLU FFN, local unpermute, ragged_a2a rev) -> global combine.
- Extended `transformer_engine/jax/permutation.py` with A2A param  helpers (`compute_ragged_all_to_all_params`, `compute_reverse_ragged_all_to_all_params`,  `local_permute_after_a2a`, `local_unpermute_before_a2a`) and the  pure-JAX `unfused_token_dispatch` / `unfused_token_combine` paths
  with custom VJPs.
- `tests/jax/test_moe_block.py` -- single-device shape, backward,
  cross-backend equivalence, aux-loss, group-topk, JIT determinism.
- `tests/jax/test_distributed_moe_block.py` -- EP=2 x FSDP=2 mesh test using the canonical Flax-Linen sharded-init pattern (`eval_shape -> get_partition_spec -> logical_to_mesh_sharding -> jit(init, out_shardings=...)`) and `data_parallelism_axes=("fsdp",)` to exercise true FSDP (batch sharded across both axes).

# Checklist:


transformer_engine/pytorch/permutation.py
```diff
@@ -852,7 +876,7 @@ def _moe_unpermute_mask_map_backward_wrapper(ctx, unpermuted_act_grad):
 def moe_permute(
     inp: torch.Tensor,
     routing_map: torch.Tensor,
-    num_out_tokens: int = -1,
+    num_out_tokens: int,
     max_token_num: int = -1,
     map_type: str = "mask",
 ) -> Tuple[torch.Tensor, torch.Tensor]:
```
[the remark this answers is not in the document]
tdophung: This has been resolved after rebasing on top of main

tests/jax/test_distributed_moe_block.py
```diff
@@ -0,0 +1,170 @@
...
+                #      place the params on-device with those shardings.
+                abstract_variables = jax.eval_shape(sharded_block.init, init_key, inputs)
+                logical_partition_spec = nn.get_partition_spec(abstract_variables)
+                out_shardings = nn.logical_to_mesh_sharding(
+                    logical_partition_spec, mesh, logical_axis_rules
+                )
+                sharded_variables = jax.jit(sharded_block.init, out_shardings=out_shardings)(
+                    init_key, inputs
+                )
+                (sharded_loss, (sharded_output, sharded_aux)), sharded_grads = jax.value_and_grad(
```
jberchtold-nvidia: SR: Do we want to `jax.jit` this function before calling it?

tests/jax/test_moe_block.py
```diff
@@ -0,0 +1,310 @@
...
+
+        assert (
+            output.shape == inputs.shape
+        ), f"Unexpected output shape {output.shape} for backend {permutation_backend}"
+        assert output.dtype == inputs.dtype
+        assert jnp.all(jnp.isfinite(output)), "Output contains NaN/Inf"
+        assert aux_loss is None, "aux_loss should be None when aux_loss_coeff=0"
+
+    @pytest.mark.parametrize("permutation_backend", ["pure_jax", "triton"])
+    def test_backward_grad(self, permutation_backend):
```
jberchtold-nvidia: nit SR: rename to `test_backward_grad_is_finite_and_nonzero` or something similar to indicate this test doesn't compare with a reference impl

tests/jax/test_moe_block.py
```diff
@@ -0,0 +1,310 @@
...
+        atol_out, rtol_out = 5e-2, 5e-2
+        assert jnp.allclose(
+            out_pj, out_tr, atol=atol_out, rtol=rtol_out
+        ), f"Forward outputs differ across backends: max diff {jnp.max(jnp.abs(out_pj - out_tr))}"
+        assert jnp.allclose(loss_pj, loss_tr, atol=atol_out, rtol=rtol_out)
+
+        for name in ("gate_kernel", "wi_0", "wi_1", "wo"):
+            g_pj = _unwrap_partitioned(grads_pj["params"][name])
+            g_tr = _unwrap_partitioned(grads_tr["params"][name])
+            assert jnp.allclose(g_pj, g_tr, atol=1e-1, rtol=1e-1), (
```
jberchtold-nvidia: 1e-1 is a pretty high tolerance for most of our tests. What error values of atol and rtol do you typically get from these tests and is that error difference expected between jax/triton backends?

tests/jax/test_moe_block.py
```diff
@@ -0,0 +1,310 @@
...
+        )
+        inputs = _make_inputs(data_key)
+        _variables, output, _aux_loss = _init_and_apply(block, inputs, init_key)
+
+        assert output.shape == inputs.shape
+        assert jnp.all(jnp.isfinite(output))
+
+    @pytest.mark.xfail(
+        reason=(
+            "TE grouped_dense FFI asserts sum(group_sizes) == M at "
```
jberchtold-nvidia: This assertion is only for the V1 grouped GEMM. For the V2 grouped GEMM sum(group_sizes) < M is supported. Can you try the following?

1. Enforce the grouped GEMM with NVTE_JAX_ENFORCE_V2_GROUPED_GEMM=1, similar to what we do here: 
https://github.com/NVIDIA/TransformerEngine/blob/4b6923dd1141272c668a96e9486a5482ec6ada40/tests/jax/utils.py#L1681
2. Remove the functools cache on this function so it can change: https://github.com/NVIDIA/TransformerEngine/blob/4b6923dd1141272c668a96e9486a5482ec6ada40/transformer_engine/jax/cpp_extensions/gemm.py#L2028
3. Run this test within a try/catch. If it runs, great. If the "catch" catches a runtime error that contains the string here, then pytest.skip("V2 grouped gemm is not supported") https://github.com/NVIDIA/TransformerEngine/blob/4b6923dd1141272c668a96e9486a5482ec6ada40/transformer_engine/jax/cpp_extensions/gemm.py#L2170

transformer_engine/jax/permutation.py
```diff
@@ -7,6 +7,17 @@
 This module provides high-level token dispatch and combine operations for
 Mixture of Experts (MoE) models with proper automatic differentiation support.
 
+Two backends are offered:
+
+* Fused, Triton-backed ``token_dispatch`` / ``token_combine`` - uses the
+  Triton kernels in ``transformer_engine.jax.triton_extensions.permutation``.
+* Unfused, pure-JAX ``unfused_token_dispatch`` / ``unfused_token_combine`` -
```
jberchtold-nvidia: I feel like "unfused" implies slower, when in practice this approach is faster at least in MaxText. Would "triton" or "pure_jax" like we have in the tests fit better. What do you think?

transformer_engine/jax/flax/moe.py
```diff
@@ -0,0 +1,974 @@
...
+    # Parallelism
+    expert_parallelism_axis: Optional[str] = None
+    data_parallelism_axes: Tuple[str, ...] = ()
+    tensor_parallelism_axis: Optional[str] = None
+    # ``jax.sharding.Mesh`` to use when ``expert_parallelism_axis`` is set.
+    # Required for the ``shard_map`` wrapper; ignored otherwise.
+    mesh: Optional[Any] = None
+
+    # Permutation
+    permutation_backend: str = "pure_jax"
```
jberchtold-nvidia: Let's make this a Python Enum for better type annotations

transformer_engine/jax/flax/moe.py
```diff
@@ -0,0 +1,974 @@
...
+    align_size: int = 0
```
jberchtold-nvidia: I think we should infer align_size based on the recipe as there isn't a benefit to the user changing it except for getting it to work with a given recipe.

But for this PR, I'd say we make this experimental `__align_size = 128` in this PR and I can make a follow-up to remove the field and infer it from the quantization config

transformer_engine/jax/flax/moe.py
```diff
@@ -0,0 +1,974 @@
...
+    # Sharding
+    gate_kernel_axes: Tuple[Optional[str], ...] = ()
+    wi_kernel_axes: Tuple[Optional[str], ...] = ("exp", "embed", "mlp")
+    wo_kernel_axes: Tuple[Optional[str], ...] = ("exp", "mlp", "embed")
+    input_axes: Tuple[Optional[str], ...] = ()
+
+    # Parallelism
+    expert_parallelism_axis: Optional[str] = None
+    data_parallelism_axes: Tuple[str, ...] = ()
+    tensor_parallelism_axis: Optional[str] = None
```
jberchtold-nvidia: Can be inferred from global MeshResource `global_mesh_resource()` if needed. But also I think we can leave TP as unsupported for now and just focus on EP+FSDP

transformer_engine/jax/flax/moe.py
```diff
@@ -0,0 +1,974 @@
...
+    aux_loss_coeff: float = 0.0
+
+    # Sharding
+    gate_kernel_axes: Tuple[Optional[str], ...] = ()
+    wi_kernel_axes: Tuple[Optional[str], ...] = ("exp", "embed", "mlp")
+    wo_kernel_axes: Tuple[Optional[str], ...] = ("exp", "mlp", "embed")
+    input_axes: Tuple[Optional[str], ...] = ()
+
+    # Parallelism
+    expert_parallelism_axis: Optional[str] = None
```
jberchtold-nvidia: Can be inferred from global MeshResource `global_mesh_resource`. We may need to add a field on MeshResource for expert_axis, I'm not sure if it exists yet.

transformer_engine/jax/flax/moe.py
```diff
@@ -0,0 +1,974 @@
...
+    # Architecture
+    num_experts: int = 8
+    num_experts_per_tok: int = 2
+    intermediate_size: int = 2048
+    activation_type: str = "silu"
+
+    # Routing
+    score_function: Union[str, ScoreFunction] = "softmax"
+    use_pre_softmax: bool = False
+    num_groups: int = -1
```
jberchtold-nvidia: Can we use `None` instead of `-1`?

transformer_engine/jax/flax/moe.py
```diff
@@ -0,0 +1,974 @@
...
+            Scalar auxiliary load-balancing loss when
+            ``aux_loss_coeff > 0``, else ``None``.
+        """
+        assert (
+            inputs.ndim == 3
+        ), f"MoEBlock expects [batch, sequence, hidden] input, got shape {inputs.shape}"
+        inputs = with_sharding_constraint_by_logical_axes(inputs, self.input_axes)
+
+        _, _, hidden_size = inputs.shape
+        params = self._make_params(hidden_size)
```
jberchtold-nvidia: Do we need this "params" dictionary? With `nn.compact` we can define each param right before it's usage inline and nn.compact will handle collectiong all params for us. Most of our other modules in module.py define params inline instead of in a dictionary upfront

transformer_engine/jax/flax/moe.py
```diff
@@ -0,0 +1,974 @@
...
+
+        if self.aux_loss_coeff <= 0.0:
+            aux_loss = None
+        return output, aux_loss
+
+    # ------------------------------------------------------------------
+    # Gate
+    # ------------------------------------------------------------------
+
+    def _gate(self, inputs: jnp.ndarray, gate_kernel: jnp.ndarray) -> jnp.ndarray:
```
jberchtold-nvidia: Is this gating proj GEMM something we should quantize?

transformer_engine/jax/flax/moe.py
```diff
@@ -0,0 +1,974 @@
...
+    # ------------------------------------------------------------------
+    # Global permute (route -> token dispatch)
+    # ------------------------------------------------------------------
+
+    def _global_permute(
+        self,
+        inputs_2d: jnp.ndarray,
+        sparse_probs: jnp.ndarray,
+        routing_map: jnp.ndarray,
+    ) -> dict:
```
jberchtold-nvidia: Can you replace the dict with a dataclass with the same fields

transformer_engine/jax/flax/moe.py
```diff
@@ -0,0 +1,974 @@
...
+        ``shard_map`` is unnecessary here.
+        """
+        batch_size, sequence_length, hidden_size = inputs.shape
+        inputs_2d = inputs.reshape(-1, hidden_size)
+        logits_2d = gate_logits.reshape(-1, self.num_experts)
+
+        sparse_probs, routing_map = self._route_topk(logits_2d, params.get("expert_bias"))
+        aux_loss = self._compute_aux_loss(logits_2d)
+        perm = self._global_permute(inputs_2d, sparse_probs, routing_map)
+        expert_outputs = self._expert_ffn(
```
jberchtold-nvidia: Since the grouped quant and grouped GEMM do not have custom partitioning rules, I think this being outside of shard_map will either raise an error about missing partitioning rules or silently replicate

transformer_engine/jax/flax/moe.py
```diff
@@ -0,0 +1,974 @@
...
+        }
+        if "expert_bias" in params:
+            captured["expert_bias"] = params["expert_bias"]
+            in_specs["expert_bias"] = P(ep_axis)
+        if "wi_0_bias" in params:
+            for name in ("wi_0_bias", "wi_1_bias", "wo_bias"):
+                captured[name] = params[name]
+                in_specs[name] = P(ep_axis, None)
+
+        def _a2a_fn(local: dict) -> Tuple[jnp.ndarray, jnp.ndarray]:
```
jberchtold-nvidia: This is a fairly large function inside another big function. Can we move this to an outer scope or is this required to capture something?

transformer_engine/common/util/multi_stream.cpp
```diff
@@ -12,40 +12,74 @@
...
 #include <mutex>
+#include <unordered_map>
 #include <vector>
 
 #include "cuda_runtime.h"
 #include "logging.h"
 
 namespace transformer_engine::detail {
 
+namespace {
```
tdophung: NOTE: I added this as I run into  `cudaErrorInvalidResourceHandle` (at `cast.cu:112` in `nvte_multi_tensor_quantize`) when trying to launch with 1 process in an independent script (that imports TE moe block) to test MoEBlock with data type FP8. This was because the global cudaStream or Event pool was created lazily via std::call_once, which leaves the resources bound to whichever device arrive first. 

 I fixed this with caching per cudaGetDevice() in an unordered map. Let me know if there is any reason why we should not do this. @jberchtold-nvidia

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -1157,12 +1157,18 @@ Error_Type GroupedGemmFFI(cudaStream_t stream, Buffer_Type lhs_data, Buffer_Type
...
+    // off group_sizes).
     if (!is_rhs_ragged) {
-      NVTE_CHECK(m == sum_group_sizes, "Unexpected group_sizes! M = ", m,
-                 ", got sum(group_sizes)=", sum_group_sizes);
+      NVTE_CHECK(sum_group_sizes <= m, "Unexpected group_sizes! sum(group_sizes)=", sum_group_sizes,
+                 " must be <= M = ", m);
     } else {
-      NVTE_CHECK(k == sum_group_sizes, "Unexpected group_sizes! K = ", k,
-                 ", got sum(group_sizes)=", sum_group_sizes);
+      NVTE_CHECK(sum_group_sizes <= k, "Unexpected group_sizes! sum(group_sizes)=", sum_group_sizes,
```
tdophung: Just pointing at this so reviewers also pay attention to this change (that is different from the initial version previous to the other comments).  When I tested with FP8, which is when the padding to align_size took effect, I start to see these checks firing, to which I then relaxed the checks because I think it should allow for garbage data on dim m to exist when there is worst case padding.

jberchtold-nvidia: This is for the V1 grouped GEMM FFI, not the one we are using now that binds to cuBLASLt. The V2 does support k >= group sizes and I have tested the V2 with it thoroughly. In theory, I think relaxing this constraint for V1 shouldn't cause issues, but I have not tested it so I am not sure.

In your message above, is FP8 = tensor-scaled FP8? If so, the reason that triggers this assertion is we don't support tensor-scaled FP8 for the V2 grouped quant + GEMM so it falls through to the old V1 codepath.

tdophung: Yes I think I tested with current scaling that's why it hit this V1 implementation. I have changedto MXFP8_1D_SCALING in my test script now. Good catch

tests/jax/test_moe_block.py
```diff
@@ -215,6 +227,134 @@ def test_aux_loss_returned(self, permutation_backend):
         # With uniform-ish routing the loss should be small-positive, not huge.
         assert jnp.abs(aux_loss) < 1e2
 
+    def test_aux_loss_uses_real_routing_under_group_topk(self):
+        """Regression test for PR #2912 review (greptile P1).
```
jberchtold-nvidia: SR: Remove this "Regression test for PR #2912 review (greptile P1).", I don't think we need those specifics. The other comments are useful though.

transformer_engine/jax/csrc/extensions/gemm.cpp
jberchtold-nvidia: @tdophung can we revert this assertion relaxing since the test above was modified to only run if the V2 grouped GEMM was available? Unless you've already tried this and it works okay, then we can update the test above to run on both V1/V2 grouped GEMMs. But if it doesn't' work or you haven't tried yet, let's revert this change so we can focus on V2.

transformer_engine/jax/flax/moe.py
```diff
@@ -87,7 +89,36 @@
 Initializer = Callable[[PRNGKey, Shape, DType], Array]
 
 
-__all__ = ["MoEBlock"]
+__all__ = ["GlobalPermuteResult", "MoEBlock"]
+
+
+# =============================================================================
```
jberchtold-nvidia: SR: Move this more detailed description into the docstring so it gets included in our generated docs

tests/jax/test_moe_block.py
jberchtold-nvidia: @tdophung following up on this tolerance

transformer_engine/common/util/multi_stream.cpp
jberchtold-nvidia: @tdophung Let's move this into a separate PR. This is an issue in the V1 codepath but not V2. It's a good fix and the code seems reasonable to me but it could impact PyTorch so I'd like to decouple from this PR to unblock MoEBlock+V2 and then we can have PyTorch and core review and test this fix separately.

transformer_engine/jax/flax/moe.py
jberchtold-nvidia: @tdophung following up on this

transformer_engine/jax/flax/moe.py
jberchtold-nvidia: @tdophung following up on this, can we make it `__align_size` for now. I will hopefully have the follow-up PR to infer it from the quantization config before our next code freeze but just in case I don't want us to make this part of the public API

transformer_engine/jax/flax/moe.py
jberchtold-nvidia: @tdophung Let's remove this field for expert_parallelism axis and make the following changes instead

1. Add `ep_resource` to MeshResource here https://github.com/NVIDIA/TransformerEngine/blob/708d7c160ad6b2bf44c9c597083d4cbb4860f068/transformer_engine/jax/sharding.py#L322 
2. Where self.expert_parallelism_axis is used below, replace with global_mesh_resource().ep_resource. You can use `get_mesh_axis_size` to determine how many GPUs are on the axis and whether we are using EP or not

transformer_engine/jax/flax/moe.py
jberchtold-nvidia: @tdophung Let's remove TP support for now

phu0ngng (changes requested):
I think we should go with exposing GroupMLP VJP first before the MoE module to enable future possible fusions.

tdophung:
Changing back to draft to not spam people's email while I push commits to this branch for the full unrolling of ops in a big VJP.

tests/jax/test_moe_vjp.py
```diff
@@ -0,0 +1,449 @@
...
+
+    return jax.grad(aux_only)(params, x)
+
+
+# -----------------------------------------------------------------------------
+# Tests
+# -----------------------------------------------------------------------------
+
+
+@pytest.mark.triton
```
jberchtold-nvidia: Will this skip these tests if triton isn't installed? If so, we should do this manually and only skip when backend_name  == triton

jberchtold-nvidia: See comment on permutation.py about removing triton dependency for pure-JAX permute backend

tests/jax/test_moe_vjp.py
```diff
@@ -0,0 +1,449 @@
...
+    """Forward shape / finiteness / parity vs pure-JAX reference."""
+
+    @pytest.mark.parametrize("backend_name", ["pure_jax", "triton"])
+    def test_forward_shape_and_finite(self, backend_name):
+        backend = PermutationBackend(backend_name)  # noqa: F821
+        key = jax.random.PRNGKey(0)
+        kp, kx = jax.random.split(key)
+        params = _init_params(kp)
+        x = _make_inputs(kx)
+        out, aux = _run_te_moe(x, params, permutation_backend=backend)
```
jberchtold-nvidia, line 271: Do we need to gate these tests by GPU arch? The grouped GEMM currently only supports blackwell

transformer_engine/jax/permutation.py
```diff
@@ -7,6 +7,19 @@
...
+* Triton-backed ``token_dispatch`` / ``token_combine`` - uses the
+  Triton kernels in ``transformer_engine.jax.triton_extensions.permutation``.
+* Pure-JAX ``pure_jax_token_dispatch`` / ``pure_jax_token_combine`` - uses
+  only ``jnp.argsort`` + gather and is therefore compiled as plain XLA.
+  Despite the name, this path is often *faster* than the Triton kernels in
+  current testing because XLA can fuse the ops with surrounding work.
+
+Both backends support optional alignment padding (``align_size > 0``) so each
+expert's group size is a multiple of ``align_size``, which is required for
+quantized grouped GEMMs.
```
jberchtold-nvidia, line 21: Now that `permutation.py` has two backends, I think we should update this file to remove our triton dependency if using the pure-JAX backend.

I think we can do this in the following way:
1. Gate imports from triton_extensions here to only apply if triton is installed and defer any errors about  missing triton installation until the  triton backend is invoked, no errors for the pure-JAX backend.
2. Then `transformer_engine/jax/moe.py` and `transformer_engine/jax/flax/moe.py` no longer depend on triton
3. We can remove the `_inject_moe` logic from the test files and also remove the `#noqa: F821` linter suppresions that `_inject_moe` requires.

transformer_engine/jax/permutation.py
```diff
@@ -655,3 +677,642 @@ def _sort_chunks_by_index_bwd_rule(
...
+#
+# The two ``compute_*_ragged_all_to_all_params`` functions translate
+# ``all_shards_tokens_per_expert`` (an EP-axis ``all_gather`` of each shard's
+# global ``group_sizes``) into the four ``ragged_all_to_all`` arguments
+# (``input_offsets``, ``send_sizes``, ``output_offsets``, ``recv_sizes``).
+# ``shard_id`` may be a traced value (e.g. from :func:`jax.lax.axis_index`),
+# which is why every slice into ``all_shards_tokens_per_expert`` uses
+# :func:`jax.lax.dynamic_slice`.
+#
+# These functions are pure JAX (no MaxText / TE dependencies) and equivalent
```
jberchtold-nvidia: I don't think we need to reference our functions we've added on our MaxText fork. Those will be replaced by our implementation in TE once this is ready.

transformer_engine/jax/moe.py
```diff
@@ -0,0 +1,2019 @@
...
+# =============================================================================
+# Enums
+# =============================================================================
+
+
+class PermutationBackend(Enum):
+    """Token-dispatch / combine backend used by :func:`moe`.
+
+    * ``PURE_JAX``: ``jnp.argsort`` + gather paths compiled as plain XLA;
+      typically faster than ``TRITON`` in current testing because XLA can
```
jberchtold-nvidia: SR: this commend may need updating

transformer_engine/jax/moe.py
```diff
@@ -0,0 +1,2019 @@
...
+from .quantize import (
+    QuantizerSet,
+    ScaledTensor,
+    TensorUsage,
+    noop_quantizer_set,
+    with_sharding_constraint_by_logical_axes,
+)
+from .router import ScoreFunction, _validate_score_function
+from .sharding import _get_mesh
+from .triton_extensions.permutation import (
```
jberchtold-nvidia: Do we need to import triton_extensions directly here?

transformer_engine/jax/moe.py
```diff
@@ -0,0 +1,2019 @@
...
+        by ``pure_jax_token_dispatch``.
+    pre_a2a_buffer_shape : tuple[int, int]
+        ``(num_real_tokens + padding_size, hidden)`` -- the per-shard
+        shape of the sorted-inputs buffer that is sent over the EP
+        ragged_all_to_all in the fwd direction.
+    post_a2a_buffer_shape : Optional[tuple[int, int]]
+        ``(recv_buffer_rows, hidden)`` when EP is active, ``None``
+        otherwise.
+    """
+    import math
```
jberchtold-nvidia: SR: move `math` import to top of file

transformer_engine/jax/moe.py
```diff
@@ -0,0 +1,2019 @@
...
+    caller can pass them as ordinary static keyword args. See the
+    module-level comment above for why this matters.
+
+    ``batch_is_per_shard`` controls whether ``batch_size`` is already
+    sharded (True -- e.g. when this is called from inside a shard_map
+    body, where ``x.shape[0]`` reports the per-shard batch size) or
+    global (False -- e.g. when computing from x.shape outside the
+    shard_map body).
+
+    Keys
```
jberchtold-nvidia: Instead of a dict let's use a dataclass with these fields

transformer_engine/jax/moe.py
```diff
@@ -0,0 +1,2019 @@
...
+            num_tokens,
+            num_experts,
+            hidden,
+        )
+    return out_2d.reshape(batch_size, sequence_length, hidden).astype(dtype)
+
+
+def _combine_bwd(
+    d_output: jnp.ndarray,
+    state: dict,
```
jberchtold-nvidia: Can we make this state dict a dataclass? If the fields different between pure-JAX  and Triton backends we could make two dataclasses and annotate the type as `Union[Type1, Type2]`

transformer_engine/jax/moe.py
```diff
@@ -0,0 +1,2019 @@
...
+        casted_sorted_x_lhs_trans = casted_sorted_x_lhs_trans.checkpoint(q_set_w0.x)
+    if isinstance(casted_wi_0_rhs_trans, ScaledTensor):
+        casted_wi_0_rhs_trans = casted_wi_0_rhs_trans.checkpoint(q_set_w0.kernel)
+
+    # GEMM 2: layer_w1 = sorted_x @ wi_1
+    casted_sorted_x_w1 = tex.grouped_quantize(
+        sorted_x, q_set_w1.x, local_group_sizes, flatten_axis=-1
+    )
+    casted_wi_1 = tex.grouped_quantize(wi_1, q_set_w1.kernel, flatten_axis=-1)
+    layer_w1 = tex.grouped_gemm(
```
jberchtold-nvidia: What does the `w` mean in `layer_w1`? When I see `w` I think weights but this is an intermediate

transformer_engine/jax/moe.py
```diff
@@ -0,0 +1,2019 @@
...
+        bias=wi_0_bias,
+    )
+    casted_sorted_x_lhs_trans = casted_sorted_x_w0.get_tensor(usage=TensorUsage.LHS_TRANS)
+    casted_wi_0_rhs_trans = casted_wi_0.get_tensor(usage=TensorUsage.RHS_TRANS)
+    if isinstance(casted_sorted_x_lhs_trans, ScaledTensor):
+        casted_sorted_x_lhs_trans = casted_sorted_x_lhs_trans.checkpoint(q_set_w0.x)
+    if isinstance(casted_wi_0_rhs_trans, ScaledTensor):
+        casted_wi_0_rhs_trans = casted_wi_0_rhs_trans.checkpoint(q_set_w0.kernel)
+
+    # GEMM 2: layer_w1 = sorted_x @ wi_1
```
jberchtold-nvidia: It shouldn't be that big of a change to fuse GEMMs 1 and 2, right? Our activation function supports gated activations already and having a single up-proj GEMM will make it easier to fuse act+quant in the future

As for the weight tensors, we could enforce a single contiguous wi weight tensor instead of wi_0 and wi_1. I don't think we need to support both fused and unfused up-proj

transformer_engine/jax/moe.py
```diff
@@ -0,0 +1,2019 @@
...
+        num_real_tokens=_static_shape["num_real_tokens"],
+        padding_size=_static_shape["padding_size"],
+        pre_a2a_buffer_shape=_static_shape["pre_a2a_buffer_shape"],
+        ep_axis=ep_axis,
+        shard_id=shard_id,
+        num_ep=num_ep,
+    )
+
+    # ---------------- Build ctx dict ----------------
+    ctx: dict = {
```
jberchtold-nvidia: There are a lot of fields here, so I agree a dict is better than a tuple. But I think it'd be even easier to read if this was a dataclass

transformer_engine/jax/moe.py
```diff
@@ -0,0 +1,2019 @@
...
+        batch_is_per_shard=False,
+    )
+
+    # Compute per-shard input shape: under the EP shard_map body, the
+    # gradient tensors live at per-shard shape, so the dispatch_bwd
+    # reshape target and ``d_x_from_dispatch.reshape(x_shape)`` below
+    # must use the per-shard shape rather than the captured global
+    # ``x_shape``.
+    if ep_active:
+        import math as _math  # local import keeps the no-EP path zero-overhead.
```
jberchtold-nvidia: I think we can import math at the top of the file. We don't have to import inline or gate it behind any logic since it's a built-in library

transformer_engine/jax/moe.py
```diff
@@ -0,0 +1,2019 @@
...
+
+
+def _moe_fwd_rule(
+    # IMPORTANT — calling convention for jax.custom_vjp fwd rule.
+    #
+    # JAX uses ``_argnums_partial`` (jax/_src/api_util.py) when wiring up
+    # the fwd rule. That helper preserves the ORIGINAL positional order
+    # of the decorated function: dyn (= diff) args sit at their original
+    # positions and static (= nondiff) args fill the remaining slots in
+    # nondiff_argnums order. So the fwd rule MUST take args in the
```
jberchtold-nvidia: This feels pretty verbose. Do we need this comment or was this WIP notes from an agent and we can remove it now.

transformer_engine/jax/moe.py
tdophung: yeah that's confusing, it means the intermediate result that was derived from using w1, I changed layer_w0 to gate_proj_out, and layer_w1 to up_proj_out.

transformer_engine/jax/moe.py
tdophung: it's WIP from an agent, trimmed it down to 2 lines now

transformer_engine/jax/permutation.py
tdophung: removed

tests/jax/test_moe_vjp.py
tdophung: yes, I now gated it with `get_device_compute_capability`

transformer_engine/jax/flax/moe.py
```diff
@@ -0,0 +1,1174 @@
...
+        # Examples:
+        #   data_parallelism_axes=()              -> P('ep', None, None)
+        #   data_parallelism_axes=('fsdp',)       -> P(('ep','fsdp'), None, None)
+        #   data_parallelism_axes=('fsdp','data') -> P(('ep','fsdp','data'), ...)
+        for ax in self.data_parallelism_axes:
+            if ax not in mesh.shape:
+                raise ValueError(
+                    f"data_parallelism_axes contains {ax!r} but mesh has"
+                    f" axes {tuple(mesh.shape.keys())}"
+                )
```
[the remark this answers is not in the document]
tdophung: moved during the refactor to big VJP, resolving

transformer_engine/jax/permutation.py
tdophung: Can we do this in a follow up PR. I think it is not really blocking this and this is blocking other work so might as well merge this and have this in the follow up PR

tests/jax/test_moe_vjp.py
tdophung: Can we resolve this and keep the other triton dependency comment for the follow up PR?

transformer_engine/jax/permutation.py
jberchtold-nvidia: Good point, keeping this PR moving and doing this in a follow-up sounds like a better plan! Might happen naturally when we switch to TE EP anyways

tests/jax/test_moe_vjp.py
jberchtold-nvidia: I'm okay for us to do all the triton dependency refactoring in a follow-up. You can ignore this comment for this PR

jberchtold-nvidia:
fyi @tdophung I've removed the `community-contributed` tag the bot added to this PR. I think you added some commits authored with nvjax, the default git author in JAX containers, instead of your git config info and the bot picked this up as a community contribution

transformer_engine/jax/moe.py
```diff
@@ -0,0 +1,2019 @@
...
+        axis_name=ep_axis,
+    )
+    # Step 1 (EP) inverse: combine fwd applied is_forward=False; the
+    # bwd is is_forward=True with the SAME row_id_map.
+    recv_buffer_rows, hidden = d_x_send_back.shape
+    d_expert_outputs, _ = sort_chunks_by_map(
+        d_x_send_back,
+        state["local_perm_row_id_map"],
+        None,
+        recv_buffer_rows,
```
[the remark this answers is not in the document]
jberchtold-nvidia: @tdophung is this comment valid? I see your recent commits fixed a Triton+EP issue that looks similar. Do we also need to make a corresponding fix to the pure-JAX version?

transformer_engine/jax/moe.py
tdophung: fused in latest commit

transformer_engine/jax/moe.py
tdophung: changed to dataclass

jberchtold-nvidia (dismissed):

qa/L0_jax_distributed_unittest/test.sh
```diff
@@ -37,6 +37,12 @@ wait
 TE_PATH=$TE_PATH bash $TE_PATH/examples/jax/collective_gemm/run_test_cgemm.sh || test_fail "run_test_cgemm.sh"
 wait
 
+# MoE custom_vjp distributed suite. Runs one Python process per GPU
+# via tests/jax/run_multiprocess_moe_vjp.sh (mirrors the pattern in
+# examples/jax/encoder/run_test_multiprocessing_encoder.sh). Requires
+# >=4 visible GPUs.
+TE_PATH=$TE_PATH bash $TE_PATH/tests/jax/run_multiprocess_moe_vjp.sh \
```
phu0ngng, line 44: Just to learn, why do we want/need to run this test with multiple processes?

phu0ngng (dismissed):

qa/L0_jax_distributed_unittest/test.sh
tdophung: I encounter a bug similar to one described here: https://docs.google.com/document/d/18XCO7WNbTvadIG1WXPpO409dmu7HQaRlHG02mm5TYDA/edit?tab=t.0#heading=h.k4ivwttoxnil, but for triton kernels, where it would hang. More details were discussed in the TE/JAX meeting

Running the test with multiple process (1 process per GPU) circumvents that

jberchtold-nvidia (approved):

PR #2916, opened 2026-04-22
[PyTorch][CP] Reduce P2P forward peak memory: O(C) _ O(1)

sudhakarsingh27:
## Summary

Reduce P2P attention forward temporary storage from O(CP) to O(1) with two independent optimizations:

1. **KV storage:** rotate two communication buffers instead of retaining one per CP step, saving 1.50 GiB per GPU for the profiled CP8 workloads.
2. **Output/LSE storage:** rotate two partial-result slots instead of retaining every step's output and LSE, saving another 1.52-3.05 GiB per GPU depending on causal rank.

The second optimization requires an online merge because each partial result must be consumed before its slot is reused. When a new step changes the combined LSE, the existing accumulator is renormalized and the new partial output is added. The THD correction kernel performs both operations in one traversal.

Together, the two optimizations reduce retained tensor payload by 3.02-4.55 GiB per GPU for these CP8 workloads.

`rng_states` and `attn_biases` remain CP-sized because backward requires them. The existing output format is preserved for downstream fused and all-to-all paths.

## Type of change

- [x] Breaking change (the experimental `nvte_cp_thd_out_correction` C API gains an `old_lse` argument)

## Performance

H100, CP8, BF16 THD, causal P2P FusedAttention, eager execution. End-to-end values are synchronized means from five paired runs on the same node.

| Fixed-token workload | Pre-PR | Final online merge | Change |
|---|---:|---:|---:|
| 16 x 32K | 152.476 ms | 152.360 ms | -0.08% |
| 64 x 8K | 70.593 ms | 70.677 ms | +0.12% |
| 256 x 2K | 57.992 ms | 57.670 ms | -0.55% |

The online merge is effectively performance-neutral end to end. Its direct correction cost returns to the pre-PR level while retaining bounded storage:

| Fixed-token workload | Pre-PR correction | Separate rescale | Fused correction |
|---|---:|---:|---:|
| 16 x 32K | 3.73 ms | 13.73 ms | 3.81 ms |
| 64 x 8K | 3.72 ms | 13.36 ms | 3.72 ms |
| 256 x 2K | 3.75 ms | 13.60 ms | 3.62 ms |
| Launches per iteration | 8 | 28 | 7 |

Correction values are cross-rank means of GPU kernel durations over the final five of seven iterations. The separate-rescale and fused variants were profiled together; the pre-PR values are a prior-profile reference.

## Memory

The fused correction keeps the same two-buffer/two-slot storage model and removes the separate rescale temporary. Rank-max allocator measurements for `B=2`, `S=262144`, `H=16`, `G=16`, and `D=128`:

| CP | Forward peak before | Bounded forward peak | Forward saving | E2E peak before | Bounded E2E peak | E2E saving |
|---:|---:|---:|---:|---:|---:|---:|
| 4 | 8752 MiB | 7704 MiB | 1048 MiB (12.0%) | 10268 MiB | 10268 MiB | 0 MiB |
| 8 | 7464 MiB | 3860 MiB | 3604 MiB (48.3%) | 7464 MiB | 5134 MiB | 2330 MiB (31.2%) |

## Numerics

BSHD/SBHD accumulation remains in the partial-output dtype (BF16/FP16 for non-FP8; FP8 partial outputs are dequantized to FP32). THD also retains the partial-output dtype; its fused correction computes each merge in FP32 and writes the result back to that dtype. The online merge changes operation ordering relative to post-loop correction, and the CP4/CP8 forward/backward comparisons remain within the existing test tolerances.

## Validation

- CP4 and CP8 BF16 THD causal P2P FusedAttention forward/backward correctness passed on all ranks, including output, dQ, dK, and dV comparisons.
- Independent allocator runs reproduced the CP4 and CP8 memory results.
- Python syntax validation and `git diff --check` pass.
- [CI checks](https://github.com/NVIDIA/TransformerEngine/pull/2916/checks) will run on the updated branch.

### DIFF
--- a/transformer_engine/common/include/transformer_engine/fused_attn.h
+++ b/transformer_engine/common/include/transformer_engine/fused_attn.h
@@ -490,7 +490,8 @@ void nvte_cp_thd_read_second_half_lse(const NVTETensor &lse, const NVTETensor &c
  *
  *  \param[out]    out                   Output tensor.
  *  \param[in]     out_per_step          THD format output of context parallelism in forward pass.
- *  \param[in]     lse                   Softmax LSE.
+ *  \param[in]     old_lse               Softmax LSE before adding this step.
+ *  \param[in]     lse                   Softmax LSE after adding this step.
  *  \param[in]     lse_per_step          Softmax LSE per step.
  *  \param[in]     cu_seqlens            Cumulative sequence lengths, [batch_size + 1].
  *  \param[in]     only_second_half      Whether or not to correct only second half.
@@ -498,9 +499,9 @@ void nvte_cp_thd_read_second_half_lse(const NVTETensor &lse, const NVTETensor &c
  *  \param[in]     stream                CUDA stream used for this operation.
  */
 void nvte_cp_thd_out_correction(NVTETensor out, const NVTETensor &out_per_step,
-                                const NVTETensor &lse, const NVTETensor &lse_per_step,
-                                const NVTETensor &cu_seqlens, int only_second_half, int lse_packed,
-                                cudaStream_t stream);
+                                const NVTETensor &old_lse, const NVTETensor &lse,
+                                const NVTETensor &lse_per_step, const NVTETensor &cu_seqlens,
+                                int only_second_half, int lse_packed, cudaStream_t stream);
 
 /*!  \brief Update the two halves of each packed THD sequence during context-parallel backward.
  *
--- a/transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
+++ b/transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
@@ -173,6 +173,41 @@ def flash_attn_fwd_second_half_softmax_lse_correction(
 
 
 @jit_fuser
+def flash_attn_fwd_incremental_out_correction(
+    out: torch.Tensor,
+    out_per_step: torch.Tensor,
+    old_softmax_lse: torch.Tensor,
+    new_softmax_lse: torch.Tensor,
+    softmax_lse_per_step: torch.Tensor,
+    seq_dim: int,
+):
+    """Online softmax merge: rescale accumulated output and add new step's contribution."""
+    scale_old = torch.exp(old_softmax_lse - new_softmax_lse).movedim(2, seq_dim).unsqueeze(-1)
+    scale_new = torch.exp(softmax_lse_per_step - new_softmax_lse).movedim(2, seq_dim).unsqueeze(-1)
+    out.mul_(scale_old)
+    out.addcmul_(out_per_step, scale_new)
+
+
+@jit_fuser
+def flash_attn_fwd_incremental_second_half_out_correction(
+    out: torch.Tensor,
+    out_per_step: torch.Tensor,
+    old_softmax_lse: torch.Tensor,
+    new_softmax_lse: torch.Tensor,
+    softmax_lse_per_step: torch.Tensor,
+    seq_dim: int,
+):
+    """Online softmax merge for second-half tokens only (causal upper-triangle steps)."""
+    out_ = out.select(seq_dim, 1)
+    old_lse_ = old_softmax_lse.view(*old_softmax_lse.shape[:-1], 2, -1)[..., 1, :]
+    new_lse_ = new_softmax_lse.view(*new_softmax_lse.shape[:-1], 2, -1)[..., 1, :]
+    scale_old = torch.exp(old_lse_ - new_lse_).movedim(2, seq_dim).unsqueeze(-1)
+    scale_new = torch.exp(softmax_lse_per_step - new_lse_).movedim(2, seq_dim).unsqueeze(-1)
+    out_.mul_(scale_old)
+    out_.addcmul_(out_per_step, scale_new)
+
+
+@jit_fuser
 def get_cu_seqlens_on_cp_rank(
     cu_seqlens: torch.Tensor,
     cu_seqlens_padded_on_cp_rank: torch.Tensor,
@@ -1347,7 +1382,7 @@ class AttnFuncWithCPAndKVP2P(torch.autograd.Function):
         amax_per_step = None
         S_quantizer_per_step = [None for _ in range(cp_size)]
         O_quantizer_per_step = [None for _ in range(cp_size)]
-        max_logit_per_step = [None for _ in range(cp_size)]
+        max_logit_per_step = [None, None]
         max_logit = None
 
         assert isinstance(k, q.__class__) and isinstance(
@@ -1439,7 +1474,7 @@ class AttnFuncWithCPAndKVP2P(torch.autograd.Function):
                 fused_attn_backend = FusedAttnBackend["F16_arbitrary_seqlen"]
             if return_max_logit:
                 max_logit_per_step = [
-                    torch.empty(q.shape[-2], dtype=q.dtype, device=q.device) for _ in range(cp_size)
+                    torch.empty(q.shape[-2], dtype=q.dtype, device=q.device) for _ in range(2)
                 ]
 
         # split qkv to two halves and prepare for load balancing
@@ -1547,8 +1582,8 @@ class AttnFuncWithCPAndKVP2P(torch.autograd.Function):
         # set up inputs for forward
         q_inputs = [None, None]
         kv_inputs = [None, None]
-        out_per_step = [None for _ in range(cp_size)]
-        softmax_lse_per_step = [None for _ in range(cp_size)]
+        out_per_step = [None, None]
+        softmax_lse_per_step = [None, None]
         rng_states = [None for _ in range(cp_size)]
         attn_biases = [None for _ in range(cp_size)]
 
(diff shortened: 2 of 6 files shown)


transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -490,17 +490,18 @@ void nvte_cp_thd_read_second_half_lse(const NVTETensor &lse, const NVTETensor &c
...
  *  \param[in]     lse_packed            Whether or the softmax LSE is in packed format.
  *  \param[in]     stream                CUDA stream used for this operation.
  */
 void nvte_cp_thd_out_correction(NVTETensor out, const NVTETensor &out_per_step,
-                                const NVTETensor &lse, const NVTETensor &lse_per_step,
-                                const NVTETensor &cu_seqlens, int only_second_half, int lse_packed,
-                                cudaStream_t stream);
+                                const NVTETensor &old_lse, const NVTETensor &lse,
+                                const NVTETensor &lse_per_step, const NVTETensor &cu_seqlens,
+                                int only_second_half, int lse_packed, cudaStream_t stream);
```
cyanguwa, line 776: This is a breaking change - not a problem, but maybe check the box in the PR description where it says this contains breaking changes. Thanks!

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2216,59 +2298,14 @@ def forward(
...
-                        cu_seqlens_q_padded,
-                        True,
-                        softmax_lse_in_packed_format,
-                    )
+        if qkv_format == "bshd":
+            out = out.view(out.shape[0], -1, *out.shape[-2:])
+            ctx.batch_size = out.shape[0]
+        elif qkv_format == "sbhd":
+            out = out.view(-1, *out.shape[-3:])
+            ctx.batch_size = out.shape[1]
```
cyanguwa: Do we need L2301-2306? It seems like we don't use ctx.batch_size anywhere, and also `out.view` is overwritten by L2309? Nit.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2216,59 +2298,14 @@ def forward(
...
-                elif o_format == "thd":
-                    tex.thd_out_correction(
-                        out,
-                        out_per_step[i],
-                        softmax_lse,
-                        softmax_lse_per_step[i],
-                        cu_seqlens_q_padded,
-                        True,
-                        softmax_lse_in_packed_format,
-                    )
```
cyanguwa, line 2272: Should we remove `flash_attn_fwd_out_correction_init`, `flash_attn_fwd_out_correction`, and `flash_attn_fwd_second_half_out_correction` then? Do we still call them anywhere else?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2138,74 +2173,121 @@ def forward(
...
                         if qkv_format == "thd":
-                            if fp8:
-                                out = torch.zeros_like(out_per_step[0]).view(o_shape)
-                            else:
-                                out = torch.zeros(o_shape, dtype=q.dtype, device=q.device)
+                            # Keep THD in the kernel's input dtype; thd_out_correction
+                            # performs its own promotion and requires matching dtypes.
+                            out = out_per_step[0].clone().view(o_shape)
+                        elif qkv_format in ["bshd", "sbhd"]:
+                            out = out_per_step[0].to(torch.float32)
```
cyanguwa: AI floated this precision change in the accumulator - could you please take a look? It may not be a serious issue, but we should document it (and/or keep it consistent with before, unless there are reasons for changing the accumulation dtype). Thanks!

```
The bshd/sbhd accumulator silently moved from bf16 to fp32, and THD went the other way. This is the change I'd most want explained in the description. On main, flash_attn_fwd_out_correction_init ends with out_corrected.to(out_init_step.dtype), so for non-FP8 the accumulator was bf16 for the whole correction loop — the # fwd output correction: out in torch.float32 comment above it was stale. The PR now does out = out_per_step[0].to(torch.float32), so it's genuinely fp32. That's a real accuracy improvement, but it costs an extra 2 bytes per output element in steady state, which eats into the reported savings and may explain the 0 MiB E2E result at CP4.

Meanwhile THD stays in out_per_step's dtype (bf16), because thd_out_correction_helper asserts out.dtype() == out_per_step.dtype(). So THD keeps a bf16 accumulator and now absorbs an extra rounding of the entire accumulator on every step from p[k] = old_out + out_per_step, where main only ever did one add per step against a fixed final scale. At CP8 that's eight extra roundings of the running sum. Either lift the dtype restriction so THD can accumulate in fp32 too, or show that the existing tolerances still hold at CP8 — and either way, state the bshd/sbhd promotion explicitly, since it's the reason existing numerics tests will shift.
```

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2138,74 +2173,121 @@ def forward(
...
                 # wait until fwd results correction of last step is done
                 if i > 1:
                     flash_attn_streams[(i - 1) % 2].wait_event(fwd_results_correction_done)
 
                 with torch.cuda.stream(flash_attn_streams[(i - 1) % 2]):
                     if use_fused_attention:
-                        # [b, h, sq, 1] -> [b, h, sq]
-                        # [t, h, 1] -> [t, h]
-                        softmax_lse_per_step[i - 1].squeeze_(-1)
+                        # [b, h, sq, 1] -> [b, h, sq] or [t, h, 1] -> [t, np]
```
cyanguwa: I think we removed all notations of `np` in context_parallel.py. Does `np` here just mean `h`?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -1936,7 +1968,10 @@ def forward(
         # MXFP8/F16 attention:    q, k, v: torch.Tensor, dtype=fwd_nominal_dtype
         # FP8DS/CS attention:     q, k, v: torch.Tensor, dtype=torch.uint8
         out = None
+        # Current fused/A2A plumbing still consumes this original format; the
+        # online merge only changes the lifetime of per-step outputs.
```
cyanguwa: Is this comment critical enough to stay here or it'll become some noise after the merge? I know it helps understand the PR. Thanks.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2216,59 +2298,14 @@ def forward(
...
-                        softmax_lse_in_packed_format,
-                    )
+        if qkv_format == "bshd":
+            out = out.view(out.shape[0], -1, *out.shape[-2:])
+            ctx.batch_size = out.shape[0]
+        elif qkv_format == "sbhd":
+            out = out.view(-1, *out.shape[-3:])
+            ctx.batch_size = out.shape[1]
+        # The backward context needs the rank-local output before any A2A
+        # restoration, matching the pre-existing post-loop correction path.
```
cyanguwa: Same for this comment. I feel it's quite specific to this PR?

transformer_engine/common/include/transformer_engine/fused_attn.h
sudhakarsingh27: Done—marked the experimental C API signature update as a breaking change in the PR description.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Agreed—removed the redundant views and unused ctx.batch_size assignments.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Yes—removed all three; they no longer have callers.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Good catch. Restored BSHD/SBHD accumulation to the partial-output dtype, documented the numerics, and validated CP8 BF16 BSHD/SBHD/THD forward and backward against no-CP on all ranks.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Yes—changed [t, np] to [t, h].

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Reworded it as a durable invariant: preserve the original format for fused output and post-attention A2A.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Reworded it as a durable invariant: save the rank-local output for backward before restoring the A2A layout.

cyanguwa (approved):

PR #2919, opened 2026-04-23
guard fuser grad checks on non-leaf nodes

CarlosGomes98:
# Description

Pass an explicit flag that controls whether the fuser forward pass sets requires_grad on outputs. This is required so that in no_grad mode we dont try to mutate this information on non-leaf nodes (which is not allowed by torch)

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/ops/fuser.py
+++ b/transformer_engine/pytorch/ops/fuser.py
@@ -509,14 +509,14 @@ class OperationFuser:
         # Fuser forward pass
         # When is_grad_enabled is False, we call forward directly.
         # This does not register a PyTorch autograd node, so
-        # no fuser backward will run. We pass set_output_requires_grad=False 
+        # no fuser backward will run. We pass set_output_requires_grad=False
         # to avoid setting requires_grad on outputs in
         # this path since they may be non-leaf tensors from the inner ops.
         args = (
             input,
             self,
             basic_op_kwargs,
-            is_grad_enabled, # set_output_requires_grad
+            is_grad_enabled,  # set_output_requires_grad
             *self._flat_basic_op_params,
             *extra_inputs,
         )
--- a/transformer_engine/pytorch/ops/fuser.py
+++ b/transformer_engine/pytorch/ops/fuser.py
@@ -65,6 +65,7 @@ class _OperationFuserAutogradFunction(torch.autograd.Function):
         input_: torch.Tensor,
         fuser: OperationFuser,
         basic_op_kwargs: list[dict[str, Any]],
+        set_output_requires_grad: bool,
         *params_and_extra_inputs: torch.Tensor,
     ) -> torch.Tensor | tuple[torch.Tensor, ...]:
         """Forward pass
@@ -79,6 +80,8 @@ class _OperationFuserAutogradFunction(torch.autograd.Function):
             Container for the pipeline of operations to run
         basic_op_kwargs: list of dict
             Keyword arguments to BasicOperation
+        set_output_requires_grad: bool
+            Whether to set ``requires_grad`` flags on returned tensors
         *params_and_extra_inputs: torch.Tensor
             Other tensor inputs to include in autograd graph. Consists
             of parameter tensors, followed by extra operation inputs.
@@ -138,7 +141,7 @@ class _OperationFuserAutogradFunction(torch.autograd.Function):
             )
             for idx, ys in zip(basic_op_idxs, fused_op_extra_outputs):
                 for y in ys:
-                    if func_ctx is not None:
+                    if set_output_requires_grad:
                         y.requires_grad_(idx >= fuser.first_op_requiring_backward)
                 extra_outputs[idx] = ys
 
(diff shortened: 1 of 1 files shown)


transformer_engine/pytorch/ops/fuser.py
```diff
@@ -138,7 +138,8 @@ def forward(
             )
             for idx, ys in zip(basic_op_idxs, fused_op_extra_outputs):
                 for y in ys:
-                    y.requires_grad_(idx >= fuser.first_op_requiring_backward)
+                    if func_ctx is not None:
```
timmoon10: This logic is not intuitive. `func_ctx` is `None` when `torch.is_grad_enabled() == False`: 
https://github.com/NVIDIA/TransformerEngine/blob/0c2e7b09c6d33109803ea089fbf80421a326e0a4/transformer_engine/pytorch/ops/fuser.py#L504-L509

It would be better to pass in `is_grad_enabled` as an arg so that we can be explicit and not rely on secret contracts.

timmoon10 (commented):
I'm not sure if this is really addressing the root cause of the issue. Two problems:
- We aren't actually protecting against setting `requires_grad` on non-leaf nodes. We're just skipping `requires_grad` logic when `torch.is_grad_enabled() == True`.
- Do we even want to skip setting `requires_grad` on non-leaf nodes? The backward expects grads from each of the outputs, so we need `requires_grad` for autograd to do the right thing.

I think the right solution is smarter logic when setting `requires_grad_`. Maybe something like:
```python
x_requires_grad = fuser.first_op_requiring_backward < fuser._num_basic_ops
if x_requires_grad != x.requires_grad:
    x = x.detach()
    if x_requires_grad:
        x.requires_grad_()

# Or maybe only detach if x is a non-leaf node?
# Need to check if the CPU overhead of checking
# is worth saving the CPU overhead of detaching.

...

return x
```
Another approach would be changing our ops to always return leaf nodes. For example, here is the forward pass of `MakeExtraOutput`:
https://github.com/NVIDIA/TransformerEngine/blob/0c2e7b09c6d33109803ea089fbf80421a326e0a4/transformer_engine/pytorch/ops/basic/make_extra_output.py#L76
This would be changed to:
```python
out = input_.detach()
return out, [(out,)] 
```

transformer_engine/pytorch/ops/fuser.py
CarlosGomes98: As I understand it, the real issue is that when the forward_func is .apply, we are free to set requires_grad_ on returned tensors. But when it is .forward, we cannot mutate this state on non-leaf tensors.

When `torch.is_grad_enabled()` is false, we bypass `.apply` and call `.forward` directly with no `func_ctx`. In that path there is no OperationFuserAutogradFunction node registered, so no fuser backward will run. So I think this

> The backward expects grads from each of the outputs, so we need requires_grad for autograd to do the right thing

is not true because we cannot run backward() through it.

I think it makes sense to pass this as an explicit argument as you say, instead of relying on the func_ctx being None. But I think the current logic is correct

timmoon10: I see, the `apply`/`forward` distinction makes sense to me. We should include a comment making that clear.

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -501,20 +507,24 @@ def __call__(
...
+        # When is_grad_enabled is False, we call forward directly.
+        # This does not register a PyTorch autograd node, so
+        # no fuser backward will run. We pass set_output_requires_grad=False
+        # to avoid setting requires_grad on outputs in
+        # this path since they may be non-leaf tensors from the inner ops.
+        args = (
             input,
             self,
             basic_op_kwargs,
+            is_grad_enabled,  # set_output_requires_grad
```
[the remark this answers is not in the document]
timmoon10: It's a reasonable contract for output tensors to have the expected `requires_grad` when grads are enabled. If grads are enabled and we are unable to set `requires_grad` in an output tensor, then that's a bug.

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -501,20 +507,24 @@ def __call__(
...
-            args = []
-        else:
-            forward_func = _OperationFuserAutogradFunction.forward
-            args = [None]
-        args += (
+        # When is_grad_enabled is False, we call forward directly.
+        # This does not register a PyTorch autograd node, so
+        # no fuser backward will run. We pass set_output_requires_grad=False
+        # to avoid setting requires_grad on outputs in
+        # this path since they may be non-leaf tensors from the inner ops.
```
timmoon10: I feel this comment is not quite right. The distinction between `forward`/`apply` makes sense, but it is not obvious what that has to do with `requires_grad`. The comment implies it has something to do with the backward pass, but that doesn't make sense. Really, it's because `apply` does some pre-processing that converts the inputs to leaf nodes.
```suggestion
        # Note: We call forward directly when is_grad_enabled=False,
        # which can expose non-leaf tensors to the inner ops. Avoid
        # problems in this case by passing set_output_requires_grad=False.
```

timmoon10 (approved):

PR #2920, opened 2026-04-23
[PyTorch] Add newton_schulz_tp optimizer step function

vcherepanov-nv:
# Description

Add a newton_schulz_tp utility function on top of the existing newton_schulz. To be used in Muon optimizer step.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add a newton_schulz_tp function
- refactor test_newton_schulz.py to do multiple check during a single torchrun

# Checklist:


transformer_engine/pytorch/optimizers/muon.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+class MuonOptimizer(Optimizer):
+    """Distributed Muon optimizer for 2D CUDA parameters.
+
+    This optimizer applies SGD-momentum followed by Newton-Schulz orthogonalization
+    on tensor-parallel parameter shards. The local parameter shard must represent a
+    partition of a logical 2D matrix across the provided NCCL process group.
+
+    Args:
```
skyw: Q: Does TE use numpy style docstring instead of Google style?

transformer_engine/pytorch/optimizers/muon.py
```diff
@@ -0,0 +1,235 @@
...
+        extra_scale_factor: Extra multiplicative scale applied after orthogonalization.
+        process_group: NCCL process group for distributed Newton-Schulz. Defaults to world.
+        partition_dim: Dimension along which each logical 2D parameter is partitioned.
+            Must be 0 or 1.
+        eps: Lower bound for the distributed normalization denominator.
+    """
+
+    def __init__(
+        self,
+        params: Iterable[torch.nn.Parameter | dict],
```
skyw: Nit: The type here doesn't match PyTorch internal. Should be fine for the purpose of this class.

transformer_engine/pytorch/optimizers/muon.py
```diff
@@ -0,0 +1,235 @@
...
+        nesterov: bool = True,
+        weight_decay: float = 0.01,
+        *,
+        use_decoupled_weight_decay: bool = True,
+        coefficient_type: NSCoeffT = "quintic",
+        num_ns_steps: int = 5,
+        scale_mode: MuonScaleT = "spectral",
+        extra_scale_factor: float = 1.0,
+        process_group: Optional[dist.ProcessGroup] = None,
+        partition_dim: int = 1,
```
skyw: Fix: partition_dim is per parameter.

transformer_engine/pytorch/optimizers/muon.py
```diff
@@ -0,0 +1,235 @@
...
+    ) -> None:
+        if lr < 0.0:
+            raise ValueError(f"Invalid learning rate: {lr}")
+        if momentum < 0.0 or momentum >= 1.0:
+            raise ValueError(f"Invalid momentum value: {momentum}")
+        if weight_decay < 0.0:
+            raise ValueError(f"Invalid weight_decay value: {weight_decay}")
+        if num_ns_steps < 1:
+            raise ValueError(f"num_ns_steps must be at least 1, got {num_ns_steps}")
+        if partition_dim not in (0, 1):
```
skyw: Q: Does this class intend to support non-distributed case? partition_dim would be -1 in TE in such case.

transformer_engine/pytorch/optimizers/muon.py
```diff
@@ -0,0 +1,235 @@
...
+            raise ValueError(f"partition_dim must be 0 or 1, got {partition_dim}")
+        get_coefficients(num_ns_steps, coefficient_type)
+
+        if process_group is None:
+            if not dist.is_initialized():
+                raise RuntimeError("MuonOptimizer requires torch.distributed to be initialized.")
```
skyw: Same question above regarding single GPU support.

transformer_engine/pytorch/optimizers/muon.py
```diff
@@ -0,0 +1,235 @@
...
+            process_group = dist.group.WORLD
```
skyw: Suggestion: This silent behavior is dangerous. If user forgot to pass the correct TP group, wrong group will be used.

transformer_engine/pytorch/optimizers/muon.py
```diff
@@ -0,0 +1,235 @@
...
+        *,
+        partition_dim: int,
+        coefficient_type: NSCoeffT,
+        num_ns_steps: int,
+        scale_mode: MuonScaleT,
+        extra_scale_factor: float,
+        eps: float,
+    ) -> torch.Tensor:
+        self._validate_param(grad, partition_dim)
+        world_size = dist.get_world_size(self.process_group)
```
skyw: Some suggestion as above. The silent behavior of None process group falling back to default is dangerous. (Understand it is from PyTorch for historical reasons)

transformer_engine/pytorch/optimizers/muon.py
```diff
@@ -0,0 +1,235 @@
...
+        global_shape = [grad.size(0), grad.size(1)]
+        global_shape[partition_dim] *= world_size
+
+        orth_grad = grad.clone()
+        transposed = partition_dim == 0
```
skyw: Attn: This is from common Row and Column wise tensor parallelism in most LLM. It would be sub optimal for anything other than that. Add comment if the assumption is made.

skyw (commented):
I'd advice NOT to expose it in public API. Keeping it in test only if that is the purpose.

Having an optimizer with most code copied invites fragmentation. 

Before this, all optimizer TE provides are more optimized fused version. I'd say a highly optimized Fused Muon with similar concept can be justified, but would need more consideration because it has more dependencies on other part of the training pipeline than elementwise optimizers.

vcherepanov-nv:
> Having an optimizer with most code copied invites fragmentation.

The idea was to give something to users, who use TE, but not Megatron-LM. By fragmentation you mean that we want to encourage everyone to use Megatron-LM? Or that the optimizer being relatively thin thing on top of newton_schulz call, and the users should have no trouble creating it themselves?

I don't think we gain anything by putting it into tests, since we already have tests for newton_schulz call. So we need to decide whether we want this PR, or should abandon it altogether. @cyanguwa

skyw:
> > Having an optimizer with most code copied invites fragmentation.
> 
> The idea was to give something to users, who use TE, but not Megatron-LM. By fragmentation you mean that we want to encourage everyone to use Megatron-LM? Or that the optimizer being relatively thin thing on top of newton_schulz call, and the users should have no trouble creating it themselves?
> 
> I don't think we gain anything by putting it into tests, since we already have tests for newton_schulz call. So we need to decide whether we want this PR, or should abandon it altogether. @cyanguwa

Fragmentation means there will be different flavor of muon in emerging optimizer and TE, also a lot of copied code. TE can have stalled feature when emerging optimizer updates. Megatron-LM will always have its own version because there are implementation specific things need to be hooked together. For example, how QKV is implemetned, or fused swighlu.
For TE, I think an example of how to build a version of emerging optimizer use TE NS backend would be good to have.  But providing optimizer (not fusion optimized version) confuses customers. 
Having said that, I would love for TE to have a more optimized version. similar idea as fusedAdam, etc.

transformer_engine/pytorch/optimizers/muon.py
cyanguwa, line 1: Should we move newton_schulz.py to this directory? Also, how do we expect Megatron to call us for this functionality? Thanks.

vcherepanov-nv: > Should we move newton_schulz.py to this directory?

No, don't think so.

> Should we move newton_schulz.py to this directory?

Megatron will call newton_shulz directly from their optimizers. This one is for other users.

cyanguwa:
1. run CI here "/te-ci torch L1"; add tests to qa/Lx_pytorch_unittest/test.sh?
2. please create `newton_schulz_tp` API to include `partition_dim`/`mode` params, and for Megatron integration 
3. please test non-distributed cases (per comment above)
4. please move newton_schulz.py to te/pytorch/optimizers; if we have more solvers to integrate to TE or more use cases of Newton-Schulz, we can definitely restructure the code, but we don't see that in the near future

cyanguwa:
@skyw Just following up on the discussion above - our purpose for this PR was two-fold. One was to provide an equivalent `newton_schulz_tp` API for Megatron; the other one was to provide a dialed-down version of Muon Optimizer class so direct TE users can access the Newton-Schulz solver. I understand this may cause divergence in TE and Megatron's Muon support, but we do want to expose this feature to direct users of TE as well. Hope that helps. Thanks.

transformer_engine/pytorch/optimizers/muon.py
timmoon10: I'd prefer moving them into something like `transformer_engine/pytorch/cusolver`. But I suppose that is orthogonal to this PR.

tests/pytorch/distributed/test_muon_optimizer.py
timmoon10, line 1: We should make sure to include this in the QA script: https://github.com/NVIDIA/TransformerEngine/blob/main/qa/L1_pytorch_distributed_unittest/test.sh

tests/pytorch/distributed/test_muon_optimizer.py
```diff
@@ -0,0 +1,84 @@
...
+    not MULTI_GPU_AVAILABLE,
+    reason="Muon optimizer distributed tests require at least 2 GPUs.",
+)
+
+TEST_ROOT = Path(__file__).parent.resolve()
+NUM_PROCS = torch.cuda.device_count()
+LAUNCH_CMD = ["torchrun", f"--nproc_per_node={NUM_PROCS}"]
+
+
+def _run_test(dtype: str, partition_dim: int, weight_decay_mode: str) -> None:
```
timmoon10: Each `torchrun` launch is somewhat expensive. Instead of launching a separate `torchrun` for each test case, it's better to launch a single `torchrun` instance and to perform multiple tests internally. See [`distributed/test_fusible_ops.py`](https://github.com/NVIDIA/TransformerEngine/blob/main/tests/pytorch/distributed/test_fusible_ops.py) for an example.

skyw:
> @skyw Just following up on the discussion above - our purpose for this PR was two-fold. One was to provide an equivalent `newton_schulz_tp` API for Megatron; the other one was to provide a dialed-down version of Muon Optimizer class so direct TE users can access the Newton-Schulz solver. I understand this may cause divergence in TE and Megatron's Muon support, but we do want to expose this feature to direct users of TE as well. Hope that helps. Thanks.

Megatron is wrapped over emgering-optimizers with megatron specific details, like TP and how QKV are organized. The most optimizer logic is in emerging-optimizers. Could TE do the same? I understand introducing a new dependency may have concern, let me know. The biggest concern is actually large portion of duplicated code. 

What I would favor is having `newton_schulz_tp`  in one release, test it out. and have a well optimized version of Muon optimizer class in the next. There are a lot of optimizations (fusion, batch, graph capturability etc.) that can and I believe should go into TE.

transformer_engine/pytorch/optimizers/newton_schulz.py
```diff
@@ -206,3 +206,115 @@ def newton_schulz(
...
+
+def newton_schulz_tp(
+    x: torch.Tensor,
+    ctx: CusolverMpCtx,
+    num_iterations: int = 5,
+    coefficients: Optional[Sequence[CoeffT]] = None,
+    partition_dim: Optional[int] = None,
+    tp_mode: Literal["duplicated", "distributed"] = "duplicated",
+) -> None:
+    """Compute tensor-parallel Newton-Schulz orthogonalization in-place.
```
cyanguwa, line 277: Is there some requirement for the operation to be in-place? I see we are doing a few `x.copy_()`s in the code and was just wondering if it's necessary.

transformer_engine/pytorch/optimizers/newton_schulz.py
```diff
@@ -206,3 +206,115 @@ def newton_schulz(
...
+    partitions are transposed before the call and transposed back afterward.
+
+    Parameters
+    ----------
+    x : torch.Tensor
+        Local tensor to orthogonalize in-place.
+    ctx : CusolverMpCtx
+        cuSolverMp context created for the tensor-parallel process group.
+    num_iterations : int, optional
+        Number of Newton-Schulz iterations. Default: 5.
```
cyanguwa, line 291: Should this be optional, or mandatory? Shouldn't users always be specifying the exact number of iterations they want to run NS with? Otherwise, they might get successful runs but incorrect results silently?

transformer_engine/pytorch/optimizers/newton_schulz.py
```diff
@@ -206,3 +206,115 @@ def newton_schulz(
...
+            f"Tensor dimension {distributed_dim} with size {x.size(distributed_dim)} "
+            f"must be divisible by tensor-parallel size {ctx.nranks}"
+        )
+
+    local_work = work.chunk(ctx.nranks, dim=1)[ctx.rank].contiguous()
+    newton_schulz(local_work, ctx, num_iterations, coefficients=coefficients)
+
+    output_shards = [torch.empty_like(local_work) for _ in range(ctx.nranks)]
+    dist.all_gather(output_shards, local_work, group=ctx.group)
+    output = torch.cat(output_shards, dim=1)
```
cyanguwa, line 265: I feel the functionality of `_orthogonalize_replicated()` is a bit odd. In the case of `partition_dim is None`, we probably don't need any chunking or all-gathering, given that tensor `x` is local, non-distributed. In the case of `tp_mode == "duplicated"`, we seem to be all-gathering, chunking, and then all-gathering again - is this necessary? Looking at [emerging-optimizers](https://github.com/NVIDIA-NeMo/Emerging-Optimizers/blob/6bf9d2d369c3c5f58ced7a3a4f6b7732fb24b55c/emerging_optimizers/orthogonalized_optimizers/muon_utils.py#L263), it seems to be all lowering down to a simple `newton_schulz()` call?

tests/pytorch/distributed/run_newton_schulz.py
```diff
@@ -72,56 +90,181 @@ def main():
...
+        newton_schulz_tp(
+            x_local,
+            ctx,
+            num_iterations,
+            coefficients=coefficients,
+            partition_dim=partition_dim,
+            tp_mode=tp_mode,
+        )
+    else:
+        newton_schulz(x_local, ctx, num_iterations, coefficients=coefficients)
```
cyanguwa, line 160: Non-TP is a special case of TP, right? So we probably can just call `newton_schulz_tp()` here regardless of whether `api == "tp"`? We just need to set up `partition_dim` and `tp_mode` properly for the non-TP case?

tests/pytorch/distributed/run_newton_schulz.py
```diff
@@ -32,34 +38,46 @@ def newton_schulz_reference(
...
-    dtype = torch.float32 if args.dtype == "float32" else torch.bfloat16
-    m = args.matrix_rows
-    n = args.matrix_cols if args.matrix_cols is not None else args.matrix_rows
-    coefficients = get_coefficients(args.num_iterations, args.coeff_type)
+def _test_tolerances(dtype: str, check: str, world_size: int) -> tuple[float, float]:
+    if dtype == "bfloat16":
+        return (5e-2, 5e-2)
+    if check == "orthogonality" and world_size == 1:
+        return (2e-2, 2e-2)
+    return (1e-2, 1e-2)
```
cyanguwa, line 55: Are the tolerances set a bit high? Usually we have 1e-2 for BF16 and even smaller for FP32. Could you do a bit of research and tighten them up if necessary? A couple of reference points are:

https://github.com/NVIDIA-NeMo/Emerging-Optimizers/blob/main/tests/test_muon_utils.py
https://github.com/vcherepanov-nv/TransformerEngine/blob/3c4dfffb788f01009bf6741d25d4b69078e95d13/tests/pytorch/test_permutation.py#L192

cyanguwa (changes requested):

transformer_engine/pytorch/optimizers/newton_schulz.py
timmoon10, line 1: Are we committing to only supporting Newton-Shulz, and to only use it for Muon? It would be more general to put this code in a `transformer_engine.pytorch.cusolvermp` submodule.

tests/pytorch/distributed/run_newton_schulz.py
timmoon10: I think it makes sense to test both APIs, even if they end up running the same thing internally.

transformer_engine/pytorch/optimizers/newton_schulz.py
cyanguwa: I think in principle that's a good idea, but if there's realistically no other use case of it in TE, it might be better to simplify the file structure?

cyanguwa: Could you check if there are other optimizers in Emerging-Optimizers that could be using cusolvermp?

transformer_engine/pytorch/optimizers/newton_schulz.py
vcherepanov-nv: Yes, it's intentional

transformer_engine/pytorch/optimizers/newton_schulz.py
vcherepanov-nv: It's necessary for this cuSolverMp. cuSolverMp interprets each rank’s input as a column shard; the wrapper must shard replicated inputs before the call and gather the distributed result afterward. Emerging-Optimizers can avoid this only because its local PyTorch implementation is different.

cyanguwa (approved):
Looks like the CI is clean as well. Thanks!

PR #2925, opened 2026-04-25
[PyTorch] Fix stale columnwise data usage

ksivaman:
# Description

This PR sets columnwise usage correctly for all quantizers instead of retaining the value in the quantizer, which may be incorrect after resuming training post validation steps as the columnwise usage is set to `False` for eval mode.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Set columnwise usage explicitly instead of retaining the value in the quantizer.

# Checklist:

### DIFF
--- a/tests/pytorch/test_sanity.py
+++ b/tests/pytorch/test_sanity.py
@@ -1233,3 +1233,92 @@ def test_inference_mode(
         with autocast(enabled=with_quantization, recipe=quantization_recipe):
             y = module(x, **kwargs)
     check_weights()
+
+
+@pytest.mark.parametrize(
+    "module_name",
+    ("Linear", "LayerNormLinear", "LayerNormMLP", "GroupedLinear"),
+)
+@pytest.mark.parametrize(
+    "quantization",
+    ("fp8_delayed_scaling", "fp8_current_scaling", "mxfp8"),
+)
+def test_quantizer_columnwise_usage_after_eval(
+    module_name: str,
+    quantization: str,
+) -> None:
+    """
+    Eval mode removes the columnwise usage from the quantizer, so modules
+    need to reset columnwise mode each time in backward instead of using it
+    from quantizer state, which may be stale.
+    """
+
+    if quantization in ("fp8_delayed_scaling", "fp8_current_scaling") and not fp8_available:
+        pytest.skip(reason_for_no_fp8)
+    if quantization == "mxfp8" and not mxfp8_available:
+        pytest.skip(reason_for_no_mxfp8)
+
+    sequence_length = 32
+    hidden_size = 32
+
+    if quantization == "fp8_delayed_scaling":
+        quantization_recipe = recipe.DelayedScaling()
+    elif quantization == "fp8_current_scaling":
+        quantization_recipe = recipe.Float8CurrentScaling()
+    else:
+        quantization_recipe = recipe.MXFP8BlockScaling()
+
+    with quantized_model_init(enabled=True, recipe=quantization_recipe):
+        if module_name == "Linear":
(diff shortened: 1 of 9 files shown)


tests/pytorch/test_sanity.py
```diff
@@ -1233,3 +1233,92 @@ def check_weights():
...
+    run_forward(is_eval=False)
+    for q in get_weight_quantizers():
+        assert (
+            q.columnwise_usage
+        ), "After an initial training forward, weight quantizer should have columnwise_usage=True"
+
+    # 2. Eval forward -- should set columnwise=False on primary FP8 weight
+    # quantizers, simulating the start of an evaluation loop.
+    run_forward(is_eval=True)
+    for q in get_weight_quantizers():
```
vthumbe1503, line 1313: The test doesnt make sense to me. I dont think we should be toggling the quantizer usages in case of quantized_model_init at all. 

This breaks the very principle that quantized tensor and its internal quantizer shouldnt be in conflict with each other.  And in here, the columnwise_data is present for the quantized_tensor even though the columnwise_usage is set to False.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -135,16 +135,11 @@ def forward(
...
                 )
-            # No need to set the quantizer states if weight is already quantized
-            # for debug mode we create quantizer every iteration, thus we need to set the quantizer states
-            if weight_quantizers[0] is not None and (
-                not isinstance(weights[0], QuantizedTensorStorage) or debug
-            ):
+            if weight_quantizers[0] is not None:
+                if isinstance(weights[0], QuantizedTensorStorage) and not debug:
+                    weight_quantizers = [weight._quantizer for weight in weights]
                 for weight_quantizer in weight_quantizers:
```
vthumbe1503, line 141: I dont think these changes are needed for this file and any of the other files in the PR. If weight is already quantized, it doesnt make sense to change its internal quantizer and have the quantized weight and its internal quantizer in a state of conflict with each other.

In general in case of quantized_model_init, if we are changing quantized_tensor's internal quantizer, quantized_tensor should also be updated to have that appropriate usages.

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -653,14 +653,14 @@ def fsdp_pre_all_gather(self, mesh, orig_size, contiguous_orig_stride, module, m
...
+        is_backward_pass = training_state == TrainingState.PRE_BACKWARD
         if reshard_after_forward:
-            training_state = param_group._training_state
-            is_backward_pass = training_state == TrainingState.PRE_BACKWARD
             rowwise_usage = not is_backward_pass
             columnwise_usage = is_backward_pass
         else:
             rowwise_usage = True
-            columnwise_usage = self._quantizer.columnwise_usage
+            columnwise_usage = is_backward_pass or torch.is_grad_enabled()
```
vthumbe1503, line 663: I think it still makes sense to use self._quantize.columwise_usage as the real truth of what data is 
"really avaliable" in the sharded quantized tensor and throw an error if that usage doesnt match
` is_backward_pass or torch.is_grad_enabled `(Similar to mxfp8 tensor)

What we are doing here is that we are silently creating columnwise data after allgather for allgathered tensor, even though original sharded data tensor didnt have that data.

In my opinion, I am against any change here since even doing such a validation and throwing error is going to incur CPU overheads when using 
`torch.is_grad_enabled `

vthumbe1503: Same comment in every other FSDP2 related changes

PR #2930, opened 2026-04-26
[Common/PyTorch] Add MXFP8 cast-and-transpose op

jeweldave:
## Summary

Add a fused MXFP8 cast-and-transpose op that takes a high-precision tensor
plus the source's existing compact column-wise E8M0 scales and emits row-wise
compact MXFP8 storage for the source's logical transpose. Surfaces in three
layers:

- **C API (additive, ABI-safe)**:
  - nvte_mxfp8_scaling_transpose_cast(input, scale_inv_colwise,
    output_rowwise, output_rowwise_scale_inv, rows, cols, stream) — minimal
    signature, E4M3 output, non-swizzled scales.
  - nvte_mxfp8_scaling_transpose_cast_v2(..., fp8_dtype,
    with_gemm_swizzled_scales, stream) — extended signature.
- **PyTorch extension**: transformer_engine_torch.mxfp8_scaling_transpose_cast
  (default kwargs map to the minimal C symbol's behavior).
- **Python**: MXFP8Quantizer.quantize_rowwise_transpose(tensor,
  columnwise_scale_inv, *, fake_dtype=None, with_gemm_swizzled_scales=None)
  returns a row-wise-only MXFP8Tensor whose logical shape is tensor.T.

No existing C symbol, Python signature, or default behavior is changed.

## Why

The standard MXFP8Quantizer path can already produce row-wise and
column-wise MXFP8 from BF16/FP16/FP32 input. There is currently no public TE
path that, given X and its compact column-wise scales S_col(X), produces
the row-wise compact MXFP8 storage for the logical transpose X.T without
either:

- emitting it from BF16 in a separate pass through the standard quantizer
  (re-reads BF16 source), or
- copying the existing column-wise MXFP8 payload and column-wise scales into
  transposed row-wise storage (extra payload+scale byte traffic).

This op closes that gap. It is the building block needed to route MXFP8
backward through TN GEMMs on hardware where cuBLASLt does not currently
support MXFP8 backward NN/NT GEMM layouts (NVIDIA Spark / sm_12.1). On
hardware where backward MXFP8 NN/NT is supported (B200 sm_10.0, H100
sm_9.0) it is unused by default; downstream code can still call it for any
path that wants direct transposed-rowwise MXFP8 emission without a payload
copy.

Detailed motivation, measurements, and out-of-scope notes are in
docs/motivation.md of the proposal directory.

## What's in the change

- transformer_engine/common/include/transformer_engine/recipe.h — declare
  nvte_mxfp8_scaling_transpose_cast and _v2.
- transformer_engine/common/recipe/mxfp8_scaling.cu — two new kernels
  (mxfp8_scaling_transpose_cast_kernel for the FP8 payload tile transpose,
  mxfp8_scaling_transpose_scales_kernel for compact-or-swizzled scale
  transpose), one new C++ entry point, two new C symbols.
- transformer_engine/pytorch/csrc/extensions.h /
  extensions/fp8_partial_cast.cpp /
  extensions/pybind.cpp — new
  mxfp8_scaling_transpose_cast PyTorch binding routed through _v2.
- transformer_engine/pytorch/tensor/mxfp8_tensor.py — new
  MXFP8Quantizer.quantize_rowwise_transpose helper.

## Numerics

For an input tensor X quantized with the standard MXFP8 column-wise path,
the new op's output is bit-for-bit equal to taking the existing column-wise
MXFP8 payload + scales and transposing those bytes. Confirmed on GB10 in the
existing cppmega probe at (M, N, K) = (64, 96, 128) and (256, 4096, 4096):
max_payload_abs_byte_delta == 0, payload_equal == True,
scale_equal == True.

The included tests/test_mxfp8_scaling_transpose_cast.py exercises this
equivalence both via the raw extension call and via the
quantize_rowwise_transpose helper.

## Tests

Drop-in pytest files added under tests/pytorch/ (this PR puts them in
tests/):

- test_mxfp8_scaling_transpose_cast.py:
  - byte equivalence vs. column-wise-then-copy reference, multiple shapes,
    E4M3 and E5M2;
  - Python helper equivalence;
  - decoded-value reconstruction is within MXFP8 quantization tolerance of
    the native re-quantized transpose;
  - error path: high-precision input is required (FP8 input rejected);
  - error path: source dims must be MXFP8-block-aligned.
- test_mxfp8_scaling_transpose_cast_swizzled.py:
  - with with_gemm_swizzled_scales=True, emitted scales match the bytes
    produced by the standard MXFP8Quantizer.quantize swizzled path on the
    actual transposed source.

All tests gate on CUDA being present and the new extension symbol being
built into the loaded transformer_engine_torch module.

## Compatibility

- C API: additive only. Original symbol name is reserved as the long-term
  stable signature; _v2 carries the extra knobs. No symbol's signature is
  changed.
- Python: additive only. New method on MXFP8Quantizer; no existing method
  signature or default behavior is changed.
- Build system: no new build flags or files; only changes existing files
  that already participate in the MXFP8 recipe build.

## Out of scope (intentionally not in this PR)

- Wiring TE Linear backward to use this op on GB10. That depends on
  cuBLASLt behavior we cannot upstream and on a downstream backward-rewrite
  shim.
- Changing default behavior of quantize / quantize_rowwise.
- Any change to cuBLASLt transposed operand consumption.
- Dequantize support for with_gemm_swizzled_scales=True MXFP8 tensors. By
  design TE rejects this in cast/mxfp8/dequantize_mxfp8.cuh (and symmetric
  paths in cast/nvfp4/dequantize_nvfp4.cuh,
  cast/mxfp8/group_dequantize_mxfp8.cuh) with
  Input must have scales in compact format: swizzled scales are a one-way
  GEMM-operand layout, and the dequantize kernels don't carry an inverse
  unswizzle_scale_idx path. Our swizzled-scale test
  (test_mxfp8_scaling_transpose_cast_swizzled.py) therefore compares the
  emitted row-wise payload and scale bytes against the standard
  MXFP8Quantizer.quantize(...,
  with_gemm_swizzled_scales=True) output byte-for-byte instead of via
  decoded values, since both paths target the same GEMM-ready layout for
  the same logical row-wise tensor.

### DIFF
--- a/tests/pytorch/mxfp8/test_mxfp8_scaling_transpose_cast.py
+++ b/tests/pytorch/mxfp8/test_mxfp8_scaling_transpose_cast.py
@@ -195,7 +195,6 @@ def test_transpose_cast_rejects_fp8_input():
 
 
 def test_transpose_cast_requires_block_aligned_dims():
-    source = _make_source(64, 128)
     quantizer = _make_quantizer(tex.DType.kFloat8E4M3)
     quantizer.set_usage(rowwise=True, columnwise=False)
     bad_source = torch.randn(48, 128, dtype=torch.bfloat16, device="cuda")
--- a/tests/pytorch/mxfp8/test_mxfp8_scaling_transpose_cast.py
+++ b/tests/pytorch/mxfp8/test_mxfp8_scaling_transpose_cast.py
@@ -101,12 +101,12 @@ def test_transpose_cast_matches_copy_adapter_bytes(rows, cols, fp8_dtype):
     )
     torch.cuda.synchronize()
 
-    assert torch.equal(rowwise_data.view(torch.uint8), expected_payload.view(torch.uint8)), (
-        "Row-wise MXFP8 payload bytes differ from copy-adapter reference"
-    )
-    assert torch.equal(rowwise_scale, expected_scale), (
-        "Row-wise MXFP8 scale bytes differ from copy-adapter reference"
-    )
+    assert torch.equal(
+        rowwise_data.view(torch.uint8), expected_payload.view(torch.uint8)
+    ), "Row-wise MXFP8 payload bytes differ from copy-adapter reference"
+    assert torch.equal(
+        rowwise_scale, expected_scale
+    ), "Row-wise MXFP8 scale bytes differ from copy-adapter reference"
 
 
 @pytest.mark.parametrize("rows,cols", [(64, 128), (256, 4096)])
--- a/tests/pytorch/mxfp8/test_mxfp8_scaling_transpose_cast.py
+++ b/tests/pytorch/mxfp8/test_mxfp8_scaling_transpose_cast.py
@@ -1,0 +1,208 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""Tests for the new MXFP8 cast-and-transpose op.
+
+These tests are written to drop into tests/pytorch/mxfp8/ of upstream
+TransformerEngine. They exercise:
+
+1. nvte_mxfp8_scaling_transpose_cast numerics vs. an in-test reference
+   reconstruction (MXFP8Quantizer.quantize columnwise + naive Python
+   transpose-and-pack).
+2. nvte_mxfp8_scaling_transpose_cast byte-for-byte equivalence to a copy
+   adapter that takes the existing column-wise MXFP8 payload, transposes it,
+   and rewrites it as row-wise storage.
+3. The MXFP8Quantizer.quantize_rowwise_transpose Python helper.
+4. The with_gemm_swizzled_scales=True variant (covered in
+   test_mxfp8_scaling_transpose_cast_swizzled.py).
+
+All tests gate on:
+
+- CUDA available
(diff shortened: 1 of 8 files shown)


ptrendx:
Hi @jeweldave, is this change intended to overcome the lack of non-TN gemm support on GB10?

jeweleugin2:
> Hi @jeweldave, is this change intended to overcome the lack of non-TN gemm support on GB10?

not quite yet but we will be adding support step by step. GB10 doesn't have tcgen05 and is essentially similar to Ada just with some FP8/NVFP4 extensions. If you're interested in exactly what can and cannot be done on GB10 we recently published our investigations here: https://megacpp.com/blog/gb10-blackwell-tensor-paths-what-we-actually-proved/

By the way we have multiple fixes related to Torch, Megatron and other components in our repository. We are currently working on porting our massive training setup with mix of homegrown kernels, Torch FSDP2 and code from DeepSeek, Megatron, and MaxText to GB10 and eventually to the full Megatron framework. You can check it out here https://github.com/DatasunriseOU/cppmega while this overcome the lack of non-TN gemm support on GB10 is part of our attempt to run end to end fp8 training on gb10 while nvidia did not provide right kernels for this and at least not end to end fused kernels

PR #2931, opened 2026-04-27
Implement row-scaled NVFP4 fprop recipe

zianglih:
# Description
@humansand

Implement ~per-token~ row-scaled NVFP4 recipe with fprop only.
Currently, the row-scaled scaling is handled by separate pytorch code.
Quantization kernels are bitwise exact with existing TE reference implementation.

The following tests passed on B200:
```
python3 -m pytest --tb=auto tests/pytorch/nvfp4/test_nvfp4_quantize_exact.py
python3 -m pytest --tb=auto tests/pytorch/nvfp4/test_nvfp4_gemm_exact.py
python3 -m pytest --tb=auto tests/pytorch/test_backward_override.py
python3 -m pytest --tb=auto tests/pytorch/test_sanity.py
python3 -m pytest --tb=auto tests/pytorch/test_recipe.py
python3 -m pytest --tb=auto tests/pytorch/test_torch_compile.py
python3 -m pytest --tb=auto tests/pytorch/test_cpu_offloading.py
PYTORCH_JIT=0 NVTE_TORCH_COMPILE=0 NVTE_ALLOW_NONDETERMINISTIC_ALGO=0 NVTE_FUSED_ATTN=0 python3 -m pytest --tb=auto tests/pytorch/test_cuda_graphs.py
NVTE_TEST_NVINSPECT_ENABLED=1 NVTE_TEST_NVINSPECT_CONFIG_FILE=tests/pytorch/debug/test_configs/dummy_feature.yaml NVTE_TEST_NVINSPECT_FEATURE_DIRS=transformer_engine/debug/features PYTORCH_JIT=0 NVTE_TORCH_COMPILE=0 NVTE_ALLOW_NONDETERMINISTIC_ALGO=0 python3 -m pytest --tb=auto tests/pytorch/test_sanity.py
```

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add a `row_scaled_activation` field in nvfp4 recipe, can be turned on by `NVTE_NVFP4_ROW_SCALED_ACTIVATION`
- ~New per-token nvfp4 quantize kernels in `transformer_engine/common/cast/nvfp4/quantize_pertoken_nvfp4.cuh`, bitwise exact with existing TE pytorch reference implementation and per-tesor nvfp4 emulated implmentation.~ New quantization kernels folded into existing nvfp4 quantization kernels.
- Expand dequant kernel `transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh` to correctly handle this row-scaled nvfp4
- In `TransformerEngine/transformer_engine/pytorch/cpp_extensions/gemm.py`, if row-scaled nvfp4 is enabled, it conducts separate per-token scaling using pytorch code, after cublas gemm
- Broad test coverage by expanding 7 python and 2 cpp test files
- Modify 1d quant reference implementation in `tests/cpp/operator/test_cast_nvfp4_transpose.cu` to align with pytorch reference numerics

# Checklist:


tests/cpp/operator/test_cast_nvfp4_transpose.cu
```diff
@@ -114,16 +114,14 @@ void quantize_nvfp4_1d(float (*OP)(const float),
...
-            const fp8e4m3 S_dec_b_fp8 = static_cast<fp8e4m3>(S_dec_b * S_enc);
+            // Compute and store the per-block FP8 decode scale
+            const float S_dec_b = block_amax * (S_enc * (1.0f / 6.0f));
+            const fp8e4m3 S_dec_b_fp8 = static_cast<fp8e4m3>(fminf(S_dec_b, Numeric_Traits<float>::maxNorm));
             const float S_dec_b_fp32 = static_cast<float>(S_dec_b_fp8);
 
             // Compute "correct" per-block encoding scaling factor
-            const float S_enc_b_fp8 = S_dec_b_fp32 == 0.f ? 0.f : S_enc / S_dec_b_fp32;
+            const float S_enc_b_fp8 = S_dec_b_fp32 == 0.f ? 0.f :
+                fminf(1.0f / (S_dec_b_fp32 * (1.0f / S_enc)), Numeric_Traits<float>::maxNorm);
```
zianglih, line 124: We have to change here to stay aligned with pytorch reference.

zianglih:
The following extended tests all passed:
```
python3 -m pytest --tb=auto tests/pytorch/nvfp4/test_nvfp4_quantize_exact.py
python3 -m pytest --tb=auto tests/pytorch/nvfp4/test_nvfp4_gemm_exact.py
python3 -m pytest --tb=auto tests/pytorch/test_backward_override.py
python3 -m pytest --tb=auto tests/pytorch/test_sanity.py
python3 -m pytest --tb=auto tests/pytorch/test_recipe.py
python3 -m pytest --tb=auto tests/pytorch/test_torch_compile.py
python3 -m pytest --tb=auto tests/pytorch/test_cpu_offloading.py
PYTORCH_JIT=0 NVTE_TORCH_COMPILE=0 NVTE_ALLOW_NONDETERMINISTIC_ALGO=0 NVTE_FUSED_ATTN=0 python3 -m pytest --tb=auto tests/pytorch/test_cuda_graphs.py
NVTE_TEST_NVINSPECT_ENABLED=1 NVTE_TEST_NVINSPECT_CONFIG_FILE=tests/pytorch/debug/test_configs/dummy_feature.yaml NVTE_TEST_NVINSPECT_FEATURE_DIRS=transformer_engine/debug/features PYTORCH_JIT=0 NVTE_TORCH_COMPILE=0 NVTE_ALLOW_NONDETERMINISTIC_ALGO=0 python3 -m pytest --tb=auto tests/pytorch/test_sanity.py

cd /root/TransformerEngine/tests/cpp
cmake --build build -j200
TEST_BIN="$(find build -type f -name test_operator -perm -u+x | head -n 1)"
"$TEST_BIN" --gtest_filter='*FusedCastTransposeNVFP4*:*DequantizeNVFP4*'
EOF
```

ptrendx:
Hi @zianglih, could you clarify why you needed the new quantization kernels? The existing NVFP4 quantization kernels should already work if you only use the rowwise mode there, no?

zianglih:
Hi @ptrendx , I have removed the standalone implementation and extended existing kernel to support this per-token nvfp4 recipe.

zianglih:
The functionality has been verified by nvfp4 rl experiment.

transformer_engine/common/cast/nvfp4/quantize_transpose_nvfp4.cuh
```diff
@@ -16,6 +16,9 @@
 #include <cuda_runtime.h>
 #include <transformer_engine/transformer_engine.h>
 
+#include <cub/cub.cuh>
```
timmoon10: Noting that we've added CUB as a dependency. This should be fine since CUB is included in the CCCL, which is included in the CUDA Toolkit.

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -370,6 +370,8 @@ enum NVTEQuantizationConfigAttribute {
    *  inconsistently between kernels.
    */
   kNVTEQuantizationConfigUseFastMath = 7,
+  /*! Whether to enable per-token (per-row) NVFP4 quantization */
+  kNVTEQuantizationConfigNVFP4PerTokenActivation = 8,
```
timmoon10: "Per-token activation" is a little over-tuned to our specific use-case. It's a general data format and  I don't see why we can't apply it in the future to non-activation tensors like weights or grads. Going further, I could see applications for quantized GEMMs outside NLP and even outside DL. "with 1D scaling" or "with row scaling" would be better names.

timmoon10: We should configure this in `NVTETensor` rather than in `NVTEQuantizationConfig`. The quantization config is used for quantization, and is not available to any downstream consumers (dequant or GEMM). However, consumers need to be aware of tensor-scaling vs row-scaling. The buffer sizes are different, and getting it wrong means incorrect values or segfaults.

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
timmoon10, line 1: Dequantization handles row-scaled NVFP4, but we need to make sure that other quantized tensor consumers also handle it. Erroring out is fine for now. Currently the only other consumers we need to handle are GEMM and attention, although we should keep this mind if we add more features in the future.

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
```diff
@@ -112,7 +113,8 @@ inline void dequantize(const Tensor &input, Tensor *output, cudaStream_t stream)
           dequantize_fp4_kernel<OType, WITH_GEMM_SWIZZLED_SCALES><<<blocks, threads, 0, stream>>>(
               input.data.dptr, reinterpret_cast<OType *>(output->data.dptr),
               reinterpret_cast<fp8e4m3 *>(input.scale_inv.dptr),
-              reinterpret_cast<float *>(input.amax.dptr), N, Mread, input.scale_inv.shape.back(),
+              reinterpret_cast<float *>(input.amax.dptr), input.amax.numel(), N, Mread,
```
timmoon10: I think it's too implicit to distinguish between row-scaling and tensor-scaling based on the amax tensor shape. We should add a `bool` flag to `Tensor`.

transformer_engine/pytorch/quantization.py
```diff
@@ -1375,6 +1375,7 @@ def _make_quantizer(idx: int) -> NVFP4Quantizer:
                     with_post_rht_amax=qparams.random_hadamard_transform,
                     with_2d_quantization=qparams.fp4_2d_quantization,
                     stochastic_rounding=qparams.stochastic_rounding,
+                    per_token_activation=self.recipe.per_token_activation and idx % 3 != 1,
```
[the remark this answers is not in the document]
timmoon10: We should be able to get rid of this hack once https://github.com/NVIDIA/TransformerEngine/pull/2620 is merged.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1747,6 +1748,9 @@ std::pair<TensorWrapper, py::object> NVFP4Quantizer::create_tensor(const std::ve
   NVTE_CHECK(flat_last_dim % NVFP4_BLOCK_SIZE == 0,
              "NVFP4 requires tensor dims that are divisible by ", NVFP4_BLOCK_SIZE,
              " (got shape=", shape, ")");
+  NVTE_CHECK(!this->per_token_activation || rowwise_usage,
+             "Per-token NVFP4 quantization requires rowwise usage.");
+  const bool columnwise_usage = this->columnwise_usage && !this->per_token_activation;
```
timmoon10: I'd prefer to error out rather than silently ignore the column-wise usage.
```suggestion
  if (this->per_token_activation) {
    NVTE_CHECK(rowwise_usage,
               "Per-token NVFP4 quantization requires rowwise usage.");
    NVTE_CHECK(!columnwise_usage,
               "Per-token NVFP4 quantization does not support columnwise usage.");
  }
  const bool columnwise_usage = this->columnwise_usage;
```
I see this pattern in several places in the C++ and Python quantizers and in `cast.cpp`.

transformer_engine/pytorch/tensor/storage/grouped_tensor_storage.py
timmoon10, line 1: Are we actually handling grouped tensors correctly? I don't see the corresponding changes in the C++ quantizer and we don't have any special handling in [`general_grouped_gemm_for_grouped_tensor`](https://github.com/NVIDIA/TransformerEngine/blob/528f16c5067a50c5a4ec2b8f4c466d3372536323/transformer_engine/pytorch/cpp_extensions/gemm.py#L321). If we don't intend to support this now (probably a good idea since this it's still highly experimental), it would be better to error out.

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -69,6 +70,44 @@ def validate_gemm_scale(scale: Optional[float], required: bool) -> float:
     return 0.0
 
 
+def _is_nvfp4_per_token_tensor(tensor: torch.Tensor) -> bool:
+    """Whether tensor carries per-token NVFP4 global amax metadata."""
+    if not isinstance(tensor, NVFP4TensorStorage):
+        return False
+    amax = tensor._amax_rowwise if tensor._amax_rowwise is not None else tensor._amax_columnwise
+    return amax is not None and amax.numel() > 1
```
timmoon10: Enabling row-scaling based on the amax tensor shape is too implicit, and it also limits us if we ever want to look at other hierarchical scaling schemes, e.g. column-scaling. Better to put a bool flag in `NVFP4TensorStorage`.

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -174,7 +213,59 @@ def general_gemm(
...
+        gemm_args[6] = TE_DType[torch.float32]  # out_dtype
+        out, bias_grad, gelu_input, extra_output = tex.generic_gemm(*gemm_args, **kwargs)
+        out_2d = out.reshape(-1, out.shape[-1])
+
+        assert per_token_scales.dtype == torch.float32 and out.dtype == torch.float32
+        assert per_token_scales.numel() == out_2d.shape[0]
+
+        if bias is not None:
+            bias_cast = bias.to(dtype=torch.float32)
+            out_2d.sub_(bias_cast)
```
timmoon10: Why not remove the bias from the GEMM so we don't need to subtract it out here?

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -1502,7 +1563,7 @@ std::vector<py::object> split_quantize(const at::Tensor &tensor,
         });
         quantization_method = QuantizationMethod::UNFUSED;
       }
-      if (!contiguous_data_and_scale) {
+      if (!all_nvfp4_per_token_activation && !contiguous_data_and_scale) {
```
timmoon10: Does the row-scaled multi-quantize kernel support non-contiguous data? I don't see any kernel changes that would allow the row-scaled kernel to work if the tensor-scaled kernel doesn't.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -1263,6 +1281,33 @@ void split_quantize_nvfp4_impl_helper(const TensorWrapper &input,
     nvte_tensor_output_list.push_back(output_list[i].data());
   }
 
+  if (quantizer.per_token_activation) {
```
timmoon10: `split_quantize_nvfp4_impl_helper` is intended to launch a fused multi-quantize kernel, but here we have an unfused implementation. It would be less roundabout for `split_quantize` to just choose `QuantizationMethod::UNFUSED` when row-scaling is enabled.

tests/pytorch/test_recipe.py
```diff
@@ -507,8 +513,28 @@ def test_quantizer_update(self, module_class):
                     y = module(x)
 
 
+@pytest.mark.skipif(not fp4_available, reason=reason_for_no_fp4)
+def test_nvfp4_per_token_quantizer_roles():
```
timmoon10: This test will need to be updated once https://github.com/NVIDIA/TransformerEngine/pull/2620 merges.

transformer_engine/pytorch/tensor/nvfp4_tensor.py
timmoon10, line 1: We should make sure we don't break tensor parallelism. This should just require returning `False` from `is_quantizable` when row-scaling is enabled:
https://github.com/NVIDIA/TransformerEngine/blob/0803102751cfb099f71348efddc50cc0664cda89/transformer_engine/pytorch/tensor/nvfp4_tensor.py#L213

This is a badly-named function that indicates whether to quantize before or after a tensor-parallel all-gather:
https://github.com/NVIDIA/TransformerEngine/blob/0803102751cfb099f71348efddc50cc0664cda89/transformer_engine/pytorch/quantized_tensor.py#L366-L372

timmoon10 (changes requested):
Overall this is a nice feature, but we should make some changes to the core design. My biggest suggestions:
- Row-scaling should be part of the tensor data and not just hidden in the quantizer. You need to be aware of it for both quantization and dequantization, so it can't be hidden in the quantizer.
- We should enable row-scaling based on a bool flag rather than the amax tensor shape. We should also make sure it is clearly documented.
- We should consider a better name like "1D scaling" or "row scaling" since I don't see any reason this is specific to tokens or activations.

transformer_engine/pytorch/cpp_extensions/gemm.py
zianglih: Done by https://github.com/NVIDIA/TransformerEngine/pull/2931/commits/94b05e37152ac176636faddb7901ed5c51902c94

transformer_engine/pytorch/csrc/extensions/cast.cpp
zianglih: Done by https://github.com/NVIDIA/TransformerEngine/pull/2931/commits/6c10ed2871b4f51140a517e58c6dbf2764ee44fa

transformer_engine/pytorch/csrc/extensions/cast.cpp
zianglih: Done by https://github.com/NVIDIA/TransformerEngine/pull/2931/commits/6c10ed2871b4f51140a517e58c6dbf2764ee44fa

transformer_engine/pytorch/csrc/quantizer.cpp
zianglih: Done by https://github.com/NVIDIA/TransformerEngine/pull/2931/commits/aa519d1f45e80872b080b93c7833446e6520dfcc

transformer_engine/common/include/transformer_engine/transformer_engine.h
zianglih: Rename done by https://github.com/NVIDIA/TransformerEngine/pull/2931/commits/1a4d3b0da24039197519537286a3edc6d316c954

transformer_engine/pytorch/tensor/nvfp4_tensor.py
zianglih: Done by https://github.com/NVIDIA/TransformerEngine/pull/2931/commits/600b4cd24d06607393c3a1ffbe0f88f600331ec3

transformer_engine/common/cast/nvfp4/dequantize_nvfp4.cuh
zianglih: Done by https://github.com/NVIDIA/TransformerEngine/pull/2931/commits/1a4d3b0da24039197519537286a3edc6d316c954

transformer_engine/pytorch/tensor/storage/grouped_tensor_storage.py
zianglih: Explicit error out in https://github.com/NVIDIA/TransformerEngine/pull/2931/commits/cc9a2103eaa60ffd6cc77f71db0f1f3468d353d5

zianglih:
Hi @timmoon10 , thank you for your comments! I have refactored accordingly and rename this to `row_scaled_nvfp4`, no per-token. In recipe we still have `row_scaled_activation`, which indicates 1d2d gemm where only the activation is modified to 1d. The control flow is now based on bool flags instead of implicitly inferred by shapes.

transformer_engine/common/cast/nvfp4/quantize_transpose_nvfp4.cuh
```diff
@@ -27,6 +30,132 @@ namespace transformer_engine {
...
+template <typename IType, int BLOCK_SIZE>
+__global__ void
+#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
+__launch_bounds__(BLOCK_SIZE)
+#endif
+    compute_rowwise_amax_kernel(const int num_rows, const int num_cols,
+                                const IType *__restrict__ input,
+                                float *__restrict__ output_rowwise_amax,
+                                const float *__restrict__ noop) {
+#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
```
timmoon10: Better to fail loudly:
```suggestion
#if !defined(__CUDA_ARCH__) || (__CUDA_ARCH__ < 1000)
  NVTE_DEVICE_ERROR("SM 10.0+ is required.");
#else
```

transformer_engine/common/cast/nvfp4/quantize_transpose_nvfp4.cuh
```diff
@@ -27,6 +30,132 @@ namespace transformer_engine {
...
+  for (int i = threadIdx.x; i < num_vec2; i += BLOCK_SIZE) {
+    const IType2 val = input_row[i];
+    abs_max_2x_update(thread_amax_2x, val);
+  }
+  const float thread_max = abs_max_2x_to_float(thread_amax_2x);
+
+  using BlockReduce = cub::BlockReduce<float, BLOCK_SIZE>;
+  __shared__ typename BlockReduce::TempStorage temp_storage;
+  const float row_amax =
+      BlockReduce(temp_storage).Reduce(thread_max, [](float a, float b) { return fmaxf(a, b); });
```
timmoon10: We already have a utility for block max reductions:
```suggestion
  const float row_amax = reduce_max<BLOCK_SIZE / THREADS_PER_WARP>(thread_max, threadIdx.x / THREADS_PER_WARP);
```
This would also remove the need for CUB. See: 
https://github.com/NVIDIA/TransformerEngine/blob/4b6923dd1141272c668a96e9486a5482ec6ada40/transformer_engine/common/utils.cuh#L863

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -229,6 +320,52 @@ def general_grouped_gemm(
...
+            out_views = []
+            for i in range(num_gemms):
+                size = m_splits[i]
+                out_views.append(out_init[start_idx : start_idx + size])
+                start_idx += size
+        else:
+            out_views = out
+        for i in range(num_gemms):
+            if out_views[i].numel() == 0:
+                continue
```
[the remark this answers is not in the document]
timmoon10: The type hint for `out` is `List[torch.Tensor]`, so this case shouldn't happen. @greptile-apps says all callers respect this, but it would be good for them to double-check.

timmoon10 (commented):
This is almost ready. The only important suggestion is to add checks in the core lib GEMM functions so we error out instead of misinterpreting the amax:
https://github.com/NVIDIA/TransformerEngine/blob/4b6923dd1141272c668a96e9486a5482ec6ada40/transformer_engine/common/gemm/cublaslt_gemm.cu#L315

My other suggestions are nits.

transformer_engine/common/cast/nvfp4/quantize_transpose_nvfp4.cuh
zianglih: Done by https://github.com/NVIDIA/TransformerEngine/pull/2931/commits/01a32ef2537c2e455f47fd3a593227e902d4fc31

transformer_engine/common/cast/nvfp4/quantize_transpose_nvfp4.cuh
zianglih: Done by https://github.com/NVIDIA/TransformerEngine/pull/2931/commits/01a32ef2537c2e455f47fd3a593227e902d4fc31

zianglih:
Hi @timmoon10 , I have cleaned up accordingly and resolved a previous ci lint error. Thank you!

timmoon10:
We see a test failure when running on A100: https://github.com/NVIDIA/TransformerEngine/actions/runs/25475147141/job/74746937329
The distributed test failures also show up in the `main` branch, so they are not blocking.

zianglih:
Previous L0_pytorch_unittest--A100_1GPU failed due to a minor test file issue. L1_pytorch_distributed_unittest--H100_4GPU failed due to numerics I did not touch. L1_pytorch_distributed_unittest--B200_8GPU reached time limit.

tests/pytorch/utils.py
```diff
@@ -151,15 +151,40 @@ def make_recipe(name: Optional[str], **recipe_kwargs: Any) -> Optional[Recipe]:
...
 
 
+def recipe_id(fp8_recipe: Optional[Recipe]) -> str:
+    """Readable pytest id for FP8/FP4 recipes."""
+    if fp8_recipe is None:
+        return "None"
+    nvfp4 = getattr(fp8_recipe, "nvfp4", None)
+    if nvfp4 is not None and nvfp4() and getattr(fp8_recipe, "row_scaled_activation", False):
+        return "NVFP4RowScaledBlockScaling"
+    return type(fp8_recipe).__name__
```
timmoon10: This messiness is working around this test failure: https://github.com/NVIDIA/TransformerEngine/actions/runs/25475147141/job/74746937329

Looking into it, the root cause is actually a pytest bug: https://github.com/pytest-dev/pytest/issues/13031
The fix landed in pytest 8.4.0: https://github.com/pytest-dev/pytest/pull/13073

The real problem is that pytest is incorrectly passing in a weird internal class, and it has nothing to do with NVFP4. We can make this more logical with:
```suggestion
def recipe_id(recipe: Optional[Recipe]) -> str:
    """Readable pytest id for a quantization recipe."""
    if not isinstance(recipe, Recipe):
        return "None"
    if recipe.nvfp4() and recipe.row_scaled_activation:
        return "NVFP4RowScaledBlockScaling"
    return type(recipe).__name__
```

timmoon10 (approved):
LGTM. The PyTorch test failures are unrelated. The JAX failure seems related to MoE kernels, so that's also unrelated.

PR #2935, opened 2026-04-27
fix: skip generating fp8 & mxfp8 checkpoints if unsupported

kainzhong:
# Description

Using `python3 tests/pytorch/test_checkpoint.py --save-checkpoint all` to generate checkpoints would fail on some hardwares that don't support fp8 and mxfp8. This PR refactors the test script so it checks if the checkpoint type is supported first.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Make `test_checkpoint.py`check if the checkpoint to be generated is supported on the current hardware.

# Checklist:


timmoon10 (changes requested):
I don't agree with this approach. If you specify `--save-checkpoint all`, you should get all the checkpoints. We don't need to put much effort into making `TestLoadCheckpoint._save_checkpoint` fully adaptable to all systems, since it's not run in the test suite and is actually a special utility for updating checkpoints in our CI infrastructure.

PR #2937, opened 2026-04-28
Fix CUDA graph parameter grad lifetime

buptzyb:
## Summary

Fix CUDA graph replay so parameter gradients returned from `Graphed.backward` do not expose CUDA graph static buffers to downstream autograd users.

The fix clones returned parameter gradients before handing them back to autograd, while preserving the existing aliasing behavior for delayed-wgrad parameters marked with `skip_backward_post_hook`.

## Root Cause

When CUDA graph replay returns parameter grad slots directly from static graph buffers, downstream autograd users can retain references to buffers that are overwritten by later graph replays. This can corrupt retained grads or break gradient accumulation semantics.

This is related to PyTorch issue https://github.com/pytorch/pytorch/issues/181723.

## Changes

- Detect parameter grad slots in the graphed autograd input surface.
- Clone returned non-delayed-wgrad parameter grads before returning from `Graphed.backward`.
- Allow reused graph input/output buffer mode to weak-ref current parameter grad static buffers after capture because returned grads are now cloned.
- Add CUDA graph tests for owned returned parameter grads, accumulation, delayed-wgrad alias preservation, and reused buffer interleaved pipeline replay.

## Summary by CodeRabbit

* **New Features**
  * Added control over whether parameter gradients are cloned when returned from CUDA graph replay.
  * Added support for reusing CUDA graph input and output buffers.
  * Preserved safe gradient ownership and aliasing behavior during replay and accumulation.
  * Added coverage for interleaved pipeline parallelism with reused buffers and configurable gradient cloning.

* **Bug Fixes**
  * Prevented returned parameter gradients from incorrectly referencing static CUDA graph buffers.


transformer_engine/pytorch/graph.py
```diff
@@ -407,6 +407,15 @@ def _make_graphed_callables(
...
     graph_callables = [None for _ in range(len(flatten_sample_args))]
 
+    def _is_returned_param_grad_slot(idx, static_grad_inputs, module_params):
+        """Return whether a static grad slot is consumed through Graphed.backward."""
+        module_param_start = len(static_grad_inputs) - len(module_params)
+        if idx < module_param_start:
+            return False
+        return not getattr(
+            module_params[idx - module_param_start], "skip_backward_post_hook", False
+        )
```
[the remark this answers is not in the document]
buptzyb: Addressed in 4077b857. The parameter-grad clone policy is now snapshotted at capture time and passed into `Graphed.backward`, so replay no longer re-reads `skip_backward_post_hook`. Added `test_make_graphed_callables_snapshots_parameter_grad_clone_policy` to cover changing the attribute after capture.

tests/pytorch/test_cuda_graphs.py
```diff
@@ -752,3 +884,23 @@ def test_make_graphed_callables_with_interleaved_pipeline_parallelism(
...
+    outputs = _test_cuda_graphs_with_interleaved_pipeline_parallelism(
+        with_graph=False,
+        **kwargs,
+    )
+    graph_outputs = _test_cuda_graphs_with_interleaved_pipeline_parallelism(
+        with_graph=True,
+        reuse_graph_input_output_buffers=True,
+        **kwargs,
+    )
+    assert_all_equal(outputs, graph_outputs)
```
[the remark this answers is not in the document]
buptzyb: Addressed in 4077b857. The interleaved pipeline helper now returns final weights in addition to outputs, and the reused-buffer test compares graph/eager final weights to cover gradient correctness. Full `tests/pytorch/test_cuda_graphs.py` passed on H100: 415 passed, 423 skipped.

ptrendx:
I have a problem with this PR. On one hand I agree that there is a bug here and we should ensure that the gradient buffers are not overwritten before being applied. On the other hand for the cases without gradient accumulation (or when the accumulation is done in a different way, like in Megatron) it is a performance loss. Could we make this behavior optional - have it on by default, but also provide an opt-out option with a clear warning that states where this would be applicable?

buptzyb:
@ptrendx I added a new argument `_clone_param_grads_on_return` (default = True). Documented that disabling this may improve performance but no longer have standard PyTorch returned-gradient lifetime semantics. Does this work for you?

buptzyb:
Hi @ptrendx can we merge this PR?

transformer_engine/pytorch/graph.py
```diff
@@ -108,6 +108,7 @@ def _make_graphed_callables(
     pool: Optional[Tuple[int, ...]] = None,
     retain_graph_in_backward: bool = False,
     _reuse_graph_input_output_buffers: bool = False,
+    _clone_param_grads_on_return: bool = True,
```
pggPL: I think we should expose it to `make_graphed_callables()`, not just `_make_graphed_callables()` and also document it in docstring. Also, I think we can make it fully public without `_`.

pggPL: Also, I think we should make sure that MCore will set it to `False` if they do their wgrad accumulation. Maybe it's a good idea to create PR there also?

buptzyb: Thanks! Addressed both suggestions in f286621a:

- `clone_param_grads_on_return` is now a public argument on `make_graphed_callables()` and is documented in the public docstring.
- I opened NVIDIA/Megatron-LM#5893 to set `clone_param_grads_on_return=False` from MCore when the public option is available, while preserving compatibility with older TE versions.

Thanks for the suggestions!

pggPL (dismissed):
LGTM

pggPL:
@buptzyb please do not merge yet, we need to make sure that we do not slow down any public nemo workload - I will reach out to nemo folks tomorrow to make sure that both PRs will land in versions that we will not do harm to this workloads.

cspades:
@buptzyb Quick question, is the bug specifically when a CG buffer initializes `param.grad` on the first microbatch? If `param.grad` already exists, then `AccumulateGrad` will just add the new gradient to the existing buffer, which may or may not be backed (and overwritten by) CG.

So if anything this is an AccumulateGrad logical loophole, where to save memory allocation they use the first gradient as the accumulation buffer.

More or less same comment on MCore side: https://github.com/NVIDIA/Megatron-LM/pull/5893#issuecomment-5093576161

buptzyb:
@cspades I think so. If you pre-assign an empty tensor for `param.grad`, then the CG output grad will just be added to that tensor, then everything is fine. The issue is only for running CG when `param.grad` is None.

transformer_engine/pytorch/graph.py
```diff
@@ -402,6 +403,19 @@ def _make_graphed_callables(
...
+    def _returned_param_grad_clone_slots(static_grad_inputs, module_params):
+        """Snapshot static grad slots that need clones before returning from Graphed.backward."""
+        if not clone_param_grads_on_return:
+            return (False,) * len(static_grad_inputs)
+        module_param_start = len(static_grad_inputs) - len(module_params)
+        return tuple(
+            idx >= module_param_start
+            and not getattr(
+                module_params[idx - module_param_start], "skip_backward_post_hook", False
+            )
```
cspades, line 473: @buptzyb I think developers reading this criterion may be a bit misled. Could we add this comment:

> Megatron DDP and FSDP have custom post-backward hooks that accumulate the gradient into a different buffer. param.grad.clone() isn't necessary in this case, because the replay and accumulation buffers are distinct and the latter will not be overwritten by the former.

In general, if there are post-backward hooks that do NOT have custom accumulation logic, then it will still trigger this bug. So currently, this `skip_backward_post_hook` naming may be slightly underspecified, right? Let's document it.

cspades (approved):
Only one nitpick comment, but let's get this merged ASAP.

transformer_engine/pytorch/graph.py
buptzyb: oh wait, this attr is not related to the DDP/FSDP self-managed accumulation. This attr is specifically for delayed wgrad, where the returned `param.grad` is a dummy tensor and will not be accumulated into main grad, so we don't need to clone it as an optimization. The flag to let normal DDP/FSDP skip cloning is the user-controlled `clone_param_grads_on_return`, unrelated to this one.

So actually here we are treating setting up `skip_backward_post_hook=True` <=> the returned `param.grad` being nonsense. At least this is how it works for now.

Added a comment to clarify this, could you take a look? Thanks!

And I'm ok to merge if you think this is good now～

pggPL (approved):

transformer_engine/pytorch/graph.py
cspades: Ah, good to know, thanks!

PR #2942, opened 2026-04-29
[JAX] Fix bf16 precision loss in TestGroupedDense reference dbias

tdophung:
# Description

`TestGroupedDense::test_grouped_dense_grad_fp8` fails on `MXFP8_1D_SCALING` with `input_shape=(8, 64, 128, 256)`: `prim_dbias` vs `ref_dbias` differs by ~6% on the largest group, exceeding the bf16 rtol (other rows pass; fwd, dgrad, wgrad pass).

Reference path adds bias in bf16 inside `_ref_grouped_dense`, so JAX autodiff
lowers the bias-grad sum-over-m *in bf16* and saturates on the largest group. The primitive's `grouped_dbias` casts to fp32 before `segment_sum`, so it's accurate. The two computed dbias values therefore disagree even though the kernel is correct. Test-construction issue, not a kernel regression.

To fix this, in the reference impl, add NO bias to the grouped dense call, then add bias externally in fp32, so the autodiff bias-grad accumulates in fp32 --> This matches what `grouped_dbias` does for the primitive. 

Reason why this bug just surface now was because of commit 70af7305 ("[JAX] MXFP8 Grouped Quant+GEMM (#2763)") introduced `group_size_multiplier=128` for MXFP8, which quadrupled M for this shape and pushed the largest group's bf16 bias-grad sum past the bf16 rtol.

Fixes # (issue) 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- `_ref_sum_grouped_dense`: pass `bias=None` to `_ref_grouped_dense`, add bias externally in fp32, sum and divisor in fp32
- `_primitive_sum_grouped_dense`: cast `out` and divisor to fp32 to mirror the reference

# Checklist:

### DIFF
--- a/tests/jax/test_custom_call_compute.py
+++ b/tests/jax/test_custom_call_compute.py
@@ -1923,17 +1923,16 @@ class TestGroupedDense:
         # If bias is added inside _ref_grouped_dense in bf16, JAX lowers the bias
         # backward as a bf16 sum-over-m and loses precision on the largest group,
         # producing a >bf16-rtol mismatch against the primitive's grouped_dbias
-        # (which casts the cotangent to fp32 before segment_sum).
+        # (which casts the cotangent to fp32 before segment_sum). Bias is required
+        # for this helper since it is only used by the grad tests below, which all
+        # set with_bias=True.
+        assert bias is not None, "_ref_sum_grouped_dense requires a non-None bias"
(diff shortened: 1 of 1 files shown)


phu0ngng (commented):
LGTM. Thanks

phu0ngng (approved):

PR #2951, opened 2026-05-01
MXFP8 + FSDP2 checkpoint resume crashes in reset_sharded_param            - add mxfp8 recpipe to fully shard 

savitha-eng:
## Bug                                                                                                                                   
                                                                                                                                           
  The `fully_shard.py` example crashes on checkpoint resume when using `MXFP8BlockScaling`. This affects any MXFP8 + FSDP2 training run that needs to resume from a checkpoint. Based on the existing
  `test_dcp_output_parity` xfails, NVFP4 is affected by the same bug.
                                                                                                                                           
  We hit this while running a Llama 7B MXFP8 convergence run with quantized model init + FusedAdam master weights. Checkpoint save succeeds, but `model.load_state_dict()` on resume crashes with:

  RuntimeError: Attempted to access the data pointer on an invalid python storage.

  at `torch/distributed/fsdp/_fully_shard/_fsdp_param.py:890` in `reset_sharded_param()`.

  ## Root Cause

  `model.load_state_dict()` calls `copy_()` on MXFP8Tensor params, which re-quantizes and allocates new internal storage. FSDP2's
  `reset_sharded_param` post-load hook then calls `data_ptr()` on the now-invalidated old storage.

  PyTorch has a `# TODO: need to support tensor subclass` comment at the crash site.

  ## Mislabeled xfails in TE tests

  In `tests/pytorch/distributed/fsdp2_tests/run_fsdp2_fused_adam.py::test_dcp_output_parity`:

  - The **MXFP8 xfail** references PR #2789 (FusedAdam kernel bug), but this is incorrect — the actual failure is the `data_ptr()` crash on
   invalid storage, not a FusedAdam issue. Commenting out the xfail and running the test confirms the same `data_ptr()` crash.
  - The **NVFP4 xfail** correctly describes the issue: "DCP load_state_dict triggers reset_sharded_param() which calls data_ptr() on
  wrapper subclass with invalid storage."

  Both MXFP8 and NVFP4 are blocked by the same underlying bug.

  ## Reproduce

  ```bash
  torchrun --nproc_per_node=2 examples/pytorch/quantized_model_init/fully_shard.py

  Requires Blackwell GPUs (B200). The only change to the example is adding recipe=MXFP8BlockScaling() to quantized_model_init and
  te.autocast calls.

  Workaround we used to train the model with checkpoint resumption: 

  We patched FSDPParam.reset_sharded_param to wrap the data_ptr() comparison in try/except RuntimeError, setting same_local_tensor = False
  on failure. This just tells FSDP to re-record _sharded_param_data, which is always correct.

  if type(self._sharded_param_data) is torch.Tensor:
      try:
          same_local_tensor = (
              self._sharded_param_data.untyped_storage().data_ptr() > 0
              and self._sharded_param_data.untyped_storage().data_ptr()
              == local_tensor.untyped_storage().data_ptr()
          )
      except RuntimeError:
          same_local_tensor = False

  Environment

  - TE: 2.15.0 (te-main)
  - PyTorch: 2.7.0
  - GPUs: B200 (Blackwell)


vthumbe1503:
Closing this PR since the bug has been fixed. @savitha-eng 
We also the pytest xfails removed for DCP checkpoint load+save tests.

vthumbe1503:
It might be good to generalize the recipe to be anything and not hardcode it to MXFP8. If you want to do that in this PR, we can revive this as well. Thanks for catching the issue!

PR #2953, opened 2026-05-01
[Common, PyTorch] Improve mHC to match DeepSeek's implementation

kainzhong:
# Description

Some enhancement for mHC to better align with DeepSeek's tilelang implementation: https://github.com/deepseek-ai/TileKernels/tree/main/tile_kernels/mhc

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes
- Add `mhc_generate_mix_and_aggregate` API that does projection, scale, sinkhorn and aggregate together
- Allow `mhc_fused_projection` to accept arguments with mixed dtype: x.dtype=bf16, phi.dtype=fp32, which matches DeepSeek's [implementation](https://github.com/deepseek-ai/TileKernels/blob/36d9e45d38e204ebb87e6f6e833821eee0482fe5/tile_kernels/modeling/mhc/ops/norm_fn.py#L58)
- `mhc_fused_projection` now outputs fp32 regardless of the input dtype, matching DeepSeek's [implementation](https://github.com/deepseek-ai/TileKernels/blob/36d9e45d38e204ebb87e6f6e833821eee0482fe5/tile_kernels/modeling/mhc/ops/norm_fn.py#L70)
- Add `fuse_grad_x_acc` optimization (default to False), which will reuse the same grad_x buffer to accumulate the initial mHC input x's gradient for `mhc_fused_expand_combine`, `mhc_fused_aggregate` and `mhc_fused_projection` 
- Support `norm_weight` for `mhc_fused_projection`, which would be equivalent to apply RMSNorm in the unfused manner with `elementwise_affine=True`, which would be the learnable per-element affine parameters for RMSNorm
- Support deterministic mode with non-spilt-K
- Refactor some kernel code to avoid duplication. I just realized if you make a triton kernel constexpr, it can be used as a macro in if branches since triton will not compile if it knows in compile time that some branch will not be taken
- Fix the bug that grid will exceed CUDA's limitation when M is too large and the autotune candidate is BLOCK_SIZE_M=1. Such invalid configs will be pruned now.
- Improve projection op by using TMA if on Hopper+

- ~~Add `main_grad` optimization for Megatron-LM integration, which will accumulate the gradient of phi, alpha, beta (they are all supposed to be `torch.nn.Parameter`) to `main_grad` if such attribute exists.~~ Removed because it's too much complexity and this parameter is not too large so this optimization only leads to negligible win
- ~~[TODO]: add checkpoint recomputing fused kernel to match DeepSeek's [implementation](https://github.com/deepseek-ai/TileKernels/blob/main/tile_kernels/modeling/mhc/ops/multilayer_recompute.py)~~ Probably need to figure out how to implement this in Megatron first since it doesn't seem too straightforward
- ~~[TODO]: add a fused projection + aggregate only kernel (no expand & combine path) for the last mHC layer, which seems to be also used for MTP. See function `learned_output_contract` in https://github.com/NVIDIA/Megatron-LM/pull/4518~~ That kernel only takes 0.06ms for fwd+bwd. Plus my vibe coded kernel doesn't save too much time here.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/triton/mhc.py
+++ b/transformer_engine/pytorch/triton/mhc.py
@@ -9,38 +9,111 @@ import torch
 import triton
 
 from transformer_engine.common.triton.mhc import (
+    _mhc_projection_bwd_fused_dphi,
+    _mhc_projection_bwd_fused_dx,
     _mhc_scale_fwd_fused,
     _mhc_scale_bwd_fused,
-    _mhc_expand_combine_with_bias_fwd,
-    _mhc_expand_combine_with_bias_bwd,
     _mhc_expand_combine_fwd,
     _mhc_expand_combine_bwd,
     _mhc_aggregate_fwd,
     _mhc_aggregate_bwd,
     _mhc_projection_fwd_fused,
-    _mhc_projection_bwd_fused,
     _mhc_sinkhorn_fwd_fused,
-    _mhc_sinkhorn_fwd_fused_recompute,
     _mhc_sinkhorn_bwd_fused,
-    _mhc_sinkhorn_bwd_fused_recompute,
 )
 from transformer_engine.pytorch.cpp_extensions.gemm import general_gemm
 
 
-def check_deterministic(operator: str):
-    """
-    Checks if the non-deterministic algorithm is allowed for the given operator. If not, raises an assertion error with instructions on how to allow it.
-    Since atomic add is used in this mHC implementation, it breaks the determinism guarantee due to non-associativity of floating point addition.
-    """
-    allow_nondeterministic = os.environ.get("NVTE_ALLOW_NONDETERMINISTIC_ALGO", "1") == "1"
-    assert allow_nondeterministic, (
-        f"[{operator}]: This operation uses atomic add which violates determinism. Set"
-        " NVTE_ALLOW_NONDETERMINISTIC_ALGO=1 to allow this non-deterministic behavior."
-    )
+def is_deterministic_enforced():
+    # We assume non-determinism is allowed if this flag is not set
+    return os.environ.get("NVTE_ALLOW_NONDETERMINISTIC_ALGO", "1") == "0"
 
+
+def mhc_generate_mix_and_aggregate(
+    x: torch.Tensor,
+    phi: torch.Tensor,
+    alpha: torch.Tensor,
+    beta: torch.Tensor,
+    norm_weight: torch.Tensor = None,
+    use_tf32: bool = True,
+    fuse_grad_x_acc: bool = False
+):
+    """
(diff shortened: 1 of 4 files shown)


transformer_engine/pytorch/triton/mhc.py
```diff
@@ -803,7 +1018,7 @@ class mHCExpandCombineOp(torch.autograd.Function):
     """
 
     @staticmethod
-    def forward(ctx, f, bias, H_post, x, H_res, n, use_tf32=True):
+    def forward(ctx, f, bias, H_post, x, H_res, n, use_tf32=True, fuse_grad_x_acc=True):
```
kainzhong: Make it default to False, and also add this to other two functions to make users aware

transformer_engine/pytorch/triton/mhc.py
```diff
@@ -996,4 +1213,9 @@ def backward(ctx, grad_output):
         if bias is not None:
             grad_bias = grad_bias.to(bias.dtype)
 
-        return grad_f, grad_bias, grad_H_post, grad_x, grad_H_res, None, None
+        if ctx.fuse_grad_x_acc:
+            # When fused x gradient accumulation is enabled, use fp32 for the accumulation buffer
+            x.untyped_storage().grad_x_acc = grad_x.to(torch.float32)
```
kainzhong, line 1269: Check if this is overwriting anything that is already there

transformer_engine/pytorch/triton/mhc.py
```diff
@@ -763,7 +970,11 @@ def backward(ctx, grad_output):
         assert n == 4, "Only n=4 is supported in this implementation"
         M = s * b
 
-        grad_x = torch.empty_like(x)
+        fuse_grad_x_acc = hasattr(x.untyped_storage(), "grad_x_acc")
```
kainzhong: Add fuse_grad_x_acc param in other functions as well

transformer_engine/pytorch/triton/mhc.py
```diff
@@ -401,6 +614,8 @@ def forward(ctx, H, alpha, beta, ms, n):
 
         ctx.dtype = H.dtype
         H = H.to(torch.float32)
+        ctx.alpha_main_grad = getattr(alpha, "main_grad", None)
```
kainzhong: With FSDP this will break
Check linear.py for reference
Talk with Varun

kainzhong: After some discussion we decide to remove this `main_grad` in my PR since it's a lot of complexity but it only saves a little overhead since the parameter here is small: (4*hidden, 32).
Will probably wait until Varun figures out how to make `main_grad` nicer in his PR

PR #2955, opened 2026-05-01
[PyTorch] Guard/document single parameter feature for grouped linear

ksivaman:
# Description

Guard/document single parameter feature for grouped linear.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Guard/document single parameter feature for grouped linear

# Checklist:


transformer_engine/pytorch/utils.py
```diff
@@ -81,6 +82,36 @@ def get_device_compute_capability() -> Tuple[int, int]:
     return _get_device_compute_capability(torch.cuda.current_device())
 
 
+def resolve_grouped_linear_single_param_flags(
+    single_grouped_weight: bool,
+    single_grouped_bias: bool,
```
timmoon10, line 87: While we are breaking backward compatibility, we might consider consolidating these options together. Do we really want to take on the burden of supporting the case with a single grouped weight and discrete bias, or discrete weights and single grouped bias?

transformer_engine/pytorch/utils.py
```diff
@@ -81,6 +82,36 @@ def get_device_compute_capability() -> Tuple[int, int]:
...
+) -> Tuple[bool, bool]:
+    """Gate ``single_grouped_weight`` / ``single_grouped_bias`` on ``NVTE_GROUPED_LINEAR_SINGLE_PARAM``."""
+    if not (single_grouped_weight or single_grouped_bias):
+        return single_grouped_weight, single_grouped_bias
+
+    env_enabled = int(os.environ.get("NVTE_GROUPED_LINEAR_SINGLE_PARAM", "0")) > 0
```
timmoon10, line 93: If we only respect the kwargs if an envvar is set, it doesn't really make sense to keep the envvars rather than just checking the envvar. I guess we're half-heartedly maintaining/preparing the stable API for this feature.

timmoon10 (dismissed):
This logic breaks backward compatibility since we can no longer rely on the module kwargs to configure single grouped params. I guess we've already been treating single grouped params as an experimental feature. We should keep this instability in mind whenever we use this feature externally, e.g. in Mcore.

timmoon10 (dismissed):

timmoon10 (approved):

PR #2957, opened 2026-05-04
[PyTorch] CPU overhead optimizations for te autocast

vthumbe1503:
# Description
te-autocast has quite a bit of CPU overheads on Grace Systems.
Here are the results on GB200 after the optimizations

- Without Optimizations  
image<br></br>

- Optimization1 --> Cache recipe string representation for getting unique autocast key. Also directly accessing the cached representation of recipe is 40 ns faster than going through a str(recipe)
image
<br></br>

- Optimization2 --> Use __enter__, __exit__ methods instead of using contextlib.contextmanager
image

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/recipe/__init__.py
```diff
@@ -60,6 +60,14 @@ class MMParams:
 
     use_split_accumulator: bool = True
 
+    def __repr__(self) -> str:
+        cached = self.__dict__.get("_cached_repr")
+        if cached is not None:
+            return cached
+        result = f"MMParams(use_split_accumulator={self.use_split_accumulator})"
+        object.__setattr__(self, "_cached_repr", result)
+        return result
```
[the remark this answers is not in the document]
timmoon10: I see that this is why we're doing the funny accesses with `__dict__`. I agree that bypassing `frozen=True` is iffy, so I wonder if we could set `_cached_repr` in `__post_init__`? If the class is frozen, its repr must also be frozen and I don't see a benefit in lazy evaluation.

transformer_engine/common/recipe/__init__.py
```diff
@@ -77,20 +85,41 @@ class QParams:
...
     """
     Base recipe class.
     """
 
+    # Cached string representation. Lazily populated by ``__repr__`` in
+    # subclasses and invalidated by ``__setattr__`` whenever any attribute
+    # changes. This makes repeated ``str(recipe)`` calls (e.g. on the hot
+    # path in ``FP8GlobalStateManager.get_unique_autocast_key``) essentially
+    # free after the first call.
+    _cached_repr: Optional[str] = None
```
timmoon10, line 113: Three problems:
- `_cached_repr` is being set as a class attr, not an instance attr.
- Accessing `_cached_repr` via `__dict__` is non-standard and bug-prone.
- Splitting the cache logic between the base class and child classes results in code duplication and more risk of bugs, especially if it involves non-standard `__dict__` accesses.

What if we concentrated the caching logic in the base class:
```python
class Recipe:

    def __init__(self) -> None:
        self._cached_repr: Optional[str] = None

    @abc.abstractmethod
    def _make_repr(self) -> str:
        ...

    def __repr__(self) -> str:
        if self._cached_repr is None:
            self._cached_repr = self._make_repr()
        return self._cached_repr

    ...

class DelayedScaling(Recipe):

    def _make_repr(self) -> str:
        return f"..."
```

transformer_engine/pytorch/quantization.py
```diff
@@ -591,7 +590,13 @@ def get_unique_autocast_key(
         Object identity is sufficient since autocast contexts never outlive a single
         training session.
         """
-        return str((str(recipe), id(group) if group is not None else None))
+        # directly getting the cached repr is about 40 ns faster than str(recipe)
+        # on grace systems.
```
timmoon10: This is good to mention in the PR description, but not that useful in the code itself. Profiling becomes outdated once we move on to the next architecture.

transformer_engine/pytorch/quantization.py
```diff
@@ -591,7 +590,13 @@ def get_unique_autocast_key(
...
+        recipe_repr = recipe.__dict__.get("_cached_repr") if recipe is not None else None
+        if recipe_repr is None:
+            recipe_repr = str(recipe)
+        group_id = id(group) if group is not None else 0
+        return f"{recipe_repr}|{group_id}"
```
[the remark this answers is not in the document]
timmoon10:
```suggestion
        group_id = id(group) if group is not None else None
        return f"recipe=({str(recipe)}),group={group_id}"
```

transformer_engine/pytorch/quantization.py
```diff
@@ -885,24 +880,53 @@ def autocast(
                           are reduced at the end of each training step.
     """
 
-    if enabled:
-        check_recipe_support(recipe)
+    # Class-based context manager (instead of ``@contextmanager`` from contextlib)
+    # to avoid the ~0.5us / invocation overhead of contextlib's generator-driven
+    # ``GeneratorContextManager``. ``__slots__`` further avoids per-instance
+    # dict allocation.
```
timmoon10: Why are we mentioning the context manager here? It makes sense for this PR, but once the code is merged it will be completely random. This comment should explain what we are doing with `__slots__`, and we should explain the custom context manager logic in `__enter__` and `__exit__`.

transformer_engine/pytorch/quantization.py
```diff
@@ -885,24 +880,53 @@ def autocast(
...
-    try:
-        yield
-    finally:
-        FP8GlobalStateManager.set_autocast_state(fp8_state)
-        FP8GlobalStateManager.autocast_exit(enabled, _graph=_graph)
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        FP8GlobalStateManager.set_autocast_state(self._fp8_state)
+        FP8GlobalStateManager.autocast_exit(self._enabled, _graph=self._graph)
+        # Do not suppress exceptions.
+        return None
```
timmoon10: Nit: The function already returns `None` and the comment is trivially true (all Python outside of a `try` statement is not suppressing exceptions).
```suggestion
```

transformer_engine/pytorch/quantization.py
vthumbe1503: Simpliefied the comment

transformer_engine/pytorch/quantization.py
```diff
@@ -591,7 +590,11 @@ def get_unique_autocast_key(
         Object identity is sufficient since autocast contexts never outlive a single
         training session.
         """
-        return str((str(recipe), id(group) if group is not None else None))
+        recipe_repr = recipe.__dict__.get("_cached_repr") if recipe is not None else None
+        if recipe_repr is None:
+            recipe_repr = str(recipe)
+        group_id = id(group) if group is not None else None
+        return f"recipe={recipe_repr},group={group_id}"
```
timmoon10, line 597: We can simplify now that the recipe handles caching internally:
```suggestion
        group_id = id(group) if group is not None else None
        return f"recipe=({recipe}),group={group_id}"
```

timmoon10 (approved):
Overall LGTM, with one simplification.

transformer_engine/pytorch/quantization.py
vthumbe1503: It turns out Getting str(recipe) is about 40ns slower than getting the cached_repr directly.

vthumbe1503:
Merging the changes since the one test failure is a pre-existing one on main.

PR #2964, opened 2026-05-06
[All] Refactor fused attention APIs with cuDNN-frontend support checks and opaque config/params handles

cyanguwa:
# Description

TE currently hand-maintains the fused-attention backend-selection logic in `nvte_get_fused_attn_backend`, duplicating cuDNN's support rules. This list drifts out of sync as cuDNN evolves, and the support check can disagree with what actually runs.

This PR replaces that logic with cuDNN-frontend's production-grade support checks. The new `nvte_get_fused_attn_backend_v2` builds the *same* graph cuDNN executes at runtime, so the probe and execution can no longer diverge. It caches the graph on success and returns a diagnostic message on failure, giving users actionable guidance (e.g. adjust the config, GPU architecture, or cuDNN version).

This PR also reworks `nvte_fused_attn_fwd` / `nvte_fused_attn_bwd` into `nvte_fused_attn_fwd_v2` / `nvte_fused_attn_bwd_v2`, which take opaque, attribute-based config/params handles instead of long flat argument lists — improving TE's API and ABI stability.

Legacy APIs are retained as deprecated shims that route through the v2 APIs, so existing callers keep working.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

### API rework (opaque config/params + v2 entry points)
- **Opaque config/params handles** (`common/fused_attn/config_and_params.{h,cpp}`, `common/include/transformer_engine/fused_attn.h`): new `NVTEFusedAttnConfig` / `NVTEFusedAttnFwdParams` / `NVTEFusedAttnBwdParams` with `create`/`destroy`/`get`/`set` attribute accessors, for better API/ABI stability. The cache key, probe, and execution now all originate from one place via `make_config` / `derive` / `make_cache_key`.
- **v2 APIs** (`common/fused_attn/fused_attn*.{cpp,cu}`): `nvte_get_fused_attn_backend_v2`, `nvte_fused_attn_fwd_v2`, and `nvte_fused_attn_bwd_v2`, taking the opaque handles above in place of flat argument lists.
- **Graph construction split from execution** (`common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu`, `fused_attn_fp8.cu`, `graph_cache.h`): graph building moved out of `fused_attn_*_{fwd,bwd}_impl` into `create_graph_{f16,fp8}_{fwd,bwd}(const FusedAttnConfig &)`; the `*_impl` bodies now only fetch the graph, finish its plans, and bind the variant pack. The probe (`support_verdict_*`) and execution reach that builder through the same `get_graph<Backend, Pass, create_graph_*>` and cache, so they can't diverge; `build_plans()` is deferred so a probe-only miss doesn't pay for it.
- **Deprecated shims**: legacy `nvte_get_fused_attn_backend` / `nvte_fused_attn_fwd` / `nvte_fused_attn_bwd` are retained, routed through the v2 APIs.
- **Bindings updated to v2**: PyTorch (`csrc/extensions/attention.cpp`) and JAX (`jax/csrc/extensions/attention.cpp`).

### Correctness & backend selection
- **Process-wide graph cache**: cache is now process-wide (was thread-local) and guarded by a mutex, so a compiled graph is reused across threads instead of rebuilt per thread (still thread-safe).
- **Bias-shape handling fix**: applied consistently across common, PyTorch, and JAX.
- **Per-step CP config checks**: `cp_per_step_configs` probes each context-parallel step instead of only the global, non-CP config.
- **`log2(0)` guard**: avoids UB when casting `-inf` to `size_t` in `get_max_batch_size` / `get_max_tokens`.

### Diagnostics
- **`NVTE_DEBUG` / `NVTE_DEBUG_LEVEL` for JAX** (parity with PyTorch): level 1 reports the selected backend; level 2 adds a diagnostic message explaining why fused attention was rejected.
- **Fused attention graph cache debug `NVTE_FUSED_ATTN_CACHE_DEBUG=<level>[:<ranks>]`**: opt-in instrumentation that reports cuDNN graph build-vs-execution counts and per-stage cudnn-frontend build timings, so cache hit/miss/build/exec behaviors and graph build time can be inspected. Off by default; `[:<ranks>]` specifies select ranks for diagnostics; available for both PyTorch and Jax.

### Cleanup / removals
- Removed `NVTE_FUSED_ATTN_BACKEND` — the two remaining backends (F16, FP8) are mutually exclusive now that max512 is gone.
- Removed dead `Q_ID`/.../`MASK_VAL_ID` macros (used only by the max512 backend).
- Removed dead `cudnn_frontend::xxx` utility functions (used only by `fp8_impl_v0` and max512).
- Unified include-guard names across `fused_attn/` headers.

### Tests
- Enabled previously skipped tests: padding + post_scale_bias in both PyTorch and Jax, D256 bprop in PyTorch, and SWA + dropout/post_scale_bias in Jax.
- Curated the L0 sweeps to keep CI time in check: deduplicated PyTorch tests, and tiered the newly enabled JAX tests across L0/L1/L2.
- Added `test_fused_attn_graph_cache` to test graph cache's behavior with `NVTE_FUSED_ATTN_CACHE_DEBUG=2` on, and `test_fused_attn_backend_message` to test the surfacing of TE-specific or cuDNN-related error messages to users.

# Checklist:


transformer_engine/jax/flax/transformer.py
```diff
@@ -779,6 +779,8 @@ def __call__(
         enable_fused_attn = int(os.getenv("NVTE_FUSED_ATTN", "1"))
 
         sequence_dim = 0 if self.transpose_batch_sequence else 1
+        batch_dim = 1 - sequence_dim
```
jberchtold-nvidia, line 798: Should this be `batch_dim = sequence_dim - 1` instead of the current `1 - sequence_dim`?

For the (B, S, H, D) case they're equivalent, in both cases batch_dim = 0. But if we wanted to support an additional axis (_, B, S, H, D), the current formula would put the batch dim as `batch_dim = 1 - seq_dim = 1 - 2 = -1`, which is really the D axis.

transformer_engine/jax/csrc/extensions/pybind.cpp
```diff
@@ -246,6 +248,14 @@ PYBIND11_MODULE(transformer_engine_jax, m) {
       .value("NVFP4_2D_SCALING", JAXX_Scaling_Mode::NVFP4_2D_SCALING)
       .export_values();
 
+  pybind11::enum_<NVTEScalingMode>(m, "NVTEScalingMode", pybind11::module_local())
```
jberchtold-nvidia: Please use the existing already exported JAXX_Scaling_Mode instead of exporting NVTEScalingMode

pggPL:
There is a limitation for torch.compile dynamic shapes here. If `b` and `s` are passed to black-box function affecting the program flow, I will need to force recompilation on every `b` or `s` change. 
If we have args like `s_greater_or_eq_512` and we check `s >= 512` in python then no recompilation would be needed for every change of `s`. It may be more problematic for `b`.

I don't think this is a blocker for this PR, but we need to keep it in mind.

transformer_engine/jax/flax/transformer.py
cyanguwa: I don't think we support (_, B, S, H, D) in TE-Jax, but we do support SBHD, so `1-sequence_dim` should be correct (similar to what `self.transpose_batch_sequence` does in the line above). But let me know if there are some use cases I don't know about!

cyanguwa:
@greptileai

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -205,16 +182,41 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
...
+      fused_attn_graph_debug::note_cache_lookup("fwd", cache_hit, cfg);       // [GRAPH-DEBUG]
+      if ((is_ragged_q || is_ragged_kv) && cudnn_runtime_version >= 90600 &&  // [GRAPH-DEBUG]
+          sm_arch_ != 120) {                                                  // [GRAPH-DEBUG]
+        fused_attn_graph_debug::note_thd_lookup(                              // [GRAPH-DEBUG]
+            "fwd", cache_hit, !cache_hit || fused_attn_graph_debug::cache_disabled(),
+            /*legacy=*/!use_cu_seqlens_directly);  // [GRAPH-DEBUG]
+      }  // [GRAPH-DEBUG]
+      if (cache_hit && !fused_attn_graph_debug::cache_disabled()) {  // [GRAPH-DEBUG]
+        return cached_graph;
       }
```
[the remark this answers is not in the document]
cyanguwa: Good catch — addressed. FusedAttnConfig now carries a device_id, set from cuda::current_device() in make_cache_key(), and it's included in operator<, so the process-wide cache no longer aliases graphs across devices in a multi-GPU single-process run. It's excluded from attribute (de)serialization on purpose, since it's a cache-locality key rather than part of the graph spec.

transformer_engine/common/fused_attn/config_and_params.cpp
```diff
@@ -0,0 +1,1157 @@
...
+  const Tensor *input_O = convertNVTETensorCheck(params.O);
+  const Tensor *input_dO = convertNVTETensorCheck(params.dO);
+  const Tensor *output_dQ = convertNVTETensorCheck(params.dQ);
+  const Tensor *output_dBias = convertNVTETensorCheck(params.dBias);
+
+  const NVTE_QKV_Format q_format = nvte_get_q_format(params.qkv_layout);
+  const NVTE_QKV_Format kv_format = nvte_get_kv_format(params.qkv_layout);
+  auto *q_dims = input_Q->data.shape.data();
+  auto *k_dims = input_K->data.shape.data();
+  auto *v_dims = input_V->data.shape.data();
```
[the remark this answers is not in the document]
cyanguwa: This is intentional. MXFP8 fused-attention uses rowwise V in backward so this is already correct. I've also added a table in FusedAttention to detail the different recipes including MXFP8.

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -545,98 +445,113 @@ void nvte_fused_attn_fwd(const NVTETensor Q, const NVTETensor K, const NVTETenso
...
+    Tensor *output_S = convertNVTETensorCheck(p.Aux_CTX_Tensors->tensors[i++]);
+    Tensor *input_rng_state = convertNVTETensorCheck(p.Aux_CTX_Tensors->tensors[i++]);
+    Tensor *input_Bias, *input_SoftmaxOffset;
+    if ((p.bias_type != NVTE_NO_BIAS) && (p.bias_type != NVTE_ALIBI)) {
+      input_Bias = convertNVTETensorCheck(p.Aux_CTX_Tensors->tensors[i++]);
+    }
+    if (p.softmax_type != NVTE_VANILLA_SOFTMAX) {
+      input_SoftmaxOffset = convertNVTETensorCheck(p.Aux_CTX_Tensors->tensors[i++]);
+    }
+    fused_attn_arbitrary_seqlen_bwd(
```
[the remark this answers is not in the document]
cyanguwa: Fixed — both input_Bias and input_SoftmaxOffset are now initialized to nullptr and only reassigned when the corresponding aux tensors are present, so no uninitialized pointer reaches the kernel for NVTE_NO_BIAS / NVTE_VANILLA_SOFTMAX configs.

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -225,308 +226,207 @@ NVTE_QKV_Format nvte_get_kv_format(NVTE_QKV_Layout qkv_layout) {
...
   } else {
-    backend = NVTE_Fused_Attn_Backend::NVTE_No_Backend;
+    const char *reject_reason =
+        (fused_attn_reject_reason != nullptr && fused_attn_reject_reason[0] != '\0')
+            ? fused_attn_reject_reason
+            : "no cuDNN fused-attention backend supports the requested parameters";
+    NVTE_ERROR("Fused attention is not supported for this configuration: ", reject_reason);
   }
-  return backend;
 }
```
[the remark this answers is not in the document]
cyanguwa: This is intentional. Backend selection goes through nvte_get_fused_attn_backend_v2 with is_forward=false, which probes both forward and backward whenever is_training (see the is_training && !is_forward guard around the is_supported_*_bwd calls). The per-execution nvte_fused_attn_fwd_v2 sets is_forward=true only to avoid rebuilding the backward graph during the forward call — so a backend that can't support the training backward pass is already rejected at selection time, not deferred to the backward call.

tests/jax/test_distributed_fused_attn.py
```diff
@@ -81,25 +81,6 @@ def impl_test_self_attn(
...
-            num_head,
-            num_head,
-            seqlen,
-            seqlen,
-            hidden,
-            hidden,
-            None,  # no window
-        ):
-            pytest.skip("No FusedAttn backend found")
-
```
cyanguwa, line 103: Note: The explicit is_fused_attn_kernel_available() pre-check is no longer needed here:
FusedAttnRunner._check_configs() now runs the cuDNN-frontend support probe during
setup and skips with a diagnostic message when no fused-attn backend is available.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -4918,6 +4918,106 @@ def backward(ctx, dout, *_args):
...
+                mask,
+                s_q,
+                s_kv,
+                num_heads,
+                num_gqa_groups,
+                br,
+                t_q,
+                num_tokens_kv * cp_size * s_kv // max_seqlen_kv if max_seqlen_kv else 0,
+            )
+            for s_kv in dict.fromkeys([s_kv_chunk, max_seqlen_kv])
```
sudhakarsingh27, line 5486: this varies for every rank, would it make a difference?

KshitijLakhani:
- The addition of `NVTE_DEBUG` / `NVTE_DEBUG_LEVEL` for JAX (parity with PyTorch) bridges a long time gap. Thank you !

- I am very curious if you've run any experiments with `NVTE_FUSED_ATTN_CACHE_DEBUG` - what sort of numbers/metrics have you seen ? 
    - For e.g. with the new API, when we query backend support, we try to build a graph which in theory seems more expensive as compared to the C++ decision tree query we had in the older API, however, we'd expect the first fused attn execution call made to have a cache hit and hence in theory be faster than the older API.
I'd expect fused attn backend supported cases (where a graph has to be built anyways) to get faster, but the unsupported cases to get slower. Would be interesting to know by how much though.
    - Another interesting case would be if we have THD segments with a very small standard deviation packed together thereby resulting in them being in the same "bucket" hence triggering a cache hit ad no new graph creation. Comparing this to a larger deviation in the THD packed segment sizes which would trigger repetitive graph creation.
    
   - It seems like in the instrumentation code there's no way for the user to know if the defensive call for getting the backend in the fwd pass fused attn  in the execution phase of the model has a cache miss resulting in creating a new graph and new cache entry thereby losing any benefits of the graph created in the query phase. It could be useful information in the future, so that if the user is doing something incorrectly (inadvertently), maybe changing/passing any tensor params between querying and fused attn execution at least we can inform them via debug logging as I'd image that penalty won't be inexpensive
   
   - Do we log the device ID when `NVTE_FUSED_ATTN_CACHE_DEBUG` is enabled ? Maybe I missed i but if not, maybe we should so that the user knows which device the debug info corresponds to ?

docs/examples/attention/attention.ipynb
```diff
@@ -346,17 +346,11 @@
...
-    "```\n",
-    "NVTE_FUSED_ATTN_BACKEND = 1/2 # user preference of cuDNN sub-backend\n",
-    "```\n",
-    "\n",
     "```\n",
     "<div class=\"alert alert-info\">\n",
     "<b>Note</b>\n",
     "    \n",
-    "Environment variables <code>NVTE_FLASH_ATTN</code>, <code>NVTE_UNFUSED_ATTN</code>, <code>NVTE_FUSED_ATTN_BACKEND</code>, and <code>NVTE_FUSED_ATTN_USE_FAv2_BWD</code> are supported in PyTorch. <code>NVTE_FUSED_ATTN</code> and <code>NVTE_ALLOW_NONDETERMINISTIC_ALGO</code> are supported in both PyTorch and JAX.\n",
+    "Environment variables <code>NVTE_FLASH_ATTN</code>, <code>NVTE_UNFUSED_ATTN</code>, and <code>NVTE_FUSED_ATTN_USE_FAv2_BWD</code> are supported in PyTorch. <code>NVTE_FUSED_ATTN</code> and <code>NVTE_ALLOW_NONDETERMINISTIC_ALGO</code> are supported in both PyTorch and JAX.\n",
```
KshitijLakhani, line 390: No `NVTE_FUSED_ATTN_USE_FAv2_BWD` support in JAX ?

transformer_engine/common/fused_attn/config_and_params.cpp
```diff
@@ -0,0 +1,1160 @@
...
+        cache_cfg.batch_size = cache_cfg.bucketed_batch_size;
+      }
+    }
+  }
+
+  // attn_scale is a pass-by-value graph input and different scales can share the same cached graph
+  cache_cfg.attn_scale = 1.0f;
+
+  // Restrict each direction's key to the fields its graph actually consumes, so
+  // no redundant graphs are built and no cache misses either
```
KshitijLakhani: That's a good change to avoid redundant graphs if fwd only

transformer_engine/common/fused_attn/config_and_params.cpp
```diff
@@ -0,0 +1,1160 @@
...
+  NVTE_CHECK(attr < kNVTEFusedAttnConfigNumAttributes, "Invalid NVTEFusedAttnConfigAttribute (got ",
+             static_cast<int>(attr), ")");
+  const auto &attr_size = FusedAttnConfig::attr_sizes[attr];
+  NVTE_CHECK(size_in_bytes >= attr_size,
+             "Buffer is too small for fused attention config attribute (attribute ",
+             static_cast<int>(attr), " needs ", attr_size, " bytes, but buffer has ", size_in_bytes,
+             " bytes)");
+  NVTE_CHECK(buf != nullptr, "Invalid buffer (got NULL)");
+
+  auto &cfg = *get_fused_attn_config_mutable(config);
```
KshitijLakhani, line 582: nit: for the setter and getter (nvte_set_fused_attn_config_attribute and nvte_get_fused_attn_config_attribute) - do you think in the future it might make sense to make this a bulk API in which one can request to set/get multiple attributes in in one call? Which would then reduce multiple calls to `get_fused_attn_config_mutable()` and `get_fused_attn_config` ?

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -225,308 +226,215 @@ NVTE_QKV_Format nvte_get_kv_format(NVTE_QKV_Layout qkv_layout) {
...
-// select a backend for fused attention
-NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
-    bool is_training, NVTEDType q_dtype, NVTEDType kv_dtype, NVTE_QKV_Layout qkv_layout,
-    NVTE_Bias_Type bias_type, NVTE_Mask_Type attn_mask_type, NVTE_Softmax_Type softmax_type,
-    float dropout, size_t num_attn_heads, size_t num_gqa_groups, size_t max_seqlen_q,
-    size_t max_seqlen_kv, size_t head_dim_qk, size_t head_dim_v, int64_t window_size_left,
-    int64_t window_size_right, bool return_max_logit, bool cuda_graph, bool deterministic) {
+namespace {
+
+// The per-thread storage for the diagnostic string; it's re-used (cleared + re-populated)
```
KshitijLakhani: nit: 

```suggestion
// The per-thread storage for the diagnostic string; it is re-used (cleared + re-populated)
```

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -225,308 +226,215 @@ NVTE_QKV_Format nvte_get_kv_format(NVTE_QKV_Layout qkv_layout) {
...
-  auto cudnn_runtime_version = cudnnGetVersion();
+  using namespace transformer_engine::fused_attn;
+  const FusedAttnConfig &cfg = *get_fused_attn_config(config);
+  set_message(message, "");
 
-  // For ragged offsets we only support 32-bit prior to cuDNN 9.5
-  // Only used when THD format is requested.
+  cudnnHandle_t handle = cudnnExecutionPlanManager::Instance().GetHandle();
+  const NVTE_QKV_Format qkv_format = nvte_get_qkv_format(cfg.qkv_layout);
+  const NVTE_QKV_Layout_Group layout_group = nvte_get_qkv_layout_group(cfg.qkv_layout);
```
KshitijLakhani: nit: Although having the explicit type like NVTE_QKV_Format and NVTE_QKV_Layout_Group is almost always the better option, consider using `auto` instead ? Especially since the var names are descriptive enough to understand the data type ?
It could help reduce code verbosity

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -225,308 +226,215 @@ NVTE_QKV_Format nvte_get_kv_format(NVTE_QKV_Layout qkv_layout) {
...
-                               (dropout == 0.0 || bias_type == NVTE_Bias_Type::NVTE_NO_BIAS)) ||
-                              (is_training && deterministic && cudnn_runtime_version >= 91801 &&
-                               dropout == 0.0 && bias_type == NVTE_Bias_Type::NVTE_NO_BIAS))))) {
-      flag_arb = true;
+      (qkv_format == NVTE_THD &&
+       fused_attn::get_ragged_offset_dtype(layout_group, cfg.num_attn_heads, cfg.num_gqa_groups,
+                                           cfg.max_seqlen_q, cfg.max_seqlen_kv, cfg.head_dim_qk,
+                                           cfg.head_dim_v) == DType::kInt64);
+  if (requires_64bit_ragged_offset && cudnn_runtime_version < 90500) {
+    set_message(message,
```
KshitijLakhani: Consider having a function: `set_message_if()` as almost all uses of `set_message()` are wrapped in an if block so it could help make it one liner : `set_message_if(condition, message, "string")`

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -225,308 +226,215 @@ NVTE_QKV_Format nvte_get_kv_format(NVTE_QKV_Layout qkv_layout) {
...
-        (qkv_format == NVTE_QKV_Format::NVTE_BSHD || qkv_format == NVTE_QKV_Format::NVTE_SBHD ||
-         qkv_format == NVTE_QKV_Format::NVTE_BHSD))) &&
-      !requires_64bit_ragged_offset &&
-      // 9.10.0: known bugs with SDPA FP8
-      (cudnn_runtime_version != 91000) && !return_max_logit) {
-    backend = NVTE_Fused_Attn_Backend::NVTE_FP8;
-  } else if ((q_dtype == NVTEDType::kNVTEFloat16) || (q_dtype == NVTEDType::kNVTEBFloat16)) {
-    bool flag_arb = false;
-    if (
-        // TODO(cyang): replace with cudnn-frontend check_support for cleaner logic and better error messaging
```
KshitijLakhani, line 309: nit: Curious to know if we've confirmed whether the TODOs have been addressed via the new API ? If yes, great. If not, maybe we should so that if it a specific TODO is not addressed, we can still keep the TODO intact in the file for later. Wouldn't want to lose our logging of TODOs if not addressed :)

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -225,308 +226,215 @@ NVTE_QKV_Format nvte_get_kv_format(NVTE_QKV_Layout qkv_layout) {
...
+                "cuDNN >= 9.5.");
+    return NVTE_Fused_Attn_Backend::NVTE_No_Backend;
+  }
+
+  // THD requires padding-style mask
+  if (qkv_format == NVTE_QKV_Format::NVTE_THD &&
+      cfg.attn_mask_type != NVTE_Mask_Type::NVTE_PADDING_MASK &&
+      cfg.attn_mask_type != NVTE_Mask_Type::NVTE_PADDING_CAUSAL_MASK &&
+      cfg.attn_mask_type != NVTE_Mask_Type::NVTE_PADDING_CAUSAL_BOTTOM_RIGHT_MASK) {
+    set_message(message,
```
KshitijLakhani: nit: All these are prime contenders for `set_message_if()` wrapper

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -225,308 +226,215 @@ NVTE_QKV_Format nvte_get_kv_format(NVTE_QKV_Layout qkv_layout) {
...
+                "THD format requires PADDING / PADDING_CAUSAL / PADDING_CAUSAL_BOTTOM_RIGHT mask.");
+    return NVTE_Fused_Attn_Backend::NVTE_No_Backend;
+  }
+
+  // cuDNN does not support pre-scale bias
```
KshitijLakhani: IIUC: 
```suggestion
  // TE's cuDNN fused-attention graph does not represent pre-scale bias.
```

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -10,103 +10,89 @@
...
+  int64_t page_size_v = static_cast<int64_t>(cfg.page_size_v);
+  int64_t max_pages_per_seq_k = static_cast<int64_t>(cfg.max_pages_per_seq_k);
+  int64_t max_pages_per_seq_v = static_cast<int64_t>(cfg.max_pages_per_seq_v);
+  int64_t bias_b = static_cast<int64_t>(cfg.bias_batch_size);
+  int64_t bias_h = static_cast<int64_t>(cfg.bias_num_heads);
+  int64_t bias_sq = static_cast<int64_t>(cfg.bias_seqlen_q);
+  int64_t bias_skv = static_cast<int64_t>(cfg.bias_seqlen_kv);
+  const bool is_training = cfg.is_training;
+  const bool return_max_logit = cfg.return_max_logit;
+  float scaling_factor = cfg.attn_scale;
```
KshitijLakhani: nit: I was confused as to why this wasn't a const but on further probing realized that is must be non-const because cuDNN's variant pack stores pass-by-value inputs as void*.
Maybe adding a comment for this would be helpful ?

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -225,308 +226,215 @@ NVTE_QKV_Format nvte_get_kv_format(NVTE_QKV_Layout qkv_layout) {
...
+                                    input_cu_seqlens_kv, input_cu_seqlens_q_padded,
+                                    input_cu_seqlens_kv_padded, input_page_table_k,
+                                    input_page_table_v, input_rng_state, wkspace, p.stream, handle);
+  } else if (fused_attention_backend == NVTE_Fused_Attn_Backend::NVTE_FP8) {
+    fused_attn_fp8_fwd(cfg, input_Q, input_K, input_V, input_SoftmaxOffset, input_output_S,
+                       output_O, p.Aux_CTX_Tensors, input_cu_seqlens_q, input_cu_seqlens_kv,
+                       input_rng_state, wkspace, p.stream, handle);
   } else {
-    backend = NVTE_Fused_Attn_Backend::NVTE_No_Backend;
+    const char *reject_reason =
```
KshitijLakhani: Consider `const char *const` so that not only can it not be modified through the pointer (const char *) but also it cannot e modified via redirection (`const char *const`) 

Currently:   reject_reason[0] = 'X';  // Error, reject_reason = "other"; // Allowed
After the suggestedchange:   reject_reason[0] = 'X';  // Error, reject_reason = "other"; // Error

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -10,103 +10,89 @@
...
-  bool is_padding = ((mask_type == NVTE_Mask_Type::NVTE_PADDING_MASK) ||
-                     (mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_MASK) ||
-                     (mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_BOTTOM_RIGHT_MASK));
-  if (is_bottom_right && s_q == s_kv && !is_padding) {
-    is_causal = true;
-    is_bottom_right = false;
-    bottom_right_diagonal = false;
-  }
+  bool is_causal_bottom_right = cfg.is_causal_bottom_right;
+  bool is_padding = cfg.is_padding;
```
KshitijLakhani: const ?

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -10,103 +10,89 @@
...
   bool is_softmax_offset = (softmax_type != NVTE_Softmax_Type::NVTE_VANILLA_SOFTMAX);
   bool is_dropout = (is_training && dropout_probability != 0.0f);
-  NVTE_QKV_Format q_format = nvte_get_q_format(qkv_layout);
-  NVTE_QKV_Format kv_format = nvte_get_kv_format(qkv_layout);
-  bool is_ragged_q = (q_format == NVTE_QKV_Format::NVTE_THD);
-  bool is_ragged_kv = (kv_format == NVTE_QKV_Format::NVTE_THD);
+  bool is_ragged_q = cfg.is_ragged_q;
+  bool is_ragged_kv = cfg.is_ragged_kv;
```
KshitijLakhani: const ?

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -205,16 +151,26 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
...
+      // Lock the map lookup, not the build, so different graphs can build in parallel
+      graph_and_tensors cached_graph{};
+      bool cache_hit = false;
+      {
+        std::lock_guard<std::mutex> shared_cache_lock(sdpa_f16_fprop_cache_mutex);
+        auto it = cache.find(descriptor);
+        cache_hit = (it != cache.end());
+        if (cache_hit) cached_graph = it->second;
+      }
+      graph_cache_debug::record_cache_lookup("fwd", cache_hit, cfg);
```
KshitijLakhani: I think it makes sense to have the recording outside the mutex's scope so that the mutex is not held when performing (slow) I/o ops, however, this would mean that the recorded logs for the cache ops may not reflect exact wall clock ordering. I think it is vital to mention this in the docs/code if not already so that the users are aware of this

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -640,52 +610,58 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
...
+  bool is_causal_bottom_right = cfg.is_causal_bottom_right;
+  bool is_padding = cfg.is_padding;
   bool is_softmax_offset = (softmax_type != NVTE_Softmax_Type::NVTE_VANILLA_SOFTMAX);
   bool is_dropout = (dropout_probability != 0.0f);
-  NVTE_QKV_Format q_format = nvte_get_q_format(qkv_layout);
-  NVTE_QKV_Format kv_format = nvte_get_kv_format(qkv_layout);
-  bool is_ragged_q = (q_format == NVTE_QKV_Format::NVTE_THD);
-  bool is_ragged_kv = (kv_format == NVTE_QKV_Format::NVTE_THD);
+  bool is_ragged_q = cfg.is_ragged_q;
+  bool is_ragged_kv = cfg.is_ragged_kv;
```
KshitijLakhani: const ?

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -640,52 +610,58 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
...
-  bool is_padding = ((mask_type == NVTE_Mask_Type::NVTE_PADDING_MASK) ||
-                     (mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_MASK) ||
-                     (mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_BOTTOM_RIGHT_MASK));
-  if (is_bottom_right && s_q == s_kv && !is_padding) {
-    is_causal = true;
-    is_bottom_right = false;
-    bottom_right_diagonal = false;
-  }
+  bool is_causal_bottom_right = cfg.is_causal_bottom_right;
+  bool is_padding = cfg.is_padding;
```
KshitijLakhani: const ?

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -1425,4 +1340,57 @@ void fused_attn_arbitrary_seqlen_bwd(
...
   }
 }
+
+std::string is_supported_f16_fwd(const FusedAttnConfig &cfg, cudnnHandle_t handle) {
+  FusedAttnConfig graph_cfg = cfg;
+  graph_cfg.is_forward = true;
+  graph_cfg.derive();
+
+  size_t workspace_size = 0;
+  try {
```
KshitijLakhani: Thanks for adding the try catch in here. I was hoping for it while reviewing the code for `fused_attn_arbitrary_seqlen_fwd_impl()`

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -772,16 +708,24 @@ void fused_attn_arbitrary_seqlen_bwd_impl(
                    std::shared_ptr<fe::graph::Tensor_attributes>,   // dropout_seed
                    std::shared_ptr<fe::graph::Tensor_attributes>>;  // dropout_offset
 
-    using CacheType = std::map<FADescriptor_v1, graph_and_tensors>;
-    static thread_local CacheType sdpa_f16_bprop_cache;
+    using CacheType = std::map<FusedAttnConfig, graph_and_tensors>;
+    static CacheType sdpa_f16_bprop_cache;
+    static std::mutex sdpa_f16_bprop_cache_mutex;
```
KshitijLakhani: maybe nit and overthinking on my part but would it better to reverse the order of creation ?
Because the order of destruction will be reverse of creation, we'd like to first destroy the resource (cache) and then the mutex guarding it, right ?

It may also be beneficial (to make it mistake proof) if we tie these together in a struct with the suggested new ordering above so that if ever anyone else touches the cache and mutex code in the future they do not need to worry about the individual object ordering (destroying the struct object is all they'd care about and we can take care of the reordering in the struct object)

KshitijLakhani (changes requested):

cyanguwa:
Note to self: integrate these changes to this PR, thanks to @sudhakarsingh27.
https://github.com/cyanguwa/TransformerEngine/pull/5/changes

cc #3092 

PR 2964 addresses the following points from the above PR:
Mixed THD incorrectly passes the C++ OR condition
Unsupported SM80 + old cuDNN reaches runtime and crashes
Clear fallback/debug reason

These points aren't fully addressed in 2964:
Enable valid SM80 THD execution with cuDNN 9.18.1+
Correct SM8x Stats/LSE/Max shapes
Mixed-layout Python prefilters

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -1108,7 +1125,7 @@ void fused_attn_fp8_bwd_impl(
       void* devActualSeqlenQ = static_cast<int8_t*>(workspace) + plan_workspace_size;
       void* devActualSeqlenKV = static_cast<int8_t*>(devActualSeqlenQ) + b * sizeof(int32_t);
       cu_seqlens_to_actual_seqlens<<<grid, nthreads_per_block, 0, stream>>>(
-          b, b, static_cast<const int32_t*>(devPtrcuSeqlensQ),  // TODO(pass max_b)
+          b, b, static_cast<const int32_t*>(devPtrcuSeqlensQ),  // TODO(pass bucketed_batch_size)
```
KshitijLakhani: Is this TODO for the future ? if yes, 
```suggestion
          b, b, static_cast<const int32_t*>(devPtrcuSeqlensQ),  // TODO(<GH username>): pass bucketed_batch_size
```

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -493,39 +502,61 @@ void fused_attn_fp8_fwd_impl(
...
   bool is_bias = (bias_type == NVTE_Bias_Type::NVTE_POST_SCALE_BIAS);
   bool is_alibi = (bias_type == NVTE_Bias_Type::NVTE_ALIBI);
   bool is_causal = ((mask_type == NVTE_Mask_Type::NVTE_CAUSAL_MASK) ||
                     (mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_MASK));
-  bool is_padding = ((mask_type == NVTE_Mask_Type::NVTE_PADDING_MASK) ||
-                     (mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_MASK));
+  bool is_causal_bottom_right = cfg.is_causal_bottom_right;
+  bool is_padding = cfg.is_padding;
   bool is_dropout = (dropout_probability != 0.0f);
   bool is_softmax_offset = (softmax_type != NVTE_Softmax_Type::NVTE_VANILLA_SOFTMAX);
```
KshitijLakhani: const ?

transformer_engine/common/fused_attn/fused_attn_fp8.cu
KshitijLakhani: outide this PR's scope but if possible: const ?

KshitijLakhani: const ?

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -1016,24 +1020,36 @@ void fused_attn_fp8_bwd_impl(
...
       auto return_tuple =
           std::tuple_cat(std::make_tuple(mha_graph), key_tensors_tuple, mxfp8_tensors_tuple,
                          bias_tuple, softmax_offset_tuple, padding_tuple, dropout_tuple);
-      cache.insert({descriptor, return_tuple});
-
-      return return_tuple;
+      graph_cache_debug::record_build("bwd");
+      // Lock the insert. If another thread inserted a graph for the same key while we were building,
+      // use their graph (it's the same as ours) and discard our graph.
+      {
```
KshitijLakhani: Maybe we already have this, but if not, it might be useful to add a cache specific test especially since that's a pretty imp component of graphing in TE attention.

Add a single-thread test that queries one config twice, then executes fused attention with matching params. 

Assert the first query causes one miss/build, while the second query and execution are hits with no additional build. Maybe then modify one graph-defining field and assert exactly one new miss/build—this directly catches broken key normalization and unintended recompilation.

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -1370,4 +1382,64 @@ void fused_attn_fp8_bwd(
...
+std::string is_supported_fp8_fwd(const FusedAttnConfig& cfg, cudnnHandle_t handle) {
+  FusedAttnConfig graph_cfg = cfg;
+  graph_cfg.is_forward = true;
+  graph_cfg.derive();
+
+  size_t workspace_size = 0;
+  try {
+    fused_attn::fused_attn_fp8_fwd_impl(
+        graph_cfg,
+        /*devPtrQ=*/nullptr, /*devPtrK=*/nullptr, /*devPtrV=*/nullptr,
```
KshitijLakhani: thanks for adding the args as comments here

transformer_engine/common/fused_attn/graph_cache_debug.h
```diff
@@ -0,0 +1,235 @@
...
+  static const bool on = [] {
+    const char *e = std::getenv("NVTE_FUSED_ATTN_CACHE_DEBUG");
+    return e != nullptr && e[0] != '\0' && e[0] != '0';
+  }();
+  return on;
+}
+
+// More readable, shorter thread IDs (0, 1, 2, ...).
+inline unsigned thread_seq_id() {
+  static std::atomic<unsigned> next{0};
```
KshitijLakhani, line 132: This got me thinking about logging device id a bit more

IIUC, the cache key includes device ID, but the debug events omit it right ?.
Could we log descriptor.device_id and pass the normalized descriptor to the recorder? This would make multi-GPU cache behavior diagnosable.

transformer_engine/common/fused_attn/graph_cache_debug.h
```diff
@@ -0,0 +1,235 @@
...
+    return true;
+  }();
+  (void)registered;
+}
+
+}  // namespace graph_cache_debug
+}  // namespace fused_attn
+}  // namespace transformer_engine
+
+#endif  // TRANSFORMER_ENGINE_COMMON_FUSED_ATTN_GRAPH_CACHE_DEBUG_H_
```
KshitijLakhani, line 531: If it is not too much effort, would be nice to see an example of what this logs looks like for the different diagnostics enabled. This is a good to have only so feel free to skip over

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -708,6 +984,488 @@ class AttentionShape {
...
+    return set_attr(kNVTEFusedAttnFwdParamsStream, val);
+  }
+
+ private:
+  // Common implementation for every setter: copy the value to a local variable,
+  // forward its address and size to the C API, and return *this for chaining.
+  template <typename T>
+  FusedAttnFwdParamsWrapper &set_attr(NVTEFusedAttnFwdParamsAttribute attr, T val) noexcept {
+    nvte_set_fused_attn_fwd_params_attribute(params_, attr, &val, sizeof(val));
+    return *this;
```
KshitijLakhani, line 1306: Chaining is a good addition to this 👍

transformer_engine/common/fused_attn/config_and_params.cpp
cyanguwa: The setters are already chainable and I suspect providing a multi-attribute setter will not provide much convenience because we will still need to specify which value goes to which attribute. Can think a bit more about this in the future though. Thanks.

transformer_engine/common/fused_attn/fused_attn.cpp
cyanguwa: Yes, they are all kept. The message above is exactly resolved by this PR. Another TODO about "BRCM + cross-attention on sm100" is now a `reject()` block in `nvte_get_fused_attn_backend_v2()`.

transformer_engine/common/fused_attn/graph_cache_debug.h
cyanguwa: Both of these are in. Please see `attention.ipynb` for an example of the Level-2 diagnostics. Thanks.

transformer_engine/common/fused_attn/graph_cache_debug.h
cyanguwa: Please see `attention.ipynb`. Some text description is also available in the header comment section. Thanks.

docs/examples/attention/attention.ipynb
cyanguwa: We don't support FA on the Jax side, so there's no point in supporting this flag.

transformer_engine/common/fused_attn/fused_attn.cpp
cyanguwa: Done.

transformer_engine/common/fused_attn/fused_attn.cpp
cyanguwa: These calls have been moved to `FusedAttnConfig.derive()` without `const`, because their values are subject to change.

transformer_engine/common/fused_attn/fused_attn.cpp
cyanguwa: I revamped the set_message a bit, and now just a `reject()` to set the message as well as returning `NVTE_NO_BACKEND`.

transformer_engine/common/fused_attn/fused_attn.cpp
cyanguwa: The padding mask determination is now a `is_padding` flag in the Config, so we only enumerate these 3 masks once and then just reuse `is_padding` in necessary places.

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
cyanguwa: The recording has now been moved outside the mutex. Thanks.

transformer_engine/common/fused_attn/fused_attn_fp8.cu
cyanguwa: Done.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
cyanguwa: Every per-step KV length is a multiple of `s_kv_chunk`, and if there's support for that chunk, a multiple of that chunk should also be supported. cuDNN so far hasn't had sequence length limitations, other than the initial %128==0 in early versions.

transformer_engine/common/fused_attn/fused_attn_fp8.cu
cyanguwa: I've added the test here: https://github.com/cyanguwa/TransformerEngine/blob/9cf595929d903061162490f76a0eda62ea759e8e/tests/pytorch/attention/test_attention.py#L313

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
cyanguwa: Done. It's implemented here: https://github.com/cyanguwa/TransformerEngine/blob/9cf595929d903061162490f76a0eda62ea759e8e/transformer_engine/common/fused_attn/graph_cache.h#L48

transformer_engine/common/fused_attn/graph_cache.h
```diff
@@ -0,0 +1,159 @@
...
+    get_graph<kBackend, kPass, kCreateGraphFn>(cfg, handle);
+    return "";
+  } catch (const std::exception &e) {
+    if (e.what()[0] != '\0') return e.what();
+    fallback = "rejected without a reason.";
+  } catch (...) {
+    fallback = "unknown failure.";
+  }
+  return std::string("support_verdict<") + backend_name(kBackend) + ", " + pass_name(kPass) +
+         ">: " + fallback;
```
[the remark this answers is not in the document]
cyanguwa: This rarely happens and can be safely ignored.

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -24,18 +24,95 @@
...
+      params.attr("o_dtype").cast<DType>(), params.attr("do_dtype").cast<DType>(),
+      params.attr("dqkv_dtype").cast<DType>(), qkv_layout, o_format, do_format, dqkv_layout,
+      params.attr("qkv_scale_inv_format").cast<NVTE_QKV_Format>(),
+      params.attr("do_scale_inv_format").cast<NVTE_QKV_Format>(),
+      params.attr("batch_size").cast<size_t>(), params.attr("num_attn_heads").cast<size_t>(),
+      params.attr("num_gqa_groups").cast<size_t>(), params.attr("head_dim_qk").cast<size_t>(),
+      params.attr("head_dim_v").cast<size_t>(), params.attr("max_seqlen_q").cast<size_t>(),
+      params.attr("max_seqlen_kv").cast<size_t>(), params.attr("bias_batch_size").cast<size_t>(),
+      params.attr("bias_num_heads").cast<size_t>(), params.attr("bias_seqlen_q").cast<size_t>(),
+      params.attr("bias_seqlen_kv").cast<size_t>());
```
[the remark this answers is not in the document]
cyanguwa: Fixed.

docs/examples/attention/attention.ipynb
KshitijLakhani: I'm not sure what i had in mind when I added this comment 🤣
Maybe I confused myself with the fused an missed the FA. Anyways, this is good.

transformer_engine/common/fused_attn/graph_cache.h
```diff
@@ -0,0 +1,159 @@
...
+
+  if (std::shared_ptr<CacheEntry<GraphAndTensors>> cached = find()) {
+    graph_cache_debug::record_hit_miss(backend, pass, LookupResult::Hit, key);
+    return cached;
+  }
+
+  graph_cache_debug::record_hit_miss(backend, pass, LookupResult::Miss, key);
+
+  auto entry = std::make_shared<CacheEntry<GraphAndTensors>>(build());
+  graph_cache_debug::record_create_graph(backend, pass);
```
KshitijLakhani, line 110: I think recording unconditionally can cause a. mismatch right ?
`record_cache_graph()` runs before we know whether a given thread won insertion. With concurrent same-key misses, both builders increment CACHE_GRAPH, although only one entry is inserted and the losing graph is discarded. 
IIUC CACHE_GRAPH is supposed to be number of graphs cached so in the above case it might increase that number by 1 incorrectly (this is relevant for what is reported in logs and does not affect the feature code)

Could we inspect insert(...).second under the mutex, then call record_cache_graph() outside the lock only for the winner? `CREATE_GRAPH` can still count both builds because it is creating those 2 graphs but only 1 is being cached.

transformer_engine/common/fused_attn/graph_cache_debug.h
```diff
@@ -0,0 +1,516 @@
...
+  s.build_plans = c.build_plans.load(std::memory_order_relaxed);
+  s.execute = c.execute.load(std::memory_order_relaxed);
+  s.hit = c.hit.load(std::memory_order_relaxed);
+  s.miss = c.miss.load(std::memory_order_relaxed);
+  return s;
+}
+
+// Per-thread counters, so the summary can break every column down by thread and backend.
+struct ThreadCounters {
+  unsigned tid = 0;
```
KshitijLakhani, line 196: The per thread structure has only one device, which should be okay for most cases, but I'm wondering if we have:
GPU0: 10 fwd events
GPU1: 4 fwd events

Summary might say: tid=2 dev=1 hit=14 which would imply that all events belong to GPU1 when then came from GPU 0

transformer_engine/common/fused_attn/graph_cache.h
```diff
@@ -0,0 +1,159 @@
...
+  const char *fallback = nullptr;
+  try {
+    get_graph<kBackend, kPass, kCreateGraphFn>(cfg, handle);
+    return "";
+  } catch (const std::exception &e) {
+    if (e.what()[0] != '\0') return e.what();
+    fallback = "rejected without a reason.";
+  } catch (...) {
+    fallback = "unknown failure.";
+  }
```
KshitijLakhani: This seems to treat every thrown exception as a support rejected which is okay for when cuDNN is unsupported but it would also potentially incorrectly report other failures like RT error, bad_alloc or some cuda error as No backend.
Would be good if we could classify this ?

transformer_engine/common/fused_attn/graph_cache.h
cyanguwa: Fixed as you suggested: the insert result is captured under the mutex, the lock is
released, and `record_cache_graph()` fires outside the lock only when
`result.second` is true. CREATE_GRAPH still counts every build, since multiple graphs
really were built.

transformer_engine/common/fused_attn/graph_cache_debug.h
cyanguwa: Fixed. `ThreadCounters` now carries a `multi_device` flag set by a `note_device()` helper,
and the summary prints `dev=mixed` instead of naming a device when a thread's events
span more than one. The per-thread block are kept un-split by device, and the accurate per-device attribution can be found out by using level 2 diagnostics: the event lines do carry the live device ID.

transformer_engine/common/fused_attn/graph_cache.h
cyanguwa: Fixed. Now `query_support()` throws a dedicated
`detail::UnsupportedByCudnn`, `support_verdict()` catches only that, and the `catch (...)`
is gone, so a non-empty return means cuDNN said no and anything genuinely broken surfaces
as an error.

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -294,11 +404,16 @@ static void FusedAttnForwardImpl(
...
+  auto [backend, _fwd_msg] = GetFusedAttnBackendImpl(
+      is_training, deterministic, false, return_max_logit, mask_type, bias_type, window_size_left,
+      window_size_right, bottom_right_diagonal, softmax_type, JAXX_Scaling_Mode::NO_SCALING,
+      dropout_probability, scaling_factor, dtype, dtype, dtype, dtype, qkv_layout,
+      nvte_get_q_format(qkv_layout), nvte_get_q_format(qkv_layout), qkv_layout,
+      NVTE_QKV_Format::NVTE_QKV_Format_NOT_SET, NVTE_QKV_Format::NVTE_QKV_Format_NOT_SET,
+      num_segments, attn_heads, num_gqa_groups, qk_head_dim, v_head_dim, q_max_seqlen,
+      kv_max_seqlen, is_ragged ? input_batch * q_max_seqlen : 0,
+      is_ragged ? input_batch * kv_max_seqlen : 0, bias_batch, bias_heads, bias_seqlen_q,
+      bias_seqlen_kv);
```
KshitijLakhani: nit: Should we check for no backend here before populating rng state or preparing the aux tensors for bwd ?

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -57,6 +67,37 @@
 ]
 
 
+# NVTE_DEBUG = 0/1 # disables/enables debug mode, default = 0
+_NVTE_DEBUG = int(os.getenv("NVTE_DEBUG", "0"))
+# NVTE_DEBUG_LEVEL = 0/1/2 # enables increasingly verbose debug messages, default = 0
+_NVTE_DEBUG_LEVEL = int(os.getenv("NVTE_DEBUG_LEVEL", "0"))
+
+
+class AttentionLogging:
```
KshitijLakhani, line 76: nit: not a blocker for this PR but it would be good if we could unify the logging for both frameworks rather than duplicating code here

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -355,7 +521,14 @@ def abstract(
             q_head_dim,
             v_head_dim,
             config.window_size,
-            config.return_max_logit,
+            return_max_logit=config.return_max_logit,
```
KshitijLakhani: nit: consider making all keywrd args ?

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -24,18 +24,98 @@
...
-    float dropout_probability, size_t q_attn_heads, size_t kv_attn_heads, size_t q_max_seqlen,
-    size_t kv_max_seqlen, size_t qk_head_dim, size_t v_head_dim, int64_t window_size_left,
-    int64_t window_size_right, bool return_max_logit, bool deterministic) {
-  auto backend = nvte_get_fused_attn_backend(
-      is_training, static_cast<NVTEDType>(q_dtype), static_cast<NVTEDType>(kv_dtype), qkv_layout,
-      bias_type, mask_type, softmax_type, dropout_probability, q_attn_heads, kv_attn_heads,
-      q_max_seqlen, kv_max_seqlen, qk_head_dim, v_head_dim, window_size_left, window_size_right,
-      return_max_logit, false, deterministic);
-  return backend;
+static std::tuple<NVTE_Fused_Attn_Backend, std::string> GetFusedAttnBackendImpl(
```
KshitijLakhani, line 27: Could GetFusedAttnBackendImpl accept a local param struct instead of this long positional arg list?
Most adjacent dims have the same type, so I'm wondering if an accidental swap might still compiles successfully. PyT seems to rightly avoids this in its backend-query path by reading named attributes directly into `FusedAttnConfigWrapper`. 

I did check with my agent and I think a named JAX parameter struct would retain reuse between the pybind and FFI/runtime paths while providing similar safety.

tests/jax/test_fused_attn_score_mod.py
```diff
@@ -534,7 +538,7 @@ def test_dot_product_attention_plumbs_score_mod_to_fused_attn(monkeypatch):
     assert captured["attn_bias_type"] is AttnBiasType.NO_BIAS
     assert captured["qkv_layout"] is QKVLayout.BSHD_BSHD_BSHD
     assert captured["softmax_type"] is AttnSoftmaxType.VANILLA_SOFTMAX
-    assert captured["kernel_checks"][0][0][3] is QKVLayout.BSHD_BSHD_BSHD
+    assert captured["kernel_checks"][0][0][4] is QKVLayout.BSHD_BSHD_BSHD
```
KshitijLakhani: nit: Could we construct FusedAttnHelper using keyword arguments and assert kwargs["qkv_layout"] here? The index changing from 3 to 4 after adding batch_size demonstrates how brittle our earlier logic was so might be good to be explicit here
Thanks

KshitijLakhani (commented):
nit: So the fact that we use an adjusted window size for CP in one of our primitives which is different than the global window size is not something this PR introduces, however, I'm wondering ff the caching becomes less efficient because of this ?  

```
workspace query with global window
      → build/cache global-window graph in the thread-local cache

  runtime with adjusted CP window
      → different descriptor
      → cache miss
      → build/cache adjusted-window graph
```
it's a minor penalty I suppose, but I do not think we need to worry about it - thoughts ?

cyanguwa:
> nit: So the fact that we use an adjusted window size for CP in one of our primitives which is different than the global window size is not something this PR introduces, however, I'm wondering ff the caching becomes less efficient because of this ?
> 
> ```
> workspace query with global window
>       → build/cache global-window graph in the thread-local cache
> 
>   runtime with adjusted CP window
>       → different descriptor
>       → cache miss
>       → build/cache adjusted-window graph
> ```
> 
> it's a minor penalty I suppose, but I do not think we need to worry about it - thoughts ?

Done. I added a helper function to get the effective window size for CP + Ring + Striped + THD + SWA. Looking through other CP variants, it seems that they are already passing the exact config that the execution will run, during the probing/support query stage. Thanks.

tests/jax/test_fused_attn_score_mod.py
cyanguwa: Done, now it reads the name `qkv_layout` rather than positional numbers. Thanks.

transformer_engine/jax/csrc/extensions/attention.cpp
cyanguwa: Done, now `GetFusedAttnBackendImpl` takes a single arg `FusedAttnConfigWrapper cfg`. Thanks.

transformer_engine/jax/cpp_extensions/attention.py
cyanguwa: I made `FusedAttnHelper(kw_only=True)` so callers need to specify kwarg names to initialize, otherwise it would raise "TypeError". This is not a public API and I think this change should be safe (and future proof). Thanks.

transformer_engine/jax/cpp_extensions/attention.py
cyanguwa: Yes, will follow up on this in the future.

transformer_engine/jax/csrc/extensions/attention.cpp
cyanguwa: Good catch - I added an NVTE_CHECK in both fwd and bwd. Thanks.

KshitijLakhani (approved):

cyanguwa:
L1 and L2 jax distributed tests have some hangs but fused_attn related tests have run and passed.

PR #2967, opened 2026-05-07
[torch.compile][PyTorch] Prepare linear for torch compile

pggPL:
# Description

Refactor of `transformer_engine/pytorch/module/linear.py` to lift the
`Linear` module into a shape that can be wrapped in a
`torch.library.custom_op` (and matching backward op) in a follow-up PR.

## Type of change

- [x] Code refactoring

## Changes

- **Pack forward/backward state into `LinearFwdArgs` / `LinearBwdArgs`
  dataclasses.** `_linear_forward_impl`, `_linear_setup_ctx`,
  `_linear_backward` and the `_Linear.{forward,backward}` autograd
  methods all take a single structured argument instead of 25+
  positional ones. A custom op requires a fully-declared signature on
  both sides; the previous pattern of writing arbitrary
  `ctx.something = ...` attributes scattered throughout forward made
  it impossible to tell from the call site what state backward
  actually consumes. The dataclasses make the read/write contract
  explicit and grep-able. Concretely, things that used to be re-queried
  from tensor objects (`input.requires_grad`, `weight.requires_grad`,
  `bias.requires_grad`) are now captured up front as
  `input_requires_grad` / `weight_requires_grad` /
  `bias_requires_grad` and consumed as `requires_dgrad` /
  `requires_wgrad` in backward — backward no longer has to assume the
  Python tensor objects survive the op boundary.
- **Move `prepare_for_saving` / `ctx.save_for_backward` to the autograd
  boundary.** `_linear_forward_impl` returns the *raw* tensors it
  produced (along with `tensors_to_save_from_forward` aliases) and
  `_linear_setup_ctx` returns the *raw* merged tensor list it wants
  saved; `_Linear.forward` is the only place that actually calls
  `prepare_for_saving(*tensors_to_save_from_setup)` and
  `ctx.save_for_backward(...)`. This shape fell out of the compile-path
  experiments: under `torch.library.register_autograd`, the
  `setup_context` callback is the only legal place to call
  `save_for_backward`, and the helper has to hand back tensors rather
  than mutate the autograd ctx itself. Same contract is now used in
  eager so a single helper serves both modes.
- **Deduplicate saved tensors that alias forward inputs.** Save-slots
  that would alias `inp` / `weight` / `bias` are emitted as `None` and
  reconstructed in `_linear_setup_ctx` from the original refs. An
  opaque custom op cannot return aliases of its inputs (the tracer
  has no way to reason about the aliasing).
- **Minimize the `ctx_attrs` blob plumbed from forward to backward
  setup.** Anything that can be re-derived from `LinearFwdArgs`
  (`weight_quantizer`, `is_fsdp2`, `owns_input`) is recomputed in
  `_linear_setup_ctx`. The compile path needs the fake forward impl
  to return a structurally identical `ctx_attrs`, so a smaller surface
  is a smaller cross-impl contract.

# Checklist:


transformer_engine/pytorch/module/linear.py
```diff
@@ -1112,73 +1258,75 @@ class _Linear(torch.autograd.Function):
...
-        )
-        if ctx.reduce_and_update_bwd_fp8_tensors and not is_graph_capturing():
+        if bwd_args.ub_name is not None:
+            nvtx_label = f"{nvtx_label}.{bwd_args.ub_name}"
+        result = _linear_backward(bwd_args) + (None,)  # fwd_args grad slot
+        reduce_and_update_bwd_fp8_tensors = bwd_args.reduce_and_update_bwd_fp8_tensors
+        # Drop all references held by bwd_args (saved tensors, quantizers, weakrefs,
+        # main_grad closure) so they don't outlive backward via ctx under retain_graph.
+        ctx.backward_objects = None
+        del bwd_args
```
timmoon10, line 1328: If we destroy the cached state, we should also mark `backward` with [`function.once_differentiable`](https://docs.pytorch.org/docs/2.11/autograd.html#custom-function-utilities).

transformer_engine/pytorch/module/linear.py
```diff
@@ -80,6 +80,163 @@
 __all__ = ["Linear"]
 
 
+TensorOrQuantized = Union[torch.Tensor, QuantizedTensorStorage]
+
+
+@dataclass(slots=True)
+class LinearFwdArgs:
```
timmoon10, line 87: This design is nicer than exposing a long list of positional args or passing in a tuple of args, but it is less nice than exposing kwargs. However, it has some advantages for this case:
- `torch.compile` infrastructure will only need to handle one non-tensor arg to the autograd function.
- Autograd functions do some processing on each arg, so minimizing the number of non-tensor args reduces CPU overhead.
- `_Linear.forward` calls `_linear_forward_impl` and it doesn't need to be aware of the impl specifics.

transformer_engine/pytorch/module/linear.py
```diff
@@ -80,6 +80,163 @@
...
+    fuse_wgrad_accumulation: bool
+    wgrad_store: Optional[Any]
+
+    # --- Misc ---
+    cpu_offloading: bool
+    is_grad_enabled: bool
+
+
+@dataclass(slots=True)
+class LinearBwdArgs:
```
timmoon10, line 158: In typical usage, `LinearBwdArgs` is somewhat redundant with the autograd context class. But we need it so that the forward-context-saving and backward are less entangled with autograd, which will allow code reuse when we implement an alternate `torch.compile` code path.

timmoon10 (approved):
LGTM. Most of my comments are just me talking through the design.

PR #2970, opened 2026-05-08
[JAX] [PyT] Tighten SWA checks in DPA, MHA and other APIs before passing onto cuDNN fused attn & unfused attn backends

KshitijLakhani:
# Description

TE PyT attn uses `check_set_window_size()` to regulate the `window_size` based on the `attn mask` and the user passed `window_size`. This is done higher in the stack, so that a limited subset of "valid" values of the `window_size` propagate to the backends.

Via this PR TE JAX attn tries to mimic this behavior to clean up the SWA mechanism in the different TE JAX attn APIs via a common updating logic. This new function `check_set_window_size()` does not constrain the user of the API (rather strips this complexity off for the user) when using (or not using SWA). TE handles the checks and modifications to the `window_size` internally and warns the user about any canonicalization performed, when needed.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Update APIs to canonicalize the SWA before passing to the backends
- Update internal CP P2P Ring helpers to re-canonicalize when it changes the mask for internal computations
- Update tests to validate the chosen SWA window before running the tests

- nit: [PyT] Fixed a small QOL check in the right side window for causal cases so as to not actively coerce that value and instead assert it. 

## Testing

Ran local fused attn + dist fused attn tests on H100*8 and GB200*4 and they pass successfully.

## Next steps

A subsequent PR will fix any issues / missing op in Unfused DPA SWA to make the SWA infrastructure and fallback paths more robust

# Checklist:

### DIFF
--- a/transformer_engine/jax/attention.py
+++ b/transformer_engine/jax/attention.py
@@ -374,16 +374,14 @@ def check_set_window_size(
             window_size = (orig_window_size[0], 0)
             if warn:
                 warnings.warn(
-                    f"window_size should be (-1, 0) or (>=0, 0) for "
+                    "window_size should be (-1, 0) or (>=0, 0) for "
                     f"attn_mask_type={attn_mask_type_str}, got {orig_window_size}; "
                     f"coercing to {window_size}."
                 )
         # Assert if invalid window size is provided.
-        elif orig_window_size != (-1, 0) and (
-            orig_window_size[0] < 0 or orig_window_size[1] != 0
-        ):
+        elif orig_window_size != (-1, 0) and (orig_window_size[0] < 0 or orig_window_size[1] != 0):
             raise AssertionError(
-                f"window_size should be (-1, 0) or (>=0, 0) for "
+                "window_size should be (-1, 0) or (>=0, 0) for "
                 f"attn_mask_type={attn_mask_type_str}, got {orig_window_size}."
             )
     elif attn_mask_type_enum in (AttnMaskType.NO_MASK, AttnMaskType.PADDING_MASK):
@@ -394,16 +392,14 @@ def check_set_window_size(
             window_size = (-1, -1)
             if warn:
                 warnings.warn(
-                    f"window_size should be (-1, -1) or (>=0, >=0) for "
+                    "window_size should be (-1, -1) or (>=0, >=0) for "
                     f"attn_mask_type={attn_mask_type_str}, got {orig_window_size}; "
                     f"coercing to {window_size}."
                 )
         # Assert if invalid window size is provided.
-        elif orig_window_size != (-1, -1) and (
-            orig_window_size[0] < 0 or orig_window_size[1] < 0
-        ):
+        elif orig_window_size != (-1, -1) and (orig_window_size[0] < 0 or orig_window_size[1] < 0):
             raise AssertionError(
-                f"window_size should be (-1, -1) or (>=0, >=0) for "
+                "window_size should be (-1, -1) or (>=0, >=0) for "
                 f"attn_mask_type={attn_mask_type_str}, got {orig_window_size}."
             )
     else:
@@ -929,7 +925,7 @@ class SequenceDescriptor:
         """
         Acquire the seqlens/offsets for cuDNN backend.
 
-        ``window_size`` must be a ``Tuple[int, int]`` (never ``None``) 
+        ``window_size`` must be a ``Tuple[int, int]`` (never ``None``)
         and already canonicalized by check_set_window_size.
         """
         q_segment_ids, kv_segment_ids = self.segment_ids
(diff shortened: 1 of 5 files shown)


transformer_engine/jax/attention.py
```diff
@@ -343,7 +427,9 @@ def is_fused_attn_kernel_available(
     """
     To check whether the fused attention kernel is supported
     """
-    window_size_tuple = (-1, -1) if window_size is None else window_size
+    # Canonicalize at the C-extension boundary so direct callers see the same
+    # canonical encoding as users of DPA/MHA API to ensure consistency.
+    window_size_tuple = check_set_window_size(attn_mask_type, window_size)
```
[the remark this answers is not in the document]
KshitijLakhani: There are only two callers of `is_fused_attn_kernel_available()` currently and both will not trigger any warnings - i) In `DPA __call__()` -> the window-size received here is already canonicalized so no need to worry about triggering the warning and ii) in `test_distributed_fused_attn.py` -> (here input is `None` so this will not trigger any warnings)
So I think we are good for current uses. I'd like warning to be displayed for future uses of `is_fused_attn_kernel_available()` and if at that time those warnings are bombarding we can turn warnings off then but for now I think we should let these be on.

The idea is that all public facing APIs should inform the user if the SWA passed by them has been canonicalized so that when they log they are not surprised by these values.

transformer_engine/jax/attention.py
```diff
@@ -323,6 +323,90 @@ def canonicalize_attn_mask_type(attn_mask_type: str):
...
+        elif orig_window_size == (-1, -1) or (
+            orig_window_size[0] >= 0 and orig_window_size[1] != 0
+        ):
+            window_size = (orig_window_size[0], 0)
+            if warn:
+                warnings.warn(
+                    "window_size should be (-1, 0) or (>=0, 0) for "
+                    f"attn_mask_type={attn_mask_type_str}, got {orig_window_size}; "
+                    f"coercing to {window_size}."
+                )
```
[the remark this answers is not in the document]
KshitijLakhani: nit: Fixed here as well as in the Pytorch side

transformer_engine/jax/attention.py
```diff
@@ -323,6 +323,88 @@ def canonicalize_attn_mask_type(attn_mask_type: str):
...
+            raise AssertionError(
+                "window_size should be (-1, 0) or (>=0, 0) for "
+                f"attn_mask_type={attn_mask_type_str}, got {orig_window_size}."
+            )
+    elif attn_mask_type_enum in (AttnMaskType.NO_MASK, AttnMaskType.PADDING_MASK):
+        if orig_window_size is None:
+            window_size = (-1, -1)
+        # Coerce the right side window to -1.
+        elif orig_window_size == (-1, 0):
+            window_size = (-1, -1)
```
cyanguwa, line 390: I wonder if we should do this, or go the other direction and change the mask? Technically, this could be a valid combination right? no_mask/padding + swa(left, 0) -> essentially causal + swa(left,0)?

KshitijLakhani: @cyanguwa I did think a bit about this especially since I had seen it in the PyT `check_set_window_size()` too. 

Because there is a lot of downstream branching on the mask type in the primitives, and none really on the SWA window size, I'd prefer _to not coerce the mask_ and instead _coerce the SWA window size instead_ with a warning. This can also make debugging difficult (because we do change the masks for some of the CP patterns internally without the user being aware and so it just increases the chances of something going wrong when the mask has the ability to be changed in multiple places and often)

Also, smaller concern but if the mask is indeed coerced, it can give the user an incorrect understanding of the support when `is_fused_attn_kernel_available()` is called by them. They maybe asking about padding masks and may get an answer for padding_causal instead now (which can also be argued for the SWA window but I believe the mask just has more ramifications in general)

cyanguwa: Could we make "no_mask/padding + swa(left,0)" officially supported then, as a first-class citizen, just like any other combination, including "causal/padding_causal/BRCM/PBRCM + swa(left,0)", which is what it's equivalent to anyway.

KshitijLakhani: I think there's also the below cases for SWA(L,R) that are currently being rejected but should be valid: 
(L=-1, R > 0) and (L >= 0, R=-1) right ?

If we change these we have to also consider the downstream problems (for e.e.g right now the JAX side loosely translates window_size[0]=-1, i.e. “no finite SWA” in THD seqlen/offset routing. That would misclassify (-1, 0) and (-1, R > 0) as no-window cases)

So here's my plan/suggestion if you are alright with it @cyanguwa :
1. Keep the scope of this PR to create parity between JAX and PyT SWA code only (i,.e. keep current changes as is) 
2. Create another PR where we think through all the case and their downstream effects and surgically address them for PyT and JAX side (now that merging of 1. has confirmed parity) 
3. Slightly out of scope but in my plan: Need to add missing pieces in Unfused SWA support in TE JAX

PR #2971, opened 2026-05-08
[common] Grouped gemm update - nvfp4 for blackwell and fp8 blockwise hopper

### LINKED ISSUE #2455: GroupedGemm: NVFP4 via cuBLAS
Implement GroupedGemm for NVFP4 format using cuBLAS kernels. Ensure integration with GroupedTensor utilities and grouped quantization pathways for Sync-Free MoE training.

pggPL:
## Description

Adds Hopper (SM90) support to cuBLAS grouped GEMM and enables NVFP4 / FP8 block scaling recipes. 

Fixes #2455 

## Type of change

- [ ] Documentation change
- [ ] Breaking change
- [ ] Infra/Build change
- [x] Code refactoring

## Checklist


tests/cpp/operator/test_grouped_gemm.cu
```diff
@@ -145,6 +140,125 @@ Tensor make_mxfp8_operand(const std::string& name, const std::vector<size_t>& sh
...
+  Tensor nvfp4_sw(name + "_sw", shape, DType::kFloat4E2M1,
+                  /*rowwise=*/true, /*columnwise=*/false, NVTE_NVFP4_1D_SCALING);
+  nvfp4_sw.set_with_gemm_swizzled_scales(true);
+  size_t data_bytes = test::bytes(nvfp4.rowwise_shape(), nvfp4.dtype());
+  NVTE_CHECK_CUDA(cudaMemcpy(nvfp4_sw.rowwise_dptr(), nvfp4.rowwise_dptr(),
+                             data_bytes, cudaMemcpyDeviceToDevice));
+  nvte_swizzle_scaling_factors(nvfp4.data(), nvfp4_sw.data(), 0);
+  NVTE_CHECK_CUDA(cudaDeviceSynchronize());
+  return nvfp4_sw;
+}
```
ptrendx: Why can't we just swizzle the scales in the first nvte_quantize_v2 call instead of going through 2 tensors?

tests/cpp/operator/test_grouped_gemm.cu
```diff
@@ -145,6 +140,125 @@ Tensor make_mxfp8_operand(const std::string& name, const std::vector<size_t>& sh
...
+                          bool is_A, bool transposed) {
+  (void)is_A;
+  (void)transposed;
+
+  // 1. Rowwise: quantize + swizzle directly
+  Tensor rowwise = make_nvfp4_rowwise(name + "_row", shape);
+
+  // 2. Columnwise: transpose input, quantize + swizzle as rowwise of transposed shape
+  std::vector<size_t> t_shape = {shape[1], shape[0]};
+  Tensor colwise = make_nvfp4_rowwise(name + "_col", t_shape);
```
ptrendx: Both of those tensors are using different inputs (fillUniform called in both invocations of the `make_nvfp4_rowwise` function).

ptrendx (dismissed):
There are issues with the test code (most notably the rowwise and columnwise taken from different inputs).

tests/cpp/operator/test_grouped_gemm.cu
```diff
@@ -145,6 +140,125 @@ Tensor make_mxfp8_operand(const std::string& name, const std::vector<size_t>& sh
...
+                             data_bytes, cudaMemcpyDeviceToDevice));
+  nvte_swizzle_scaling_factors(nvfp4.data(), nvfp4_sw.data(), 0);
+  NVTE_CHECK_CUDA(cudaDeviceSynchronize());
+  return nvfp4_sw;
+}
+
+// Creates an NVFP4 operand with both rowwise and columnwise data, swizzled scales.
+// NVFP4 "columnwise" data is the transposed tensor quantized rowwise.
+// We quantize rowwise directly, and for columnwise we quantize the transposed input rowwise.
+Tensor make_nvfp4_operand(const std::string& name, const std::vector<size_t>& shape,
```
ptrendx: In general I don't understand why you can't just quantize the tensor in both directions at the same time, then all of those issues with using different data for both would not be there.
The FP8 blockwise counterpart is doing just that.

tests/cpp/operator/test_grouped_gemm.cu
```diff
@@ -168,104 +282,202 @@ std::vector<std::tuple<size_t, size_t, size_t>> make_shapes(ShapeCase scase) {
       // Same N (last dim), varying M and K
       return {{128, 256, 384}, {256, 256, 512}, {384, 256, 640}};
     case ShapeCase::kAllDifferent:
-    default:
       return {{128, 256, 384}, {256, 384, 512}, {384, 512, 640}};
+    case ShapeCase::kUnalignedAllSame:
+    default:
+      // (M, N, K) all multiples of 32 (MXFP8 block) and 16 (NVFP4 block), but NONE
```
ptrendx: The multiple of 16 for NVFP4 is wrong as TMA requires the alignment to 16B, which is 32 elements in case of the NVFP4. So if anything for the "not nice" shapes we should test multiple of 16 for MXFP8 (although I'm not sure if that would be currently passed by the rest of the logic in cublaslt_gemm.cu, I do have a separate PR to relax some of those requirements) and multiple of 32 for NVFP4.

ptrendx: That said, the values that you have here are actually common to both of the types and both have K being multiple of 32, so this comment is just wrong.

tests/cpp/operator/test_grouped_gemm.cu
```diff
@@ -168,104 +282,202 @@ std::vector<std::tuple<size_t, size_t, size_t>> make_shapes(ShapeCase scase) {
...
-void run_grouped_gemm_case(const TestParams& params) {
+// Compile-time version macro for Hopper grouped GEMM support (mirrors cublaslt_grouped_gemm.cu)
+#define CUBLAS_GROUPED_GEMM_HOPPER_VERSION 130400
+
+inline std::string grouped_gemm_skip_reason(InputCase input_case) {
 #if CUBLAS_VERSION < 130300
-  GTEST_SKIP() << "Grouped GEMM requires cuBLAS 13.3+, but compile-time cuBLAS version is "
-               << CUBLAS_VERSION << ".";
+  return "Grouped GEMM requires cuBLAS 13.3+, but compile-time cuBLAS version is " +
+         std::to_string(CUBLAS_VERSION) + ".";
```
ptrendx: I don't believe this is correct here. The code in TE itself needs to check both compile and runtime versions of cublas. The test should not care at all about the compilation version (since it doesn't actually use any API from cublas) and instead should check against the runtime version.

tests/cpp/operator/test_grouped_gemm.cu
```diff
@@ -168,104 +282,202 @@ std::vector<std::tuple<size_t, size_t, size_t>> make_shapes(ShapeCase scase) {
...
+                                                                b_shape, /*is_A=*/false,
+                                                                params.transb));
         break;
-      }
     }
-    D_multi.emplace_back(Tensor("D_multi" + std::to_string(i),
-                                std::vector<size_t>{M, N},
-                                DType::kBFloat16));
+    s.D_multi.emplace_back(Tensor("D_multi" + std::to_string(i),
+                                  std::vector<size_t>{M, N}, DType::kBFloat16));
```
ptrendx: Do we only support BF16 output?

tests/cpp/operator/test_grouped_gemm.cu
```diff
@@ -299,152 +511,47 @@ void run_grouped_gemm_case(const TestParams& params) {
...
-  NVTE_CHECK_CUDA(cudaMemcpy(alpha_tensor.rowwise_dptr(), alpha_vals.data(),
-                             num_gemms * sizeof(float), cudaMemcpyHostToDevice));
-  NVTE_CHECK_CUDA(cudaMemcpy(beta_tensor.rowwise_dptr(), beta_vals.data(),
-                             num_gemms * sizeof(float), cudaMemcpyHostToDevice));
-
-  const size_t setup_ws_bytes = grouped_setup_workspace_size(num_gemms);
+  AlphaBetaTensors ab = make_alpha_beta(num_gemms);
+
+  constexpr size_t cublas_ws_bytes = 32ull * 1024 * 1024;
+  const size_t setup_ws_bytes = nvte_get_grouped_gemm_setup_workspace_size(num_gemms);
```
ptrendx, line 475: TBH not a fan of the name of this function, but I guess that ship has sailed already.

tests/cpp/operator/test_grouped_gemm.cu
```diff
@@ -299,152 +511,47 @@ void run_grouped_gemm_case(const TestParams& params) {
...
-#if CUBLAS_VERSION < 130300
-  GTEST_SKIP() << "Grouped GEMM requires cuBLAS 13.3+, but compile-time cuBLAS version is "
-               << CUBLAS_VERSION << ".";
-#else
-  if (getDeviceComputeCapability() < blackwellComputeCapability) {
-    GTEST_SKIP() << "Grouped GEMM requires Blackwell (SM100) or newer.";
+  if (auto reason = grouped_gemm_skip_reason(params.input_case); !reason.empty()) {
+    GTEST_SKIP() << reason;
   }
+#if CUBLAS_VERSION >= 130300
```
ptrendx: Why do we need this guard? We are not using any cublas API here so compilation is OK and we already skipped the test at this point anyway if the cublas version is too low.

tests/cpp/operator/test_grouped_gemm.cu
```diff
@@ -34,25 +35,19 @@ enum class InputCase {
   kFP8Current,
   kBF16,
   kMXFP8,
+  kNVFP4,
```
ptrendx, line 39: BTW, why do we have another enum here that is basically the same as the scaling mode?

tests/cpp/test_common.cu
```diff
@@ -1231,67 +1271,146 @@ GroupedBuffers build_grouped_tensor(const std::vector<Tensor*>& tensors,
...
-        NVTE_CHECK_CUDA(cudaGetLastError());
-        void* dst = static_cast<char*>(buffer.get()) + scale_offsets[i];
-        const void* src = get_cpu_ptr_fn(tensors[i]);
-        NVTE_CHECK_CUDA(cudaMemcpy(dst, src, numels[i], cudaMemcpyHostToDevice));
+        NVTE_CHECK(first_dims[i] % 32 == 0,
+                   "MXFP8 grouped GEMM test: first_dim must be divisible by 32, got ",
+                   first_dims[i]);
+        NVTE_CHECK(last_dims[i] % 32 == 0,
+                   "MXFP8 grouped GEMM test: last_dim must be divisible by 32, got ",
+                   last_dims[i]);
```
ptrendx: We should be able to have 16 here, but probably not yet, see my PR #2894.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -132,119 +142,117 @@ inline int64_t compute_avg_last_dim(const transformer_engine::GroupedTensor *t)
...
-    ws.b_rows = reinterpret_cast<int *>(setup_ws_ptr + offset);
-    offset += int_size;
-    ws.b_cols = reinterpret_cast<int *>(setup_ws_ptr + offset);
-    offset += int_size;
-    ws.d_rows = reinterpret_cast<int *>(setup_ws_ptr + offset);
-    offset += int_size;
-    ws.d_cols = reinterpret_cast<int *>(setup_ws_ptr + offset);
-
+    // 8 pointer arrays (each 16-byte aligned), then 6 int arrays, then 1 float array.
+    align_ptr();
```
ptrendx, line 191: This is wrong if the base is not aligned already (since it only takes the offset into account and that at the beginning is always 0). Fixing this then produces an issue where the returned workspace size is not actually between the get workspace size call and the actual execution, since in the get workspace size call the base is set to nullptr, so always aligned. This call should therefore assume the worst-case alignment requirement when calculating the workspace size.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -303,11 +314,22 @@ inline size_t grouped_gemm_setup_workspace_size(size_t num_tensors) {
...
+             " requires cuBLAS 13.3+, but run-time cuBLAS version is ", cublas_ver);
+  if (sm < 100) {
+    NVTE_CHECK(cublas_ver >= CUBLAS_GROUPED_GEMM_HOPPER_VERSION, api_name,
+               " on Hopper (SM90) requires cuBLAS 13.4+, but run-time cuBLAS version is ",
+               cublas_ver);
+  }
+#else
+  NVTE_CHECK(sm >= 100, api_name, " requires Blackwell (SM100) or newer architecture.");
+  NVTE_CHECK(cublas_ver >= CUBLAS_GROUPED_GEMM_VERSION, api_name,
+             " requires cuBLAS 13.3+, but run-time cuBLAS version is ", cublas_ver);
```
ptrendx, line 349: This seems wrong - what about the situation where we compiled against the cublas version that is not enough for any grouped gemm support (and so some stuff is not compiled I think?) but then run it on a system with newer cublas? This would pass, but the functionality would still not be there.

tests/cpp/operator/test_grouped_gemm.cu
pggPL: This is not supported for nvfp4 (and it is not easy to workaround, for example swizlle in both directions also does not work). I refactored the nvfp4 tests to use only rowwise or only columnwise.

tests/cpp/operator/test_grouped_gemm.cu
pggPL: as above, it is not supported for nvfp4

tests/cpp/operator/test_grouped_gemm.cu
pggPL: fixed

tests/cpp/operator/test_grouped_gemm.cu
pggPL: I added tests for fp16 and fp32, I haven't received any example from cublas people with quantized output. We can ofc ask them.

tests/cpp/operator/test_grouped_gemm.cu
pggPL: I add it in this PR, but I think the name is ok. Any proposition?

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
pggPL: I added alignment requirement for base pointer.

tests/cpp/test_common.cu
pggPL: I see, but currently we have 32 in docs.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
pggPL: We have compile time checks, they are in different places in the code.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -720,6 +820,75 @@ inline void set_mxfp8_scale_pointers(cublasLtMatmulDescOpaque_t &matmulDesc,
...
+                                                   &a_scale_inv_ptrs, sizeof(a_scale_inv_ptrs)));
+  NVTE_CHECK_CUBLAS(cublasLtMatmulDescSetAttribute(&matmulDesc,
+                                                   CUBLASLT_MATMUL_DESC_B_SCALE_POINTER,
+                                                   &b_scale_inv_ptrs, sizeof(b_scale_inv_ptrs)));
+#else
+  NVTE_CHECK(false,
+             "NVFP4 grouped GEMM requires cuBLAS 13.4+, but compile-time "
+             "cuBLAS version is ",
+             CUBLAS_VERSION);
+#endif  // CUBLAS_VERSION >= CUBLAS_NVFP4_GROUPED_GEMM_VERSION
```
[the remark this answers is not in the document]
vthumbe1503: This might be outdated, since check_grouped_gemm_requirements happens before grouped_gemm_supports_per_group_alpha_beta is called and so, NVFP4 support on a hopper GPU will throw error before itself.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -303,11 +325,22 @@ inline size_t grouped_gemm_setup_workspace_size(size_t num_tensors) {
...
+    NVTE_CHECK(cublas_ver >= CUBLAS_GROUPED_GEMM_HOPPER_VERSION, api_name,
+               " on Hopper (SM90) requires cuBLAS 13.4+, but run-time cuBLAS version is ",
+               cublas_ver);
+  }
+#else
+  NVTE_CHECK(sm >= 100, api_name, " requires Blackwell (SM100) or newer architecture.");
+  NVTE_CHECK(cublas_ver >= CUBLAS_GROUPED_GEMM_VERSION, api_name,
+             " requires cuBLAS 13.3+, but run-time cuBLAS version is ", cublas_ver);
+#endif
 }
```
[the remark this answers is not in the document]
vthumbe1503: Fixed

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -132,124 +142,119 @@ inline int64_t compute_avg_last_dim(const transformer_engine::GroupedTensor *t)
...
-    constexpr size_t kPtrAlignment = 16;  // Must match from_buffers
-
-    // Layout: 8 ptr arrays (each 16-byte aligned), then 6 int arrays
-    // Each ptr array takes ptr_size bytes but needs to start at 16-byte boundary
-    auto aligned_ptr_size = ((ptr_size + kPtrAlignment - 1) / kPtrAlignment) * kPtrAlignment;
-    size_t size = 8 * aligned_ptr_size + 6 * int_size;
-    size = ((size + alignment - 1) / alignment) * alignment;
-    return size;
+    const size_t raw = from_buffers(nullptr, num_tensors).total_bytes;
+    return ((raw + alignment - 1) / alignment) * alignment;
```
vthumbe1503: returning raw + alignment makes sense

vthumbe1503: Done

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -319,19 +359,90 @@ inline transformer_engine::GroupedMatmulConfig parse_grouped_gemm_config(
...
+  if (a_fp8_block || b_fp8_block) {
+    return a_fp8_block && b_fp8_block &&
+           !(a_mode == NVTE_BLOCK_SCALING_2D && b_mode == NVTE_BLOCK_SCALING_2D);
+  }
+  return a_mode == b_mode;
+}
+
+// Validates A/B scaling-mode pairing and Hopper-only FP8 block scaling support.
+// Call from every grouped GEMM entry point after operand scaling modes are known.
+inline void validate_grouped_gemm_scaling_modes(NVTEScalingMode a_mode, NVTEScalingMode b_mode,
```
ptrendx, line 418: Shouldn't this also have some guard against mxfp8 on Hopper?

ptrendx (approved):
Apart from one question about mxfp8 on Hopper guard, LGTM.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
vthumbe1503: Will add this guard in future PR

PR #2974, opened 2026-05-11
[Pytorch][Bug] DCP Checkpoint Loading Fixes for FSDP2 with QuantizedModelInit

vthumbe1503:
# Description

Fixes DCP Sync checkpoint loading for MXFP8/NVFP4. 
Fixes DCP Async checkpoint loading for all Quantization recipes
Fixes NVFP4 allgather + dequant numerical errors for fsdp2. Turns out this was due to us not setting the fsdp group as the  amax reduction group in the quantizer

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- **Untyped_storage implementation needed for FSDP2 + DCP**
  - untyped_storage is now defined for the base QuantizedTensor to return empty storage. Untyped_storage refers to the backing storage that we use to create all the internal tensors. Since we use make_wrapper_subclass to create TE QuantizedTensors, we dont have any backing storage associated/registered with the QuantizedTensor.
  - The main issue is that FSDP2 maintains sharded param tensor for checkpointing. It does so by calling view(-1) on our Quantized sharded model parameters. We return back a dequantized 1D tensor in TE. So, the sharded tensor that FSDP2 maintains for checkpointing is BF16 and Quantized sharded param is our custom FP8 tensor. It evaluates untyped_storage(BF16 sharded tensor reloaded from disk) == untyped_storage(Quantized sharded parameter) to see if the same_tensor. With us returning empty storage now, this would never be equal to sharded tensor's untyped storage.
  - 
 - **DCP Aync/Sync Checkpoint loading**
   - For Sync cases previously we were going through the route of dequantization to BF16 before saving to disk, which happened through the to_new_empty function
   - For both syn/async, dequantizing is not ideal. And so we now have .cpu() and .to() implemented for QuantizedTensor which dont go through dequantization and rather just copy inner tensors of QuantizedTensor to cpu if needed in blocking/non-blocking way.

- **NVFP4 Allgather Correctness issues** 
  - Allgather with FSDP2 was very far away from fp32 allgather for the same values. This was due to us not setting the amax reduction group in the quantizer.
  

- **TE_DType Serialization issues with DCP Checkpointing**
  - DCP uses torch.load(weights_only=True), whose Unpickler rejects every GLOBAL reference that isn't in add_safe_globals — and getattr is intentionally not allow-listed.
  - So we override the default enum reduction in pybind:
```
default:      (getattr, (tex.DType, "kFloat8E4M3"))   # needs getattr + tex.DType allow-listed
pybind override: (tex.DType, (int_value,))            # only needs tex.DType allow-listed
```
# Checklist:


transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -596,6 +607,23 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
             dtype = kwargs.get("dtype", tensor.dtype)
             device = kwargs.get("device", tensor.device)
             pin_memory = kwargs.get("pin_memory", False)
+
+            target_device = torch.device(device) if device is not None else tensor.device
+            if target_device.type == "cpu":
+                # When a CPU copy of a quantized tensor is requested (e.g. by
+                # torch DCP staging via ``x.new_empty(..., device="cpu")``), we
+                # save the high-precision values in a plain CPU dense tensor.
+                # For the DCP load path, we will re-quantize the high-precision values.
```
timmoon10: This fix seems ad hoc to me. It's not obvious why `qtensor.new_empty(..., device="cpu")` returns a quantized tensor while `qtensor.new_empty(..., device="cuda")` returns a plain tensor. I wonder if it would be cleaner to just return a plain tensor in all cases. Thoughts:
- It's uncomfortable how `new_empty` and `empty_like` would have different behavior. I suppose we could interpret `empty_like` as "make a tensor that matches the input" and `new_empty` as "call torch.empty with defaults taken from input", but that would be a private interpretation that no one else follows.
- Would this affect FSDP or CPU offloading?
- Given the weirdness, would it be worthwhile raising a warning if `new_empty` is called outside of DCP?

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -596,6 +607,23 @@ def __torch_dispatch__(cls, func, types, args, kwargs=None):
...
+                target_size = torch.Size(size) if len(size) > 0 else tensor.size()
```
timmoon10: An empty size is valid and it corresponds to a tensor with 1 entry (for the same reason `2^0=1`).
```
>>> import torch
>>> x = torch.ones(123).new_empty([])
>>> print(x.numel())
1
```
```suggestion
                target_size = size
```

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -533,6 +533,17 @@ def cpu(self, memory_format=torch.preserve_format) -> torch.Tensor:
...
+    def untyped_storage(self) -> torch.UntypedStorage:
+        """Return an empty UntypedStorage on the tensor's device.
+
+        ``QuantizedTensor`` is a ``_make_wrapper_subclass`` and has no real
+        backing storage of its own; the actual bytes live in the inner
+        buffers (e.g. ``_rowwise_data`` / ``_columnwise_data``) which are
+        an implementation detail of the quantization scheme. Need to define
+        this method to avoid DCP staging errors with FSDP2.
+        """
+        return torch.UntypedStorage(0, device=self.device)
```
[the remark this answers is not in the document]
timmoon10: The correct behavior for these functions is to fall back to the slow path for `QuantizedTensor` s, unless it has a dedicated implementation to handle quantized data.

tests/pytorch/distributed/fsdp2_tests/run_fsdp2_fused_adam.py
```diff
@@ -852,6 +817,15 @@ def test_dcp_output_parity(recipe_name, async_save):
                 atol=0.1,
                 msg=lambda x: f"Fresh model loaded from DCP checkpoint produces different output: {x}",
             )
+        elif recipe_name == "NVFP4BlockScaling":
+            # NVFP4 DCP load goes through a dequant + quant, so neec to relax tolerances
```
ptrendx: Why do we need dequant + quant here?

tests/pytorch/distributed/fsdp2_tests/run_fsdp2_model.py
```diff
@@ -243,7 +243,7 @@ def _check_fp8_fsdp2_allgather(model):
         # NVFP4 scale unpad/repad through FSDP2 introduces small numerical
         # differences vs the manual dequantize-then-allgather path.
         if isinstance(param, NVFP4Tensor):
-            tols = dict(atol=5e-4, rtol=5e-3)
+            tols = dict(atol=0.125, rtol=0.25)
```
ptrendx: Why are the tolerances so much bigger? Is it also due to the dequant+quant path? If so, the comment above is no longer relevant and should be replaced with a better one (but I would still like an explanation why we cannot just load the nvfp4 values from the checkpoint).

transformer_engine/pytorch/quantized_tensor.py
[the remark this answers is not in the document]
ptrendx: Yeah, while I don't think we use QuantizedTensors in the SplitAlongDim ever, the concat sounds plausible to be hit.

transformer_engine/pytorch/quantized_tensor.py
ptrendx: Ok, I see now why you want to dequantize. I don't think this is needed though - we should be able to create the QuantlizedTensor on the CPU and save it, no? I remember that the CPU offloading of the activations faced similar problem and already had to support some CPU ops on the QuantizedTensor anyway.

transformer_engine/pytorch/__init__.py
```diff
@@ -91,3 +91,61 @@
...
+            Float8BlockwiseQTensor._make_in_reduce_ex,
+            # The pickle stream produced by ``__reduce_ex__`` references
+            # the pybind11 enum ``transformer_engine_torch.DType`` (e.g.
+            # the ``fp8_dtype`` argument) and uses ``builtins.getattr`` to
+            # resolve both the enum members and the bound-classmethod
+            # ``_make_in_reduce_ex`` callables above. Both must be
+            # allow-listed for ``torch.load(weights_only=True)`` (used
+            # internally by DCP async-staging) to accept the stream.
+            _TE_DType,
+            getattr,
```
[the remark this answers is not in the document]
vthumbe1503: Fixed

transformer_engine/pytorch/quantized_tensor.py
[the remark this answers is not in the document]
vthumbe1503: Need to resolve this comment after going thoroughly over noop_cat consequences on Quantizedtensors

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -1000,6 +981,32 @@ def _set_data(self, tensor: torch.Tensor) -> None:
...
+    stays free of enum reductions as well.
+    """
+    return Float8Tensor(
+        data=data,
+        fp8_dtype=TE_DType(fp8_dtype),
+        fp8_scale_inv=fp8_scale_inv,
+        dtype=dtype,
+        shape=shape,
+        device=data.device,
+    )
```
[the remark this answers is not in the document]
vthumbe1503: Weight tensors that need to be checkpointed always need to have data. If not we orbably want to fail here itself rather than silently continuing

transformer_engine/pytorch/quantized_tensor.py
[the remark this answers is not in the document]
vthumbe1503: The behavior is unchanged with the change. And I would argue the implementation now is more correct with the change. untyped_storage() default implementation from QuantizedTensor(torch.Tensor) before this change, gives a storage with two properties.

1.  storage.nbytes() returns bytes based on the fake_dtype that we use to register our QuantizedTensor as a torchTensor using make_wrapper_subclass method of torch.

2.  storage.data_ptr() gives an error saying it is an invalid storage and there is no data_ptr()

Both of them is not ideal.
The first one is grossly incrorrect due to two reasons. First we manage the backing storage for the inner tensors of QuantizedTensor and torch has no idea about it. Second nbytes based on fake_dtype is misleading since that might not actually be the number of bytes we actually allocate.
Second one is causing problems with FSDP2 now since it expects some storage for identity check.

For QuantizedTensor, noop_cat today always returns an actual torch.cat which goes through a dequantization luckily due to this [condition](https://github.com/NVIDIA/TransformerEngine/blob/76c2a9e90275a6856039fbba9fcb09bc98c48605/transformer_engine/pytorch/module/_common.py#L128-L129) being true. This condition is going to be true now with the change as well since nbytes() would return 0.


If we do QuantizedTensor.data_ptr() today it gives you 0. QuantizedTensor.untyped_storage().data_ptr() will give invalid storage error which is inconsistent. And giving empty storage as empty storage will fix this inconsitency.

As far as idenity checking goes, FSDP2 does all the comparisong logic only if data_ptr() is not 0. And it also doesnt really make sense to compare two empty storages.

transformer_engine/pytorch/quantized_tensor.py
vthumbe1503: I agree it was ugly. Now we have a new solution where we have implemented the to_copy function in torch dispatch. This allows for staging the inner tensors of QuantizedTensor on CPU in a blocking/non-blocking way for sync/async DCP checkpointing.

We only do this in to_copy if dtype is unchanged. Otherwise we still go through the dequantize route.

pstjohn (approved):

tests/pytorch/distributed/fsdp2_tests/run_fsdp2_model.py
vthumbe1503: Fixed this now. amax_reduction_group was not updated for NVFP4 Dtensors before

transformer_engine/pytorch/quantized_tensor.py
vthumbe1503: Now we have CPU QuantizedTensors in PR

tests/pytorch/distributed/fsdp2_tests/run_fsdp2_fused_adam.py
vthumbe1503: We are doing it anymore

transformer_engine/pytorch/quantized_tensor.py
vthumbe1503: Changed the torch dispatch function now. So we dont have size here

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -898,44 +904,35 @@ def is_cpu(self):
...
+        with ``torch.serialization.add_safe_globals`` to keep
+        ``torch.load(weights_only=True)`` compatibility.
+        """
+        return (
+            _make_float8_tensor_in_reduce_ex,
+            (self._data, self._fp8_dtype, self._scale_inv, self.dtype, self.shape),
+        )
+
     @classmethod
     def _make_in_reduce_ex(
```
timmoon10, line 922: Now that we've moved the pickling logic to `_make_float8_tensor_in_reduce_ex`, the `_make_in_reduce_ex` class method only exists for backward compatibility. We should never update it again and it should not be copy-pasted into future quantized tensors.

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -1000,6 +997,29 @@ def _set_data(self, tensor: torch.Tensor) -> None:
...
+    fp8_dtype: TE_DType,
+    fp8_scale_inv: torch.Tensor,
+    dtype: torch.dtype,
+    shape: torch.Size,
+) -> Float8Tensor:
+    """Reconstruct a ``Float8Tensor`` from its ``__reduce_ex__`` payload.
+    Defined at module level (not as a Float8Tensor classmethod) so the pickle stream
+    references it via a single ``GLOBAL`` opcode rather than the
+    ``(getattr, (cls, name))`` reduction that bound classmethods/static
+    methods produce.
```
timmoon10: Nit: Once we merge the code, this comment will not be very useful. `__reduce_ex__` doesn't call any class method and the reader will be confused why you're bringing something irrelevant. However, it may be helpful to include this in `Float8Tensor._make_in_reduce_ex` to explain why it was deprecated.

This pattern is repeated in the other quantized tensors.

timmoon10 (dismissed):
Overall LGTM

transformer_engine/pytorch/tensor/float8_tensor.py
vthumbe1503: Added a comment that is kept only for backward compatibility

transformer_engine/pytorch/tensor/float8_tensor.py
vthumbe1503: Fixed it.

timmoon10 (approved):

PR #2975, opened 2026-05-11
[JAX] Size autotuned Triton grids per config 

tdophung:
# Description
The autotuned path in triton_call_lowering compiled all BLOCK_SIZE configs but dispatched every one with the same fixed grid sized for the smallest BLOCK_SIZE, so larger configs over-launched by the BLOCK_SIZE ratio. Make grid accept a callable(meta)->tuple evaluated per config, matching the jax-triton API. Update _permute_kernel, _unpermute_kernel, and _sort_chunks_by_map_kernel lowerings. Measured 22.6ms -> 7.4ms (3.06x) on GB200 for sort_chunks at 524k tokens, hidden=4096, fp32.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Pass the grid callable format to triton kernel call
- Make grid passed into the triton lowering call for all triton permutation kernels to be a a callable(meta)->tuple

# Checklist:


transformer_engine/jax/triton_extensions/permutation.py
```diff
@@ -589,10 +589,17 @@ def lowering(
...
+        # grid sized for the smallest BLOCK_SIZE, larger configs over-launch by
+        # the BLOCK_SIZE ratio (every extra block masks out and exits) — the
+        # cause of the perm-kernel perf regression vs jax-triton's autotuner.
+        def grid(meta):
+            return (num_tokens, triton.cdiv(hidden_size, meta["BLOCK_SIZE"]))
+
+        # block_size is only a placeholder for the constexpr signature; the
+        # autotune loop overrides BLOCK_SIZE per config (and on the old-JAX
+        # fallback path, the first config's BLOCK_SIZE is used).
         block_size = _get_min_block_size(_permute_kernel)
```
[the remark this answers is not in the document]
tdophung: updated in latest commit

transformer_engine/jax/triton_extensions/permutation.py
```diff
@@ -589,10 +589,19 @@ def lowering(
...
-        # Grid function equivalent: (num_tokens, cdiv(hidden_size, BLOCK_SIZE))
-        # Use minimum BLOCK_SIZE from autotune configs to ensure grid covers all elements
+        # Per-config grid: size by autotune-selected BLOCK_SIZE. With a fixed
+        # grid sized for the smallest BLOCK_SIZE, larger configs over-launch by
+        # the BLOCK_SIZE ratio (every extra block masks out and exits) — the
+        # cause of the perm-kernel perf regression vs jax-triton's autotuner.
+        def grid(meta):
+            return (num_tokens, triton.cdiv(hidden_size, meta["BLOCK_SIZE"]))
+
+        # block_size populates the BLOCK_SIZE entry in constexprs so the wrapper
```
jberchtold-nvidia: SR: some of these comments are more about this particular fix and reference the previous approach, which I think will be confusing in the future after this PR merges and we can no longer see the previous approach

I think we can simplify this and the comment above to something like "We use BLOCK_SIZE  in grid calculation to ensure the grid is the proper size. If the grid size is an overestimate it can significantly hurt  performance"

jberchtold-nvidia: Same for other cases where `def grid` is defined below too

transformer_engine/jax/triton_extensions/utils.py
```diff
@@ -431,15 +439,23 @@ def lowering(ctx, x, *, block_size):
...
+    # matching jax-triton's behavior).
+    def _normalize_grid(grid_tuple):
+        if isinstance(grid_tuple, int):
+            return (grid_tuple, 1, 1)
+        if len(grid_tuple) == 1:
+            return (grid_tuple[0], 1, 1)
+        if len(grid_tuple) == 2:
+            return (grid_tuple[0], grid_tuple[1], 1)
+        return tuple(grid_tuple[:3])
+
```
jberchtold-nvidia, line 458: SR: `assert isinstance(grid, (callable, tuple)), f"Argument 'grid' must be a tuple or a callable but received: type={type(grid)}, value={grid}"`

jberchtold-nvidia (commented):
Overall looks good! Left a few comments. Thanks!

transformer_engine/jax/triton_extensions/permutation.py
tdophung: Done, comments at all 3 `def grid` sites in this file
(edited from AI comments! I did NOT set it up to comment on PR )

tdophung:
2 notes:
1. The num_warp arg is now recovered to default to Triton default num warp =4 num_stages=3. I deleted the TODO to expose them as I do not think hand tuning the kernels is sustainable, but want to note that we could also try to do autotuning on num_warps and num_stages as metrics. Let me know what you think @jberchtold-nvidia 
2. this recover most of the speed lost from using our own jax triton binding. However, there is still the dummy buffer issue that causes 1 specific kernel `sort_chunks_by_map` to slow down, but only about 1.172ms slow down from 3.02ms to 4.19 ms. This I will not disable in this PR because it compromise the correctness of the kernel in CI if done so. But another direction is to introduce yet another env var to control whether the dummy buffer is there or not, for benchmarking/pef purposes while waiting for the root cause to be fixed from JAX side

jberchtold-nvidia:
> The num_warp arg is now recovered to default to Triton default num warp =4 num_stages=3. I deleted the TODO to expose them as I do not think hand tuning the kernels is sustainable, but want to note that we could also try to do autotuning on num_warps and num_stages as metrics. Let me know what you think @jberchtold-nvidia

I think the current approach gives a big speedup for now and we can leave it as you have it in this PR. We can try out the autotuning num_warps and num_stages in the future if needed

> this recover most of the speed lost from using our own jax triton binding. However, there is still the dummy buffer issue that causes 1 specific kernel sort_chunks_by_map to slow down, but only about 1.172ms slow down from 3.02ms to 4.19 ms. This I will not disable in this PR because it compromise the correctness of the kernel in CI if done so. But another direction is to introduce yet another env var to control whether the dummy buffer is there or not, for benchmarking/pef purposes while waiting for the root cause to be fixed from JAX side

Sounds good! Makes sense we can't remove due to the correctness issue. I think the current speedups in this PR are big enough that we should move forward with it as is. If we end up needing to optimize further in the future we can explore the env var to remove the dummy buffer.

@tdophung

jberchtold-nvidia (approved):
LGTM pending CI, thanks!

PR #2981, opened 2026-05-12
GGEMM+srelu kernels for MxFP8 Nemotron

sraman-rgb:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


ksivaman:
Please sign-off your commits @sraman-rgb

transformer_engine/pytorch/ops/fused/backward_grouped_mlp.py
```diff
@@ -358,8 +369,13 @@ def fuser_backward(
...
 
-        # Saved tensors from scaled SwiGLU forward
-        swiglu_in, scales = swiglu_ctx.saved_tensors
+        # Saved tensors from activation forward
+        activation_saved_tensors = activation_ctx.saved_tensors
+        if len(activation_saved_tensors) == 1:
+            swiglu_in = activation_saved_tensors[0]
+            scales = None
+        else:
+            swiglu_in, scales = activation_saved_tensors
```
vthumbe1503: I wonder if we should always use ScaledSRelu.
In cases like Srelu we can just use scales = ones tensor with SRelu

Saving different context.saved_tensors and context's other attributes for different use-cases can become quite difficult to maintain and can also make fusion logic more convoluted.

It might be easier to just support Scaled Variants of Activations in the grouped mlp block

transformer_engine/pytorch/ops/basic/activation.py
```diff
@@ -345,6 +347,100 @@ def _activation_backward_impl(self, *args, **kwargs) -> torch.Tensor:
         return tex.dsrelu(*args, **kwargs)
 
 
+class ScaledSReLU(BasicOperation):
+    r"""Squared ReLU with per-row post-scaling.
+
+    If the SReLU output has shape ``(d_1, ..., d_n)``, it is multiplied
```
vthumbe1503, line 353: Could we add a standalone unit test for this?

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -89,17 +89,30 @@ def __init__(
...
         if not self.is_supported():
             self.grouped_gemm_glu_kernel()  # Try triggering import error
             raise RuntimeError(f"{self.__class__.__name__} is not supported on this system.")
-        validate_grouped_mlp_dims(fc1, swiglu, fc2)
+        validate_grouped_mlp_dims(fc1, activation, fc2)
         # The cuDNN geglu implementation corresponds to ScaledClampedQGeGLU.
         # The act_func string should be fixed on the cuDNN FE side.
-        self._cudnn_act_func: str = "geglu" if isinstance(swiglu, ScaledClampedQGeGLU) else "swiglu"
+        if isinstance(activation, (SReLU, ScaledSReLU)):
+            self._cudnn_act_func: Optional[str] = None
```
vthumbe1503: I am not sure I understand. Is default activation function for cute dsl kernel = ScaledSrelu and thats why we pass None?

sraman-rgb: yes right

transformer_engine/pytorch/ops/fused/backward_grouped_mlp.py
```diff
@@ -306,19 +306,32 @@ def __init__(
         self,
         *,
         fc1: GroupedLinear,
-        swiglu: ScaledSwiGLU | ScaledClampedQGeGLU,
+        swiglu: Optional[ScaledSwiGLU | ScaledClampedQGeGLU] = None,
+        srelu: Optional[ScaledSReLU] = None,
```
timmoon10: Why not have a single arg?
```suggestion
        activation: Optional[FusibleOperation] = None,
```
It seems like we're adding one activation function after another, so we want interfaces that scale gracefully. Also, fused ops are basically internal to TE and these ops in particular are experimental, so backward compatibility is not a major concern.

The forward fused op should have a similar design. Changing to a consistent arg name would also let us get rid of the kwarg name messiness in the op fusion function.

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -520,6 +543,18 @@ def fuser_forward(
         return fc2_out, [(), (), ()]
 
 
+class ForwardGroupedMLP_CuTeGEMMSReLU_MXFP8(ForwardGroupedMLP_CuTeGEMMSwiGLU_MXFP8):
```
timmoon10: This is an awkward class hierarchy. It would be better to have a virtual base class that both the GLU and non-GLU functions inherit from. The backward fused ops should have a similar design.

While we're messing with the existing classes, we should reconsider the names. The "SwiGLU" op is actually used for both SwiGLU and ClampedQGeGLU, so a name like "GLU" would be better. And there's no reason to expect "SReLU" won't be applied to other activations later, so maybe "Unary" would be more general.

tests/pytorch/test_fusible_ops.py
```diff
@@ -3608,9 +3663,14 @@ def test_grouped_mlp(
             pytest.skip("single_grouped_bias requires bias=True")
         if with_quantization and dtype not in (torch.bfloat16, torch.float16):
             pytest.skip("Quantized group GEMM is only supported with BF16/FP16")
+        if activation == "scaled_srelu" and quantization != "mxfp8":
+            pytest.skip("ScaledSReLU grouped MLP fusion is only supported with MXFP8")
+        if activation == "scaled_srelu" and glu_interleave_size is not None:
```
timmoon10: Nit: This is assuming that activations are GLUs by default, and SReLU is weird. Isn't that kind of backward? In any case, it would be more logical to have a single point where we check `is_glu_activation`, and then use that everywhere.

timmoon10 (commented):
Overall looks good, but we've gotten to the point where we need to start thinking about how to gracefully handle adding new activations. It seems that every model has a different activation function.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -252,30 +273,24 @@ def fuse_grouped_mlp_ops(
...
-            or window[2].out_features % 64 != 0
-        ):
-            matches_pattern = False
-        elif window[1].glu_interleave_size != 32:
-            matches_pattern = False
+        else:
+            try:
+                validate_grouped_mlp_dims(window[0], window[1], window[2])
+            except (TypeError, ValueError):
+                matches_pattern = False
```
vthumbe1503, line 303: We would want to disable srelu fusion based on cudnn version here eventually before the merge

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -89,17 +109,26 @@ def __init__(
         self,
         *,
         fc1: GroupedLinear,
-        swiglu: ScaledSwiGLU | ScaledClampedQGeGLU,
+        activation: Optional[FusibleOperation] = None,
```
timmoon10: Nit: Python supports kwargs without defaults.
```suggestion
        activation: Optional[FusibleOperation],
```

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -516,10 +560,35 @@ def fuser_forward(
             fc2_ctx.dtype = dtype
             fc2_ctx.input_requires_grad = input_requires_grad
             fc2_ctx.weight_requires_grad = weight_requires_grad
+            fc2_ctx.recompute_input_from_dsrelu = recompute_srelu_fc2_x
```
timmoon10, line 560: This option isn't supported in the unfused `GroupedLinear` op. This is a problem because the forward and backward fusions are performed indendently, so everything needs to be compatible with the unfused op interfaces in case there are different forward and backward fusions. However, I also don't want to include this in the unfused op because this is so hyper-specific to this particular fusion.

The requirement that the fused and unfused ops are interchangeable has  causing some trouble with the grouped MLP block. It may be worth relaxing, but we would need to have some guarantee that the forward and backward fusions match exactly. I propose we change the op fuser to operate in three stages: fuse forward-backward ops together, fuse forward ops, fuse backward ops. For fused ops with matching forward and backward, we can tolerate tighter forward-backward integration.

vthumbe1503 (commented):
LGTM. We might want to wait on the cudnn release and apt cudnn guards are added.

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -281,30 +310,35 @@ def fuser_forward(
...
             "sfa_tensor": fc1_x_scales,
             "padded_offsets": split_points,
             "alpha_tensor": alpha_tensor,
             "bias_tensor": fc1_bias_packed,
             "norm_const_tensor": norm_const_tensor,
-            "prob_tensor": scales.detach().to(dtype=dtype).reshape(-1, 1, 1),
+            "prob_tensor": (
+                scales.detach().to(dtype=dtype).reshape(-1, 1, 1)
+                if scales is not None
+                else torch.ones((in_shape[0], 1, 1), dtype=torch.float32, device=device)
```
vthumbe1503: This might be a hold over from before right? And we do expect scales passed to be never None. So we can revert the change?

transformer_engine/pytorch/ops/basic/activation.py
```diff
@@ -352,10 +352,20 @@ class ScaledSReLU(BasicOperation):
...
+    Parameters
+    ----------
+    activation_recompute : bool, default = ``False``
+        Enable fused grouped MLP kernels to recompute activation outputs
+        during backward when supported instead of saving them.
     """
 
     num_extra_inputs: int = 1
 
+    def __init__(self, *, activation_recompute: bool = False) -> None:
```
timmoon10: This is not a general feature, but is specific to MLP fusions (currently grouped MLP, maybe dense MLP in the future). I'd suggest making the behavior explicit.
```suggestion
    def __init__(self, *, activation_recompute_in_mlp: bool = False) -> None:
```
We should also error out if you perform the unfused forward/backward pass with this option.

vthumbe1503: I was wondering about the same thing...Currently it is used only for fused operation.. But i dont see why we cant use it even for unfused version.


I was thinking that we can implement the feature in this basic op by actually saving None in the forward and actually recomputing srelu in backward to maintain consistency. @timmoon10  what do you think

vthumbe1503: Given that the main intent of this flag is to save memory.. even the unfused path will benefit from that if needed.

transformer_engine/pytorch/ops/_common.py
sraman-rgb: changed in the future commits

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
sraman-rgb: changed in the future commits

transformer_engine/pytorch/ops/basic/activation.py
vthumbe1503: Had a discussion offline. activation_recompute doesnt really make sense for an activation op and is almost never a use-case. So, we should rather error out.

sraman-rgb: Added in the latest commit, renamed the variable as well

vthumbe1503 (approved):
LGTM.

PR #2984, opened 2026-05-13
[PyTorch] Support for cuDNN-backed flex attention

### LINKED ISSUE #2492: [PyTorch/Jax/common] Flex attention via cuDNN
cuDNN flex attention offers more flexible support for attention masks or attention score modifications. We'd like to integrate this to TE on both the PyTorch side and Jax side.

References:
https://pytorch.org/blog/flexattention/
https://github.com/NVIDIA/cudnn-frontend/blob/0258951d4d512f4714eb1574496f4d57669b1b93/test/python/test_flexible_sdpa.py#L217

vcherepanov-nv:
# Description

Adds experimental PyTorch support for cuDNN-backed flex attention in `DotProductAttention` via a new `score_mod` callback path.

Users can pass:
- `score_mod(graph, score, tensors) -> score` for forward score modification
- optional `score_mod_bprop(graph, dP, tensors) -> dP` for backward
- optional runtime tensor dictionaries for forward/backward score-mod graph inputs

When `score_mod_bprop` is supplied, it is the user's responsibility to make it mathematically consistent with `score_mod`. TE forwards this callback to cuDNN as provided and does not derive or validate the backward score transformation automatically.

## Supported score_mod configuration

The current cuDNN-backed Flex Attention path supports:

- PyTorch `DotProductAttention` / `FusedAttention`
- FP16 or BF16 unquantized `torch.Tensor` Q/K/V inputs
- SBHD or BSHD Q/K/V layouts
- cuDNN F16/BF16 arbitrary-seqlen fused attention backend
- `attn_mask_type="no_mask"`
- `core_attention_bias_type="no_bias"` with no explicit bias tensor
- vanilla softmax
- `attention_dropout=0.0`
- `num_splits=1`

The path is currently not supported with FP8, `fp8_output`, THD format, explicit `cu_seqlens` inputs, `pad_between_seqs`, attention masks, attention bias, ALiBi, sliding-window attention, sink attention, dropout, KV cache, context parallelism, CUDA graph capture, checkpointed core attention, or `return_max_logit`.

For deterministic execution, TE passes the deterministic setting through backend selection and forwards it to cuDNN Frontend `sdpa_backward` as `use_deterministic_algorithm`. The score_mod forward `sdpa` call does not take a separate deterministic flag.

Fixes #2492.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Adds `FusedAttentionWithScoreModFunc`, a cuDNN frontend Python graph path for SDPA forward/backward with `score_mod` and `score_mod_bprop`.
- Extends `DotProductAttention` / `FusedAttention` APIs with `score_mod`, `score_mod_bprop`, `score_mod_tensors`, and `score_mod_bprop_tensors`.
- Adds backend-selection filtering so `score_mod` only selects supported cuDNN fused attention configurations.
- Adds execution-plan caching for forward and backward score-mod graphs, keyed by tensor metadata, layout, scale, callback topology, and runtime tensor metadata.
- Supports explicit `score_mod_graph_cache_key()` for stateful callbacks, while leaving unsafe unkeyed bound methods uncached.
- Executes cuDNN graphs on PyTorch's current CUDA stream and preserves SBHD/BSHD layouts without extra BHSD copies.
- Adds validation for unsupported combinations including FP8, context parallelism, THD, KV cache, explicit masks/biases, dropout, non-vanilla softmax, CUDA graph capture, and checkpointed core attention.
- Adds tests for cache-key behavior, unsafe callback caching, runtime tensor version checking, and CUDA correctness cases covering causal masking, softcap, and post-scale-bias-style score modification.

# Checklist:


transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -2075,15 +2731,49 @@ def forward(
...
+                all_quantizers = dpa_utils.get_attention_quantizers(fp8, fp8_recipe, quantizers)
                 for q in all_quantizers:
                     if isinstance(q, Float8CurrentScalingQuantizer):
                         q.with_amax_reduction = True
                         q.amax_reduction_group = (
                             cp_group[0] if cp_comm_type == "a2a+p2p" else cp_group
                         )
 
-        if context_parallel:
+        if score_mod is not None:
```
cyanguwa: I think this should be in the `else` branch, because it doesn't support context parallelism. Something like this:
`if context_parallel:
  elif score_mod is not None:
  else:`

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1014,6 +892,10 @@ def forward(
         pad_between_seqs: Optional[bool] = None,
         fp8_output: Optional[bool] = False,
         num_splits: Optional[int] = 1,
+        score_mod: Optional[Callable] = None,
+        score_mod_bprop: Optional[Callable] = None,
+        score_mod_tensors: Optional[Dict[str, torch.Tensor]] = None,
+        score_mod_bprop_tensors: Optional[Dict[str, torch.Tensor]] = None,
```
cyanguwa, line 898: Do you think it'd be clearer if we add "fprop" to the names? i.e. score_mod_fprop, score_mod_bprop, score_mod_fprop_tensors, score_mod_bprop_tensors?

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1528,6 +1432,84 @@ def forward(
...
+                    ), "score_mod_tensors must map string names to torch.Tensor instances!"
+                if score_mod_bprop_tensors is not None:
+                    assert isinstance(
+                        score_mod_bprop_tensors, dict
+                    ), "score_mod_bprop_tensors must be a dict!"
+                    assert all(
+                        isinstance(k, str) and isinstance(v, torch.Tensor)
+                        for k, v in score_mod_bprop_tensors.items()
+                    ), "score_mod_bprop_tensors must map string names to torch.Tensor instances!"
+
```
cyanguwa, line 1468: I think all these checks can go into `dpa_utils.get_attention_backend()`, and with `score_mod_xxx` args passed in (to `AttentionParams`), that utility function can return `use_fused_attention=False` if one of the checks if violated. `dpa_utils.get_attention_backend()` is used in the tests as well (by `get_available_attention_backends()`).

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1565,7 +1547,39 @@ def forward(
...
+                    -1,
+                    False,
+                    is_graph_capturing(),
+                    self.deterministic,
+                )
+                if fused_attention_backend == tex.NVTE_Fused_Attn_Backend.NVTE_No_Backend:
+                    raise ValueError(
+                        "score_mod requires a cuDNN FusedAttention backend, but no fused "
+                        "attention backend supports the provided inputs."
+                    )
```
cyanguwa: For the score_mod path, I don't think we need to call `tex.get_fused_attn_backend()` and check if it's supported or not. If anything, we should add `graph.validate() -> .... graph.build_plans()` to `dpa_utils.get_attention_backend(attention_params)`, but if that's too heavy-handed, we can only do the checks you had above (the asserts). Once those checks were added to `dpa_utils.get_attention_backend`, whether FusedAttention backend is run or not will be controlled by the following logic (just like with non-score_mod cases): 

```
(
                        use_flash_attention,
                        flash_attention_backend,
                        use_fused_attention,
                        fused_attention_backend,
                        use_unfused_attention,
                        _,
                    ) = dpa_utils.get_attention_backend(attention_params)
```

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1528,6 +1432,84 @@ def forward(
                 else:
                     pad_between_seqs = False
 
+            if score_mod is None:
```
cyanguwa, line 1437: Please label this "experimental".

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1263,6 +1247,671 @@ def convert_to_torch_float8(tensor, dtype):
...
+            _cudnn_score_mod_handles[device] = handle
+
+        stream = torch.cuda.current_stream(device).cuda_stream
+        cudnn.set_stream(handle=handle, stream=stream)
+    return handle
+
+
+def _build_cudnn_pygraph(dtype: torch.dtype, device: torch.device):
+    """Create a cuDNN frontend Python graph for F16/BF16 SDPA."""
+    import cudnn  # pylint: disable=import-outside-toplevel
```
cyanguwa: Can you import the cudnn from 3rdparty/cudnn-frontend, instead of from the environment/system-wide installation? We have control over the version in 3rdparty/cudnn-frontend, but not the system one.

tests/pytorch/attention/test_attention.py
```diff
@@ -1390,6 +1391,628 @@ def get_dummy_cuda_rng_tracker() -> CudaRNGStatesTracker:
...
+    out = _from_bhsd(torch.matmul(probs, v_bhsd), qkv_format).to(v.dtype)
+    return out.reshape(*out.shape[:-2], out.shape[-2] * out.shape[-1])
+
+
+@pytest.mark.skipif(not torch.cuda.is_available(), reason="CUDA is required.")
+@pytest.mark.skipif(get_cudnn_version() < (9, 6, 0), reason="cuDNN 9.6.0+ is required.")
+@pytest.mark.parametrize("dtype", param_types)
+@pytest.mark.parametrize("qkv_format", ["sbhd", "bshd"])
+@pytest.mark.parametrize("scalar_loss", [False, True])
+def test_dot_product_attention_score_mod(dtype, qkv_format, scalar_loss):
```
cyanguwa: Would `@pytest.mark.parameterize("score_mod", ["causal", "softcap", "post_scale_bias"])` simplify the tests a bit, so that we don't have 3 separate tests with a lot of repeated code?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1263,6 +1247,671 @@ def convert_to_torch_float8(tensor, dtype):
...
+    query_layer: torch.Tensor,
+    key_layer: torch.Tensor,
+    value_layer: torch.Tensor,
+    q_format: str,
+    kv_format: str,
+    attn_scale: float,
+    score_mod: Callable,
+    score_mod_tensors: Optional[Dict[str, torch.Tensor]],
+    output_layer: torch.Tensor,
+    stats_bhs1: Optional[torch.Tensor],
```
cyanguwa: I think we can just call this `stats`, even though it might only support bhs1 shape right now. On the C++ side, cuDNN does support th1 (for THD format) as well. Could we leave the name generic for now in case we want to add more support to it in the future?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -1263,6 +1247,671 @@ def convert_to_torch_float8(tensor, dtype):
         return output.contiguous()
 
 
+def _bhsd_dim_stride(
```
cyanguwa: We have a lot of small utility functions here - is there a way to pack them up a bit or group them in some way, so the code is easier to read? I know this is Python and we probably do need more than 2 functions (fwd+bwd) but could you please have a look into this? Thanks.

tests/pytorch/attention/test_attention.py
```diff
@@ -1390,6 +1391,628 @@ def get_dummy_cuda_rng_tracker() -> CudaRNGStatesTracker:
...
+    )
+    keep.set_data_type(cudnn.data_type.BOOLEAN)
+    return score_mod_graph.binary_select(
+        input0=dP_tensor,
+        input1=tensors["zero"],
+        mask=keep,
+    )
+
+
+def _score_mod_relative_position(score_mod_graph, score_tensor, _tensors):
```
cyanguwa: We can just call this "post_scale_bias" to be consistent with our nomenclature elsewhere.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -246,16 +246,30 @@ class AttentionParams:
...
+    has_attention_mask : bool, default = False
+        Whether an explicit attention mask tensor was provided.
+    has_core_attention_bias : bool, default = False
+        Whether an explicit core attention bias tensor was provided.
+    user_supplied_seqlens : bool, default = False
+        Whether explicit cu_seqlens metadata was provided.
+    score_mod : bool, default = False
+        Whether a score_mod callback was provided.
+    score_mod_bprop : bool, default = False
+        Whether a score_mod bprop callback was provided.
```
KshitijLakhani: nit: If this is a bool, to match `has_attention_mask`, consider `has_score_mod` and `has_score_mod_bprop` instead ?

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -647,6 +675,90 @@ def get_attention_backend(
             use_unfused_attention = False
             logger.debug("Disabling all backends for max_logit with FP8 attention")
 
+    # Filter: score_mod
+    if score_mod_bprop and not score_mod:
```
KshitijLakhani: What happens (is expected to happen) if `score_mod_bprop=False` and `score_mod=True` ?

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -647,6 +675,90 @@ def get_attention_backend(
...
+        logger.debug("Disabling all backends because score_mod_bprop requires score_mod")
+        use_flash_attention = False
+        use_flash_attention_2 = False
+        use_flash_attention_3 = False
+        use_flash_attention_4 = False
+        use_fused_attention = False
+        use_unfused_attention = False
```
KshitijLakhani: nit: Outside the scope of this PR but would be good to do in this or subsequent PR: having a function or something similar for when performing an action/query on all flash_attention vars

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -647,6 +675,90 @@ def get_attention_backend(
...
+    if score_mod:
+        if use_flash_attention_2 or use_flash_attention_3 or use_flash_attention_4:
+            logger.debug("Disabling FlashAttention for score_mod")
+        use_flash_attention = False
+        use_flash_attention_2 = False
+        use_flash_attention_3 = False
+        use_flash_attention_4 = False
```
KshitijLakhani: Consider this maybe ?: 
```suggestion
        if use_flash_attention_2 or use_flash_attention_3 or use_flash_attention_4:
            logger.debug("Disabling FlashAttention for score_mod")
            use_flash_attention = False
            use_flash_attention_2 = False
            use_flash_attention_3 = False
            use_flash_attention_4 = False
```

unless there's a good reason to do otherwise ?

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -647,6 +675,90 @@ def get_attention_backend(
...
+        if use_unfused_attention:
+            logger.debug("Disabling UnfusedDotProductAttention for score_mod")
+        use_unfused_attention = False
```
KshitijLakhani: Consider this maybe ?
```suggestion
        if use_unfused_attention:
            logger.debug("Disabling UnfusedDotProductAttention for score_mod")
            use_unfused_attention = False
```

unless there's a good reason to do otherwise ?

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -647,6 +675,90 @@ def get_attention_backend(
...
+        if cuda_graph:
+            score_mod_unsupported_reasons.append("CUDA graph capture is enabled")
+        if num_splits != 1:
+            score_mod_unsupported_reasons.append(f"num_splits = {num_splits}")
+        if score_mod_unsupported_reasons and use_fused_attention:
+            logger.debug(
+                "Disabling FusedAttention for score_mod because %s",
+                "; ".join(score_mod_unsupported_reasons),
+            )
+            use_fused_attention = False
```
KshitijLakhani, line 742: Seems like the list of supported cases is much smaller than the unsupported cases
Could you please add the supported case list in the PR description ?
That would make it easier for users to know the supported cases (especially since TE attn documentation is not yet fully up to date). Also, if a user were to ever ask supported cases we can point them to the PR description

OR

You could add a supported cases table in the comments like we have for KV cache filter in this file: see lines 762 to 770

Basically, right now a user of Flex attention does not easily know what is supported and what isn't very easily. I'd strongly suggest we add some documentation (in the comments or in the PR description for this)

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -647,6 +675,90 @@ def get_attention_backend(
...
+            or kv_format
+            not in [
+                "sbhd",
+                "bshd",
+            ]
+        ):
+            score_mod_unsupported_reasons.append(
+                f"unsupported QKV format: q_format = {q_format}, kv_format = {kv_format}"
+            )
+        if user_supplied_seqlens:
```
KshitijLakhani: I see no `if` block for `determinism` - is it because it is not needed or was it missed ?

transformer_engine/pytorch/attention/dot_product_attention/utils.py
KshitijLakhani: I did notice some sense of mutual exclusion being documented in the DotProductAttention API so maybe it is okay but I'm not sure it's expliit enough of cases supported

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
KshitijLakhani: Agreed.
Also please add a comment on top (similar to other sections in `forward()`. 
Something like : `checking compatibility for flex attn`

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1528,6 +1432,84 @@ def forward(
...
+                    "sbhd",
+                    "bshd",
+                ], "score_mod only supports SBHD/BSHD QKV formats!"
+                if score_mod_tensors is not None:
+                    assert isinstance(score_mod_tensors, dict), "score_mod_tensors must be a dict!"
+                    assert all(
+                        isinstance(k, str) and isinstance(v, torch.Tensor)
+                        for k, v in score_mod_tensors.items()
+                    ), "score_mod_tensors must map string names to torch.Tensor instances!"
+                if score_mod_bprop_tensors is not None:
```
KshitijLakhani, line 1457: Aren't these checks duplicates of the checks in `get_attention_backend()` ?
`get_attention_backend()` alone can be the source of truth and this duplications should not be needed here. The only checks that should be added in here are those that cannot be / should not be added in the `get_attention_backend()`
I'd suggest to change this to accommodate for the same

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1558,14 +1540,21 @@ def forward(
...
                 num_splits=num_splits,
+                has_attention_mask=attention_mask is not None,
+                has_core_attention_bias=core_attention_bias is not None,
+                user_supplied_seqlens=user_supplied_seqlens,
+                score_mod=score_mod is not None,
+                score_mod_bprop=score_mod_bprop is not None,
             )
             global _attention_backends
-            if is_in_onnx_export_mode():
+            if is_in_onnx_export_mode() and score_mod is None:
```
KshitijLakhani: Is this necessary here if `dpa_utils.get_attention_backend(attention_params)` does get called in the `else` block below ?
The flash, fused, unfused would be set in there anyways rgiht ?
Or am I missing something ?
cc: @cyanguwa

transformer_engine/pytorch/attention/dot_product_attention/backends.py
KshitijLakhani: I agree with this and was my first thought too.
We should club these function into a couple classes that can sit in this file at the very least.

However, I think this approach is still not the right approach. We should have a separate `flex_attention.py` file similar to `context_parallel.py` and `backends.py` can import it similar to how it imports the CP functions right now.
I strongly recommend this for two reasons : 

1. When we refactored attention as a whole early last year, the idea was to modularize attention. That was the reason CP was moved out of attention. With Flex attention's functionality and code in here being fairly decoupled from vanilla DPA, it should be easier to move it out. Leaving this code in here would add ~1000 lines of code that is not related to the vanilla DPA and would practically be undoing the refactoring work we did early last year. The same reason for moving CP to it's own file should also apply to Flex attention.
2. A developer/user of TE PyT DPA should not have to worry about the details of flex attention. Similarly someones modifying flex should not be bogged down by the details of vanilla fused attn. Hence, decoupling is important to aid with debugging as well

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -89,6 +94,25 @@
...
+    """Import the vendored cuDNN frontend if built, otherwise use the installed package."""
+    cudnn_frontend_path = str(_CUDNN_FRONTEND_PYTHON_PATH)
+    cudnn_frontend_package = _CUDNN_FRONTEND_PYTHON_PATH / "cudnn"
+    if (
+        any(cudnn_frontend_package.glob("_compiled_module*"))
+        and cudnn_frontend_path not in sys.path
+    ):
+        sys.path.insert(0, cudnn_frontend_path)
+    return importlib.import_module("cudnn")
+
```
KshitijLakhani: How about this?: 
```
def _import_cudnn_frontend():
    cudnn_frontend_path = str(_CUDNN_FRONTEND_PYTHON_PATH)
    cudnn_frontend_package = _CUDNN_FRONTEND_PYTHON_PATH / "cudnn"
    if (
        any(cudnn_frontend_package.glob("_compiled_module*"))
        and cudnn_frontend_path not in sys.path
    ):
        sys.path.insert(0, cudnn_frontend_path)
        return importlib.import_module("cudnn")

    # Fall back
    if importlib.util.find_spec("cudnn") is not None:
        return importlib.import_module("cudnn")

    # Fail with a  message
    raise ImportError(
        "cuDNN Frontend Python package not found. "
        "Install it with: pip install nvidia-cudnn-frontend"
    )
```

tests/pytorch/attention/test_attention.py
```diff
@@ -1390,6 +1391,661 @@ def get_dummy_cuda_rng_tracker() -> CudaRNGStatesTracker:
                 return out, max_logit, (None, None, None, d_softmax_offset)
 
 
+def _score_mod_causal(score_mod_graph, score_tensor, tensors):
```
KshitijLakhani: I would strongly recommend that similar to the CP tests we have a separate Flex attention test file. Firstly for modularization and secondly because the Flex attention tests do not really end up using the test_dot_product_attention() base test like other DPA tests in the file do so there's no code reuse reasons for it either.

These isolated ~800 lines of code can sit in it's own file if it isn't really using of the funtions in here directly but writing the flex tests as "new" tests or else the flex tests must reuse the DPA setup in here and integrate into that.

I've also shared more details on this in my comment in the backends.py file

cc: @cyanguwa

KshitijLakhani (changes requested):

KshitijLakhani:
Thanks for creating this PR @vcherepanov-nv 
This is great !

I was curious about:
1. Do you have benchmark numbers bases on any toy test cases you might have run ? - would be good to have those in here for users of the API. 
    1. native PyT flex vs TE PyT flex 
    2. traditional causal TE via cuDNN vs flex expressed causal TE via cuDNN
2. I've linked the GH issue in the PR description. Could you please update / close it appropriately when this PR is merged
Thanks !

transformer_engine/pytorch/attention/dot_product_attention/utils.py
vcherepanov-nv: It's a perfectly legal case, if, for instance, score_mod is used only for masking.

vcherepanov-nv:
Thanks for the thorough review!

> 1. Do you have benchmark numbers bases on any toy test cases you might have run ? - would be good to have those in here for users of the API.
>    
>    1. native PyT flex vs TE PyT flex
>    2. traditional causal TE via cuDNN vs flex expressed causal TE via cuDNN

I haven't done any benchmarking. Reportedly (from a Slack thread) score_mod can lead to significant perf gains if it allows to avoid mask materialization. For causal, I think I observed cuDNN choosing exactly the same kernel with score_mod and the explicit causal flag.

> 2. I've linked the GH issue in the PR description. Could you please update / close it appropriately when this PR is merged

Sure, thanks for linking!

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -2084,6 +2074,16 @@ def forward(
...
+            assert score_mod is None, "score_mod is not supported with context parallelism!"
+            assert (
+                score_mod_bprop is None
+            ), "score_mod_bprop is not supported with context parallelism!"
+            assert (
+                score_mod_tensors is None
+            ), "score_mod_tensors is not supported with context parallelism!"
+            assert (
+                score_mod_bprop_tensors is None
+            ), "score_mod_bprop_tensors is not supported with context parallelism!"
```
sudhakarsingh27: If all of them result in the same assert, combine the conditions and error msg to something like "flex attn not supported with context parallelism"?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -2129,6 +2129,39 @@ def forward(
                     layer_number=self.layer_number,
                     return_max_logit=self.return_max_logit,
                 )
+        elif score_mod is not None:
+            assert not fp8, "score_mod is not supported with FP8 FusedAttention!"
```
sudhakarsingh27: IMO, these and other error messages throughout the PR should say "Flex Attention" instead of "score_mod"

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
```diff
@@ -0,0 +1,821 @@
...
+    dk: Any
+    dv: Any
+    score_mod_graph_tensors: Dict[str, Any]
+    score_mod_bprop_graph_tensors: Dict[str, Any]
+    workspace_size: int
+
+
+def _finalize_cudnn_graph(graph) -> int:
+    """Build a cuDNN frontend Python graph and return its workspace size."""
+    cudnn = _import_cudnn_frontend()
```
sudhakarsingh27, line 281: Can we cache repeated use or pass `cudnn` as an argument?

tests/pytorch/attention/test_flex_attention.py
```diff
@@ -0,0 +1,682 @@
...
+    def __init__(self):
+        self.before_tanh_activation = None
+
+    def score_mod_graph_cache_key(self):
+        """Graph topology key for softcap score_mod."""
+        return ("softcap",)
+
+    def forward(self, score_mod_graph, score_tensor, tensors):
+        """Apply softcap * tanh(score / softcap)."""
+        cudnn = flex_attention._import_cudnn_frontend()
```
sudhakarsingh27, line 110: this could use some caching instead of calling `_import_cudnn_frontend` every time as I mention later

tests/pytorch/attention/test_flex_attention.py
```diff
@@ -0,0 +1,682 @@
...
+    key_0 = flex_attention._score_mod_callback_cache_key(softcap.forward)
+    key_1 = flex_attention._score_mod_callback_cache_key(softcap.forward)
+    other_key = flex_attention._score_mod_callback_cache_key(_ScoreModSoftcap().forward)
+
+    assert key_0 == key_1
+    assert key_0 == other_key
+
+
+def test_score_mod_cache_explicit_key_distinguishes_topology():
+    """Stateful score_mods can opt into caching with topology-specific keys."""
```
sudhakarsingh27, line 183: Something like this would be quite handsome info in some user-facing doc

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
sudhakarsingh27: Is Flex Attention supposed to be a backend? Should we use explicitly use `use_flex_attention` as a flag then?

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -246,16 +246,30 @@ class AttentionParams:
...
+    has_attention_mask : bool, default = False
+        Whether an explicit attention mask tensor was provided.
+    has_core_attention_bias : bool, default = False
+        Whether an explicit core attention bias tensor was provided.
+    user_supplied_seqlens : bool, default = False
+        Whether explicit cu_seqlens metadata was provided.
+    has_score_mod : bool, default = False
+        Whether a score_mod callback was provided.
+    has_score_mod_bprop : bool, default = False
+        Whether a score_mod bprop callback was provided.
```
sudhakarsingh27: it seems like they can be derived inside the function as well so that would make the number of additions to this API lesser?

sudhakarsingh27 (commented):
Thanks for the PR! A few comments;
0. Agree with all the comments from @KshitijLakhani and @cyanguwa, so just +1ed them
1. A user doc specifying the design choices and the building blocks of graph caching would be valuable.
2. `score_mod` seems like a argument more than a feature and so the error messaging could use something more substantial like "(TE/cuDNN) Flex Attention"
3. New arguments of the form `has_*` in `AttentionParams` could be avoided. If passing `score_mod`, `score_mod_tensors` (which are hefty) is the blocker, could we create a encapsulating dataclass and pass that instead? 
4. `user_supplied_seqlens` is a big vague, it seems like just a derived variable - does it degenerate to mean `pad_between_seqs=True`? 
Among other nits

tests/pytorch/attention/test_flex_attention.py
```diff
@@ -0,0 +1,682 @@
...
+    ModelConfig,
+    get_available_attention_backends,
+)
+
+param_types = [torch.float16]
+if torch.cuda.is_available() and is_bf16_available():
+    param_types.append(torch.bfloat16)
+
+
+def _score_mod_causal(score_mod_graph, score_tensor, tensors):
```
sudhakarsingh27, line 30: Could you also add type hinting to these and other methods?

transformer_engine/pytorch/attention/dot_product_attention/utils.py
vcherepanov-nv: Done. I added a `Supported score_mod configuration` section to the PR description with the supported configuration list and the main unsupported combinations, so users have a concise reference in the PR body.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
vcherepanov-nv: Not missed. Determinism is handled by the existing backend-selection path through `AttentionParams.deterministic`, and the score_mod path forwards `self.deterministic` into `FusedAttentionWithScoreModFunc`. In the cuDNN Frontend graph it is passed to `sdpa_backward` as `use_deterministic_algorithm=deterministic`. The score_mod forward `sdpa` call does not take a separate deterministic flag.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
vcherepanov-nv: Done. The DPA validation block is now explicitly labeled as experimental flex attention validation (`Validate experimental flex attention API inputs that backend selection cannot represent.`), and the PR description also calls the feature experimental.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
vcherepanov-nv: Done. The broad compatibility filtering now lives in `get_attention_backend()`. The DPA-side block only keeps API/value validation that backend selection cannot represent cleanly: callback callability, ONNX export, Q/K/V dtype consistency, unquantized tensor inputs for all Q/K/V tensors, and the structure of the runtime tensor dictionaries.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
vcherepanov-nv: Most of the broad compatibility checks have been moved into `dpa_utils.get_attention_backend()` via `AttentionRuntimeFlags`. I am keeping this as a score_mod-triggered cuDNN FusedAttention path rather than introducing a separate `use_flex_attention` backend flag.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -89,6 +90,8 @@
 _flash_attn_bwd = None
 _flash_attn_varlen_fwd = None
 _flash_attn_varlen_bwd = None
```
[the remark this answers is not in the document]
vcherepanov-nv: Cache eviction is out of scope for this PR.

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
vcherepanov-nv: Keeping this uncached for now to optimize for simplicity / line count.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
vcherepanov-nv: Keeping the public name as `score_mod` to match the existing PyTorch/FlexAttention term; `score_mod_bprop` remains the explicit backward counterpart.

tests/pytorch/attention/test_flex_attention.py
vcherepanov-nv: This is now captured in the PR description supported-configuration and cache-key notes.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
vcherepanov-nv: Keeping `score_mod` in these messages since it is more specific than the broader Flex Attention label.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1406,6 +1432,43 @@ def forward(
...
+                        score_mod_bprop_tensors, dict
+                    ), "score_mod_bprop_tensors must be a dict!"
+                    assert all(
+                        isinstance(k, str) and isinstance(v, torch.Tensor)
+                        for k, v in score_mod_bprop_tensors.items()
+                    ), "score_mod_bprop_tensors must map string names to torch.Tensor instances!"
+
             # gather attention params for get_attention_backend
             attention_params = dpa_utils.AttentionParams(
                 qkv_type=type(query_layer),
```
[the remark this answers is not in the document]
vcherepanov-nv: Addressed in 9420479a by asserting that `score_mod_bprop_tensors` requires `score_mod_bprop` and adding a regression test.

tests/pytorch/attention/test_flex_attention.py
vcherepanov-nv: Leaving these test helpers without broader type hints for now; this feels overkill for testing code.

tests/pytorch/attention/test_flex_attention.py
vcherepanov-nv: Same rationale here: keeping the test-side import helper call simple instead of adding caching.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1202,6 +1084,18 @@ def forward(
...
+            and is mutually exclusive with masks, bias, ALiBi, sink attention, dropout, FP8,
+            context parallelism, THD format, KV caching, and return_max_logit. The callback
+            signature is ``score_mod(graph, score, tensors) -> score``.
+        score_mod_bprop: Optional[Callable], default = None
+            Optional cuDNN frontend callback for the backward pass of score_mod. The callback
+            signature is ``score_mod_bprop(graph, dP, tensors) -> dP``.
+        score_mod_tensors: Optional[Dict[str, torch.Tensor]], default = None
+            Runtime tensors exposed to score_mod as cuDNN graph tensors.
+        score_mod_bprop_tensors: Optional[Dict[str, torch.Tensor]], default = None
+            Runtime tensors exposed to score_mod_bprop as cuDNN graph tensors.
```
cyanguwa: To declare the score_mod_xxx features experimental, would it be better to add a note for it in the docstring, for example, right after ".. _max_seqlen note:"?

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1528,6 +1432,46 @@ def forward(
...
+                ), "score_mod_bprop_tensors requires score_mod!"
+            else:
+                assert callable(score_mod), "score_mod must be callable!"
+                assert score_mod_bprop is None or callable(
+                    score_mod_bprop
+                ), "score_mod_bprop must be callable when provided!"
+                assert not is_in_onnx_export_mode(), "score_mod is not supported with ONNX export!"
+                assert (
+                    key_layer.dtype == query_layer.dtype and value_layer.dtype == query_layer.dtype
+                ), "score_mod requires Q, K and V tensors to have the same dtype!"
```
cyanguwa: I think this is duplicate to [this](https://github.com/NVIDIA/TransformerEngine/blob/79821e2b0f7af2eb2ba7e2e36f311d4c05bb0027/transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py#L1248) check.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1528,6 +1432,46 @@ def forward(
...
+                assert (
+                    type(query_layer) is torch.Tensor  # pylint: disable=unidiomatic-typecheck
+                    and type(key_layer) is torch.Tensor  # pylint: disable=unidiomatic-typecheck
+                    and type(value_layer) is torch.Tensor  # pylint: disable=unidiomatic-typecheck
+                ), "score_mod only supports unquantized torch.Tensor Q, K and V inputs!"
```
cyanguwa: This check can probably be moved to get_attention_backend() since we have qkv_type as a param in AttentionParams.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1348,6 +1249,9 @@ def forward(
                         seqlens_kv = cu_seqlens_kv[1:] - cu_seqlens_kv[:-1]
                     max_seqlen_kv = int((seqlens_kv.max().item() + 63) // 64 * 64)
 
+            if score_mod is not None:
+                assert inference_params is None, "score_mod is not supported with KV caching!"
+
```
cyanguwa: This check can be moved to get_attention_backend() because we have has_score_mod and inference_params in AttentionParams.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1210,6 +1104,13 @@ def forward(
             allow_non_contiguous=True,
             allow_different_data_and_param_types=self.softmax_type != "vanilla",
         ) as query_layer:
+            has_user_provided_cu_seqlens = (
+                cu_seqlens_q is not None
+                or cu_seqlens_kv is not None
+                or cu_seqlens_q_padded is not None
+                or cu_seqlens_kv_padded is not None
+            )
+
```
cyanguwa: This can be removed, and replaced by checking if `padding in attention_mask_type` because that's when those tensors are used (i.e. THD, or non-THD + padding_xxx mask).

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1563,6 +1507,15 @@ def forward(
...
                 num_splits=num_splits,
+                runtime_flags=dpa_utils.AttentionRuntimeFlags(
+                    fp8_output=fp8_output,
+                    checkpoint_core_attention=checkpoint_core_attention,
+                    has_attention_mask=attention_mask is not None,
+                    has_core_attention_bias=core_attention_bias is not None,
+                    has_user_provided_cu_seqlens=has_user_provided_cu_seqlens,
+                    has_score_mod=score_mod is not None,
+                    has_score_mod_bprop=score_mod_bprop is not None,
+                ),
```
cyanguwa: I feel the addition of AttentionRuntimeFlags deepens the structure of AttentionParams without adding much value. Can we flatten it? Also, there's already core_attention_bias_type/shape/etc in AttentionParams.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -453,6 +506,24 @@ def get_attention_backend(
...
+
+    def _disable_all_flash_attention() -> None:
+        nonlocal use_flash_attention
+        nonlocal use_flash_attention_2
+        nonlocal use_flash_attention_3
+        nonlocal use_flash_attention_4
+        use_flash_attention = False
+        use_flash_attention_2 = False
+        use_flash_attention_3 = False
+        use_flash_attention_4 = False
```
cyanguwa: Wouldn't use_flash_attention=False disable all use_flash_attention_x? I thought the relationship was use_flash_attention=True when one of use_flash_attention_x is True, but when we set use_flash_attention=False, we're effectively disabling all use_flash_attention_x.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -647,6 +718,84 @@ def get_attention_backend(
...
+            score_mod_unsupported_reasons.append("context parallelism is enabled")
+        if (
+            qkv_format == "thd"
+            or q_format not in ["sbhd", "bshd"]
+            or kv_format
+            not in [
+                "sbhd",
+                "bshd",
+            ]
+        ):
```
cyanguwa: I think you can do `if "thd" in (qkv_format, q_format, kv_format)`?

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -2084,6 +2074,12 @@ def forward(
                         )
 
         if context_parallel:
+            assert (
+                score_mod is None
+                and score_mod_bprop is None
+                and score_mod_tensors is None
+                and score_mod_bprop_tensors is None
+            ), "(TE/cuDNN) Flex Attention is not supported with context parallelism!"
```
cyanguwa: This check should have already happened through the get_attention_backend() call, right? So we don't need to make another assert here? Asserts adds CPU time.

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -2129,6 +2125,39 @@ def forward(
...
+            assert (
+                fused_attention_backend == tex.NVTE_Fused_Attn_Backend.NVTE_F16_arbitrary_seqlen
+            ), "score_mod requires the F16/BF16 cuDNN fused attention backend!"
+            assert (
+                attn_mask_type == "no_mask"
+                and core_attention_bias_type == "no_bias"
+                and core_attention_bias is None
+                and self.softmax_type == "vanilla"
+                and self.attention_dropout == 0.0
+            ), "score_mod is mutually exclusive with masks, bias, sink attention and dropout!"
```
cyanguwa: Same for these asserts regarding CPU time.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1528,6 +1432,46 @@ def forward(
...
+                    type(query_layer) is torch.Tensor  # pylint: disable=unidiomatic-typecheck
+                    and type(key_layer) is torch.Tensor  # pylint: disable=unidiomatic-typecheck
+                    and type(value_layer) is torch.Tensor  # pylint: disable=unidiomatic-typecheck
+                ), "score_mod only supports unquantized torch.Tensor Q, K and V inputs!"
+                if score_mod_tensors is not None:
+                    assert isinstance(score_mod_tensors, dict), "score_mod_tensors must be a dict!"
+                    assert all(
+                        isinstance(k, str) and isinstance(v, torch.Tensor)
+                        for k, v in score_mod_tensors.items()
+                    ), "score_mod_tensors must map string names to torch.Tensor instances!"
```
cyanguwa, line 1456: Should we let users know what the accepted keys are for this dict? Same for bprop.

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
```diff
@@ -0,0 +1,821 @@
...
+            score_mod_tensors,
+            score_mod_bprop_tensors,
+            deterministic,
+        )
+        _cudnn_score_mod_graph_cache[key] = entry
+    return entry
+
+
+class FusedAttentionWithScoreModFunc(torch.autograd.Function):
+    """cuDNN frontend Python SDPA path with score_mod callback support."""
```
cyanguwa: nit: use consistent "cuDNN f/Frontend naming"

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
```diff
@@ -0,0 +1,821 @@
...
+            q_format,
+            kv_format,
+            attn_scale,
+            score_mod,
+            score_mod_bprop,
+            score_mod_tensors,
+            score_mod_bprop_tensors,
+            deterministic,
+        )
+        _cudnn_score_mod_graph_cache[key] = entry
```
cyanguwa, line 604: Could the tensors be packed as a tuple, i.e. (query_layer, ...., deterministic), so it doesn't have to be spelt out 3 times in this function? Also, could the logic be simplified to `if key is None or _cudnn_score_mod_graph_cache.get(key) is None`?

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
```diff
@@ -0,0 +1,821 @@
...
+    else:
+        raise ValueError(f"score_mod only supports FP16/BF16 tensors, got {dtype}.")
+
+    graph = cudnn.pygraph(
+        io_data_type=io_data_type,
+        intermediate_data_type=cudnn.data_type.FLOAT,
+        compute_data_type=cudnn.data_type.FLOAT,
+        handle=_get_cudnn_current_stream_handle(cudnn, device),
+    )
+    return cudnn, graph
```
cyanguwa: Why does it return cudnn? It doesn't look like it's been used either? `_, graph = _build_cudnn_pygraph()`

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
```diff
@@ -0,0 +1,821 @@
...
+        items = (_freeze_score_mod_cache_key(item) for item in value)
+        return tuple(sorted(items, key=repr))
+    try:
+        hash(value)
+    except TypeError as exc:
+        raise TypeError(
+            "score_mod_graph_cache_key() must return a hashable value or a nested "
+            "combination of dict/list/tuple/set values."
+        ) from exc
+    return value
```
cyanguwa, line 95: cudnn-frontend seems to have a key() method on the C++ side, which hashes the json file of the important params. Not sure if they have an equivalent method on the Python side, or if their implementation is better or this in your PR, but please take a look:

https://github.com/NVIDIA/cudnn-frontend/blob/7cd11d734f843b882715a49f7441e3c5f20fd590/include/cudnn_frontend/graph_interface.h#L271

Also, for the caching of the graph:
https://github.com/NVIDIA/cudnn-frontend/blob/7cd11d734f843b882715a49f7441e3c5f20fd590/python/cudnn/experimental/ops/sdpa.py#L193

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -647,6 +718,84 @@ def get_attention_backend(
...
+                f"unsupported QKV format: q_format = {q_format}, kv_format = {kv_format}"
+            )
+        if has_user_provided_cu_seqlens:
+            score_mod_unsupported_reasons.append("explicit sequence length metadata was provided")
+        if pad_between_seqs:
+            score_mod_unsupported_reasons.append("pad_between_seqs is enabled")
+        if has_attention_mask:
+            score_mod_unsupported_reasons.append("attention_mask was provided")
+        if attn_mask_type != "no_mask":
+            score_mod_unsupported_reasons.append(f"attn_mask_type = {attn_mask_type}")
```
cyanguwa, line 716: I know you implemented causal mask as an example in the test, but does cuDNN really not work with any other mask type than "no_mask"? Should a simple mask "causal" at least work??

cyanguwa (changes requested):

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
cyanguwa: Yeah, probably not necessary. We already have all the checks related to score_mod encapsulated in the "Filter: score_mod" section I think.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
cyanguwa: Maybe we can use the wording "Flex attention only supports...", rather than "score_mod ....", because it could be a little confusing to users whether score_mod means the feature or the argument.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
vcherepanov-nv: This is just for external tensors, so names are completely up to the user, as long as naming is consistent: the same name used in callbacks and the caller code.

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
vcherepanov-nv: Hmm, not sure how any function in cuDNN FE C++ code can possibly help here. What we're constructing is a key for a callable Python object. The logic depends on whether it's a global function, a lambda, a bound method, etc.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
vcherepanov-nv: Well, that was the recommendation from the cuDNN team, not to mix it with any other options, even if it might work for some of them.
But personally, I think it's cleaner / less confusing for the user, to either go flex attn or custom options. Any mask can be implemented via score_mod, right?

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -647,6 +683,74 @@ def get_attention_backend(
...
+        if cuda_graph:
+            score_mod_unsupported_reasons.append("CUDA graph capture is enabled")
+        if num_splits != 1:
+            score_mod_unsupported_reasons.append(f"num_splits = {num_splits}")
+        if score_mod_unsupported_reasons and use_fused_attention:
+            logger.debug(
+                "Disabling FusedAttention for score_mod because %s",
+                "; ".join(score_mod_unsupported_reasons),
+            )
+            use_fused_attention = False
```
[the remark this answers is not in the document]
vcherepanov-nv: Thanks for flagging this. I do not think the main failure mode described here applies to the current code.

It is true that an unsupported `score_mod` configuration can leave FlashAttention, FusedAttention, and UnfusedDotProductAttention all disabled in `get_attention_backend()`. However, `DotProductAttention.forward()` checks that immediately after backend selection:

```python
if sum([use_flash_attention, use_fused_attention, use_unfused_attention]) == 0:
    raise ValueError(...)
```

That guard runs before the backend dispatch section and before the trailing `return None`, so this path raises a `ValueError` rather than silently returning `None`.

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
```diff
@@ -0,0 +1,792 @@
...
+
+
+def _import_cudnn_frontend():
+    """Import the vendored cuDNN frontend if built, otherwise use the installed package."""
+    cudnn_frontend_path = str(_CUDNN_FRONTEND_PYTHON_PATH)
+    cudnn_frontend_package = _CUDNN_FRONTEND_PYTHON_PATH / "cudnn"
+    if any(cudnn_frontend_package.glob("_compiled_module*")):
+        if cudnn_frontend_path not in sys.path:
+            sys.path.insert(0, cudnn_frontend_path)
+        return importlib.import_module("cudnn")
```
cyanguwa, line 32: Do we expect users to compile in the 3rdparty/cudnn-frontend directory first before using this feature? i.e. how do they get the _compiled_module? Do we need to set this up in our setup.py file so users won't have this issue?

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
```diff
@@ -0,0 +1,792 @@
...
+    q_format: str,
+    kv_format: str,
+    attn_scale: float,
+    score_mod: Callable,
+    score_mod_tensors: Optional[Dict[str, torch.Tensor]],
+    output_layer: torch.Tensor,
+    stats: Optional[torch.Tensor],
+) -> _CudnnScoreModFwdGraphEntry:
+    """Build a cached cuDNN frontend graph for score_mod fprop."""
+    cudnn = _import_cudnn_frontend()
```
cyanguwa, line 408: We're calling this in pretty much every function in this file. We could do this once at the top of the file. Thanks.

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
cyanguwa: I meant that maybe it could be a neater implementation. They usually have a matching Python API if they have a C++ key(). I wasn't sure - just wanted to point you to it in case it helps.

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
cyanguwa: This was just a question of submodule FE vs system FE that we discussed on Slack.

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
```diff
@@ -0,0 +1,792 @@
...
+            value_layer,
+            q_format,
+            kv_format,
+            attn_scale,
+            score_mod,
+            score_mod_tensors,
+            output_layer,
+            stats,
+        )
+        _cudnn_score_mod_graph_cache[key] = entry
```
cyanguwa, line 485: We should make similar changes here like in _get_cudnn_score_mod_bwd_graph. Thanks.

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
vcherepanov-nv: So what's your concern here? Python caches repeated importlib.import_module("cudnn").

vcherepanov-nv: I think it's written like this because backends.py imports flex_attention.py eagerly, and cudnn python package might be optional?

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1563,6 +1501,12 @@ def forward(
                 return_max_logit=self.return_max_logit,
                 cuda_graph=is_graph_capturing(),
                 num_splits=num_splits,
+                fp8_output=fp8_output,
+                checkpoint_core_attention=checkpoint_core_attention,
+                has_attention_mask=attention_mask is not None,
+                has_core_attention_bias=core_attention_bias is not None,
```
cyanguwa: Isn't this duplicate to `core_attention_bias_type='no_bias'`, which we already have in `AttentionParams`?

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1563,6 +1501,12 @@ def forward(
                 return_max_logit=self.return_max_logit,
                 cuda_graph=is_graph_capturing(),
                 num_splits=num_splits,
+                fp8_output=fp8_output,
+                checkpoint_core_attention=checkpoint_core_attention,
+                has_attention_mask=attention_mask is not None,
```
cyanguwa: I think we can get away without having this entry to `AttentionParams`. Instead, we can restrict `"padding" not in attn_mask_type and attn_mask_type != "arbitrary"`, because attention_mask is only passed in when there's padding, or when the mask type is "arbitrary".

cyanguwa (changes requested):

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
cyanguwa: CPU overhead. I think this could be called once at the beginning of the file?

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
cyanguwa: Will investigate this in a follow-up PR.

transformer_engine/pytorch/attention/dot_product_attention/flex_attention.py
cyanguwa: Will investigate this in a follow-up PR.

Also, if there's any way to make the feature easier to understand to users.

cyanguwa (approved):

KshitijLakhani (approved):
Deferring to @cyanguwa's thorough review of this PR and @vlad having perused/addressed my review comment sfrom before
Approving the PR so as to not hold it back
Good to merge once CI passes

PR #2985, opened 2026-05-13
[JAX] Support for cuDNN-backed flex attention

vcherepanov-nv:
# Description

Adds experimental JAX fused-attention `score_mod` support through cuDNN frontend SDPA graphs.

This introduces a `score_mod(graph, score, tensors)` callback path for `fused_attn`, plus optional `score_mod_bprop(graph, dscore, tensors)` support for backward. The Python side builds and serializes cuDNN frontend forward/backward graphs, caches graph metadata with stable callback keys, supports auxiliary tensor operands, and supports Python/NumPy scalar operands as cuDNN pass-by-value tensors. The C++ JAX extension deserializes and caches the graphs per device, then executes them through new forward/backward FFI handlers.

The Flax API now plumbs `score_mod` through `DotProductAttention`, `MultiHeadAttention`, and `TransformerLayer`. Packed QKV/KV layouts are unpacked to the separate BSHD layout when score modification is requested.

Users are responsible for supplying a mathematically correct `score_mod_bprop` for the corresponding `score_mod`; Transformer Engine wires the callback into the cuDNN graph but does not validate gradient semantics.

Current score_mod limitations:

- Requires fused attention to be enabled.
- Supports separate rank-4 `BSHD_BSHD_BSHD` Q/K/V tensors only.
- Supports FP16/BF16 Q/K/V tensors.
- Mutually exclusive with attention bias, masks, sequence descriptors, dropout, sliding-window attention, packed/ragged metadata, context parallelism, and non-vanilla softmax/softmax offset.
- Requires matching cuDNN frontend Python package and C++ headers.

Fixes # (issue)
https://github.com/NVIDIA/TransformerEngine/issues/2492

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- A new score_mod code path for the JAX FusedAttention backend
- cuDNN frontend graph serialization and JAX FFI execution for score_mod forward/backward
- Flax plumbing for DotProductAttention, MultiHeadAttention, and TransformerLayer
- Tests

# Checklist:


transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -267,6 +279,698 @@ def check_seed(self, seed, dropout_probability, is_training):
...
+            and self.is_training == other.is_training
+            and self.deterministic == other.deterministic
+        )
+
+
+_SCORE_MOD_UID_Q = 1
+_SCORE_MOD_UID_K = 2
+_SCORE_MOD_UID_V = 3
+_SCORE_MOD_UID_O = 4
+_SCORE_MOD_UID_STATS = 5
```
jberchtold-nvidia: Where do these _SCORE_MOD_UID_XXXXX come from? Is it a C/C++ enum? If so, we should make this a Python Enum that derives its values from the C/C++ enum exposed via pybind

See this enum for reference: https://github.com/NVIDIA/TransformerEngine/blob/76c2a9e90275a6856039fbba9fcb09bc98c48605/transformer_engine/jax/quantize/scaling_modes.py#L795

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -267,6 +279,698 @@ def check_seed(self, seed, dropout_probability, is_training):
...
+    )
+
+
+def _shape_dtype(value) -> jax.ShapeDtypeStruct:
+    return jax.ShapeDtypeStruct(tuple(value.shape), value.dtype)
+
+
+def _import_cudnn_for_score_mod():
+    cudnn_frontend_path = str(_CUDNN_FRONTEND_PYTHON_PATH)
+    cudnn_frontend_package = _CUDNN_FRONTEND_PYTHON_PATH / "cudnn"
```
jberchtold-nvidia: Does this work for all installation types? Editable install, system installation?

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -689,5 +696,221 @@ XLA_FFI_DEFINE_HANDLER_SYMBOL(FusedAttnBackwardHandler, FusedAttnBackwardFFI,
...
+  return registry;
+}
+
+std::mutex &ScoreModGraphRegistryMutex() {
+  static std::mutex mutex;
+  return mutex;
+}
+
+std::atomic<int64_t> &NextScoreModGraphId() {
+  static std::atomic<int64_t> next_id{1};
```
jberchtold-nvidia: Why do we need these static values? Can we simplify this to track things on the Python side instead of C++ or as explicit FFI args?

```
  static std::unordered_map<int64_t, std::shared_ptr<ScoreModGraphEntry>> registry;
...
static std::mutex mutex;
...
  static std::atomic<int64_t> next_id{1};
```

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -689,5 +696,221 @@ XLA_FFI_DEFINE_HANDLER_SYMBOL(FusedAttnBackwardHandler, FusedAttnBackwardFFI,
...
+};
+
+struct ScoreModGraphEntry {
+  PyObject *py_graph = nullptr;
+  std::vector<int64_t> user_uids;
+  std::vector<int64_t> input_uids;
+  std::vector<int64_t> output_uids;
+  std::vector<int64_t> scalar_uids;
+  std::vector<ScoreModScalarStorage> scalar_values;
+};
```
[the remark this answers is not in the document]
jberchtold-nvidia: @vcherepanov-nv This seems like a valid comment from greptile about leaking pygraphs. But I'm also not sure if that is the intended design to prevent GC freeing up a graph too early by mistake before we use it in the XLA C++ FFI. I'm not sure what the best option is here

transformer_engine/jax/attention.py
```diff
@@ -1391,10 +1391,127 @@ def _fused_attn_bwd_rule(
...
+    window_size: Optional[Tuple[int, int]],
+    context_parallel_strategy: CPStrategy,
+    context_parallel_causal_load_balanced: bool,
+    context_parallel_axis: str,
+    softmax_offset: Optional[jnp.ndarray],
+    stripe_size: int | None,
+):
+    """Validate arguments for the cuDNN frontend score_mod path."""
+    header = "score_mod fused_attn"
+    if qkv_layout is not QKVLayout.BSHD_BSHD_BSHD:
```
jberchtold-nvidia: Thanks for the detailed error checks!

transformer_engine/jax/cpp_extensions/attention.py
vcherepanov-nv: These are just arbitrary numbers, really. In fact, assigning UIDs is completely optional, cuDNN can auto-assign. UIDs are added here just for determinism / to make future troubleshooting easier, e.g. so that we know that 4 is the output tensor.

transformer_engine/jax/cpp_extensions/attention.py
vcherepanov-nv: Source/editable installs are covered by the repo-relative 3rdparty/cudnn-frontend/python path. Normal/system installs work if the cudnn Python package is installed on sys.path; otherwise the code raises an explicit import error.

transformer_engine/jax/csrc/extensions/attention.cpp
vcherepanov-nv: We cannot carry cuDNN graph directly via FFI, right? The Python trace-time code registers the graph and passes a graph_id; the C++ FFI handler later uses that ID to find the graph and UID metadata when XLA executes. Moving the registry to Python would still require a global lookup and more GIL / module-lifetime coupling.

transformer_engine/jax/csrc/extensions/attention.cpp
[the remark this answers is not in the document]
vcherepanov-nv: Not sure if I'd call it a leak, but yes, currently the cache is process-lifetime. If we ever encounter an issue with it's growth, then we'll need to implement some kind of eviction policy. But it is out of scope of this PR.

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -754,55 +775,98 @@ cudnnHandle_t GetScoreModCudnnHandle() {
...
+  auto handle = GetScoreModCudnnHandle();
+  NVTE_CHECK_CUDNN(cudnnSetStream(handle, stream));
+
+  auto graph = std::make_shared<cudnn_frontend::graph::Graph>();
+  auto status = graph->deserialize(handle, serialized_data);
+  NVTE_CHECK(status.is_good(), "Failed to deserialize cuDNN score_mod SDPA graph: ",
+             status.get_message());
+
+  std::lock_guard<std::mutex> lock(ScoreModGraphCacheMutex());
+  auto &cache = ScoreModGraphCache();
```
jberchtold-nvidia: SR: Can we name `ScoreModGraphCache()` something like  `getScoreModeGraphCache()`? On my first read-thru I read `ScoreModeGraphCache` as a constructing a new object and thought this was always using a  fresh cache.

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -754,55 +775,98 @@ cudnnHandle_t GetScoreModCudnnHandle() {
...
+
+  auto handle = GetScoreModCudnnHandle();
+  NVTE_CHECK_CUDNN(cudnnSetStream(handle, stream));
+
+  auto graph = std::make_shared<cudnn_frontend::graph::Graph>();
+  auto status = graph->deserialize(handle, serialized_data);
+  NVTE_CHECK(status.is_good(), "Failed to deserialize cuDNN score_mod SDPA graph: ",
+             status.get_message());
+
+  std::lock_guard<std::mutex> lock(ScoreModGraphCacheMutex());
```
jberchtold-nvidia: Same here about `ScoreModGraphCacheMutex()` -> `getScoreModGraphCacheMutex()`

transformer_engine/jax/attention.py
```diff
@@ -1411,6 +1528,10 @@ def fused_attn(
     context_checkpoint_name: str = "context",
     softmax_offset: Optional[jnp.ndarray] = None,
     stripe_size: int | None = None,
+    score_mod: Optional[Callable] = None,
+    score_mod_bprop: Optional[Callable] = None,
+    score_mod_tensors: Optional[Mapping[str, Any]] = None,
+    score_mod_bprop_tensors: Optional[Mapping[str, Any]] = None,
 ):
```
KshitijLakhani, line 1471: Looks like this is the highest API that score_mod has been plumbed to.
There are higher APIs that would need to be plumbed to as well - please do take a look 
At the very least FusedDPA and DPA

transformer_engine/jax/attention.py
```diff
@@ -1391,10 +1391,127 @@ def _fused_attn_bwd_rule(
...
+        raise ValueError(f"{header} is mutually exclusive with context parallelism.")
+    if stripe_size is not None:
+        raise ValueError(f"{header} is mutually exclusive with striped context parallelism.")
+
+
 def fused_attn(
     qkv: Tuple[jnp.ndarray, ...],
     bias: Optional[jnp.ndarray],
-    sequence_descriptor: SequenceDescriptor,
+    sequence_descriptor: Optional[SequenceDescriptor],
```
KshitijLakhani, line 1450: If you are making sequence_descriptor Optional then please add a check in the function body to ensure that SequenceDescriptor is passed when the score_mod / flex attn is not being used

Users should not use non-flex attn / non-score-mod attn without `sequence_descriptor`

transformer_engine/jax/attention.py
```diff
@@ -1485,6 +1620,48 @@ def fused_attn(
                              AttnBiasType.NO_BIAS, AttnMaskType.PADDING_CAUSAL_MASK,
                              QKVLayout.T3HD, 0.125, 0, True, 3)
     """
+    if score_mod is None:
+        if score_mod_bprop is not None:
+            raise ValueError("score_mod_bprop requires score_mod to be provided.")
+        if score_mod_tensors is not None:
+            raise ValueError("score_mod_tensors requires score_mod to be provided.")
+        if score_mod_bprop_tensors is not None:
+            raise ValueError("score_mod_bprop_tensors requires score_mod to be provided.")
```
KshitijLakhani: While being descriptive with the asserts is great in general, I am wondering if this overkill ?
The essence seems to be that we error out when `score_mod is None` irrespective of values of other params, so why not consolidate these three ?
And if `score_mod is not None` we use that as a flag that the user wants to use flex attn right ?

transformer_engine/jax/attention.py
KshitijLakhani: Never mind, I think we have a check that I had forgotten about: 
`if sequence_descriptor is None or isinstance(sequence_descriptor, jnp.ndarray):`

Do confirm please

transformer_engine/jax/attention.py
```diff
@@ -1391,10 +1391,127 @@ def _fused_attn_bwd_rule(
...
+    softmax_type: AttnSoftmaxType,
+    dropout_probability: float,
+    max_segments_per_seq: int,
+    window_size: Optional[Tuple[int, int]],
+    context_parallel_strategy: CPStrategy,
+    context_parallel_causal_load_balanced: bool,
+    context_parallel_axis: str,
+    softmax_offset: Optional[jnp.ndarray],
+    stripe_size: int | None,
+):
```
KshitijLakhani: I am wondering if there's merit in moving this check one step downstream to cpp_extensions similar to how there are checks for `fused_attn` fwd in `fused_attn_fwd()`

Note: There's merit in having a checking function like this and I think that would be a good addition for `fused_attn` as well but outside the scope of this PR

transformer_engine/jax/attention.py
```diff
@@ -1391,10 +1391,127 @@ def _fused_attn_bwd_rule(
...
+        config,
+    )
+    return (
+        grad_qkv,
+        tuple(None for _ in score_mod_tensors),
+        tuple(None for _ in score_mod_bprop_tensors),
+    )
+
+
+_fused_attn_score_mod.defvjp(_fused_attn_score_mod_fwd_rule, _fused_attn_score_mod_bwd_rule)
```
KshitijLakhani, line 1444: Initially my thought was if we could combine fused_attn_score_mode* functions with the pre-existing fused_attn functions but as the score_mod functions are anyways not being coupled with a lot of the other features such as determinism, dropout, bias, cp, etc, I guess it is okay to keep it as a separate API

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -689,5 +702,242 @@ XLA_FFI_DEFINE_HANDLER_SYMBOL(FusedAttnBackwardHandler, FusedAttnBackwardFFI,
...
+           frontend_version == other.frontend_version;
+  }
+};
+
+struct ScoreModGraphCacheKeyHash {
+  size_t operator()(const ScoreModGraphCacheKey &key) const {
+    size_t seed = std::hash<int>{}(key.device_id);
+    auto combine = [&seed](int64_t value) {
+      seed ^= std::hash<int64_t>{}(value) + 0x9e3779b97f4a7c15ULL + (seed << 6) + (seed >> 2);
+    };
```
KshitijLakhani, line 730: nit: Comment explaining choice of magic constant would be good

transformer_engine/jax/cpp_extensions/attention.py
KshitijLakhani: Could you please add comments explaining this reason in the file ?
Thanks

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -267,6 +280,809 @@ def check_seed(self, seed, dropout_probability, is_training):
         return seed
 
 
+@dataclass(frozen=True)
```
KshitijLakhani: I would recommend to move this newly added piece of into a separate file for flex attention.
As part of a larger refactor in the near future I will anyways be moving out the CP classes to a new CP only file and `attention.py` would be fused attn only so moving flex attn to it's own file right would be a good way to not accumulate more technical debt.

Also, it would be good not have those may single functions if possible but instead have them combined into a one/two helper classes like the other fused attn classes in this `attention.py` file

When moving to flex attention file, do be mindful to ideally not have circular dependencies with `attention.py` - if there's any happy to discuss them

tests/jax/test_distributed_fused_attn.py
```diff
@@ -272,6 +284,114 @@ def test_cross_attn(
         runner.test_backward()
 
 
+DISTRIBUTED_SCORE_MOD_DATA_SHAPES = {
+    "L0": [],
+    "L1": [(4, 16, 4, 64)],
+    "L2": [(4, 16, 4, 64)],
+}
```
KshitijLakhani, line 286: Why are we running tests with the same shape ? 
If only 1 shape is needed, only L1 should suffice - no need for L2 then

tests/jax/test_distributed_fused_attn.py
```diff
@@ -272,6 +284,114 @@ def test_cross_attn(
...
+        self,
+        device_count,
+        mesh_shape,
+        mesh_axes,
+        mesh_resource,
+        data_shape,
+        dtype,
+    ):
+        _require_cudnn_frontend_score_mod()
+        batch, seqlen, num_heads, head_dim = data_shape
```
KshitijLakhani, line 304: There seems to be a lot of input setup here,
Can we not use `_setup_inputs()` with minor modifications ?
You could also use `_setup_inputs()` and "undo" some of the setup in the test if you do not wish to modify `setup_inputs()` - that is fine too.

All tests in attention end up using it so would make sense to maintain uniformity if possible.
However, if it involves too much branching and changes for flex attn, I understand keeping it separate.

Additionally all tests end up using the FusedAttnRunner() - are you not able to use it for the flex tests with minor changes ? Using the runner also means that you do not need to import fused_attn in the distributed tests.
The current setup only imports the fused_attn in the test_fused_attn.py tests the the distributed tests setup and use it directly from there

The above two might help reduce duplication of code and maintain uniformity and the test inputs generated. Moreover, we would not want to add setup, breakdown for any new attn types we add, especially when we are using the same fused_attn API for all.

tests/jax/test_distributed_fused_attn.py
```diff
@@ -25,6 +36,7 @@
     inverse_reorder_causal_load_balancing,
     CPStrategy,
     ReorderStrategy,
+    fused_attn,
```
KshitijLakhani: Please see below comment
Would be good to avoid this import here if possible 
The CP tests also do not end up importing it but use it via the runner customcall_fused_dpa

KshitijLakhani: A lot fo the other imports like assert_allclose acan also be prevented by using the FusedAttnRunner nad setup_inputs in the fused_attn tests

tests/jax/test_fused_attn.py
```diff
@@ -54,6 +55,157 @@
...
+    if post_scale_bias:
+        q_pos = jnp.arange(query.shape[1], dtype=jnp.float32)[:, None]
+        kv_pos = jnp.arange(key.shape[1], dtype=jnp.float32)[None, :]
+        scores = scores + q_pos - kv_pos
+    if softcap is not None:
+        scores = softcap * jnp.tanh(scores / softcap)
+    probs = jax.nn.softmax(scores, axis=-1)
+    return jnp.einsum("bhqk,bkhd->bqhd", probs, value).astype(query.dtype)
+
+
```
KshitijLakhani: WHy create your own reference and not use the jax native reference in the test file already ?

tests/jax/test_fused_attn.py
```diff
@@ -54,6 +55,157 @@
 _deterministic = not bool(int(os.getenv("NVTE_ALLOW_NONDETERMINISTIC_ALGO", "1")))
 
 
+def _has_cudnn_frontend_python():
```
KshitijLakhani: This is a general comment for the newly added items to conform to the existing test infrastructure and not add completely "new" tests:
1. Have you considered reusing the `setup_inputs()` rather than each test regenerating the inputs themselves ? If not,please do
2. Have you consider using the `FusedAttnRunner` setup to maintain uniformity across tests ? If not, please consider 

If for whatever reason, you are unable to conform to the existing test infrastructure in fused attn for the flex attn tests, please move all the flex attn items a different test file

KshitijLakhani (changes requested):

transformer_engine/jax/attention.py
```diff
@@ -1411,6 +1528,10 @@ def fused_attn(
     context_checkpoint_name: str = "context",
     softmax_offset: Optional[jnp.ndarray] = None,
     stripe_size: int | None = None,
+    score_mod: Optional[Callable] = None,
+    score_mod_bprop: Optional[Callable] = None,
```
jberchtold-nvidia, line 1468: Instead of a separate `score_mod` and `score_mod_bprop` definitions would it be fitting to JAX-style to only provide a `score_mod` function and we instead use JAX's VJP rules for determining the backprop. This would allow the backprop to be inferred automatically for most JAX operations. And if a user wanted a custom backprop they could define a custom VJP rule, like we do here, to package their forward and backward operations together in the typical JAX way.

https://github.com/NVIDIA/TransformerEngine/blob/aab7bc947bb09a72a9b068f3205c28dcba6e7064/transformer_engine/jax/dense.py#L239

What do you think? @KshitijLakhani @vcherepanov-nv

vcherepanov-nv: That would be neat, but the missing piece is jax-to-cuDNN translation. I actually did some experiments with it awhile back, but eventually decided against. There will always be some corner (or even not so corner) cases, where perfectly valid jax cannot be expressed in cuDNN ops graph, which would be quite confusing to the user.
We can think and revisit it later, I guess. But for now - I just want to enable the most direct way to use cuDNN API.

jberchtold-nvidia: Got it, sounds good. Thanks for the explanation!

KshitijLakhani: I agree with Vlad's approach based on his experiments

jberchtold-nvidia (dismissed):

transformer_engine/jax/flax/transformer.py
```diff
@@ -760,25 +796,30 @@ def __call__(
...
-            seqlen_q,
-            seqlen_kv,
-            head_dim_qk,
-            head_dim_v,
-            self.window_size,
-        )
+        if score_mod_requested:
+            if not enable_fused_attn:
+                raise ValueError("score_mod requires fused attention, but NVTE_FUSED_ATTN=0.")
+            has_fused_attn_kernel = True
```
KshitijLakhani: Why do we force this to `True` ?
If user wants to perform flex attn and has enabled fused attn, then shouldn't we check via `is_fused_attn_kernel_available()` right ?
This seems to be our contract API to determine whether fused is available+requested so I think we should rely on that API rather than force it to True here

transformer_engine/jax/flax/transformer.py
```diff
@@ -715,6 +745,12 @@ def __call__(
         qkv_layout = QKVLayout[self.qkv_layout.upper()]
         softmax_type = AttnSoftmaxType.from_str(self.softmax_type)
         del self.attn_bias_type, self.attn_mask_type, self.qkv_layout
+        score_mod_requested = (
+            self.score_mod is not None
+            or self.score_mod_bprop is not None
+            or score_mod_tensors is not None
+            or score_mod_bprop_tensors is not None
+        )
```
KshitijLakhani, line 749: Why repeat this check in DPA and _FusedDPA ?
The expectation is that _FusedDPA is an internal class (with the leading underscore indicating the same) and so if the checks exist in DPA I donot think we need them again in _FusedDPA. Thoughts ?
The user should be exposed to DPA only IIRC

transformer_engine/jax/attention.py
KshitijLakhani: Left some comments for this

tests/jax/test_fused_attn.py
KshitijLakhani: Just to clarify, I believe the idea behind this comment was to check the possibility of code reuse. However, it seems like the solutions chosen is to move the contents of the flex attention tests to a different file altogether.

I do not think it is a good practice for us to have different ways of creating the reference for different types of fused attn. It would have been best to use the reference implementation in `test_fused_attn.py` tweaked for the test case we have fro flex attention. 

However, I'll try to not hold the PR for this but I definitely think this should be looked into a follow up PR
cc: @cyanguwa

tests/jax/test_distributed_fused_attn.py
```diff
@@ -272,6 +285,136 @@ def test_cross_attn(
...
+            mesh_axes=mesh_axes,
+            mesh_resource=mesh_resource,
+        )
+        runner._setup_inputs()
+
+        qkv_sharding = NamedSharding(runner.mesh, PartitionSpec(dp_axis, None, tp_axis, None))
+        query = (0.125 * runner.q).astype(dtype)
+        key_tensor = (0.125 * runner.k).astype(dtype)
+        value = (0.125 * runner.v).astype(dtype)
+        doutput = random.normal(random.PRNGKey(2025), data_shape, dtype=dtype)
```
KshitijLakhani: Why do we do this ? 0.125*

tests/jax/test_distributed_fused_attn.py
```diff
@@ -272,6 +285,136 @@ def test_cross_attn(
...
+            None,
+            None,
+            SeqDescFormat.Mask,
+            number_of_devices=device_count,
+            mesh_shape=mesh_shape,
+            mesh_axes=mesh_axes,
+            mesh_resource=mesh_resource,
+        )
+        runner._setup_inputs()
+
```
KshitijLakhani: So it seems like you are using the runner to only setup the inputs but then are following that up with "duplicate" code that `test_forward` and `test_backward` in `test_fused_attn.py`
The suggestions was to try and use the runner to call forward(), which does the setup using the runner and also runs the test. 

The idea is if you can integrate the non distributed tests with the Runner infrastrucutre then the distributed tests can directly use it here for free. The approach here seems to be somewhere in between your older approach and the suggested approach. You can refer to other tests in this file for reference, incase I've not done a good job explaining

i'm curiosu to know if the reason for that was because you were unable to fully integrate the score mod tests into the Fused Attn runner ? Because it seems like it is the same reason for creating a separate test_fused_attn_score_mod.py as compared to integrating the flex attn tests in the Fused attn runner in test_fused_attn.py

If test_fused_attn_score_mod must be created then a Runner should be ceated in the too and the distributed flex attn tests can then use that runner in here (similar to how other tests do)

tests/jax/test_distributed_fused_attn.py
```diff
@@ -272,6 +285,136 @@ def test_cross_attn(
...
+            out_shardings=((None, qkv_sharding), (qkv_sharding, qkv_sharding, qkv_sharding)),
+        )
+        jitted_ref = jax.jit(jax.value_and_grad(ref_loss, argnums=(0, 1, 2), has_aux=True))
+
+        sharded_args = (
+            jax.device_put(query, qkv_sharding),
+            jax.device_put(key_tensor, qkv_sharding),
+            jax.device_put(value, qkv_sharding),
+            jax.device_put(doutput, qkv_sharding),
+        )
```
KshitijLakhani: All of this can come for free if the flex attn is integrated with the FusedAttn Runner

tests/jax/test_fused_attn.py
```diff
@@ -138,6 +138,30 @@ def general_dot_product_attention(
...
+    num_groups = h_q // h_kv
+    grouped_query = jnp.reshape(query, (b, s_q, h_kv, num_groups, d))
+    logits = scale_factor * jnp.einsum("...qhgd,...khd->...hgqk", grouped_query, key)
+    logits = score_mod_reference(logits.astype(jnp.float32))
+    softmax_out = jax.nn.softmax(logits, axis=-1).astype(input_dtype)
+    context = jnp.einsum("...hgqk,...khd->...qhgd", softmax_out, value)
+    context_shape = query.shape[:-1] + (value.shape[-1],)
+    return jnp.reshape(context, context_shape)
+
+
```
KshitijLakhani: The original comment here: https://github.com/NVIDIA/TransformerEngine/pull/2985#discussion_r3275717169 and the follow up comment here: https://github.com/NVIDIA/TransformerEngine/pull/2985#discussion_r3314012083 was to suggest that we do not create a new reference but instead use the available reference. 

This new function defeats the purpose and does not do so. This is just duplicating `general_dot_product_attention()` with an extra score_mod_reference param right ? so why not just `general_dot_product_attention` and pass a a score_mod param to it instead ?

tests/jax/test_fused_attn.py
```diff
@@ -267,6 +291,30 @@ def jax_dpa(query, key, value, bias, softmax_offset, mask, dropout_rng, **kwargs
...
+            or dropout_rng is not None
+        ):
+            raise ValueError(
+                "score_mod reference path expects no bias, mask, softmax offset, or dropout."
+            )
+        if kwargs["qkv_layout"] != QKVLayout.BSHD_BSHD_BSHD:
+            raise ValueError(
+                "score_mod reference path expects separate BSHD query/key/value tensors."
+            )
+        return score_mod_reference_dot_product_attention(
```
KshitijLakhani: Why not use `general_dot_product_attention` instead with a `score_mod` arg passed to it so that the reference call used is the same ?

tests/jax/test_fused_attn.py
```diff
@@ -516,6 +575,22 @@ def _setup_inputs(self):
         self.q = jax.random.uniform(q_key, q_shape, self.dtype, -1.0)
         self.k = jax.random.uniform(k_key, k_shape, self.dtype, -1.0)
         self.v = jax.random.uniform(v_key, v_shape, self.dtype, -1.0)
+        if self.input_scale != 1.0:
+            self.q = (self.input_scale * self.q).astype(self.dtype)
+            self.k = (self.input_scale * self.k).astype(self.dtype)
+            self.v = (self.input_scale * self.v).astype(self.dtype)
```
KshitijLakhani: Can this not collapse to `self.q = self.input_scale * jax.random.uniform(q_key, q_shape, self.dtype, -1.0)` ? - thereby reducing the bloat ?
Or is this intentional for some reason?

tests/jax/test_fused_attn.py
```diff
@@ -516,6 +575,22 @@ def _setup_inputs(self):
...
+        else:
+            output_shape = (
+                self.batch_size,
+                self.max_seqlen_q,
+                self.num_heads_q,
+                self.head_dim_v,
+            )
+            self.doutput = jax.random.normal(
+                jax.random.PRNGKey(self.doutput_seed), output_shape, dtype=self.dtype
+            )
```
KshitijLakhani: I would consider moving it to post_init
Reason:
In this class function, `self.doutput` is being defined and then is used in a different class function
IF someone were to incorrectly call that class function prior to this one, they will see failures arising from incorrect init
If we need this var `self.doutput` to be a class var as you need it to be persistent across class member functions, I'd suggest to initialize this correctly in post_init instead

tests/jax/test_fused_attn.py
```diff
@@ -810,7 +885,18 @@ def test_forward(self):
         """
         self._setup_inputs()
 
-        args = [self.q, self.k, self.v, self.bias, self.softmax_offset, self.mask, self.dropout_rng]
+        reference_sequence_descriptor = (
```
KshitijLakhani: nit: change to reference_mask to make the intent cleare

tests/jax/test_fused_attn_score_mod.py
```diff
@@ -0,0 +1,740 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
KshitijLakhani, line 1: Why do test_fused_attn_score_mod_post_scale_bias_optional_bprop and  test_fused_attn_score_mod_causal_with_bprop not use the Runner insfrastructure?

KshitijLakhani: Doc strings / concise comments for tests would be great !

tests/jax/test_fused_attn_score_mod.py
```diff
@@ -0,0 +1,740 @@
...
+import pytest
+from jax import value_and_grad
+
+import transformer_engine.jax.cpp_extensions.flex_attention as tex_attention
+from transformer_engine.jax.attention import (
+    AttnBiasType,
+    AttnMaskType,
+    AttnSoftmaxType,
+    QKVLayout,
+    fused_attn,
```
KshitijLakhani: Fused Attn Runner (if used properly) takes care of calling fused_attn . Seems like bewlo tests need fused_attn:

  1. test_fused_attn_score_mod_validation_rejects_masks_without_cudnn_frontend
     This is a validation/unit test. Direct fused_attn call is fine here because it specifically checks
     argument rejection.

  2. test_fused_attn_score_mod_post_scale_bias_optional_bprop
     This should be converted to runner use.

  3. test_fused_attn_score_mod_causal_with_bprop
     This should be converted to runner use.

If we think 1. makes sense in test__fused_attn or is not needed then we can ensure no imports of fused_attn here and only FusedAttNRUnner

tests/jax/test_fused_attn_score_mod.py
```diff
@@ -0,0 +1,740 @@
...
+    )
+
+
+def _stable_score_mod_input(key, shape, dtype):
+    """Create low-magnitude inputs for fp16/bf16 score_mod reference comparisons."""
+    input_scale = 1.0 / sqrt(shape[-1])
+    return (input_scale * jax.random.normal(key, shape, dtype=dtype)).astype(dtype)
+
+
+class ScoreModFusedAttnRunner(FusedAttnRunner):
```
KshitijLakhani, line 181: Can we not make ScoreModFusedAttnRunner generic and then have it be used byt the 
  - softcap
  - causal
  - post-scale-bias

Right now ScoreModFusedAttnRunner is very softcap specific r

KshitijLakhani: Subsequently, how about moving require_cudnn_frontend to ScoreModFusedAttnRunner class ?

I'd also consider moving these to this class once generic: 
  - Score_mod fixed defaults:
    NO_BIAS, NO_MASK, VANILLA_SOFTMAX, BSHD_BSHD_BSHD, dropout=0.0, SeqDescFormat.Mask,
    cp_load_balanced=False

  - Input scaling logic: input_scale = 1.0 / sqrt(head_dim)
  - Default score_mod backward settings: doutput_seed, rtol, atol
  - Convenience constructors:
    softcap(...), causal(...), post_scale_bias(...)

tests/jax/test_fused_attn_score_mod.py
```diff
@@ -234,30 +250,128 @@ def softcap(
...
+            number_of_devices=number_of_devices,
+            mesh_shape=mesh_shape,
+            mesh_axes=mesh_axes,
+            cp_load_balanced=cls.CP_LOAD_BALANCED,
+            score_mod=score_mod,
+            score_mod_bprop=score_mod_bprop,
+            score_mod_tensors=score_mod_tensors,
+            score_mod_bprop_tensors=score_mod_bprop_tensors,
+            score_mod_reference=score_mod_reference,
+            input_scale=cls.input_scale(head_dim),
```
KshitijLakhani: suggestion: rename the input_scale function to something like `_compute_input_scale()` to avoid name collision between the datclass field and method

tests/jax/test_fused_attn.py
```diff
@@ -306,28 +286,27 @@ def jax_dpa(query, key, value, bias, softmax_offset, mask, dropout_rng, **kwargs
...
-            jnp.float32,
-        ).astype(query.dtype)
-
+        deterministic = not kwargs.get("is_training", False)
+        dropout_probability = kwargs.get("dropout_probability", 0.0)
+        softmax_type = kwargs.get("softmax_type", AttnSoftmaxType.VANILLA_SOFTMAX)
+    else:
+        deterministic = not kwargs["is_training"]
+        dropout_probability = kwargs["dropout_probability"]
+        softmax_type = kwargs["softmax_type"]
```
KshitijLakhani: Maybe I'm wrong but: 
Why the discrepancy - The score-mod branch uses `.get()` with fallbacks; the normal branch requires the keys
A score-mod call that omits `is_training` silently becomes `deterministic=True` (i.e. no backward path) rather than raising. For a test reference that's a quiet way to mask a bug. Either require the same keys in both branches, or add a comment explaining why the defaults are safe.

tests/jax/test_fused_attn_score_mod.py
```diff
@@ -181,51 +178,70 @@ def score_mod(scores):
...
+    QKV_LAYOUT = QKVLayout.BSHD_BSHD_BSHD
+    BIAS_SHAPE = None
+    WINDOW_SIZE = None
+    SEQ_DESC_FORMAT = SeqDescFormat.Mask
+    CP_LOAD_BALANCED = False
+    DOUTPUT_SEED = None
+    RTOL = 5e-2
+    ATOL = 5e-2
+
+    @staticmethod
```
KshitijLakhani, line 198: suggestion (worth a try): 
softcap, post_scale_bias, and causal each repeat the identical 5-arg signature plus the four
  number_of_devices/mesh_shape/mesh_axes/mesh_resource passthrough lines. You could forward **kwargs to generic and cut ~30 lines. Tradeoff is
  explicitness vs. brevity — fine to leave, but worth a mention.

tests/jax/test_fused_attn_score_mod.py
```diff
@@ -181,51 +178,70 @@ def score_mod(scores):
...
+    def require_cudnn_frontend():
```
KshitijLakhani, line 199: Quick question: The tests use a decorator for the cudnn version and skip if conditions nto matched and then we have a require_cudnn_frontend here - is this a redundancy in here that we might be missing?

tests/jax/test_fused_attn.py
```diff
@@ -106,6 +107,9 @@ def general_dot_product_attention(
             mask = jnp.expand_dims(mask, axis=-3)
         logits = jnp.where(mask, jnp.finfo(dtype).min, logits)
 
+    if score_mod_reference is not None:
+        logits = score_mod_reference(logits.astype(jnp.float32))
+
```
KshitijLakhani, line 113: nit: score_mod is applied right after the mask. It's correct for the current tests (score_mod always runs with NO_MASK, and fused_attn rejects mask+score_mod), but the reference now technically supports a combination the kernel forbids. A one-line comment noting that intent would save a future reader confusion. 
Also logits.astype(jnp.float32) is a no-op here since jax_dpa always passes dtype=jnp.float32 — harmless, but worth knowing it's defensive only.

KshitijLakhani (dismissed):

tests/jax/test_fused_attn_score_mod.py
vcherepanov-nv: I kept these wrappers explicit intentionally. The repeated part is the public test-helper signature, while generic() owns the actual runner setup. Keeping number_of_devices, mesh_shape, mesh_axes, and mesh_resource visible in each convenience constructor makes the distributed test usage easier to read without jumping through **kwargs. I am inclined to leave this explicit unless you feel strongly about the shorter form.

tests/jax/test_fused_attn_score_mod.py
vcherepanov-nv: They are related but not identical. The skipif decorator only checks whether the cuDNN frontend Python package can be imported, so pytest can skip these tests early when the package is missing. require_cudnn_frontend() also checks that the frontend version supports score_mod SDPA (>= 9.6) before running the integration path, and the same guard is reused by the distributed score_mod test. So this is partly redundant on import availability, but not on the version requirement.

tests/jax/test_fused_attn_score_mod.py
KshitijLakhani: I'm okay with leaving as is

KshitijLakhani (approved):
LGTM !
Thanks. 
Approved for merge when CI passes

PR #2991, opened 2026-05-14
Improve TE Group MLP CPU Overhead 

zhongbozhu:
# Description

NOTE: this PR will be closed and the refactored PR is at: https://github.com/NVIDIA/TransformerEngine/pull/3075

Improve TE grouped mlp CPU overhead, suppose cuda graph is not enabled. 

This is for issue: https://github.com/NVIDIA/TransformerEngine/issues/2897 

E2E Model: Qwen3.5 35B-A3B, nano scale model more prone to CPU overhead
What we measure: suppose we didn't turn on cuda graph, so that `split_sizes` live on CPU, we measure the CPU side time between the end of H2D copy of `split_sizes` , to the launch of grouped quantize kernel. 

before: 355us
image

after: 167us
image

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
+++ b/transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
@@ -159,10 +159,19 @@ class ForwardGroupedMLP_CuTeGEMMSwiGLU_MXFP8(FusedOperation):
         split_sizes = fc1_split_sizes
         if int(split_sizes.numel()) != num_groups:
             raise ValueError(f"Expected {num_groups} splits, but got {int(split_sizes.numel())}.")
-        split_sizes = split_sizes.to(dtype=torch.int64, device=device)
-        base_split_offsets = tex.splits_to_offsets(split_sizes, 1)
-        split_points = base_split_offsets[1:].to(dtype=torch.int)
-        fc2_x_tensor_offsets = base_split_offsets * fc2_weight_shape[1]
+        # Prepare all split metadata in one CUDA kernel.  The returned split_sizes is the
+        # canonical TE representation: int64[num_groups].  Python uses it from here
+        # onward for grouped quantization and backward state.
+        #
+        # base_split_offsets: int64[num_groups + 1], [0, cumsum(split_sizes)]
+        # split_points: int32[num_groups], cumsum(split_sizes) without the leading 0
+        # fc2_x_tensor_offsets: int64[num_groups + 1], base_split_offsets * fc2 K
+        (
+            split_sizes,
+            base_split_offsets,
+            split_points,
+            fc2_x_tensor_offsets,
+        ) = tex.prepare_grouped_splits(split_sizes, num_groups, fc2_weight_shape[1])
 
         # Extract post-scales from extra input
         scales = basic_op_extra_inputs[1][0]
--- a/benchmarks/linear/benchmark_graph_safe_grouped_linear.py
+++ b/benchmarks/linear/benchmark_graph_safe_grouped_linear.py
@@ -1,0 +1,376 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""Benchmark MXFP8 graph-safe grouped MLP.
+
+This mirrors ``benchmark_grouped_linear.py`` but targets the graph-safe TE ops
+path used by grouped MLP:
+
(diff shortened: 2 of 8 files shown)


transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -159,10 +159,19 @@ def fuser_forward(
         split_sizes = fc1_split_sizes
         if int(split_sizes.numel()) != num_groups:
             raise ValueError(f"Expected {num_groups} splits, but got {int(split_sizes.numel())}.")
-        split_sizes = split_sizes.to(dtype=torch.int64, device=device)
```
zhongbozhu, line 162: one issue found in E2E training, it's possible that we are doing a H2D transfer suppose split_sizes is not on device already, needs to patch it in the `misc.cpp` to handle this case as well, also add this to the unit test

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -159,10 +159,19 @@ def fuser_forward(
...
+        #
+        # base_split_offsets: int64[num_groups + 1], [0, cumsum(split_sizes)]
+        # split_points: int32[num_groups], cumsum(split_sizes) without the leading 0
+        # fc2_x_tensor_offsets: int64[num_groups + 1], base_split_offsets * fc2 K
+        (
+            split_sizes,
+            base_split_offsets,
+            split_points,
+            fc2_x_tensor_offsets,
+        ) = tex.prepare_grouped_splits(split_sizes, num_groups, fc2_weight_shape[1])
```
timmoon10, line 174: I'm not a fan of this name because it seems like it's a general purpose function for grouped tensors, but actually it's a very ad hoc thing specific to grouped MLP. We could give it a more descriptive name, but that would become very long and unwieldy. I'll take a look and see if we can come up with a more logical interface.

As a general practice, we should pick function names that accurately describe what they do or what contracts they expect. If the names become unreasonably long, that's a sign that the interface is not well designed. Basically, it's a tax on hackiness.

vthumbe1503, line 174: I think what we really want is the following

 --> Make sure user passes int64 splits so that we dont have to D2D copy to get it to int64
--> Make CUDNN accept int64 split_points instead of int32 split points
--> fc2_tensor_offsets is technically not even needed. Its not used by both CUTE dsl GGEMM or CUBLAS GGEMM. We should remove it.

zhongbozhu: I agree, do we have a tracker for fixing these issues?

zhongbozhu: About "Make sure user passes int64 splits so that we dont have to D2D copy to get it to int64", this one is okay, you can the D2D copy is part of the fused kernel that takes care of of the converting to int64 regardless, so there is no overhead here. 

"Make CUDNN accept int64 split_points instead of int32 split points", yes, this part of the code is ugly and I hate it. 

"fc2_tensor_offsets is technically not even needed." I agree, but it's okay to have it, because I already used bulk allocate & a single fused kernel for it, so if it's not needed, you can simply ignore it in python layer.

vthumbe1503: I was essentially thinking if we can avoid having these fusion APIs that are so specialised that they might not be used anywhere else. I know that your PR solves all of the issues above but I was hoping for a cleaner way to do this.

With the above 3 points that i mentioned we wont need a special kernel and no overhead would be there. 1 and 3 are trivial. But as you pointed trickiest one is point2

CUDNN should not only accept int32 split offsets, but its size should be one lesser than our offsets. Fixing point 2 might take a while, so we might want a immediate solution, but i would want to keep the API as simple as possible.

As I said, since fc2_x_tensor_offsets is not needed, API can be simplified
`tex.prepare_grouped_splits(split_sizes, num_groups)`
and we can remove fc2_x_tensor_offsets from the flow completely.

The only case I see where fc2_x_tensor_offsets can be useful --> is if group_quantize can use the offsets. Right now we are building the offsets from splits in group_quantize every time and we can reuse this there. but this is not done today

timmoon10:
https://github.com/NVIDIA/TransformerEngine/pull/3075 added a fused kernel for converting split sizes to offsets.

PR #2994, opened 2026-05-14
Plumb FP8+THD 

sudhakarsingh27:
## Summary

Adds FP8 support to cuDNN fused attention for packed THD inputs, including context-parallel execution.

## Final support matrix

| FP8 recipe (cuDNN 9.23+) | H100 / SM90, non-CP | H100 / SM90, CP | B200 / SM100, non-CP | B200 / SM100, CP |
|---|---|---|---|---|
| `DelayedScaling` | Forward only | Forward only (not directly covered by PR tests) | Forward and backward | Forward and backward (P2P, all-gather, A2A, A2A+P2P) |
| `Float8CurrentScaling` | Not supported | Not supported | Forward and backward | Forward and backward (P2P, all-gather, A2A, A2A+P2P) |
| `MXFP8BlockScaling` | Not supported with THD | Not supported with THD | Not supported with THD | Not supported with THD |

## Changes

- Plumbs THD sequence metadata and ragged offsets through FP8 forward and backward.
- Supports delayed and current FP8 scaling while keeping MXFP8+THD disabled where backend support is unavailable.
- Enables FP8+THD across the existing P2P, all-gather, and A2A context-parallel paths.
- Handles bottom-right causal masking, padded-token cleanup, and output and gradient initialization for packed buffers.
- Adds focused coverage for reference parity, context parallelism, padding cleanup, CUDA graphs, and FP8 recipe selection.

## Validation

- FP8+THD forward and backward with combined and separate QKV layouts.
- Context-parallel FP8+THD canaries.
- THD padding-mask and CUDA graph tests.
- Backend-selection coverage for delayed scaling, current scaling, and MXFP8.


tests/pytorch/attention/test_attention.py
```diff
@@ -2526,6 +2528,8 @@ def get_dummy_cuda_rng_tracker() -> CudaRNGStatesTracker:
             attn_mask_type=config.attn_mask_type,
             checkpoint_core_attention=False,
             core_attention_bias_type=config.attn_bias_type,
+            fp8_output=fp8_dpa,
+            fast_zero_fill=False,
```
cyanguwa: cuDNN doesn't touch the pad tokens (between seqs or at the end of the batch) so we had to zero out the entire output for F16 THD (see [here](https://github.com/NVIDIA/TransformerEngine/blob/21ba49c59f35947a617874e5274f597d572755f9/transformer_engine/pytorch/csrc/extensions/attention.cpp#L164)). I wonder if we need to do the same for FP8?

tests/pytorch/attention/test_attention.py
```diff
@@ -2512,6 +2512,8 @@ def get_dummy_cuda_rng_tracker() -> CudaRNGStatesTracker:
     out_grad_shape_new = [*out_grad_shape[:-2], out_grad_shape[-2] * out_grad_shape[-1]]
     out_grad = torch.randn(out_grad_shape_new, dtype=dtype, device="cuda")
 
+    print(f"qkv_format: {qkv_format}")
+    print(f"inp shape: {inp[0].shape}, {inp[1].shape}, {inp[2].shape}")
```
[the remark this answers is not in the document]
cyanguwa: We can probably remove these debug prints once everything is verified to be fine?

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -281,12 +281,15 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
...
+      // 9.23+ sm100+: {thd} (ragged/packed variable-length)
       ((cudnn_runtime_version < 92100 &&
         (qkv_format == NVTE_QKV_Format::NVTE_BSHD || qkv_format == NVTE_QKV_Format::NVTE_SBHD) &&
         softmax_type == NVTE_Softmax_Type::NVTE_VANILLA_SOFTMAX) ||
        (cudnn_runtime_version >= 92100 &&
         (qkv_format == NVTE_QKV_Format::NVTE_BSHD || qkv_format == NVTE_QKV_Format::NVTE_SBHD ||
-         qkv_format == NVTE_QKV_Format::NVTE_BHSD))) &&
+         qkv_format == NVTE_QKV_Format::NVTE_BHSD)) ||
+       (cudnn_runtime_version >= 92300 && sm_arch_ >= 100 &&
+        qkv_format == NVTE_QKV_Format::NVTE_THD && supported_ragged_offset_size)) &&
```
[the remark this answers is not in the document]
cyanguwa: I agree. We probably should add a mask constraint - we only accept THD with padding masks.

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -60,6 +62,27 @@ void fused_attn_fp8_fwd_impl(
   NVTE_CHECK(!is_mxfp8 || cudnn_runtime_version >= 92100,
              "MXFP8 fused attention requires cuDNN 9.21.0 or later!");
 
+  NVTE_QKV_Format q_format = nvte_get_q_format(qkv_layout);
+  NVTE_QKV_Format kv_format = nvte_get_kv_format(qkv_layout);
+  bool is_ragged_q = (q_format == NVTE_QKV_Format::NVTE_THD);
+  bool is_ragged_kv = (kv_format == NVTE_QKV_Format::NVTE_THD);
+  const int device_id = cuda::current_device();
+  const int sm_arch_ = cuda::sm_arch(device_id);
+  bool use_ragged_stats = is_ragged_q && cudnn_runtime_version >= 90600 && sm_arch_ != 120;
```
cyanguwa: I wonder if all the 90600 occurrences should be 92300, given that we only start to support FP8 THD from 9.23?

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -60,6 +62,27 @@ void fused_attn_fp8_fwd_impl(
...
+
+  NVTE_QKV_Layout_Group layout_group = nvte_get_qkv_layout_group(qkv_layout);
+  const DType ragged_offset_type = cudnn_runtime_version >= 90500 ? DType::kInt64 : DType::kInt32;
```
cyanguwa: Could you check with cuDNN see if FP8 THD is supported with both types, or just one, i.e. kInt64? Also, should the runtime version here really be 92300 (if both types were supported)?

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -60,6 +62,27 @@ void fused_attn_fp8_fwd_impl(
...
+
+  int64_t actual_b = b;
+  if ((is_ragged_q || is_ragged_kv) && cudnn_runtime_version >= 90600) {
```
cyanguwa: Same as above.

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -121,6 +144,11 @@ void fused_attn_fp8_fwd_impl(
                    std::shared_ptr<fe::graph::Tensor_attributes>,   // softmax_offset
                    std::shared_ptr<fe::graph::Tensor_attributes>,   // seq_q
                    std::shared_ptr<fe::graph::Tensor_attributes>,   // seq_kv
+                   std::shared_ptr<fe::graph::Tensor_attributes>,   // offset_q
+                   std::shared_ptr<fe::graph::Tensor_attributes>,   // offset_o
+                   std::shared_ptr<fe::graph::Tensor_attributes>,   // offset_k
+                   std::shared_ptr<fe::graph::Tensor_attributes>,   // offset_v
+                   std::shared_ptr<fe::graph::Tensor_attributes>,   // offset_stats
```
cyanguwa, line 176: Nit: do we want to follow the same order of listing as in F16, i.e. offset_q/k/v/o/stats?

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -367,22 +443,37 @@ void fused_attn_fp8_fwd_impl(
...
       return return_tuple;
     };
 
     auto [mha_graph, Q, K, V, descale_q, descale_k, descale_v, descale_s, scale_s, scale_o,
-          attn_scale, O, amax_s, amax_o, Stats, bias, softmax_offset, seq_q, seq_kv, dropout_seed,
-          dropout_offset] = get_graph(sdpa_fp8_fprop_cache, descriptor);
-
-    auto plan_workspace_size = mha_graph->get_workspace_size();
+          attn_scale, O, amax_s, amax_o, Stats, bias, softmax_offset, seq_q, seq_kv, offset_q,
+          offset_o, offset_k, offset_v, offset_stats, dropout_seed, dropout_offset] =
```
cyanguwa: Same comment about the ordering for q/k/v/o/stats.

cyanguwa (changes requested):
Overall looks good, but please address the few comments and pass the CI. Thanks for the PR!

transformer_engine/common/fused_attn/fused_attn_fp8.cu
```diff
@@ -427,10 +430,10 @@ void fused_attn_fp8_fwd_impl(
           is_softmax_offset ? std::make_tuple(softmax_offset) : std::make_tuple(nullptr);
       auto padding_tuple =
           is_padding ? std::make_tuple(seq_q, seq_kv) : std::make_tuple(nullptr, nullptr);
-      auto offset_qo_tuple =
-          is_ragged_q ? std::make_tuple(offset_q, offset_o) : std::make_tuple(nullptr, nullptr);
+      auto offset_q_tuple = is_ragged_q ? std::make_tuple(offset_q) : std::make_tuple(nullptr);
       auto offset_kv_tuple =
           is_ragged_kv ? std::make_tuple(offset_k, offset_v) : std::make_tuple(nullptr, nullptr);
+      auto offset_o_tuple = is_ragged_q ? std::make_tuple(offset_o) : std::make_tuple(nullptr);
```
cyanguwa, line 479: Ok, maybe I was making a bigger deal out of this than it is. It looks like we are following the q/k/v/o/stats order in the shared_ptr declaration but not in the tuple creation, in the F16 file. So feel free to revert this change, or just change the shared_ptr order if you like. Thanks.

bbuschkaemper:
@sudhakarsingh27 Are there any plans for cudnn thd sm90 fp8 support?

sudhakarsingh27:
@bbuschkaemper, are you looking for both fwd and bwd in fp8+thd? SM90 in cudnn only supports fp8+thd in fwd right now

bbuschkaemper:
> @bbuschkaemper, are you looking for both fwd and bwd in fp8+thd? SM90 in cudnn only supports fp8+thd in fwd right now

Yes, full fp8 training. It would be interesting to see if this is possible, although it would prob. be quite unstable I think.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -2923,22 +2939,11 @@ def backward(ctx, dout, *_args):
...
-            if cu_seqlens_kv_padded is not None:
-                kv_actual_t = cu_seqlens_kv_padded[-1]
-                for d_tensor in [dk, dv]:
-                    if isinstance(d_tensor, torch.Tensor) and d_tensor.shape[0] > 0:
-                        kv_pad_mask = (
-                            torch.arange(d_tensor.shape[0], device=d_tensor.device) >= kv_actual_t
-                        )
-                        d_tensor[kv_pad_mask] = 0
+        # FP8 partial-gradient reduction can write THD inter-sequence padding.
+        if ctx.qkv_format == "thd" and ctx.fp8:
```
sudhakarsingh27: previously, this was being done for non fp8 as well, why the switch to only fp8?

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -407,6 +407,8 @@ def test_cp_with_flash_attention(cp_pool, dtype, model, qkv_format, cp_comm_type
         2, 4096, 12, 128, attn_bias_type="post_scale_bias", bias_shape="bhss"
     ),  # MHA
     "cp_1_5": ModelConfig(2, 4096, 12, 128, attn_mask_type="causal", window_size=(512, 512)),  # MHA
+    # Noncausal MHA without bias/max-logit provides an FP8+THD+CP backend-compatible row.
+    "cp_1_6": ModelConfig(2, 4096, 12, 128),
```
sudhakarsingh27: is causal supported now in the newer commits? Should this be removed?

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -283,14 +283,15 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
             attn_mask_type == NVTE_Mask_Type::NVTE_CAUSAL_BOTTOM_RIGHT_MASK))) &&
       // pre-9.21: {bshd, sbhd}, {vanilla}
       // 9.21+: {bshd, sbhd, bhsd}, {vanilla, off-by-one, learnable}
-      // 9.23+ sm100+: {thd} (ragged/packed variable-length)
+      // 9.23+ sm100+: {thd}; 9.25+ sm90+: {thd} inference-only
```
sudhakarsingh27: forward pass only instead of inference-only

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -105,6 +105,23 @@ def _zero_thd_padding(tensor, cu_seqlens, cu_seqlens_padded):
     tensor[padding_mask] = 0
 
 
+def _copy_fp8_thd_half(dst, src, cu_seqlens, half_idx):
```
sudhakarsingh27: is this needed now that we have fused version?

transformer_engine/pytorch/csrc/extensions/attention.cpp
```diff
@@ -12,27 +12,9 @@ namespace {
...
-  int32_t start_row = start_index.data_ptr<int32_t>()[0];
-  void *base_ptr = static_cast<char *>(self.get_rowwise_data().data_ptr) +
-                   static_cast<size_t>(start_row) * fcd_size * element_size_bits / 8;
-  size_t num_rows_to_zero = max_tokens - start_row;
-  size_t total_bytes = num_rows_to_zero * fcd_size * element_size_bits / 8;
-
-  NVTE_SCOPED_GIL_RELEASE(
-      { nvte_memset(base_ptr, 0, total_bytes, at::cuda::getCurrentCUDAStream()); });
+// Keep zeroing stream-ordered; deriving a suffix offset from CUDA cu_seqlens on the host is unsafe.
+void mha_fill(transformer_engine::TensorWrapper &self) {
```
cyanguwa: Why don't we remove `mha_fill` entirely, and let all the `if` branches go through the `else` branch, for example, `te_O.zero_(stream)` on L152?

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -1225,12 +1225,6 @@ def _is_fa3_supported(num_heads, num_gqa_groups, head_dim_qk, head_dim_v, qkv_dt
                 " bias for THD format"
             )
             use_fused_attention = False
-        elif fp8 and fp8_meta["recipe"].fp8_dpa and qkv_format == "thd":
```
cyanguwa, line 1228: Should we let FP8 DS and CS off the hook here, but keep MXFP8 restricted by this condition?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -93,6 +93,27 @@ def get_bsh_dims(tensor_format):
     return batch_dim, seq_dim, head_dim
 
 
+def _get_thd_padding_mask(num_tokens, cu_seqlens, cu_seqlens_padded):
```
cyanguwa: I wonder if you could reuse this function instead: [get_padding_mask](https://github.com/NVIDIA/TransformerEngine/blob/d4ca825c782c59412e2ddb21c19537004d59fb89/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L1681)?

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -3405,6 +3415,11 @@ def forward(
                         )
                         max_seqlen_kv_ = kv_range[1]
                         cu_seqlens_kv_per_step[i] = thd_cu_seqlens_kv_per_step[i]
+                        if fp8 and not fp8_recipe.mxfp8():
```
cyanguwa: If we disabled MXFP8 in `utils.py`, this shouldn't need the `not fp8_recipe.mxfp8()`, right? Also, do we only need to make this one change here? Nothing for `o_part`, or in backward?

transformer_engine/common/include/transformer_engine/fused_attn.h
```diff
@@ -509,9 +509,11 @@ void nvte_cp_thd_out_correction(NVTETensor out, const NVTETensor &out_per_step,
...
  *  \param[in]     grad_per_step       THD format gradient of context parallelism.
  *  \param[in]     cu_seqlens          Cumulative sequence lengths, [batch_size + 1].
- *  \param[in]     first_half          One of ("add", "copy", "none") correction op for first half.
- *  \param[in]     second_half         One of ("add", "copy", "none") correction op for second half.
-                                       Must be different from first_half.
+ *  \param[in]     first_half          One of ("add", "copy", "none", "zero") correction op for
+ *                                     first half.
+ *  \param[in]     second_half         One of ("add", "copy", "none", "zero") correction op for
+ *                                     second half. Byte gradients support copy/zero pairs only.
+ *                                     Must be different from first_half.
```
cyanguwa: I guess "Byte gradients" means FP8 data? Would "Byte/FP8 gradients" be more obvious to readers?

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -283,13 +285,19 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
...
        (cudnn_runtime_version >= 92100 &&
         (qkv_format == NVTE_QKV_Format::NVTE_BSHD || qkv_format == NVTE_QKV_Format::NVTE_SBHD ||
-         qkv_format == NVTE_QKV_Format::NVTE_BHSD))) &&
-      !requires_64bit_ragged_offset &&
+         qkv_format == NVTE_QKV_Format::NVTE_BHSD)) ||
+       ((cudnn_runtime_version >= 92300 && sm_arch_ >= 100) &&
+        qkv_format == NVTE_QKV_Format::NVTE_THD && supported_ragged_offset_size &&
+        (attn_mask_type == NVTE_Mask_Type::NVTE_PADDING_MASK ||
+         attn_mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_MASK ||
+         (sm_arch_ >= 100 &&
```
cyanguwa: Is this sm_arch constraint duplicate to the one in L295?

transformer_engine/common/fused_attn/context_parallel.cu
```diff
@@ -42,6 +42,12 @@ struct CopyFunctor {
   }
 };
 
+struct ZeroFunctor {
+  __forceinline__ __device__ static void run(void *token, void *token_per_step, int idx) {
+    reinterpret_cast<float4 *>(token)[idx] = make_float4(0.f, 0.f, 0.f, 0.f);
```
cyanguwa, line 47: Should this be in float4 or 8-bit types?

tests/pytorch/attention/test_cp_utils.py
```diff
@@ -809,6 +811,84 @@ def _legacy_valid_copy(out, inp, cu_seqlens_padded, cu_seqlens):
...
+                )
+                for seqlens, padded_seqlens in layouts:
+                    cu_seqlens = torch.cat((torch.zeros(1, dtype=torch.int32), seqlens.cumsum(0)))
+                    cu_seqlens_padded = torch.cat(
+                        (torch.zeros(1, dtype=torch.int32), padded_seqlens.cumsum(0))
+                    )
+                    expected = self._reference_mask(cu_seqlens, cu_seqlens_padded)
+                    cu_seqlens = cu_seqlens.cuda()
+                    cu_seqlens_padded = cu_seqlens_padded.cuda()
+                    actual = _get_thd_padding_mask(expected.numel(), cu_seqlens, cu_seqlens_padded)
```
cyanguwa: Again, should we use [get_padding_mask](https://github.com/NVIDIA/TransformerEngine/blob/d4ca825c782c59412e2ddb21c19537004d59fb89/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L1681) for this functionality?

transformer_engine/pytorch/csrc/extensions/attention.cpp
sudhakarsingh27: Done. I removed `mha_fill`.

For the forward output, `TensorWrapper::zero_()` remains the right operation because TE's implementation clears both [the tensor data and its `amax`](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/transformer_engine/common/transformer_engine.cpp#L1012-L1030). In backward, dQ/dK/dV can be strided views of combined storage, so their data is cleared through PyTorch's elementwise [`fill_(0)`](https://docs.pytorch.org/docs/stable/generated/torch.Tensor.fill_.html), followed by an explicit reset of the shared delayed-scaling `amax`. The resulting split is [here](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/transformer_engine/pytorch/csrc/extensions/attention.cpp#L433-L448).

This avoids treating a strided view as one contiguous byte range while still clearing the FP8 metadata.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
sudhakarsingh27: Yes. Delayed scaling and current scaling are now allowed through this check. The THD restriction is inside the `fp8_recipe.mxfp8()` branch [here](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L689-L706), so MXFP8+THD remains disabled while the other FP8 recipes can use the supported fused-attention path.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: I checked whether these could share an implementation, but they return different representations. [`get_padding_mask()`](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L1667-L1706) builds a dense `[batch, 1, 1, max_seqlen]` mask for attention scores. [`get_thd_padding_mask()`](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L1658-L1664) instead builds a flat `[num_tokens]` mask for the physical holes between sequences in a packed THD buffer.

I kept the implementations separate, but moved the THD helper beside the dense helper in `utils.py`. Its boundary lookup follows the documented [`torch.searchsorted` semantics](https://docs.pytorch.org/docs/stable/generated/torch.searchsorted.html).

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
sudhakarsingh27: Yes, I removed the redundant `not fp8_recipe.mxfp8()` at this site. The communicated THD Q/K/V views are now rewrapped whenever FP8 is active [here](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py#L3404-L3415).

MXFP8+THD is already rejected during backend selection [here](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L689-L706), so this path serves delayed and current scaling. This block only reconstructs the communicated Q/K/V inputs before attention; `o_part` and backward use their existing conversion and saved-tensor paths, so they do not need a matching change at this location.

transformer_engine/common/include/transformer_engine/fused_attn.h
sudhakarsingh27: Yes, I changed the wording to make FP8 explicit. CUDA defines FP8 storage as an [8-bit unsigned storage type](https://docs.nvidia.com/cuda/archive/12.6.0/cuda-math-api/group__CUDA__MATH__FP8__MISC.html). This kernel receives those encoded bytes directly, so numerically adding them would add the encodings rather than the represented values.

The updated [API documentation](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/transformer_engine/common/include/transformer_engine/fused_attn.h#L507-L524) therefore states that FP8 gradients support only the copy/zero and zero/copy correction pairs.

transformer_engine/common/fused_attn/fused_attn.cpp
sudhakarsingh27: Yes, it was duplicate, so I removed it. The enclosing THD condition already requires `sm_arch_ >= 100`; bottom-right causal masking is now simply one of the accepted mask choices inside that condition. The consolidated check is [here](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/transformer_engine/common/fused_attn/fused_attn.cpp#L288-L301).

transformer_engine/common/fused_attn/context_parallel.cu
sudhakarsingh27: I kept `float4` because it is being used only as an aligned 16-byte transport type. The [copy and zero functors](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/transformer_engine/common/fused_attn/context_parallel.cu#L39-L48) only reinterpret, assign, or write zero bits; they do not perform floating-point arithmetic on the FP8 payload.

The CUDA Programming Guide lists `float4` as [16-byte types with 16-byte alignment](https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/cpp-language-extensions.html#built-in-types).

tests/pytorch/attention/test_cp_utils.py
sudhakarsingh27: The tests now import the shared [`get_thd_padding_mask()` helper from `utils.py`](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/tests/pytorch/attention/test_cp_utils.py#L7-L17). The dense `get_padding_mask()` cannot replace it because it masks attention-score coordinates, whereas this test needs a flat mask over physical padding rows in the packed THD buffer.

The focused tests cover the computed mask, preservation of valid rows, zeroing of padding rows, and CUDA graph replay [here](https://github.com/NVIDIA/TransformerEngine/blob/a34fdc1e7e14557c067a0e8f26bb11cc613a539c/tests/pytorch/attention/test_cp_utils.py#L814-L889).

cyanguwa (dismissed):
Please run a full PyTorch CI before merging. Just wanted to be double sure. Thanks!

cyanguwa (approved):

PR #3001, opened 2026-05-16
Optimize function that loads pointers on GPU

timmoon10:
# Description

[`tex.get_device_pointer_for_data_and_scales`](https://github.com/NVIDIA/TransformerEngine/blob/eca05d3b554e36d99106c368f45df2cc350ddebc/transformer_engine/pytorch/csrc/extensions/utils.cpp#L51) has two problems:
1. It has significant CPU overhead (see https://github.com/NVIDIA/TransformerEngine/issues/2897). In a representative benchmark on a GB200, it takes ~70 us per call.
2. The meaning is extremely unintuitive. The most natural interpretation is that it takes a FP8/MXFP8/NVFP4 tensor and returns pointers as two `int` s. But actually it takes the buffers from multiple MXFP8/NVFP4 tensors (all assumed to have the same shape), swizzles the scaling factors, and transfers the pointers to a GPU array in a CUDA Graph-friendly way.

This PR makes several optimizations to reduce CPU overhead, mostly by avoiding heap allocations and mutex acquisition. I've also attempted to make the functionality more general and logical:
- `nvte_copy_host_to_device_via_kernel_args`: A general function for copying a small amount of data to GPU in a CUDA Graph-friendly way. Unlike `nvte_convert_pointers_to_tensor`, it makes no assumptions that the data is a list of pointers.
- `tex.copy_data_ptrs_to_device`: Takes a list of tensors and puts their data pointers into a GPU buffer.
- `tex.transform_and_copy_data_ptrs_to_device`: Performs a user-specified transform on a list of tensors and puts the resulting data pointers into a GPU buffer. Currently it only supports scale swizzles on uniformly shaped tensors, but the transform names help make the contracts explicit.

With these changes, per-call CPU runtime has dropped from 70 us to 31 us on a GB200 node.

This is progress toward https://github.com/NVIDIA/TransformerEngine/issues/2897.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring
- [x] Performance optimization

## Changes

- Add `transformer_engine::Tensor::flat_2d_dims` to compute first and last dims simultaneously
- Generalize and rename `nvte_load_value_on_device`
- Refactor and rename `tex.load_data_ptrs_on_device` and `tex.transform_and_load_data_ptrs_on_device`
- Add internal wrapper class for `NVTEShape` with similar API as `std::vector`
- Remove heap allocations in `transformer_engine::SimpleTensor`
- Remove heap allocations in `transformer_engine::Tensor` shape functions
- Add batched tensor allocation and deallocation to reduce mutex overhead
- Avoid heap allocations in tensor checking functions

# Checklist:


transformer_engine/common/common.h
```diff
@@ -121,12 +122,19 @@ struct SimpleTensor {
...
 
+  SimpleTensor() : SimpleTensor(nullptr, std::vector<size_t>{0}, DType::kFloat32) {}
+
   SimpleTensor(const NVTEBasicTensor &tensor)  // NOLINT
       : dptr(tensor.data_ptr),
         shape(tensor.shape.data, tensor.shape.data + tensor.shape.ndim),
         dtype(static_cast<DType>(tensor.dtype)) {}
 
-  SimpleTensor() : SimpleTensor(nullptr, std::vector<size_t>{0}, DType::kFloat32) {}
+  SimpleTensor &operator=(const NVTEBasicTensor &tensor) {
```
timmoon10: Without this assignment operator, assigning from a `NVTEBasicTensor` triggers a heap allocator in the `NVTEBasicTensor` constructor. We do this assignment frequently within [`nvte_set_tensor_param_v2`](https://github.com/NVIDIA/TransformerEngine/blob/eca05d3b554e36d99106c368f45df2cc350ddebc/transformer_engine/common/transformer_engine.cpp#L800).

transformer_engine/common/util/utils.cu
```diff
@@ -7,45 +7,75 @@
...
+  uint32_t vectors[kMaxVectors];
 };
 
-__global__ void write_pointers_kernel(HostPointersArgs args, uint64_t *out, int64_t count,
-                                      int64_t offset) {
-  const int64_t idx = static_cast<int64_t>(blockIdx.x) * blockDim.x + threadIdx.x;
-  if (idx < count) {
-    out[offset + idx] = args.ptrs[idx];
+constexpr size_t block_size = 512;
+constexpr size_t num_blocks = DIVUP(Payload::kMaxBytes / Payload::kVectorSize, block_size);
```
timmoon10:
```suggestion
constexpr size_t num_blocks = DIVUP(Payload::kMaxVectors, block_size);
```

transformer_engine/pytorch/csrc/extensions/utils.cpp
```diff
@@ -6,160 +6,172 @@
...
-    std::vector<at::Tensor> data_tensors, std::vector<at::Tensor> scale_tensors, bool swizzle,
-    bool rowwise, transformer_engine::DType data_dtype) {
-  const size_t num_tensors = data_tensors.size();
-  NVTE_CHECK(num_tensors > 0, "data_tensors must not be empty.");
-  NVTE_CHECK(num_tensors == scale_tensors.size(),
-             "data_tensors and scale_tensors must have the same size.");
-  NVTE_CHECK(data_tensors[0].is_cuda(), "data_tensors must be on CUDA.");
-  const auto device = data_tensors[0].device();
-  auto stream = at::cuda::getCurrentCUDAStream();
+std::tuple<at::Tensor, std::optional<at::Tensor>> transform_and_load_data_ptrs_on_device(
```
timmoon10: I'm not committed to this name. I based it on [`std::transform`](https://en.cppreference.com/cpp/algorithm/transform). I suppose "map" would be more Python-focused, but that sounds worse.

transformer_engine/common/util/utils.cu
```diff
@@ -7,45 +7,75 @@
...
+    std::memcpy(payload.bytes, src + offset, chunk_size);
+    kernel<<<num_blocks, block_size, 0, stream>>>(payload, chunk_size, dst + offset);
+    NVTE_CHECK_CUDA(cudaGetLastError());
+  }
+}
 
 void nvte_convert_pointers_to_tensor(const uint64_t *host_ptrs, NVTETensor output, int64_t count,
                                      cudaStream_t stream) {
-  NVTE_API_CALL(nvte_convert_pointers_to_tensor);
   using namespace transformer_engine;
```
timmoon10, line 78: https://github.com/NVIDIA/TransformerEngine/pull/3001/changes#r3252732073
```suggestion
  NVTE_API_CALL(nvte_convert_pointers_to_tensor);
  using namespace transformer_engine;
```

transformer_engine/common/transformer_engine.cpp
```diff
@@ -91,54 +92,55 @@ void CheckScaleTensorShape(const Tensor &t, const std::string &name) {
...
+        constexpr std::array<size_t, 2> block_shape{1, 16};
+        constexpr std::array<size_t, 2> block_alignment{128, 4};
+        const std::array<size_t, 2> expected{
+            DIVUP_TO_MULTIPLE(DIVUP(last_dim, block_shape[0]), block_alignment[0]),
+            DIVUP_TO_MULTIPLE(DIVUP(first_dim, block_shape[1]), block_alignment[1])};
+        NVTE_CHECK(t.columnwise_scale_inv.shape.size() == 2 &&
+                       t.columnwise_scale_inv.shape[0] == expected[0] &&
+                       t.columnwise_scale_inv.shape[1] == expected[1],
+                   "Tensor \"", name, "\" has invalid columnwise_scale_inv shape (expected ",
+                   expected, ", got ", t.scale_inv.shape, ")");
```
timmoon10: https://github.com/NVIDIA/TransformerEngine/pull/3001/changes#r3252731973
```suggestion
                   expected, ", got ", t.columnwise_scale_inv.shape, ")");
```

ptrendx:
Seems a lot of those changes would basically not be needed if we did not use the std::vector in Tensor/SimpleTensor and just used NVTEShape everywhere - this would effectively make SimpleTensor and NVTEBasicTensor the same thing (we could even do the constructor in the public header, just behing the `if cplusplus` guard).

transformer_engine/common/util/utils.cu
```diff
@@ -7,45 +7,76 @@
...
+      static_cast<uint8_t *>(out)[i] = payload.bytes[i];
+    }
   }
 }
 
 }  // namespace
+}  // namespace load_value_on_device
+}  // namespace transformer_engine
+
+void nvte_load_value_on_device(const void *host_ptr, void *device_ptr, size_t num_bytes,
```
vthumbe1503: Is there a reason we are exposing this as an NVTE function? In other words do you see a use-case where one might want to use this function over nvte_convert_pointers_to_tensor?

transformer_engine/pytorch/ops/fused/backward_grouped_mlp.py
```diff
@@ -498,12 +498,14 @@ def fuser_backward(
...
                 [w._columnwise_data for w in grouped_fc2_weight],
+                device,
+            )
+            fc2_sfb_ptrs, _fc2_sfb_buffer = tex.transform_and_load_data_ptrs_on_device(
+                "uniform_mxfp8_columnwise_swizzle",
                 [w._columnwise_scale_inv for w in grouped_fc2_weight],
-                swizzle=True,
-                rowwise=False,
-                data_dtype=grouped_fc2_weight[0]._fp8_dtype,
+                device,
```
vthumbe1503: The other optimization can be to load both fc1 and fc2 data and scale inv togegther at the start of backward. I am hoping it wouldnt make the code ugly.

transformer_engine/common/common.h
```diff
@@ -121,12 +122,19 @@ struct SimpleTensor {
...
+    dptr = tensor.data_ptr;
+    shape.assign(tensor.shape.data, tensor.shape.data + tensor.shape.ndim);
```
vthumbe1503: So when you say heap allocations being done redundantly again and again. Do you mean the vector to NVTEShape conversions?

I rememember  this problem being observed even with a basic te linear profiling. And I hadnt gotten this PR merged.
https://github.com/NVIDIA/TransformerEngine/pull/2514

which essentially standadizes to use NVTEShape everywhere instead of using vector at all to avoid bouncing back and forth between the two allocations. Maybe it might be worth to revive the PR?

vthumbe1503 (commented):
Thanks for cleaning up the APIs. Looks much nicer now. CPU overheads being caused by heap allocations of shape, makes me wonder whether we should revive this [PR](https://github.com/NVIDIA/TransformerEngine/pull/2514) to standardize on NVTEShape yo avoid back and forth between vector<size_t> and NVTE_Shape

transformer_engine/common/util/utils.cu
timmoon10: My intention is to deprecate `nvte_convert_pointers_to_tensor` and replace with `nvte_load_value_on_device` since its more general.

transformer_engine/common/common.h
timmoon10: Previously, assigning an `NVTEBasicTensor` to a `SimpleTensor` would trigger the constructor and then the move operator. This would allocate an `std::vector`, move it, and deallocate the old `std::vector`.

One other approach I was thinking about was implementing a `Shape` class that wraps around `NVTEShape` and has a similar API as `std::vector`. That way we can keep the nice ergonomics, while avoiding heap allocations.

vthumbe1503: I had tried your other approach in the 2514 PR above, but eventually had removed it due to some complications. I have refer back to my notes on why it didnt work out for me.

But here is the commit that reverted it
https://github.com/NVIDIA/TransformerEngine/pull/2514/changes/b5997763942fb2a4bd85603643db6d494f402cb1

I had called it NVTEShapeWrapper and implemented all the vector based APIs.

vthumbe1503: One complication I do remember was to change a lot of attention interfaces to have NVTEShapeWrapper instead of using vector.

timmoon10: I've found that adding a cast operator to `std::vector` helps reduce the number of places we need to change the interfaces.

transformer_engine/common/util/utils.cu
Oleg-Goncharov: I think this function name is a bit confusing for public API. The name says "load_value", but the arguments are raw pointers and num_bytes, so it looks more like copying host bytes to device. Also the implementation launches multiple kernels if `num_bytes > Payload::kMaxBytes`, which may be unexpected for users who read this as a generic H2D copy helper.

Would it be better if we make it more explicit, e.g.:
- rename to `nvte_copy_small_host_payload_to_device`
- add a check to make sure this is only used for small CUDA graph-friendly payloads

transformer_engine/pytorch/csrc/extensions/utils.cpp
```diff
@@ -6,160 +6,177 @@
...
-      const size_t scale_numel = static_cast<size_t>(scale_tensors[i].numel());
-      const size_t dtype_bits = transformer_engine::pytorch::typeToNumBits(scale_dtype);
-      output_bytes = roundup(output_bytes, 16);
-      output_offsets.push_back(output_bytes);
-      output_bytes += ceildiv(scale_numel * dtype_bits, 8);
+    // Data types
+    transformer_engine::DType data_dtype, scale_dtype;
+    switch (scaling_mode) {
+      case NVTE_MXFP8_1D_SCALING:
+        data_dtype = transformer_engine::DType::kFloat8E4M3;
```
Oleg-Goncharov, line 77: Do we really want to hardcode `data_dtype = kFloat8E4M3` here?

transformer_engine/common/util/utils.cu
timmoon10: I'm not quite happy with the name either. My intention was "load an arbitrarily-sized object into GPU, but it's optimized for small things like structs". Some other ideas I had:
- "memcpy" or "copy": Perfect match for the API, but it also doesn't give any hint that it is optimized for very small copies. Also, it doesn't communicate that the data is immediately passed as a kernel arg, so the host buffer can be immediately freed and this is compatible with CUDA Graphs.
- "fill": Implies that we are copying a single thing and that the data is included as a kernel arg. However, [`std::fill`](https://en.cppreference.com/cpp/algorithm/fill) and [`torch.Tensor.fill_`](https://docs.pytorch.org/docs/2.12/generated/torch.Tensor.fill_.html) repeat the value multiple times in the output buffer, rather than copying directly.
- "load": Consistent with my intended meaning, but vague. Reminds me of [`cuModuleLoadDataEx`](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MODULE.html), which has some similarities but is also different enough that we need to avoid confusion.
- "store": Similar to "load".

I don't think we should enforce a single kernel launch though. The intended use-case is to copy lists of device pointers, which can become large for MoE models with large numbers of experts. The implementation can handle arbitrarily-sized data correctly, although perf may be bad.

transformer_engine/pytorch/csrc/extensions/utils.cpp
timmoon10: We don't actually access the fp8e4m3 values when swizzling, this is a fake configuration so the tensor passes validation checks.

transformer_engine/common/util/utils.cu
timmoon10: Perplexity likes `nvte_copy_host_to_device_via_kernel`, `nvte_copy_host_to_device_immediate`, `nvte_copy_host_to_device_graph_safe`. The first one seems the clearest and least awkward to me.

transformer_engine/common/common.h
```diff
@@ -1066,9 +1185,9 @@ GroupedTensor *convertNVTEGroupedTensor(const NVTEGroupedTensor tensor);
 GroupedTensor *convertNVTEGroupedTensorCheck(const NVTEGroupedTensor tensor);
 
 // Helper functions for GroupedTensor validation
-void CheckGroupedTensorShapeArrays(const GroupedTensor &t, const std::string &name);
-void CheckInputGroupedTensor(const GroupedTensor &t, const std::string &name);
-void CheckOutputGroupedTensor(const GroupedTensor &t, const std::string &name,
+void CheckGroupedTensorShapeArrays(const GroupedTensor &t, const std::string_view &name);
```
vthumbe1503: std::string_view is already a reference. Additional & in the API might be an overkill

transformer_engine/common/util/utils.cu
vthumbe1503: +1 for nvte_copy_host_to_device_via_kernel

transformer_engine/common/util/utils.cu
```diff
@@ -7,45 +7,76 @@
...
+             "Device pointer is not aligned to ", Payload::kVectorSize, " bytes.");
+
+  // Chunk data to fit in kernel arguments and launch kernels
+  const uint8_t *src = static_cast<const uint8_t *>(host_ptr);
+  uint8_t *dst = static_cast<uint8_t *>(device_ptr);
+  for (size_t offset = 0; offset < num_bytes; offset += Payload::kMaxBytes) {
+    const size_t chunk_size = std::min(num_bytes - offset, Payload::kMaxBytes);
+    Payload payload{};
+    std::memcpy(payload.bytes, src + offset, chunk_size);
+    kernel<<<num_blocks, block_size, 0, stream>>>(payload, chunk_size, dst + offset);
```
vthumbe1503, line 70: I am curious about why we are doing std::memcpy.. Is it that Payload struct enables a cleaner way to enable vectorization?

timmoon10: We aren't making assumptions on the input type, so the API should take a `const void *`. Within the implementation, it's easier just to call `std::memcpy` since a loop would require an extra cast to `uint8_t` or something.

transformer_engine/common/normalization/layernorm/ln_api.cpp
```diff
@@ -100,7 +100,7 @@ void layernorm_fwd(const Tensor& x,      // BxSxhidden_size
       gamma_in_weight_dtype);
 
   if (workspace->data.numel() == 0) {
-    workspace->data.shape = plan->getWorkspaceShape();
+    workspace->data.shape = Shape(plan->getWorkspaceShape());
```
ptrendx: getWorkspaceShape is a purely internal function, why can't we change it to actually return `Shape`?

transformer_engine/common/swizzle/swizzle.cu
```diff
@@ -25,14 +25,15 @@ constexpr int MXFP8_BLOCK_SIZE = 32;
...
+  auto query_max_smem = []() -> int {
+    int device{0}, max_smem{0};
     NVTE_CHECK_CUDA(cudaGetDevice(&device));
     NVTE_CHECK_CUDA(
         cudaDeviceGetAttribute(&max_smem, cudaDevAttrMaxSharedMemoryPerBlockOptin, device));
-  }
-  return max_smem;
+    return max_smem;
+  };
+  static int cached_val = query_max_smem();
```
ptrendx: This is nice but wrong - we need to be able to support this value per GPU rather than a single global, similarly to the other CUDA properties that we have in the CUDA runtime helpers (also, this function should live together with those other functions).

transformer_engine/common/swizzle/swizzle.cu
```diff
@@ -1456,10 +1457,8 @@ void multi_tensor_swizzle_scaling_factors(const std::vector<Tensor*>& input,
 
     // We don't allow empty tensors. They should be filtered out before calling this function.
     NVTE_CHECK(input[i]->numel() != 0, "Tensor input[", i, "] is empty.");
-    CheckInputTensor(*input[i], "scaling_factor_input[" + std::to_string(i) + "]",
-                     check_scale_inv_shapes);
-    CheckInputTensor(*output[i], "scaling_factor_output[" + std::to_string(i) + "]",
-                     check_scale_inv_shapes);
+    CheckInputTensor(*input[i], "scaling_factor_input", check_scale_inv_shapes);
```
ptrendx, line 1462: This is making the error quality worse. I think the better approach would be to have a version of this function that still performs this string concatenation, but does so only when we actually want to raise an error.

transformer_engine/common/common.h
```diff
@@ -113,24 +119,135 @@ size_t get_buffer_size_bytes(const size_t N, const DType buffer_dtype);
...
+  void push_back(size_type value) {
+    NVTE_CHECK(data_.ndim < max_ndim, "Cannot add dimension: shape is at maximum capacity (",
+               max_ndim, ").");
+    data_.data[data_.ndim++] = value;
+  }
+
+  void resize(size_type count) {
+    NVTE_CHECK(count <= max_ndim, "Too many dimensions (requested ", count, ", max is ", max_ndim,
+               ").");
+    data_.ndim = count;
```
ptrendx, line 210: If you wanted to be compatible with the vector then this is incomplete since resize in vector would also default-initialize the contents that were added, whereas this will leave them uninitialized.

transformer_engine/common/common.h
```diff
@@ -220,10 +337,11 @@ struct Tensor {
...
   /*! Number of tensor elements. */
   size_t numel() const {
-    if (!has_data() && has_columnwise_data()) {
-      return product(columnwise_data.shape);
+    size_t ret = 1;
+    for (const size_t dim : shape()) {
+      ret *= dim;
     }
-    return product(data.shape);
+    return ret;
```
ptrendx: You have the product function defined, why not use it here?

transformer_engine/common/transformer_engine.cpp
```diff
@@ -437,6 +433,32 @@ class TensorAllocator {
                MAX_TENSOR_NUM, ". There is probably a memory leak in your application.");
   }
 
+  void Allocate(NVTEScalingMode mode, NVTETensor *out, size_t N) {
```
ptrendx, line 405: We should then implement regular Allocate as
```cpp
NVTETensor Allocate(NVTEScalingMode mode) {
  NVTETensor t;
  Allocate(mode, &t, 1);
  return t;
}```

File: transformer_engine/pytorch/csrc/extensions/recipe.cpp [hunk 18]
```diff
@@ -35,35 +35,37 @@ void fused_amax_and_scale_update_after_reduction(const at::Tensor& amax_reductio
...
-  std::vector<NVTETensor> te_scales;
-  te_amax_histories.reserve(num_tensors);
-  te_scales.reserve(num_tensors);
+
+  // Helper to deallocate a batch of NVTETensors
+  struct DestroyGuard {
+    NVTETensor* data;
+    size_t n;
+    ~DestroyGuard() { nvte_destroy_tensors(data, n); }
+  };
```
ptrendx: Why can't we just create a TensorVector or something like that which would do the allocation and deallocation in the RAII way and which could give the reference to the std::vector inside rather than doing this partial thing here?

ptrendx: Or something like
```cpp
TensorVector amax_histories(num_tensors);
auto& te_amax_histories = amax_histories.vec();
```
and then use the te_amax_histories like you would normally.

transformer_engine/pytorch/csrc/extensions/swizzle.cpp
```diff
@@ -202,75 +204,86 @@ std::optional<at::Tensor> multi_tensor_swizzle_scales_for_gemm_impl(
...
+  // Allocate input/output NVTETensors as a single batch. The first
+  // n_swizzle entries are inputs; the next n_swizzle are outputs.
+  const size_t n_swizzle = tensors_needing_swizzle.size();
+  std::vector<NVTETensor> nvte_tensors(2 * n_swizzle);
+  nvte_create_tensors(scaling_mode, nvte_tensors.data(), nvte_tensors.size());
+  struct DestroyGuard {
+    NVTETensor *data;
+    size_t n;
+    ~DestroyGuard() { nvte_destroy_tensors(data, n); }
+  } destroy_guard{nvte_tensors.data(), nvte_tensors.size()};
```
ptrendx: Oh wow - yeah, this DestroyGuard (or the other solution) should be defined somewhere where it will be visible to multiple extensions.

transformer_engine/pytorch/csrc/extensions/utils.cpp
```diff
@@ -6,160 +6,178 @@
...
-  std::vector<uint64_t> scale_host_ptrs(num_tensors);
+  // CUDA stream
+  auto stream = at::cuda::getCurrentCUDAStream();
 
-  if (swizzle) {
-    NVTEScalingMode scaling_mode;
-    transformer_engine::DType scale_dtype;
-    if (is_fp8_dtype(data_dtype)) {
+  // Swizzle scales for GEMM, with uniform tensor sizes
+  const bool uniform_mxfp8_rowwise_swizzle = transform_type == "uniform_mxfp8_rowwise_swizzle";
```
ptrendx, line 61: If we want to make it possibly the fastest while still using strings, we should start the strings with the things that are different, rather than start with the "uniform_" part that will be the same in every case (and therefore the loop that this will become will need to go over that part every time no matter what).

transformer_engine/common/swizzle/swizzle.cu
timmoon10: Updated to handle the multi-device case.

transformer_engine/common/swizzle/swizzle.cu
timmoon10: I guess we could implement the `Check*Tensor` functions with variadic templates and unpack the args in `NVTE_CHECK`. Alternatively we could pass in a closure that creates the name. These seem overkill though. Both approaches seem excessive for what should be a quick sanity check.

transformer_engine/pytorch/csrc/extensions/recipe.cpp
timmoon10: Added a `MultiTensorWrapper` class. Right now it just exposes raw `NVTETensor`s, but in the future we could consider getter/setters or we could make it construct non-owning `TensorWrapper`s.

transformer_engine/pytorch/csrc/extensions/utils.cpp
timmoon10: It turns out this microoptimization won't be helpful. A quick benchmark on a GB200 node shows that string comparisons are vectorized into 16 byte chunks and that each chunk takes ~0.5 ns.

image

PR #3009, opened 2026-05-18
Bitmap topk

### LINKED ISSUE #2999: Output routing map from fused_topk_with_scores in bitmap format
Currently the format is bytemap. We need to change this to bitmap format for optimization (less space, more compact) and for compatibility with TE EP.

tdophung:
# Description
Add a new path to our topk kernel to output the routing map in bitmap format instead of bytemap alone. The default still stay at bytemap so no regression for existing consumers downstream of this op. However, since the op now requires an additional arg to specify the routing map type (bytemap or bitmap), we introduce a V2 of the API to accomplish this, while keeping the original API the same not to break customers.

This helps NCCL EP not have to do the token_indices (sparse format) conversion to bitmap format for comms later. 

Fixes #2999 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes
- Change in the kernel for an if path that does atomicOr for all expert indices shifted by bit position to create the expert indices -> bitmap conversion.
- Plumb the arg for routing map mode (byte map or bitmap) through pytorch and jax primitives

# Checklist:


transformer_engine/common/fused_router/fused_score_for_moe_aux_loss.cu
```diff
@@ -20,7 +20,8 @@ template <typename DataType, TopkFuncType TopkFunc = TopkFuncType::Naive>
 __global__ void fused_score_for_moe_aux_loss_forward_kernel(const DataType *logits, int num_tokens,
                                                             int num_experts, int topk,
                                                             int score_function, float *scores,
-                                                            bool *routing_map,
+                                                            uint8_t *routing_map,
+                                                            NVTERoutingMapFormat routing_map_format,
```
jberchtold-nvidia: Should we make `NVTERoutingMapFormat` a template arg and put the branch in the host code that calls the kernel? I've seen our other kernels do that but I'm unsure if this branch will introduce any runtime overhead.

transformer_engine/common/fused_router/fused_score_for_moe_aux_loss.cu
```diff
@@ -349,14 +391,15 @@ void fused_score_for_moe_aux_loss_backward(const Tensor &intermediate_output,
 
 void nvte_fused_score_for_moe_aux_loss_forward(const NVTETensor logits, int num_tokens,
```
jberchtold-nvidia, line 443: This changes our public C API so we may need to introduce a `_v2` API. Can you ask Przemek if he thinks that is required here? If so, in the V2 API it may be useful to have a NVTERouterConfig or something similar like we have on other C API functions that stores most or all static args. That way if we need to add more static args in the future we can add defaults to the struct instead without breaking the API.

transformer_engine/common/include/transformer_engine/fused_router.h
```diff
@@ -13,6 +13,21 @@
...
 #endif
 
+/*! \brief Output format of the routing_map tensor.
+ *
+ *  BYTEMAP   — bool/uint8 tensor of shape [num_tokens, num_experts]; one byte
+ *              per (token, expert) pair, 0 or 1.
+ *  BITMAP_U8 — uint8 tensor of shape [num_tokens, ceil(num_experts/8)]; bit
+ *              (e % 8) of byte (e / 8) of row t is 1 iff token t routes to
+ *              expert e (little-endian / LSB-first packing along the expert
+ *              axis). This is the layout NCCL EP dispatch is expected to
```
jberchtold-nvidia: SR: Let's remove "This is the layout NCCL EP dispatch is expected to consume directly.". This is a specific detail of another component that may change in the future or we may introduce different components that also consume this bitmap format

transformer_engine/common/include/transformer_engine/fused_router.h
```diff
@@ -26,19 +41,24 @@ extern "C" {
...
- *  \param[out]    routing_map     Output tensor for routing map.
+ *  \param[out]    routing_map     Output tensor for routing map. Shape depends on
+ *                                 routing_map_format (see NVTERoutingMapFormat).
+ *  \param[in]     routing_map_format NVTERoutingMapFormat value selecting the routing_map
+ *                                    output layout. The caller is responsible for
+ *                                    allocating routing_map with the matching shape.
  *  \param[out]    intermediate_output  Output tensor for intermediate output. (Softmax/sigmoid output)
  *  \param[in]     stream          CUDA stream used for the operation.
  */
 void nvte_fused_topk_with_score_function_forward(
```
jberchtold-nvidia: Same comment as above about C API change

transformer_engine/common/fused_router/fused_topk_with_score_function.cu
```diff
@@ -502,7 +561,7 @@ void fused_topk_with_score_function_backward(const Tensor &routing_map,
   TE_ROUTER_PROBS_TYPE_SWITCH_ALL(
```
jberchtold-nvidia: Were there no existing shape `NVTE_CHECK`s for this function to verify inputs and output shapes are valid? We should add some for both the bytemap and bitmap cases

transformer_engine/jax/cpp_extensions/router.py
```diff
@@ -28,6 +29,20 @@ class ScoreFunction(IntEnum):
     SOFTMAX = int(JAXX_Score_Function.SOFTMAX)
 
 
+class RoutingMapFormat(IntEnum):
+    """Routing-map output layout, synced with C++ JAXX_Routing_Map_Format / NVTERoutingMapFormat.
+
+    BYTEMAP   — bool/uint8 tensor of shape [num_tokens, num_experts].
+    BITMAP_U8 — uint8 tensor of shape [num_tokens, ceil(num_experts/8)]; bit
+                (e % 8) of byte (e / 8) of row t is 1 iff token t routes to
+                expert e (LSB-first packing along the expert axis). This is the
```
jberchtold-nvidia: SR: remove consumer-specific details (NCCL  EP)

transformer_engine/jax/csrc/extensions/misc.h
```diff
@@ -127,6 +127,13 @@ enum class JAXX_Score_Function : int64_t {
   SOFTMAX = 1,
 };
 
+// Mirror of NVTERoutingMapFormat for JAX FFI plumbing. Values must stay in sync
+// with the C enum in transformer_engine/common/include/transformer_engine/fused_router.h.
+enum class JAXX_Routing_Map_Format : int64_t {
```
jberchtold-nvidia, line 134: For the values here can we use `BYTEMAP = NVTE_ROUTING_MAP_FORMAT_BYTEMAP` rather  than `BYTEMAP  = 0`

transformer_engine/jax/router.py
```diff
@@ -35,11 +36,29 @@
...
+        return routing_map_format
+    if isinstance(routing_map_format, int):
+        return RoutingMapFormat(routing_map_format)
+    try:
+        return RoutingMapFormat[routing_map_format.upper()]
+    except (KeyError, AttributeError):
+        raise ValueError(
+            "routing_map_format must be 'bytemap', 'bitmap_u8', a RoutingMapFormat enum, "
+            f"or the matching int; got {routing_map_format!r}"
+        ) from None
```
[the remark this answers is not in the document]
tdophung: Done

transformer_engine/common/fused_router/fused_score_for_moe_aux_loss.cu
tdophung: Done in latest commit

transformer_engine/common/fused_router/fused_score_for_moe_aux_loss.cu
tdophung: I asked him and he said I should have a v2, and also put in aa warning that V1 will be obsoleted, and in rerlease note, and in PR description also. So updated to have v2 in latest commit

transformer_engine/pytorch/router.py
```diff
@@ -63,6 +114,7 @@ def forward(
         ctx.topk = topk
         ctx.scaling_factor = scaling_factor
         ctx.score_function = score_function
+        ctx.routing_map_format = routing_map_format
         ctx.logits_dtype = logits.dtype
         return probs, routing_map
```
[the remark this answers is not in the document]
tdophung: added shape checking

transformer_engine/common/fused_router/fused_topk_with_score_function.cu
tdophung: added checks now in latest committ

transformer_engine/jax/csrc/extensions/misc.h
tdophung: Done

jberchtold-nvidia (dismissed):
LGTM, thanks! I reviewed core and JAX changes but not PyTorch

transformer_engine/pytorch/router.py
```diff
@@ -13,12 +13,61 @@
...
+RoutingMapFormat = tex.NVTERoutingMapFormat
+
+
+_ROUTING_MAP_FORMAT_FROM_STRING = {
+    "bytemap": RoutingMapFormat.BYTEMAP,
+    "bitmap_u8": RoutingMapFormat.BITMAP_U8,
+}
+
+
+def _validate_routing_map_format(
```
ptrendx: Could we use regular ints instead of the enums and only do this just before calling the tex function? This function screams CPU overhead so can we simplify it, to not e.g. do lower on the string every time etc.?

transformer_engine/pytorch/router.py
```diff
@@ -36,6 +85,7 @@ def forward(
         scaling_factor: Optional[float],
         score_function: str,
         expert_bias: Optional[torch.Tensor],
+        routing_map_format: "RoutingMapFormat",
```
ptrendx: Why do you put the string here as the type?

transformer_engine/pytorch/router.py
```diff
@@ -53,16 +103,23 @@ def forward(
...
             expert_bias,
+            routing_map_format,
         )
-        # Restore the shape
-        probs = probs.view(tensor_shape)
+        # Save the flat 2D routing_map for backward (kernel indexes by
+        # num_tokens x trailing_dim), then restore the leading dims of the
+        # input on the returned outputs. The trailing dim of routing_map
+        # depends on the format: num_experts for BYTEMAP, ceil(num_experts/8)
+        # for BITMAP_U8.
```
ptrendx: Please don't do that and instead use flat_first_dim and flat_last_dim on the C++ side to avoid Python overheads here - so the routing map should be created with the proper shape and we should not do the view afterwards.

transformer_engine/pytorch/router.py
```diff
@@ -164,13 +238,19 @@ def forward(
...
         ctx.topk = topk
         ctx.score_function = score_function
         ctx.num_tokens = num_tokens
         ctx.num_experts = num_experts
         ctx.logits_dtype = logits.dtype
+        # Restore the leading dims of the input on both outputs. The trailing
+        # dim of routing_map depends on the format: num_experts for BYTEMAP,
+        # ceil(num_experts/8) for BITMAP_U8.
+        scores = scores.view(tensor_shape)
+        routing_map = routing_map.view(*tensor_shape[:-1], routing_map.shape[-1])
```
ptrendx: As before - please don't do that, let's avoid the views on the Python side.

transformer_engine/pytorch/router.py
tdophung: no good reason. This is AI slop. I have changed it to int to match the int that is passed around

transformer_engine/pytorch/router.py
tdophung: okay done, see line 120-123 in router.cpp for the flat_first_dim and flat_last_dmi calculation

tests/pytorch/test_fused_router.py
```diff
@@ -458,6 +459,145 @@ def test_fused_moe_aux_loss(dtype, num_tokens, num_experts, topk):
     torch.testing.assert_close(probs.grad, probs_clone.grad, atol=atol, rtol=rtol)
 
 
+# =============================================================================
+# Test: routing_map BITMAP_U8 vs BYTEMAP parity (fwd + bwd)
+# Mirrors tests/jax/test_fused_router.py::test_topk_bitmap_vs_bytemap.
```
ptrendx: This line is not really needed.

transformer_engine/common/fused_router/fused_score_for_moe_aux_loss.cu
```diff
@@ -127,8 +149,22 @@ __global__ void fused_score_for_moe_aux_loss_forward_kernel(const DataType *logi
...
-    for (int i = lane_id; i < topk; i += kThreadsPerWarp) {
-      routing_map[pos_offset + topk_indices[i]] = true;
+    if constexpr (!kIsBitmap) {
+      for (int i = lane_id; i < topk; i += kThreadsPerWarp) {
+        routing_map[pos_offset + topk_indices[i]] = 1;
+      }
+    } else {
+      for (int i = lane_id; i < topk; i += kThreadsPerWarp) {
+        int e = topk_indices[i];
+        atomicOr(&local_bitmap_words[e / 32], 1u << (e % 32));
```
ptrendx, line 159: What is the performance you are getting with this mode?

transformer_engine/common/fused_router/fused_topk_with_score_function.cu
```diff
@@ -229,21 +253,39 @@ __global__ void fused_topk_with_score_function_forward_kernel(
...
+        routing_map[pos_offset + topk_indices[i]] = 1;
+        probs[pos_offset + topk_indices[i]] = scaling_factor * topk_scores[i];
+      }
+    } else {
+      // BITMAP_U8: OR the selected-expert bit into the per-warp uint32 accumulator
+      // (shmem atomicOr handles same-word collisions across the topk lanes), then
+      // copy the bytemap-equivalent bytes out to the global uint8 bitmap row.
+      for (int i = lane_id; i < topk; i += kThreadsPerWarp) {
+        int e = topk_indices[i];
+        atomicOr(&local_bitmap_words[e / 32], 1u << (e % 32));
```
ptrendx, line 262: What is the performance of this? I'm pretty scared by this atomic or.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -129,6 +129,43 @@ void init_extension() {
...
 
+// Pybind11 registrations for the fused MoE router kernels. Split out of
+// PYBIND11_MODULE() to keep that function under the cpplint readability/fn_size
+// limit.
+void init_router_bindings(pybind11::module &m) {
+  pybind11::enum_<NVTERoutingMapFormat>(m, "NVTERoutingMapFormat", pybind11::module_local())
+      .value("BYTEMAP", NVTE_ROUTING_MAP_FORMAT_BYTEMAP)
+      .value("BITMAP_U8", NVTE_ROUTING_MAP_FORMAT_BITMAP_U8);
+  // routing_map_format is passed as int (not the enum) on the PyTorch hot
+  // path; see CLAUDE.md "CPU overhead in PyTorch wrappers".
```
ptrendx: LOL.

transformer_engine/pytorch/csrc/extensions/router.cpp
```diff
@@ -41,100 +65,166 @@ std::tuple<at::Tensor, at::Tensor, at::Tensor> fused_topk_with_score_function_fw
   int num_groups_value = num_groups.has_value() ? num_groups.value() : -1;
   float scaling_factor_value = scaling_factor.has_value() ? scaling_factor.value() : 1.0f;
 
-  // Construct the output tensor
-  at::Tensor probs =
-      at::empty({num_tokens, num_experts}, at::dtype(logits.scalar_type()).device(at::kCUDA));
+  // Allocate outputs at the user-facing N-D shape (probs / intermediate_output
+  // match logits exactly; routing_map's trailing dim depends on format).
+  // No caller-side .view() needed.
+  at::Tensor probs = at::empty(sizes, at::dtype(logits.scalar_type()).device(at::kCUDA));
```
ptrendx: Please clean those comments.

transformer_engine/pytorch/csrc/extensions/router.cpp
```diff
@@ -41,100 +65,166 @@ std::tuple<at::Tensor, at::Tensor, at::Tensor> fused_topk_with_score_function_fw
...
+  at::Tensor intermediate_output = at::empty(sizes, at::dtype(at::kFloat).device(at::kCUDA));
 
-  auto logits_cu = makeTransformerEngineTensor(logits);
-  auto probs_cu = makeTransformerEngineTensor(probs);
-  auto routing_map_cu = makeTransformerEngineTensor(routing_map);
-  auto intermediate_output_cu = makeTransformerEngineTensor(intermediate_output);
+  // Wrap with explicit 2D shape for the kernel — the common-layer NVTE_CHECKs
+  // expect {num_tokens, num_experts} (or {num_tokens, ceil(num_experts/8)} for
+  // the bitmap routing_map). This is ~100ns of std::vector + TensorWrapper
+  // construction vs ~1.5us per .view() in Python.
```
ptrendx: While I appreciate the benchmark, I don't think it is needed in this comment.

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -26,22 +26,27 @@ namespace transformer_engine::pytorch {
  * Router fusion
  **************************************************************************************************/
 
+// PyTorch-extension boundary uses int for routing_map_format (not the enum) to
+// keep the Python hot path free of pybind11 enum construction. The int is
+// cast to NVTERoutingMapFormat once inside each function. See CLAUDE.md
+// "CPU overhead in PyTorch wrappers".
```
ptrendx: ...

transformer_engine/pytorch/router.py
```diff
@@ -36,14 +87,14 @@ def forward(
...
-        tensor_shape = logits.shape
-        logits = logits.view(-1, tensor_shape[-1])
-        # Get the metadata of the viewed logits
-        num_tokens = logits.size(0)
-        num_experts = logits.size(1)
+        # The C++ extension handles N-D logits directly: it computes
+        # num_tokens = product of leading dims and num_experts = last dim,
+        # allocates probs / routing_map / intermediate_output at full N-D
+        # shape, and wraps tensors with an explicit 2D view *only* for the
+        # kernel call. No .view() is needed on either side of the FFI here.
```
ptrendx: This is not needed comment.

transformer_engine/common/fused_router/fused_topk_with_score_function.cu
tdophung: there was not any meaningful difference to bytemap. I did a 3 ways comparison (between original kernel time, after the change but bytemap, and after the change choosing bitmap) and the difference were < 6% between any pair on all the sizes tested. The larger the number of tokens, the less the overhead of atomicOr is visble. Here is a chart that sumarized the numbers for kernel-only duration: 
image


And here is a chart that summarized the measurement including the python side (I did not do bytemap post-PR for this one because it was similar in kernel times based on the above ^ chart). "fb" is forward + backward:
image

As you can see, the time actually decreased because we removed the .view() reshapes that was there before this PR started (in ORIG).

transformer_engine/common/fused_router/fused_score_for_moe_aux_loss.cu
tdophung: I did not benchmark this specific kernel, but I got numbers for the fused_topk_with_score_functions as pasted below i n reply to your other comment. I would assume it is similar, because the way we use the atomicOr was the same.

transformer_engine/pytorch/csrc/extensions/router.cpp
tdophung: Im removing the benchmark but still keeping the explanation on why we want to wrap with explicit 2D shape

tests/pytorch/test_fused_router.py
```diff
@@ -461,6 +462,139 @@ def test_fused_moe_aux_loss(dtype, num_tokens, num_experts, topk, expert_multipl
     torch.testing.assert_close(probs.grad, probs_clone.grad, atol=atol, rtol=rtol)
 
 
+def _bytemap_to_bitmap_u8(bytemap: torch.Tensor) -> torch.Tensor:
+    """Reference packer: bool[T, E] -> uint8[T, ceil(E/8)] LSB-first.
+
+    Matches numpy.packbits(..., bitorder='little'), which is what the JAX-side
+    parity test uses.
```
ptrendx: The comment about JAX is not needed.

ptrendx (dismissed):

ptrendx (approved):

PR #3011, opened 2026-05-19
[PyTorch] NVFP4 RHT cast-fusion: emit GEMM-swizzled scale factors directly

cael-ling:
# Description

Before this PR every NVFP4 RHT-cast-fusion quantize was followed by two standalone swizzle kernels (rowwise + columnwise) whose only job was to move scale factors into the layout cuBLAS LT consumes. The cast-fusion kernel already had a `kEnableSwizzleSFOutput` switch for that, but the framework never set the matching `with_gemm_swizzled_scales` flag on
NVFP4 outputs -- it was a `false` with a `TODO`. This PR wires it through and saves ~25 us per quantize on LLaMA-class shapes (**1.18x – 1.36x** on the `quant + swizzle` path that `te.Linear` runs).

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

**Kernel side** (`transformer_engine/common/hadamard_transform/`): 
- `row_cast_col_hadamard_transform_cast_fusion.cu` &
  `group_row_cast_col_hadamard_transform_cast_fusion.cu`: drive the
  existing `kEnableSwizzleSFOutput` template parameter from
  `output.with_gemm_swizzled_scales`. The grouped kernel additionally
  `NVTE_CHECK`s the flag is consistent across all tensors in a group
  (it honours a single boolean).
- The graph-safe grouped variant already had this wired correctly --
  no change.
  
**Framework side** (`transformer_engine/pytorch/csrc/`):
- New static helper `NVFP4Quantizer::is_eligible_for_rht_cast_fusion(rows, cols)`
  mirroring the dispatch-time eligibility check in
  `NVFP4Quantizer::quantize_impl` (rows%64==0 && cols%128==0 && SM100/110).
- `NVFP4Quantizer::create_tensor`, `NVFP4Quantizer::convert_and_update_tensor`,
  and `bulk_allocate_nvfp4_tensors` now set
  `with_gemm_swizzled_scales = optimize_for_gemm && with_rht && shape_eligible`.
  For the grouped allocator the flag is True only if *every* tensor in
  the group is eligible.
- Belt-and-suspenders `NVTE_CHECK(!out.with_gemm_swizzled_scales)` at
  the entry of `quantize_with_rht_unfused_helper`. The framework gate
  already keeps user code from tripping it; this only fires if a future
  low-level caller bypasses the gate.
  
## Performance
SM100a, bf16 input, rowwise + columnwise SF, RHT + post-RHT amax.
Per-quantize wall-clock median via `torch.utils.benchmark.Timer.blocked_autorange`.
`quant + swizzle` = `quantizer(x); tex.swizzle_scales_for_gemm_(t)` --
exactly what `te.Linear` runs before its GEMM.
| shape          | baseline | SUT      | saved   | speedup | note         |
| -------------- | -------- | -------- | ------- | ------- | ------------ |
| (8192, 5120)   | 108.6 us | 81.9 us  | 26.6 us | 1.33x   | eligible     |
| (8192, 10240)  | 107.8 us | 90.2 us  | 17.5 us | 1.19x   | eligible     |
| (8192, 2560)   | 107.7 us | 79.9 us  | 27.8 us | 1.35x   | eligible     |
| (8192, 11328)  | 236.3 us | 236.3 us |  0.0 us | 1.00x   | *ineligible* |
| (8192, 3584)   | 106.0 us | 78.6 us  | 27.4 us | 1.35x   | eligible     |
| (5120, 8192)   | 101.2 us | 76.0 us  | 25.3 us | 1.33x   | eligible     |
| (10240, 8192)  | 107.8 us | 90.4 us  | 17.4 us | 1.19x   | eligible     |
| (2560, 8192)   | 101.4 us | 74.9 us  | 26.4 us | 1.35x   | eligible     |
| (11328, 8192)  | 114.4 us | 93.2 us  | 21.2 us | 1.23x   | eligible     |
| (3584, 8192)   | 101.6 us | 74.9 us  | 26.7 us | 1.36x   | eligible     |
| (4096, 16384)  | 100.2 us | 75.0 us  | 25.2 us | 1.34x   | eligible     |
| (14336, 16384) | 232.1 us | 197.5 us | 34.6 us | 1.18x   | eligible     |

- 11/12 shapes get **1.18x – 1.36x** on the `quant + swizzle` path.
- The single ineligible shape `(8192, 11328)` shows **1.00x** as expected;
  the gate clamped, the unfused fallback ran, and the result is byte-
  identical to baseline (no regression, no crash).
- `quant_only` is unchanged on all shapes within noise -- writing
  swizzled SF inside the cast-fusion kernel is essentially free; the
  entire win comes from eliminating the standalone swizzle pass.
Repro: `benchmarks/benchmark_rht_cast_swizzle_fusion.py` (also has a
`--profile` mode for ncu / nsys).

# Checklist:


transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -730,7 +730,27 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>, bool> bulk_alloc
...
+  for (size_t i = 0; i < num_tensors; ++i) {
+    const auto [rows, cols] = get_2d_dims(shape_list[i]);
+    if (!NVFP4Quantizer::is_eligible_for_rht_cast_fusion(rows, cols)) {
+      all_tensors_rht_cast_fusion_eligible = false;
+      break;
+    }
+  }
+  const bool with_gemm_swizzled_scales = quantizer_cpp_list[0]->optimize_for_gemm &&
+                                         quantizer_cpp_list[0]->with_rht &&
+                                         all_tensors_rht_cast_fusion_eligible;
```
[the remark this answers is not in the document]
cael-ling: with_gemm_swizzled_scales was derived from quantizer_cpp_list[0]->optimize_for_gemm / with_rht without checking that other quantizers in the group agreed; if any later quantizer had a different value, its tensors would be silently allocated with the wrong SF layout. 

Following the precedent of the split-quantize path at line 1276
(// Assume all quantizers have identical config), this commit:
  * adds an explicit comment block calling out the group-wide
    identical-config assumption and which fields this PR enforces
    vs. which are pre-existing;
  * adds an NVTE_CHECK loop enforcing identical optimize_for_gemm
    and with_rht across the group (the two fields the
    with_gemm_swizzled_scales gate depends on), with error messages
    that print the offending tensor index and the disagreeing values;
  * extracts the [0] reads into group_optimize_for_gemm and
    group_with_rht locals so the same value feeds both the check
    and the gate.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -719,7 +719,17 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>, bool> bulk_alloc
...
+  // (see split_quantize_nvfp4_impl in this file), we assume all
+  // quantizers in the group share an identical config and read
+  // group-wide flags from quantizer_cpp_list[0]. The grouped RHT
+  // cast-fusion kernel honours a single with_gemm_swizzled_scales
+  // boolean across the whole group, so optimize_for_gemm and with_rht
+  // must in particular agree across all quantizers; the NVTE_CHECK
+  // loop below enforces that for the fields the swizzled-SF gate
+  // depends on. (The other group-wide reads from [0] -- rowwise_usage,
+  // row_scaled_nvfp4, columnwise_usage, scaling_mode, dtype -- are
+  // pre-existing assumptions and out of scope for this PR.)
```
ptrendx: I'm not super happy about the style of those comments - they reference multiple other files, and while right now the comment matches the reality, it will easily drift. We should concentrate on commenting the invariants and assumptions needed for this file only.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -730,7 +740,40 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>, bool> bulk_alloc
...
+  // Only the RHT cast-fusion quant kernel supports direct swizzled SF
+  // emission. Other NVFP4 quant kernels (e.g. nvte_quantize_v2 ->
+  // quantize_nvfp4.cuh, quantize_transpose_nvfp4.cuh) NVTE_CHECK reject
+  // a swizzled-flagged output, so we gate on with_rht to avoid silent
+  // data corruption / hard aborts on non-RHT paths. Additionally we
+  // require *all* tensors in the group to be shape-eligible for RHT
+  // cast-fusion, because the grouped kernel honours a single boolean
+  // and the unfused fallback rejects swizzled output (see NVTE_CHECK
+  // at group_row_cast_col_hadamard_transform_cast_fusion.cu and
+  // quantize_with_rht_unfused_helper).
```
ptrendx: Same.

transformer_engine/pytorch/csrc/common.h
```diff
@@ -371,6 +371,19 @@ class NVFP4Quantizer : public Quantizer {
...
+  /*! @brief Whether a 2D shape (rows, cols) is eligible for the
+   *  NVFP4 RHT cast-fusion kernel.
+   *
+   *  Matches the dispatch logic in NVFP4Quantizer::quantize_impl.
+   *  The dtype check (BF16) is implicit -- with_rht=True requires
+   *  BF16 input by construction, so callers gate on with_rht first.
+   *  When false, the dispatch falls back to quantize_with_rht_unfused
+   *  which cannot emit GEMM-swizzled SF; framework gates that opt
+   *  into with_gemm_swizzled_scales must therefore also check this
+   *  to avoid mismatched-flag aborts in the fallback path.
```
ptrendx: Again, this is mostly talking about the internal implementation choices rather than what that function actually does (which is covered by the first sentence).

transformer_engine/pytorch/csrc/common.h
```diff
@@ -371,6 +371,19 @@ class NVFP4Quantizer : public Quantizer {
...
+   */
+  static bool is_eligible_for_rht_cast_fusion(size_t rows, size_t cols);
```
ptrendx: Shouldn't it take arbitrary shape rather than assuming it will be 2D?

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1760,18 +1760,38 @@ void NVFP4Quantizer::set_quantization_params(TensorWrapper* tensor) const {
                               columnwise_data.shape);
 }
 
+bool NVFP4Quantizer::is_eligible_for_rht_cast_fusion(size_t rows, size_t cols) {
+  // Must mirror the eligibility check in NVFP4Quantizer::quantize_impl
+  // (search for "eligible_for_rht_cast_fusion" in this file). The dtype
+  // check (BF16) is implicit: with_rht is only valid for BF16 input by
+  // construction.
```
ptrendx: Why does it have to mirror the check in that other function? Considering that both of these functions are in the same file and in the same class, can't we just call one from the other to keep a single source of truth?

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -2205,6 +2229,18 @@ void NVFP4Quantizer::quantize_with_rht_unfused_helper(
     QuantizationConfigWrapper& quant_config, QuantizationConfigWrapper& quant_config_columnwise,
     cudaStream_t stream) {
   // only triggered for irregular shapes where RHT cast fusion kernel is not eligible
+  // The unfused fallback dispatches to nvte_quantize_v2 / nvte_hadamard_transform,
+  // neither of which supports emitting SF in the GEMM-swizzled layout (their
+  // backing kernels NVTE_CHECK reject swizzled-flagged output). Surface a clean
+  // error here instead of letting it abort deep inside the kernel with an
+  // opaque message. JAX hard-asserts eligibility upfront; PyTorch matches that
```
ptrendx: Why do we mention JAX in the PyTorch source files?

ptrendx:
Please also handle the convert_and_update_tensor path since it also needs changes.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -730,7 +740,40 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>, bool> bulk_alloc
...
+               quantizer_cpp_list[i]->optimize_for_gemm, ").");
+    NVTE_CHECK(quantizer_cpp_list[i]->with_rht == group_with_rht,
+               "NVFP4 bulk allocation requires all quantizers in the group to share "
+               "the same with_rht value (tensor 0=",
+               group_with_rht, ", tensor ", i, "=", quantizer_cpp_list[i]->with_rht, ").");
+  }
+  bool all_tensors_rht_cast_fusion_eligible = true;
+  for (size_t i = 0; i < num_tensors; ++i) {
+    const auto [rows, cols] = get_2d_dims(shape_list[i]);
+    if (!NVFP4Quantizer::is_eligible_for_rht_cast_fusion(rows, cols)) {
```
ptrendx: The grouped kernel that supports the swizzle will only run for rows being divisible by 128, but this function will  allow tensors divisible by 64.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1760,18 +1760,38 @@ void NVFP4Quantizer::set_quantization_params(TensorWrapper* tensor) const {
                               columnwise_data.shape);
 }
 
+bool NVFP4Quantizer::is_eligible_for_rht_cast_fusion(size_t rows, size_t cols) {
+  // Must mirror the eligibility check in NVFP4Quantizer::quantize_impl
+  // (search for "eligible_for_rht_cast_fusion" in this file). The dtype
+  // check (BF16) is implicit: with_rht is only valid for BF16 input by
+  // construction.
+  return rows % 64 == 0 && cols % 128 == 0 && transformer_engine::cuda::sm_arch() >= 100 &&
```
ptrendx: Why is the rows % 64 == 0 a requirement here rather than rows % 128 == 0?

transformer_engine/pytorch/csrc/extensions/cast.cpp
cael-ling: New commit removed the prose about dispatch internals and caller responsibilities.

transformer_engine/pytorch/csrc/extensions/cast.cpp
cael-ling: New commit removed the prose about dispatch internals and caller responsibilities.

transformer_engine/pytorch/csrc/common.h
cael-ling: New commit removed the prose about dispatch internals and caller responsibilities.

transformer_engine/pytorch/csrc/common.h
cael-ling: Good point. Changed the signature to take the full tensor shape (`const std::vector<size_t>& shape`) and moved the `get_2d_dims(...)` flatten inside the function. All four call sites (`create_tensor`, `convert_and_update_tensor`, `quantize_impl`, and the grouped path in `cast.cpp`) now pass the shape directly without pre-flattening. The bulk loop in `cast.cpp` also no longer calls `get_2d_dims` per iteration since the function takes care of it.

transformer_engine/pytorch/csrc/quantizer.cpp
cael-ling: Correct, the proper fix is to call one from the other. `quantize_impl` now delegates the shape/arch predicate to `NVFP4Quantizer::is_eligible_for_rht_cast_fusion(...)` instead of re-inlining the same check. The BF16 dtype guard stays as an explicit `&&` at the call site because it's specific to `quantize_impl` (the allocation callers don't have an input tensor to check). I also replaced the hand-rolled `rows = product(input.shape[:-1])` loop with `get_2d_dims(input.shape())` so the flattening rule isn't duplicated either. The shape/arch eligibility now has a single source of truth.

transformer_engine/pytorch/csrc/quantizer.cpp
cael-ling: New commit dropped the JAX reference and the surrounding narration. The remaining 2-line comment just explains why this `NVTE_CHECK` is here.

transformer_engine/pytorch/csrc/quantizer.cpp
cael-ling: The 64 here is correct for the single-tensor cast-fusion kernel — its entry check is `NVTE_CHECK(M % 64 == 0, ...)` at `row_cast_col_hadamard_transform_cast_fusion.cu:1161`. The 128 you're thinking of is the grouped kernel's stricter requirement at `graph_safe_group_row_cast_col_hadamard_transform_cast_fusion.cu:1385`.

transformer_engine/pytorch/csrc/extensions/cast.cpp
cael-ling: Good catch — this was a real bug, not just an over-permissive style.

Before this fix, `is_eligible_for_rht_cast_fusion(shape)` used a single row-alignment constraint of `rows % 64 == 0` (the single-tensor RHT cast-fusion kernel's entry check at `row_cast_col_hadamard_transform_cast_fusion.cu:1161`). The bulk-allocation path in `cast.cpp` was calling this same lax check, so shapes like rows in {64, 192, 320, ...} — all satisfying `% 64 == 0` — would pass eligibility, get `with_gemm_swizzled_scales=True`, and then hard-abort inside the grouped kernel whose entry asserts `first_logical_dim % 128 == 0`
(`graph_safe_group_row_cast_col_hadamard_transform_cast_fusion.cu:1385`).

The fix adds a `for_grouped_kernel` parameter on `is_eligible_for_rht_cast_fusion` so callers select the constraint
that matches the kernel they will actually invoke:
- `false` (default): `rows % 64 == 0`, single-tensor kernel
- `true`: `rows % 128 == 0`, grouped kernel

The bulk-allocation caller in `cast.cpp` passes `/*for_grouped_kernel=*/true`; the three single-tensor callers
(`create_tensor`, `convert_and_update_tensor`, `quantize_impl`) keep the default `false`. Shapes with rows in {64, 192, 320, ...} now correctly fail the grouped-path eligibility and fall back to the unfused path instead of reaching the grouped kernel.

cael-ling:
> Please also handle the convert_and_update_tensor path since it also needs changes.

Done. Both `create_tensor` and `convert_and_update_tensor` now have the same 2-line comment on the gating; removed the previous "See NVFP4Quantizer::create_tensor for the rationale" cross-reference. I also trimmed `create_tensor`'s long rationale block (which referenced specific `.cu`/`.cuh` filenames and `quantize_with_rht_unfused`'s internal behavior) in the same pass, so the two functions are consistent.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -2084,6 +2091,12 @@ std::pair<TensorWrapper, py::object> NVFP4Quantizer::convert_and_update_tensor(
   }
 
   const auto [flat_first_dim, flat_last_dim] = get_2d_dims(shape);
+
+  // Swizzled SF is only valid when the RHT cast-fusion path runs;
+  // other quantize paths reject it.
+  const bool with_gemm_swizzled_scales = this->optimize_for_gemm && this->with_rht &&
```
ptrendx, line 2140: This is then set for the out_cpp TensorWrapper (at the end of this function), but not in the actual Python object. See handing of this in the MXFP8 quantizer:
```cpp
  tensor.attr("_with_gemm_swizzled_scales") = with_gemm_swizzled_scales;
```

cael-ling: Good catch — fixed in the latest commit (pushed), mirroring the MXFP8 quantizer

ptrendx (dismissed):

ptrendx (dismissed):

ptrendx (dismissed):

ptrendx (approved):

PR #3017, opened 2026-05-20
[Common] NVTETensor peer-handle annotation + nccl_comm backend

phu0ngng:
# Description

 - Add a generic peer-handle annotation on `NVTETensor` so consumers can issue one-sided RMA against a tensor's storage without owning the underlying resource.
  - Register NCCL symmetric-memory windows as the first concrete backend behind this annotation (`nccl_comm.h`).
  - Gate `newton_schulz` on `NVTE_WITH_CUSOLVERMP` to unblock builds without cuSOLVERMp.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

  - `common/include/transformer_engine/comm_handle.h`: payload-agnostic C API — kind tag (`NVTEPeerHandleKind`), `nvte_tensor_peer_handle_kind`, `nvte_tensor_detach_peer_handle`. Reserves slots for future backends (NVSHMEM, CUDA-IPC, UCX).
  - `common/include/transformer_engine/nccl_comm.h`: NCCL-specific setter/getter — `nvte_tensor_attach_nccl_window(t, window, offset)` / `nvte_tensor_nccl_window(t, &window, &offset)`. Window is borrowed; tensor never owns it.
  - `common/common.h`: extend `Tensor` with `peer_handle_kind / peer_handle_data / peer_handle_offset` fields (ignored by paths that don't issue cross-rank ops).
  - `common/comm_handle.cpp`: implementations.
  - `common/CMakeLists.txt`: build the new `comm_handle.cpp`.
  - `pytorch/csrc/extensions/newton_schulz.cpp` + `build_tools/pytorch.py`: compile/include newton_schulz only when `NVTE_WITH_CUSOLVERMP` is set.
  - Bump `3rdparty/cudnn-frontend` and minor doc tweak under `examples/pytorch/comm_gemm_overlap/README.md`.
  
# Testing 
 - Covered by the cpp-distributed EP tests in the follow-up branch (`phuong/ep-2-common-cpp`), which exercise the NCCL-window attach/detach path end-to-end.
  - Standard TE build sanity (`NVTE_CUDA_ARCHS="90;100" NVTE_FRAMEWORK=jax pip install --no-build-isolation -e .`).

# Checklist:

### DIFF
--- a/build_tools/pytorch.py
+++ b/build_tools/pytorch.py
@@ -76,6 +76,13 @@ def setup_pytorch_extension(
 
     setup_mpi_flags(include_dirs, cxx_flags)
 
+    # Mirror the cuSOLVERMp gate. newton_schulz.cpp is conditionally compiled
+    # in the common lib; the pytorch ext glob pulls the same source so it must
+    # see the same define, otherwise the pybind layer refers to undefined
+    # cusolvermp_ctx_* symbols.
+    if bool(int(os.getenv("NVTE_WITH_CUSOLVERMP", "0"))):
+        cxx_flags.append("-DNVTE_WITH_CUSOLVERMP")
+
     library_dirs = []
     libraries = []
     if bool(int(os.getenv("NVTE_ENABLE_NVSHMEM", 0))):
--- a/transformer_engine/common/CMakeLists.txt
+++ b/transformer_engine/common/CMakeLists.txt
@@ -147,6 +147,7 @@ set(transformer_engine_cuda_sources)
 set(transformer_engine_cuda_arch_specific_sources)
 
 list(APPEND transformer_engine_cpp_sources
+     comm_handle.cpp
      cudnn_utils.cpp
      transformer_engine.cpp
      fused_attn/fused_attn.cpp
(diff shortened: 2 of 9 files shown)


build_tools/pytorch.py
```diff
@@ -76,6 +76,13 @@ def setup_pytorch_extension(
 
     setup_mpi_flags(include_dirs, cxx_flags)
 
+    # Mirror the cuSOLVERMp gate. newton_schulz.cpp is conditionally compiled
+    # in the common lib; the pytorch ext glob pulls the same source so it must
+    # see the same define, otherwise the pybind layer refers to undefined
+    # cusolvermp_ctx_* symbols.
+    if bool(int(os.getenv("NVTE_WITH_CUSOLVERMP", "0"))):
+        cxx_flags.append("-DNVTE_WITH_CUSOLVERMP")
+
```
ptrendx, line 85: ? What is the error you are seeing without that? At least from the cursory look at those sources I don't see how the pytorch files would be affected here. Is the comments implying that the PyTorch compilation is somehow taking the common files and compiling them again?

ptrendx:
Ok, a general question - this looks like a thing that is basically not compatible with everything else in nvtetensor, so why do we even want to use NVTETensor at all? If you just need a C struct that would hold the pointer and offset then we should make that separate. It is hard to review this without seeing what the concrete implementation would actually use this mechanism for.

phu0ngng:
Close this PR in favor of having a separate data structure for the NCCL Window instead of attaching it to the NVTETensor.

PR #3018, opened 2026-05-21
[Common] Fix fused MoE aux loss for sequence aux loss

harryzhou2000:
## Summary
- Allow fused MoE aux loss inputs whose column count is a multiple of the logical expert count.
- Document that sequence aux loss batches independent sequences along the expert dimension.
- Extend the PyTorch fused router test to cover expanded expert columns.

## Details
Sequence aux loss reshapes router scores from `[seq_length * batch_size, num_experts]` to `[seq_length, batch_size * num_experts]` before calling the same aux-loss helper. The logical `num_experts` is still used in the loss coefficient, while `probs.shape[1]` is the actual reduction width. The previous assertion required these values to be equal and rejected valid sequence aux-loss inputs when `batch_size > 1`.

## Testing
- PyTorch fused router UT on B200: `912 passed, 120 skipped, 3 warnings in 24.52s`

### DIFF
--- a/tests/pytorch/test_fused_router.py
+++ b/tests/pytorch/test_fused_router.py
@@ -451,7 +451,7 @@ def test_fused_moe_aux_loss(dtype, num_tokens, num_experts, topk, expert_multipl
         coeff=coeff,
     )
 
-    atol, rtol = _get_tolerances(dtype, num_experts)
+    atol, rtol = _get_tolerances(dtype, num_cols)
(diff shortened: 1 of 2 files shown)


denera (approved):
LGTM. This separation between `num_cols` and `num_experts` was a point of confusion when I re-wrote this kernel. This PR clarifies it especially in the test parameterization. Thanks!

PR #648, opened 2024-01-31
FP8 Support for MCore MoE

Victarry:
Add FP8 support for MoE in MCore.

Related MR in MCore:
https://gitlab-master.nvidia.com/ADLR/megatron-lm/-/merge_requests/1089

Implementation details:
- Add `rng_tracker_name` for initialize with EP
- Add if statement in kernel to handle zeros tokens passed to an expert.

PR #3019, opened 2026-05-21
fix(grouped_linear): handle all-zero-token forward and backward

jubick1337:
When a rank receives zero tokens for every local expert (`sum(m_splits) == 0`), `_GroupedLinear.forward` previously fed an empty input into `general_grouped_gemm`, which can crash or hang in the underlying cuBLAS grouped GEMM. Skipping the call entirely would also remove the weights from the autograd graph on that rank, desynchronizing DDP / EP collectives across ranks that did have tokens.

Short-circuit `_GroupedLinear.forward` when `sum(m_splits) == 0`: produce an empty output of the correct shape and dtype, register the weights and biases via `ctx.save_for_backward` so DDP sees matching shapes across ranks, and record an `empty_input` flag on ctx.

Mirror this in `_GroupedLinear.backward`: when `ctx.empty_input` is set, synthesize correctly-shaped zero gradients (dgrad, wgrad, bgrad) without touching the quantization / GEMM machinery. The `fuse_wgrad_accumulation`
+ Mcore-DDP path is preserved by flipping `grad_added_to_main_grad` on weights that opt into it, matching the existing
`handle_custom_ddp_from_mcore` semantics.

Add an `"all"` value to the existing `empty_split` parametrization in `test_sanity_grouped_linear`, exercising the zero-token forward and backward path across the same dtype / FP8 recipe / single_param matrix already in use, and assert that every weight / bias gradient comes back with the correct full shape (zeros are allowed but `None` is not, since that would desynchronize DDP / EP allreduce).

The `te.ops.GroupedLinear` (basic-ops API) was verified to already handle the all-zero-token case correctly without changes, so this PR is scoped to `_GroupedLinear` in `module/grouped_linear.py`.

Reported in NVIDIA/Megatron-LM#4851; per @Victarry's review, the fix belongs in TE (analogous to TE PR #648 for the FP8 path).

# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/test_sanity.py
+++ b/tests/pytorch/test_sanity.py
@@ -566,8 +566,8 @@ def test_sanity_grouped_linear(
     # Small batch size used to catch bug from https://github.com/NVIDIA/TransformerEngine/pull/1527.
     bs = bs * 16
     # The "all" empty_split exercises the path where every local expert
-    # receives zero tokens, e.g. a Megatron-LM MoE rank that didn't receive
-    # any locally-routed tokens for a microbatch. In that case num_tokens
+    # receives zero tokens (e.g. a Megatron-LM MoE rank that didn't receive
+    # any locally-routed tokens for a microbatch). In that case num_tokens
     # is 0; for the other empty_split values exactly one of num_gemms splits
     # is empty so num_tokens covers the remaining (num_gemms - 1) groups.
     if empty_split == "all":
--- a/tests/pytorch/test_sanity.py
+++ b/tests/pytorch/test_sanity.py
@@ -547,7 +547,7 @@ def test_sanity_linear_with_zero_tokens(
 @pytest.mark.parametrize("fp8_model_params", all_boolean)
 @pytest.mark.parametrize("use_bias", all_boolean)
 @pytest.mark.parametrize("single_param", all_boolean)
-@pytest.mark.parametrize("empty_split", ["first", "last", "middle"])
+@pytest.mark.parametrize("empty_split", ["first", "last", "middle", "all"])
 @pytest.mark.parametrize("num_gemms", [4])
 def test_sanity_grouped_linear(
     dtype,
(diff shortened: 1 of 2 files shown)


ptrendx (dismissed):
There are a few issues with this, but the most important one is that we cannot really do that check. We are moving towards the device-based m_splits in order to have CUDA graphs compatibility and the comparison `sum(m_splits) == 0` is not going to work in that context. We need to make sure that cuBLAS works in that case too.

jubick1337:
> There are a few issues with this, but the most important one is that we cannot really do that check. We are moving towards the device-based m_splits in order to have CUDA graphs compatibility and the comparison `sum(m_splits) == 0` is not going to work in that context. We need to make sure that cuBLAS works in that case too.

Thanks @ptrendx!
Dropping `sum(m_splits) == 0`.

Some context: we hit this on a Megatron-LM training run (32n H100 MoE SFT) where some EP ranks got zero tokens for a microbatch and the whole job deadlocked at the next NCCL allreduce. We worked around it in Megatron first (#4851 there), then @Victarry pointed us at TE as the proper place to fix.

But re-running my coverage on current TE main without the Python check, the all-zero case already works end-to-end on bf16, fp32, and every supported FP8 recipe my Blackwell can execute (2568 pass, 0 fail). The per-group early return at gemm.cpp:509 + the empty nvte_multi_tensor_gemm handling seem to cover it.

So I'm reducing this PR to just the regression test: new empty_split="all" parametrize value + a check that weight grads come back with full shape (not None) so DDP/EP allreduce doesn't desync. Keeps this from regressing without adding any code surface.

What do you think?

ptrendx:
Sounds good. The changes look good to me, the only point I'm not sure about is not testing the single parameter path. Adding @ksivaman to look at that.

jubick1337:
@ksivaman what's your opinion?

PR #3033, opened 2026-05-22
Add MXFP8 attention unit test with linear and rope layers

layalir:
Add a DSv3-shaped MXFP8 attention unit test covering the training path:

- Adds MLA RoPE utilities for the DSv3 671B attention shape.
- Adds an end-to-end MXFP8 path: `Linear(QKV) -> MLA RoPE -> DotProductAttention -> Linear(out)`.
- Exercises MXFP8 forward and backward through TE's real `DotProductAttention` wrapper.
- Runs BF16 reference comparison by default.
- Runs the performance benchmark by default and reports fprop and bprop timing separately from the same benchmark collection.

## Validation

Local checks:

- `python -m py_compile tests/pytorch/attention/test_linear_mxfp8_attention.py tests/pytorch/attention/mla_rope_utils.py`
- `git diff --check`

GB300 dlcluster validation:

- Job: `1062811`
- GPU: NVIDIA GB300
- CUDA capability: `(10, 3)`
- cuDNN: `(9, 21, 1)`
- MXFP8 available: `(True, '')`
- Command: `python -m pytest tests/pytorch/attention/test_linear_mxfp8_attention.py -v -s`
- Result: `3 passed`

Perf output:

```text
[PERF] b=1 s=4096:
  BF16 fprop:  7.219 ms  (567397 tok/s)
  BF16 bprop:  15.179 ms  (269844 tok/s)
  MXFP8 fprop: 4.718 ms  (868181 tok/s)
  MXFP8 bprop: 9.215 ms  (444492 tok/s)
  Fprop speedup: 1.53x
  Bprop speedup: 1.65x
```


cyanguwa:
Thanks for the contribution! Could you please:
- fix the DCO (there are instructions in the DCO link)
- address Greptile comments
- add test_linear_mxfp8_attention.py to qa/L0_pytorch_unittest/test.sh, similar to [this](https://github.com/NVIDIA/TransformerEngine/blob/5f1eaffde87336ef3f1bd877044a7159e02e569d/qa/L0_pytorch_unittest/test.sh#L51-L52)

tests/pytorch/attention/test_linear_mxfp8_attention.py
```diff
@@ -0,0 +1,378 @@
...
+        qkv_dtype=torch.float8_e4m3fn,
+        qkv_layout="sbhd_sbhd_sbhd",
+        fp8=True,
+        fp8_meta=fp8_meta,
+        is_training=True,
+        deterministic=_DETERMINISTIC,
+    )
+    flash_attn_supported, fused_attn_supported_fp8, _ = fp8_backends
+    if flash_attn_supported + fused_attn_supported_fp8 < 1:
+        pytest.skip("No FP8 attention backend available for DSv3 MLA shape.")
```
cyanguwa: I think we should check "if not fused_attn_supported_fp8" here. FlashAttention doesn't support MXFP8, so it's kind of irrelevant here.

tests/pytorch/attention/test_linear_mxfp8_attention.py
```diff
@@ -0,0 +1,378 @@
...
+        is_training=True,
+        deterministic=_DETERMINISTIC,
+    )
+    if sum(bf16_backends) < 1:
+        pytest.skip("No BF16 attention backend available for DSv3 MLA shape.")
+
+    _attention_backends["backend_selection_requires_update"] = True
+
+
+def _split_qkv(qkv: torch.Tensor) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
```
cyanguwa: Does this work for MXFP8 tensors? Should we do qkv.to() first before splitting?

tests/pytorch/attention/test_linear_mxfp8_attention.py
```diff
@@ -0,0 +1,378 @@
...
+        assert not torch.isinf(out_mxfp8).any(), "MXFP8 output contains Inf"
+        assert out_mxfp8.float().abs().max() > 0, "MXFP8 output is all zeros"
+
+        max_abs_qkv, rms_qkv = _compute_errors(qkv_bf16, qkv_mxfp8)
+        print(f"\n[QKV] b={batch_size} s={seq_len}: max_abs={max_abs_qkv:.6f}  rms={rms_qkv:.6f}")
+        torch.testing.assert_close(
+            qkv_mxfp8,
+            qkv_bf16,
+            atol=2.0,
+            rtol=0.5,
```
cyanguwa: These tolerances are a bit too high, aren't they? For regular MXFP8 tests, I've been using atol = 5e-1, rtol = 5e-2. Would they work for these tests?

layalir: I tried the suggested atol=5e-1, rtol=5e-2 on GB300. It fails only the raw QKV projection comparison: max_abs=0.632812, rms=0.110919. I think the difference from the regular MXFP8 attention tests is that those tolerances are used with the FP8 comparison helper, which tolerates per-element assert_close failures and gates on RMSE. Here we were applying assert_close directly to the full QKV projection tensor, which has ~268M elements, so the MXFP8 quantization tail is enough to exceed 0.5 even though RMS remains low. I’ll keep rtol=5e-2, tighten QKV atol to the smallest passing value above the observed tail, and keep the printed max/RMS so this remains visible.

tests/pytorch/attention/test_linear_mxfp8_attention.py
layalir: I removed the packed QKV split path entirely. The test now uses separate Q and KV LayerNormLinear modules and feeds quantized tensors into MXFP8 DPA which matches the real workload.

tests/pytorch/attention/test_linear_mxfp8_attention.py
```diff
@@ -0,0 +1,418 @@
...
+    q, k, v = apply_mla_rope(q, k, v)
+
+    with te.fp8_autocast(enabled=True, fp8_recipe=recipe):
+        attn_out = dpa(q, k, v, qkv_format="sbhd")
+
+    with te.fp8_autocast(enabled=True, fp8_recipe=recipe):
+        out = out_linear(
+            attn_out.view(x.shape[0], x.shape[1], HIDDEN_SIZE),
+            is_first_microbatch=is_first_microbatch,
+        )
```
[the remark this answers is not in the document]
layalir: I addressed this by collapsing the MXFP8 forward path into one enclosing fp8_autocast scope.

tests/pytorch/attention/test_linear_mxfp8_attention.py
```diff
@@ -0,0 +1,418 @@
...
+            dtype=torch.bfloat16,
+            device="cuda",
+            requires_grad=True,
+        )
+
+        with torch.no_grad():
+            _run_forward_mxfp8(mxfp8_modules, x, fp8_recipe, is_first_microbatch=True)
+
+        mxfp8_fprop_ms, mxfp8_bprop_ms = _benchmark_training_step(
+            _run_forward_mxfp8, mxfp8_modules, x, fp8_recipe, False
```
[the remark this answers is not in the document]
layalir: For the performance test, I removed the pre-cache call with is_first_microbatch=True and stopped timing with is_first_microbatch=False

tests/pytorch/attention/mla_rope_utils.py
```diff
@@ -0,0 +1,546 @@
...
+        sin_left = tl.load(SIN + token_idx * emb_dim + tl.arange(0, emb_dim // 2))
+        cos_right = tl.load(COS + token_idx * emb_dim + emb_dim // 2 + tl.arange(0, emb_dim // 2))
+        sin_right = tl.load(SIN + token_idx * emb_dim + emb_dim // 2 + tl.arange(0, emb_dim // 2))
+        cos_left = cos_left.expand_dims(0).broadcast_to(BLOCK_H, emb_dim // 2)
+        sin_left = sin_left.expand_dims(0).broadcast_to(BLOCK_H, emb_dim // 2)
+        cos_right = cos_right.expand_dims(0).broadcast_to(BLOCK_H, emb_dim // 2)
+        sin_right = sin_right.expand_dims(0).broadcast_to(BLOCK_H, emb_dim // 2)
+        Q = Q + pid_m * stride_x_seq + pid_head * BLOCK_H * stride_x_nheads
+        x_off = tl.arange(0, BLOCK_H)[:, None] * stride_x_nheads + qk_head_dim
+        mask = x_off < head_num * stride_x_nheads
```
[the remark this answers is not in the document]
layalir: I changed the Triton masks to use absolute head offsets instead of only intra-block offsets.

cyanguwa (approved):

PR #3034, opened 2026-05-22
Expert Parallelism: common C API + NCCL EP backend

phu0ngng:
## Summary

First PR in the TE Expert Parallelism (EP) series. Lands the common C API and NCCL EP backend that later framework PRs (PyTorch, JAX) build on. No Python bindings yet — common-lib foundation plus build wiring only. Build/load works on any arch; SM and NCCL version gates fire at runtime.

Every network-bound payload tensor takes an **optional** `NVTECommWindow`. When the window is provided, the backend uses NCCL EP's symmetric-memory zero-copy path, which skips the D2D Memcpy from the user buffers to the Symmetric Staging Buffers.
 
  ## Implementation

  ### Public C API (`transformer_engine/common/include/transformer_engine/{ep.h,comm_window.h}`)

Types: `NVTEEpGroupConfig`, `NVTEEpLayerConfig`, `NVTEEpHandle`, `NVTECommWindow` (side-band `{ncclWindow_t window, size_t offset}`; NCCL peer handles are not carried on `NVTETensor`).

Lifecycle (host-only, eager):

```c
void     nvte_ep_initialize(void* ep_comm, NVTEEpGroupConfig group_config);
void     nvte_ep_shutdown(void);

uint64_t nvte_ep_register_layer(NVTEEpLayerConfig layer_config, size_t* handle_mem_size);
```

- `nvte_ep_initialize` — borrow an external `ncclComm_t` for the EP sub-group and init the singleton backend.
- `nvte_ep_shutdown` — tear down the backend; idempotent; does not destroy `ep_comm`.

- `nvte_ep_register_layer` — reserve a `handle_id` for a layer config and report the `handle_mem` buffer size the caller must allocate. The pair `{id, mem}` becomes the per-step `NVTEEpHandle`.

Per-step (allocation-free, CUDA-graph capturable)
```c
void nvte_ep_prepare(NVTEEpHandle handle, NVTETensor topk_idx, NVTETensor token_counts,
                     size_t dispatch_output_per_expert_alignment, cudaStream_t stream);

void nvte_ep_dispatch(NVTEEpHandle handle, NVTETensor topk_idx,
                      NVTETensor tokens, NVTECommWindow tokens_win,
                      NVTETensor topk_weights, NVTECommWindow topk_weights_win,
                      NVTETensor recv_tokens, NVTECommWindow recv_tokens_win,
                      NVTETensor recv_topk_weights,  NVTECommWindow recv_topk_weights_win,
                      cudaStream_t stream);

void nvte_ep_combine(NVTEEpHandle handle, NVTETensor expert_out, NVTECommWindow expert_out_win,
                     NVTETensor result, cudaStream_t stream);

void nvte_ep_dispatch_bwd(NVTEEpHandle handle, NVTETensor grad, NVTECommWindow grad_win,
                          NVTETensor g_recv_topk_weights, NVTECommWindow g_recv_topk_weights_win,
                          NVTETensor grad_tokens, NVTETensor grad_topk_weights, cudaStream_t stream);

void nvte_ep_combine_bwd(NVTEEpHandle handle, NVTETensor grad, NVTECommWindow grad_win,
                         NVTETensor grad_expert_out, NVTECommWindow grad_expert_out_win,
                         cudaStream_t stream);
```
 - `nvte_ep_prepare` — all-gather the routing map and write routing maps to `handle.mem`.
 - `nvte_ep_dispatch` — scatter tokens and routing weights from source ranks to expert ranks. `tokens`, `topk_weights`, `recv_tokens`, `recv_topk_weights` each accept an optional symm-mem window for zero-copy.
 - `nvte_ep_combine` — scatter-sum expert outputs back to source ranks (unweighted; caller pre-multiplies by `recv_topk_weights`). `expert_out` accepts a window.
 - `nvte_ep_dispatch_bwd` — backward of dispatch; routes token and weight grads back to source. `grad` and `g_recv_topk_weights` accept windows; the gathered outputs (`grad_tokens`, `grad_topk_weights`).
 - `nvte_ep_combine_bwd` — backward of combine; `grad` and `grad_expert_out` accept windows. Padded slots in `grad_expert_out` are zeroed.

### Backend + build

- **NCCL EP backend** (`transformer_engine/common/ep/`): `EPBackend` singleton, HT-mode dispatch/combine over NCCL EP (`libnccl_ep.so`), group/layer registration. Internal helper `make_payload_tensor()` builds the per-call `ncclEpTensor_t`: when the caller's `NVTECommWindow.window != nullptr` it sets `win_hdl` + `win_offset` (zero-copy); otherwise it sets `data` from `nvte_tensor_data(t)` (HBM fallback).
- **Runtime gates** (in `EPBackend::initialize`): SM>=90 (via `cudaDeviceGetAttribute`), NCCL>=2.30.4 (via `ncclGetVersion`), CUDA multicast/NVLS support.
- **Stub path**: when `NVTE_WITH_NCCL_EP=OFF`, `ep/ep_api_stub.cpp` provides throwing `nvte_ep_*` stubs so framework bindings link unconditionally; failure surfaces at first `nvte_ep_initialize`.
- **Build wiring**
  - `setup.py` builds `libnccl_ep.so` from `3rdparty/nccl` by default; auto-disables NCCL EP when no requested CUDA arch >= 90. Explicit `NVTE_BUILD_WITH_NCCL_EP=1` with all archs < 90 is treated as user error `NVTE_BUILD_WITH_NCCL_EP=0` to opt out.
  - `NCCL_HOME` resolved dynamically: explicit env → `/opt/nvidia/nccl`, `/usr/local/nccl`, `/usr` → `ldconfig -p` fallback.

## Testing

- C++ distributed tests under `tests/cpp_distributed/`.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


transformer_engine/common/ep/ep_backend.cpp
```diff
@@ -0,0 +1,514 @@
...
+void EPBackend::init(ncclComm_t ep_comm, NVTEEpGroupConfig group_config) {
+  NVTE_CHECK(!initialized_, "EPBackend already initialized");
+
+  group_config_ = group_config;
+
+  ncclEpGroupConfig_t cfg = NCCL_EP_GROUP_CONFIG_INIT;
+  cfg.algorithm = NCCL_EP_ALGO_HIGH_THROUGHPUT;
+  cfg.num_experts = static_cast<unsigned int>(group_config.num_experts);
+  cfg.max_dispatch_tokens_per_rank = static_cast<unsigned int>(group_config.max_tokens_per_rank);
+  cfg.max_token_bytes = static_cast<unsigned int>(group_config.hidden_dim * sizeof(nv_bfloat16));
```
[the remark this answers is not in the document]
phu0ngng: Note for myself: Need to expose this option for users to set in `ep_bootstrap`.

tests/cpp_distributed/CMakeLists.txt
```diff
@@ -30,7 +30,7 @@ if(NOT DEFINED TE_LIB_PATH)
     get_filename_component(TE_LIB_PATH ${TE_LIB_FILE} DIRECTORY)
 endif()
 
-find_library(TE_LIB NAMES transformer_engine PATHS "${TE_LIB_PATH}/.." ${TE_LIB_PATH} ENV TE_LIB_PATH REQUIRED)
+find_library(TE_LIB NAMES transformer_engine PATHS "${TE_LIB_PATH}/.." ${TE_LIB_PATH} ENV TE_LIB_PATH REQUIRED NO_CMAKE_SYSTEM_PATH)
```
ptrendx: Why do we need that?

tests/cpp_distributed/CMakeLists.txt
```diff
@@ -46,12 +46,99 @@ add_executable(test_comm_gemm
...
 target_include_directories(test_comm_gemm PRIVATE ${MPI_CXX_INCLUDE_PATH} $ENV{CUBLASMP_HOME}/include)
 target_link_libraries(test_comm_gemm PUBLIC CUDA::cuda_driver CUDA::cudart GTest::gtest ${TE_LIB} CUDA::nvrtc MPI::MPI_CXX ${NCCL_LIB} OpenMP::OpenMP_CXX)
 
 include(GoogleTest)
 gtest_discover_tests(test_comm_gemm DISCOVERY_TIMEOUT 600)
+
+# ── EP distributed tests (HT mode) ─────────────────────────────────────────
+# No MPI dependency — processes are spawned by run_test_ep.sh with
+# --rank / --nranks flags.  ncclUniqueId exchange uses a
+# shared temp file (see test_ep_common.h for details).
```
ptrendx: I believe that the other distributed tests do rely on MPI, so why don't we also do that here?

tests/cpp_distributed/CMakeLists.txt
```diff
@@ -46,12 +46,99 @@ add_executable(test_comm_gemm
...
+    CUDA::nvrtc
+    GTest::gtest
+    ${TE_LIB}
+    ${NCCL_LIB}
+    ${NCCL_EP_LIB})
+
+# nvrtc symbols are referenced from libtransformer_engine.so but not in its
+# DT_NEEDED list (loaded via dlopen in Python).  For cpp tests we link nvrtc
+# explicitly with --no-as-needed so the linker keeps the dependency.
+set(EP_TEST_LINK_OPTS "LINKER:--no-as-needed")
```
ptrendx: This sounds like a bug actually, but the other tests do not need to do this, they instead specify the nvrtc after the TE_LIB in the LINKER_LIBS variable.

tests/cpp_distributed/test_ep_common.h
```diff
@@ -0,0 +1,308 @@
...
+  } while (false)
+
+#define ASSERT_NCCL_OK(expr)                                                      \
+  do {                                                                            \
+    ncclResult_t _err = (expr);                                                   \
+    if (_err != ncclSuccess) {                                                    \
+      fprintf(stderr, "NCCL error %d: %s\n", _err, ncclGetErrorString(_err));    \
+      exit(EXIT_FAILURE);                                                         \
+    }                                                                             \
+  } while (false)
```
ptrendx: Why not use logging.h?

tests/cpp_distributed/test_ep_common.h
```diff
@@ -0,0 +1,308 @@
...
+  }
+  TensorHandle& operator=(TensorHandle&& o) noexcept {
+    if (this != &o) {
+      if (tensor) nvte_destroy_tensor(tensor);
+      tensor = o.tensor; dev_ptr = o.dev_ptr;
+      o.tensor = nullptr; o.dev_ptr = nullptr;
+    }
+    return *this;
+  }
+};
```
ptrendx: Why not TensorWrapper?

tests/cpp_distributed/test_ep_common.h
```diff
@@ -0,0 +1,308 @@
...
+  bt.dtype    = dtype;
+  bt.shape    = s;
+  nvte_set_tensor_param_v2(h.tensor, kNVTERowwiseData, &bt, sizeof(bt));
+
+  return h;
+}
+
+// RAII owner for a cudaMalloc'd device buffer; frees on destruction.
+template <typename T>
+struct DevBuf {
```
ptrendx, line 60: We have a very similar thing already in the test_common.h

tests/cpp_distributed/test_ep_init.cu
```diff
@@ -0,0 +1,64 @@
+/*************************************************************************
```
ptrendx: I feel that having those tests as separate entities does not really make sense and will introduce overhead to the CI - the actual functionality tests would already be able to cover those initialization issues, no?

tests/cpp_distributed/test_ep_pipeline.cu
```diff
@@ -0,0 +1,890 @@
...
+    grad_result.alloc(num_tokens * hidden_dim);
+    grad_expert.alloc(recv_capacity * hidden_dim);
+    grad_tokens.alloc(num_tokens * hidden_dim);
+    g_recv_topk_weights.alloc(recv_capacity);
+    grad_topk_weights.alloc(num_tokens * top_k);
+  }
+};
+
+// Bundled NVTETensor views over an EPBuffers — one place to update the shape
+// conventions when the C-API evolves.
```
ptrendx: What do you mean by "when the C-API evolves"? We should aim for stability of the C API.

tests/cpp_distributed/test_ep_pipeline.cu
```diff
@@ -0,0 +1,890 @@
...
+                                   t.recv_topk_weights.tensor, NVTECommWindow{}, stream));
+  ASSERT_NO_THROW(nvte_ep_combine(NVTEEpHandle{handle_id, t.handle_mem.tensor}, t.recv_tokens.tensor, NVTECommWindow{},
+                                  t.result.tensor, stream));
+  CHECK_CUDA(cudaStreamSynchronize(stream));
+
+  std::vector<nv_bfloat16> h_result(num_tokens_ * hidden_dim_);
+  CHECK_CUDA(cudaMemcpy(h_result.data(), buf.result.get(),
+                        h_result.size() * sizeof(nv_bfloat16), cudaMemcpyDeviceToHost));
+  auto h_tok = generate_tokens(g_process_id, num_tokens_, hidden_dim_);
+  // Spot-check 3 hidden-dim positions per token to catch partial-row writes.
```
ptrendx: What? Why don't we check the full data?

tests/cpp_distributed/test_ep.cu
```diff
@@ -0,0 +1,890 @@
...
+  CHECK_CUDA(cudaStreamSynchronize(stream));
+
+  std::vector<nv_bfloat16> h_result(num_tokens_ * hidden_dim_);
+  CHECK_CUDA(cudaMemcpy(h_result.data(), buf.result.get(),
+                        h_result.size() * sizeof(nv_bfloat16), cudaMemcpyDeviceToHost));
+  auto h_tok = generate_tokens(g_process_id, num_tokens_, hidden_dim_);
+  // Spot-check 3 hidden-dim positions per token to catch partial-row writes.
+  const int probes[3] = {0, hidden_dim_ / 2, hidden_dim_ - 1};
+  for (int tok = 0; tok < num_tokens_; ++tok) {
+    float exp = __bfloat162float(h_tok[tok * hidden_dim_]) * static_cast<float>(top_k_);
```
ptrendx: Why do we hardcode BF16 everywhere? I assume that NCCL EP works with the other datatypes, right?

tests/cpp_distributed/test_ep_pipeline.cu
```diff
@@ -0,0 +1,890 @@
...
+  }
+  std::sort(vals.begin(), vals.end());
+  return vals;
+}
+
+// BF16 has 7 mantissa bits; relative ULP ≈ 2^-7. Use 4× headroom for
+// accumulation noise inside dispatch/combine.
+static float bf16_tol(float magnitude) {
+  return 4.f * std::ldexp(std::fabs(magnitude) + 1e-3f, -7);
+}
```
ptrendx: So why can't we just use rtol 2^-5 rather than this formula? In general the error checking here is very custom, could we integrate it better with the rest of the tests?

tests/cpp_distributed/test_ep_coverage.cu
```diff
@@ -0,0 +1,562 @@
+/*************************************************************************
```
ptrendx: What are the cases that this test would catch that the ep_pipeline one would not?

transformer_engine/common/ep/ep_backend.h
```diff
@@ -0,0 +1,122 @@
...
+#include <atomic>
+#include <cstddef>
+#include <cstdint>
+#include <mutex>
+#include <unordered_map>
+
+namespace transformer_engine {
+namespace ep {
+
+/*! \brief EP backend singleton — owns the NCCL EP group; borrows the comm. */
```
ptrendx: If it borrows the communicator then on the framework side we need to make sure that it stays alive.
Also, if it is a singleton, how does it work with multiple GPUs per process?

transformer_engine/common/ep/ep_backend.h
```diff
@@ -0,0 +1,122 @@
...
+   *  and must span exactly config.ep_size ranks.
+   */
+  static void initialize(ncclComm_t ep_comm, NVTEEpGroupConfig config);
+
+  /*! \brief Tear down the backend. Idempotent. Does not destroy ep_comm_. */
+  static void shutdown();
+
+  // Host-only: reserve a fresh handle_id, cache the layer config, and report
+  // the handle_mem buffer size the caller must allocate.
+  uint64_t register_layer(NVTEEpLayerConfig layer_config, size_t* handle_mem_size);
```
ptrendx: Is it ever-growing? I don't see any free_layer API.

transformer_engine/common/include/transformer_engine/ep.h
```diff
@@ -0,0 +1,161 @@
...
+} NVTEEpGroupConfig;
+
+/*! \brief Per-layer EP configuration. */
+typedef struct {
+  int num_local_experts; /*!< Reserved for ABI stability (derived from group config). */
+  int top_k;             /*!< Per-token expert fan-out. Required. */
+  size_t dispatch_output_per_expert_alignment;
+  /*!< Per-expert zone alignment in tokens (pow2; 0/1 = no padding). Must match
+   *   between nvte_ep_register_layer and nvte_ep_prepare. */
+} NVTEEpLayerConfig;
```
ptrendx, line 68: If we make this a public API then we should probably version those?

transformer_engine/common/ep/ep_backend.cpp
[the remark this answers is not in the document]
phu0ngng: Done.

tests/cpp_distributed/test_ep_common.h
phu0ngng: Done

transformer_engine/common/include/transformer_engine/ep.h
phu0ngng: Hi, no other TE public struct is versioned, so I think EP should follow the same convention for now. We can add versioning for all structs in a follow-up PR.

tests/cpp_distributed/CMakeLists.txt
phu0ngng: Cleaned.

tests/cpp_distributed/CMakeLists.txt
phu0ngng: Cleaned.

tests/cpp_distributed/test_ep_common.h
phu0ngng: Cleaned.

tests/cpp_distributed/test_ep_init.cu
phu0ngng: Remove + solidated the tests

tests/cpp_distributed/test_ep_pipeline.cu
phu0ngng: Cleaned

tests/cpp_distributed/test_ep_pipeline.cu
phu0ngng: Done

tests/cpp_distributed/test_ep_pipeline.cu
phu0ngng: done

transformer_engine/common/ep/ep_backend.h
phu0ngng: > If it borrows the communicator then on the framework side we need to make sure that it stays alive.

Added docstring note for this. Will see what to do in PyTorch.

> Also, if it is a singleton, how does it work with multiple GPUs per process?

Added note that only a single process, single device is supported for now.

transformer_engine/common/ep/ep_backend.h
phu0ngng: Added doc note saying registrations are released at shutdown, and the cache exhausts if registered per-step

tests/cpp_distributed/test_ep_coverage.cu
phu0ngng: Cleaned.

tests/cpp_distributed/test_ep.cu
phu0ngng: Added tests for FP16 and FP32 in the pipeline tests and found out that NCCL EP has assertions and expects the input to be in BF16 ~~.
I requested NCCL EP to support other types. FP16 and FP32 unit tests will be skipped for now.

transformer_engine/common/include/transformer_engine/ep.h
ptrendx: A lot of other structs like this (like the quantization config) are opaque though. I'm fine with it being not versioned for now if we are fairly sure that there is not going to be churn with them.

tests/cpp_distributed/CMakeLists.txt
```diff
@@ -46,12 +46,84 @@ add_executable(test_comm_gemm
...
+             PATH_SUFFIXES lib lib64
              REQUIRED)
+
+# NCCL headers: prefer submodule build output (has the handle_init API),
+# then submodule src, then system (CUDA toolkit).
+set(NCCL_SUBMODULE_INCLUDE "${CMAKE_CURRENT_SOURCE_DIR}/../../3rdparty/nccl/build/include")
+set(NCCL_SUBMODULE_SRC_INCLUDE "${CMAKE_CURRENT_SOURCE_DIR}/../../3rdparty/nccl/src/include")
+if(EXISTS "${NCCL_SUBMODULE_INCLUDE}/nccl.h")
+  set(NCCL_INCLUDE_DIR "${NCCL_SUBMODULE_INCLUDE}")
+elseif(EXISTS "${NCCL_SUBMODULE_SRC_INCLUDE}/nccl.h")
```
ptrendx: Why would this ever not exist (other than the submodule not being downloaded, but then that is a problem by itself)?

tests/cpp_distributed/CMakeLists.txt
```diff
@@ -46,12 +46,84 @@ add_executable(test_comm_gemm
...
+# ── EP distributed tests ──────────────────────────────────────────────────────
+# Launched via mpirun; ncclUniqueId exchange uses MPI_Bcast (see test_ep_common.h).
+# Headers + libs come from the in-tree 3rdparty/nccl submodule build.
+set(NCCL_EP_SUBMODULE_ROOT
+    "${CMAKE_CURRENT_SOURCE_DIR}/../../3rdparty/nccl")
+find_library(NCCL_EP_LIB
+             NAMES nccl_ep libnccl_ep
+             HINTS ${NCCL_EP_SUBMODULE_ROOT}/build/lib
+             NO_DEFAULT_PATH
+             REQUIRED)
```
ptrendx: The fact that you need to do that in tests is worrying - it is possible that this will make the experience of installing and using TE worse if people need to care about the NCCL EP installation and library loading. Could we link it statically into libtransformer_engine.so instead and just rely on libnccl?

tests/cpp_distributed/CMakeLists.txt
```diff
@@ -46,12 +46,84 @@ add_executable(test_comm_gemm
...
+
+set(NCCL_EP_INCLUDE_DIR "${NCCL_EP_SUBMODULE_ROOT}/contrib/nccl_ep/include")
+if(NOT EXISTS "${NCCL_EP_INCLUDE_DIR}/nccl_ep.h")
+  message(FATAL_ERROR
+    "NCCL EP header not found at ${NCCL_EP_INCLUDE_DIR}/nccl_ep.h. "
+    "Run `git submodule update --init --recursive` to checkout 3rdparty/nccl.")
+endif()
+message(STATUS "EP test: NCCL EP headers: ${NCCL_EP_INCLUDE_DIR}")
```
ptrendx: Why do we need NCCL EP headers to compile tests?

transformer_engine/common/ep/ep_api_stub.cpp
```diff
@@ -0,0 +1,61 @@
+/*************************************************************************
```
ptrendx: The rest of the codebase does not use this convention of a full file with stubs. Instead, please consolidate into ep_api.cpp with `#if defined(NVTE_WITH_NCCL_EP)`.

transformer_engine/common/ep/ep_backend.h
```diff
@@ -0,0 +1,122 @@
...
+  EPBackend(const EPBackend&) = delete;
+  EPBackend& operator=(const EPBackend&) = delete;
+
+  // ep_comm is borrowed — caller retains ownership across the backend lifetime.
+  void init(ncclComm_t ep_comm, NVTEEpGroupConfig config);
+
+  static EPBackend& instance();  // Meyers singleton accessor
+  static void validate_config(const NVTEEpGroupConfig& config);
+
+  static ncclDataType_t nvte_dtype_to_nccl(NVTEDType dtype);
```
ptrendx: This looks like a pretty generic function, maybe it should be somewhere else?

transformer_engine/common/ep/ep_backend.h
```diff
@@ -0,0 +1,122 @@
...
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+
+/*! \file ep_backend.h
+ *  \brief Internal NCCL EP singleton; not part of the public API.
+ *
+ *  Per handle_id the cache stores config only (no device pointers), so
+ *  handle_mem may be relocated between ops. Cap: NVTE_EP_HANDLE_CACHE_SIZE
+ *  (default 8192); overflow throws.
```
ptrendx: This limit makes me a little worried. I don't have strong feelings but we will need to monitor if this leads to issues.

transformer_engine/common/include/transformer_engine/ep.h
```diff
@@ -0,0 +1,177 @@
...
+/*! \brief Group-level EP configuration (fixed for the EP group lifetime). */
+typedef struct {
+  int ep_size;             /*!< EP world size. */
+  int num_experts;         /*!< Total experts across all ranks. */
+  int max_tokens_per_rank; /*!< Upper bound on tokens this rank sends per dispatch. */
+  /*! Upper bound on tokens received per dispatch (worst-case top_k fan-out; must be > 0). */
+  int max_recv_tokens_per_rank;
+  int hidden_dim;  /*!< Token hidden dimension. */
+  int max_num_sms; /*!< Max SMs for EP kernels. 0 = auto. */
+  /*! 0 (default): throw on relocated handle_mem for a cached handle_id. 1: silently rebuild. */
```
ptrendx: This would benefit from a better description

transformer_engine/common/include/transformer_engine/ep.h
```diff
@@ -0,0 +1,177 @@
...
+#endif
+
+/* ── Config structs ─────────────────────────────────────────────────────── */
+/* TODO: add a struct_size/version field to these configs (and align with other
+ *       TE public structs) once a TE-wide convention for ABI versioning lands. */
+
+/*! \brief Group-level EP configuration (fixed for the EP group lifetime). */
+typedef struct {
+  int ep_size;             /*!< EP world size. */
+  int num_experts;         /*!< Total experts across all ranks. */
```
ptrendx:
```suggestion
  int num_experts;         /*!< Total number of experts across all ranks. */
```

transformer_engine/common/include/transformer_engine/ep.h
```diff
@@ -0,0 +1,177 @@
...
+  int max_tokens_per_rank; /*!< Upper bound on tokens this rank sends per dispatch. */
+  /*! Upper bound on tokens received per dispatch (worst-case top_k fan-out; must be > 0). */
+  int max_recv_tokens_per_rank;
```
ptrendx: Is there a default that we could set here if somebody doesn't know the right values?

transformer_engine/common/include/transformer_engine/ep.h
```diff
@@ -0,0 +1,177 @@
...
+  int allow_handle_mem_reloc;
+  /*! Widest token dtype the group will dispatch. Sizes NCCL EP staging buffers
+   *  at group create. Tensors passed to nvte_ep_dispatch may use any dtype whose
+   *  element size is <= sizeof(max_token_dtype). */
+  NVTEDType max_token_dtype;
+} NVTEEpGroupConfig;
+
+/*! \brief Per-layer EP configuration. */
+typedef struct {
+  int num_local_experts; /*!< Reserved for ABI stability (derived from group config). */
```
ptrendx: So is this field required to be filled by the user or derived?

transformer_engine/common/include/transformer_engine/ep.h
```diff
@@ -0,0 +1,177 @@
...
+  int top_k;             /*!< Per-token expert fan-out. Required. */
+  size_t dispatch_output_per_expert_alignment;
+  /*!< Per-expert zone alignment in tokens (pow2; 0/1 = no padding). Must match
+   *   between nvte_ep_register_layer and nvte_ep_prepare. */
```
ptrendx: Those docstrings are sometimes above and sometimes below the actual member. I'm not a Doxygen expert but I think only one way is the right one?

ptrendx: Also, to be honest, this docstring doesn't tell me anything as I think it is talking about some internal requirements? It should be expanded in order to be actionable by the user.

transformer_engine/common/include/transformer_engine/ep.h
```diff
@@ -0,0 +1,177 @@
...
+ *  \param[in]     recv_topk_weights_win  Optional symmem window for ``recv_topk_weights``.
+ *  \param[in]     stream                 CUDA stream.
+ */
+void nvte_ep_dispatch(NVTEEpHandle handle, NVTETensor topk_idx, NVTETensor tokens,
+                      NVTECommWindow tokens_win, NVTETensor topk_weights,
+                      NVTECommWindow topk_weights_win, NVTETensor recv_tokens,
+                      NVTECommWindow recv_tokens_win, NVTETensor recv_topk_weights,
+                      NVTECommWindow recv_topk_weights_win, cudaStream_t stream);
+
+/*! \brief Scatter-sum expert outputs back to originating ranks. Unweighted —
```
ptrendx: What is "unweighted caller"?

transformer_engine/common/include/transformer_engine/ep.h
```diff
@@ -0,0 +1,177 @@
+/*************************************************************************
```
ptrendx, line 1: A general comment here that the docstrings here should be expanded to say more of what actually happens.

transformer_engine/common/util/logging.h
```diff
@@ -98,6 +98,14 @@
     }                                                                            \
   } while (false)
 
+#define NVTE_CHECK_NCCL(expr)                                                 \
```
ptrendx: The same function is defined 30 lines later.

transformer_engine/common/CMakeLists.txt
```diff
@@ -391,6 +379,96 @@ if (NVTE_WITH_CUSOLVERMP)
     message(STATUS "Using cuSolverMp at: ${CUSOLVERMP_DIR}")
 endif()
 
+# ── NCCL EP (on by default, HT mode only) ─────────────────────────────────
+# Set -DNVTE_WITH_NCCL_EP=OFF (or NVTE_BUILD_WITH_NCCL_EP=0 in setup.py) to
+# skip NCCL EP entirely — useful on older images whose system NCCL is below
+# the 2.30.4 EP minimum.
+option(NVTE_WITH_NCCL_EP "Build NCCL EP into libtransformer_engine.so" ON)
+if(NVTE_WITH_NCCL_EP)
+# SM>=90 and NCCL>=2.30.4 are gated at runtime in EPBackend::initialize.
```
ptrendx, line 446: Is the gate only on the functionality or also on the symbols? As in, can we compile with the support and then just fail at runtime with "unsupported" or will the loading of the so fail on systems with unsupported NCCL?

transformer_engine/common/CMakeLists.txt
```diff
@@ -391,6 +379,96 @@ if (NVTE_WITH_CUSOLVERMP)
...
+set(NCCL_EP_LIB_DIR "${NCCL_EP_SUBMODULE_ROOT}/build/lib")
+find_library(NCCL_EP_LIB
+    NAMES nccl_ep libnccl_ep
+    HINTS ${NCCL_EP_LIB_DIR}
+    NO_DEFAULT_PATH
+    REQUIRED)
+
+# ── NCCL + GIN headers ─────────────────────────────────────────────────────
+# libnccl.so and all GIN headers (ncclGin.h, ncclWindow_t, ncclDevComm_t)
+# ship with the base CUDA Toolkit OR the 3rdparty/nccl submodule build
```
ptrendx: NCCL does not ship with the base CUDA Toolkit AFAIK. It is also a separate pip package if installed from pypi.

transformer_engine/common/CMakeLists.txt
```diff
@@ -391,6 +379,96 @@ if (NVTE_WITH_CUSOLVERMP)
...
+# (preferred when present; auto-built by setup.py via build_nccl_ep_submodule).
```
ptrendx: So does that mean that we are building our own libnccl.so if the first hint is our submodule?

transformer_engine/common/CMakeLists.txt
```diff
@@ -391,6 +379,96 @@ if (NVTE_WITH_CUSOLVERMP)
...
+# ── NCCL EP (on by default, HT mode only) ─────────────────────────────────
+# Set -DNVTE_WITH_NCCL_EP=OFF (or NVTE_BUILD_WITH_NCCL_EP=0 in setup.py) to
+# skip NCCL EP entirely — useful on older images whose system NCCL is below
+# the 2.30.4 EP minimum.
+option(NVTE_WITH_NCCL_EP "Build NCCL EP into libtransformer_engine.so" ON)
+if(NVTE_WITH_NCCL_EP)
+# SM>=90 and NCCL>=2.30.4 are gated at runtime in EPBackend::initialize.
+# ── NCCL EP headers ────────────────────────────────────────────────────────
+# Headers + libs are produced by the in-tree 3rdparty/nccl submodule build
+# (auto-built by setup.py via build_nccl_ep_submodule).
```
ptrendx, line 449: What do we ship then e.g. inside the wheel? Are we installing libnccl_ep.so system-wide once we built it? How is it found during the importing of TE in Python?

.gitmodules
```diff
@@ -7,3 +7,7 @@
 [submodule "3rdparty/cutlass"]
 	path = 3rdparty/cutlass
 	url = https://github.com/NVIDIA/cutlass.git
+[submodule "3rdparty/nccl"]
+	path = 3rdparty/nccl
+  url = https://github.com/NVIDIA/nccl.git
+  branch = v2.30u1
```
ptrendx: Do we need to specify the branch here rather than normally via the commit hash?

setup.py
```diff
@@ -83,6 +83,34 @@ def setup_common_extension() -> CMakeExtension:
...
+            if explicit_nccl_ep:
+                raise RuntimeError(
+                    "NVTE_BUILD_WITH_NCCL_EP=1 requires at least one CUDA arch >= 90 in "
+                    f"NVTE_CUDA_ARCHS (got '{archs}'). Add '90' or unset NVTE_BUILD_WITH_NCCL_EP."
+                )
+            print(
+                "[NCCL EP] No CUDA arch >= 90 in NVTE_CUDA_ARCHS"
+                f" ('{archs}'); auto-disabling NCCL EP (nvte_ep_* will throw at runtime)."
+            )
+            build_with_nccl_ep = False
```
ptrendx, line 105: Is this needed? Considering that we are fine if they compile for anything AND Hopper+, then we should also be fine if they compile for the same things without Hopper+ (and then it should just say unsupported at runtime once somebody tries to run it on that hardware).

setup.py
```diff
@@ -128,6 +156,109 @@ def setup_requirements() -> Tuple[List[str], List[str]]:
...
+    nccl_include.mkdir(parents=True, exist_ok=True)
+    for cand in (Path(nccl_home) / "include", Path("/usr/include")):
+        p = Path(cand)
+        if (p / "nccl.h").exists():
+            for name in ("nccl.h", "nccl_net.h", "nccl_tuner.h"):
+                src = p / name
+                dst = nccl_include / name
+                if src.exists() and not dst.exists():
+                    dst.symlink_to(src)
+            break
```
ptrendx: Do we really have to do it like that?

transformer_engine/common/ep/ep_backend.cpp
```diff
@@ -0,0 +1,513 @@
...
+#include "../util/logging.h"
+
+namespace transformer_engine {
+namespace ep {
+
+namespace {
+
+// Build a by-value ncclEpTensor_t descriptor. `sizes` is caller-owned and must
+// outlive any NCCL EP call that consumes the descriptor.
+inline ncclEpTensor_t make_tensor(void* data, unsigned int ndim, ncclDataType_t datatype,
```
ptrendx: I would have expected this to take NVTETensor instead of the raw data.

transformer_engine/common/ep/ep_backend.cpp
```diff
@@ -0,0 +1,513 @@
...
+  t.ndim = ndim;
+  t.datatype = datatype;
+  t.data = data;
+  t.sizes = sizes;
+  return t;
+}
+
+// Payload descriptor: prefer the symmem window when set, else fall back to the
+// NVTETensor's raw device pointer.
+inline ncclEpTensor_t make_payload_tensor(const NVTETensor t, const NVTECommWindow& win,
```
ptrendx: same, commwindow and nvtetensor should be enough.

transformer_engine/common/ep/ep_backend.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+  int device, major;
+  NVTE_CHECK_CUDA(cudaGetDevice(&device));
+  NVTE_CHECK_CUDA(cudaDeviceGetAttribute(&major, cudaDevAttrComputeCapabilityMajor, device));
+  NVTE_CHECK(major >= 9,
+             "NCCL EP requires SM_90+ (Hopper or later), "
+             "but current device has compute capability ",
+             major, ".x");
+
+  // NCCL EP needs CUDA multicast (NVLS); init hangs without it.
```
ptrendx: That's a pretty big requirement, no?

transformer_engine/common/ep/ep_backend.cpp
```diff
@@ -0,0 +1,513 @@
...
+  NVTE_CHECK(nccl_version >= kMinNcclVersion, "NCCL EP requires NCCL >= 2.30.4, found ",
+             nccl_version / 10000, ".", (nccl_version / 100) % 100, ".", nccl_version % 100,
+             " at runtime.");
+
+  validate_config(config);
+
+  int comm_size = 0;
+  NVTE_CHECK_NCCL(ncclCommCount(ep_comm, &comm_size));
+  NVTE_CHECK(comm_size == config.ep_size, "ep_comm size (", comm_size, ") must equal ep_size (",
+             config.ep_size, "). Pass the EP sub-communicator, not the world comm.");
```
ptrendx, line 139: The second sentence may not be true, I would drop the part about the world comm.

transformer_engine/common/ep/ep_backend.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+  inst.init(ep_comm, config);
+}
+
+void EPBackend::shutdown() {
```
ptrendx, line 144: When is this called when TE is called from Python? Shouldn't we have some on_exit hook in TE common Python part?

tests/cpp_distributed/CMakeLists.txt
phu0ngng: Clean

transformer_engine/common/ep/ep_backend.h
phu0ngng: You mean the relocation? 
In the latest code, I removed this option in the API and added an env var for JAX WAR usage only instead. We will remove the env var after XLA fixes the bug that the tensor ptr may relocate between runs.

transformer_engine/common/include/transformer_engine/ep.h
phu0ngng: I would say I would prefer not to have a worst-case size by default so that users know their memory consumption when setting it.

transformer_engine/common/include/transformer_engine/ep.h
phu0ngng: It was derived but no longer needed, thus removed.

transformer_engine/common/include/transformer_engine/ep.h
phu0ngng: I will rewrite it to make it clear. By "unweighted", I mean we don't do a weighted sum in Combine. Caller needs to apply `topk_weight * out_tokens` before calling Combine. This weighted product apparently could be fused to the activation function after FC1.

transformer_engine/common/ep/ep_backend.h
phu0ngng: If we move it to another file, we will need to include `nccl.h` there. I would leave it here for now and move it when we need to reuse it.

transformer_engine/common/CMakeLists.txt
phu0ngng: This gate is for capability, not symbols. So it would fail at load time if the NCCL version does not meet.
I can add `dlopen` so that it fails at runtime. Would it be helpful?

phu0ngng: Added lazy load via existing infrastructure in `shared_lib_wrapper.h`.

transformer_engine/common/CMakeLists.txt
phu0ngng: Polished. We don't.

transformer_engine/common/CMakeLists.txt
phu0ngng: Bundled libnccl_ep.so with libte.so.

.gitmodules
phu0ngng: We don't need the branch.

setup.py
phu0ngng: Cleaned up this fallback. Skipped NCCL EP build when arch < 90.

setup.py
phu0ngng: Removed this symlink copy. We don't really need this after we fixed NCCL EP.

transformer_engine/common/ep/ep_backend.cpp
phu0ngng: *multicast is required, not NVLS. Reworded.

setup.py
```diff
@@ -128,6 +146,104 @@ def setup_requirements() -> Tuple[List[str], List[str]]:
...
+        )
+
+    # Stage libnccl_ep.so.0 alongside libtransformer_engine.so so $ORIGIN-rpath
+    # finds it in the installed wheel.
+    soname = "libnccl_ep.so.0"
+    src = (build_dir / "lib" / soname).resolve()
+    dst = current_file_path / "transformer_engine" / soname
+    if dst.is_symlink() or dst.exists():
+        dst.unlink()
+    shutil.copy2(src, dst)
```
phu0ngng: Is it ok if we bundle libnccl_ep.so like this so that it's in the same dir as libtransformer_engine.so? @ptrendx

transformer_engine/common/CMakeLists.txt
phu0ngng: Removed the bundle, statically link libnccl_ep.a with libte.so

.gitignore
```diff
@@ -43,3 +43,6 @@ tensor_dumps/
 artifacts/
 .DS_Store
 .claude/
+
+# NCCL EP shared library staged by setup.py for wheel packaging.
+/transformer_engine/libnccl_ep.so*
```
phu0ngng: To revert.

qa/L1_cpp_distributed/test.sh
```diff
@@ -2,16 +2,52 @@
...
-	cd $TE_PATH/tests/cpp_distributed
-	cmake -GNinja -S. -Bbuild
-	cmake --build build
-	mpirun --allow-run-as-root --np 4 --oversubscribe ./build/test_comm_gemm
+    cd $TE_PATH/tests/cpp_distributed
+    configure_ok=1
+    cmake -GNinja -S. -Bbuild || { test_fail "configure"; configure_ok=0; }
+
+    # Build each suite independently so one suite's build failure does not
+    # mask the other's results. Skip mpirun when the binary is missing.
```
ptrendx, line 28: In general this is fine, but still interested to know why you needed to do those changes?

setup.py
```diff
@@ -84,6 +84,24 @@ def setup_common_extension() -> CMakeExtension:
         cusolvermp_dir = os.getenv("CUSOLVERMP_HOME", "/usr")
         cmake_flags.append(f"-DCUSOLVERMP_DIR={cusolvermp_dir}")
 
+    # NCCL EP (Hopper+): on by default; auto-skipped when no arch >= 90 is
+    # targeted. Set NVTE_BUILD_WITH_NCCL_EP=0 to force off.
+    build_with_nccl_ep = bool(int(os.getenv("NVTE_BUILD_WITH_NCCL_EP", "1")))
```
ptrendx: We should have some consistency wrt naming of those flags. The other ones are named `NVTE_WITH_X` so I think we should continue that (even though I actually like the `BUILD_WITH` version).

setup.py
```diff
@@ -84,6 +84,24 @@ def setup_common_extension() -> CMakeExtension:
...
+    if build_with_nccl_ep:
+        arch_tokens = [a.strip() for a in str(archs or "").split(";") if a.strip()]
+        has_hopper_or_newer = any(
+            t.lower() == "native" or (t.rstrip("af").isdigit() and int(t.rstrip("af")) >= 90)
+            for t in arch_tokens
+        )
+        if not has_hopper_or_newer:
+            print(f"[NCCL EP] No arch >= 90 in NVTE_CUDA_ARCHS ('{archs}'); skipping build.")
+            build_with_nccl_ep = False
```
ptrendx, line 105: If the user explicitly requested build with NCCL EP (by setting the env variable) then this most probably should be an error.

setup.py
```diff
@@ -130,6 +148,89 @@ def setup_requirements() -> Tuple[List[str], List[str]]:
     return [remove_dups(reqs) for reqs in [install_reqs, test_reqs]]
 
 
+def _discover_nccl_home() -> str:
```
ptrendx, line 158: This will not find pip installed NCCL.

setup.py
```diff
@@ -130,6 +148,89 @@ def setup_requirements() -> Tuple[List[str], List[str]]:
...
+    nproc = os.cpu_count() or 8
+    env = os.environ.copy()
+    env["NVCC_GENCODE"] = gencode
+    # NCCL EP needs the core NCCL headers + libnccl.so; write NCCL EP build
+    # outputs to the submodule's local build/ tree.
+    nccl_home = _discover_nccl_home()
+    env["NCCL_HOME"] = nccl_home
+    env["NCCL_EP_BUILDDIR"] = str(build_dir)
+
+    if not nccl_ep_lib.exists():
```
ptrendx: This is too weak of a check. What if somebody builds only for some architectures and then tries to build for a different one (e.g. in a directory mounted to different machines)? That would result in a stale NCCL EP being used.

tests/cpp_distributed/CMakeLists.txt
```diff
@@ -55,9 +55,18 @@ target_include_directories(test_comm_gemm PRIVATE ${test_comm_gemm_INCLUDES})
...
+    HINTS /opt/nvidia/nccl/include /usr/local/nccl/include)
+if(NOT NCCL_INCLUDE_DIR)
+  message(FATAL_ERROR
+    "nccl.h not found. Pass -DNCCL_INCLUDE_DIR=<prefix>/include.")
+endif()
 find_library(NCCL_LIB
              NAMES nccl libnccl
-             PATH_SUFFIXES lib
+             PATH_SUFFIXES lib lib64
              REQUIRED)
```
ptrendx, line 67: There is a mismatch here - you use the NCCL header from one place and then the library from a different place potentially. Is that intentional?

tests/cpp_distributed/test_ep.cu
```diff
@@ -0,0 +1,808 @@
...
+    }
+  }
+  std::sort(got_vals.begin(), got_vals.end());
+
+  auto exp_vals = expected_recv_values_sorted(g_process_id, g_num_processes, num_tokens_,
+                                              top_k_, num_experts_, num_local_experts_);
+
+  ASSERT_EQ(got_vals.size(), exp_vals.size());
+  for (size_t i = 0; i < exp_vals.size(); ++i)
+    EXPECT_NEAR(got_vals[i], exp_vals[i], bf16_tol(exp_vals[i]))
```
ptrendx: Do we need to use any tolerance when testing dispatch?

tests/cpp_distributed/test_ep_common.h
```diff
@@ -0,0 +1,200 @@
...
+// -- Shared routing helper -----------------------------------------------------
+
+// Balanced round-robin routing: token t on rank r maps top_k experts to
+//   (r * num_local_experts + t * top_k + k) % num_experts
+static inline std::vector<int64_t> routing_balanced(
+    int rank, int num_tokens, int top_k, int num_experts, int num_local_experts) {
+  std::vector<int64_t> idx(num_tokens * top_k);
+  for (int t = 0; t < num_tokens; ++t)
+    for (int k = 0; k < top_k; ++k)
+      idx[t * top_k + k] = (rank * num_local_experts + t * top_k + k) % num_experts;
```
ptrendx: Shouldn't there be `rank * num_tokens * topk` instead of `rank * num_local_experts`?

tests/cpp_distributed/test_ep.cu
```diff
@@ -0,0 +1,808 @@
...
+            ep_size_, max_tokens_per_rank_);
+  upload_inputs(buf);
+  EPTensors<> t(buf, num_tokens_, top_k_, hidden_dim_, num_local_experts_);
+
+  // Distinct per-(rank, t, k) weights so each slot carries a unique value.
+  std::vector<float> h_w(num_tokens_ * top_k_);
+  for (int tok = 0; tok < num_tokens_; ++tok)
+    for (int k = 0; k < top_k_; ++k)
+      h_w[tok * top_k_ + k] = 0.1f + 0.01f * tok + 0.001f * k +
+                              0.0001f * (g_process_id + 1);
```
ptrendx: This is pretty peculiar formula

tests/cpp_distributed/test_ep.cu
ptrendx: Ok, that makes sense, but even once they add those datatypes these tests would not work due to the hardcoding of the BF16 operations.

transformer_engine/__init__.py
```diff
@@ -6,10 +6,56 @@
...
+        raise RuntimeError(
+            f"NCCL EP requires libnccl.so.2 (>= {mn}); could not load libnccl.so.2 "
+            "or query its version. Install NCCL or rebuild Transformer Engine with "
+            "NVTE_BUILD_WITH_NCCL_EP=0."
+        )
+    if cur < _NCCL_EP_MIN_VERSION:
+        raise RuntimeError(
+            f"NCCL EP requires NCCL >= {mn} at runtime; found "
+            f"{'.'.join(str(x) for x in cur)}. Upgrade libnccl.so or rebuild "
+            "Transformer Engine with NVTE_BUILD_WITH_NCCL_EP=0."
```
ptrendx: This error should only fire if somebody tried to run with NCCL EP, right? If so, then telling them to rebuild without NCCL EP is counterproductive.

transformer_engine/common/CMakeLists.txt
```diff
@@ -437,6 +437,91 @@ if (NVTE_WITH_CUSOLVERMP)
...
+    "NCCL EP header not found at ${NCCL_EP_INCLUDE_DIR}/nccl_ep.h. "
+    "Run `git submodule update --init --recursive` and rebuild TE.")
+endif()
+message(STATUS "NCCL EP headers: ${NCCL_EP_INCLUDE_DIR}")
+
+# -- libnccl_ep.a -----------------------------------------------------------
+# Statically linked into libtransformer_engine.so. EPBackend::initialize checks
+# NCCL >= 2.30.4 before any nccl_ep call, so the newer NCCL symbols nccl_ep
+# imports stay unresolved (and harmless) under default ELF lazy binding when
+# the gate trips. LD_BIND_NOW environments lose this property.
```
ptrendx, line 464: Ok, so let's try to understand the impact here. Do we know what are the symbols that nccl ep uses from NCCL and in which version those were added (even if they don't meet our API requirements)? In other words, what is the minimum version of NCCL that TE will load properly with (even if EP itself would not work in that environment)?

transformer_engine/common/CMakeLists.txt
```diff
@@ -437,6 +437,91 @@ if (NVTE_WITH_CUSOLVERMP)
...
+file(READ "${NCCL_INCLUDE_DIR}/nccl.h" _nvte_nccl_h)
+string(REGEX MATCH "#define[ \t]+NCCL_MAJOR[ \t]+([0-9]+)" _ "${_nvte_nccl_h}")
+set(_nvte_nccl_major "${CMAKE_MATCH_1}")
+string(REGEX MATCH "#define[ \t]+NCCL_MINOR[ \t]+([0-9]+)" _ "${_nvte_nccl_h}")
+set(_nvte_nccl_minor "${CMAKE_MATCH_1}")
+string(REGEX MATCH "#define[ \t]+NCCL_PATCH[ \t]+([0-9]+)" _ "${_nvte_nccl_h}")
+set(_nvte_nccl_patch "${CMAKE_MATCH_1}")
+if(_nvte_nccl_major AND _nvte_nccl_minor AND _nvte_nccl_patch)
+  message(STATUS "NCCL header: ${NCCL_INCLUDE_DIR}/nccl.h (version ${_nvte_nccl_major}.${_nvte_nccl_minor}.${_nvte_nccl_patch})")
+endif()
```
ptrendx: I don't think this would do anything - after all you are checking nccl.h from the submodule so it will always be ok, whereas it is the library that may be too old.

qa/L1_cpp_distributed/test.sh
phu0ngng: When I run the suite on a local node, I don't have cublasMp in the image. So the comm_gemm tests failed, and the EP tests are skipped. The change is to make sure EP test runs even comm_gemm tests failed.

setup.py
phu0ngng: Added a probe for the pip-installed nvidia-nccl-cu* wheel.

setup.py
phu0ngng: Stamped the gencode used to build NCCL EP.

tests/cpp_distributed/CMakeLists.txt
phu0ngng: It shouldn't. Let me add search for lib first, then use that path to expand the search for the header.

tests/cpp_distributed/test_ep_common.h
phu0ngng: We can. Both produce balanced loads.

tests/cpp_distributed/test_ep.cu
phu0ngng: Simlified.

tests/cpp_distributed/test_ep.cu
phu0ngng: I added FP32 and FP16 to the `EPPipelineTest` which covers both Dispatch and Combine. Other tests will test a subset that is BF16 only. 
But let me add dtype template for all the tests, in case if we want to include all dtypes for all tests in the future.

transformer_engine/__init__.py
phu0ngng: Ask users to update NCCL instead.

transformer_engine/common/CMakeLists.txt
phu0ngng: Working on it.

phu0ngng: Actually, with EFL lazy binding, even if users have an old NCCL version that does not have the symbol NCCL EP references, users should still be able to load TE, since ELF's default is to resolve function symbols only on first call. 
It's only problematic when users enforce binding early (via LD_BIND_NOW/RTLD_NOW).

phu0ngng:
Pipeline 54600556

ptrendx (approved):

transformer_engine/common/include/transformer_engine/ep.h
timmoon10, line 1: Nit: `expert_parallel.h` or `expert_parallelism.h` would be a less ambiguous file name, especially in a public-facing header.

Asking a few AIs "What does EP mean in deep learning?":
- Claude: "Expectation Propagation", "Energy Propagation", or "Episode"
- Gemini: "Epoch", various non-MoE models (EP-DNN, EP-PINN, EP-RNN)
- ChatGPT: "Expert Parallelism", "Epoch", "Embedded Projection"

transformer_engine/common/include/transformer_engine/ep.h
timmoon10: EP optimization is specific to particular MoE models and workflows, and the field is advancing too quickly for me to have confidence that we can treat it as stable.

Best to get the design right before anyone starts using this code. Opaque structs are better for experimentation since we can more easily add or deprecate options, and it also allows us to have defaults that are more complicated than zero-initialization. It is more annoying to implement, but we already have a few examples and it's something Codex can do very well.

setup.py
```diff
@@ -130,6 +155,138 @@ def setup_requirements() -> Tuple[List[str], List[str]]:
     return [remove_dups(reqs) for reqs in [install_reqs, test_reqs]]
 
 
+def _discover_nccl_home() -> str:
```
timmoon10, line 158: This could be a standalone utility funciton in [`build_utils/utils.py`](https://github.com/NVIDIA/TransformerEngine/blob/main/build_tools/utils.py). It's actually quite general and not specific to NCCL EP.

setup.py
```diff
@@ -84,6 +84,31 @@ def setup_common_extension() -> CMakeExtension:
...
+        if not has_hopper_or_newer:
+            if nccl_ep_explicit:
+                raise RuntimeError(
+                    f"NVTE_WITH_NCCL_EP=1 was set but NVTE_CUDA_ARCHS ('{archs}') "
+                    "contains no arch >= 90. NCCL EP requires Hopper or newer."
+                )
+            print(f"[NCCL EP] No arch >= 90 in NVTE_CUDA_ARCHS ('{archs}'); skipping build.")
+            build_with_nccl_ep = False
+    if build_with_nccl_ep:
+        nccl_home = build_nccl_ep_submodule()
```
timmoon10, line 107: Flagging that we compile `libnccl_ep.a` at a different time than when we compile `libtransformer_engine.so.`
- `libnccl_ep.a`: Compiled during the Setuptools configuration stage, while processing `setup.py`.
- `libtransformer_engine.so`: Compiled during the Setuptools build stage, configured via a custom `setuptools.Extension`.

It's not obviously wrong, but it does seem messy. I suppose there are also some edge case problems (if NCCL is not installed but is listed as a dependency, compiling NCCL EP in the configuration stage will fail but building during the build stage can succeed).

It seems the "proper" approach would be to treat `libnccl_ep.a` compilation as basically a part of compiling `libtransformer_engine.so`. However, [`CMakeExtension`](https://github.com/NVIDIA/TransformerEngine/blob/f95573f128c8539c2c7bc6ad27425202eb981ffb/build_tools/build_ext.py#L30) is a very general class for CMake projects and I'm not sure how messy it would be to make add the NCCL EP compilation logic.

setup.py
```diff
@@ -130,6 +155,138 @@ def setup_requirements() -> Tuple[List[str], List[str]]:
...
+                arch_list.append(bare)
+    if not arch_list:
+        raise RuntimeError(
+            "NCCL EP requires Hopper or newer (SM >= 90); none found in"
+            f" NVTE_CUDA_ARCHS={cuda_archs()!r}. Re-run with NVTE_WITH_NCCL_EP=0 to skip the NCCL"
+            " EP build (the rest of TE still builds)."
+        )
+    gencode = " ".join(f"-gencode=arch=compute_{a},code=sm_{a}" for a in arch_list)
+
+    nproc = os.cpu_count() or 8
```
timmoon10, line 257: This seems likely to break users with wimpy nodes. We can reuse an existing API to control build parallelism:
```suggestion
    nproc = get_max_jobs_for_parallel_build()
```

phu0ngng:
Pipeline 54600556 - all tests passed.

timmoon10 (approved):
This still has some minor issues, but deadlines are what they are and the changes are well-contained. We should incorporate suggestions into future PRs.

PR #3036, opened 2026-05-22
[JAX] Expert Parallelism: JAX primitives + VJPs

phu0ngng:
## Summary

Third PR in the TE Expert Parallelism (EP) series, built on top of https://github.com/NVIDIA/TransformerEngine/pull/3034. Lands the JAX bindings: an XLA FFI layer over the `nvte_ep_*` C API, a Python wrapper with `custom_vjp` for autograd, mesh-aware sharding rules, a multi-process test suite, and an end-to-end MoE example. NCCL `ncclEpDispatch`/`ncclEpCombine` are exposed as XLA primitives and work with CUDA-graph capture.

### Public Python API (`transformer_engine/jax/ep.py`)

```python
from transformer_engine.jax.ep import (
    EpHandle,        # opaque (id, handle_mem) pair from ep_prepare
    ep_bootstrap,    # one-shot per-process: init NCCL comm + nvte_ep_initialize
    ep_dispatch,     # custom_vjp-wrapped dispatch 
    ep_combine,      # custom_vjp-wrapped combine
```
`ep_dispatch` / `ep_combine` are `jax.custom_vjp` functions: forward is the FFI primitive, backward calls the matching `nvte_ep_*_bwd` FFI primitive directly (no `ep_prepare` in the bwd — routing state is already cached in `handle.mem`). Note that `ep_dispatch` also calls `ep_prepare` in the forward path, which all-gathers and preprocesses routing maps.

### XLA FFI bindings (`transformer_engine/jax/csrc/extensions/ep.cpp`)

Five `XLA_FFI_DEFINE_HANDLER_SYMBOL` entries — `EpPrepareHandler`, `EpDispatchHandler`, `EpCombineHandler`, `EpDispatchBwdHandler`, `EpCombineBwdHandler` — each calling the corresponding `nvte_ep_*` C entry point. All marked `FFI_CudaGraph_Traits` so they capture cleanly. `handle_id` is a static FFI attribute baked at jit trace time.

### Primitives + Python layer (`transformer_engine/jax/cpp_extensions/ep.py`, +951 lines)

Standard TE primitive plumbing: `abstract_eval` (shape/dtype inference), `lowering`, `impl`, `outer_primitive` registration, and partitioning rules so the EP collective is treated as a single sharded op by XLA (no spurious resharding around it).

### Sharding (`transformer_engine/jax/sharding.py`, +12 lines)

Adds the EP mesh axis to the global mesh resource set so downstream sharding rules can reference it.

### Build wiring (`build_tools/jax.py`, +41 lines)

Threads NCCL EP linkage through the JAX `transformer_engine_jax` extension. No new top-level build flags; rides on the parent PR's `NVTE_BUILD_WITH_NCCL_EP`.

### Tests & example

- `tests/jax/test_multi_process_ep.py` (+690 lines): 13 tests covering bootstrap, `ep_prepare` shape/handle contracts, primitive-level dispatch/combine identity (uniform + skewed routing), `custom_vjp` fwd+bwd correctness, and HLO inspection (must not insert XLA collectives outside the EP FFI).
- `tests/jax/multi_process_launch_ep.sh`: 4-rank launcher; sets `XLA_FLAGS` to keep XLA command-buffer capture off for the EP FFI sequence (NCCL EP graph-destroy interaction).
- `examples/jax/ep/ep_moe.py` (+394 lines) + `run_test_ep.sh`: end-to-end MoE with EP, dp=ep=2 mesh, includes a ref-comparison `--check` that verifies fwd+bwd vs a single-process reference.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


transformer_engine/jax/csrc/extensions/ep.cpp
```diff
@@ -0,0 +1,457 @@
...
+
+  void shutdown() {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (comm_ == nullptr) return;
+    ncclCommDestroy(comm_);
+    comm_ = nullptr;
+  }
+
+ private:
+  EpCommManager() = default;
```
mgoldfarb-nvidia: If we use stateful FFI calls we could tie to EP communicator to the lifetime of the jax computation rather than the process.

transformer_engine/jax/csrc/extensions/ep.cpp
```diff
@@ -0,0 +1,457 @@
...
+  uint64_t handle_id = nvte_ep_register_layer(layer_cfg, &handle_mem_size);
+  return pybind11::make_tuple(handle_id, handle_mem_size);
+}
+
+// ── ep_prepare ────────────────────────────────────────────────────────────────
+
+Error_Type EpPrepareFFI(cudaStream_t stream, Buffer_Type topk_idx, Result_Type token_counts,
+                        Result_Type handle_mem, Result_Type workspace, EpPrepareConfig config) {
+  auto topk_dims = topk_idx.dimensions();
+  NVTE_CHECK(topk_dims.size() >= 2,
```
mgoldfarb-nvidia, line 199: nit: can we return FFI InvalidArgument instead of a NVTE_CHECK for these inputs?

phu0ngng:
I would appreciate your help to review this PR @tdophung @jberchtold-nvidia! 
Please focus on the changes in the JAX side, as the TE/Common ones will be discussed in https://github.com/NVIDIA/TransformerEngine/pull/3034

transformer_engine/jax/csrc/extensions/ep.cpp
phu0ngng: Cool to learn! I will update it.

transformer_engine/jax/csrc/extensions/ep.cpp
phu0ngng: This is probably a good idea.  I suggest we make another follow-up MR to do so for all the FFIs.

examples/jax/ep/ep_moe.py
```diff
@@ -0,0 +1,394 @@
...
+    in_spec = PartitionSpec(("dp", "ep"), None, None)  # [B, S, ...]
+    ep3 = PartitionSpec(("dp", "ep"), None, None)  # [num_procs, recv_pr, H]
+    ep2 = PartitionSpec(("dp", "ep"), None)  # [num_procs, recv_pr]
+    # Kernels are EP-replicated across dp colors; shard only the ep-rank axis.
+    kernel_spec = PartitionSpec("ep", None, None, None)
+
+    kernels = kernels.reshape(ep_size, NLE, *kernels.shape[1:])
+
+    @jax.jit
+    def step(idx, toks, w, lk):
```
jberchtold-nvidia: What does `lk` stand for?

tests/jax/multi_process_launch_ep.sh
```diff
@@ -0,0 +1,72 @@
...
+
+#!/bin/bash
+
+SCRIPT_NAMES="${SCRIPT_NAMES:-test_multi_process_ep.py}"
+TEST_TIMEOUT_S="${TEST_TIMEOUT_S:-180}"
+
+
+# Keep XLA command-buffer capture OFF for the EP FFI sequence. Cuda-graph
+# capture of dispatch→combine fails with CUDA_ERROR_INVALID_VALUE on graph
+# destroy under recent XLA/NCCL; the non-cudaGraph path is correct since the
```
jberchtold-nvidia: Does this mean we always have to disable cuda graph capture for now when using TE EP until this is fixed? If so, would it be better to remove the CudaGraph trait from the EP FFIs instead

transformer_engine/jax/cpp_extensions/ep.py
```diff
@@ -0,0 +1,951 @@
...
+        assert (
+            len(topk_idx_aval.shape) >= 2
+        ), f"topk_idx must be at least 2D [..., top_k], got shape {topk_idx_aval.shape}"
+        handle_mem_size = _ep_handle_mem_size(handle_id)
+        leading = _ep_leading_dims(is_outer)
+        token_counts_aval = jax.core.ShapedArray(leading + (num_local_experts,), jnp.int32)
+        handle_mem_aval = jax.core.ShapedArray(leading + (handle_mem_size,), jnp.uint8)
+        # FFI scratch for the int32 -> int64 topk_idx upcast.
+        # TODO(phuong): drop once NCCL EP supports int32 topk_idx.
+        workspace_aval = jax.core.ShapedArray(topk_idx_aval.shape, jnp.int64)
```
jberchtold-nvidia: I think this will fallback to jnp.int32 unless JAX_ENABLE_X64 is set, without adjusting the shape accordingly so you get a smaller amount of bytes. We may need to make this jnp.int32 and make the shape 2x bigger.

Here is jnp.zeros, I believe it also applies to primitive outputs
```
>>> jnp.zeros((128,), jnp.int64)
<stdin>:1: UserWarning: Explicitly requested dtype int64 requested in zeros is not available, and will be truncated to dtype int32. To enable more dtypes, set the jax_enable_x64 configuration option or the JAX_ENABLE_X64 shell environment variable. See https://github.com/jax-ml/jax#current-gotchas for more.
Array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
       0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0], dtype=int32)
>>> a = jnp.zeros((128,), jnp.int64)
>>> a.dtype
dtype('int32')
>>> a.shape
(128,)
```

transformer_engine/jax/cpp_extensions/ep.py
```diff
@@ -0,0 +1,951 @@
...
+        assert (
+            len(tokens_aval.shape) >= 2
+        ), f"tokens must be at least 2D [..., H], got shape {tokens_aval.shape}"
+        recv_pr = recv_capacity_per_rank
+        tok_dtype = dtypes.canonicalize_dtype(tokens_aval.dtype)
+        hidden_dim = tokens_aval.shape[-1]
+        leading = _ep_leading_dims(is_outer)
+        recv_tokens_aval = jax.core.ShapedArray(leading + (recv_pr, hidden_dim), tok_dtype)
+        recv_topk_weights_aval = jax.core.ShapedArray(leading + (recv_pr,), jnp.float32)
+        workspace_aval = jax.core.ShapedArray(topk_idx_aval.shape, jnp.int64)
```
jberchtold-nvidia: Same comment as above about int64

examples/jax/ep/ep_moe.py
```diff
@@ -0,0 +1,394 @@
...
+
+def _distributed_init(args):
+    jax.distributed.initialize(
+        coordinator_address=args.coordinator_address,
+        num_processes=args.num_processes,
+        process_id=args.process_id,
+        local_device_ids=[args.process_id],
+    )
+    assert (
+        jax.local_device_count() == 1
```
jberchtold-nvidia, line 69: Does this `local_device_count() == 1` restriction only apply to this example or does it apply to all TE EP usage? If all TE EP usage, can we also add this assertion into `ep_bootstrap`

tests/jax/multi_process_launch_ep.sh
phu0ngng: Wait, this is an old command. Let me clean them up and enable CUDAGraph! I fixed this in NCCL EP already :D.

transformer_engine/jax/cpp_extensions/ep.py
phu0ngng: You are right. 
There is another int32-int64 conversion with a CUDA kernel triggered underneath, and this JAX conversion should be cleaned up.

examples/jax/ep/ep_moe.py
phu0ngng: This will apply to all, as we are expecting a single process per device setup. The current TE EP does not support single process multiple devices. Let me add this assertion to the ep_bootstrap.

jberchtold-nvidia (dismissed):
LGTM pending CI

transformer_engine/jax/ep.py
```diff
@@ -0,0 +1,293 @@
...
+# ── Bootstrap ────────────────────────────────────────────────────────────────
+
+
+def ep_bootstrap(
+    world_size,
+    rank,
+    ep_size,
+    num_experts,
+    max_tokens_per_rank,
+    recv_capacity_per_rank,
```
tdophung, line 77: for the ep.h file def of `NVTEEpGroupConfig`, you used `max_recv_tokens_per_rank` instead of this. Just for consistency, maybe we should use the same names?

transformer_engine/jax/ep.py
```diff
@@ -0,0 +1,293 @@
...
+    """Initialize the EP communicator. Call once per process before any EP op.
+
+    max_num_sms caps the SMs allotted to EP kernels (0 = auto).
+    """
+    if world_size < 4:
+        raise ValueError(
+            f"ep_bootstrap requires world_size >= 4 (got {world_size}); NCCL EP requires"
+            " at least 4 ranks on the node for its HT mode."
+        )
+    UID_SIZE = 128
```
tdophung, line 144: nit: after looking into the headers in nccl I figured out what this is. However might be helpful to have an inline comment to say what this is. Like `# NCCL_UNIQUE_ID_BYTES from nccl.h to store host name, listening port, etc.`

transformer_engine/jax/ep.py
```diff
@@ -0,0 +1,303 @@
...
+    )
+    out_leading = tuple(tokens.shape[:-1])
+    primal = (recv_tokens, recv_topk_weights, handle, token_counts)
+    return primal, (handle, out_leading, top_k)
+
+
+def _dispatch_bwd(recv_capacity_per_rank, dispatch_output_per_expert_alignment, res, g_outputs):
+    del recv_capacity_per_rank, dispatch_output_per_expert_alignment
+    handle, out_leading, top_k = res
+    # Re-pin cotangent sharding: XLA transpose can drop the EP axis on a
```
tdophung: I now understand that the sharding for other fwd-output cotangents can be lost when propagated to bwd. But is this a fault of JAX that we should ask to be fixed? Did you write this defensively because you ran into a bug where it was trying to read the entire global tensor?

transformer_engine/jax/cpp_extensions/ep.py
```diff
@@ -0,0 +1,961 @@
...
+    """Exchange routing metadata; return ``(token_counts, EpHandle)``."""
+    import sys as _sys
+
+    top_k = int(topk_idx.shape[-1])
+    alignment = int(dispatch_output_per_expert_alignment)
+    # Cache handle_id by caller (file:lineno, top_k, alignment): JAX re-traces
+    # the same call site (e.g. custom_vjp fwd vs primal) and the resulting
+    # EpHandles must share the same id to compare equal in pytree aux.
+    f = _sys._getframe(1)
+    cache_key = (f.f_code.co_filename, f.f_lineno, top_k, alignment)
```
tdophung: would the public ep_dispatch wrapper interacts badly with this? Every ep_dispatch(...) call in a user's program ultimately routes through
`token_counts, handle = tex.ep_prepare(topk_idx, dispatch_output_per_expert_alignment)`
from `jax/ep.py:191`

so  `_sys._getframe(1)` always sees the same `(jax/ep.py, 191)`. That means a model with multiple MoE layers all using `ep_dispatch` end up sharing one handle_id across layers, which would corrupt the cache for the routing state?

tdophung: maybe you don't see it because the test is only 1 layer? I am not entirely surer because I have not tried running a case with multiple layers. This is just from reading the code

phu0ngng: This is valid. Thanks for finding it!
I will expose an option so that the user can host the handle_id per layer.

transformer_engine/jax/csrc/extensions/ep.cpp
```diff
@@ -0,0 +1,539 @@
...
+                              FFI_CudaGraph_Traits);
+
+// ── ep_dispatch ───────────────────────────────────────────────────────────────
+
+Error_Type EpDispatchFFI(cudaStream_t stream, EpInstanceState* ep_state, Buffer_Type handle_mem,
+                         Buffer_Type topk_idx, Buffer_Type tokens, Buffer_Type topk_weights,
+                         Result_Type recv_tokens, Result_Type recv_topk_weights,
+                         Result_Type workspace, EpDispatchConfig config) {
+  (void)ep_state;
+  auto token_dims = tokens.dimensions();
```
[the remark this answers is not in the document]
tdophung: Currently our router is fixed to fp32, so there is no chance it will become bf16 or f16. Whenever we change the router to output more than just 1 datatype, then we can change this one too

transformer_engine/jax/cpp_extensions/ep.py
```diff
@@ -0,0 +1,961 @@
...
+                "EpDispatchBwd: g_recv_topk_weights must be sharded as"
+                " PartitionSpec(ep_resource, None) (or ((dp, ep), None) when dp/fsdp is set)"
+                f" over [num_procs, recv_pr]; got spec={gw_spec}."
+            )
+        resolved = _resolve_out_partition_spec(out_partition_spec, len(out_leading_shape))
+        per_shard_leading = _per_shard_leading(out_leading_shape, resolved, mesh)
+        arg_shardings = tuple(a.sharding for a in arg_infos)
+        out_shardings = [
+            NamedSharding(mesh, PartitionSpec(*resolved)),
+            NamedSharding(mesh, PartitionSpec(*resolved, None)),
```
tdophung: given that len(resolved) is always num_leading_dims + 1 based on the `_resolve_out_partition_spec`, wouldnt it make this partitionspec have 1 more than needed dim? (so num_leading_dims + 1 + 1?)

maybe we should fix it with:

```suggestion
            NamedSharding(mesh, PartitionSpec(*resolved[:-1], None)),
```

I notice that in your test script you already clarified that XLA will drop all trailing None in the partitionspec, hence there wasn't any error message for mismatching number of dims. However, why not size it exactly the same number of dimensions as grad_topk_weights_aval?

transformer_engine/jax/csrc/extensions.h
```diff
@@ -199,6 +199,26 @@ XLA_FFI_DECLARE_HANDLER_SYMBOL(FusedTopkWithScoreFunctionBackwardHandler);
 XLA_FFI_DECLARE_HANDLER_SYMBOL(FusedMoEAuxLossForwardHandler);
 XLA_FFI_DECLARE_HANDLER_SYMBOL(FusedMoEAuxLossBackwardHandler);
 
+// Bootstrap EP (eager NCCL comm init); anchor released by ReleaseEpResources.
+void SetEpBootstrapParams(pybind11::bytes unique_id_bytes, int ep_size, int rank_within_group,
+                          int num_experts, int max_tokens_per_rank, int max_recv_tokens_per_rank,
+                          int hidden_dim, int max_num_sms, int allow_handle_mem_reloc);
```
tdophung: While integrating this into a MoE block, I hit tokens dtype (6) wider than group max_token_dtype (0) on the first bf16 dispatch. `nvte_ep_create_group` requires `max_token_dtype` to be set (added in #3034, enforced at ep_backend.cpp:349), but `SetEpBootstrapParams` here never plumb it through, so the field falls back to zero-init kByte (1 byte) and any bf16 dispatch fails immediately. Could you expose max_token_dtype through ep_bootstrap → SetEpBootstrapParams → NVTEEpGroupConfig? There's no MoEBlock workaround as the staging buffer is sized at create-time so the bootstrap has to know the widest dtype before hand. For my MoeBlock development, i made have it plumb through locally to avoid this

phu0ngng: done

transformer_engine/jax/cpp_extensions/ep.py
phu0ngng: Reworked.

phu0ngng:
pipeline 54608589

transformer_engine/jax/csrc/extensions/pybind.cpp
```diff
@@ -107,6 +107,25 @@ pybind11::dict Registrations() {
   dict["te_fused_moe_aux_loss_forward_ffi"] = EncapsulateFFI(FusedMoEAuxLossForwardHandler);
   dict["te_fused_moe_aux_loss_backward_ffi"] = EncapsulateFFI(FusedMoEAuxLossBackwardHandler);
 
+#ifdef NVTE_WITH_NCCL_EP
```
jberchtold-nvidia, line 110: Since the TE core library and TE/JAX may be built separately, e.g. when pip installing from PyPi the TE core is pre-built but TE/JAX is a source dist and rebuilt. In that case, it's possible TE core is built with NVTE_WITH_NCCL_EP support but when installing TE/JAX the user doesn't have this flag set.

Is this intended? If not, would it be better to have TE core expose a `nvte_is_nccl_ep_supported()` and we call that here instead of using the NVTE_WITH_NCCL_EP macro?

transformer_engine/jax/ep.py
```diff
@@ -0,0 +1,311 @@
...
+    g_in = jax.make_array_from_process_local_data(sharded, local, (world_size, uid_size))
+    g_out = jax.jit(lambda x: x, out_shardings=replicated)(g_in)
+    return np.asarray(g_out).reshape(world_size, uid_size)
+
+
+# ── Bootstrap ────────────────────────────────────────────────────────────────
+
+
+def ep_bootstrap(
+    world_size,
```
jberchtold-nvidia, line 73: Can we add a description of each arg into the docstring? I think I could infer what "world_size" is but would be good to be explicit

phu0ngng:
pipeline 54620197

transformer_engine/jax/ep.py
phu0ngng: Done

transformer_engine/jax/csrc/extensions/pybind.cpp
phu0ngng: This flag is enabled by default. In the case that they disable it, then those ep FFIs are not exported.

jberchtold-nvidia (dismissed):
LGTM pending CI

jberchtold-nvidia (approved):

PR #3038, opened 2026-05-22
[PyTorch] Make `modules.GroupedLinear` graph-safe

yaox12:
# Description

- Enable grouped quantization and cuBLASLt grouped gemm for `modules.GroupedLinear` to benefit cases where cuteDSL fused grouped gemm is not available.  
    1. Reduce CPU overhead by reducing number of kernels.
    2. Be CUDA-Graph-safe.
    3. Improve kernel performance.

- Move grouped gemm and grouped linear related tests to a standalone file.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1105,7 +1645,7 @@ def _load_from_state_dict(
     def forward(
         self,
         inp: torch.Tensor,
-        m_splits: List[int],
+        m_splits: Union[List[int], torch.Tensor],
```
timmoon10: I see that the grouped tensor path can handle both cases where `m_splits` is `list[int]` or `torch.Tensor`, but the split-quantize path still assumes it is `list[int]`. It should be simple just the add a `tolist` to the split-quantize path so that the caller can pass in a `torch.Tensor` without needing to care about the TE implementation.

tests/pytorch/test_grouped_linear.py
timmoon10, line 1: In the future, we should consider moving the grouped MLP tests from `test_fusible_ops.py` into this file.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -349,12 +702,199 @@ def forward(
...
+            if ctx.fuse_wgrad_accumulation:
+                wgrad_list = main_grads
+            else:
+                weight_shape = [ctx.weights_shape_0, ctx.weights_shape_1]
+                wgrad_list = tex.bulk_allocate(
+                    [weight_shape] * N,
+                    [ctx.activation_dtype] * N,
+                    ctx.device,
+                    [256] * N,
+                )
```
timmoon10: It turns out this can cause race conditions (https://github.com/NVIDIA/TransformerEngine/pull/3049). The preferred idiom right now:
```suggestion
                wgrad_packed = torch.empty(
                    N, ctx.weights_shape_0, ctx.weights_shape_1,
                    dtype=ctx.activation_dtype,
                    device=ctx.device,
                )
                wgrad_list = [wgrad_packed[i] for i in range(N)]
```

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1115,8 +1662,8 @@ def forward(
         ----------
         inp : torch.Tensor
              Input tensor.
-        m_splits : List[int]
-                 List of integers representing the split of the input tensor.
+        m_splits : torch.Tensor
+                 Split sizes for the input tensor.
```
timmoon10, line 1666: Going into the future, I think we should prefer passing in the splits as a `torch.Tensor`, even if they are on CPU. This makes the API more consistent. We need to support lists of ints for backward compatibility, but it should be considered deprecated.

timmoon10 (approved):
LGTM, pending CI.

As a followup, we should consolidate the implementation so that we can reuse the same code in `te.ops.GroupedLinear` and `te.GroupedLinear`.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1132,18 +1681,26 @@ def forward(
...
+            m_splits = torch.tensor(m_splits, dtype=torch.int64, device="cpu")
+        elif m_splits.dtype != torch.int64:
+            m_splits = m_splits.to(dtype=torch.int64)
+        if m_splits.size() != (num_gemms,):
             raise ValueError(
-                f"Number of splits ({len(m_splits)}) should match number of"
-                f" GEMMs ({self.num_gemms})."
+                f"Shape of splits tensor ({tuple(m_splits.size())}) "
+                f"does not match number of GEMMs ({num_gemms})."
             )
```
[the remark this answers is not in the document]
timmoon10: If the splits are passed in as a list or on CPU, there's no hope of CUDA Graph capture anyways. A blocking H2D memcpy is the best we can do. Warning is also excessive, since CPU splits are perfectly valid (if suboptimal) when running without CUDA Graphs.

muutot:
May I ask why wgrad's calculation does not include bias calculation

PR #3039, opened 2026-05-22
[PyTorch] Introduce transformer_engine.pytorch.DType enum

vthumbe1503:
# Description

Replace the pybind tex.DType with a canonical Python DType IntEnum throughout transformer_engine.pytorch. CPP still sees the transformer_engine::DType(C++ def for tex.DType) and so pybind caster is defined to make sure Python DType is mapped CPP Dtype correctly.

**Motivation**
- CPU overheads: tex.DType is a pybind enum, so every access/compare/convert in Python crosses into C-extension code.
- torch.compile: tex.DType won't work with torch.compile — TorchDynamo doesn't understand pybind enums, so it graph-breaks (or fails to trace) when one flows through a compiled region.
- Checkpointing: tex.DType lives in tensor/quantizer state and lands in checkpoints; pickling a pybind enum is fragile and awkward to allow-list vs. a stdlib python enum.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- **Canonical DType:** Add a pure-Python DType(IntEnum) in constants.py as the single source of truth. Its members are defined from the C++ enum values, and an import-time assert verifies it stays in sync with the pybinded enum tex.DType 
- **Migration:** Repoint TE_DType maps and move pytorch modules, examples, benchmarks, and tests off raw tex.DType onto constants.DType.
- **Backward compatibility:** We support Union[DType, tex.DType] at all interfaces that accepted tex.DType; i.e tex.DType is still accepted at Quantizer/QuantizedTensor constructor boundaries and stays allow-listed for loading old checkpoints.
- **Python → C++:** Register a type_caster for transformer_engine::DType in Pytorch extensions so that all our pybind registered functions see them. This enables us to pass int/pythonDType values into tex functions from python and enables pybind to convert it to transformer_engine::DType accepted by tex C++ API interface
- **C++ → Python:** Add cached MakePythonDType (csrc/common.*) and use it at quantizer/quantizedtensor construction.

# Checklist:


transformer_engine/pytorch/csrc/common.cpp
```diff
@@ -86,6 +86,16 @@ transformer_engine::DType getTransformerEngineFP8Type(bool e4m3_if_hybrid,
...
 }
 
+pybind11::object MakeTEDType(transformer_engine::DType dtype) {
+  // Cache the Python ``TE_DType`` class object on first call so subsequent
+  // invocations avoid re-importing the module. ``static`` initialization is
+  // thread-safe under C++11. We are always inside a pybind11-invoked function
+  // when this runs, so the GIL is held and Python imports are legal.
+  static pybind11::object te_dtype_cls =
+      pybind11::module_::import("transformer_engine.pytorch.constants").attr("TE_DType");
+  return te_dtype_cls(static_cast<int>(dtype));
```
vthumbe1503: Find a way to bind C++ and python Dtype through pybind cast mechanism

transformer_engine/pytorch/__init__.py
```diff
@@ -132,8 +128,8 @@
             MXFP8Quantizer,
             NVFP4Quantizer,
             Float8BlockQuantizer,
-            # pybind11 enum used as Quantizer.dtype
-            tex.DType,
+            # Python IntEnum used as Quantizer.dtype
+            TE_DType,
```
vthumbe1503: save/load backward compatibilty should be there

vthumbe1503: Done

transformer_engine/pytorch/csrc/common.cpp
vthumbe1503: This is done for Python. -> C++

For C++ to Python. --> Cant avoid this.

benchmarks/benchmark_rht_cast.py
```diff
@@ -8,7 +8,7 @@
 import torch.utils.benchmark as benchmark
 
 import transformer_engine.pytorch as te
-import transformer_engine_torch as tex
+from transformer_engine.pytorch import constants
```
ptrendx: Considering how frequent it is, shouldn't we just expose the DType in the top module (transformer_engine.pytorch)?

transformer_engine/common/util/dtype_pybind_conversion.h
```diff
@@ -0,0 +1,97 @@
...
+    cached_dtype_object = construct_dtype_object(value, type);
+    if (cached_dtype_object == nullptr) {
+      return nullptr;
+    }
+  }
+  Py_INCREF(cached_dtype_object);
+  return cached_dtype_object;
+}
+
+/*! @brief Register the Python -> C++ ``DType`` implicit conversion.
```
ptrendx: Not sure if that is accurate to be honest. My understanding of this function is that it converts the Python object of constants.DType to tex.DType, and only then the real conversion to the C++ type happens - this would actually increase the overhead I think. I believe the right approach is to use the custom type_caster from pybind to get the DType from either type.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -1260,6 +1260,8 @@ def _is_fa3_supported(num_heads, num_gqa_groups, head_dim_qk, head_dim_v, qkv_dt
     # Filter: cuDNN support
     fused_attention_backend = None
     if use_fused_attention:
+        # ``constants.DType`` is implicitly convertible to ``transformer_engine::DType``
+        # on the C++ side, so pass it straight to the pybind function.
```
ptrendx: Do we need those comments? They do not really give that much value.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -575,10 +575,12 @@ std::tuple<std::vector<py::object>, std::vector<TensorWrapper>> bulk_allocate_fp
     py::object columnwise_scale =
         (columnwise_usage ? py::cast(columnwise_scale_list[i]) : py::none());
 
-    // Construct Python tensor
-    tensor_py_list.emplace_back(
-        Float8BlockwiseQTensorClass(rowwise_data, rowwise_scale, columnwise_data, columnwise_scale,
-                                    fp8_dtype, quantizer_py_list[i], is_2D_scaled));
+    // Construct Python tensor. ``MakePythonDType`` returns the cached
+    // ``constants.DType`` IntEnum member, so ``_fp8_dtype`` ends up as that
+    // IntEnum without a per-call ``DType.cast`` normalization.
```
ptrendx: Same, please clean those comments (especially duplicate ones).

transformer_engine/pytorch/csrc/common.cpp
```diff
@@ -86,6 +86,25 @@ transformer_engine::DType getTransformerEngineFP8Type(bool e4m3_if_hybrid,
   return transformer_engine::DType::kFloat8E5M2;
 }
 
+pybind11::object MakePythonDType(transformer_engine::DType dtype) {
```
ptrendx, line 89: I am slightly worried about the thread safety of this function when there is no GIL.

transformer_engine/pytorch/csrc/common.h
```diff
@@ -392,6 +392,21 @@ std::vector<size_t> getTensorShape(const at::Tensor& t);
...
+ *
+ * The returned object is cached per enum value (one ``py::object`` per
+ * ``DType``), so repeated calls avoid both re-importing the module and
+ * re-running ``IntEnum.__call__``. C++ sites that set tensor/quantizer dtype
+ * attributes (``_fp8_dtype``, the ``fp8_dtype`` / ``fp4_dtype`` kwargs, etc.)
+ * should go through this helper instead of ``py::cast(dtype)``: ``py::cast``
+ * produces a fresh ``tex.DType`` pybind enum object on every call, which the
+ * Python constructors then have to normalize via ``DType.cast``.
+ *
+ * Must be called with the GIL held (always true inside pybind-invoked code).
```
ptrendx: That is not necessarily true with the GIL-less Python becoming a thing.

transformer_engine/pytorch/__init__.py
```diff
@@ -132,7 +133,10 @@
             MXFP8Quantizer,
             NVFP4Quantizer,
             Float8BlockQuantizer,
-            # pybind11 enum used as Quantizer.dtype
+            # Python IntEnum used as Quantizer.dtype.
+            constants.DType,
```
ptrendx: Do we still need to add it here even if it is a regular Python object?

transformer_engine/pytorch/constants.py
```diff
@@ -3,42 +3,82 @@
...
+    kInt32 = int(tex.DType.kInt32)
+    kFloat32 = int(tex.DType.kFloat32)
+    kFloat16 = int(tex.DType.kFloat16)
+    kBFloat16 = int(tex.DType.kBFloat16)
+    kFloat8E4M3 = int(tex.DType.kFloat8E4M3)
+    kFloat8E5M2 = int(tex.DType.kFloat8E5M2)
+    kFloat4E2M1 = int(tex.DType.kFloat4E2M1)
+
+    @classmethod
+    def cast(cls, dtype: "DTypeSupported") -> "DType":
```
ptrendx: DTypeSupported?

transformer_engine/pytorch/constants.py
```diff
@@ -3,42 +3,82 @@
...
+# on the C++ side without being mirrored above.
+assert {m.name for m in DType} == set(tex.DType.__members__), (
+    "DType is out of sync with transformer_engine_torch.DType; "
+    "add the new pybind enumerator to DType in constants.py."
+)
+
+
+# tex.DType is the pybind enum kept for backward compatibility.
+# in the constructors for QuantizedTensors and Quantizers.
+DTypeSupported = Union[DType, tex.DType]
```
ptrendx: Hmmm, why can't you just use this union type directly in the cast function declaration rather than having this indirection?

transformer_engine/pytorch/constants.py
```diff
@@ -3,42 +3,82 @@
...
+
+
+# One-to-one mapping ``torch.dtype -> DType`` (mirrors the enum order in
+# ``transformer_engine.h``). Use the bracket syntax ``TE_DType[torch_dtype]``
+# to resolve a ``torch.dtype`` to its matching ``DType`` member.
+# Used for passing dtypes into cuda extension.
 TE_DType = {
```
ptrendx, line 93: nit: we sometimes need int64

transformer_engine/pytorch/constants.py
```diff
@@ -3,42 +3,82 @@
...
-    tex.DType.kFloat32: torch.float32,
-    tex.DType.kFloat16: torch.half,
-    tex.DType.kBFloat16: torch.bfloat16,
+    DType.kByte: torch.uint8,
+    DType.kFloat8E4M3: torch.float8_e4m3fn,
+    DType.kFloat8E5M2: torch.float8_e5m2,
+    DType.kInt32: torch.int32,
+    DType.kFloat32: torch.float32,
+    DType.kFloat16: torch.half,
+    DType.kBFloat16: torch.bfloat16,
```
ptrendx:
```python
{value: key for key, value in TE_DType.items()}
```

transformer_engine/pytorch/constants.py
vthumbe1503: This is reused in multiple places including Quantizer and QuantizedTensor constructors. So I defined it once here.

transformer_engine/pytorch/tensor/float8_tensor.py
pggPL: Please make sure that the change is visible in the docs.
I think we should explicitly document the dtype enum and change all transformer_engine_torch.DType occurences like above.

transformer_engine/pytorch/constants.py
```diff
@@ -3,42 +3,96 @@
...
+
+    def __hash__(self) -> int:
+        return int.__hash__(self)
+
+
+# Fail fast at import time if a new enumerator is added
+# on the C++ side without being mirrored above.
+assert {m.name for m in DType} == set(tex.DType.__members__), (
+    "DType is out of sync with transformer_engine_torch.DType; "
+    "add the new pybind enumerator to DType in constants.py."
```
pggPL: This is good comment for TE devs, but if someone loads newer/older c++ library version of TE with different dtypes this error may be confusing.

transformer_engine/pytorch/csrc/common.cpp
vthumbe1503: Made it thread safe now. Entire array is initialized once as a static const

transformer_engine/pytorch/constants.py
```diff
@@ -3,42 +3,82 @@
...
+        """Normalize any ``DTypeSupported`` value to the canonical ``DType`` ``IntEnum``.
+        ``DType`` is the canonical dtype tag used internally throughout
+        ``transformer_engine.pytorch``, and is what this function always outputs.
+        The pybind ``transformer_engine_torch.DType`` enum is an additional type
+        accepted as input (for backward compatibility), which this function maps
+        to the matching ``DType`` member so stored attributes are always ``DType``.
+        """
+        if isinstance(dtype, cls):
+            return dtype
+        return cls(int(dtype))
```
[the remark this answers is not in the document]
vthumbe1503: Fixed this now. We have equality comparators defined for both DType and tex.DType to handle this

transformer_engine/pytorch/constants.py
vthumbe1503: Have made the comment better now.

transformer_engine/pytorch/constants.py
vthumbe1503: Done

transformer_engine/pytorch/csrc/extensions/cast.cpp
vthumbe1503: Done

transformer_engine/pytorch/attention/dot_product_attention/utils.py
vthumbe1503: removed

transformer_engine/pytorch/csrc/common.h
vthumbe1503: Now the function is thread safe

transformer_engine/pytorch/constants.py
vthumbe1503: Union is used now, to be more explicit and clear.

transformer_engine/pytorch/tensor/float8_tensor.py
vthumbe1503: Added it in Docs

transformer_engine/pytorch/__init__.py
vthumbe1503: Yes for every python object that is used as an argument for checkpoint loading function needs its class allow listed

transformer_engine/pytorch/constants.py
vthumbe1503: Now using Union directly

transformer_engine/common/util/dtype_pybind_conversion.h
vthumbe1503: Added type caster

transformer_engine/pytorch/csrc/extensions/pybind_dtype_caster.h
timmoon10, line 1: It's uncomfortable to have such PyTorch-specific logic here in the core lib. Some options:
1. Move this into the PyTorch extensions. We don't import this header in JAX, so it's no different than the current implementation. That said, your warning about ODR violations makes me worried about problems in cases we have PyTorch and JAX together. 
2. Rename this to something like `pybind_casters.h` and make sure both the PyTorch and JAX C++ extensions import it everywhere they deal with Pybind11. The header isn't really about dtype conversion, it's making sure we have consistent type casters.
3. Move the `DType` `IntEnum` into `transformer_engine.common`. `transformer_engine.pytorch.constants` can just import from `common`. This is weird for JAX though since it strictly uses `jnp.dtype` in Python.

2 sounds the best to me.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -170,6 +170,32 @@ void init_router_bindings(pybind11::module &m) {
...
+  // Register __eq__/__ne__ on the pybind ``DType`` enum so it compares by integer.
+  py::object dtype_class = m.attr("DType");
+  dtype_class.attr("__eq__") = py::cpp_function(
+      [dtype_class](transformer_engine::DType self, py::object other) -> py::object {
+        if (py::isinstance(other, dtype_class)) {
+          return py::cast(self == other.cast<transformer_engine::DType>());
+        }
+        if (py::isinstance<py::int_>(other)) {
+          return py::cast(static_cast<int>(self) == other.cast<int>());
+        }
```
timmoon10: We expect to compare against `IntEnum` and `int`, both of which can be cast to `int`. Could we simplify?
```suggestion
        return py::cast(static_cast<int>(self) == other.cast<int>());
```
This would remove the need for the custom type casters in `dtype_pybind_conversion.h`.

transformer_engine/pytorch/csrc/extensions/pybind_dtype_caster.h
vthumbe1503: The logic here is not Pytorch specific except for the fact that its used only for pytorch. The comment might be and I ll fix the comment to make sure the Pytorch specific constant is referred to in the context of being an example

cast logic is essentially base class implementation and is just there for the purpose of completeness of implementing both load and cast

load logic --> enables us to pass int/intenum values into tex functions from python and enables  pybind to convert it to transformer_engine::DType accepted by tex C++ API interface

2 sounds good to me too. The header at the moment is specific to transformer_engine::DType, but the src can be any int/int enum. Now in future we can probably add more casters in the same header. So, renaming it to pybind_casters.h is reasonable.

Also can we even have jax and pytorch in the same environment and should we be concerned about the ODR violations about such a use-case? I actually prefer to keep changes limited to pytorch for this.

vthumbe1503: Considering this further, There can be a possibility in future of implementing the cast function to return the Pytorch DType in which case it ll become Pytorch specific.


So I think I ll go with option 1 that you suggested. Moving this to pytorch extension seems like the good way to go for me.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
vthumbe1503: Done

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -170,6 +170,27 @@ void init_router_bindings(pybind11::module &m) {
...
   NVTE_DECLARE_COMMON_PYBIND11_HANDLES(m)
+
+  // Register __eq__/__ne__ on the pybind ``DType`` enum so it compares by integer.
+  py::object dtype_class = m.attr("DType");
+  dtype_class.attr("__eq__") = py::cpp_function(
+      [dtype_class](transformer_engine::DType self, py::object other) -> py::object {
+        return py::cast(static_cast<int>(self) == other.cast<int>());
+        return py::reinterpret_borrow<py::object>(Py_NotImplemented);
+      },
+      py::is_method(dtype_class));
```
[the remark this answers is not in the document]
vthumbe1503: fixed this now

transformer_engine/common/util/dtype_pybind_conversion.h
```diff
@@ -0,0 +1,97 @@
...
+ * Allows a Python object of type ``transformer_engine.pytorch.constants.DType``
+ * to be passed wherever a pybind-bound ``transformer_engine::DType`` argument is expected.
+ * pybind-bound ``transformer_engine::DType`` argument is expected.
+ * Must be called after the pybind ``DType`` enum has been registered.
+ */
```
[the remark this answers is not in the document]
vthumbe1503: resolved

transformer_engine/pytorch/csrc/extensions/pybind_dtype_caster.h
timmoon10, line 1: Nit: This is kind of a strange place for this header since it's not exposed externally in `tex`. That said, the `csrc` dir is generally kind of disorganized right now. Maybe reorganizing would be good for a future PR, something like:

- `common.h`: Header for internal utilities
- `api.h`: Header for functions exposed in `tex`
- `pybind`: Dir with Pybind11 logic
- `api`: Dir with functions exposed in `tex`
- `quantizers`: Dir with quantizer impls
- `utils`: Dir with misc utilities

This is far beyond the scope of this PR though.

timmoon10 (dismissed):
LGTM. Aside from the functional considerations with checkpointing/pickling/torch.compile, this is quite a nice from a design perspective. It was uncomfortable that downstream dependencies needed to import `tex` to do advanced quantizer configuration.

transformer_engine/pytorch/csrc/common.h
```diff
@@ -55,6 +55,7 @@
 
 #include "c10/util/ArrayRef.h"
 #include "common/util/logging.h"
+#include "extensions/pybind_dtype_casters.h"
```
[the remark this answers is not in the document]
vthumbe1503: Fixed

timmoon10 (approved):

ptrendx (approved):

vthumbe1503:
Pipeline: https://gitlab-master.nvidia.com/dl/transformerengine/transformerengine/-/pipelines/53689424

PR #3041, opened 2026-05-25
[PyTorch Debug] Fix scale_inv_min returning 0 for MXFP8/NVFP4

### LINKED ISSUE #2628: debug api scale_inv_min is always zero when using padding or divisibility requirements
The debug_api provides the ability to acquire log information specific to FP8 through FP8TensorStats. However, we noticed that when we add padding (32 for MXFP8 or 16 for NVFP4) to the end of a sequence to make it divisible, the `scale_inv_min` is always zero.

This is because, there exists a consecutive block of 32 0s, which means that one of your `scale_inv` values will be 0, and therefore your `scale_inv_min` will always be zero. 

See this slide

Image

Thus, we are always going to have some zeros in that batch, because its a requirement, which makes `scale_inv_min` a non useful metric for us.

Otherwise, please provide:
- OS version
- PyTorch version
- Python version
- Transformer Engine version
- CUDA version
- CUDNN version

**Device details**
- GPU model

**Additional context**

pggPL:
# Description

  The debug API stat `scale_inv_min` was always reporting `0` for MXFP8 and NVFP4 recipes.

  Root cause: `MXFP8Quantizer.get_scale_shape` and `NVFP4Quantizer.get_scale_shape` round the `scale_inv` shape up to multiples of 128 along one axis and 4 along the other:

  - MXFP8 rowwise:    `[round_up(M, 128), round_up(N/32, 4)]`
  - MXFP8 columnwise: `[round_up(M/32, 4), round_up(N, 128)]`
  - NVFP4 rowwise:    `[round_up(M, 128), round_up(⌈K/16⌉, 4)]`
  - NVFP4 columnwise: `[round_up(K, 128), round_up(⌈M/16⌉, 4)]`

  The padded buffer is allocated upfront at the padded shape (`at::empty(padded_shape)` in `quantizer.cpp`), and the cast kernel issues `cudaMemsetAsync(scale_inv, 0, ...)` over the whole buffer before filling the valid region (see `quantize_mxfp8.cuh:780-794`), so padded
   slots end up as literal zeros. `add_scale_inv_stats` in `stats_computation.py` was calling `.min()` over the whole padded tensor, which therefore always returned `0` whenever the unpadded scale shape did not already meet the alignment.

  A real `scale_inv` entry is never zero: `compute_scale_from_amax` (`recipe_common.cuh`) returns `scale = 1.0` for all-zero data blocks and clamps inf to a finite fallback, so zero in `scale_inv` uniquely identifies padding. Masking zeros before `.min()` is therefore
  exact, not heuristic. `.max()` is unaffected (zero cannot be the maximum of a non-negative tensor) and is left as-is.

  Fixes #2628

  ## Type of change

  - [ ] Infra/Build change
  - [ ] Code refactoring

  ## Changes

  - Add `nonzero_min` helper in `add_scale_inv_stats` (`transformer_engine/debug/features/utils/stats_computation.py`) that masks out padding zeros before computing the minimum of `scale_inv`.
  - Route the `scale_inv_min` per-step lambda through `nonzero_min`; `scale_inv_max` is unchanged.
  - Degenerate all-zero case returns a `0` scalar of the right dtype/device so the buffer aggregator (`min(_get(...))`) stays well-defined.

  # Checklist:

### DIFF
--- a/transformer_engine/debug/features/utils/stats_computation.py
+++ b/transformer_engine/debug/features/utils/stats_computation.py
@@ -348,6 +348,16 @@ def add_scale_inv_stats(recipe_name: str, columnwise: bool = False):
             return getattr(quantized_tensor, "_columnwise_scale_inv")
         return getattr(quantized_tensor, "_rowwise_scale_inv")
 
+    def nonzero_min(scale_inv):
+        # MXFP8/NVFP4 scale_inv is padded to a multiple of [128, 4] (or [4, 128]
+        # for columnwise) with zeros, so a plain .min() always returns 0. Mask
+        # those padding zeros out; if everything is zero (degenerate case) fall
+        # back to 0 so the buffer aggregation stays well-defined.
+        nz = scale_inv[scale_inv != 0]
+        if nz.numel() == 0:
+            return scale_inv.new_zeros(())
+        return nz.min()
+
     columnwise_suffix = "_columnwise" if columnwise else ""
     # Prepare stat names.
     stat_name_min = (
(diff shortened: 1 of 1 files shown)


jomitchellnv (approved):
looks good thanks for fixing

negvet (approved):

PR #3044, opened 2026-05-26
[PyTorch Debug] Add scale_inv_std stat and skip NVFP4 layers in LogFp8TensorStats

### LINKED ISSUE #2801: Request for additional features from TE Debug modules for low precision training.
**Is your feature request related to a problem? Please describe.**

1. For LogFp4Tensor stats and LogFP8TensorStats we want to obtain the standard deviation of the scale factors.
2. LogFp8 and LogFP4 tensor stats can't currently be used at the same time, which is useful for situations when you have some layers as FP8 and some as FP4.
3. [most important] When using MXFP8 or NVFP4 micro-scaling methods we are only obtaining global statistics over the entire tensor instead of per each block.

**Describe the solution you'd like**
1. We want to add the standard deviation metric for the scale factors because simply having min and max doesn't tell us enough information. We could have serious clipping going on but won't know if we only have a max value.
2. Currently, I've had to hack in support by modifying yaml files online in order to perform FP8 and FP4 logging for networks that have mixed FP4/FP8 layers https://github.com/NVIDIA/bionemo-framework/blob/main/bionemo-recipes/recipes/esm2_native_te/quantization.py#L94 I am hoping this can be supported natively.
3. For LogFP8TensorStats and LogFP4TensorStats specifically for micro-block formats such as MXFP8 / NVFP4 we need an optional mechanism to extract block level metadata in addition to per-tensor level metadata including scale factor statistics per block of (scale_value, block_indices, tensor_chunk) etc saved to files. By default this should be able to work on the first block (0:32 for MXFP8, 0:16 for NVFP4) unless otherwise specified etc. It would be super useful in order to identify issues on a more fine grained scale.

**Describe alternatives you've considered**

**Additional context**

pggPL:
# Description

Addresses items 1 and 2 of #2801:

- Adds `scale_inv_std` for delayed/current FP8 scaling, MXFP8, FP8 block scaling and NVFP4, including supported `_columnwise` variants. NVFP4 also gains `scale_inv_min` and `scale_inv_max`.
- Allows `LogFp8TensorStats` and `LogNvfp4TensorStats` to share a layer selection. On NVFP4 layers, the FP8 logger skips bare stats with a warning and preserves FP8-recipe-prefixed what-if stats such as `mxfp8_mse`.

Scale standard deviation uses population variance, so a single scale has `std=0`. Microbatch/rank reduction uses the Chan/Welford combination. This also fixes the existing between-group sign error and preserves the sample-variance convention of the high-precision `variance`/`std` statistics.

Per-block metadata dumping (item 3 of #2801) remains outside this PR.

## Validation

- Blackwell (`gb102`): `tests/pytorch/debug/test_log.py` — **31 passed, 8 skipped**. Tested the changed Python features with TE 2.20.0.dev0 / PyTorch 2.14.0a0 native binaries from the current TE devel image.
- Full `PYTHON_ONLY=1 qa/L0_pytorch_lint/test.sh`: passed. Black 24.4.2: passed.
- Tests cover unequal microbatches, every supported recipe, rowwise/columnwise scales, zero-padded scale storage, and simultaneous FP8/NVFP4 logging.
- The numerical test reference excludes zero scale entries consistently with the logger.
- The high-precision reduction regression checks reduced statistics against concatenated tensors.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Documentation updates


transformer_engine/debug/features/log_nvfp4_tensor_stats.py
```diff
@@ -85,13 +96,18 @@ class LogNvfp4TensorStats(BaseLogTensorStats):
 
     def check_if_stat_is_supported(self, stat: str):
         """Returns True if stat is supported, raises ValueError otherwise."""
+        bare = stat[: -len("_columnwise")] if stat.endswith("_columnwise") else stat
```
ptrendx: mse_columnwise and underflow%_columnwise should not pass this function but they would with the logic as written.

pggPL: I added them to exception, even though we should have stats for them I guess, but this is not scope of the PR.

PR #3045, opened 2026-05-26
Add NVFP4 per-token quantization recipe

cael-ling:
# Description

This PR adds an NVFP4 per-token quantization recipe. The default NVFP4BlockScaling recipe computes a single per-tensor outer amax (s_global) per tensor. The per-token variant instead computes a per-row outer amax (length M) for rowwise data and a per-col outer amax (length K) for columnwise data, giving each token/row its own global scale. 

# Changes

- Per-token cast kernels: vector-amax + encode/swizzle producing NVFP4 tensors whose _amax_rowwise / _amax_columnwise are per-row/per-col vectors.
- CUTLASS GEMM (nvfp4_cutlass_per_token_gemm) that rescales with the per-row/per-col outer-amax vectors inside the epilogue;
- Forward + backward coverage (dgrad NN / wgrad NT layouts).
- NVFP4PerTokenBlockScaling recipe (re-exported from transformer_engine.pytorch.fp8), plus an equivalent NVTE_NVFP4_PER_TOKEN=1 env-var switch on a plain NVFP4BlockScaling so frameworks that only build a default recipe (e.g. Megatron-Core) can opt in with no code change.
- Opt-in RHT / SR (per_token_rht / per_token_sr) — off by default on the per-token path.
- Opt-in 2D weight quantization (per_token_weight_2d): transposition-invariant 16×16 cast emitted in per-token layout.
- Docs: API reference entry, NVTE_NVFP4_* env-var docs, and a "Per-token NVFP4" feature section with Megatron-Core launch instructions.
- Example: examples/pytorch/nvfp4_per_token_megatron — single-GPU MoE example comparing per-token vs per-tensor vs BF16 with identical model/data/seed.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -287,6 +287,48 @@ Kernel Configuration
    :Default: ``0``
    :Description: Enable row-scaled NVFP4 tensors for forward activation quantizers in the ``NVFP4BlockScaling`` recipe. When set to ``1`` (or when ``NVFP4BlockScaling(row_scaled_activation=True)`` is used), rowwise ``amax`` metadata is stored as one FP32 value per tensor row instead of a single scalar.
 
+.. envvar:: NVTE_NVFP4_DISABLE_RHT
+
+   :Type: ``int`` (0 or 1)
+   :Default: ``0``
+   :Description: Opt out of the random Hadamard transform (RHT) in the per-tensor ``NVFP4BlockScaling`` recipe. RHT is applied by default to the forward activation and backward gradient quantizers. Set to ``1`` (or use ``NVFP4BlockScaling(disable_rht=True)``) to disable it. No effect on the per-token path (see :envvar:`NVTE_NVFP4_PER_TOKEN_RHT`).
+
+.. envvar:: NVTE_NVFP4_DISABLE_STOCHASTIC_ROUNDING
+
+   :Type: ``int`` (0 or 1)
+   :Default: ``0``
+   :Description: Opt out of stochastic rounding (SR) in the per-tensor ``NVFP4BlockScaling`` recipe. SR is applied by default to the backward gradient quantizer. Set to ``1`` (or use ``NVFP4BlockScaling(disable_stochastic_rounding=True)``) to disable it. No effect on the per-token path (see :envvar:`NVTE_NVFP4_PER_TOKEN_SR`).
+
+.. envvar:: NVTE_NVFP4_DISABLE_2D_QUANTIZATION
+
+   :Type: ``int`` (0 or 1)
+   :Default: ``0``
+   :Description: Opt out of 2D (16x16 inner tile + scalar outer amax) weight quantization in the per-tensor ``NVFP4BlockScaling`` recipe. 2D weight quantization is enabled by default. Set to ``1`` (or use ``NVFP4BlockScaling(disable_2d_quantization=True)``) to fall back to 1D (16-element block) weight quantization. Forced on the per-token path (the per-token cast hard-disables 2D); see :envvar:`NVTE_NVFP4_PER_TOKEN_WEIGHT_2D` for the per-token weight-2D route.
+
+.. envvar:: NVTE_NVFP4_PER_TOKEN
+
+   :Type: ``int`` (0 or 1)
+   :Default: ``0``
+   :Description: Flip a plain ``NVFP4BlockScaling`` recipe into per-token mode (per-row / per-col outer ``amax`` cast plus the fused-EVT CUTLASS GEMM) without changing the recipe class. This lets frameworks that already construct a default ``NVFP4BlockScaling`` (e.g. Megatron-Core with ``--fp4-format e2m1``) opt into per-token purely from the launch environment. Equivalent to constructing the explicit ``NVFP4PerTokenBlockScaling`` recipe. The per-token forward path currently requires the unfused norm+amax path: also set ``NVTE_NORM_FWD_USE_CUDNN=1`` (the fused norm+amax path rejects per-token quantizers).
+
+.. envvar:: NVTE_NVFP4_PER_TOKEN_RHT
+
+   :Type: ``int`` (0 or 1)
+   :Default: ``0``
+   :Description: Per-token only. Opt into the random Hadamard transform (RHT) on the per-token forward activation and backward gradient quantizers. Per-token disables RHT by default (its per-row outer amax already mitigates the long-tail outliers RHT targets); set to ``1`` (or use ``NVFP4PerTokenBlockScaling(per_token_rht=True)``) to re-enable it. No effect on the per-tensor path.
+
+.. envvar:: NVTE_NVFP4_PER_TOKEN_SR
+
+   :Type: ``int`` (0 or 1)
+   :Default: ``0``
+   :Description: Per-token only. Opt into stochastic rounding (SR) on the per-token backward gradient quantizer (the K2 encode kernel implements a Philox-dithered FP4 cast). Per-token disables SR by default; set to ``1`` (or use ``NVFP4PerTokenBlockScaling(per_token_sr=True)``) to re-enable it. No effect on the per-tensor path.
+
+.. envvar:: NVTE_NVFP4_PER_TOKEN_WEIGHT_2D
+
+   :Type: ``int`` (0 or 1)
+   :Default: ``0``
+   :Description: Per-token only. Quantize the forward weight with the per-tensor 2D cast (16x16 inner tile + scalar outer amax) emitted in per-token layout, instead of the per-token 1D weight cast. 2D weight quantization is transposition-invariant, so forward (rowwise) and dgrad (columnwise) see the same weight, removing the 1D path's weight-gradient bias. Activations and gradients stay on the standard per-token 1D cast. Set to ``1`` (or use ``NVFP4PerTokenBlockScaling(per_token_weight_2d=True)``). No effect on the per-tensor path.
+
 .. envvar:: NVTE_NVFP4_4OVER6
 
    :Type: ``str`` (``none``, ``weights``, ``activations``, or ``all``)
--- a/docs/envvars.rst
+++ b/docs/envvars.rst
@@ -309,7 +309,7 @@ Kernel Configuration
 
    :Type: ``int`` (0 or 1)
    :Default: ``0``
-   :Description: Flip a plain ``NVFP4BlockScaling`` recipe into per-token mode (per-row / per-col outer ``amax`` cast plus the fused-EVT CUTLASS GEMM) without changing the recipe class. This lets frameworks that already construct a default ``NVFP4BlockScaling`` (e.g. Megatron-Core with ``--fp4-format e2m1``) opt into per-token purely from the launch environment. Equivalent to constructing the explicit ``NVFP4PerTokenBlockScaling`` recipe. The per-token forward path currently requires the unfused norm+amax path: also set ``NVTE_NORM_FWD_USE_CUDNN=1`` (the fused norm+amax path rejects per-token quantizers).
+   :Description: Flip a plain ``NVFP4BlockScaling`` recipe into per-token mode (per-row / per-col outer ``amax`` cast plus the fused-EVT CUTLASS GEMM) without changing the recipe class. This lets frameworks that already construct a default ``NVFP4BlockScaling`` (e.g. Megatron-Core with ``--fp4-format e2m1``) opt into per-token purely from the launch environment. Equivalent to constructing the explicit ``NVFP4PerTokenBlockScaling`` recipe.
 
 .. envvar:: NVTE_NVFP4_PER_TOKEN_RHT
 
--- a/docs/features/low_precision_training/nvfp4/nvfp4.rst
+++ b/docs/features/low_precision_training/nvfp4/nvfp4.rst
@@ -250,13 +250,15 @@ There are two ways to select per-token, both equivalent:
 * ``row_scaled_activation`` and 4over6 are forced off (mutually exclusive with
   the per-token amax layout).
 
-**Requirement: unfused norm forward.** The per-token forward path requires the
-unfused norm+amax implementation; the fused norm+amax path rejects per-token
-quantizers. When the first GEMM consumes a fused norm output (for example
-``LayerNormLinear``), also set ``NVTE_NORM_FWD_USE_CUDNN=1``.
+**Norm forward.** The per-token forward path uses the unfused norm+amax
+implementation. Its per-row/per-col outer amax is computed by the per-token
+cast's first kernel (K1, the amax pass), a per-row ``(M,)`` + per-col ``(K,)``
+vector that the fused norm+amax kernel (which only emits a scalar amax) cannot
+produce. This is selected automatically for per-token quantizers -- no
+``NVTE_NORM_FWD_USE_CUDNN=1`` needed.
 
-**Currently unsupported on the per-token path**: ``fuse_wgrad_accumulation=True``,
-forward/backward output quantization, and communication/bulk overlap.
+**Currently unsupported on the per-token path**: forward/backward output
+quantization, and communication/bulk overlap.
 
 Running per-token NVFP4 with Megatron-Core
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
@@ -269,7 +271,6 @@ environment variables. A minimal launch looks like:
 
    # Select NVFP4 per-token via TE env vars (read at recipe construction).
    export NVTE_NVFP4_PER_TOKEN=1
-   export NVTE_NORM_FWD_USE_CUDNN=1          # required: unfused norm forward
    # Optional per-token knobs:
    # export NVTE_NVFP4_PER_TOKEN_RHT=1
    # export NVTE_NVFP4_PER_TOKEN_SR=1
--- a/docs/features/low_precision_training/nvfp4/nvfp4.rst
+++ b/docs/features/low_precision_training/nvfp4/nvfp4.rst
@@ -207,6 +207,89 @@ NVFP4 all-gather is supported.
 
 *Figure 6. Quantization and all-gather flow for NVFP4 showing amax synchronization and hierarchical scaling.*
 
+Per-token NVFP4
+---------------
+
+The default ``NVFP4BlockScaling`` recipe computes a single per-tensor outer
+``amax`` (``s_global``) for each tensor. The **per-token** variant instead
+computes a per-row outer ``amax`` (length ``M``) for rowwise data and a per-col
+outer ``amax`` (length ``K``) for columnwise data, giving each token/row its own
+global scale. This finer outer-scale granularity can improve accuracy, and the
+per-token cast feeds a dedicated fused-EVT CUTLASS GEMM that consumes the vector
+outer ``amax`` directly (cuBLASLt cannot).
+
+There are two ways to select per-token, both equivalent:
+
+* **Explicit recipe class** ``NVFP4PerTokenBlockScaling`` (recommended for code
+  that constructs its own recipe).
+* **Environment variable** ``NVTE_NVFP4_PER_TOKEN=1`` on a plain
+  ``NVFP4BlockScaling``. This lets frameworks that only ever build a default
+  ``NVFP4BlockScaling`` (for example Megatron-Core) opt into per-token purely
+  from the launch environment, with no framework-side code change.
+
+.. code-block:: python
+
+   from transformer_engine.common.recipe import NVFP4PerTokenBlockScaling
+   import transformer_engine.pytorch as te
+
+   # RHT and SR are OFF by default on the per-token path; opt in as needed.
+   recipe = NVFP4PerTokenBlockScaling(per_token_rht=True, per_token_sr=True)
+   with te.fp8_autocast(enabled=True, fp8_recipe=recipe):
+       out = model(inp)
+
+**Differences from the per-tensor default**
+
+* RHT and stochastic rounding are **off by default** on the per-token path (the
+  per-row outer ``amax`` already mitigates the long-tail outliers RHT targets).
+  Opt in with ``per_token_rht=True`` / ``per_token_sr=True`` (env vars
+  :envvar:`NVTE_NVFP4_PER_TOKEN_RHT` / :envvar:`NVTE_NVFP4_PER_TOKEN_SR`).
+* 2D weight quantization is disabled by default. The per-token weight-2D route
+  (``per_token_weight_2d=True`` / :envvar:`NVTE_NVFP4_PER_TOKEN_WEIGHT_2D`)
+  quantizes the forward weight with the transposition-invariant 2D cast emitted
+  in per-token layout, removing the 1D weight-gradient bias.
+* ``row_scaled_activation`` and 4over6 are forced off (mutually exclusive with
+  the per-token amax layout).
+
+**Requirement: unfused norm forward.** The per-token forward path requires the
+unfused norm+amax implementation; the fused norm+amax path rejects per-token
+quantizers. When the first GEMM consumes a fused norm output (for example
+``LayerNormLinear``), also set ``NVTE_NORM_FWD_USE_CUDNN=1``.
+
+**Currently unsupported on the per-token path**: ``fuse_wgrad_accumulation=True``,
+forward/backward output quantization, and communication/bulk overlap.
+
+Running per-token NVFP4 with Megatron-Core
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Megatron-Core builds a plain ``NVFP4BlockScaling`` for ``--fp4-format e2m1`` and
+has no CLI for per-token, so per-token is selected entirely through TE
+environment variables. A minimal launch looks like:
+
+.. code-block:: bash
+
+   # Select NVFP4 per-token via TE env vars (read at recipe construction).
+   export NVTE_NVFP4_PER_TOKEN=1
+   export NVTE_NORM_FWD_USE_CUDNN=1          # required: unfused norm forward
+   # Optional per-token knobs:
+   # export NVTE_NVFP4_PER_TOKEN_RHT=1
+   # export NVTE_NVFP4_PER_TOKEN_SR=1
+   # export NVTE_NVFP4_PER_TOKEN_WEIGHT_2D=1
+
+   python pretrain_gpt.py \
+       --transformer-impl transformer_engine \
+       --fp4-format e2m1 \
+       --no-gradient-accumulation-fusion \
+       ...   # remaining model / data / optimizer args
+
+Notes:
+
+* ``--no-gradient-accumulation-fusion`` is required because the per-token kernel
+  does not yet support fused wgrad accumulation.
+* To keep the first/last transformer layers in BF16, use Megatron's
+  ``--first-last-layers-bf16 --num-layers-at-start-in-bf16 N
+  --num-layers-at-end-in-bf16 M`` flags (those layers simply skip the FP4
+  autocast; the recipe is unchanged).
+
 Examples
 --------
 
--- a/examples/pytorch/nvfp4_per_token_megatron/README.md
+++ b/examples/pytorch/nvfp4_per_token_megatron/README.md
@@ -18,7 +18,6 @@ recipe is constructed:
 | Variable | Effect |
 | --- | --- |
 | `NVTE_NVFP4_PER_TOKEN=1` | **Required**: Flip the recipe into per-token mode (per-row/per-col outer amax + fused CUTLASS GEMM) |
-| `NVTE_NORM_FWD_USE_CUDNN=1` | **Required** with per-token: forces the unfused norm forward (the fused norm+amax path rejects per-token currently) |
 | `NVTE_NVFP4_PER_TOKEN_RHT=1` | Opt into the random Hadamard transform (off by default) |
 | `NVTE_NVFP4_PER_TOKEN_SR=1` | Opt into stochastic rounding (off by default) |
 | `NVTE_NVFP4_PER_TOKEN_WEIGHT_2D=1` | Use the transposition-invariant 2D weight cast in per-token layout |
@@ -122,10 +121,6 @@ The per-token recipe is currently intended for **accuracy evaluation and
 comparison** (per-token vs per-tensor vs BF16), **not** for optimized production
 deployment. Concretely:
 
-- **Requires `NVTE_NORM_FWD_USE_CUDNN=1`** (the unfused cuDNN norm forward).
-  The fused norm+amax path (`NVTE_NORM_FWD_USE_CUDNN=0`, the default) does **not**
-  support per-token and is rejected at the C++ quantizer. The launcher sets this
-  for you in `pertoken` mode.
 - **Not tested with CUDA graphs.** The per-token path has not been validated under
   Megatron's CUDA graph capture; leave CUDA graphs disabled for now.
 - **Kernels are not yet performance-optimal.** Several per-token cast / GEMM
--- a/examples/pytorch/nvfp4_per_token_megatron/README.md
+++ b/examples/pytorch/nvfp4_per_token_megatron/README.md
@@ -1,0 +1,140 @@
+# NVFP4 per-token training with Megatron-Core
+
+This example shows how to train a small Mixture-of-Experts (MoE) model with
+the **NVFP4 per-token** quantization recipe on a single GPU using
+[Megatron-Core](https://github.com/NVIDIA/Megatron-LM), and how to compare it
+against the per-tensor NVFP4 recipe and an unquantized BF16 baseline.
+
+The same model / data / seed are used across all modes; only the GEMM precision
+changes, so the runs are directly comparable.
+
+## How per-token interacts with Megatron-Core
+
+Megatron-Core builds a plain `transformer_engine.common.recipe.NVFP4BlockScaling`
+for `--fp4-format e2m1` and has **no CLI flag for per-token**. Per-token is
+selected entirely through Transformer Engine environment variables, read when the
+recipe is constructed:
+
+| Variable | Effect |
+| --- | --- |
+| `NVTE_NVFP4_PER_TOKEN=1` | **Required**: Flip the recipe into per-token mode (per-row/per-col outer amax + fused CUTLASS GEMM) |
+| `NVTE_NORM_FWD_USE_CUDNN=1` | **Required** with per-token: forces the unfused norm forward (the fused norm+amax path rejects per-token currently) |
+| `NVTE_NVFP4_PER_TOKEN_RHT=1` | Opt into the random Hadamard transform (off by default) |
+| `NVTE_NVFP4_PER_TOKEN_SR=1` | Opt into stochastic rounding (off by default) |
+| `NVTE_NVFP4_PER_TOKEN_WEIGHT_2D=1` | Use the transposition-invariant 2D weight cast in per-token layout |
+
+For the per-tensor recipe, the analogous knobs are
+`NVTE_NVFP4_DISABLE_RHT`, `NVTE_NVFP4_DISABLE_STOCHASTIC_ROUNDING`, and
+`NVTE_NVFP4_DISABLE_2D_QUANTIZATION`.
+
+See the
+[NVFP4 documentation](../../../docs/features/low_precision_training/nvfp4/nvfp4.rst)
+("Per-token NVFP4") and `docs/envvars.rst` for full details. Equivalently, code
+that constructs its own recipe can use the public
+`transformer_engine.common.recipe.NVFP4PerTokenBlockScaling` class instead of the
+env var.
+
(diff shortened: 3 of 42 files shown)


transformer_engine/common/cast/nvfp4/quantize_nvfp4_per_token.cu
```diff
@@ -0,0 +1,1192 @@
...
+
+  alignas(64) CUtensorMap tmap_in{};
+  create_2D_tensor_map(tmap_in, input.data, M, K, TILE_DIM_Y, TILE_DIM_X, K, 0, sizeof(IType) * 8);
+
+  constexpr int buff_elems_total_in = BUFFS_NUM * BUFF_IN_SIZE;
+  constexpr int buff_size_aligned_in =
+      DIVUP_TO_MULTIPLE(buff_elems_total_in * sizeof(IType), TMA_SHMEM_ALIGNMENT);
+  constexpr int dshmem_size = buff_size_aligned_in + TMA_SHMEM_ALIGNMENT;  // + align pad
+
+  dim3 grid(static_cast<unsigned>(K / CHUNK_DIM_X), static_cast<unsigned>(M / CHUNK_DIM_Y), 1);
```
qijiaxing, line 936: maybe use `DIVUP` here to handle the remainder case?

transformer_engine/common/cast/nvfp4/quantize_nvfp4_per_token.cu
```diff
@@ -0,0 +1,1192 @@
...
+//   tid t in [0, 128):
+//     row partial: max over (cols 0..127) for row (row_base + t)
+//     col partial: max over (rows 0..127) for col (col_base + t)
+//   For each 64x64 sub-tile in stage (stage_Y, stage_X):
+//     if t in [stage_Y*64, stage_Y*64+64):  scan 64 cols of sub-tile for row t
+//     if t in [stage_X*64, stage_X*64+64):  scan 64 rows of sub-tile for col t
+// After all 4 stages, emit one atomicMaxFloat per row slot + one per col slot.
+//
+// kWithRht=true: col-wise amax over RHT-rotated 16-row strips (per-thread
+// FHT with random_sign_mask_t). Row direction never sees RHT.
```
qijiaxing: typo: `Row direction never sees RHT` -> `Row direction never uses RHT`

transformer_engine/common/cast/nvfp4/quantize_nvfp4_per_token.cu
```diff
@@ -0,0 +1,1192 @@
...
+  }
+
+  if (leading_thread) {
+#pragma unroll
+    for (int buff = 0; buff < BUFFS_NUM; ++buff) {
+      ptx::mbarrier_invalid(&IN_buff_readable_mbar[buff]);
+    }
+  }
+#else
+  NVTE_DEVICE_ERROR("Per-token amax kernel requires SM 10.0+ (Blackwell).");
```
qijiaxing, line 889: For these quantization kernel, TMA only require SM 9.0+ only. Is there any other constraints that limit to sm 10.0+?

transformer_engine/common/cast/nvfp4/quantize_nvfp4_per_token.cu
cael-ling: Good catch

transformer_engine/common/cast/nvfp4/quantize_nvfp4_per_token.cu
cael-ling: The __CUDA_ARCH__ >= 1000 guard is intentional but not because of a hardware op in this kernel. Two reasons:

1. The shared TE PTX wrappers it calls — cp_async_bulk_tensor_2d_global_to_shared and mbarrier_wait_parity_acquire_cta_shared_cta in util/ptx.cuh — are themselves guarded to >= 1000 and emit NVTE_DEVICE_ERROR below that. They were authored/validated only for the Blackwell path.
2. The whole NVFP4 quantize path is host-gated to SM100 anyway (NVTE_ERROR("NVFP4 requires SM100 ...")), since NVFP4 is a Blackwell datatype and the downstream FP4 GEMM that consumes these scales only exists on SM100. So the amax kernel is never launched off <SM100; the per-arch guard just yields a clean error instead of an undefined symbol.

transformer_engine/common/cast/nvfp4/quantize_nvfp4_per_token.cu
cael-ling: This fast path has a hard precondition that M and K are exact multiples of CHUNK_DIM (128): validate() does NVTE_CHECK(M % CHUNK_DIM_Y == 0) / NVTE_CHECK(K % CHUNK_DIM_X == 0), and is_supported() returns false unless both hold — so any non-multiple shape is rejected / routed to the generic per-token fallback before it ever reaches this launcher.

docs/envvars.rst
```diff
@@ -287,6 +287,48 @@ Kernel Configuration
...
+
+   :Type: ``int`` (0 or 1)
+   :Default: ``0``
+   :Description: Per-token only. Opt into stochastic rounding (SR) on the per-token backward gradient quantizer (the K2 encode kernel implements a Philox-dithered FP4 cast). Per-token disables SR by default; set to ``1`` (or use ``NVFP4PerTokenBlockScaling(per_token_sr=True)``) to re-enable it. No effect on the per-tensor path.
+
+.. envvar:: NVTE_NVFP4_PER_TOKEN_WEIGHT_2D
+
+   :Type: ``int`` (0 or 1)
+   :Default: ``0``
+   :Description: Per-token only. Quantize the forward weight with the per-tensor 2D cast (16x16 inner tile + scalar outer amax) emitted in per-token layout, instead of the per-token 1D weight cast. 2D weight quantization is transposition-invariant, so forward (rowwise) and dgrad (columnwise) see the same weight, removing the 1D path's weight-gradient bias. Activations and gradients stay on the standard per-token 1D cast. Set to ``1`` (or use ``NVFP4PerTokenBlockScaling(per_token_weight_2d=True)``). No effect on the per-tensor path.
```
ptrendx, line 330: While those environment variables are great for debugging and development purposes, we should strive
to not have them (or at least limit them) in the actual production software and instead rely on the
regular options in Python to limit the surprising behavior situations where a left-over env variable
affects the execution negatively.

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -207,6 +207,90 @@ NVFP4 all-gather is supported.
...
 *Figure 6. Quantization and all-gather flow for NVFP4 showing amax synchronization and hierarchical scaling.*
 
+Per-token NVFP4
+---------------
+
+The default ``NVFP4BlockScaling`` recipe computes a single per-tensor outer
+``amax`` (``s_global``) for each tensor. The **per-token** variant instead
+computes a per-row outer ``amax`` (length ``M``) for rowwise data and a per-col
+outer ``amax`` (length ``K``) for columnwise data, giving each token/row its own
+global scale. This finer outer-scale granularity can improve accuracy, and the
```
ptrendx, line 217: We should not give such statements ("can improve accuracy") without many more experiments. I would
make this statement weaker saying that "it may improve accuracy in some situations".

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -207,6 +207,90 @@ NVFP4 all-gather is supported.
 
 *Figure 6. Quantization and all-gather flow for NVFP4 showing amax synchronization and hierarchical scaling.*
 
+Per-token NVFP4
+---------------
+
+The default ``NVFP4BlockScaling`` recipe computes a single per-tensor outer
+``amax`` (``s_global``) for each tensor. The **per-token** variant instead
+computes a per-row outer ``amax`` (length ``M``) for rowwise data and a per-col
+outer ``amax`` (length ``K``) for columnwise data, giving each token/row its own
```
ptrendx, line 216: While M and K are kind of standard GEMM terminology, they are not part of this document earlier so
we should not just randomly use them here. I believe just omitting those is OK here.

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -207,6 +207,90 @@ NVFP4 all-gather is supported.
...
+global scale. This finer outer-scale granularity can improve accuracy, and the
+per-token cast feeds a dedicated fused-EVT CUTLASS GEMM that consumes the vector
+outer ``amax`` directly (cuBLASLt cannot).
```
ptrendx, line 219: This is user-facing guide and so I don't think we should provide here the details about which kernel
we are using for that, since it is not really relevant for that audience.

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -207,6 +207,90 @@ NVFP4 all-gather is supported.
...
+
+There are two ways to select per-token, both equivalent:
+
+* **Explicit recipe class** ``NVFP4PerTokenBlockScaling`` (recommended for code
+  that constructs its own recipe).
+* **Environment variable** ``NVTE_NVFP4_PER_TOKEN=1`` on a plain
+  ``NVFP4BlockScaling``. This lets frameworks that only ever build a default
+  ``NVFP4BlockScaling`` (for example Megatron-Core) opt into per-token purely
+  from the launch environment, with no framework-side code change.
```
ptrendx, line 228: The fact that you can do it suggests that you probably don't actually need it to be its own class and
it probably could just be the option to NVFP4BlockScaling recipe. Not sure yet how you handle the
tensor though so maybe the class is needed. Anyway, having both of these is confusing and should be
avoided - please choose one and also make a PR to MCore to adopt the new recipe.

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -207,6 +207,90 @@ NVFP4 all-gather is supported.
...
+
+.. code-block:: python
+
+   from transformer_engine.common.recipe import NVFP4PerTokenBlockScaling
+   import transformer_engine.pytorch as te
+
+   # RHT and SR are OFF by default on the per-token path; opt in as needed.
+   recipe = NVFP4PerTokenBlockScaling(per_token_rht=True, per_token_sr=True)
+   with te.fp8_autocast(enabled=True, fp8_recipe=recipe):
```
ptrendx, line 237: fp8_autocast is deprecated, use te.autocast instead.

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -207,6 +207,90 @@ NVFP4 all-gather is supported.
...
+       out = model(inp)
+
+**Differences from the per-tensor default**
+
+* RHT and stochastic rounding are **off by default** on the per-token path (the
+  per-row outer ``amax`` already mitigates the long-tail outliers RHT targets).
```
ptrendx, line 243: This remark is not true - the outer scale granularity and RHT are nearly completely orthogonal,
since RHT effect applies to the values inside the block of 16 and that has its own scale. So RHT
should still be able to help the accuracy in this recipe. Adding @negvet - we could try running some
experiments with this recipe to see what the effect of RHT would be here.

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -207,6 +207,90 @@ NVFP4 all-gather is supported.
...
+  Opt in with ``per_token_rht=True`` / ``per_token_sr=True`` (env vars
+  :envvar:`NVTE_NVFP4_PER_TOKEN_RHT` / :envvar:`NVTE_NVFP4_PER_TOKEN_SR`).
+* 2D weight quantization is disabled by default. The per-token weight-2D route
+  (``per_token_weight_2d=True`` / :envvar:`NVTE_NVFP4_PER_TOKEN_WEIGHT_2D`)
+  quantizes the forward weight with the transposition-invariant 2D cast emitted
+  in per-token layout, removing the 1D weight-gradient bias.
```
ptrendx, line 249: I don't quite understand this sentence - what does "2D cast emitted in per-token layout" mean here?

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -207,6 +207,90 @@ NVFP4 all-gather is supported.
...
+* ``row_scaled_activation`` and 4over6 are forced off (mutually exclusive with
+  the per-token amax layout).
+
+**Norm forward.** The per-token forward path uses the unfused norm+amax
+implementation. Its per-row/per-col outer amax is computed by the per-token
+cast's first kernel (K1, the amax pass), a per-row ``(M,)`` + per-col ``(K,)``
+vector that the fused norm+amax kernel (which only emits a scalar amax) cannot
+produce. This is selected automatically for per-token quantizers -- no
+``NVTE_NORM_FWD_USE_CUDNN=1`` needed.
```
ptrendx, line 258: I believe this paragraph is not necessary in the user guide.

docs/features/low_precision_training/nvfp4/nvfp4.rst
```diff
@@ -207,6 +207,90 @@ NVFP4 all-gather is supported.
...
+       ...   # remaining model / data / optimizer args
+
+Notes:
+
+* ``--no-gradient-accumulation-fusion`` is required because the per-token kernel
+  does not yet support fused wgrad accumulation.
+* To keep the first/last transformer layers in BF16, use Megatron's
+  ``--first-last-layers-bf16 --num-layers-at-start-in-bf16 N
+  --num-layers-at-end-in-bf16 M`` flags (those layers simply skip the FP4
+  autocast; the recipe is unchanged).
```
ptrendx, line 292: Could we have that example in MCore PR instead?

examples/pytorch/nvfp4_per_token_megatron/README.md
```diff
@@ -0,0 +1,135 @@
...
+
+The same model / data / seed are used across all modes; only the GEMM precision
+changes, so the runs are directly comparable.
+
+## How per-token interacts with Megatron-Core
+
+Megatron-Core builds a plain `transformer_engine.common.recipe.NVFP4BlockScaling`
+for `--fp4-format e2m1` and has **no CLI flag for per-token**. Per-token is
+selected entirely through Transformer Engine environment variables, read when the
+recipe is constructed:
```
ptrendx, line 16: Considering those caveats, could we actually not include this example in this PR and instead make a
followup PR with the example once MCore support becomes available?

tests/pytorch/nvfp4/bench_nvfp4_per_token.py
```diff
@@ -0,0 +1,2173 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
ptrendx, line 1: We have benchmarks/ directory in the repository where such files should land, rather than in tests/.

tests/pytorch/nvfp4/bench_nvfp4_per_token_group.py
```diff
@@ -0,0 +1,460 @@
+"""Bench NVFP4 per-token grouped K1+K2 quant vs per-tensor RHT+SR baseline.
```
ptrendx, line 1: Same comment about the wrong directory. Also, this file lacks the Copyright and license information.

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -140,6 +140,14 @@ class NVFP4Quantizer(Quantizer):
     rht_matrix_random_sign_mask_t: int
     rht_matrix: torch.Tensor
 
+    """Per-token NVFP4 cast (replaces the per-tensor 1A / 2A paths)."""
```
ptrendx, line 143: What are 1A/2A paths?

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -140,6 +140,14 @@ class NVFP4Quantizer(Quantizer):
...
+    per_token: bool
+
+    """Per-token weight only: emit a per-tensor 2D-quantized weight (16x16 inner +
+    scalar outer amax) dressed in per-token layout. See quantize_impl in
+    csrc/quantizer.cpp and recipe.per_token_weight_2d."""
+    per_token_weight_2d: bool
```
ptrendx, line 149: How is this different from the other 2D quantization option? Why can't we reuse that one instead of
creating a new one?

transformer_engine/pytorch/quantization.py
```diff
@@ -1701,22 +1701,64 @@ def _make(tensor_type: str) -> NVFP4Quantizer:
...
+            #
+            #   forward:  input / output / weight  -> per-token cast
+            #   backward: grad_output / grad_input -> per-token cast
+            #
+            # Per-token bwd uses the same cast kernel as fwd (per-row + per-
+            # col vector amaxes) and dispatches dgrad/wgrad through the same
+            # nvte_nvfp4_cutlass_per_token_gemm by selecting rowwise vs
+            # columnwise quantized data + amax at the GEMM call site (see
+            # _nvfp4_per_token_gemm in cpp_extensions/gemm.py for the NN/NT
+            # layout dispatch table).
```
ptrendx, line 1715: Comments like this very easily become stale - please avoid them.

transformer_engine/pytorch/custom_recipes/quantization_nvfp4_per_token.py
```diff
@@ -0,0 +1,422 @@
...
+    ``random_sign_mask_t`` low 16 bits = sign pattern (default ``0xACE1``).
+
+    ``with_swizzle=True``: rowwise ``scale_inv`` in cuBLAS LT layout
+    (colwise stays compact). Downstream ``nvfp4_per_token_gemm`` must
+    use ``sf_swizzled=True`` to skip its built-in swizzle.
+
+    Raises ``ValueError`` on non-bf16 input or non-128-aligned shapes.
+    """
+    # Import lazily so the module does not require the binary at import time.
+    # (Mirrors the pattern in ``gemm_nvfp4_blocked.py``.)
```
ptrendx, line 232: I don't believe such file exists anywhere.

transformer_engine/pytorch/custom_recipes/quantization_nvfp4_per_token.py
```diff
@@ -0,0 +1,422 @@
...
+def nvfp4_per_token_quantize(
+    x: torch.Tensor,
+    *,
+    rowwise: bool = True,
+    columnwise: bool = False,
+    with_rht: bool = False,
+    random_sign_mask_t: int = 0xACE1,
+    with_swizzle: bool = False,
+) -> RefNVFP4TensorPerToken:
+    """Production NVFP4 per-token cast through ``tex.nvfp4_per_token_quantize``.
```
ptrendx, line 219: There is a big misunderstanding here - custom recipes are meant to show how a user can extend TE with
their own completely custom implementations of quantization and GEMM and, for this purpose, store
only basic pyTorch-only reference implementations of some recipes as examples. Actual production-grade
quantization logic should be in a completely different place (tex.quantize).

transformer_engine/pytorch/custom_recipes/__init__.py
```diff
@@ -3,3 +3,35 @@
 # See LICENSE for license information.
 
 """Experimental features and APIs."""
+
```
ptrendx, line 6: This file should not expose anything - please remove this code.

transformer_engine/pytorch/custom_recipes/quantization_nvfp4_per_token_group.py
```diff
@@ -0,0 +1,122 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""Grouped (multi-tensor) NVFP4 per-token quantize Python wrapper.
+
+Dispatches through ``tex.nvfp4_per_token_group_quantize_bulk`` -- the bulk
+C++ binding owns allocation, view-slicing, and the composite K1+K2 kernel
+dispatch. Requires bf16 input with K and every split_sections[i] a multiple
+of 128; up to 64 splits.
```
ptrendx, line 10: If so it should not be in custom recipes.

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1870,6 +1871,32 @@ NVFP4Quantizer::NVFP4Quantizer(const py::handle& quantizer) : Quantizer(quantize
 
   this->rht_matrix_random_sign_mask_t = quantizer.attr("rht_matrix_random_sign_mask_t").cast<int>();
   this->rht_matrix = quantizer.attr("rht_matrix").cast<at::Tensor>();
+
+  // Per-token NVFP4 (replaces the per-tensor 1A / 2A paths). Default False keeps
+  // the per-tensor path byte-equal; when True, quantize_impl / split_quantize_nvfp4_impl
+  // dispatch to per-token K1+K2 kernels instead.
```
ptrendx, line 1877: What are "K1+K2" kernels?

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -1936,9 +1971,12 @@ std::pair<TensorWrapper, py::object> NVFP4Quantizer::create_tensor(
...
     rowwise_scale_inv_tensor = at::empty(scale_inv_shape_int64, bit8_tensor_opts);
-    const int64_t amax_rows = row_scaled_nvfp4 ? static_cast<int64_t>(flat_first_dim) : 1;
-    // hadamard amax kernel will zero out pointer with ZeroAmaxKernel
-    // nvte_compute_amax_with_config will zero out the pointer if needed
+    // Per-token: amax_rowwise is a per-row vector of length M (matches the
+    // (M,) row_amax expected by nvfp4_per_token_quantize K1).
+    // Row-scaled: same vector layout, kept for back-compat.
+    // Otherwise: scalar (length 1) consumed by cuBLASLt amax slot.
+    const int64_t amax_rows =
+        (per_token || row_scaled_nvfp4) ? static_cast<int64_t>(flat_first_dim) : 1;
```
ptrendx, line 1979: Considering this - what is then the difference between this per_token functionality and the existing
row_scaled_nvfp4 functionality? What I suspect is that this new one is an extension of the other one
 - shouldn't those 2 be combined then?

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -2478,6 +2527,123 @@ void NVFP4Quantizer::quantize_impl(const TensorWrapper& input, TensorWrapper& ou
...
+          std::nullopt, at::cuda::detail::getDefaultCUDAGenerator());
+      auto opts = at::TensorOptions().dtype(torch::kInt64).device(torch::kCUDA);
+      at::PhiloxCudaState philox_args = init_philox_state(gen, rng_elts_per_thread);
+      rng_state_at = torch::empty({2}, opts);
+      philox_unpack(philox_args, static_cast<int64_t*>(rng_state_at.data_ptr()));
+      te_rng_state = makeTransformerEngineTensor(rng_state_at);
+      rng_state_nvte = te_rng_state.data();
+    }
+    NVTE_SCOPED_GIL_RELEASE({
+      nvte_nvfp4_per_token_quantize(input.data(), nullptr, out.data(), this->with_rht ? 1 : 0,
```
ptrendx, line 2639: Why do we need to have a special function here? This sounds like it should be either an option in
the quantization params or it should correspond to its own NVTEScalingMode (and still be routed via
nvte_quantize_v2.

transformer_engine/pytorch/csrc/extensions/activation.cpp
```diff
@@ -43,8 +43,11 @@ py::object activation_helper(const at::Tensor& input, py::handle quantizer, int
     auto nvfp4_quantizer_cpp = dynamic_cast<NVFP4Quantizer*>(quantizer_cpp.get());
     NVTE_CHECK(nvfp4_quantizer_cpp != nullptr, "Could not cast to NVFP4 quantizer");
     if (nvfp4_quantizer_cpp->row_scaled_nvfp4 ||
-        (nvfp4_quantizer_cpp->with_rht && nvfp4_quantizer_cpp->with_post_rht_amax)) {
-      // Amax is handled within NVFP4 quantizer
+        (nvfp4_quantizer_cpp->with_rht && nvfp4_quantizer_cpp->with_post_rht_amax) ||
+        nvfp4_quantizer_cpp->per_token) {
+      // Per-token's amax is a per-row/per-col vector from its K1 amax kernel;
+      // the fused activation+amax path only emits a scalar, so per-token must
+      // use UNFUSED (activation in high precision, then K1 amax + K2 cast).
```
ptrendx, line 50: With this I got the information on what K1 and K2 are, but I believe that terminology is not useful
here and should be omitted. In fact, this whole comment does not really say anything useful - it
just describes what the unfused implementation option will do.

transformer_engine/pytorch/csrc/extensions/activation.cpp
```diff
@@ -156,8 +159,11 @@ py::object dactivation_helper(const at::Tensor& grad_output, const at::Tensor& i
...
```
ptrendx, line 166: Same.

transformer_engine/pytorch/csrc/extensions/bias.cpp
```diff
@@ -153,8 +153,11 @@ std::vector<py::object> dact_dbias(
     auto nvfp4_quantizer_cpp = dynamic_cast<NVFP4Quantizer *>(quantizer_cpp.get());
     NVTE_CHECK(nvfp4_quantizer_cpp != nullptr, "Could not cast to NVFP4 quantizer");
     if (nvfp4_quantizer_cpp->row_scaled_nvfp4 ||
-        (nvfp4_quantizer_cpp->with_rht && nvfp4_quantizer_cpp->with_post_rht_amax)) {
-      // Amax is handled within NVFP4 quantizer
+        (nvfp4_quantizer_cpp->with_rht && nvfp4_quantizer_cpp->with_post_rht_amax) ||
+        nvfp4_quantizer_cpp->per_token) {
+      // Per-token's amax is a per-row/per-col vector from its K1 amax kernel;
+      // the fused dact+amax path only emits a scalar, so per-token must use
+      // UNFUSED (dact in high precision, dbias reduction, then K1 amax + K2 cast).
```
ptrendx, line 160: Same comment as in activation.cpp.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -407,6 +407,11 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
   m.def("compute_amax", &transformer_engine::pytorch::compute_amax,
         "Compute absolute max value in tensor", py::arg("input"), py::arg("amax"),
         py::call_guard<py::gil_scoped_release>());
+  m.def("hadamard_transform_amax", &transformer_engine::pytorch::hadamard_transform_amax,
```
ptrendx, line 410: Don't we already have something very similar in the regular nvFP4 path?

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -455,6 +460,113 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
         py::arg("output_rowwise"), py::arg("output_colwise"), py::arg("scale_inv_rowwise"),
         py::arg("scale_inv_colwise"), py::arg("rows"), py::arg("cols"), py::arg("start_offset"),
         py::call_guard<py::gil_scoped_release>());
+  m.def("nvfp4_per_token_quantize", &transformer_engine::pytorch::nvfp4_per_token_quantize,
```
ptrendx, line 463: We should just use regular tex.quantize rather than introduce a new function here.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -455,6 +460,113 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
...
+        "with_swizzle=True writes rowwise scale_inv directly in the swizzled layout.",
+        py::arg("input"), py::arg("q_row"), py::arg("s_dec_row"), py::arg("row_amax"),
+        py::arg("q_col"), py::arg("s_dec_col"), py::arg("col_amax"), py::arg("rowwise"),
+        py::arg("columnwise"), py::arg("with_rht") = false,
+        py::arg("random_sign_mask_t") = static_cast<int64_t>(0xACE1),
+        py::arg("with_swizzle") = false, py::arg("with_sr") = false,
+        py::arg("rng_state") = std::nullopt);
+  m.def("nvfp4_cutlass_gemm", &transformer_engine::pytorch::nvfp4_cutlass_gemm,
+        "Stage-1 forked CUTLASS NVFP4 x NVFP4 -> BF16 GEMM with scalar (alpha, "
+        "beta) epilogue. Drop-in replacement for cuBLAS LT NVFP4. Pair with "
```
ptrendx, line 491: This description is not good - what does "Stage-1 forked" mean? Also, this whole function should be
folded to the gemm function instead.

ptrendx (commented):
Additional comments from Codex review:
```
  - [high] /tmp/te-pr3045/transformer_engine/common/gemm/nvfp4_cutlass_grouped_gemm.cu:232 uses process-global scratch, workspace, and host staging buffers with no device, stream, or thread isolation. Two grouped
    GEMMs on different streams can overwrite metadata/workspace while the first kernel is still consuming it. A call on another CUDA device can reuse a pointer allocated on the first device; the SM count and
    device_id are also hard-coded to device 0. This can cause corrupted results, use-after-free during growth, or invalid-device-pointer failures.

  - [high] /tmp/te-pr3045/transformer_engine/pytorch/csrc/quantizer.cpp:1952 rejects every per-token tensor whose flattened row count is not divisible by 128. GroupedLinear passes dynamic expert token counts
    directly into split_quantize, so ordinary unbalanced MoE routing commonly fails before reaching the grouped kernel. The tests only use aligned splits, and the example does not document or explicitly enable
    expert padding. The direct grouped wrapper also rejects zero-token experts.

  - [medium] /tmp/te-pr3045/transformer_engine/pytorch/cpp_extensions/gemm.py:596 validates alpha and beta, but the per-token branch at line 684 does not forward either value. Therefore alpha != 1 is silently
    ignored, and accumulate=True, beta != 1 always behaves as beta=1. Current module calls use the defaults, but this violates general_gemm’s public contract and produces silently wrong output for other callers.

  - [medium] /tmp/te-pr3045/transformer_engine/pytorch/csrc/quantizer.cpp:2576 restores out.amax using rowwise_amax_ptr before the weight-2D cast. A valid columnwise-only NVFP4Quantizer(per_token_weight_2d=True) has
    a null rowwise pointer, so nvte_quantize_v2 receives a null scalar-amax descriptor. Recipe-created weight quantizers currently request both directions, but the newly exposed constructor permits this crashing
    configuration.

  - [low] /tmp/te-pr3045/transformer_engine/common/recipe/__init__.py:766 and /tmp/te-pr3045/docs/features/low_precision_training/nvfp4/nvfp4.rst:287 still say that the cuDNN norm workaround and fused wgrad
    accumulation restrictions apply. This PR now automatically selects unfused normalization and implements FP32 accumulation, so the new API documentation contradicts the implementation.```

File: transformer_engine/common/include/transformer_engine/nvfp4_cutlass_gemm.h [hunk 33]
```diff
@@ -0,0 +1,79 @@
...
+#include "transformer_engine.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif  // __cplusplus
+
+/*! \brief D = alpha * (A @ B^T) + beta * C. A row-major (M,K), B col-major
+ *  (K,N), D/C row-major (M,N). A/B FP4-e2m1 packed; SFs FP8-e4m3 in CUTLASS
+ *  Sm1xxBlkScaledConfig layout; D BF16. M, N, K must be multiples of 256. */
+void nvte_nvfp4_cutlass_gemm(const NVTETensor a_data, const NVTETensor b_data,
```
timmoon10, line 25: My instinct is that this should be included within `nvte_cublas_gemm_v2` instead of creating a new API, but I have mixed opinions:
- Arguments for `nvte_cublas_gemm_v2`
  - cuBLAS and CUTLASS are computing the same thing, so it should be considered an internal implementation detail. Exposing it to the user is just cluttering the API.
  - Creating separate APIs for each CUTLASS kernel is not scalable if we add more CUTLASS kernels for other data formats.
- Arguments for `nvte_nvfp4_cutlass_gemm`
  - It is misleading for a function called `nvte_cublas_*` to not call cuBLAS. Perhaps it would be better to create a new API called something like `nvte_gemm` or `nvte_matmul`.
  - This PR doesn't expose a good way to distinguish between per-tensor and per-token NVFP4, so the only way to avoid incorrect behavior is to make extremely explicit API calls. However, this is more an argument for including this information properly within `NVTETensor`.

transformer_engine/common/include/transformer_engine/nvfp4_per_token.h
```diff
@@ -0,0 +1,193 @@
...
+ *  \param[in] with_swizzle  non-zero -> K2 emits rowwise scale_inv directly
+ *                       in the cuBLAS LT swizzled tile layout (rowwise only;
+ *                       colwise stays compact M-major).
+ *  \param[in] with_sr   non-zero -> the K2 FP4 cast uses stochastic rounding
+ *                       (per-element Philox dither). K1 amax stays
+ *                       deterministic. Zero is byte-equal to the RN path.
+ *  \param[in] rng_state int64 device tensor of shape [2] = {seed, offset}
+ *                       (host-unpacked Philox). May be NULL iff with_sr == 0.
+ */
+void nvte_nvfp4_per_token_quantize(const NVTETensor input, const NVTETensor noop, NVTETensor output,
```
timmoon10, line 37: We should use `nvte_quantize_v2` instead of creating a separate API that's specific to per-token NVFP4.

The root design flaw is that this PR uses `NVTETensor` in a non-standard way, with no way to distinguish between per-tensor and per-token NVFP4. This PR expects the user to call different APIs for per-token NVFP4, and if they call the general APIs then they'll have silent incorrect behavior. It would be better to expand our existing APIs instead of creating something custom from scratch.

Please take a look at https://github.com/NVIDIA/TransformerEngine/pull/2931 and how it incorporates row-scaled NVFP4 options. In particular, notice how we put quantizaiton options in two places:
- Options that are needed for both quantization and GEMM belong within `NVTETensor` (see [list of options](https://github.com/NVIDIA/TransformerEngine/blob/e201f228e69d9ef0e8780be32fc69e6b55cec5e7/transformer_engine/common/include/transformer_engine/transformer_engine.h#L66))
  - Examples: tensor-scale/row-scale, whether scales are swizzled
  - Without these options, you are unable to interpret the data bits. They are part of the tensor, just like the tensor dims, dtype, etc.
- Options that are only needed for quantization belong in `NVTEQuantizationConfig` (see [list of options](https://github.com/NVIDIA/TransformerEngine/blob/e201f228e69d9ef0e8780be32fc69e6b55cec5e7/transformer_engine/common/include/transformer_engine/transformer_engine.h#L400))
  - Examples: stochastic rounding, noop tensor
  - After quantization is finished, no one cares how the bits were generated. Stochastic rounding, no stochastic rounding, neither has an effect on the GEMM.

docs/features/low_precision_training/nvfp4/nvfp4.rst
timmoon10, line 216: How is this different from the row-scaled NVFP4 recipe from https://github.com/NVIDIA/TransformerEngine/pull/2931? I'd prefer to reuse and expand the existing APIs instead of creating redundant ones from scratch.

timmoon10 (changes requested):

docs/features/low_precision_training/nvfp4/nvfp4.rst
zhongbozhu: I agree, given the complexity of features like `--fp4-param-gather`, introducing new recipe class & quantized tensors would make it impossible to be integrated into Mcore in a performant way, till the point the users can just use mxfp8.

docs/features/low_precision_training/nvfp4/nvfp4.rst
zhongbozhu: Agreed

zianglih:
row scaled grouped gemm is already supported in cudnn fe:
- https://github.com/NVIDIA/cudnn-frontend/pull/251

transformer_engine/pytorch/tensor/storage/nvfp4_tensor_storage.py
```diff
@@ -124,6 +129,7 @@ def __new__(
         row_scaled_nvfp4: bool = False,
         nvfp4_use_4over6: bool = False,
         nvfp4_e4m3_max: int = 448,
+        per_token: bool = False,
```
zianglih, line 132: `row_scaled_nvfp4` above is the exact same thing

cael-ling: Thanks, we are exploring a way to integrate the logic from this PR into existing strategies, rather than merging the PR in its entirety.

PR #3047, opened 2026-05-27
[PyTorch] Debug CPU offloading in grouped linear and grouped MLP

lhb8125:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
+++ b/transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
@@ -619,7 +619,7 @@ class _ForwardGroupedMLP_CuTeGEMMBase(FusedOperation):
                     else:
                         fc2_out_buf = fc2_out_buf + token_bias
             else:
-                fc2_out_grouped = GroupedTensor(
+                fc2_out_grouped = GroupedTensorStorage(
                     shape=(in_shape[0], fc2_weight_shape[0]),
                     dtype=dtype,
                     num_tensors=num_groups,
--- a/transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
+++ b/transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
@@ -13,6 +13,7 @@ from typing import Any, Optional
 import torch
 
 import transformer_engine_torch as tex
+from ...cpu_offload import is_cpu_offload_enabled, mark_not_offload
 from ...cpp_extensions import general_gemm, general_grouped_gemm_for_grouped_tensor
 from ...quantization import Recipe
 from ...tensor import NVFP4Quantizer, NVFP4Tensor, Quantizer
@@ -23,6 +24,7 @@ from ...utils import (
     mark_grouped_tensor,
 )
 from ...tensor.grouped_tensor import GroupedTensor
+from ...tensor.storage.grouped_tensor_storage import GroupedTensorStorage
 from ...tensor.mxfp8_tensor import MXFP8Quantizer
 from ...constants import MXFP8_BLOCK_SCALING_SIZE, NVFP4_BLOCK_SCALING_SIZE
 from ..basic import GroupedLinear, ScaledSReLU, ScaledClampedQGeGLU
@@ -316,6 +318,7 @@ class _ForwardGroupedMLP_CuTeGEMMBase(FusedOperation):
         # Group-quantize input tensor and convert dtypes if needed
         fc1_input_quantizer.set_usage(rowwise=True, columnwise=weight_requires_grad)
         fc1_input_quantizer.optimize_for_gemm = True
+        fc1_input_quantizer.internal = True
         input_quantizer = getattr(input_, "quantizer", None)
         if isinstance(input_, GroupedTensor) and (
             isinstance(fc1_input_quantizer, MXFP8Quantizer)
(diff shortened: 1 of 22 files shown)


transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -500,6 +524,11 @@ def fuser_forward(
         if requires_grad:
             mark_grouped_tensor(grouped_fc1_x, activation_in, scales, grouped_fc2_x)
             activation_op = self.basic_ops[1]
+            cpu_offloading = is_cpu_offload_enabled()
+            offload_fc1_input = bool(getattr(fc1_op, "fine_grained_activation_offloading", False))
+            offload_activation_input = bool(
+                getattr(activation_op, "fine_grained_activation_offloading", False)
+            )
```
timmoon10: - Do these options give us value? The dense linear op and activation ops don't expose this fine-grained control:

https://github.com/NVIDIA/TransformerEngine/blob/ace2a9653a2da74a8576a39ef7d1181c7c7923cb/transformer_engine/pytorch/ops/basic/basic_linear.py#L1052-L1053

https://github.com/NVIDIA/TransformerEngine/blob/ace2a9653a2da74a8576a39ef7d1181c7c7923cb/transformer_engine/pytorch/ops/basic/activation.py#L116-L117

- For consistency with the rest of the CPU offloading behavior, shouldn't the default be to enable offloading? Disabling offloading should be the explicit path.
- These secret undocumented attrs are delicate and unmaintainable. Better to make them arguments in the unfused ops. However, this also means we should update the unfused impls so that they disable activation checkpointing if the option is set.

Easiest just to not make this configurable.

timmoon10 (commented):
Overall looks good, but with one design suggestion.

Followup tasks after merging this PR:
- Enable activation checkpointing in the unfused grouped linear op.
- Update activation checkpointing to support v2 infrastructure from https://github.com/NVIDIA/TransformerEngine/pull/1762, which is opt-out rather than opt-in.

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
lhb8125: Thanks Tim, I may not fully get you points, but let me try to clarify.

MCore's fine grained activation offloading captures the tensors by adding a hook to torch's save_for_backward(), and it doesn't have a knowledge that where the tensors is from, so the default strategy is offloading all captured tensors and filtering out the tensors which are marked as not offload.

As for the weight tensors, they are marked as not offload by TE's offloading strategy, so for the unfused group linear, we reuse the attr from TE's offloading. For fused group mlp, I followed the group linear's style to mark weight tensors as not offload.

TE's v1 offloading strategy marks tensors as not offload by default, while v2 offloading strategy marks tensors as offload by default, which is consistent with MCore's offloading strategy.

For the unfused impls, we didn't insert too much codes for MCore's offloading because we only need to filter out the weight tensors, which can reuse TE's strategy.

timmoon10: It makes sense that we need to mark the activations so that the v2 offloading strategy can access them. However, my complaint:
- `te.Linear` offloads all activations no matter what.
- Forward grouped MLP only offloads activations if you set a secret attr.

Why not just make grouped MLP have the same behavior as `te.Linear`? If I were a user, there's no way I could figure this out without first hitting an unexpected OOM and then digging into the code. We want our APIs to be simple, general, and obvious. Having hyperspecific hacky interfaces might be needed if we're scrambling against a deadline, but it's sloppy software development.

lhb8125: I see, so you proposal is letting offloading all activations of grouped MLP as the default strategy unless we specify "do not offload ***". I agree it's more consistent and will push a fix.

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -650,6 +676,9 @@ def fuser_forward(
         if requires_grad:
             mark_grouped_tensor(grouped_fc1_x, activation_in, scales, grouped_fc2_x)
             activation_op = self.basic_ops[1]
+            cpu_offloading = is_cpu_offload_enabled()
+            no_offload_fc1_activation = bool(getattr(fc1_op, "no_offload_activation", False))
+            no_offload_moe_activation = bool(getattr(activation_op, "no_offload_activation", False))
```
timmoon10: Fine, but we should avoid relying on these in Mcore. They should be considered as debugging tools and we don't make guarantees to maintain them.

If we do want to control this in Mcore, we should do it through proper, user-facing APIs. We can add it as an arg to the unfused ops, but we should also make sure the unfused ops actually respect it too.

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -298,14 +300,38 @@ def fuser_forward(
...
+        fc1_input_quantizer.internal = True
         input_quantizer = getattr(input_, "quantizer", None)
         if isinstance(input_, GroupedTensor) and (
             isinstance(fc1_input_quantizer, MXFP8Quantizer)
             and isinstance(input_quantizer, MXFP8Quantizer)
             or isinstance(fc1_input_quantizer, NVFP4Quantizer)
             and isinstance(input_quantizer, NVFP4Quantizer)
         ):
-            grouped_fc1_x = input_
+            grouped_fc1_x = GroupedTensorStorage(
```
timmoon10, line 335: Now that we set the input quantizer with `.internal = True`, isn't it redundant to repack `grouped_fc1_x` into a `GroupedTensorStorage`?

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
lhb8125: Yea, I also thought about the same thing, but didn't want to add too many changes. Now I add ```offload_activation``` for all related modules, now the consistence would be better.

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
lhb8125: I think they are targeting different cases. The input quantizer with ```.interal= True```` takes effects on bf16 input, where we need to quantize it by ```fc1_input_quantizer```. The second case is that the input is already a quantized fp8 tensor, where we need to repack it into a ```GroupedTensorStorage```.

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
timmoon10: I don't like that this is ballooning the scope of this PR and adds a maintenance burden to so many ops. It just makes it that much more painful to implement a new fusion. I've reverted the new arg so that this PR just fixes the offloading bug in grouped MLP.

For a more general solution to selective offloading, I wonder if we can expand the existing APIs for CPU offloading. If you want to explicitly disable in one layer, what if we allowed nesting like this:
```python
with get_cpu_offload_context(...):
    x = layer0(x)  # CPU offloading
    with get_cpu_offload_context(enabled=False, ...):
        x = layer1(x)  # No CPU offloading
    x = layer2(x)  # CPU offloading
```
This would work automatically, with no extra logic needed in every op.

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
timmoon10: ~How could it already be quantized? The only way to create it is from the quantizer, either just now or from a previous step (e.g. with activation recompute). If the quantizer has `.internal=True`, it can only be `GroupedTensorStorage`.~

~If something is incorrectly producing `GroupedTensor`, then that's a bug. Fixing it here is papering over the real problem.~

Actually, on second thought, it makes sense that `input_` can be a `GroupedTensor` since it comes from outside the op. It would be useful to know what use-case hit this bug though. Activation recompute?

Really the root cause is that CPU offloading doesn't handle `GroupedTensor` gracefully, but that would be a more involved effort.

timmoon10 (approved):
I propose merging this quickly to unblock CPU offloading with grouped MLP. Afterwards, we can work on a more general interface for selectively enabling/disabling in specific modules.

PR #3048, opened 2026-05-27
Enable NVFP4 fused grouped MLP 

sraman-rgb:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -1083,6 +1080,12 @@ def _fuser_forward_split_quantize(
 
         # Split input tensor and convert dtypes if needed
         x = maybe_dequantize(input_, dtype)
+        if num_groups == 1:
+            # Avoid CUDA->CPU sync from split_sizes.tolist() during CUDA graph capture.
+            split_sizes_int = [x.numel() // x.size(-1)]
+        else:
+            # Need CPU split sizes for split_quantize / general_grouped_gemm.
+            split_sizes_int = [int(s) for s in split_sizes.tolist()]
```
vthumbe1503: We dont need this change right?..for CUDA graph capture.. we want users to go to the grouped tensor path and not the split quantize path

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -1329,8 +1332,12 @@ def _fuser_backward_split_quantize(
         ws, saved_tensors = saved_tensors[:num_groups], saved_tensors[num_groups:]
 
         # Split grad output tensor and convert dtypes if needed
-        split_sizes_int = [int(s) for s in split_sizes.tolist()]
         dy = maybe_dequantize(grad_output, ctx.dtype)
+        if num_groups == 1:
+            # Avoid CUDA->CPU sync from split_sizes.tolist() during CUDA graph capture.
+            split_sizes_int = [dy.numel() // dy.size(-1)]
+        else:
+            split_sizes_int = [int(s) for s in split_sizes.tolist()]
```
vthumbe1503: Same comment as above

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -649,6 +945,8 @@ def fuse_forward_srelu_ops(
 ) -> list[FusibleOperation]:
     """Apply GroupedLinear + ScaledSReLU + GroupedLinear fusion for forward pass."""
 
+    if recipe is None or not recipe.mxfp8():
```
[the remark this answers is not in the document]
sraman-rgb: That is intended

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -35,6 +36,159 @@
...
+    if not tensors:
+        return
+    row_amaxes = [getattr(tensor, "_amax_rowwise", None) for tensor in tensors]
+    if all(amax is not None for amax in row_amaxes):
+        packed_row_amax = torch.cat([amax.view(-1) for amax in row_amaxes], dim=0).contiguous()
+        for idx, tensor in enumerate(tensors):
+            tensor._amax_rowwise = packed_row_amax[idx : idx + 1]
+    col_amaxes = [getattr(tensor, "_amax_columnwise", None) for tensor in tensors]
+    if all(amax is not None for amax in col_amaxes):
+        packed_col_amax = torch.cat([amax.view(-1) for amax in col_amaxes], dim=0).contiguous()
```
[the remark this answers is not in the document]
sraman-rgb: Updated in the latest commit

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -274,6 +284,8 @@ def fuser_forward(
                 else:
                     quantized_fc2_weights.append(weight)
             grouped_fc2_weight = quantized_fc2_weights
+            if isinstance(fc2_input_quantizer, NVFP4Quantizer):
+                _pack_nvfp4_amax_list(grouped_fc2_weight)
```
vthumbe1503: Is this needed given that we are already concatenating the amaxes below in _nvfp4_amax function?

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -407,94 +491,169 @@ def fuser_forward(
...
+            fc2_bias_scale = None
+            if fc2_bias_packed is not None:
+                fc2_bias_for_gemm = fc2_op._get_grouped_bias_for_gemm(dtype)
+                if fc2_scales is not None:
+                    fc2_bias_scale = fc2_scales.reshape(-1)
+                    if fc2_bias_scale.dtype != torch.float32:
+                        fc2_bias_scale = fc2_bias_scale.to(dtype=torch.float32)
+
+            fc2_in = fc1_kernel_out["d_tensor"]
+            fc2_in = fc2_in.view(in_shape[0], fc2_weight_shape[1]).contiguous()
```
vthumbe1503, line 497: Can we abstract out nvfp4 fc2 logic in a different function?
It is getting tougher to read with cute dsl kernels for fc2 being used in the same place.

vthumbe1503 (commented):
LGTM mostly. Have left minor comments. We need to eventually abstract out Dense logic into a different layer called DenseMLP (and that can eventually use fusion of Linear + Swiglu + Linear in te_ops). but that is beyond the scope of this PR.

transformer_engine/pytorch/ops/fused/backward_grouped_mlp.py
```diff
@@ -625,21 +777,39 @@ def fuser_backward(
...
+            fc1_grad_output_quantizer.optimize_for_gemm = True
+            _enable_nvfp4_rht_for_group_quantize(fc1_grad_output_quantizer)
+            grouped_fc1_dy = _group_quantize_for_grouped_mlp(
+                fc1_dy_bf16,
+                fc1_grad_output_quantizer,
+                num_groups,
+                split_sizes,
+                tensor_offsets=fc1_dy_tensor_offsets,
+            )
+            _mark_with_gemm_swizzled_scales(grouped_fc1_dy)
```
vthumbe1503: Why do we have to manually mark it as swizzled? group_quantize should already do it no?

transformer_engine/pytorch/ops/fused/backward_grouped_mlp.py
```diff
@@ -674,66 +844,107 @@ def fuser_backward(
...
-                    MXFP8_BLOCK_SCALING_SIZE,
-                    4,
-                    4,
-                )
-                fc1_w_scales = fc1_w_scales.permute(3, 4, 1, 5, 2, 0)
-
-                fc1_dgrad_kwargs["b_tensor"] = fc1_w_data
-                fc1_dgrad_kwargs["sfb_tensor"] = fc1_w_scales
+            if use_nvfp4:
+                _mark_with_gemm_swizzled_scales(grouped_fc1_weight)
```
vthumbe1503: Where are we swizzling them before marking them as swizzled?

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -407,94 +491,157 @@ def fuser_forward(
...
+                general_grouped_gemm_for_grouped_tensor(
+                    grouped_fc2_weight,
+                    grouped_fc2_x,
+                    fc2_out_grouped,
+                    layout="TN",
+                )
+            fc2_out = fc2_out_buf
+            if fc2_bias_packed is not None:
+                bias_per_group = fc2_bias_packed.transpose(0, 1).contiguous()
+                token_bias = torch.repeat_interleave(bias_per_group, split_sizes, dim=0)
```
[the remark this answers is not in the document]
sraman-rgb: Updated in the latest commit

transformer_engine/pytorch/ops/fused/backward_grouped_mlp.py
sraman-rgb: the actual swizzles are done at various places: For normal single-tensor NVFP4 quantization, actual swizzle is here: /transformer_engine/pytorch/ops/_common.py:94 -> tex.swizzle_scales_for_gemm_(quantized); For grouped NVFP4 RHT/quantize: common/hadamard_transform/graph_safe_group_row_cast_col_hadamard_transform_cast_fusion.cu:1420; const bool use_swizzle_sf_output = output->with_gemm_swizzled_scales; This function is just setting up the meta data;

vthumbe1503: metadata should be set when the swizzling actually happens..we shouldnt have to do it manually ourselves afterwards unless necessary. swizzle_scales_for_gemm already sets the metadata correctly..
is graph_safe_group_row_cast_col_hadamard_transform_cast_fusion not setting metadata correctly?, then we should fix that

transformer_engine/pytorch/ops/basic/grouped_linear.py
sraman-rgb: Updated in the latest commit

transformer_engine/pytorch/ops/basic/grouped_linear.py
sraman-rgb: Updated in the latest commit

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
sraman-rgb: Updated in the latest commit

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
sraman-rgb: Updated in the latest commit

transformer_engine/pytorch/ops/_common.py
```diff
@@ -57,6 +60,152 @@ def _nvidia_cudnn_frontend_supports_wgrad() -> bool:
...
+    split_sizes: Optional[torch.Tensor],
+    *,
+    tensor_offsets: Optional[torch.Tensor] = None,
+) -> GroupedTensor:
+    """Quantize into grouped storage, using regular quantize for one-group NVFP4."""
+    if num_groups != 1 or not isinstance(quantizer, NVFP4Quantizer):
+        return tex.group_quantize(tensor, quantizer, num_groups, split_sizes)
+
+    quantized = tex.quantize(tensor, quantizer)
+    with_gemm_swizzled_scales = getattr(quantized, "_with_gemm_swizzled_scales", False)
```
timmoon10: This logic is NVFP4-specific and we've already checked that we're quantizing with an `NVFP4Quantizer`, so we can just access the attrs directly. I worry that these `getattr` s will add extra overhead and hide errors with invalid tensor types. I see this pattern throughout this file.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -57,6 +60,152 @@ def _nvidia_cudnn_frontend_supports_wgrad() -> bool:
...
+    num_groups: int,
+    split_sizes: Optional[torch.Tensor],
+    *,
+    tensor_offsets: Optional[torch.Tensor] = None,
+) -> GroupedTensor:
+    """Quantize into grouped storage, using regular quantize for one-group NVFP4."""
+    if num_groups != 1 or not isinstance(quantizer, NVFP4Quantizer):
+        return tex.group_quantize(tensor, quantizer, num_groups, split_sizes)
+
+    quantized = tex.quantize(tensor, quantizer)
```
timmoon10, line 82: This single-group path is a hacky expedient to expose the fused grouped GEMM + activation kernel in the dense MLP (also see https://github.com/NVIDIA/Megatron-LM/pull/4786). However, we should not commit to this long-term. The right approach is to create a new fused op specifically for the dense MLP. Better to have some code duplication contained within the fused op than to make Mcore contort itself in weird way.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -57,6 +60,152 @@ def _nvidia_cudnn_frontend_supports_wgrad() -> bool:
...
+        columnwise_amax=columnwise_amax,
+        first_dims=split_sizes,
+        tensor_offsets=tensor_offsets,
+        with_gemm_swizzled_scales=with_gemm_swizzled_scales,
+    )
+
+
+def _nvfp4_logical_data_view(data: torch.Tensor) -> torch.Tensor:
+    """View packed NVFP4 data with its logical K dimension for scale swizzling."""
+    return data.as_strided((data.shape[0], data.shape[1] * 2), (data.stride(0), 0))
```
timmoon10: I can see why we might want to compute the logical tensor shape, but I don't see the value of creating this tensor view. The strides are wrong (since each entry takes half of a byte), so tensor accesses will be wrong.

If this is a hack to work with `tex.get_device_pointer_for_data_and_scales`, be advised that it is disappearing in https://github.com/NVIDIA/TransformerEngine/pull/3001.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -285,7 +434,7 @@ def fuse_grouped_mlp_ops(
 
     if not fused_op_cls.is_supported():
         return ops
-    if recipe is None or not recipe.mxfp8():
+    if recipe is None or not (recipe.mxfp8() or recipe.nvfp4()):
```
timmoon10, line 432: We should modify the test to make sure this fusion is applied with NVFP4: https://github.com/NVIDIA/TransformerEngine/blob/71427e7c10c88455f33d10bffb4abed6655f725b/tests/pytorch/test_fusible_ops.py#L3978

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
timmoon10, line 1: Long-term, it would be better to have separate fused ops for MXFP8 and NVFP4. There will be some code duplication, but the logic will also be simpler and it will be easier to have recipe-specific behavior. Right now the cuDNN kernels have matching APIs, but we expect they'll start diverging as we add more things like RHT. Better to let the kernels evolve according to what gives the best performance, and not to straitjacket them to having the same pattern.

If this is the best design for now, we should at least rename the fused op to something like `_ForwardGroupedMLP_CuTeGEMMBase`, and similar for the backward op.

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -288,13 +295,34 @@ def fuser_forward(
         # Group-quantize input tensor and convert dtypes if needed
         fc1_input_quantizer.set_usage(rowwise=True, columnwise=weight_requires_grad)
         fc1_input_quantizer.optimize_for_gemm = True
-        if isinstance(input_, GroupedTensor) and isinstance(
-            getattr(input_, "quantizer", None), MXFP8Quantizer
+        _enable_nvfp4_rht_for_group_quantize(fc1_input_quantizer)
```
timmoon10: The quantizers should already have RHT configured. If the NVFP4 recipe has RHT disabled, we need to respect that. This pattern is repeated in a few other places.

transformer_engine/pytorch/ops/_common.py
sraman-rgb: Yeah I agree. That should be the long term strategy

tests/pytorch/test_fusible_ops.py
```diff
@@ -3884,71 +3891,88 @@ def test_grouped_mlp(
...
+                    delay_wgrad_compute=delay_wgrad_compute,
+                    scale_bias=bias,
+                )
+                return te_ops.Sequential(fc1_op, _make_scaled_act(), fc2_op), fc1_op, fc2_op
+
+        module, fc1, fc2 = _make_module()
+        if use_nvfp4_rht_recipe:
+            module_ref, fc1_ref, fc2_ref = _make_module()
+        else:
+            module_ref, fc1_ref, fc2_ref = None, None, None
```
timmoon10: Why do we need a separate reference impl for the RHT case? It shouldn't change the underlying math, only the approximation scheme for that math. In fact, the point of RHT is to reduce NVFP4 quantization error, so in principle it could support tighter tols than non-RHT.

I'm also uneasy about using TE as the reference since that exposes us to situations where the reference and test impls both have the same bug.

tests/pytorch/test_fusible_ops.py
```diff
@@ -3863,66 +3869,135 @@ def test_grouped_mlp(
...
+        if quantization == "nvfp4_rht":
+            for group_idx in range(group_size):
+                fc1_dy = fc1_outputs[group_idx].grad
+                if fc1_dy is None:
+                    fc1_dy = torch.zeros_like(fc1_outputs[group_idx])
+                fc1_ws_ref[group_idx].grad = _nvfp4_rht_wgrad(fc1_inputs[group_idx], fc1_dy)
+                fc2_ws_ref[group_idx].grad = _nvfp4_rht_wgrad(
+                    fc2_inputs[group_idx],
+                    dys[group_idx],
+                )
```
timmoon10: The reference implementation shouldn't be affected by RHT (https://github.com/NVIDIA/TransformerEngine/pull/3048#discussion_r3326877933):
```suggestion
```

tests/pytorch/test_fusible_ops.py
sraman-rgb: Removed in the latest commit

tests/pytorch/test_fusible_ops.py
sraman-rgb: Removed in the latest commit

timmoon10 (approved):
LGTM, pending CI

PR #3053, opened 2026-05-28
[PyTorch] [torch.compile] torch.compile support for Linear

pggPL:
# Description

This PR adds `torch.compile` support for `te.pytorch.Linear`, building on the `TensorSpec` mechanism already in `main`.

`_Linear`'s forward and backward are registered as `torch.library` custom ops, so a module containing `te.Linear` traces under `torch.compile(fullgraph=True)` without graph breaks. The fake (meta) implementations describe the produced tensors through `TensorSpec` instead of allocating them, which is what makes the quantized outputs traceable — the compiler sees the full quantized-tensor structure (data, scales, transposes) without any device allocation at trace time.

The bulk of the diff is `transformer_engine/pytorch/dynamo/custom_op.py`: a declarative `register_custom_op` helper. Custom ops require flat lists of tensors, while the TE forward/backward take dataclass "argument bundles" holding tensors, quantized tensors, quantizers, process groups and plain Python values. The helper derives the op schema from the dataclass field annotations, flattens each field to op slots via a per-kind adapter, and rebuilds the bundle on the other side, so ops are declared by writing a dataclass rather than by hand-maintaining a schema string.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- `dynamo/custom_op.py` (new): `register_custom_op` — declarative registration of forward/backward custom ops from dataclass argument bundles. Handles per-field adapters for plain tensors, quantized tensors, quantizers, opaque value bundles and reference-opaque types (e.g. process groups), schema generation, `TensorSpec`-based fake outputs and autograd wiring. Falls back to eager with a single warning if registration fails.
- `dynamo/__init__.py`: export `register_custom_op`.
- `module/linear.py`: split the forward into pure computation and context saving, add allocation-free fake forward/backward on `TensorSpec`, and register `_Linear` through `register_custom_op`. Eager behavior is unchanged.
- `dynamo/quantizer_opaque.py`, `dynamo/tensor_spec.py`, `tensor/_quantization_helpers.py`, `tensor/float8_tensor.py`, `tensor/storage/float8_tensor_storage.py`, `utils.py`: small supporting changes (idempotent spec conversion, weight-workspace quantizer preservation, keeping attributes attached to quantized parameters across `_apply`).
- `tests/pytorch/test_torch_compile.py`: coverage for the compiled Linear — fullgraph compilation, quantized FP8 weights, FP8 output, `is_first_microbatch`, dynamic shapes, parametrized over the supported recipes (FP8 per-tensor/current scaling, MXFP8, NVFP4).
- `tests/pytorch/distributed/*`: exercise the compiled path in the distributed numerics and comm-GEMM-overlap runs.

# Checklist:


tests/pytorch/distributed/run_layer_with_overlap.py
```diff
@@ -285,6 +298,9 @@ def _parse_args(argv=None, namespace=None):
     )
     args = parser.parse_args(argv, namespace)
 
+    if args.compile and args.use_cuda_graphs:
+        parser.error("--compile and --use-cuda-graphs are mutually exclusive.")
```
ptrendx: So how do we test the lower overhead mode? Shouldn't we just do that mode of the compilation if this
option is set?

tests/pytorch/distributed/run_numerics.py
```diff
@@ -310,22 +310,45 @@ def _copy_params(model_distributed, model_single):
...
     _alloc_main_grad(model_single_node, model_distributed)  # for fuse_wgrad_accumulation=True
     input_single_node.requires_grad_()
     input_distributed.requires_grad_()
+    forward_single_node = model_single_node
+    forward_distributed = model_distributed
+    if use_compile:
+        # Each parametrized case compiles the same module.forward code object with
+        # a different shape/recipe; with dynamic=False those guards accumulate and
+        # eventually trip Dynamo's recompile_limit. Reset so every case starts from
+        # a clean compile cache (mirrors the single-GPU torch.compile tests).
```
ptrendx: One concern here is that when more things are going to try testing the compilation flow we would
spend a lot of time in the compilation during the CI. Maybe it would be good to have some thoughts
on which tests should actually be run in L0 and which could be done in the nightly CI for torch.compile.

tests/pytorch/distributed/run_numerics.py
```diff
@@ -310,22 +310,45 @@ def _copy_params(model_distributed, model_single):
...
+        torch._dynamo.reset()
+        # dynamic=False for now: a symbolic shape would land in an OpaqueValueBundle
+        # (value-opaque op arg) whose hash chokes on non-nested SymInt. Force static
+        # shapes (recompile per shape) until the bundle handles symbolic shapes.
```
ptrendx: Is the dynamic=False a limitation that we require users to set? If so, we need to have some documentation
about that somewhere.

tests/pytorch/distributed/test_comm_gemm_overlap.py
```diff
@@ -281,6 +287,40 @@ def test_layers_with_overlap_bf16(
     )
 
 
+@pytest.mark.parametrize("compile_mode", ["default", "reduce-overhead"])
```
ptrendx, line 293: Kind of a general comment, but do we expect to ever see a case that would work under reduce
overhead mode but not work under the default mode? If so then maybe we could just test the stricter
mode if things are supposed to work under both of them?

tests/pytorch/test_torch_compile.py
```diff
@@ -93,6 +94,73 @@ def nvfp4_4over6():
...
+
+def _cudagraph_warmup(fn, inp, *, backward: bool) -> None:
+    """Force TE's lazily-created global scratch to be allocated before capture."""
+    out = fn(inp)
+    if backward:
+        out.sum().backward()
+
+
+@contextlib.contextmanager
+def _assert_no_cudagraph_skips(enabled: bool):
```
ptrendx, line 142: Generally speaking I think it makes sense, but since the counters map is technically an internal
thing in PyTorch, I think that we should guard ourselves a little bit in case the API changes.
E.g. our CI should not fail (but probably issue some kind of warning?) if the counters map is not
there or the specific keys are not found.

tests/pytorch/test_torch_compile.py
```diff
@@ -93,6 +94,73 @@ def nvfp4_4over6():
...
+    if enabled:
+        skipped = counters["inductor"]["cudagraph_skips"] - before
+        assert skipped == 0, (
+            f"reduce-overhead fell back to eager: {skipped} cudagraph skip(s); "
+            "see the 'skipping cudagraphs due to ...' log for the reason"
+        )
+
+
+# bf16 output tolerance: eager and compiled run the same kernels, so they should
+# agree closely; the slack only absorbs reduction-order / cuda-graph differences.
```
ptrendx: CUDA graphs and non-CUDA graph execution should actually be exactly identical. The compiled vs eager
could have some difference if you capture more that just TE, but should not have any difference in
the TE ops themselves.

tests/pytorch/test_torch_compile.py
```diff
@@ -715,17 +783,32 @@ def _hw_available(quantizer):
 
 # (factory, kwargs producing a different-but-valid config)
 _VALUE_QUANTIZERS = [
-    pytest.param(_mxfp8, id="mxfp8"),
-    pytest.param(_blockwise, id="float8_blockwise"),
-    pytest.param(_current_scaling, id="float8_current_scaling"),
-    pytest.param(_nvfp4, id="nvfp4"),
+    pytest.param(_mxfp8, {"dtype": tex.DType.kFloat8E5M2}, id="mxfp8"),
```
ptrendx: Why?

tests/pytorch/test_torch_compile.py
```diff
@@ -715,17 +783,32 @@ def _hw_available(quantizer):
...
     """Value semantics + ``__fx_repr__`` round-trip via the production FX path."""
-    a = factory()
+    a, b = factory(), factory()
+    # Same config -> equal, same hash, interchangeable as a dict/set key.
+    assert a is not b
+    assert a == b
+    assert hash(a) == hash(b)
+    assert {a: "x"}[b] == "x"
+    # Different config -> not equal.
+    assert a != factory(**other_kwargs)
```
ptrendx: I think this is an artifact of the rebase or something, I vaguely recall we got rid of something
very similar from the previous PR.

tests/pytorch/test_torch_compile.py
```diff
@@ -1164,3 +1250,278 @@ def test_to_tensor_spec_quantized(factory, shape):
...
+    ids=lambda r: "bf16" if r is None else type(r).__name__,
+)
+def test_te_linear_compiles(fp8_recipe, compile_mode):
+    """
+    torch.compile(fullgraph=True) of ``te.Linear`` under every built-in
+    recipe (plus the bf16-only baseline with no autocast), for both the default
+    backend and ``mode="reduce-overhead"`` (CUDA-graph trees).
+    """
+    if fp8_recipe is not None and not fp8_available:
+        pytest.skip(reason_for_no_fp8)
```
ptrendx: The `fp8_recipe` takes values from `all_recipes` which already skips unavailable recipes, so this
check is useless.

tests/pytorch/test_torch_compile.py
```diff
@@ -1164,3 +1236,240 @@ def test_to_tensor_spec_quantized(factory, shape):
...
+            backward=True,
+        )
+        model.zero_grad(set_to_none=True)
+    compiled = torch.compile(fn, fullgraph=True, mode=compile_mode)
+
+    with _assert_no_cudagraph_skips(compile_mode == "reduce-overhead"):
+        for is_first in schedule:
+            base = torch.randn(32, 64, dtype=dtype, device=device)
+            _assert_close_eager_compiled(fn, compiled, model, base)
+
```
ptrendx, line 2220: A similar test that should be here is to run the Linear in training and eval modes back to back to
see whether changing from training -> inference (and more importantly from inference -> training)
works well.

tests/pytorch/distributed/run_numerics.py
pggPL: It is supported now.

tests/pytorch/distributed/run_numerics.py
pggPL: All distributed are in L1, and torch compile tests for L0 are quite fast, so this is concern for future.

tests/pytorch/distributed/run_layer_with_overlap.py
pggPL: The lower-overhead mode is covered by `--compile --compile-mode=reduce-overhead`, `--use-cuda-graphs` represents completely different manual logic.

tests/pytorch/distributed/test_comm_gemm_overlap.py
pggPL: The benefit of such approach would be time saved. If torch.compile + TE CI time will be big we may do that.

tests/pytorch/distributed/run_layer_with_overlap.py
ptrendx: Then maybe we could add to this error message that if somebody wants to test torch.compile with
graphs then they could do that by setting reduce overhead option.

tests/pytorch/distributed/run_numerics.py
```diff
@@ -740,12 +777,20 @@ def test_linear():
...
+        {"use_compile": True},
+        {"use_compile": True, "compile_mode": "reduce-overhead"},
     ]
 
     for kwargs in kwargs_list:
         if kwargs.get("save_original_input", False) and QUANTIZATION == "fp8":
             continue
-        if kwargs.get("delay_wgrad_compute", False) and NVTE_TEST_NVINSPECT_ENABLED:
+        # use_compile: debug instrumentation forces the eager fallback, so
+        # compile is a no-op there.
```
ptrendx, line 788: Are we testing that anywhere (that debug tools do not crash with torch.compile and just fall back to eager)?

tests/pytorch/distributed/test_comm_gemm_overlap.py
ptrendx: So did you measure the time increase of the CI due to this PR?

tests/pytorch/test_torch_compile.py
```diff
@@ -37,9 +43,9 @@
     is_mxfp8_available,
     is_fp8_block_scaling_available,
     is_nvfp4_available,
+    Float8Quantizer,
```
ptrendx: Are you actually using it?

tests/pytorch/test_torch_compile.py
```diff
@@ -26,9 +32,9 @@
 from transformer_engine.common import recipe
 from transformer_engine.pytorch.constants import FP8FwdTensorIdx, FP8BwdTensorIdx
 from transformer_engine.pytorch.module.base import TransformerEngineBaseModule
+from transformer_engine.pytorch.quantization import QuantizerRole
 from transformer_engine.pytorch.ops.basic.basic_linear import BasicLinear
 from transformer_engine.pytorch.tensor.float8_tensor import Float8CurrentScalingQuantizer
-from transformer_engine.pytorch.quantization import QuantizerRole
 from transformer_engine.pytorch.tensor.nvfp4_tensor import NVFP4Quantizer
```
ptrendx, line 43: Hmmm, there is some inconsistency here - some quantizers you are importing from the exact files,
but some others are imported just from transformer_engine.pytorch a few lines below.

tests/pytorch/test_torch_compile.py
```diff
@@ -37,9 +43,9 @@
     is_mxfp8_available,
     is_fp8_block_scaling_available,
     is_nvfp4_available,
+    Float8Quantizer,
     Float8BlockQuantizer,
     MXFP8Quantizer,
-    NVFP4Quantizer,
 )
 from utils import recipe_id
```
ptrendx: There is currently an issue with cutedsl, where they are exporting their own utils package when
you import them, so this could potentially not work (or stop working when somebody adds a test
that introduces importing cutedsl). See the other files dealing with utils import and do the
same workaround here.

tests/pytorch/test_torch_compile.py
```diff
@@ -1164,3 +1242,321 @@ def test_to_tensor_spec_quantized(factory, shape):
...
+@pytest.mark.parametrize(
+    "fp8_recipe",
+    [None, *_all_recipes],
+    ids=lambda r: "bf16" if r is None else type(r).__name__,
+)
+def test_te_linear_compiles(fp8_recipe, compile_mode):
+    """
+    torch.compile(fullgraph=True) of ``te.Linear`` under every built-in
+    recipe (plus the bf16-only baseline with no autocast), for both the default
+    backend and ``mode="reduce-overhead"`` (CUDA-graph trees).
```
ptrendx, line 1898: Trees?

tests/pytorch/test_torch_compile.py
```diff
@@ -1164,3 +1242,321 @@ def test_to_tensor_spec_quantized(factory, shape):
...
+            assert workspace is not None, f"no cached FP8 weight after step {step}"
+            if step == 0:
+                cached_workspace = workspace
+            else:
+                assert workspace is cached_workspace, f"cache rebuilt at step {step}"
+
+
+@pytest.mark.skipif(not _opaque_available, reason="torch opaque object API not available")
+@pytest.mark.skipif(not fp8_available, reason=reason_for_no_fp8)
+@pytest.mark.xfail(reason="waiting for a PyTorch fix", strict=False)
```
ptrendx: Could we have a PyTorch issue number here?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+
+
+# ``None`` entries in an op's flat ``Tensor[]`` return are smuggled through a
+# 0-element sentinel tensor: a non-nullable ``Tensor[]`` schema is required for
+# ``register_autograd`` to attach a ``grad_fn`` to the outputs. The sentinel is
+# recognized by (numel == 0, dtype), so its dtype must be one no real payload
+# tensor can have -- complex is never a TE payload (quantized data / scales are
+# uint8 / fp32, outputs are float), while e.g. a uint8 sentinel would collide
+# with a genuinely empty FP8 data buffer (empty batch).
+_NONE_SENTINEL_DTYPE = torch.complex32
```
ptrendx, line 127: That is hacky as hell.

ptrendx: Do we have a proposal to pytorch to fix this and allow normal None?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+
+
+class OpaqueValueBundle:
+    """Opaque value-type bundle of simple Python values.
+
+    Wraps a ``{name: value}`` dict so many small non-Tensor args pass through a
+    single custom-op input; registered as a torch.compile *value* opaque type
+    (Dynamo specializes the graph on its contents). Allowed values: primitives
+    in :attr:`PRIMITIVE_TYPES` (incl. ``torch.Size``), ``enum.Enum``, classes,
+    any registered value-opaque type (e.g. TE quantizers), plus nested tuples /
```
ptrendx, line 158: If quantizers are ok here then why do we need a special encoder for quantizers?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+        float,
+        str,
+        torch.dtype,
+        torch.device,
+        torch.Size,
+    )
+
+    @classmethod
+    def is_simple_value(cls, value: Any) -> bool:
+        """Whether ``value`` may be stored inside an instance (recursive)."""
```
ptrendx, line 176: And N

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+        if isinstance(value, dict):
+            return all(isinstance(k, str) and cls.is_simple_value(v) for k, v in value.items())
+        if isinstance(value, (list, tuple)):
+            return all(cls.is_simple_value(v) for v in value)
+        return False
+
+    @classmethod
+    def _to_hashable(cls, value: Any) -> Any:
+        # Tag with the concrete type so e.g. [1] / (1,) / Size([1]) or True / 1
+        # stay distinct under __eq__ / __hash__ (graph guards compare bundles).
```
ptrendx, line 194: Is that really needed though? If the actual held value is the same then we would want the graph
to stay the same as well, no? As it is right now this could lead to more compilations than
necessary.

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+        if isinstance(value, dict):
+            return ("dict", tuple(sorted((k, cls._to_hashable(v)) for k, v in value.items())))
+        if isinstance(value, (list, tuple)):  # incl. torch.Size
+            return (type(value).__name__, tuple(cls._to_hashable(v) for v in value))
+        return (type(value).__name__, value)
+
+    @classmethod
+    def _fmt_simple(cls, value: Any) -> str:
+        """Repr for a value, evaluable in a context with ``torch`` globals."""
```
ptrendx, line 203: Do we actually call eval on those?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+        self._frozen: Tuple[Tuple[str, Any], ...] = tuple(
+            (k, OpaqueValueBundle._to_hashable(v)) for k, v in sorted(data.items())
+        )
+
+    def __getitem__(self, key: str) -> Any:
+        return self._data[key]
+
+    def __getattr__(self, name: str) -> Any:
+        # Underscored names raise cleanly: copy/pickle probe dunders on a clone
+        # created without __init__, where reading ``self._data`` would recurse.
```
ptrendx: ??? This requires some more explanation.

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+        except KeyError as e:
+            raise AttributeError(name) from e
+
+    def get(self, key: str, default: Any = None) -> Any:
+        """Return ``self._data.get(key, default)``."""
+        return self._data.get(key, default)
+
+    def as_dict(self) -> Dict[str, Any]:
+        """Return a shallow copy of the stored mapping."""
+        return dict(self._data)
```
ptrendx, line 255: Do we need all of these functions?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, OpaqueValueBundle):
+            return NotImplemented
+        return self._frozen == other._frozen
+
+    def __hash__(self) -> int:
+        return hash(self._frozen)
```
ptrendx: How often is this function called? Considering during init we already made everything hashable,
we could just compute that hash there and store it.

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+                return
+            if isinstance(value, Enum):
+                globals_[type(value).__name__] = type(value)
+                return
+            if isinstance(value, type):
+                globals_[value.__name__] = value
+                return
+            if isinstance(value, OpaqueValueBundle.PRIMITIVE_TYPES):
+                return
+            if _is_opaque_value_type(type(value)):
```
ptrendx, line 288: Considering that the function called here is actually imported later (and in a try block that
could fail), shouldn't we actually guard this entire class as valid only if the opaque_object
support is there?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+        is_opaque_value_type as _is_opaque_value_type,
+        register_opaque_type,
+    )
+
+    register_opaque_type(OpaqueValueBundle, typ="value")
+    _OPAQUE_VALUE_BUNDLE_TYPE_NAME: Optional[str] = get_opaque_type_name(OpaqueValueBundle)
+# Older torch without opaque_object support.
+except Exception as e:  # pylint: disable=broad-exception-caught  # pragma: no cover
+    record_compile_disabled(
+        f"could not register OpaqueValueBundle as an opaque type ({e}); use a newer PyTorch build"
```
ptrendx, line 309: When will that code execute? Are we going to raise a warning only when somebody actually tries
torch.compile with TE, or every time?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+
+# --------------------------------------------------------------------------- #
+# Storage flatten / unflatten (value-opaque quantizer; no ProcessGroup)
+# --------------------------------------------------------------------------- #
+
+
+def _storage_flatten(
+    value: Any, extra_meta: Optional[Dict[str, Any]] = None
+) -> Tuple["OpaqueValueBundle", List[torch.Tensor]]:
+    """Split a ``QuantizedTensor`` / bare storage into ``(meta, Tensor[])``.
```
ptrendx, line 330: Isn't this essentially what prepare_for_saving() does? Why do we need a new function here?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+
+    A custom op only takes flat, simply-typed arguments, but a TE op takes a
+    single ``@dataclass`` of mixed fields. Each adapter knows how to translate
+    its kind of field both ways. ``try_build`` and ``schema_slots`` run once at
+    registration (to build the op's schema); ``to_slots`` and ``from_slots`` run
+    on each call and must agree on the slot layout that ``schema_slots`` declares.
+    """
+
+    @classmethod
+    def try_build(cls, name: str, annot: Any) -> Optional["_Adapter"]:
```
ptrendx: This is pretty strange design to be honest. Shouldn't there be just a factory of the adapters
from the annotations instead, rather than holding all of those classes and trying one by one?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+        if value is None:
+            return {
+                self.tensor_slot(): None,
+                self.inner_slot(): [],
+                self.meta_slot(): OpaqueValueBundle({self.KIND_KEY: _TensorOrQuantizedKind.NONE}),
+            }
+        if isinstance(value, torch.Tensor):
+            # Plain tensor *and* subclass (e.g. Float8Tensor) pass through the
+            # ``Tensor?`` slot; subclass flattening (if any) is done by the
+            # wrapper op's ``register_torch_dispatch`` rule.
```
ptrendx: Wouldn't it be easier then to just check for storage first and treat the QuantizedTensor as
storage?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+    if _is_union(annot):
+        args = get_args(annot)
+        if type(None) in args:
+            non_none = [a for a in args if a is not type(None)]
+            if len(non_none) == 1:
+                return non_none[0], True
+    return annot, False
+
+
+class _Adapter:
```
ptrendx: To be honest this entire structure seems to be pretty overcomplicated for what I believe it does.
I think that a simple factory function that takes a list of annotations and produces the schema (I assume that is needed for torch.compile machinery) as well
as list or dictionary of the "adapters" (which could be just regular functions, not bespoke objects)

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+    def grad_slot(self) -> Optional[int]:
+        return 0
+
+
+class _QuantizerAdapter(_Adapter):
+    """``Quantizer`` / ``Optional[Quantizer]`` -> one own ``OpaqueValueBundle`` slot.
+
+    Each quantizer gets its own dedicated slot. The field is annotated with the
+    base ``Quantizer`` (not itself a registered opaque type), so the simple
+    bundle would not claim it.
```
ptrendx: I don't understand the concept of ownership and "claiming" here. Could you explain that?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+    """
+
+    NAME_KEY = "group_name"
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+    def meta_slot(self) -> str:
+        """Group-name slot name."""
+        return self.name + "__pg"
```
ptrendx: Why do we need to attach suffixes to the names of the arguments?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+
+    def from_slots(self, args: Dict[str, Any], kwargs: Dict[str, Any]) -> None:
+        name = args[self.meta_slot()][self.NAME_KEY]
+        kwargs[self.name] = None if name is None else _resolve_process_group(name)
+
+
+class _SimpleBundleAdapter(_Adapter):
+    """Aggregates every simple-typed field into a single OpaqueValueBundle.
+
+    Unlike the per-field adapters, at most one of these exists per op (none if the
```
ptrendx: I don't understand why we cannot put everything into a single bundle.

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+    Such a field cannot cross the op boundary, so it emits no slot and is
+    tolerated only when its runtime value carries nothing: ``to_slots`` accepts
+    ``None`` / an all-``None`` sequence (e.g. an unset ``Optional[Any]`` field,
+    or an empty list, on the compiled path) and ``from_slots`` restores it as
+    ``None``. A non-trivial value means the config is genuinely unsupported
+    under torch.compile, and ``to_slots`` raises.
+
+    The check must run at call time (not in ``_get_adapters``): the annotation
+    alone -- e.g. ``Optional[Any]`` -- is valid when the value is ``None``, so
+    only the runtime value can decide.
```
ptrendx: Does that mean that this will run every time the compiled function is invoked?

transformer_engine/pytorch/dynamo/custom_op.py
```diff
@@ -0,0 +1,1475 @@
...
+
+
+def _spec_slot_count(spec: Optional[TensorSpec]) -> int:
+    """Flat ``Tensor[]`` slots the value for ``spec`` occupies."""
+    if spec is None:
+        return 1
+    return len(spec.inner_names())
+
+
+def _unflatten_value(
```
ptrendx: This function is an example of code that has 0 value (as it is a 1-liner 
`spec.assemble(chunk) if spec is not None else None`) and lives here just to increase the
line count.

tests/pytorch/test_torch_compile.py
pggPL: this is name of this mechanism, heavily used

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: We had https://github.com/pytorch/pytorch/pull/187434, but pytorch reviewers were not enthusiastic due to reason: "In particular, Optional tensor returns are not actually supported well due to None<>UndefinedTensor confusion"

tests/pytorch/test_torch_compile.py
pggPL: done

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: good point, removed separate adapter

tests/pytorch/distributed/run_numerics.py
pggPL: added the test

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: what do you mean?

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: I don't think it will matter substantially - someone would need to use torch.size/list/int as one argument in TE. 
Graphs which would have torch.size as an arg may be different than graphs which use int - note that it is reconstructed from bare value and then used - something in op may accept torch.size but not accept list.

So I think it is ok.

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: Yes

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: This code executes at import, but it only saves the reason. Warning is raised when somebody actually tries
torch.compile with TE

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: It's similar, but it uses tensor_flatten inside. I can imagine we can use it instead  of prepare_for_saving in modules, but I guess CPU overhead will be higher. Also, this method does not destroy input objects.

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: it turned out I didn't need __get_attr__ at all in current version, removed

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: Good point, done

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: removed getattr and get, other are needed by pytorch or used

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: removed anyways

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: Yes.

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: We can put qunatizers and processgroup (it was not possible some multiple commits ago, but now we can do it). 
We cannot put any tensors inside it (it is opaque value object which cannot contain tensors).

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: removed this line anyways, sometimes (in different adapters) we have multiple cusotom op arguments per Te op argument -> list of tensors and meta bundle for qunatized tensor argument

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: The Float8Tensor is passed as a tensor and dismantled in torch dispatch. It enables gradient flow through it.

transformer_engine/pytorch/dynamo/custom_op.py
pggPL: The agent comment I agree with


> Largely addressed in the current revision, in the direction you describe: adapter selection is now a plain factory (`_build_field_adapter(name, annot)`), `_build_schema` derives the schema from the adapters, and the structure itself shrank — the per-field registry, `_QuantizerAdapter` and `_ProcessGroupAdapter` are gone (quantizers and process groups now ride in the single shared bundle), netting about −100 lines in this file versus the revision you reviewed.
The one part I'd keep as-is are the adapter *objects*: each field kind is four operations that must agree on its slot layout (`schema_slots`/`to_slots`/`from_slots`/`grad_slot`) plus shared per-field state, and downstream code dispatches on the adapter kind (`_tensor_field_names`, `_tensor_or_quantized_offsets`) — as plain functions that becomes an anonymous tuple of closures plus parallel tagging structures, i.e. the same complexity without names. Only two per-field kinds remain, both tensor-shaped.

tests/pytorch/distributed/test_comm_gemm_overlap.py
pggPL: 1.5 min for L0(not just this PR, all torch.compile tests), ~5min for L1

transformer_engine/pytorch/dynamo/custom_op.py
ptrendx: Hmmm, not sure to be honest. NVM.

ptrendx (dismissed):

ptrendx (approved):

PR #3056, opened 2026-05-28
[JAX] [PyT] [Common] Enable D=256 BWD cuDNN fused attn for Blackwell CC 10.x 

KshitijLakhani:
# Description

Support for D=256 BWD for Blackwell CC 10x via the C++ API (which TE uses) was added in cuDNN 9.23 + cuDNN FE 1.24 (BSHD) and cuDNN 9.25 + cuDNN FE 1.26 (THD)
Enabling this support in TE attention and adding conditional tests for the same

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes
 
 Add guard when picking the backend (sub backend) in TE common.
 Add tests for D=256 case in TE PyT and TE JAX

# Checklist:

### DIFF
--- a/tests/jax/test_fused_attn.py
+++ b/tests/jax/test_fused_attn.py
@@ -511,16 +511,16 @@ class FusedAttnRunner:
                     "D=256 BWD on Blackwell requires cuDNN 9.23 or newer;"
                     f" got cuDNN {cudnn_version}."
                 )
-            # Non-learnable bias is fine (bias is allowed as an input); only dBias is
-            # unsupported. The JAX runner asks for dBias iff the bias shape is [1, h, s, s]
-            # (see test_backward), so gate on that.
+            # TODO(KshitijLakhani): cuDNN FE can model bias input separately from dBias,
+            # but TE does not yet plumb whether dBias is requested into the common backend selector. 
+            # Until that distinction is available, the D=256 SM10x gate requires no bias.
             unsupported = None
             if self.attn_bias_type == AttnBiasType.PRE_SCALE_BIAS:
                 unsupported = "pre-scale bias"
-            elif self.attn_bias_type != AttnBiasType.NO_BIAS and self.bias_shape == BiasShape._1HSS:
+            elif self.attn_bias_type != AttnBiasType.NO_BIAS:
                 unsupported = (
-                    "bias gradients (dBias); frozen/non-learnable bias inputs"
-                    " (i.e. non-1HSS bias shapes) are supported"
+                    "post-scale bias in TE's D=256 backend gate; bias-input-only"
+                    " support needs TE to distinguish between bias input and dBias"
                 )
             elif self.dropout_prob != 0.0:
                 unsupported = "dropout"
--- a/tests/jax/test_fused_attn.py
+++ b/tests/jax/test_fused_attn.py
@@ -497,7 +497,7 @@ class FusedAttnRunner:
         cudnn_version = get_cudnn_version()
         # D=256 bprop on SM10x uses the deterministic algorithm path only. BSHD support
         # starts with cuDNN FE 1.24 / BE 9.23; THD execution-plan support starts with
-        # cuDNN FE 1.26 / BE 9.30. The kernel rejects dBias, dropout, and ALiBi, supports vanilla
+        # cuDNN FE 1.26 / BE 9.25. The kernel rejects dBias, dropout, and ALiBi, supports vanilla
         # softmax only, and allows SWA together with a causal mask only.
         is_sm10x = 100 <= compute_capability < 110
         if self.is_training and is_sm10x and (self.head_dim_qk == 256 or self.head_dim_v == 256):
@@ -506,8 +506,8 @@ class FusedAttnRunner:
                     "D=256 BWD on Blackwell only supports d_qk == d_v == 256;"
                     f" got d_qk={self.head_dim_qk}, d_v={self.head_dim_v}."
                 )
-            required_cudnn_version = 93000 if self.qkv_layout.is_thd() else 92300
-            required_cudnn_version_label = "9.30" if self.qkv_layout.is_thd() else "9.23"
+            required_cudnn_version = 92500 if self.qkv_layout.is_thd() else 92300
+            required_cudnn_version_label = "9.25" if self.qkv_layout.is_thd() else "9.23"
             if cudnn_version < required_cudnn_version:
                 pytest.skip(
                     f"D=256 BWD on Blackwell with {self.qkv_layout} requires cuDNN"
@@ -1694,7 +1694,7 @@ class TestFusedAttn:
             id="2-1024-2048-12-6-128-64-BF16-CROSS-GQA-RAGGED_SEPARATE",
         ),
         # D=256 deterministic backward on the SM100 dedicated SDPA bprop kernel.
-        # BSHD requires cuDNN FE 1.24 / BE 9.23+; THD requires cuDNN FE 1.26 / BE 9.30+.
+        # BSHD requires cuDNN FE 1.24 / BE 9.23+; THD requires cuDNN FE 1.26 / BE 9.25+.
         pytest.param(
             4,
             128,
--- a/tests/jax/test_fused_attn.py
+++ b/tests/jax/test_fused_attn.py
@@ -469,7 +469,6 @@ class FusedAttnRunner:
             return 1
 
     def _check_configs(self):
-        # TODO(rewang): probably adds this in is_fused_attn_available
         # TODO(KshitijLakhani): probably add/move this to is_fused_attn_available
         if self.qkv_layout.is_thd() and not self.attn_mask_type.is_padding():
             pytest.skip("THD format requires padding masks.")
--- a/tests/jax/test_fused_attn.py
+++ b/tests/jax/test_fused_attn.py
@@ -1695,6 +1695,21 @@ class TestFusedAttn:
             QKVLayout.THD_THD_THD,
             id="2-1024-2048-12-6-128-64-BF16-CROSS-GQA-RAGGED_SEPARATE",
         ),
+        # D=256 deterministic backward on the SM100 dedicated SDPA bprop kernel
+        # (cuDNN FE 1.24 / BE 9.23+). Unsupported configs (e.g. dBias, non-256 head dims)
+        # are skipped by FusedAttnRunner._check_configs.
+        pytest.param(
+            4,
+            128,
+            128,
+            16,
+            16,
+            256,
+            256,
+            jnp.float16,
+            QKVLayout.BSHD_BS2HD,
+            id="4-128-128-16-16-256-256-FP16-SELF-KV_PACKED",
+        ),
     ],
 )
 @pytest.mark.parametrize(
@@ -1807,113 +1822,6 @@ class TestFusedAttnWithDeterminism:
     ):
         """
         Test backward with parameterized configs
-        """
-        TestFusedAttn.test_backward(
-            b,
-            s_q,
-            s_kv,
-            h_q,
-            h_kv,
-            d_qk,
-            d_v,
-            attn_bias_type,
-            attn_mask_type,
-            softmax_type,
-            dropout_prob,
-            dtype,
-            qkv_layout,
-            bias_shape,
-            swa,
-            seq_desc_format,
-        )
-
-
-@pytest.mark.parametrize(
-    "attn_mask_type",
-    [
-        pytest.param(AttnMaskType.NO_MASK, id="NO_MASK"),
-        pytest.param(AttnMaskType.PADDING_MASK, id="PADDING"),
-        pytest.param(AttnMaskType.CAUSAL_MASK, id="CAUSAL"),
-        pytest.param(AttnMaskType.PADDING_CAUSAL_MASK, id="PADDING_CAUSAL"),
-        pytest.param(
-            AttnMaskType.PADDING_CAUSAL_BOTTOM_RIGHT_MASK, id="PADDING_CAUSAL_BOTTOM_RIGHT"
-        ),
-    ],
-)
-@pytest.mark.parametrize(
-    "softmax_type",
-    [
-        pytest.param(AttnSoftmaxType.VANILLA_SOFTMAX, id="VANILLA_SOFTMAX"),
-    ],
-)
-@pytest.mark.parametrize(
-    "dropout_prob",
-    [
-        pytest.param(0.0, id="DROP_0.0"),
-    ],
-)
-@pytest.mark.parametrize(
-    "swa",
-    [
-        pytest.param(False, id="NO_SWA"),
-    ],
-)
-@pytest.mark.parametrize(
-    "seq_desc_format",
-    [
-        pytest.param(SeqDescFormat.Seqlens, id="Seqlens"),
-    ],
-)
-@pytest.mark.skipif(not _deterministic, reason="Test determinism only")
-class TestFusedAttnD256WithDeterminism:
-    """
-    Fused attention D=256 deterministic backward tester.
-    """
-
-    @staticmethod
-    @pytest.mark.parametrize(
-        "b, s_q, s_kv, h_q, h_kv, d_qk, d_v, dtype, qkv_layout",
-        [
-            pytest.param(
-                4,
-                128,
-                128,
-                16,
-                16,
-                256,
-                256,
-                jnp.float16,
-                QKVLayout.BSHD_BS2HD,
-                id="4-128-128-16-16-256-256-FP16-SELF-KV_PACKED",
-            ),
-        ],
-    )
-    @pytest.mark.parametrize(
-        "attn_bias_type, bias_shape",
-        [
-            pytest.param(AttnBiasType.NO_BIAS, None, id="NO_BIAS"),
-        ],
-    )
-    def test_backward(
-        b,
-        s_q,
-        s_kv,
-        h_q,
-        h_kv,
-        d_qk,
-        d_v,
-        attn_bias_type,
-        attn_mask_type,
-        softmax_type,
-        dropout_prob,
-        dtype,
-        qkv_layout,
-        bias_shape,
-        swa,
-        seq_desc_format,
-    ):
-        """
-        Test D=256 backward with the supported deterministic SM100 bprop configuration.
         """
         TestFusedAttn.test_backward(
             b,
--- a/tests/jax/test_fused_attn.py
+++ b/tests/jax/test_fused_attn.py
@@ -512,7 +512,7 @@ class FusedAttnRunner:
                     f" got cuDNN {cudnn_version}."
                 )
             # TODO(KshitijLakhani): cuDNN FE can model bias input separately from dBias,
-            # but TE does not yet plumb whether dBias is requested into the common backend selector. 
+            # but TE does not yet plumb whether dBias is requested into the common backend selector.
             # Until that distinction is available, the D=256 SM10x gate requires no bias.
             unsupported = None
             if self.attn_bias_type == AttnBiasType.PRE_SCALE_BIAS:
--- a/tests/jax/test_fused_attn.py
+++ b/tests/jax/test_fused_attn.py
@@ -1705,6 +1705,22 @@ class TestFusedAttn:
             QKVLayout.BSHD_BS2HD,
             id="4-128-128-16-16-256-256-FP16-SELF-KV_PACKED",
         ),
+        pytest.param(
+            4,
+            128,
+            128,
+            16,
+            16,
+            256,
+            256,
+            jnp.float16,
+            QKVLayout.THD_T2HD,
+            id="4-128-128-16-16-256-256-FP16-SELF-RAGGED_KV_PACKED",
+            marks=pytest.mark.xfail(
+                reason="cuDNN 9.23 D=256 BWD currently does not build a THD execution plan.",
+                strict=True,
+            ),
+        ),
     ],
 )
 @pytest.mark.parametrize(
--- a/tests/jax/test_fused_attn.py
+++ b/tests/jax/test_fused_attn.py
@@ -520,6 +520,59 @@ class FusedAttnRunner:
                 "is either BSHD_BSHD_BSHD or THD_THD_THD"
             )
 
+        # D=256 bprop on SM10.x uses cuDNN's dedicated SDPA bprop kernel
+        # (cuDNN FE 1.24 / BE 9.23+). FE forces this path onto the deterministic algorithm path,
+        # which rejects dBias, dropout, and ALiBi. It supports vanilla softmax only and allows SWA
+        # together with a causal mask only.
+        compute_capability = get_device_compute_capability(0)
+        is_sm10x = 100 <= compute_capability < 110
+        if self.is_training and is_sm10x and (self.head_dim_qk == 256 or self.head_dim_v == 256):
+            if self.head_dim_qk != 256 or self.head_dim_v != 256:
+                pytest.skip(
+                    "D=256 BWD on Blackwell only supports d_qk == d_v == 256;"
+                    f" got d_qk={self.head_dim_qk}, d_v={self.head_dim_v}."
+                )
+            cudnn_version = get_cudnn_version()
+            if cudnn_version < 92300:
+                pytest.skip(
+                    "D=256 BWD on Blackwell requires cuDNN 9.23 or newer;"
+                    f" got cuDNN {cudnn_version}."
+                )
+            # Non-learnable bias is fine (bias is allowed as an input); only dBias is
+            # unsupported. The JAX runner asks for dBias iff the bias shape is [1, h, s, s]
+            # (see test_backward), so gate on that.
+            unsupported = None
+            if self.attn_bias_type == AttnBiasType.PRE_SCALE_BIAS:
+                unsupported = "pre-scale bias"
+            elif (
+                self.attn_bias_type != AttnBiasType.NO_BIAS
+                and self.bias_shape == BiasShape._1HSS
+            ):
+                unsupported = (
+                    "bias gradients (dBias); frozen/non-learnable bias inputs"
+                    " (i.e. non-1HSS bias shapes) are supported"
+                )
+            elif self.dropout_prob != 0.0:
+                unsupported = "dropout"
+            elif self.softmax_type != AttnSoftmaxType.VANILLA_SOFTMAX:
+                unsupported = "non-vanilla softmax"
+            if unsupported is not None:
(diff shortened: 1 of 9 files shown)


transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -315,6 +315,21 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
...
+           // Blackwell rejects dBias, dropout, and ALiBi (and supports vanilla softmax only).
+           bias_type == NVTE_Bias_Type::NVTE_NO_BIAS && dropout == 0.0 &&
+           softmax_type == NVTE_Softmax_Type::NVTE_VANILLA_SOFTMAX &&
+           // Non-causal D=256 supports only full-window attention; SWA is allowed only for causal masks.
+           ((window_size_left == -1 && window_size_right == -1) ||
+            ((attn_mask_type == NVTE_Mask_Type::NVTE_CAUSAL_MASK ||
+              attn_mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_MASK ||
+              attn_mask_type == NVTE_Mask_Type::NVTE_CAUSAL_BOTTOM_RIGHT_MASK ||
+              attn_mask_type == NVTE_Mask_Type::NVTE_PADDING_CAUSAL_BOTTOM_RIGHT_MASK) &&
+             (window_size_right == -1 || window_size_right == 0)))) ||
```
cyanguwa: Could these changes be moved to before "\\ bias type" just so it's following an increasing order of the cuDNN version?

tests/pytorch/attention/test_attention.py
```diff
@@ -377,6 +377,38 @@ def test_dpa_fa4_hdim256(dtype, model_configs, model):
     test_dot_product_attention(dtype, model_configs, model, False, True, None, False, False)
 
 
+# cuDNN FusedAttention D=256 bprop is supported on sm10x from cuDNN 9.23 (FE 1.24),
+# via the dedicated deterministic SDPA bprop kernel, which supports d_qk == d_v == 256 only,
+# vanilla type of softmax only, no dropout, no ALiBi, and (for non-causal masks) full-window attention only.
+# (for non-causal masks) full-window attention.
```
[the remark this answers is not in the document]
cyanguwa: Seems like some editing glitch :)

tests/jax/test_fused_attn.py
```diff
@@ -444,6 +444,56 @@ def _check_configs(self):
...
+                    pytest.skip(
+                        "D=256 BWD on Blackwell uses the SM100 D=256 SDPA BWD kernel"
+                        " which requires window_size=(-1, -1) for non-causal masks."
+                    )
+                if self.window_size[1] not in (-1, 0):
+                    pytest.skip(
+                        "D=256 BWD on Blackwell only supports right window -1 or 0"
+                        " for causal masks."
+                    )
+
```
cyanguwa, line 547: Aren't these checks duplicate to the checks we added on the C++ side? Would the call `FusedAttnHelper().get_fused_attn_backend()` give you the same gating effect?

transformer_engine/common/fused_attn/fused_attn.cpp
cyanguwa: Does this new feature support BSHD/SBHD and THD? It looks like the tests are focused on BSHD/SBHD only.

cyanguwa (changes requested):

transformer_engine/common/fused_attn/fused_attn.cpp
KshitijLakhani: RE: THD support
I did test BSHD and BSHD+CP and it did pass on the JAX side and the CI for the PyT side did not fail either so I think that works.
My testing revealed that THD support is not yet available (Bwd plan compialtion issue) so I've filed a bug and shared a reproducer for the same with the cuDNN team: https://github.com/NVIDIA/cudnn-frontend/issues/276

tests/jax/test_fused_attn.py
KshitijLakhani: So if we are just interested in the gating effect, you are right. The `get_fused_attn_backend()` will return `NVTE_No_Backend` and then there's a catch-all at the end which basically skip the tests as there is no fused attn backend avalable.

However, the reason for this to be here is to give a meaningful reason as to why a test is being skipped as compared to a generic "Unsupported inputs combination or device compute capability." message which does not qualify the reason for the skip. Unfortunately, on the JAX attn side we do not log the reason for disabling fused attn in the feature code like we have on the Pytorch side in `d_p_a/utils.py`. So there is no way for the user to know why the test was skipped. Hence, we need to rely on test code to log this on the JAX side.

I'd suggest we leave this in here for now. And when your PR for generating log messages in the C++ level when selecting the attn backend is ready, I can plumb it through onto the JAX side and then as part of that clean up, get rid of all the skip messages in check_configs()

tests/pytorch/attention/test_attention.py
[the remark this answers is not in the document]
KshitijLakhani: Fixed in 379242be28c19d05140544fe38a3079f7d02e224

tests/jax/test_fused_attn.py
```diff
@@ -444,6 +444,56 @@ def _check_configs(self):
...
+            # unsupported. The JAX runner asks for dBias iff the bias shape is [1, h, s, s]
+            # (see test_backward), so gate on that.
+            unsupported = None
+            if self.attn_bias_type == AttnBiasType.PRE_SCALE_BIAS:
+                unsupported = "pre-scale bias"
+            elif self.attn_bias_type != AttnBiasType.NO_BIAS and self.bias_shape == BiasShape._1HSS:
+                unsupported = (
+                    "bias gradients (dBias); frozen/non-learnable bias inputs"
+                    " (i.e. non-1HSS bias shapes) are supported"
+                )
```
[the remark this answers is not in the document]
KshitijLakhani: Fixed in b8fe919a1746e5b3d2a824115602b204e1e594a2

transformer_engine/common/fused_attn/fused_attn.cpp
KshitijLakhani: > Could these changes be moved to before "\ bias type" just so it's following an increasing order of the cuDNN version?

Fixed in a264de10ad69201d28ae22e5c600235955b313dd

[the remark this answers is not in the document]
KshitijLakhani: This support is forwarding looking, i.e., support is added for THD and BSHD in the TE common fused attn backend checking code, however, the PR is still waiting on cuDNN to fix support for THD.
The current PR will not be merged as is. One of two things will happen:
1. cuDNN will fix THD support and only then will this PR be merged (most likely) - after fixing the XFAIL for THD cases to skips for a specific cuDNNv version
2. cuDNN will not fix this soon in which case I will switch the support to BSHD only prior to merging this PR
3.

KshitijLakhani:
Latest testing: Pipeline [56553160]
I manually set the cuDNN BE in the CI to be 9.30+ and the cuDNN FE to point at latest `develop` as of 07/01/26 to test the THD support for D=256
I confirmed that the newly added tests for BSHD and THD (PyT and JAX, both) pass

tests/jax/test_fused_attn.py
```diff
@@ -493,11 +494,62 @@ def _check_configs(self):
...
+            required_cudnn_version = 93000 if self.qkv_layout.is_thd() else 92300
+            required_cudnn_version_label = "9.30" if self.qkv_layout.is_thd() else "9.23"
+            if cudnn_version < required_cudnn_version:
+                pytest.skip(
+                    f"D=256 BWD on Blackwell with {self.qkv_layout} requires cuDNN"
+                    f" {required_cudnn_version_label} or newer; got cuDNN {cudnn_version}."
+                )
+            # TODO(KshitijLakhani): cuDNN FE can model bias input separately from dBias,
+            # but TE does not yet plumb whether dBias is requested into the common backend selector.
+            # Until that distinction is available, the D=256 SM10x gate requires no bias.
```
KshitijLakhani, line 518: I've opened a GH issue for this: https://github.com/NVIDIA/TransformerEngine/issues/3188

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -327,7 +330,23 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
              attn_mask_type != NVTE_Mask_Type::NVTE_PADDING_CAUSAL_MASK))) ||
           // 9.11: d_qk = 192, d_v = 128 + Blackwell + bprop + non-paged
           (head_dim_qk == 192 && head_dim_v == 128 && is_training && sm_arch_ >= 100 &&
-           cudnn_runtime_version >= 91100)) &&
+           cudnn_runtime_version >= 91100) ||
+          // 9.23: d_qk = d_v = 256 + SM10x (cuDNN FE 1.24 / BE 9.23+) + bprop + non-paged.
+          // THD layouts require cuDNN FE 1.26 / BE 9.30+ for execution-plan support.
+          (head_dim_qk == 256 && head_dim_v == 256 && is_training && sm_arch_ >= 100 &&
+           sm_arch_ < 110 && cudnn_runtime_version >= (is_thd_layout ? 93000 : 92300) &&
```
jepio: The PR body says this should be `92500`, not `93000`. Which one is correct?

KshitijLakhani: Good catch !
I've updated the PR body to reflect cuDN 9.30. I believe the body was a preemptive action from me while I was waiting for cuDNN to add support for THD via the C++ API.

KshitijLakhani:
Latest testing: Pipeline [58978691]
I manually set the cuDNN BE in the CI to be 9.30+ and the cuDNN FE to be 1.26 to test the THD support for D=256
I confirmed that the newly added tests for BSHD and THD (PyT and JAX, both) pass for non-CP and CP

transformer_engine/common/fused_attn/fused_attn.cpp
cyanguwa: I think cuDNN has renamed 9.30 to 9.25? Please double check!

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -240,6 +240,9 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
   NVTE_QKV_Format qkv_format = nvte_get_qkv_format(qkv_layout);
   NVTE_QKV_Format q_format = nvte_get_q_format(qkv_layout);
   NVTE_QKV_Format kv_format = nvte_get_kv_format(qkv_layout);
+  const bool is_thd_layout = q_format == NVTE_QKV_Format::NVTE_THD ||
+                             kv_format == NVTE_QKV_Format::NVTE_THD ||
+                             qkv_format == NVTE_QKV_Format::NVTE_THD;
```
cyanguwa: is_thd_layout ORs q_format, kv_format, and the combined qkv_format against NVTE_THD — the third is redundant with the first two, but harmless.

tests/pytorch/attention/test_attention.py
```diff
@@ -371,6 +372,58 @@ def test_dpa_fa4_hdim256(dtype, model_configs, model):
...
+    "fused_hd256_padding": ModelConfig(2, 512, 16, 256, attn_mask_type="padding"),
+    # SWA is allowed only together with a causal mask on the D=256 bprop kernel.
+    "fused_hd256_causal_swa": ModelConfig(
+        2, 1024, 16, 256, attn_mask_type="causal", window_size=(128, 0)
+    ),
+    # GQA variant (num_gqa_groups < num_heads).
+    "fused_hd256_padding_causal_gqa": ModelConfig(
+        2, 1024, 16, 256, num_gqa_groups=4, attn_mask_type="padding_causal"
+    ),
+}
```
cyanguwa, line 398: Can we reuse some of the configs in other tests (I think there are some h256 ones as well), or can we keep this dict and remove those other h256 ones to deduplicate?

tests/pytorch/attention/test_attention.py
```diff
@@ -371,6 +372,58 @@ def test_dpa_fa4_hdim256(dtype, model_configs, model):
...
+        pytest.param(
+            "thd_t2hd",
+            id="THD_KV_PACKED",
+            marks=pytest.mark.skipif(
+                get_cudnn_version() < (9, 30, 0),
+                reason="cuDNN 9.30+ is required for THD D=256 fused-attn backward.",
+            ),
+        ),
+    ],
+)
```
cyanguwa, line 426: No problem with the "id", but it seems inconsistent with other PyTorch tests. Also, please check the 9.30 vs 9.25 version guard.

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -483,6 +483,16 @@ def test_cp_with_flash_attention(cp_pool, dtype, model, qkv_format, cp_comm_type
 }
 
 
+model_configs_fused_attn_hdim256 = {
+    # cuDNN FusedAttention D=256 bprop is Blackwell server only. Keep these
+    # outside the generic CP sweep so the D=256 signal is focused and inexpensive.
+    "cp_hd256_causal": ModelConfig(2, 1024, 16, 256, attn_mask_type="causal"),
+    "cp_hd256_causal_swa": ModelConfig(
+        2, 1024, 16, 256, attn_mask_type="causal", window_size=(128, 0)
+    ),
```
cyanguwa: Should we test no_mask, padding_causal as well, or not really?

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -696,3 +706,125 @@ def test_cp_with_fused_attention(
...
+]
+
+
+@pytest.mark.skipif(
+    get_device_compute_capability() not in ((10, 0), (10, 3)),
+    reason="cuDNN FusedAttention head_dim=256 backward is SM100/SM103-only.",
+)
+@pytest.mark.parametrize("dtype", ["bf16", "fp16"])
+@pytest.mark.parametrize("model,qkv_format,cp_comm_type", DPA_CP_FUSED_HD256_CASES)
+def test_cp_with_fused_attention_hdim256(cp_pool, dtype, model, qkv_format, cp_comm_type):
```
cyanguwa: Could we fold this into the regular test_cp_with_fused_attention() tests instead of creating this extra test code and extra tests? You can modify some of the existing tests to cover the d256 cases as well, or force some skips if you think it's adding too many tests (I feel your pytest params above might have enumerated all the combinations anyway). Thanks.

Also, please use d256 as the reference, same as in the title, instead of hdim256. (nit)

tests/jax/test_fused_attn.py
```diff
@@ -493,11 +493,62 @@ def _check_configs(self):
...
                 "seqlen_q > seqlen_kv is not supported with sliding window attention in cuDNN"
             )
+        compute_capability = get_device_compute_capability(0)
+        cudnn_version = get_cudnn_version()
+        # D=256 bprop on SM10x uses the deterministic algorithm path only. BSHD support
+        # starts with cuDNN FE 1.24 / BE 9.23; THD execution-plan support starts with
+        # cuDNN FE 1.26 / BE 9.30. The kernel rejects dBias, dropout, and ALiBi, supports vanilla
+        # softmax only, and allows SWA together with a causal mask only.
+        is_sm10x = 100 <= compute_capability < 110
+        if self.is_training and is_sm10x and (self.head_dim_qk == 256 or self.head_dim_v == 256):
```
cyanguwa, line 503: Shouldn't these checks already be covered on the C side? Are these by any chance duplicate?

tests/jax/test_fused_attn.py
cyanguwa: #2964 will plumb bias/dbias shapes/requirements to the C side, so hopefully when that's merged, this wouldn't be an issue.

cyanguwa (changes requested):

transformer_engine/common/fused_attn/fused_attn.cpp
KshitijLakhani: That's right. Let me get rid of it and push a commit

tests/pytorch/attention/test_attention.py
KshitijLakhani: Will do

tests/jax/test_fused_attn.py
KshitijLakhani: IIUC, is your concern the same as you had in your earlier review: https://github.com/NVIDIA/TransformerEngine/pull/3056#discussion_r3359564828. ? I think my comment goes into more details. Happy to answer any follow ups on that.

I will get rid of these extraneous checks in the tests (if https://github.com/NVIDIA/TransformerEngine/pull/2964 does not already) once https://github.com/NVIDIA/TransformerEngine/pull/2964 is merged in. For now, I think we should let them be in. The clean up for these test messages can follow up once we've confirmed that a more descriptive string is plumbed through to the tests

tests/pytorch/attention/test_attention_with_cp.py
KshitijLakhani: > Also, please use d256 as the reference, same as in the title, instead of hdim256. (nit)

Done. Do you want me to change it for the Flas Attn tests too ?

cyanguwa: Yes, please!

tests/jax/test_fused_attn.py
cyanguwa: Sounds good - are there any CP specific skips/conditions we should add?

#2964 will surface the reject reason to Jax FusedAttnHelper.get_fused_attn_backend() so it'd be nice to use that as a single source of rejection/pass reasons. The d256 feature reject reason does come from cudnn-frontend so we should be able to remove some of the code (and reduce some maintenance cost).

tests/pytorch/attention/test_attention.py
KshitijLakhani: I considered reusing the existing D=256 configs, but they are tied to other coverage: decode-style generic tests, mask/layout sweeps (using post scale bias which would be rejected by cuDNN D=256 bprop path), FA4, or MLA where d_v != 256 (hence not the same D=256 cuDNN kernel path either).

Changing them would either remove existing coverage or mix cuDNN D=256-specific gates into generic tests. I kept this as a small dedicated dict so the cuDNN FusedAttention D=256 bprop surface is explicit and isolated. 

If you have a strong opinion on this, I can arm twist the code to do so

tests/pytorch/attention/test_attention_with_cp.py
KshitijLakhani: I think it's okay to not test those two, if you are fine with it ?
The non-CP tests cover no_mask and I believe THD CP anyways translates to padding_causal under the hood, so I decided to keep this lightweight.

tests/pytorch/attention/test_attention_with_cp.py
KshitijLakhani: Regarding folding this into the regular test_cp_with_fused_attention() tests, that was my first approach actually but I dropped it in favor of the current approach.

The current test_cp_with_fused_attention() is a pretty broad sweep which would add quite a few test cases (while still missing on quite a few as essential mode does not cover BSHD and FP16, both of which we want for D=256). Most of the collected test cases would just be skips adding unnecessary noise in the logs (e.g. fp8, diff mask combos, etc.)  so I decided to take a more funneled approach.

The focused test is clearer: it directly covers BSHD and THD, fp16 and bf16, P2P and all-gather, plus THD SWA, without polluting the broad existing CP matrix.

Are you okay with it being as is @cyanguwa ?

transformer_engine/common/fused_attn/fused_attn.cpp
KshitijLakhani: You are right indeed. Thanks for the note. I've made changes for the same

tests/pytorch/attention/test_attention_with_cp.py
cyanguwa: "essential mode does not cover BSHD and FP16, both of which we want for D=256"

Do we have SBHD tests in the essential mode instead? They should be equivalent to testing BSHD though.

I feel the D256 feature shouldn't be that special to have its own test function, even though the kernels are written in cutedsl, different from the regular SDPA. But we seem to be always expanding the range of head dims we support so do we want to add a new test every time, and would the test code become a bit repetitive.

tests/pytorch/attention/test_attention.py
cyanguwa: Could you remove the 'id' fields actually - they just look really odd sitting next to other tests with no 'id's. :D

tests/pytorch/attention/test_attention_with_cp.py
KshitijLakhani: > Do we have SBHD tests in the essential mode instead? They should be equivalent to testing BSHD though.

You are right about that

KshitijLakhani: I folded in the  D=256 tests into the test_cp_with_fused_attention and added a bunch of skips in there for D=256 in commit https://github.com/NVIDIA/TransformerEngine/pull/3056/changes/9921b9983f1ee3d5e202d21dadb66c0cb9b21b41
This should cover most of the earlier D=256 tests I had added. 
The only differences are: 
- Older D=256 tests ran BF16 and FP16 in essential mode but the Newer D=256 tests only run BF16 (FP8 is not supported) and FP16 is ran as non-essential, which is consistent with other fused attn tests
- Older D=256 tests ran BSHD in essential mode but the Newer D=256 tests run SBHD instead, which is consistent with the other fused attn tests

So with this commit I think we are consistent throughout

KshitijLakhani:
Latest testing: Pipeline 59713381
I manually set the cuDNN BE in the CI to be 9.25+ (cuDNN FE defaults to 1.26) to test the full support for D=256.  This pipeline uses base image  [ghcr.io/nvidia/jax:jax](http://ghcr.io/nvidia/jax:jax) for JAX and defaults for PyT and Core.
I confirmed that the newly added tests for BSHD and THD (PyT and JAX, both) pass for non-CP and CP

tests/pytorch/attention/test_attention.py
```diff
@@ -376,6 +372,58 @@ def test_dpa_fa4_hdim256(dtype, model_configs, model):
...
+            "thd_t2hd",
+            id="THD_KV_PACKED",
+            marks=pytest.mark.skipif(
+                get_cudnn_version() < (9, 25, 0),
+                reason="cuDNN 9.25+ is required for THD D=256 fused-attn backward.",
+            ),
+        ),
+    ],
+)
+def test_dpa_fused_attn_d256(dtype, model_configs, model, qkv_layout):
```
cyanguwa: I think we can remove the "fused_xx" keyword from model_configs_fused_d256, fused_d256_no_mask, and test_dpa_fused_attn_d256. Those tests will run both unfused/flash and fused for comparison. Maybe there's no need to restrict it to "fused". Essentially what we want to test here is d256.

cyanguwa (dismissed):
Other than the two nits (test name change and 'id' removal), looks good!

transformer_engine/common/fused_attn/fused_attn.cpp
```diff
@@ -327,7 +329,23 @@ NVTE_Fused_Attn_Backend nvte_get_fused_attn_backend(
...
+          // 9.23: d_qk = d_v = 256 + SM10x (cuDNN FE 1.24 / BE 9.23+) + bprop + non-paged.
+          // THD layouts require cuDNN FE 1.26 / BE 9.25+ for execution-plan support.
+          (head_dim_qk == 256 && head_dim_v == 256 && is_training && sm_arch_ >= 100 &&
+           sm_arch_ < 110 && cudnn_runtime_version >= (is_thd_layout ? 92500 : 92300) &&
+           layout_group != NVTE_QKV_Layout_Group::NVTE_Paged_KV_HD_HD_HD &&
+           // The FE forces this path onto the deterministic bprop algorithm, which on
+           // Blackwell rejects dBias, dropout, and ALiBi (and supports vanilla softmax only).
+           bias_type == NVTE_Bias_Type::NVTE_NO_BIAS && dropout == 0.0 &&
+           softmax_type == NVTE_Softmax_Type::NVTE_VANILLA_SOFTMAX &&
+           // Non-causal D=256 supports only full-window attention; SWA is allowed only for causal masks.
```
KshitijLakhani, line 342: @cyanguwa Here's the cuDNN FE restriction in this class for non-causal SWA : https://github.com/NVIDIA/cudnn-frontend/blob/c5012bd3011d28d8ffdf0307b8c213d1b97c68e0/python/cudnn/sdpa/fwd/api.py#L22 and the check is at https://github.com/NVIDIA/cudnn-frontend/blob/c5012bd3011d28d8ffdf0307b8c213d1b97c68e0/python/cudnn/sdpa/fwd/api.py#L201

tests/pytorch/attention/test_attention.py
KshitijLakhani: Done SHA 5ef0d4f29bf7518fa3d71c7e5dfe4f6fb0c7b01c

tests/pytorch/attention/test_attention.py
KshitijLakhani: Done SHA eda1a7431571564181f590790ed4c040308fd7b1

cyanguwa (dismissed):

KshitijLakhani:
**TL; DR:** New pipeline passes all tests fully after latest pushed commits


**Details:** 
I added two commits (top two commits in the branch): f61ef61d7882efbf8d6229cb805ed8b89c7c9e1e and 72958ee09f2ad84ab1104f687ac0957423c60e07 after your approval @cyanguwa as I noticed a couple failures in the L3 FA tests. These commits were pushed to not fix an issue in the functionality added in this PR (i.e. support for the cuDNN backend for D=256) but rather to fix an issue in flash attention 4 support configuration in TE.
The tests and feature support added in this PR were inadvertently triggering Flash attention 4 code paths that aren't supported  (or need some tweaking before being supported) by FA4

_Added two focused FA4 fixes : 
- disable FA4 for SM10x D=256 local/SWA, which FA4 does not support yet, and 
- materialize non-contiguous packed Q/K/V views before FA4 for SM10x D=256 so packed BSHD/THD layouts compare correctly._

 _I also added a change that keeps the pre-existing FA4 hdim256 test forward-only, matching its prior effective coverage (in `main`) before cuDNN fused D=256 backward became available in this PR_

Latest PyT only pipeline launched (as changes were local to PyT only): #60370958
Tests look clean :white_check_mark:

cyanguwa (approved):

PR #3057, opened 2026-05-28
[Common/PyTorch] bugfix: Token-linear fused RoPE impl. for THD tensors.

### LINKED ISSUE #2866: [Performance] Fused RoPE THD kernel becomes dominant bottleneck in long-context training with many packed sequences
## Problem

In long-context training (e.g., `seq_length=65536`) with variable-length sequence packing (THD format), the fused RoPE kernel becomes the **dominant bottleneck** when a micro-batch contains many packed sequences.

Most training samples are shorter than `max_seq_length`, so a single micro-batch routinely packs hundreds to thousands of spans. In this regime, RoPE time grows linearly with `n_seqs` and can exceed the combined cost of attention + MLP.

## Cause

The kernel grid is `dim3(max_seqlen, n_seqs)`, launching `max_seqlen × n_seqs` CUDA blocks. Only `total_tokens` blocks do useful work; the rest read `cu_seqlens` and early-exit. When `n_seqs` is large, the vast majority of blocks are wasted.

```cpp
// fused_rope.cu — forward & backward launchers
dim3 blocks(s, b);   // s = max_seqlen, b = n_seqs
```

```cpp
// Inside kernel — THD path
int t_id = s_id + start;
if (t_id >= end) return;   // most blocks exit here
```

Example: `total_tokens=65536`, `n_seqs=2401` → **157M blocks launched, 65K useful (99.96% wasted)**.

## Profiling Data

0.9B model, 40 layers, H100, TP=2, `seq_length=65536`:

| n_seqs | RoPE / layer (×24) | % of layer time |
|--------|-------------------|-----------------|
| <50 | 22 ms | ~10% |
| 50–200 | 201 ms | ~59% |
| 200–500 | 488 ms | ~79% |
| ~2400 | 4,620 ms | **~97%** |

## Environment

- Transformer Engine 2.6.0.post1
- H100 80GB, CUDA 12.8
- Megatron-LM, THD format, span-based attention

plugyawn:
# Description

Adds a token-linear implementation of the existing THD fused RoPE path to remove a launch-scaling bug. 

Addresses #2866, which finds an interesting case with RoPE scales by freqs_len × n_spans, which is pathological; it should scale by total tokens. I reproduced the issue and found that it's causing a noticeable drops on even plausibly routine shapes. For eg: the [128/512] and [512/128] cases here. 

The new kernel reuses the existing `fused_rope_block_forward` and `fused_rope_block_backward` device helpers, so the math doesn't change. All we need to do is add a THD-only path that launches one bloc/packed token.

| n_seqs | max span | old layer fwd+bwd (ms) | new layer fwd+bwd (ms) | layer speedup | old paired-RoPE share | new paired-RoPE share |
| -----: | -------: | ---------------------: | ---------------------: | ------------: | --------------------: | --------------------: |
|    128 |      512 |                41.8151 |                23.0284 |        1.816x |                49.12% |                 6.14% |
|    512 |      128 |               102.1047 |                23.0167 |        4.436x |                79.38% |                 6.59% |
|   1024 |       64 |               182.9933 |                23.3783 |        7.827x |                88.36% |                 6.77% |
|   2401 |       28 |               401.0516 |                24.5668 |       16.325x |                94.40% |                 6.41% |

This is mostly pathological, however, so I've added a condition on the dispatch to avoid the unnecessary binary search overhead, although the overhead appears to be not-that-relevant. The condition is: token-linear only when `b >= 64` and the old launch would issue ≥ 8× as many blocks as there are tokens. I'm not sure if this the usual shape of TE updates, so I could remove it!

Some more relevant tests:
Microbenchmark on H100 (bf16, `h=32`, `d=d2=128`, `freqs_len=T_local=65536`, single GPU):
| n_seqs | old fwd+bwd (ms) | new fwd+bwd (ms) | speedup |
| -----: | ---------------: | ---------------: | ------: |
|      1 |           1.2746 |           1.2734 |  1.001x |
|      8 |           1.8860 |           1.3827 |  1.364x |
|     32 |           3.9359 |           1.4462 |  2.722x |
|    128 |          12.1849 |           1.5024 |  8.110x |
|    512 |          44.9411 |           1.5600 | 28.808x |
|   1024 |          89.1110 |           1.5919 | 55.977x |
|   2401 |         208.4182 |           1.6373 | 127.296x |

Fixes: #2866.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add token-linear THD fused RoPE forward/backward kernels that launch one CUDA block per packed local token row.
- Add `NVTE_FUSED_ROPE_THD_TOKEN_LINEAR=0|1`.
- Reuses existing `fused_rope_block_forward` and `fused_rope_block_backward` device helpers.

# Checklist:

- [x] I have made corresponding changes to the documentation <<(none?)>>


transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -215,6 +237,98 @@ __global__ void fused_rope_backward_kernel(
...
+// divided cumulative sequence boundaries, then defers to the same
+// `fused_rope_block_forward` device function as the original kernel.
+template <typename scalar_t>
+__global__ void fused_rope_thd_token_forward_kernel(
+    const scalar_t *src, const int *cu_seqlens, const float *freqs, const int *start_positions,
+    scalar_t *dst, const bool interleaved, const int cp_size, const int cp_rank, const int nseq,
+    const int h, const int d, const int d2, const int stride_t, const int stride_h,
+    const int stride_d, const int o_stride_t, const int o_stride_h, const int o_stride_d) {
+  int t_id = blockIdx.x;
+  int b_id = fused_rope_thd_find_seq_id(cu_seqlens, nseq, t_id, cp_size);
```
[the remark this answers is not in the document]
plugyawn: Smart bot!

ptrendx:
@plugyawn Hi, could you sign your commits? See https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst#sign-your-work
Nice improvement :-).

@sudhakarsingh27 Could you take a look?

plugyawn:
Thanks! Signed!

fwiw I think the binary search overhead on normal cases can be reduced also, I'll probably add some improvements.

transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -215,6 +237,98 @@ __global__ void fused_rope_backward_kernel(
...
+  if (total_tokens <= 0) return false;
+
+  const int env_override = transformer_engine::getenv<int>("NVTE_FUSED_ROPE_THD_TOKEN_LINEAR", -1);
+  if (env_override == 0) return false;
+  if (env_override == 1) return true;
+
+  // Heuristic: only worth it when the old launch would issue at least 8x as
+  // many blocks as there are useful tokens, and when there are enough
+  // sequences for binary-search overhead to be amortized.
+  if (b < 64) return false;
```
sudhakarsingh27: Blocking: this misses profitable cases below `b=64`. In our fixed-`T=4096` sweep, `n_seqs=63` stayed on the old path and measured about `0.201/0.399 ms` fwd/bwd for CP1 and `0.331/0.658 ms` for CP2, while forced token-linear was about `0.085/0.165 ms`.

Please use an overlaunch-ratio gate, accounting for CP, instead of an absolute sequence-count gate. Suggested shape from the local patch:

```cpp
constexpr size_t kCompactTHDOverlaunchThreshold = 2;

inline bool fused_rope_thd_use_compact_launch(const size_t legacy_blocks,
                                              const size_t compact_blocks,
                                              const int cp_size) {
  return legacy_blocks >
         kCompactTHDOverlaunchThreshold * static_cast<size_t>(cp_size) * compact_blocks;
}
```

transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -487,6 +600,16 @@ void fused_rope_forward_launcher(const scalar_t *input, const int *cu_seqlens, c
   const int o_stride_h = d;
   const int o_stride_d = 1;
 
+  if (fused_rope_thd_use_token_linear(qkv_format, b, s, total_tokens)) {
```
sudhakarsingh27: Please make the compact launch decision use the actual local THD rows and the legacy launch blocks. The local patch uses this shape:

```cpp
const size_t compact_thd_blocks = input.data.shape[0];
const size_t legacy_thd_blocks = static_cast<size_t>(s) * b;

if (fused_rope_thd_use_compact_launch(legacy_thd_blocks, compact_thd_blocks, cp_size)) {
  const int t = input.data.shape[0];
  dim3 blocks(t);
  ...
}
```

This also avoids routing the heuristic through a `total_tokens` value whose CP/global semantics are easy to confuse.

tests/pytorch/test_fused_rope.py
```diff
@@ -495,3 +496,134 @@ def test_rotary_position_embedding_forward_with_autocast_gives_same_result_as_wi
...
+
+    rotary_pos_emb = RotaryPositionEmbedding(hidden_size, rotary_percent, interleaved=interleaved)
+    # `freqs` must cover (max span length per CP rank + start_positions offset +
+    # CP dual-chunk offset). Use the global cu_seqlens[-1] length as an upper
+    # bound, matching how callers size the freqs tensor in practice.
+    emb = rotary_pos_emb(int(cu_seqlens[-1].item()) + 32)
+
+    def run(force_path: str) -> Tuple[torch.Tensor, torch.Tensor]:
+        monkeypatch.setenv("NVTE_FUSED_ROPE_THD_TOKEN_LINEAR", force_path)
+        cp_rank = 0
```
sudhakarsingh27: Blocking for coverage: this hardcodes `cp_rank=0`, but the new kernel's CP frequency offset depends on `cp_rank`. Please cover all ranks for `cp_size > 1`.

I would fold this into the existing `test_fused_rope_thd`, which already does the right correctness comparison:

```python
for cp_rank in range(cp_size):
    output_unfused = apply_rotary_pos_emb(..., fused=False, cp_rank=cp_rank).to(dtype)
    output_fused = apply_rotary_pos_emb(..., fused=True, cp_rank=cp_rank)
    torch.testing.assert_close(output_fused, output_unfused)
    torch.testing.assert_close(grad_fused, grad_unfused)
```

Keep one small old-vs-new bitwise parity case for the binary-search/zero-length-span edge case, but do not make that the main correctness test.

benchmarks/attention/benchmark_rope_thd_token_linear.py
```diff
@@ -0,0 +1,264 @@
...
+        yield
+    finally:
+        if prev is None:
+            os.environ.pop(name, None)
+        else:
+            os.environ[name] = prev
+
+
+def build_cu_seqlens(total_tokens: int, n_seqs: int, cp_size: int = 1) -> torch.Tensor:
+    """Build a cu_seqlens whose local packed length equals ``total_tokens``.
```
sudhakarsingh27: For `cp_size > 1`, the measured local token count is `actual_total // cp_size`, not `--total-tokens`, despite the docstring. Please make the benchmark explicit: either rename the argument to `--global-total-tokens`, or make `cu[-1] // cp_size == --local-total-tokens`.

Also expose/sweep `cp_rank`; the current benchmark hardcodes `cp_rank=0`.

benchmarks/attention/benchmark_rope_thd_full_layer.py
```diff
@@ -0,0 +1,361 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""Full TransformerLayer benchmark for token-linear THD fused RoPE.
```
sudhakarsingh27: I would drop the full-layer benchmark from this PR. It mixes TransformerLayer, attention backend behavior, and RoPE, so it is not a clean merge artifact for a kernel dispatch heuristic.

Please keep one isolated RoPE kernel benchmark with CSV output. Suggested required columns:

```text
mode, local_total_tokens, global_total_tokens, n_seqs, cp_size, cp_rank,
freqs_len, max_seq, legacy_blocks, compact_blocks, overlaunch_ratio,
fwd_ms, bwd_ms, env_override
```

sudhakarsingh27 (commented):
Posted the RoPE THD token-linear review comments from the local benchmark/coverage analysis. The main concerns are the dispatch heuristic, CP-local token accounting, CP-rank coverage, and benchmark scope.

plugyawn:
Fixed some of the review comments, resolving the rest now.

----

Additional CP-rank validation for the THD token-linear RoPE path on the rebased PR tip:

- Commit: `eaee5a1731141654a006f9872fcfe10132cdcf76` (`Cover CP ranks in THD RoPE token-linear tests`)
- Hardware/runtime: Prime Datacrunch A100 80GB, driver 580.126.09, CUDA 12.8, PyTorch 2.8.0+cu128
- Build: editable TE PyTorch build passed; `fused_rope.cu` and `apply_rope.cpp` compiled on this exact tip
- `test_fused_rope_thd_token_linear_parity`: 288 passed / 96 skipped / 0 failed. The skips are invalid `cp_rank >= cp_size`; the JUnit/log includes 96 passing `cp_rank=1, cp_size=2` cases.
- `test_fused_rope_thd` with the token-linear path forced: 384 passed / 0 failed

This closes the earlier proof gap where old-vs-new parity only covered `cp_rank=0`.

tests/pytorch/test_fused_rope.py
plugyawn: Fixed. Test now covers cp_rank=0/1. Checked on an A100 with cp_rank=1, cp_size=2 parity cases.

benchmarks/attention/benchmark_rope_thd_full_layer.py
plugyawn: Dropped!

transformer_engine/common/fused_rope/fused_rope.cu
plugyawn: Fixed. Also renamed the variable from `total_tokens`, so no CP/global ambiguity. Could you check if it's fine now?

transformer_engine/common/fused_rope/fused_rope.cu
plugyawn: Removed the absolute `b >= 64` gate and switched to an overlaunch-ratio gate with CP factored in: `legacy_blocks > 2 * cp_size * token_linear_blocks`. 
For `local_total_tokens=4096, n_seqs=63`, gate now selects token-linear and matches forced-new timings for both CP1 and CP2.

benchmarks/attention/benchmark_rope_thd_token_linear.py
plugyawn: benchmark now takes `--local-total-tokens`, constructs `cu_seqlens` so `cu[-1] // cp_size == local_total_tokens`, and exposes `--cp-ranks` with default `range(cp_size)`. 

Verified on A100 for `cp_size=2` with both `cp_rank=0` and `cp_rank=1`.

transformer_engine/common/fused_rope/fused_rope.cu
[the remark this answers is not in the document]
plugyawn: Resolved!

transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -215,6 +237,98 @@ __global__ void fused_rope_backward_kernel(
...
+    const scalar_t *src, const int *cu_seqlens, const float *freqs, const int *start_positions,
+    scalar_t *dst, const bool interleaved, const int cp_size, const int cp_rank, const int nseq,
+    const int h, const int d, const int d2, const int stride_t, const int stride_h,
+    const int stride_d, const int o_stride_t, const int o_stride_h, const int o_stride_d) {
+  int t_id = blockIdx.x;
+  int b_id = fused_rope_thd_find_seq_id(cu_seqlens, nseq, t_id, cp_size);
+  int start = cu_seqlens[b_id] / cp_size;
+  int end = cu_seqlens[b_id + 1] / cp_size;
+  int s_id = t_id - start;
+  int cur_seqlens = end - start;
```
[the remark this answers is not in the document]
plugyawn: Added a t_id guard: compute `valid_token = t_id < cu_seqlens[nseq] / cp_size` before lookup, synchronize + return early for invalid token rows.

transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -215,6 +237,119 @@ __global__ void fused_rope_backward_kernel(
...
+  if (token_linear_blocks == 0) return false;
+
+  const int env_override = transformer_engine::getenv<int>("NVTE_FUSED_ROPE_THD_TOKEN_LINEAR", -1);
+  if (env_override == 0) return false;
+  if (env_override == 1) return true;
+
+  // Heuristic: use the token-linear path when the legacy launch would issue
+  // enough extra blocks to amortize one sequence lookup per useful token. The
+  // CP factor keeps the gate conservative because local rows shrink with
+  // context parallelism while legacy launch space does not.
```
sudhakarsingh27: What's the legacy launch space? Does it refer to the one without heuristic or the previous heuristic?

transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -215,6 +237,119 @@ __global__ void fused_rope_backward_kernel(
                             offset_block_dst, h, d, d2, stride_h, stride_d, o_stride_h, o_stride_d);
 }
 
+// Token-linear THD forward kernel. Each block handles exactly one packed local
```
sudhakarsingh27: A bit iffy about calling it token-linear THD. Maybe the right term is THD linear-grid forward kernel? Pls make that change across the file(s)

transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -215,6 +237,119 @@ __global__ void fused_rope_backward_kernel(
...
+// token row. The block locates its owning sequence via binary search over the
+// divided cumulative sequence boundaries, then defers to the same
+// `fused_rope_block_forward` device function as the original kernel.
+template <typename scalar_t>
+__global__ void fused_rope_thd_token_forward_kernel(
```
sudhakarsingh27: similarly, this could be `fused_rope_thd_linear_grid_forward_kernel` and other function below could follow the suit

transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -487,6 +621,18 @@ void fused_rope_forward_launcher(const scalar_t *input, const int *cu_seqlens, c
   const int o_stride_h = d;
   const int o_stride_d = 1;
 
+  const size_t token_linear_blocks = static_cast<size_t>(local_tokens);
+  const size_t legacy_blocks = static_cast<size_t>(s) * static_cast<size_t>(b);
+  if (fused_rope_thd_use_token_linear(qkv_format, legacy_blocks, token_linear_blocks, cp_size)) {
```
sudhakarsingh27: similarly, `use_fused_rope_thd_linear_grid_launch`

transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -579,6 +738,12 @@ void fused_rope_forward(const Tensor &input, const Tensor &cu_seqlens, const Ten
                         const int cp_rank, const int s, const int b, const int h, const int d,
                         const int d2, const int stride_s_or_t, const int stride_b,
                         const int stride_h, const int stride_d, cudaStream_t stream) {
+  // For THD the packed local token count is the first dimension of the input
+  // tensor. SBHD/BSHD ignore this value.
+  const int64_t local_tokens =
```
sudhakarsingh27: Maybe we should call this `total_tokens` or `total_tokens_in_input`. `local_tokens` seems to convey a different unrelated meaning but I understand where you're coming from.

plugyawn:
@ptrendx @sudhakarsingh27 the last review comments are addressed.

sudhakarsingh27:
I'm a bit iffy about adding a benchmark since we aren't actively maintaining benchmarks. @cyanguwa wdyt?

transformer_engine/common/fused_rope/fused_rope.cu
```diff
@@ -215,6 +237,121 @@ __global__ void fused_rope_backward_kernel(
...
+  if (env_override == 0) return false;
+  if (env_override == 1) return true;
+
+  // Heuristic: use the linear-grid path when the original THD grid,
+  // `dim3(s, b)`, would issue enough extra blocks to amortize one sequence
+  // lookup per useful token. The CP factor keeps the gate conservative because
+  // rows in the input shrink with context parallelism while the original
+  // `s * b` launch space does not.
+  return legacy_grid_blocks >
+         kTHDLinearGridOverlaunchThreshold * static_cast<size_t>(cp_size) * linear_grid_blocks;
```
sudhakarsingh27: I benchmarked this PR on my end on hoppers and blackwells with the following updated heuristic and found that for the sweep that I ran, there were no regressions. 

```
current ratio gate AND wasted_blocks >= 65,536
```
Could you verify it on your end too?

---
Once you confirm, I feel we should make the heuristic the default i.e. remove the ENV override as it's not really needed (it makes sense that for larger sequence lengths/workloads, there is more block wastage) and without it the code is much leaner i.e. not needing the newly added test and the associated benchmark - the previous tests would be able to catch the errors if any in the future.

sudhakarsingh27 (commented):
Thanks for working through my comments!

transformer_engine/common/fused_rope/fused_rope.cu
plugyawn: Verifying asap! Sorry, was in terrible internet land for a week or two.

sudhakarsingh27 (approved):
LGTM. Thanks for the PR! 
I'm sure it'll be helpful for large group of people who're training models using TE, good work!

plugyawn:
Thank you for helping merge this! Was very fun, learned a lot!

PR #3058, opened 2026-05-28
[JAX] Grouped quant+GEMM custom partitioning rules

jberchtold-nvidia:
# Description

Adds custom partitioning rules to the grouped quantization and grouped GEMM primitives to support DP/FSDP and EP shardings

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add custom partitioning to grouped quantize and grouped GEMM primitives
- Add tests to validate custom partitioning functions directly, then a real test on a JIT'd program to verify the shardings are correct

# Checklist:


transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -52,6 +60,54 @@
 __all__ = ["quantize", "quantize_dbias", "grouped_quantize", "grouped_dbias"]
 
 
+def _flat_data_spec(input_spec):
+    return (merge_axis_specs(*input_spec),)
```
phu0ngng, line 65: Do we really need a helper for this simple task?

transformer_engine/jax/sharding.py
```diff
@@ -234,6 +234,155 @@ def get_padded_spec(spec, ndim):
...
+    return None if axis_spec == axis else axis_spec
+
+
+def strip_axis_from_spec(spec, axis):
+    """Remove one mesh axis from a PartitionSpec-like tuple."""
+    return tuple(strip_axis_from_axis_spec(axis_spec, axis) for axis_spec in spec)
+
+
+def merge_axis_specs(*axis_specs):
+    """Merge dimension axis specs while preserving first-seen axis order."""
```
phu0ngng, line 307: For `(tupleA, ) + (tupleB,)` ?

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -52,6 +60,54 @@
...
+
+def _normalize_flatten_axis(flatten_axis, ndim):
+    return flatten_axis + ndim if flatten_axis < 0 else flatten_axis
+
+
+def _contiguous_flat_input_spec(input_spec, flatten_axis):
+    flatten_axis = _normalize_flatten_axis(flatten_axis, len(input_spec))
+    if flatten_axis <= 0 or len(input_spec) == 0:
+        return (None,) * len(input_spec)
+    return (*input_spec[:flatten_axis], *((None,) * (len(input_spec) - flatten_axis)))
```
phu0ngng, line 112: Not sure if this is needed.

In every `partition()`, we already have something like this:
```
x_spec = get_padded_spec(arg_infos[0])
```
You can always do `x_spec[:flatten_axis]`.

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -1236,6 +1292,152 @@ def impl(
...
+                group_sizes,
+                out_dtype=out_dtype,
+                scaling_mode=scaling_mode,
+                q_layout=q_layout,
+                flatten_axis=flatten_axis,
+                scale_dtype=scale_dtype,
+            )
+            if ScalingMode(scaling_mode).is_block_scaling:
+                rowwise_scale_inv = _pad_or_slice_to_shape(rowwise_scale_inv, local_out_shapes[2])
+                colwise_scale_inv = _pad_or_slice_to_shape(colwise_scale_inv, local_out_shapes[3])
```
phu0ngng, line 1480: Is it correct that this is needed for the weight only?

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -1236,6 +1292,152 @@ def impl(
         )
         return rowwise_out, colwise_out, rowwise_scale_inv, colwise_scale_inv, updated_amax
 
+    @staticmethod
+    def _parse_partition_specs(scaling_mode, q_layout, flatten_axis, mesh, arg_infos):
```
phu0ngng: This helper is called only once, so I don't think we need it as a separate helper. 

The reason GemmPrimitive has it is that in the past, we needed to generate the partition spec twice (i.e., in `infer_sharding_for_operands` besides `partition`).

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -1236,6 +1292,152 @@ def impl(
...
+        allowed_axes = supported_grouped_partition_axes(mesh)
+        original_x_spec = get_padded_spec(arg_infos[0])
+        x_spec = filter_spec_axes(original_x_spec, allowed_axes)
+        x_spec = _contiguous_flat_input_spec(x_spec, flatten_axis)
+        _warn_if_axes_ignored("x", original_x_spec, x_spec)
```
phu0ngng, line 1371: I would merge `supported_grouped_partition_axes`, `filter_spec_axes`, and `_warn_if_axes_ignored` into a single helper and make it more generic, i.e.
```
def filter_unsupported_partition_axes(input_specs, supported_axes)
```
Then we can define a global `supported_axes` for MoE inputs or local specs case by case. We can still reuse this generic function for other cases.  We should print the warning when we overwrite an axis, not after, to avoid TOCTOU.

The other question here is whether we really should filter the unsupported spec out and replicate that axis, or should we validate and error out. I'm leaning toward erroring out so that users will really know what they are doing.

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -1236,6 +1292,152 @@ def impl(
...
+        x_spec, group_spec, out_specs = GroupedQuantizePrimitive._parse_partition_specs(
+            scaling_mode, q_layout, flatten_axis, mesh, arg_infos
+        )
+        local_out_shapes = (
+            tuple(
+                local_shape_from_spec(info.shape, spec, mesh)
+                for info, spec in zip(result_infos, out_specs)
+            )
+            if result_infos
+            else (None,) * len(out_specs)
```
phu0ngng, line 1447: When will result_infos be None?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1738,6 +1761,320 @@ def impl(
...
+        if (
+            ep_axis is not None
+            and not rhs_is_ragged
+            and spec_contains_axis(active_group_spec, ep_axis)
+        ):
+            if len(rhs_data_spec) > 0 and not spec_contains_axis(rhs_data_spec, ep_axis):
+                rhs_data_spec = (
+                    merge_axis_specs(rhs_data_spec[0], ep_axis),
+                    *rhs_data_spec[1:],
+                )
```
phu0ngng, line 1857: I'm not sure if we should insert/force the rhs to be partitioned by `ep`. I'm in favor of validating if the inputs are partitioned correctly, then continue.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1738,6 +1761,320 @@ def impl(
...
+            if len(rhs_scale_spec) > 0 and not spec_contains_axis(rhs_scale_spec, ep_axis):
+                rhs_scale_spec = (
+                    merge_axis_specs(rhs_scale_spec[0], ep_axis),
+                    *rhs_scale_spec[1:],
+                )
+            if len(bias_spec) > 0 and not spec_contains_axis(bias_spec, ep_axis):
+                bias_spec = (merge_axis_specs(bias_spec[0], ep_axis), *bias_spec[1:])
```
phu0ngng, line 1864: Same, bias should be expected to be partitioned as in the RHS non-contracting specs.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1738,6 +1761,320 @@ def impl(
...
+
+        gather_rhs_fsdp = (
+            fsdp_axis is not None
+            and not rhs_is_ragged
+            and (
+                spec_contains_axis(rhs_data_spec, fsdp_axis)
+                or spec_contains_axis(rhs_scale_spec, fsdp_axis)
+                or spec_contains_axis(bias_spec, fsdp_axis)
```
phu0ngng, line 1892: I think the scale and the bias should be validated to ensure that it has the same/the right partition wrt to rhs_data, then here we should check for one instead of OR.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1738,6 +1761,320 @@ def impl(
...
+
+        if gather_rhs_fsdp:
+            rhs_data_spec = strip_axis_from_spec(rhs_data_spec, fsdp_axis)
+            rhs_scale_spec = strip_axis_from_spec(rhs_scale_spec, fsdp_axis)
+            bias_spec = strip_axis_from_spec(bias_spec, fsdp_axis)
+
+        reducible_axes = tuple(
+            axis for axis in (gsr.dp_resource, gsr.fsdp_resource) if axis is not None
+        )
+        reduce_axis = common_spec_axis(lhs_data_spec, rhs_data_spec, reducible_axes)
```
phu0ngng, line 1904: Here, we should check if there is a common axis in the contracting dims, not in all dims.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1738,6 +1761,320 @@ def impl(
...
+        if reduce_axis is not None and gather_rhs_fsdp:
+            reduce_axis = None
```
phu0ngng, line 1911: I think we can add a comment here that when `father_rhs_fsdp = True`, it's not the dW GGEMM, thus, we should not expect an AllReduce or ReduceScatter. But in this case, we should error out if `reduce_axis` is not None.

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1738,6 +1761,320 @@ def impl(
...
+                if lhs_is_trans
+                else range(0, lhs_axis_boundary)
+            )
+            lhs_data_spec = list(lhs_data_spec)
+            for out_idx, lhs_dim in enumerate(lhs_non_contracting_dims, start=1):
+                if out_idx < len(out_spec):
+                    lhs_data_spec[lhs_dim] = merge_axis_specs(
+                        lhs_data_spec[lhs_dim], out_spec[out_idx]
+                    )
+            lhs_data_spec = tuple(lhs_data_spec)
```
phu0ngng, line 1932: Not sure about what we are doing here for the dW GEMM. 
At least, should we have a validation on what is the input specs we are expecting here?

transformer_engine/jax/cpp_extensions/gemm.py
```diff
@@ -1738,6 +1761,320 @@ def impl(
...
+            arg_infos,
+            result_infos,
+            out_shape,
+            lhs_is_trans=lhs_is_trans,
+            lhs_axis_boundary=lhs_axis_boundary,
+        )
+        arg_shardings = tuple(NamedSharding(mesh, PartitionSpec(*spec)) for spec in arg_specs)
+        out_sharding = (NamedSharding(mesh, PartitionSpec(*out_spec)),)
+        local_out_shape = local_shape_from_spec(out_shape, out_spec, mesh)
+        local_lhs_left_size, local_lhs_right_size = local_2d_sizes_from_spec(
```
phu0ngng, line 2009: I could not follow why we need 2D sizes here.

transformer_engine/jax/cpp_extensions/quantization.py
```diff
@@ -52,6 +60,54 @@
...
+        axis for axis in spec_axes(original_spec) if axis not in spec_axes(partition_spec)
+    )
+    if ignored_axes:
+        warnings.warn(
+            "Grouped quantize custom partitioning will ignore/replicate sharding "
+            f"axes {ignored_axes} from {arg_name}; only supported packed grouped "
+            "data axes are preserved.",
+            RuntimeWarning,
+            stacklevel=3,
+        )
```
tdophung, line 126: This looks exactly like the one in gemm.py, maybe we define it somewhere common and reuse for both

tests/jax/test_distributed_grouped_gemm.py
```diff
@@ -0,0 +1,503 @@
...
+    assert tuple(arg_shardings[5].spec) == ("expert",)
+    assert tuple(out_sharding[0].spec) == ("expert", None, None)
+    for spec in (
+        *(tuple(sharding.spec) for sharding in arg_shardings),
+        tuple(out_sharding[0].spec),
+    ):
+        assert not _spec_contains_axis(spec, "tp")
+
+
+def test_grouped_gemm_reduce_axis_skips_ep_and_uses_dp():
```
tdophung, line 352: there is no full test (execute a JIT + VJP run) that uses "dp" axis. The only full test `test_grouped_dense_mxfp8_ep_fsdp_outside_shard_map_single_process` I see does not use the `dp` axis. And this test on this line only check if reduce axis is "dp"

tests/jax/test_distributed_grouped_gemm.py
```diff
@@ -0,0 +1,503 @@
...
+        _, _, reduce_axis = GroupedGemmPrimitive._parse_partition_specs(
+            mesh,
+            arg_infos,
+            (),
+            out_shape=(1, 128, 64),
+            lhs_is_trans=False,
+            lhs_axis_boundary=1,
+        )
+
+    assert reduce_axis == "dp"
```
tdophung: This was really verbose to read. I wonder if you can refactor it into:
1. a wrapper over the partition() call that takes in (mesh, input spec, layout, scaling) and return plain spec tuples that can apply for all the different cases here
2. parametrize with `pytest.mark.parametrize`
3. a common helper function to build the big arg_infos tuple for each case
4. rename test cases

tests/jax/test_distributed_grouped_gemm.py
```diff
@@ -0,0 +1,503 @@
...
+
```
tdophung, line 383: Also, I notice that the tests here focuses on testing the subtlety of the axis stripping/merging logic, which diverges from the usual pattern of other `tests/jax/test_distributed_*.py`, where we mostly have a pure jax implementation results as ground truth to compare with the defined primitive's result. Is this inconsistency intentional as you implement way more axis handling logic in the `_parse_partition_specs` in `partition` than in other primitives?

transformer_engine/jax/cpp_extensions/gemm.py
tdophung, line 1546: according to this, the addittional_arg_0 and additional_arg_1 used in the partition func definition below are alpha and beta here, should we name it so, so it is consistent and less confusing? it also make it hard to infer why these seem to be per-group array with sharding spec inherited from active_group_spec

transformer_engine/jax/cpp_extensions/gemm.py
jberchtold-nvidia: This was to attempt to WAR a Shardy bug that limited the num elements on a single data axis in the tensor's global shape to INT32_MAX. There is now a PR to fix this here so I will go back to 1D
https://github.com/openxla/shardy/pull/1468

PR #3060, opened 2026-05-29
increasing precision tolerance

francesco-bertolotti:
Here’s a cleaner Markdown rewrite of your PR description:

---

## Description

`test_kv_cache` compares full-sequence attention against incremental KV-cache decoding. In the `TransformerLayer` configuration where `head_dim > 128` in `fp16`, these two execution paths use different kernels and masking strategies (e.g., `causal` vs. `padding_causal_bottom_right`, and full-matrix vs. single-query-row kernels). As a result, their outputs diverge slightly due to accumulated `fp16` rounding differences.

On Ampere, this divergence can reach the current tolerance threshold in rare cases, producing a spurious failure. In one observed instance, a single element out of 4096 showed an absolute difference of ~0.0107, which narrowly exceeds the existing `1e-2` tolerance.

This change slightly relaxes the `fp16` tolerance for the affected configuration to make the test robust across architectures. No runtime or library code is modified.

**Fixes:** N/A (spurious `test_kv_cache` failure on sm80 / fp16 / head_dim=256)

---

## Type of change

* [ ] Documentation change
* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature
* [ ] Breaking change
* [ ] Infra/Build change
* [ ] Code refactoring

---

## Checklist

* [ ] I have read and followed the contributing guidelines ([https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst))
* [x] The functionality is complete
* [x] I have commented my code, particularly in hard-to-understand areas
* [ ] I have made corresponding changes to the documentation (N/A — test-only change)
* [x] My changes generate no new warnings
* [ ] I have added tests that prove my fix is effective or that my feature works (N/A — test-only tolerance adjustment)
* [ ] New and existing unit tests pass locally with my changes

---

## Notes

* Contributing guidelines and full test runs still need to be verified before final submission.
* `bfloat16` tolerance in this branch is unchanged, as it was not failing.
* The Pyright import warnings are unrelated noise from missing stubs in the local environment.
* Potential reviewer feedback: this could be further refined to apply the tolerance only conditionally by compute capability (e.g., sm < 90), rather than broadly for the branch.

---

## Diff summary

```diff
# tests/pytorch/attention/test_kv_cache.py (get_tols)
- torch.half: (1e-2, 1e-2),
+ torch.half: (1.5e-2, 1.5e-2),

# plus explanatory comment added
```

---


cyanguwa:
Hi @francesco-bertolotti, thanks for the contribution. The rational behind the PR makes sense, and indeed we have seen sporadic single-element mismatch failures for `test_kv_cache.py`, but I think as of last night, the nightly CI is passing with the existing tolerances, on all architectures (Ampere, Hopper and Blackwell). Were you observing some failures? What's the architecture and cuDNN version you're running with? Thanks!

francesco-bertolotti:
> Were you observing some failures?

At the moment, I'm mainly interested in the ring attention kernel, which I plan to integrate into a fork of TorchTitan. However, since I need to build Transformer Engine to get it working properly on the Leonardo HPC system, I'd like to first reach a state where the full test suite passes.

> What's the architecture and cuDNN version you're running with?

I'm currently building against the cuDNN version that ships with the `torch+cu126` wheels, which is cuDNN 9.10.2.

One thing I haven't tried yet is building against a newer cuDNN release. My concern is whether that could lead to issues if PyTorch is using cuDNN 9.10 while Transformer Engine is built against a different version.

timmoon10:
> test_kv_cache compares full-sequence attention against incremental KV-cache decoding. In the TransformerLayer configuration where head_dim > 128 in fp16, these two execution paths use different kernels and masking strategies (e.g., causal vs. padding_causal_bottom_right, and full-matrix vs. single-query-row kernels). As a result, their outputs diverge slightly due to accumulated fp16 rounding differences.

Wait, if it's true that the masking is different, why do we expect the full-sequence and KV-cache paths to match? Is it just a representation difference, or are they computing something actually different? If they are computing something different, then that's the true bug. Increasing tols to make two unlike things match is the same as not testing numerics at all.

ptrendx (approved):

ptrendx:
@timmoon10 It is expected that the padding masks will be different - the full non-cached execution uses the full triangle (so the causal mask) and the cached execution works on 1 row at a time, so bottom right causal padding is correct (It could maybe even be None, depending on the shapes provided to the kernel).

cyanguwa:
> One thing I haven't tried yet is building against a newer cuDNN release. My concern is whether that could lead to issues if PyTorch is using cuDNN 9.10 while Transformer Engine is built against a different version.

I think that's ok - you can build TE against a different version of cuDNN than PyTorch. If it's in the system path, TE should be able to pick it up automatically; if not, you can specify the path with CUDNN_PATH.

With 9.10.2, I believe we did test it with TE 2.5 and the CI was clean too, but now we're on TE 2.16 and cuDNN 9.23, and I don't think we've gone back to those older versions for testing. Could you please try 9.23 and see if there's any tolerance issues? I don't have a problem with the changes in this PR, but just wanted to make sure we're not making any unnecessary changes :). Thanks.

francesco-bertolotti:
> Could you please try 9.23 and see if there's any tolerance issues?

Just tried, I am still getting the same issue unfortunately. I am pasting a bit from the debug logs. It may be helpful. The test in question is `test_kv_cache[False-False-TransformerLayer-FusedAttention-False-bshd-infer_1-dtype0]`. It is run with: 

DotProductAttention
```python
{
   "transformer_engine_version":"2.17.0.dev0+80ea3133",
   "compute_capability":"sm80",
   "cuda_version":"12.6",
   "flash_attn_version":"2.8.3",
   "flash_attn_3_version":"not installed",
   "flash_attn_4_version":"not installed",
   "cudnn_version":"9.23.0",
   "qkv_type":"<class""torch.Tensor"">",
   "qkv_dtype":torch.float16,
   "qkv_layout":"bshd_bshd_bshd",
   "batch_size":2,
   "num_heads":16,
   "num_gqa_groups":4,
   "max_seqlen_q":66,
   "max_seqlen_kv":128,
   "head_dim_qk":256,
   "head_dim_v":256,
   "attn_mask_type":"no_mask",
   "window_size":(-1,
   -1),
   "bottom_right_diagonal":true,
   "alibi_slopes_shape":"None",
   "core_attention_bias_type":"no_bias",
   "core_attention_bias_shape":"None",
   "core_attention_bias_requires_grad":false,
   "pad_between_seqs":false,
   "attention_dropout":0.0,
   "context_parallel":false,
   "cp_comm_type":"p2p",
   "cp_size":1,
   "deterministic":false,
   "is_training":false,
   "fp8":false,
   "fp8_meta":{
      "recipe":"recipe_type=DelayedScaling",
      margin=0,
      "format=HYBRID",
      amax_history_len=1,
      "reduce_amax=True",
      fp8_dpa=False,
      fp8_mha=False,
      "backward_override=None"
   },
   "inference_params":dtype=torch.float16,
   "is_paged=False",
   max_batch_size=2,
   max_seqlen=128,
   num_heads=4,
   head_dim_k=256,
   head_dim_v=256,
   "softmax_type":"vanilla",
   "return_max_logit":false,
   "cuda_graph":false,
   "num_splits":1
}
```

I am still getting a single 1 element mismatch:
```bash
Mismatched elements: 1 / 4096 (0.0%)
Greatest absolute difference: 0.0107421875 at index (3169,) (up to 0.01 allowed)
Greatest relative difference: 0.1617431640625 at index (3169,) (up to 0.01 allowed)
```

cyanguwa:
> Just tried, I am still getting the same issue unfortunately. I am pasting a bit from the debug logs. It may be helpful. The test in question is `test_kv_cache[False-False-TransformerLayer-FusedAttention-False-bshd-infer_1-dtype0]`.

No worries, thanks for trying it out. I can see there's a bit of difference in the CUDA version you used and in our CI, but I feel this problem (the exact mismatch values) has shown up in our CI sporadically too (if not on A100, on L40). I'm going to merge the PR, thanks for making the changes!

PR #3063, opened 2026-05-31
fix unfused padding causal sdpa

hungryGeek16:
Adds a targeted PyTorch SDPA fallback for unfused THD padding_causal self-attention so TransformerEngine does not materialize the full quadratic padding/causal mask. Includes a regression test that fails if get_full_mask is called on this path.

### DIFF
--- a/transformer_engine/pytorch/attention/dot_product_attention/backends.py
+++ b/transformer_engine/pytorch/attention/dot_product_attention/backends.py
@@ -365,6 +365,115 @@ class UnfusedDotProductAttention(torch.nn.Module):
         """Fast attribute set for non-parameter fields."""
         self.__dict__[name] = value
 
+    def _use_varlen_sdpa(
+        self,
+        attn_mask_type: str,
+        attention_mask: Optional[Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]]],
+        window_size: Optional[Tuple[int, int]],
+        max_seqlen_q: int,
+        max_seqlen_kv: int,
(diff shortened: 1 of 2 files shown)


cyanguwa:
Thanks for the contribution @hungryGeek16, but it looks like your base branch may be out of date - could you rebase please? Thanks!

hungryGeek16:
@cyanguwa , I have rebased and resolved conflicts, let me know if this works. Thanks!

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -365,6 +365,115 @@ def fast_setattr(self, name: str, value: Any) -> None:
...
+        if attention_mask is not None:
+            seqlens_q = attention_mask.logical_not()[:, 0, 0, :].sum(dim=1)
+        else:
+            seqlens_q = torch.full(
+                (batch_size,), max_seqlen_q, dtype=torch.int64, device=query_layer.device
+            )
+
+        dropout_p = self.attention_dropout.p if self.training else 0.0
+        with self.attention_dropout_ctx():
+            for batch_id in range(batch_size):
```
cyanguwa, line 453: This seems to be serializing the sequences in a batch. Is this really more efficient than the original implementation in UnfusedDotProductAttention?

Also, is there any particular reason to use Torch SDPA in TE's unfused path, when THD padding_causal self-attention is supported by FusedAttention and FlashAttention? Thanks.

PR #3065, opened 2026-05-31
[PyTorch] Propagate skip_fp8_weight_update in GroupedLinear during FP8 CUDA graph capture

LeSingh1:
## Description

`GroupedLinear.forward` hardcoded `None` for `skip_fp8_weight_update`, so the FP8 graph-capture skip tensor was never forwarded during CUDA graph replay — the bug reported in #3051. `Linear.forward` handles this correctly: when `FP8GlobalStateManager.fp8_graph_capturing()` is true it reads `quantization_state.skip_fp8_weight_update_tensor`, forces `is_first_microbatch = False`, and threads the tensor into its forward args. `GroupedLinear` omitted that retrieval block and passed a literal `None`, even though the autograd `_GroupedLinear.forward` already unpacks and uses the flag (`quantize_weight(..., skip_update_flag=skip_fp8_weight_update)`).

## Fix

Mirror the `Linear` reference path in `GroupedLinear.forward`: add the same `fp8_graph_capturing()` retrieval block and thread `skip_fp8_weight_update` into `non_tensor_args` (positionally aligned with the existing autograd unpack). `FP8GlobalStateManager` is already imported in the module.

## Verification status

> **Runtime-unverified — developed without a CUDA GPU, so the FP8 / CUDA-graph numerics and the Nemotron-MoE gradient-suppression reproducer were not run.** The change is a near-mechanical parity fix against the established `Linear` path (same accessor, same `is_first_microbatch` forcing, same threading), and the positional unpack lines up with `_GroupedLinear.forward`. A GPU regression test (comparing GroupedLinear FP8 grads under graph-replay-active vs delayed) belongs alongside the existing CUDA-graph tests but can't run on this platform — I'd ask a maintainer with a GPU to confirm numerics before merge, and I'm happy to add the test.
>
> Developed with AI assistance.

Addresses #3051


ptrendx:
@LeSingh1 Could you add the test that would exercise this functionality?

LeSingh1:
@ptrendx Added a test in `tests/pytorch/test_cuda_graphs.py`: `test_make_graphed_callables_grouped_linear_with_fp8_weight_caching`.

It wraps `GroupedLinear` (flattening the `[seqlen, batch, hidden]` harness input to 2D and splitting evenly across 2 GEMMs) so it slots into the existing `_test_cuda_graphs` harness, then runs the `full` / `individual` / `none` graph modes with `fp8_weight_caching=True` and asserts the graphed outputs equal the eager reference. That equality only holds when `is_first_microbatch` is threaded into the weight-update skip tensor on every microbatch — on the pre-fix `None` hardcode the cached FP8 weights diverge, so the test fails without the change. Parametrized over DelayedScaling + Float8CurrentScaling, both `fp8_params` settings, and the fp32/fp16 dtypes already used in this file.

One disclosure: I don't have FP8-capable GPU hardware locally, so I verified the test by construction (mirroring `test_make_graphed_callables_with_fp8_weight_caching`) rather than by running it — it'll need a CI/GPU run to confirm. Happy to adjust the recipe coverage or wrapper approach if you'd prefer it structured differently.

LeSingh1:
Rebased onto main to resolve a conflict with #3038 (the GroupedLinear graph-safe refactor). Worth noting: #3038 plumbs skip_fp8_weight_update all the way through _GroupedLinear's autograd function, but the module-level GroupedLinear.forward still passes `None,  # skip_fp8_weight_update` into non_tensor_args, so the value never reaches that plumbing during graph capture. This change wires it up the same way Linear.forward does (read the skip tensor from FP8GlobalStateManager.quantization_state when fp8_graph_capturing()). The test from the previous commit comes along with the rebase. Still GPU/CI-only on my end since I don't have FP8 hardware locally.

ksivaman (approved):

PR #3070, opened 2026-06-01
[PyTorch] Use data shape for MXFP8 pointer swizzle

jepio:
# Description
We've started seeing NaN loss in gpt_oss_20b training using MXFP8. gpt_oss_20b uses expert MLP dimensions based on hidden size 2880, producing FC1 weights of (5760, 2880) and FC2 weights of (2880, 2880). I bisected this to NVIDIA/TransformerEngine@9e5a847c90c436c9c9c5f3e756977cf568967ccc. That change is not safe in the presence of MXFP8 scale padding: without the actual unpadded data shape, `transform_and_copy_data_ptrs_to_device` infers the data shape from the padded scale shape, so padded E8M0 scale bytes can be copied into the swizzled GEMM scale buffer and consumed by GEMM, causing NaNs. Fix it by passing the actual data shape to transform_and_copy_data_ptrs_to_device. The patch includes a regression test that poisons padded scale entries and verifies they are masked out.

Fixes #

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Fix MXFP8 scale padding being read for non-128-aligned dimensions

# Checklist:

### DIFF
--- a/tests/pytorch/mxfp8/test_mxfp8_quantize_swizzle_fusion.py
+++ b/tests/pytorch/mxfp8/test_mxfp8_quantize_swizzle_fusion.py
@@ -20,6 +20,30 @@ from mxfp8_utils import swizzle_mxfp8_scale, get_mxfp8_scale_shape_no_padding
 recipe_available, reason_for_no_recipe = te.is_mxfp8_available(return_reason=True)
 
 
+def poison_mxfp8_scale_padding_with_nan(
+    scale: torch.Tensor,
+    valid_shape: Tuple[int, int],
+) -> torch.Tensor:
+    """Fill only the padded part of a compact MXFP8 scale tensor with NaNs."""
+    scale = scale.clone()
+    scale_e8m0 = scale.view(dtype=torch.float8_e8m0fnu)
+    scale_e8m0[valid_shape[0] :, :] = float("nan")
+    scale_e8m0[:, valid_shape[1] :] = float("nan")
+    return scale
+
+
+def zero_mxfp8_scale_padding(
+    scale: torch.Tensor,
+    valid_shape: Tuple[int, int],
+) -> torch.Tensor:
+    """Zero only the padded part of a compact MXFP8 scale tensor."""
+    scale = scale.clone()
+    scale_uint8 = scale.view(dtype=torch.uint8)
+    scale_uint8[valid_shape[0] :, :] = 0
+    scale_uint8[:, valid_shape[1] :] = 0
+    return scale
+
+
 def unpack_quantized_tensor(
     quantized_tensor: MXFP8TensorStorage,
 ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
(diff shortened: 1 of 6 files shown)


tests/pytorch/mxfp8/test_mxfp8_quantize_swizzle_fusion.py
```diff
@@ -130,3 +154,57 @@ def test_mxfp8_quantize_swizzle_fusion(
...
+        inferred_padded_shape = (scale.shape[0] * 32, scale.shape[1])
+    else:
+        scale = quantized._rowwise_scale_inv
+        transform_type = "uniform_mxfp8_rowwise_swizzle"
+        inferred_padded_shape = (scale.shape[0], scale.shape[1] * 32)
+
+    # Poison padded scale values with E8M0 NaNs. If swizzle reconstructs the
+    # data shape from the padded scale shape, these values are considered real
+    # and survive.
+    scale = poison_mxfp8_scale_padding_with_nan(scale, valid_scale_shape)
```
vthumbe1503, line 187: Why is the scales getting padded with poision values in the first place?

jepio:
I think that part of the scales is normally uninitialized, I poison the
buffer in the test to illustrate what would happen if the uninitialized
values are actually read and included in the computation. This is what
seems to happen in actual training.

tests/pytorch/mxfp8/test_mxfp8_quantize_swizzle_fusion.py
jepio: @vthumbe1503 I think that part of the scales is normally uninitialized, I poison the
buffer in the test to illustrate what would happen if the uninitialized
values are actually read and included in the computation. This is what
seems to happen in actual training.

transformer_engine/pytorch/csrc/extensions/utils.cpp
```diff
@@ -40,7 +40,7 @@ at::Tensor copy_data_ptrs_to_device(const std::vector<at::Tensor> &tensors,
 
 std::tuple<at::Tensor, std::optional<at::Tensor>> transform_and_copy_data_ptrs_to_device(
     const std::string &transform_type, const std::vector<at::Tensor> &tensors,
-    const c10::Device &device) {
+    const c10::Device &device, const std::vector<int64_t> &actual_data_shape) {
```
timmoon10, line 43: `transform_and_copy_data_ptrs_to_device` is named the way it is so that it can be a general-purpose function. You could imagine, for instance, using it for things like SReLU or dtype casts. In practice, though, we only use it for swizzling uniformly-sized MXFP8 and NVFP4 scales.

Hard-coding swizzle-specific logic like `actual_data_shape` into a general API is not great. But your bug report makes me realize that the general-purpose function is probably not possible, so we should give in and make this a swizzle-specific function. The biggest problem is not the implementation (easy), but that the function name will have to be something ugly like `uniform_swizzle_scales_and_copy_data_ptrs_to_device`.

timmoon10 (commented):
Thanks for the bugfix. One of the goals of https://github.com/NVIDIA/TransformerEngine/pull/3001 was to generalize the helper functions and avoid hard-coding swizzle-specific logic in places where it wasn't needed. However, it seems that this is probably unavoidable.

I'll take a crack at fixing up the function interface. While we're at it, this gives us the opportunity to consolidate the copy-to-device kernel launches for the data pointers and scale pointers.

ptrendx:
I have some problem understanding where the problem comes from here. Where is the padding of the weights happening? It seems that the padding happens only after the weight was already cast to MXFP8 and the weight and the scale are padded separately? Why are we not padding with 0s - that padding should happen only once?

timmoon10:
https://github.com/NVIDIA/TransformerEngine/pull/3076 fixes this bug.

PR #3073, opened 2026-06-01
Enable NVFP4 grouped MLP GLU RHT amax path

sraman-rgb:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/csrc/extensions/cast.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/cast.cpp
@@ -92,9 +92,9 @@ py::object nvfp4_quantize_with_amax(const at::Tensor &tensor, py::handle quantiz
   NVTE_CHECK(tensor.dim() >= 2, "Tensor must be at least 2D");
   NVTE_CHECK(rowwise_amax.is_cuda() && columnwise_amax.is_cuda(),
              "Precomputed amax tensors must be CUDA tensors.");
-  NVTE_CHECK(rowwise_amax.scalar_type() == at::kFloat &&
-                 columnwise_amax.scalar_type() == at::kFloat,
-             "Precomputed amax tensors must be float32.");
+  NVTE_CHECK(
+      rowwise_amax.scalar_type() == at::kFloat && columnwise_amax.scalar_type() == at::kFloat,
+      "Precomputed amax tensors must be float32.");
   NVTE_CHECK(rowwise_amax.numel() == 1 && columnwise_amax.numel() == 1,
              "nvfp4_quantize_with_amax expects scalar rowwise and columnwise amaxes.");
 
@@ -294,9 +294,9 @@ py::object nvfp4_group_quantize_with_amax(const at::Tensor &tensor, py::handle q
   NVTE_CHECK(tensor.dim() == 2, "Tensor must be 2D");
   NVTE_CHECK(rowwise_amax.is_cuda() && columnwise_amax.is_cuda(),
              "Precomputed amax tensors must be CUDA tensors.");
-  NVTE_CHECK(rowwise_amax.scalar_type() == at::kFloat &&
-                 columnwise_amax.scalar_type() == at::kFloat,
-             "Precomputed amax tensors must be float32.");
+  NVTE_CHECK(
+      rowwise_amax.scalar_type() == at::kFloat && columnwise_amax.scalar_type() == at::kFloat,
+      "Precomputed amax tensors must be float32.");
   NVTE_CHECK(rowwise_amax.numel() == static_cast<int64_t>(num_tensors),
              "Rowwise amax must contain one value per group.");
   NVTE_CHECK(columnwise_amax.numel() == static_cast<int64_t>(num_tensors),
--- a/transformer_engine/pytorch/csrc/extensions/cast.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/cast.cpp
@@ -198,9 +198,9 @@ void group_quantize_nvfp4_impl(const GroupedTensorWrapper &grouped_input_tensor,
 
   auto rht_matrix_nvte = makeTransformerEngineTensor(nvfp4_quantizer_cpp->rht_matrix);
   NVTE_SCOPED_GIL_RELEASE({
-      nvte_group_hadamard_transform_cast_fusion_graph_safe(
-          grouped_input_tensor.data(), grouped_output_tensor.data(), rht_matrix_nvte.data(),
-          quant_config_cpp, nvte_tile_scheduler_workspace.data(), stream);
+    nvte_group_hadamard_transform_cast_fusion_graph_safe(
+        grouped_input_tensor.data(), grouped_output_tensor.data(), rht_matrix_nvte.data(),
+        quant_config_cpp, nvte_tile_scheduler_workspace.data(), stream);
   });
 }
 
--- a/transformer_engine/pytorch/csrc/extensions/cast.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/cast.cpp
@@ -46,7 +46,7 @@ void allreduce_nvfp4_amax_tensors(NVFP4Quantizer *nvfp4_quantizer_cpp,
 }  // namespace
 
 py::object quantize(const at::Tensor &tensor, py::handle quantizer, const py::object &output,
-                    std::optional<at::Tensor> noop_flag, bool compute_amax) {
+                    std::optional<at::Tensor> noop_flag) {
   // Convert quantizer to C++ object
   auto quantizer_cpp = convert_quantizer(quantizer);
 
--- a/transformer_engine/pytorch/csrc/extensions/cast.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/cast.cpp
@@ -31,6 +31,18 @@ std::vector<size_t> get_tensor_shape(const TensorWrapper &tensor) {
   return std::vector<size_t>(shape.data, shape.data + shape.ndim);
 }
 
+void allreduce_nvfp4_amax_tensors(NVFP4Quantizer *nvfp4_quantizer_cpp,
+                                  std::vector<at::Tensor> &&amax_tensors) {
+  if (!nvfp4_quantizer_cpp->with_amax_reduction || amax_tensors.empty()) {
+    return;
+  }
+  c10d::AllreduceCoalescedOptions opts;
+  opts.reduceOp = c10d::ReduceOp::MAX;
+  NVTE_SCOPED_GIL_RELEASE({
+    nvfp4_quantizer_cpp->amax_reduction_group->allreduce_coalesced(amax_tensors, opts)->wait();
+  });
+}
+
 }  // namespace
 
 py::object quantize(const at::Tensor &tensor, py::handle quantizer, const py::object &output,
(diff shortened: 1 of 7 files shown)


transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -65,6 +65,111 @@ py::object quantize(const at::Tensor &tensor, py::handle quantizer, const py::ob
   return output_py;
 }
 
+py::object quantize_with_amax(const at::Tensor &tensor, py::handle quantizer,
+                              const at::Tensor &rowwise_amax, const at::Tensor &columnwise_amax) {
+  using namespace transformer_engine::pytorch::detail;
+  init_extension();
+
```
vthumbe1503, line 91: https://github.com/NVIDIA/TransformerEngine/blob/1609c890ee7da24393ce62a2680e0ebb0b4eb5f6/transformer_engine/pytorch/csrc/quantizer.cpp#L2300

Most the logic is already part of NVFP4 Quantize implementation
Although it seems like compute_amax is not properly hooked up in RHT impl. After correcting that we should be able to reuse that function

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -149,6 +254,56 @@ void group_quantize_nvfp4_impl(const GroupedTensorWrapper &grouped_input_tensor,
   }
 }
 
+void group_quantize_nvfp4_impl_with_amax(const GroupedTensorWrapper &grouped_input_tensor,
```
vthumbe1503: Most of the implementation is exactly the same as
` group_quantize_nvfp4_impl`
How about we pass a flag compute_amax to the API as well and reuse the logic

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -395,25 +396,49 @@ def fuser_forward(
...
+        )
+        enable_fc1_glu_hadamard = (
+            use_nvfp4 and self._cudnn_act_func == "swiglu" and enable_fc1_glu_hadamard_env
+        )
+        fc1_glu_hadamard_kernel = None
+        if enable_fc1_glu_hadamard:
+            kernel_getter = getattr(self, "grouped_gemm_glu_hadamard_kernel", None)
+            if kernel_getter is not None:
+                fc1_glu_hadamard_kernel = kernel_getter()
+        has_precomputed_amax_quantize = (
```
vthumbe1503: No need for this boolean. If TE has all the above code, this API would be present as part of lib

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -395,25 +396,49 @@ def fuser_forward(
         else:
             fc1_alpha_tensor = alpha_tensor
 
+        enable_fc1_glu_hadamard_env = (
```
vthumbe1503: We can simply decide based on whether the kernel is available in cudnn.

If this is performant, we always want this to be true right?

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -395,25 +396,49 @@ def fuser_forward(
...
+            os.environ.get("NVTE_CUTEDSL_FUSED_GROUPED_MLP_FC1_GLU_RHT_AMAX", "0") == "1"
+        )
+        use_tmem_post_rht_amax = (
+            os.environ.get("NVTE_CUTEDSL_FUSED_GROUPED_MLP_FC1_GLU_RHT_AMAX_TMEM", "0") == "1"
+        )
```
vthumbe1503: I guess we should enable this through something like fusion args being passed to TE Sequential rather env variable. @timmoon10 thoughts?

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
```diff
@@ -465,7 +490,10 @@ def fuser_forward(
             fc1_activation_kwargs["b_dtype"] = data_dtype
             fc1_activation_kwargs["b_major"] = "k"
 
-        fc1_kernel_out = self.grouped_gemm_activation_kernel()(**fc1_activation_kwargs)
+        if use_fc1_glu_hadamard:
+            fc1_kernel_out = fc1_glu_hadamard_kernel(**fc1_activation_kwargs)
```
vthumbe1503: Lets just compute boolean above and call
`self.grouped_gemm_glu_hadamard_kernel()(**fc1_activation_kwargs)`
for consistency

vthumbe1503 (commented):
Mostly LGTM. Left a few comments on code duplication and other minor issues.

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -31,10 +31,22 @@ std::vector<size_t> get_tensor_shape(const TensorWrapper &tensor) {
...
+  NVTE_SCOPED_GIL_RELEASE({
+    nvfp4_quantizer_cpp->amax_reduction_group->allreduce_coalesced(amax_tensors, opts)->wait();
+  });
+}
+
 }  // namespace
 
 py::object quantize(const at::Tensor &tensor, py::handle quantizer, const py::object &output,
-                    std::optional<at::Tensor> noop_flag) {
+                    std::optional<at::Tensor> noop_flag, bool compute_amax) {
```
timmoon10: This is leaking NVFP4-specific details into a generic API.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -130,6 +120,80 @@ def _group_quantize_for_grouped_mlp(
...
+        list(tensor.shape),
+        tensor.dtype,
+        tensor.device,
+        False,
+    )
+    if getattr(quantized, "_amax_rowwise", None) is not None:
+        quantized._amax_rowwise = rowwise_amax.view(-1)[:1]
+    if getattr(quantized, "_amax_columnwise", None) is not None:
+        quantized._amax_columnwise = columnwise_amax.view(-1)[:1]
+    quantized = tex.quantize(tensor, quantizer, quantized, compute_amax=False)
```
timmoon10: If the NVFP4 quantizer has `with_amax_reduction=True`, then the amax reduction is part of the `tex.quantize` contract. Rather than baking in an NVFP4-specific carveout into `tex.quantize`, it would be more natural to change the quantizer config. It's still a little messy, but it reveals the exact point we are doing something strange: we are ignoring the user if they manually set `with_amax_reduction=True`.
```suggestion
    quantizer.with_amax_reduction, old_with_amax_reduction = False, quantizer.with_amax_reduction
    quantized = tex.quantize(tensor, quantizer, quantized, compute_amax=False)
    quantizer.with_amax_reduction = old_with_amax_reduction
```

Alternatively, the real root cause is that amax reduction is primarily intended for tensor-parallel communication, but the grouped linear op doesn't support tensor parallelism. The right fix is to disable it in the quantizer, e.g. here: https://github.com/NVIDIA/TransformerEngine/blob/b24049b6fc54690cb3a3bd290a7665a3341b9302/transformer_engine/pytorch/ops/basic/grouped_linear.py#L701

transformer_engine/pytorch/ops/_common.py
```diff
@@ -61,36 +61,26 @@ def _nvidia_cudnn_frontend_supports_wgrad() -> bool:
     return _cudnn_frontend_version_supported()
 
 
-def _group_quantize_for_grouped_mlp(
+def _wrap_single_nvfp4_as_grouped(
     tensor: torch.Tensor,
-    quantizer: Quantizer,
-    num_groups: int,
+    quantized: object,
```
timmoon10: Do we actually handle `object`? We make so many internal assumptions that `quantized` has the attrs of an `NVFP4TensorStorage` that it's not really meaningful if we pass in anything else.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -61,36 +61,26 @@ def _nvidia_cudnn_frontend_supports_wgrad() -> bool:
...
-
-    # --------------------------------------------------
-    # Special case: single-tensor NVFP4 quantize
-    # --------------------------------------------------
-
-    quantized = tex.quantize(tensor, quantizer)
-    with_gemm_swizzled_scales = quantized._with_gemm_swizzled_scales
-    if quantizer.optimize_for_gemm:
+    """Wrap a single NVFP4 tensor in GroupedTensor storage."""
+    with_gemm_swizzled_scales = getattr(quantized, "_with_gemm_swizzled_scales", False)
```
timmoon10: The defensive `getattr` usage in this function and elsewhere is an antipattern. If we expect an attr that isn't there, then it's better to fail quickly and loudly than to tolerate incorrect behavior happening in secret. `getattr` also adds non-trivial CPU overhead compared to accessing attrs directly.

transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -129,14 +129,21 @@ void init_extension() {
   });
 }
 
+void bind_quantize_with_amax_extensions(py::module_ &m) {
+  m.def("group_quantize_with_amax", group_quantize_with_amax, py::arg("tensor"),
```
timmoon10: I'd prefer giving this a very specific name since some data formats don't have any notion of amaxes (MXFP8, DSv3 FP8) and other data formats use amaxes in a completely different way (FP8 delayed scaling). If you can't make a principled general API, give up and make a narrowly focused API.

```suggestion
  m.def("nvfp4_group_quantize_with_amax", nvfp4_group_quantize_with_amax, py::arg("tensor"),
```

timmoon10 (changes requested):

transformer_engine/pytorch/ops/fused/forward_grouped_mlp.py
timmoon10: Hm, all of these envvars show we don't have a great way of configuring the fused ops. Our external-facing API is based on unfused ops, which means that only general, non-fusion-specific options are exposed to the user. We tend toward envvars because we don't want to give the user access to the fused ops (they should get correct results even if with the unfused impl). I guess an alternative is to have a bunch of global variables, or APIs to set global variables, since the user can set these programmatically in Python rather than relying on the environment to be setup precisely right. However, I'm not too worried about envvars at the moment because these options are evolving so quickly that we're not sure which ones are going to stick.

transformer_engine/pytorch/ops/_common.py
vthumbe1503: We would need amax_reduction_group to be set for FSDP2 as well. We wont be able to support it if we disable it in quantizer.

transformer_engine/pytorch/csrc/extensions/cast.cpp
timmoon10: I think the best fix would be exposing a separate `nvfp4_quantize_with_amax` function, similar to `nvfp4_group_quantize_with_amax`.

timmoon10 (approved):
LGTM, pending CI

PR #3074, opened 2026-06-02
[PyTorch] Isolate CP pool worker stdout from NCCL/library banners

sudhakarsingh27:
# Problem

The CP attention test pool worker (added in #2993) uses rank-0 **stdout** as a line-delimited JSON protocol channel between the worker and the parent pytest process. `PoolWorker._submit_once` reads one line and `json.loads` it.

When `NCCL_DEBUG` is set to `VERSION` or `INFO`, NCCL writes a banner line — `NCCL version 2.x.y+cudaA.B\n` — to **stdout** (fd 1), not stderr. That banner reaches the parent ahead of the first JSON response, `json.loads` raises, and because the `cp_pool` fixture is session-scoped and the pool is killed on first failure, **every subsequent CP test in the file fails** with:

```
AssertionError: pool worker JSON protocol broke:
JSONDecodeError('Expecting value: line 1 column 1 (char 0)');
line='NCCL version 2.30.5+cuda13.3\n'
```

CI runners that export `NCCL_DEBUG=VERSION` (e.g. the cuDNN nightly GB200 job) hit this on all ~200 non-skipped `test_cp_with_flash_attention` / `test_cp_with_fused_attention` cases. With `NCCL_DEBUG` unset the banner isn't emitted, so the suite passes — which is why the regression only shows up on those runners.

The existing mitigation only redirected **non-rank-0** stdout to `/dev/null`, so rank 0's own banner still corrupted the stream.

# Fix

**Worker — `run_attention_with_cp_pool.py`:** before `import torch`, `dup()` rank-0's original fd 1 into a private stream (`_JSON_STDOUT`) reserved for the JSON protocol, then `dup2(2, 1)` so fd 1 points at stderr. Any banner from NCCL / cuBLAS / cuDNN / torch — Python-level or C-level, import-time or runtime — now lands on stderr (still drained into CI logs by the parent's stderr thread) instead of the protocol pipe. Responses are written through `_JSON_STDOUT`. The redirect runs at module scope, before `import torch`, so even import-time/C-level writes are covered.

The pre-existing `_silence_non_rank0_stdout` (rank>0 → `/dev/null`) is kept, so between the two redirects every rank's fd 1 is diverted away from the pipe. The asymmetry is intentional and documented in the code: rank 0 keeps a dup of fd 1 for the JSON channel and sends the rest to stderr (preserved in CI logs); rank>0 has nothing to preserve, so it goes to `/dev/null`.

**Parent — `test_attention_with_cp.py`:** comment-only update. The read path is unchanged (single `select` + `readline` + `json.loads`), since with both redirects in place the pipe carries only rank-0's JSON line. A non-JSON line is still turned into a clear `AssertionError` ("pool worker JSON protocol broke") rather than a raw `JSONDecodeError`.

### Why fd-level isolation rather than a stdout sentinel/skip

An earlier revision prefixed responses with a `[CP_POOL_RESP] ` sentinel and had the parent skip non-protocol lines. That was dropped after measuring it: with the rank-0 fd redirect in place, instrumenting the parent to count non-sentinel lines on the pipe under `NCCL_DEBUG=INFO` (max verbosity) across multiple cases yields **zero** — the torchrun agent logs to stderr, not this stdout pipe, so the sentinel filtered nothing. fd isolation also behaves better under `NCCL_DEBUG=INFO`: the (large) banner volume goes to stderr instead of forcing the parent to read+skip hundreds of lines within the response deadline. Net result is a smaller, single-mechanism fix.

# Validation

8×H100, TE built from this commit, with the NCCL banner confirmed emitted in every run (`grep "NCCL version"`):

| Scenario | Result |
|---|---|
| `NCCL_DEBUG=VERSION`, flash p2p + all_gather + a2a (one session, pool reused across the banner) | ✅ 3 passed |
| `NCCL_DEBUG=INFO` (verbose), flash p2p + fused p2p | ✅ 2 passed |
| `NCCL_DEBUG=VERSION`, fused p2p | ✅ 1 passed |
| `NCCL_DEBUG` unset (control) | ✅ 1 passed |
| **L1 det/non-det parallel suite** (`qa/L1_pytorch_distributed_unittest/test.sh` pattern, `NCCL_DEBUG=VERSION`) | ✅ non-det 45 passed / det 33 passed, 0 protocol breaks |
| **Without this fix**, `NCCL_DEBUG=VERSION` | ❌ `pool worker JSON protocol broke` (regression reproduced) |

# Type of change

- [x] Bug fix (test infrastructure; no library/runtime code change)

# Checklist

- [x] Contributing guidelines followed
- [x] Changes commented where non-obvious
- [x] No new warnings
- [x] Existing test suite serves as input + validation
- [x] Existing tests pass locally (8×H100, `NCCL_DEBUG=VERSION`/`INFO`/unset; L1 det+non-det)


tests/pytorch/attention/run_attention_with_cp_pool.py
```diff
@@ -51,10 +73,19 @@ def _recv_request(rank: int) -> dict:
...
+# pipe and is not covered by it; the prefix lets the parser pick out responses.
+_RESP_PREFIX = "[CP_POOL_RESP] "
+
+
 def _send_response(rank: int, payload: dict) -> None:
     if rank == 0:
-        sys.stdout.write(json.dumps(payload) + "\n")
-        sys.stdout.flush()
+        if _JSON_STDOUT is None:
+            raise RuntimeError("rank-0 JSON stdout was not initialized")
```
sudhakarsingh27: Unreachable guard. `_send_response` only writes when `rank == 0`, and on rank 0 `_redirect_rank0_stdout_to_stderr()` (called unconditionally at import, when `RANK=="0"`) always sets `_JSON_STDOUT`. The only way `_JSON_STDOUT` stays None on rank 0 is if `RANK` is unset/non-zero at import but `rank` resolves to 0 in main() — which can't happen since both read the same env var. Drop the check; it adds noise without covering a real path.

tests/pytorch/attention/run_attention_with_cp_pool.py
```diff
@@ -51,10 +73,19 @@ def _recv_request(rank: int) -> dict:
     return box[0]
 
 
+# Sentinel prefix on every response line so the parent can skip any chatter that
+# still reaches the protocol pipe. The rank-0 fd redirect handles this worker's
+# own library/Python writes, but the torchrun *agent* process shares the same
+# pipe and is not covered by it; the prefix lets the parser pick out responses.
+_RESP_PREFIX = "[CP_POOL_RESP] "
```
sudhakarsingh27: Two mechanisms now solve overlapping problems: the rank-0 fd redirect (covers worker process writes) and the sentinel prefix (covers the torchrun agent process, which the worker can't touch). That split is genuinely justified — keep both. But the redirect makes the *worker*'s contribution to pipe noise effectively zero, so the only remaining producer the sentinel protects against is the torchrun agent. Worth stating that explicitly in one place. Also consider whether agent-process output on this pipe was actually observed or is speculative — if speculative, the sentinel alone is the cheaper standalone fix.

tests/pytorch/attention/run_attention_with_cp_pool.py
```diff
@@ -32,6 +32,28 @@
...
+    """Reserve rank-0's original stdout fd for the JSON channel; send the rest to stderr."""
+    global _JSON_STDOUT
+    if os.environ.get("RANK") != "0" or _JSON_STDOUT is not None:
+        return
+    _JSON_STDOUT = os.fdopen(os.dup(1), "w", buffering=1)
+    os.dup2(2, 1)
+    sys.stdout = sys.stderr
+
+
+_redirect_rank0_stdout_to_stderr()
```
sudhakarsingh27, line 55: Import-time side-effect call is correct here and must stay at module scope — the redirect has to happen before `import torch` to catch C-level/import-time writes, so it can't move into main(). Good. Minor: `_redirect_rank0_stdout_to_stderr` is only ever called once, immediately below its def; the `_JSON_STDOUT is not None` re-entry guard is dead. Either inline the body (no function) or drop the guard. The function wrapper does aid readability, so keeping the function and dropping the guard is the cleaner choice.

tests/pytorch/attention/run_attention_with_cp_pool.py
```diff
@@ -32,6 +32,28 @@
...
+
+
+def _redirect_rank0_stdout_to_stderr() -> None:
+    """Reserve rank-0's original stdout fd for the JSON channel; send the rest to stderr."""
+    global _JSON_STDOUT
+    if os.environ.get("RANK") != "0" or _JSON_STDOUT is not None:
+        return
+    _JSON_STDOUT = os.fdopen(os.dup(1), "w", buffering=1)
+    os.dup2(2, 1)
+    sys.stdout = sys.stderr
```
sudhakarsingh27, line 52: fd ops are idiomatic and correct: `os.dup(1)` snapshots the original stdout, `os.fdopen(..., buffering=1)` gives a line-buffered text wrapper, `os.dup2(2, 1)` repoints fd 1 at stderr. One asymmetry worth a comment: rank 0's library output now goes to stderr (preserved in CI logs), but `_silence_non_rank0_stdout` sends rank>0 output to /dev/null (discarded). That's defensible (rank 0 is representative) but the two functions now implement deliberately different policies — note the rationale so a future reader doesn't "unify" them.

tests/pytorch/attention/test_attention_with_cp.py
```diff
@@ -214,34 +218,46 @@ def _submit_once(self, kwargs: dict, timeout: float) -> None:
...
+                msg = self._diag("pool worker died mid-request")
+                self._kill()
+                raise AssertionError(msg)
+
+            if line.startswith(self._RESP_PREFIX):
+                resp_line = line[len(self._RESP_PREFIX) :]
+                break
+            # Non-protocol line — echo to stderr so it stays in CI logs.
+            sys.stderr.write(line)
+            sys.stderr.flush()
```
sudhakarsingh27: Read loop is clean and the deadline correctly bounds the whole loop. Two edge cases: (1) `select()` says ready but `readline()` can still block if the worker emitted a partial line without a newline — pre-existing, low risk over a pipe, fine to leave. (2) A pathological worker that emits unbounded non-sentinel lines faster than the deadline would keep this loop echoing forever-ish until the deadline; acceptable since the deadline still caps it. No change required, just confirming the loop is sound.

sudhakarsingh27 (commented):
Targeted fix, mostly minimal. Main asks: drop the unreachable `_JSON_STDOUT is None` raise in `_send_response` and the dead re-entry guard in `_redirect_rank0_stdout_to_stderr`. The fd-redirect + sentinel pair is justified (worker process vs torchrun agent process — different coverage), but confirm the agent actually writes to this pipe; if that's speculative the sentinel alone suffices. fd-level ops are correct and the import-time placement is right. Comments are on the verbose side but explain real non-obvious intent, so fine to keep.

tests/pytorch/attention/run_attention_with_cp_pool.py
sudhakarsingh27: Done — guard removed in `e543ef89`. `_send_response` now writes directly to `_JSON_STDOUT`.

tests/pytorch/attention/run_attention_with_cp_pool.py
sudhakarsingh27: Done — dropped the `_JSON_STDOUT is not None` re-entry guard, kept the function. `e543ef89`.

tests/pytorch/attention/run_attention_with_cp_pool.py
sudhakarsingh27: Tested the speculative part: instrumented the parent to count non-sentinel lines reaching the pipe and ran 3 cases under `NCCL_DEBUG=INFO` (max verbosity) — **0 lines** skipped. With the rank-0 fd redirect in place, the torchrun agent writes nothing to this stdout pipe (its logs go to stderr). So the sentinel wasn't load-bearing.

Removed the sentinel + parent skip-loop entirely in `e543ef89`. The fd redirect alone fully isolates the pipe, and it's actually *better* under `NCCL_DEBUG=INFO` (banner volume goes to stderr instead of forcing the parent to read+skip hundreds of lines within the deadline). The parent reverts to a single read + `json.loads`.

tests/pytorch/attention/run_attention_with_cp_pool.py
sudhakarsingh27: Added a note in the `_silence_non_rank0_stdout` docstring explaining the deliberate asymmetry: rank 0 keeps a dup of fd 1 for the JSON channel and sends the rest to stderr (preserved in CI logs), while rank>0 has nothing to preserve so it goes to /dev/null. `e543ef89`.

tests/pytorch/attention/test_attention_with_cp.py
sudhakarsingh27: Moot now — the skip-loop was removed (see the sentinel thread). With the fd redirect the pipe carries only the rank-0 JSON line, so the parent is back to a single `select` + `readline` + `json.loads`, no loop. The two edge cases you noted no longer apply.

tests/pytorch/attention/run_attention_with_cp_pool.py
```diff
@@ -51,10 +73,19 @@ def _recv_request(rank: int) -> dict:
...
 
 
+# Sentinel prefix on every response line so the parent can skip any chatter that
+# still reaches the protocol pipe. The rank-0 fd redirect handles this worker's
+# own library/Python writes, but the torchrun *agent* process shares the same
+# pipe and is not covered by it; the prefix lets the parser pick out responses.
+_RESP_PREFIX = "[CP_POOL_RESP] "
+
+
 def _send_response(rank: int, payload: dict) -> None:
```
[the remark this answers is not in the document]
sudhakarsingh27: Resolved by removing the sentinel entirely in `e543ef89` — the rank-0 fd redirect keeps the pipe clean on its own (verified: 0 non-protocol lines under `NCCL_DEBUG=INFO`), so there's no longer a duplicated `_RESP_PREFIX` to keep in sync across the two files.

cyanguwa (approved):
Please make sure you test with NCCL_DEBUG turned on. Thanks for the PR!

PR #3078, opened 2026-06-03
[PyTorch] Expose interleave and de-interleave function for GLU tensor for fused grouped MLP via fused ops

ksivaman:
# Description

Add utilities to be used to convert FC1 weight format when loading checkpoints. The fused grouped MLP op uses interleaved format for FC1 weight. This functionality was originally added to https://github.com/NVIDIA-NeMo/Megatron-Bridge/pull/2841.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add utility functions to interleave and de-interleave FC1 weight.

# Checklist:


transformer_engine/pytorch/utils.py
```diff
@@ -82,6 +88,96 @@ def get_device_compute_capability() -> Tuple[int, int]:
     return _get_device_compute_capability(torch.cuda.current_device())
 
 
+def deinterleave_glu_tensor(tensor: torch.Tensor, interleave_size: int) -> torch.Tensor:
+    """Convert a block-interleaved SwiGLU fc1 tensor to contiguous gate/linear layout.
+
+    Fused SwiGLU kernels (for example :class:`~transformer_engine.pytorch.ops.SwiGLU`
+    with ``glu_interleave_size`` set) expect fc1 weights in a block-interleaved
```
vthumbe1503, line 95: Isnt it activations that need SwiGLU? Is it that we need to interleave the weights the same way as the activations? Clarification in the comment would be helpful.

transformer_engine/pytorch/utils.py
```diff
@@ -82,6 +88,96 @@ def get_device_compute_capability() -> Tuple[int, int]:
...
+    """
+    shape = tensor.shape
+    x = tensor.reshape(
+        2,
+        shape[0] // (2 * interleave_size),
+        interleave_size,
+        *shape[1:],
+    )
+    x = x.transpose(0, 1).contiguous()
+    return x.reshape(shape)
```
vthumbe1503, line 185: I am curious if we have done perf analysis of this function with and without torch compile and whether it is worth jitting it with torch.compile. I would assume that would generate a triton kernel(not sure if it is a performant one?). Although I wouldnt block this PR for that

transformer_engine/pytorch/utils.py
```diff
@@ -82,6 +88,96 @@ def get_device_compute_capability() -> Tuple[int, int]:
...
+        shape[0] // (2 * interleave_size),
+        2,
+        interleave_size,
+        *shape[1:],
+    )
+    x = x.transpose(0, 1).contiguous()
+    return x.reshape(shape)
+
+
+def interleave_glu_tensor(tensor: torch.Tensor, interleave_size: int) -> torch.Tensor:
```
vthumbe1503, line 139: I guess we should use this function in our ScaledActivation Sequential Op?
That way we are also in some way testing this API.

vthumbe1503 (commented):
LGTM. Left few minor comments.

transformer_engine/pytorch/utils.py
timmoon10: We could probably improve perf with `tex.swap_first_dims`. That said, I'd expect perf is not critical if this is primarily for checkpointing.

transformer_engine/pytorch/utils.py
```diff
@@ -82,6 +88,96 @@ def get_device_compute_capability() -> Tuple[int, int]:
     return _get_device_compute_capability(torch.cuda.current_device())
 
 
+def deinterleave_glu_tensor(tensor: torch.Tensor, interleave_size: int) -> torch.Tensor:
+    """Convert a block-interleaved SwiGLU fc1 tensor to contiguous gate/linear layout.
```
timmoon10: Nit: Prefer "GLU" since this trick is used for both SwiGLU and QGEGLU.

transformer_engine/pytorch/utils.py
vthumbe1503: I was thinking whether we could reuse this for activations as well. But seems like not. The logic in terms of dim ordering is a bit different
https://github.com/NVIDIA/TransformerEngine/blob/c1e827f80637113a98f81b50a9e3ad4ebec8dfc2/transformer_engine/pytorch/ops/basic/swiglu.py#L338

transformer_engine/pytorch/utils.py
vthumbe1503: Seems like this cant be used for activation and is specific to the weights. I am wondering if we can make this API a bit more general purpose to reuse it for activation as well

https://github.com/NVIDIA/TransformerEngine/blob/c1e827f80637113a98f81b50a9e3ad4ebec8dfc2/transformer_engine/pytorch/ops/basic/swiglu.py#L338

timmoon10: If the helper function makes the op implementation easier to understand, I'd support this. This is more important than perf though. GLU interleaving is only really beneficial when the activation is fused into a GEMM, and it'll always be a useless memory-bound kernel in the unfused op.

ksivaman: Yes this is meant to be used a one time helper during checkpoint loading. Since this is not the most performant solution, we want to avoid use in the actual network runtime.

transformer_engine/pytorch/utils.py
ksivaman:
```suggestion
    """Convert a block-interleaved GLU fc1 tensor to contiguous gate/linear layout.
```

transformer_engine/pytorch/utils.py
```diff
@@ -82,6 +88,96 @@ def get_device_compute_capability() -> Tuple[int, int]:
...
+
+    Fused SwiGLU kernels (for example :class:`~transformer_engine.pytorch.ops.SwiGLU`
```
ksivaman:
```suggestion
    Fused GLU kernels (for example :class:`~transformer_engine.pytorch.ops.SwiGLU`
```

transformer_engine/pytorch/utils.py
```diff
@@ -82,6 +88,96 @@ def get_device_compute_capability() -> Tuple[int, int]:
...
+    remaining dimensions) and ``linear_fc1.bias`` (dimension 0 only).
+
+    Parameters
+    ----------
+    tensor : torch.Tensor
+        Tensor in block-interleaved layout. The length of dimension 0 must be
+        divisible by ``2 * interleave_size``.
+    interleave_size : int
+        Number of rows (for weights) or elements (for bias) per gate/linear block.
+        Fused TE SwiGLU paths commonly use ``32``.
```
ksivaman:
```suggestion
        Fused TE GLU paths commonly use ``32``.
```

transformer_engine/pytorch/utils.py
ksivaman: Done

timmoon10 (approved):
LGTM

vthumbe1503 (approved):

PR #3088, opened 2026-06-05
fix(topk): fix UB and prevent vector load splitting in standalone_topk

solos:
# Description

This PR refactors the vectorized processing logic in standalone_topk to eliminate C++ undefined behavior and ensure optimal GPU instruction emission. The changes address a critical issue where conditional vector loads were causing the compiler to split 128-bit memory transactions into scalar operations, degrading performance. Additionally, it fixes potential edge cases related to unsigned integer underflow and out-of-bounds memory access.

Fixes #

Fix C++ UB by replacing union type-punning with __builtin_memcpy.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Fix C++ UB by replacing union type-punning with `__builtin_memcpy`.
- Prevent NVCC from splitting 128-bit vector loads by using unconditional pointer selection for safe companion reads.
- Fix `len_cast_for_sync` underflow when `len_cast == 0`.
- Guard against OOB reads when input length is smaller than `sizeof(WideT)`.

# Checklist:

### DIFF
--- a/transformer_engine/common/util/standalone_topk.cuh
+++ b/transformer_engine/common/util/standalone_topk.cuh
@@ -252,9 +252,9 @@ __device__ void vectorized_process(const T *in, idxT len, Func f, int sync_width
         const bool valid = i < len_cast;
         // Unconditional 128-bit vector load: invalid threads read in_cast[0] (cached,
         // discarded via valid=false) so NVCC emits LDG.E.128 instead of predicated load.
-        // Safe because len_cast > 0 guarantees at least one valid WideT at in_cast[0].
-        const WideT *load_ptr = valid ? &in_cast[i] : &in_cast[0];
-        const WideT wide_data = *load_ptr;
+        // Index clamping (not pointer ternary) avoids C++ UB from &in_cast[i] when i >= len_cast.
+        const idxT safe_i = valid ? i : static_cast<idxT>(0);
+        const WideT wide_data = in_cast[safe_i];
         T local_array[items_per_scalar];  // NOLINT(runtime/arrays)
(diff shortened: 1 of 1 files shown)


ptrendx (changes requested):
No. Using the union to achieve the vectorized accesses is a very wide practice in CUDA and endorsed by the NVCC compiler experts. As for the second issue - what is the case where you saw the issue (shape of the inputs etc.)? If len_cast is less than 0 then len_cast_for_sync is also going to be less than 0 and so the loop will just not run - I don't believe this PR actually changes the behavior of the kernel in any way.

solos:
> No. Using the union to achieve the vectorized accesses is a very wide practice in CUDA and endorsed by the NVCC compiler experts. As for the second issue - what is the case where you saw the issue (shape of the inputs etc.)? If len_cast is less than 0 then len_cast_for_sync is also going to be less than 0 and so the loop will just not run - I don't believe this PR actually changes the behavior of the kernel in any way.

Regarding the claim that the loop won't run when len_cast is negative: that's not always true. Due to C++'s truncation-toward-zero integer division. For example, if len_cast = -1 and sync_width = 32, then len_cast - 1 = -2, and -2 / 32 is 0 (not -1), so len_cast_for_sync becomes (0 + 1) * 32 = 32. The loop then executes multiple iterations even though len_cast is negative, with valid always false. While the invalid iterations are logically guarded by valid, the expression &in_cast[i] (or the pointer arithmetic in in_cast[i]) is still evaluated for out-of-bounds indices, which is undefined behavior. The added if (len_cast > 0) guard correctly prevents this edge case and does change the behavior for safety.

PR #3090, opened 2026-06-05
Make NVTE tensor handle pool size configurable

lhb8125:
## Summary
- Add runtime environment variables to configure the internal NVTETensor and NVTEGroupedTensor handle pool sizes.
- Keep the default pool size at 20 MiB to preserve existing behavior.
- Improve exhaustion errors to mention the configured pool size and the env var to increase.

## Motivation
Large model initialization paths can legitimately create more TE tensor handles than the current fixed-size pool allows, even when GPU and CPU memory are otherwise sufficient. Exposing the pool size as an environment variable avoids downstream source patches for these scale-dependent cases.

## Testing
- git diff --check

### DIFF
--- a/transformer_engine/common/transformer_engine.cpp
+++ b/transformer_engine/common/transformer_engine.cpp
@@ -408,12 +408,11 @@ size_t GetTensorHandlePoolSizeMB(const char *env_var) {
   return pool_size_mb;
 }
 
-size_t GetTensorHandlePoolCapacity(size_t pool_size_mb, size_t handle_size,
-                                   const char *handle_name, const char *env_var) {
+size_t GetTensorHandlePoolCapacity(size_t pool_size_mb, size_t handle_size, const char *handle_name,
+                                   const char *env_var) {
   const size_t pool_size_bytes = pool_size_mb * kBytesPerMB;
   NVTE_CHECK(pool_size_bytes >= handle_size, env_var, "=", pool_size_mb,
-             " MiB is too small for one ", handle_name, " handle of size ", handle_size,
-             " bytes.");
+             " MiB is too small for one ", handle_name, " handle of size ", handle_size, " bytes.");
   return pool_size_bytes / handle_size;
 }
 
@@ -432,8 +431,8 @@ class TensorAllocator {
     std::lock_guard<std::mutex> lock(mutex);
     const size_t available = free_list.size() + (memory.capacity() - memory.size());
     NVTE_CHECK(available >= N, "Cannot allocate ", N,
-               " new NVTETensors. Maximum number of tensors reached: ", MAX_TENSOR_NUM,
-               " (", TENSOR_HANDLE_POOL_SIZE_MB,
+               " new NVTETensors. Maximum number of tensors reached: ", MAX_TENSOR_NUM, " (",
+               TENSOR_HANDLE_POOL_SIZE_MB,
                " MiB handle pool). If your application legitimately needs more tensor handles, "
                "increase NVTE_TENSOR_HANDLE_POOL_SIZE_MB.");
     for (size_t i = 0; i < N; ++i) {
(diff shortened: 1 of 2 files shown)


transformer_engine/common/transformer_engine.cpp
```diff
@@ -393,6 +395,29 @@ void CheckOutputGroupedTensor(const GroupedTensor &t, std::string_view name, boo
...
+namespace {
+
+constexpr size_t kDefaultTensorHandlePoolSizeMB = 20;
+constexpr size_t kBytesPerMB = 1024 * 1024;
+
+size_t GetTensorHandlePoolSizeMB(const char *env_var) {
+  const size_t pool_size_mb = getenv<size_t>(env_var, kDefaultTensorHandlePoolSizeMB);
+  NVTE_CHECK(pool_size_mb > 0, env_var, " must be a positive integer.");
+  NVTE_CHECK(pool_size_mb <= std::numeric_limits<size_t>::max() / kBytesPerMB, env_var,
+             " is too large.");
```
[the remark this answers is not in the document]
lhb8125: Fixed in 5c9ee293b44662576c9442b4296d61d9f96a7361 and kept in the latest PR head 935054d3d2f3baef740d3eeb9773d2f7a25f7326. The pool-size env vars now use local full-string validation instead of `getenv<size_t>`, so negative values such as `-1` and partially parsed values such as `50bad` are rejected with the positive-integer diagnostic before the overflow guard.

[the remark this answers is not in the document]
lhb8125: Fixed in 5c9ee293b44662576c9442b4296d61d9f96a7361 and kept in the latest PR head 935054d3d2f3baef740d3eeb9773d2f7a25f7326. The pool-size env vars now use local full-string validation instead of `getenv<size_t>`, so negative values such as `-1` and partially parsed values such as `50bad` are rejected with the positive-integer diagnostic before the overflow guard.

ptrendx:
I am not opposed to creating such a variable, but I would really like to see an example of such legitimate use which goes over this limit. Could you run the experiment that is failing for you with https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/common/transformer_engine.cpp#L487 set to true and send me the log of that?

lhb8125:
> I am not opposed to creating such a variable, but I would really like to see an example of such legitimate use which goes over this limit. Could you run the experiment that is failing for you with https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/common/transformer_engine.cpp#L487 set to true and send me the log of that?

[te_handle_pool_debug_fakepg_20260616-010629_2134338.pruned.log](https://github.com/user-attachments/files/28993375/te_handle_pool_debug_fakepg_20260616-010629_2134338.pruned.log)
I attached a pruned version, which replaces the repeated lines with a comment. The case is EP1 for Qwen3-235B training, so the number of local experts is 128, where we adopt full recompute to control the dynamic activation's memory footprint.
I sent the whole log to you via slack.

transformer_engine/common/transformer_engine.cpp
```diff
@@ -393,6 +395,49 @@ void CheckOutputGroupedTensor(const GroupedTensor &t, std::string_view name, boo
...
+
+  size_t pool_size_mb = 0;
+  for (size_t i = first; i <= last; ++i) {
+    NVTE_CHECK(value[i] >= '0' && value[i] <= '9', env_var, " must be a positive integer, got \"",
+               value, "\".");
+    const size_t digit = static_cast<size_t>(value[i] - '0');
+    NVTE_CHECK(pool_size_mb <= (std::numeric_limits<size_t>::max() - digit) / 10, env_var,
+               " is too large.");
+    pool_size_mb = pool_size_mb * 10 + digit;
+  }
```
ptrendx: Why not just use `std::atoi`?

ptrendx:
Ok, I see why it happens. I'm ok with this PR (modulo one comment about string to integer conversion). Now, the real reason why this happens, while legitimate, is also very bad performance-wise (logic for the master weight casting to lower precision) and this should also be fixed in a separate PR.

transformer_engine/common/transformer_engine.cpp
lhb8125: I avoided std::atoi here because it does not give us enough validation/error handling for an env var that controls an allocation size. In particular, it cannot distinguish parse failure from 0, accepts prefixes such as "50bad", returns int rather than size_t, and gives poor behavior/diagnostics for negative or overflowing inputs.

The current parser is a bit more verbose, but it intentionally enforces a full-string positive integer and checks overflow before converting to bytes, so invalid values fail with a clear NVTE_CHECK message.

If you prefer using a standard parser, I can switch this to std::strtoull/std::from_chars with explicit end-pointer/overflow checks, but I would avoid std::atoi for this case.

ptrendx: Considering that it is supposed to take the value in MBs, at least some of the concerns here are not really relevant - e.g. I do not expect anybody to have a problem with int vs size_t here as the values would be < 1000. I'm fine with strtoul (again, we don't need ull here) - basically trying to make sure that the code is as understandable as possible.

lhb8125: Thanks, agreed. I simplified this to use std::strtoul and kept the validation intentionally small:

- trim surrounding whitespace
- require the first non-space character to be a digit
- require strtoul to consume the full trimmed value
- require the parsed value to be > 0
- check ERANGE for overflow

ptrendx (approved):

lhb8125:
@ptrendx could you help launch the merging process?

PR #3093, opened 2026-06-05
[PyTorch] Enable Weight Preswizzling during Quantization in TE Modules

cael-ling:
# Description

## Motivation
For block-scaled NVFP4, a cached weight is used in two GEMMs per step — fprop (row-wise scales) and dgrad (column-wise scales) — and each GEMM needs its scale factors in the GEMM-swizzled layout. Today that swizzle is recomputed lazily inside `general_gemm` on every micro-batch and thrown away, so with `N` micro-batches the weight scale swizzle runs `2*N` times per step even though the weight is quantized only once, which hurts performance. (Activation quantizers already set `optimize_for_gemm=True` and were pre-swizzled; only the weight was missed.)

## What this PR does
This PR sets `weight_quantizer.optimize_for_gemm=True` in both MXFP8/NVFP4 recipes to enable pre-swizzling of weights. This also allows us to save the preswizzled quantized weights in cache when we enable workspace caching. NOTE: optimize_for_gemm isnt set in case of FP8/FP4 primary weights(quantized_model_init) since it we cant dequantize swizzled weights for now(needed in optimizer.step)

- Applied to `Linear`, `LayerNormLinear`, `LayerNormMLP` (fc1 + fc2) and `GroupedLinear` (per expert).
- No-op for recipes whose scales do not require swizzling (e.g. per-tensor FP8).
- Swizzling is a pure layout permutation, so **numerics are unchanged**.

- New `tests/pytorch/nvfp4/test_weight_swizzle_in_layers.py`
  -  When workspace is cached, this makes sure`_with_gemm_swizzled_scales` is set and persisted on the cached workspace.
  - `_with_gemm_swizzled_scales` is not set in case of module initialized with fp8 primary weights(quantized_model_init)
- `pytest tests/pytorch/test_numerics.py -k "linear or layernorm or mlp"` — no regressions.
- `pytest tests/pytorch/test_grouped_linear.py -k "not grouped_tensor and not fused_path"` — no regressions.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


tests/pytorch/nvfp4/test_nvfp4_weight_swizzle_cache.py
```diff
@@ -0,0 +1,192 @@
...
+
+
+@pytest.mark.parametrize("kind", ["Linear", "LayerNormLinear"])
+def test_lazy_path_not_swizzled(kind):
+    """Without weight caching (is_first_microbatch=None) no workspace is created
+    and the optimization stays off — guards against accidentally always-on."""
+    torch.manual_seed(0)
+    device = "cuda"
+    recipe = NVFP4BlockScaling(disable_stochastic_rounding=True)
+    module = _make_module(kind, 1024, 1024, device)
```
vthumbe1503: Nit:
```suggestion
@pytest.mark.parametrize("layer_type", ["Linear", "LayerNormLinear"])
def test_lazy_path_not_swizzled(kind):
    """Without weight caching (is_first_microbatch=None) no workspace is created
    and the optimization stays off — guards against accidentally always-on."""
    torch.manual_seed(0)
    device = "cuda"
    recipe = NVFP4BlockScaling(disable_stochastic_rounding=True)
    module = _make_module(layer_type, 1024, 1024, device)
```

tests/pytorch/nvfp4/test_nvfp4_weight_swizzle_cache.py
```diff
@@ -0,0 +1,192 @@
...
+        dst_params = dict(dst.named_parameters())
+        for name, param in src.named_parameters():
+            dst_params[name].copy_(param)
+
+
+def _step(module, x, is_first, recipe):
+    x = x.detach().clone().requires_grad_(True)
+    module.zero_grad(set_to_none=True)  # per-micro-batch grads (no accumulation)
+    with te.autocast(enabled=True, recipe=recipe):
+        out = module(x, is_first_microbatch=is_first)
```
vthumbe1503: If absence of m_splits argument is the only reason for creating new test for grouped_linear below, then can we add a check on the module in terms of passing m_splits only if module is GroupedLinear, instead of duplicating the test?

vthumbe1503 (commented):
Apart from FSDP2 condition being irrelevant, LGTM

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1735,6 +1735,18 @@ def forward(
...
+            # (fprop rowwise + dgrad columnwise, redone every microbatch) collapses
+            # from 2*num_microbatches kernels to 2 per step per expert. Gated to the
+            # cached, non-FSDP path (FSDP/FSDP2 all-gather weights with un-swizzled
+            # scales; see NVFP4Tensor.fsdp_pre_all_gather), so pre-swizzling is
+            # unsupported there. No-op for non-swizzled recipes (e.g. per-tensor FP8).
+            if cache_weight and self.fsdp_group is None and not self.is_fsdp2:
+                for weight_quantizer in weight_quantizers:
+                    if weight_quantizer is not None:
+                        weight_quantizer.optimize_for_gemm = True
+
```
vthumbe1503: I dont think the comment is relevant In case of FSDP/FSDP2,
For FSDP, The scales are not sharded, and the whole scales are replicated across ranks today. So it doesnt matter if scales are swizzled or not. cc: @denera. Also NVFP4 pre allgather is function specific to FSDP2 not FSDP.
For FSDP2, we havent been caching weights as it causes memory bloating. And weight caching as a mechanism doesnt fit well with fsdp2.  This was done for linear and layer_norm_linear but apparently not for grouped_linear in this PR https://github.com/NVIDIA/TransformerEngine/pull/2805. But fixing that for grouped_linear might be byond scope of this PR. Even if weight caching is still kept as it is, current behavior is to save the entire weight instead of shard in the workspace and so swizzling being present shouldnt cause any issue.

```suggestion
            # Pre-swizzle (and cache) the weight scale factors when the quantized
            # weights are cached across microbatches, so the per-GEMM scale swizzle
            # (fprop rowwise + dgrad columnwise, redone every microbatch) collapses
            # from 2*num_microbatches kernels to 2 per step per expert.
            # No-op for non-swizzled recipes (e.g. per-tensor FP8).
            if cache_weight:
                for weight_quantizer in weight_quantizers:
                    if weight_quantizer is not None:
                        weight_quantizer.optimize_for_gemm = True

```

vthumbe1503: Same applies in other files.

cael-ling: Applied to all four module files.

tests/pytorch/nvfp4/test_nvfp4_weight_swizzle_cache.py
cael-ling: Done — test_lazy_path_not_swizzled now parametrizes over all four module kinds.

tests/pytorch/nvfp4/test_nvfp4_weight_swizzle_cache.py
cael-ling: Done — folded GroupedLinear into the parametrized test_weight_swizzle_cache_numerics, passing m_splits only for GroupedLinear (in _step); removed the duplicated grouped-only test.

tests/pytorch/nvfp4/test_nvfp4_weight_swizzle_cache.py
```diff
@@ -0,0 +1,192 @@
...
+    out.sum().backward()
+    return out.detach().float(), x.grad.detach().float()
+
+
```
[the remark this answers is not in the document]
cael-ling: Added LayerNormMLP coverage (fc1+fc2 two-quantizer path) to both parametrized tests.

cael-ling:
Pushed a commit addressing the review: removed the irrelevant FSDP gating across all four modules, merged the GroupedLinear test, and added LayerNormMLP coverage. Please take a look, thanks. @vthumbe1503

vthumbe1503 (commented):
Thanks for making the changes. I think that optimize_for_gemm(weight preswizzling) should be enabled for mostly all use-cases except for primary weights in fp8/fp4.

For primary weights in fp8/fp4 it wont work because of
1. Dequantization needs in optimizer step update which wont work on swizzled weights
2. FSDP2 allgather is currently supported only for unswizzled weights.

So lets enable it for most use-cases @cael-ling instead of restricting it to weight caching only.
https://github.com/cael-ling/TransformerEngine/pull/1

vthumbe1503 (approved):

vthumbe1503:
Pipeline: 55637043
Errors are pre-existing in main

PR #3098, opened 2026-06-05
[JAX] Skip score_mod tests on older cuDNN

vcherepanov-nv:
# Description

Skip score-mod tests if cuDNN backend version is less than 9.23

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Skip score-mod tests if cuDNN backend version is less than 9.23

# Checklist:

### DIFF
--- a/tests/jax/test_fused_attn_score_mod.py
+++ b/tests/jax/test_fused_attn_score_mod.py
@@ -24,6 +24,8 @@ from test_fused_attn import FusedAttnRunner, SeqDescFormat
 
 _CONFIG_TEST_HEAD_DIM = 128
 _CONFIG_TEST_SCALING_FACTOR = 1.0 / sqrt(_CONFIG_TEST_HEAD_DIM)
+_SCORE_MOD_MIN_CUDNN_VERSION = (9, 23)
+_SCORE_MOD_MIN_CUDNN_VERSION_STRING = ".".join(map(str, _SCORE_MOD_MIN_CUDNN_VERSION))
 
 
 def _has_cudnn_frontend_python():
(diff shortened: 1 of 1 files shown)


tests/jax/test_fused_attn_score_mod.py
```diff
@@ -197,14 +199,17 @@ class ScoreModFusedAttnRunner(FusedAttnRunner):
...
         except ImportError:
             pytest.skip("cuDNN Python frontend is required for score_mod")
         version = tuple(int(part) for part in cudnn.backend_version_string().split(".")[:2])
-        if version < (9, 6):
-            pytest.skip("cuDNN score_mod SDPA requires cuDNN frontend 9.6 or newer")
+        if version < _SCORE_MOD_MIN_CUDNN_VERSION:
+            pytest.skip(
+                "cuDNN frontend score_mod SDPA requires "
+                f"cuDNN {_SCORE_MOD_MIN_CUDNN_VERSION_STRING} or newer"
+            )
```
KshitijLakhani, line 212: The function name can be misleading as the test now seems to skip on FE and BE versions now 
The new check is a BE version skip so i'd just modify that :)

KshitijLakhani: Another quick clarification: as this change is in the runner class so it should benefit the distributed tests as well right ?

vcherepanov-nv: > The function name can be misleading 

I wanted to avoid unnecessary code churn (actually, at first I renamed it, but then reverted seeing multiple files changes). Besides, it can make sense if you think in this way: "require _compatible_ cuDNN FE".

> Another quick clarification: as this change is in the runner class so it should benefit the distributed tests as well right 

I'd assume so,

KshitijLakhani: I think it's okay for now as it is test code, but if it was feature code, i think we'd want explicit verbiage.
Thanks for the PR. Approved !

KshitijLakhani (approved):

PR #3099, opened 2026-06-06
Introduce Mega-C++ to reduce CPU overhead

zhongbozhu:
# Description

UPDATES: We might prefer having a Mega-Kernel in favor of this approach. We can also refactor this PR once we have C++ interface of cudnn fusion kernels for FC1 + ACT + FC2 quant for mxfp8 recipe otherwise users have to choose between unfused GPU kernels or better host-side implementation.  

Assistant: GPT5.5 codex 

Issue: https://github.com/NVIDIA/TransformerEngine/issues/2897 

Get rid of CPU overhead whenever CUDA Graph is not applicable. Guarded by NVTE_MEGACPP_GROUPED_LINEAR. 

Drop-in replace grouped MLP, ie. FC1 - act - FC2. Target BF16 grouped gemm with cublas grouped gemm backend. 

In the future, we can extend to mxfp8 / nvfp4 with cublas backend or even cuteDSL grouped gemm and call `cute.jit` in C++: https://github.com/NVIDIA/cutlass/pull/3289 

Recommend CUDA >= 13.2.1

Dependency of merge: https://github.com/NVIDIA/TransformerEngine/pull/3132  => this PR is rebased on top of this branch.

TODO: 

- [x] E2E training with some multimodal THD packing workloads 
- [x] Develop a better fused scaled swiglu kernel to replace torch ops
- [x] Attach before & after screenshots 
- [ ] Merge the dependency PR first to main and rebase 

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/csrc/extensions/pybind.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/pybind.cpp
@@ -357,6 +357,25 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
   m.def("te_general_grouped_gemm_for_discrete_out",
         &transformer_engine::pytorch::te_general_grouped_gemm_for_discrete_out,
         "Grouped GEMM for discrete output list");
+  m.def("megacpp_grouped_mlp_forward", &transformer_engine::pytorch::megacpp_grouped_mlp_forward,
+        "Mega C++ grouped MLP forward", py::arg("input"), py::arg("act_dtype"),
+        py::arg("split_sizes"), py::arg("fc1_weight"), py::arg("fc1_bias"), py::arg("fc2_weight"),
+        py::arg("fc2_bias"), py::arg("act_scales"), py::arg("activation"),
+        py::arg("glu_interleave_size"), py::arg("activation_limit") = 0.0,
+        py::arg("activation_alpha") = 0.0, py::arg("activation_glu_linear_offset") = 0.0,
+        py::arg("gemm_scratch") = py::none());
+  m.def("megacpp_grouped_mlp_backward", &transformer_engine::pytorch::megacpp_grouped_mlp_backward,
+        "Mega C++ grouped MLP backward", py::arg("grad_output"), py::arg("act_dtype"),
+        py::arg("split_sizes"), py::arg("x_offsets"), py::arg("fc1_offsets"),
+        py::arg("fc2_offsets"), py::arg("fc2_dy_offsets"), py::arg("base_offsets"), py::arg("x"),
+        py::arg("fc1_activation_input"), py::arg("fc2_x"), py::arg("act_scales"),
+        py::arg("fc1_weight"), py::arg("fc2_weight"), py::arg("fc1_wgrad_output"),
+        py::arg("fc1_compute_wgrad"), py::arg("fc1_accumulate_wgrad"), py::arg("fc2_wgrad_output"),
+        py::arg("fc2_compute_wgrad"), py::arg("fc2_accumulate_wgrad"), py::arg("activation"),
+        py::arg("glu_interleave_size"), py::arg("activation_limit") = 0.0,
+        py::arg("activation_alpha") = 0.0, py::arg("activation_glu_linear_offset") = 0.0,
+        py::arg("act_scales_requires_grad") = true, py::arg("input_requires_grad") = true,
+        py::arg("gemm_scratch") = py::none());
   m.def("fp8_transpose", &transformer_engine::pytorch::fp8_transpose, "Transpose with FP8 I/O",
         py::arg("input"), py::arg("dtype"), py::kw_only(), py::arg("out"),
         py::call_guard<py::gil_scoped_release>());
--- a/transformer_engine/pytorch/csrc/megacpp/grouped_mlp.cpp
+++ b/transformer_engine/pytorch/csrc/megacpp/grouped_mlp.cpp
@@ -1,0 +1,812 @@
+/*************************************************************************
+ * Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+
+#include <ATen/cuda/CUDAContext.h>
+#include <c10/cuda/CUDAGuard.h>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include <array>
+#include <optional>
+#include <string>
+#include <tuple>
+#include <utility>
+#include <vector>
+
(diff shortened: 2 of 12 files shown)


transformer_engine/pytorch/csrc/extensions/pybind.cpp
```diff
@@ -357,6 +357,23 @@ PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
   m.def("te_general_grouped_gemm_for_discrete_out",
         &transformer_engine::pytorch::te_general_grouped_gemm_for_discrete_out,
         "Grouped GEMM for discrete output list");
+  m.def("megacpp_grouped_mlp_forward", &transformer_engine::pytorch::megacpp_grouped_mlp_forward,
```
timmoon10, line 360: We should expose these functions within the `tex.grouped_mlp_experimental` submodule:
https://github.com/NVIDIA/TransformerEngine/blob/3fffa55da9e68ee814086cb3c7f4571ea5532574/transformer_engine/pytorch/csrc/extensions/pybind.cpp#L647-L650

transformer_engine/pytorch/csrc/megacpp/grouped_mlp.cpp
timmoon10, line 1: It would make more sense to organize:
```
csrc/
├── extensions/
│   ├── grouped_mlp_experimental/
│   │   ├── megacpp.cpp
│   │   └── grouped_mlp_experimental.cpp
│   ├── pybind.cpp
│   └── ...
```

If we implement more mega-C++ impls in the future, I don't see a reason why they would be more similar to each other than to the block they are fusing.

transformer_engine/pytorch/ops/fused/forward_grouped_mlp_megacpp.py
```diff
@@ -0,0 +1,380 @@
...
+    return False
+
+
+class _MegaCppActivationConfig(NamedTuple):
+    """Activation semantics consumed by the C++ grouped MLP path."""
+
+    name: str
+    is_scaled: bool
+    is_gated: bool
+    glu_interleave_size: int
```
timmoon10, line 89: Is it worth supporting GLU interleaving in the mega-C++ path? The only benefit is to support the fused GEMM+GLU kernel, and otherwise the unnecessary memory-bound kernel means perf is a lost cause. If we can simplify our optimized code paths, then it's worth it.

transformer_engine/pytorch/ops/fused/forward_grouped_mlp_megacpp.py
```diff
@@ -0,0 +1,380 @@
...
+            ops = ops[1:]
+
+    out.extend(window)
+    return out
+
+
+# Explicit env opt-in gives megacpp first chance. Unsupported recipes intentionally
+# return the ops unchanged so lower-priority recipe-specific fusers remain the
+# fallback path.
+register_forward_fusion(fuse_forward_megacpp_ops, prepend=True)
```
timmoon10, line 421: The GEMM+act fusions provide better GPU perf, so I think they should take higher priority than mega-C++. Basically, I see mega-C++ as "we can't do any better on GPU than the unfused impl, but at least we can make the CPU overhead very small".

transformer_engine/pytorch/ops/fused/forward_grouped_mlp_megacpp.py
zhongbozhu: > The only benefit is to support the fused GEMM+GLU kernel

I do hope in the future we can launch CuteDSL fused kernels in C++ with some TVM-FFI tricks, otherwise we are forced to choose either better kernel fusions or less CPU overhead. Currently the CuteDSL fusion path is very CPU bounded for small models and we rely on CUDA graph and paged stashing for it to work well

transformer_engine/pytorch/ops/fused/forward_grouped_mlp_megacpp.py
zhongbozhu: Current order is follows: 
1. check env var
2. env var = 1, then check supported recipe for mega-C++, so bf16 is supported, not mxfp8 / nvfp4
3. then for mxfp8, nvfp4, mega-C++ does fallback and check for the next fusion. 

The reasoning is that, I do not want the compromise of either better fusion or less host bound, so for future mxfp8 support, we can do the following two things:
1. directly do cuteDSL integration directly with tvm-ffi and do cublas as a backup plan
2. maybe add a new value to NVTE_MEGACPP_GROUPED_LINEAR=forced, so for users who cannot enable cuda graph for some reason, they can enforce C++ when they know that their training is more host bound

transformer_engine/pytorch/ops/fused/forward_grouped_mlp_megacpp.py
timmoon10: Why not make a separate fused op for mega-C++ CuTe DSL? It'll make the implementations less entangled, so there are fewer edge cases or complications that an agent might misunderstand.

zhongbozhu: yes, if there is a cutedsl version with mega C++, it's gonna take its own code path with zero code reuse since it's gonna be agent-assisted coding anyway

zhongbozhu:
E2E test - Qwen3.5 VL 35B-A3B, BSHD layout - about 6% E2E
image

transformer_engine/pytorch/ops/fused/forward_grouped_mlp_megacpp.py
```diff
@@ -0,0 +1,421 @@
...
+
+def _megacpp_bias_arg(linear_op: GroupedLinear, dtype: torch.dtype) -> Optional[torch.Tensor]:
+    """Return a packed [G, N] high-precision bias tensor or None."""
+    grouped_bias = linear_op._get_grouped_bias_for_gemm(dtype)
+    if grouped_bias is None:
+        return None
+    return grouped_bias.rowwise_data.view(linear_op.num_groups, linear_op.out_features)
+
+
+class ForwardGroupedMLP_MegaCpp(FusedOperation):
```
timmoon10, line 208: Instead of having separate fused ops for forward and backward, we should implement a single forward-backward fused op (see https://github.com/NVIDIA/TransformerEngine/pull/3117). If the forward and backward fusions are different, then things are going to explode horribly. Also, if we know the forward and backward are together, then we can simplify and reduce unnecessary checks (e.g. we can eliminate things like `fc1_ctx.use_megacpp_grouped_mlp`).

PR #3101, opened 2026-06-06
Fix GroupedLinear FP8 calibration loop

fallintoplace:
## Summary

- remove the unused outer calibration loop in `GroupedLinear`
- calibrate each input and weight quantizer once per GEMM
- avoid repeating calibration `num_gemms` times for every GEMM

## Validation

- ran `python3 -m py_compile transformer_engine/pytorch/module/grouped_linear.py`
- ran `git diff --check -- transformer_engine/pytorch/module/grouped_linear.py`


vthumbe1503 (approved):
LGTM

PR #3104, opened 2026-06-08
[PyTorch][torch.compile] Decouple amax reduction group from the quantizer

pggPL:
# Description

There are 2 problems with amax_reduction_group on Quantizers:

- I want to declare quantizers as **opaque value objects** which is some kind of Python constant - the ProcessGroup and tensor inside are problematic, 
- there is PyTorch for dealing with custom tensor classes like Float8Tensor in torch.compile: **__tensor_flatten__()** and  **__tensor_unflatten__()** which assume that all internal tensors or **opaque reference objects** like process groups are directly parameters of a tensor. Currently, they are parameters of parameter (quantizer). I change that in this PR - amax reduction group is also stored on a QuantizedTensor when applicable.

Also, current design is prone to bugs. Amax reduction group is set in module forward(). So if we have forward for tp_group1, then forward for tp_group2, this second forwad overrides the amax_reduction_group of quantizers which are used in both backwards. So I think it is better design to set amax reduction group in forward and backward directly.
We may need to slightly refactor the quantizer to mitigate this kinds of bugs, but for now the change in this PR will be sufficient for torch.compile support.

Fixes # (N/A)

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- **TP sequence parallel:** set the amax reduction group on the input / grad-output
  quantizer *at point of use* in the fwd/bwd impls (`linear`, `layernorm_linear`,
  `layernorm_mlp`, `ops/basic_linear`) via a new
  `set_quantizer_amax_reduction_group` helper, instead of once at module setup.
  Removed the group wiring from `_customize_quantizers_float8_current_scaling`
  and dropped `_customize_quantizers_nvfp4` entirely.
- **FSDP2:** store the group on `Float8Tensor` / `NVFP4Tensor`
  (`amax_reduction_group` attribute, set in `fsdp_pre_all_gather`) and apply it
  to a throwaway quantizer copy during the in-place re-quant
  (`update_quantized` / `_set_data`) — the weight's own quantizer is never mutated.
- `Quantizer.quantize()` strips the group off the output tensor's quantizer, so
  it never persists on any tensor's `_quantizer`.
- No C++ changes.

# Checklist:


pggPL:
Blocked by FSDP bug, refactor in progress.

I plan to store .amax_reduction_group in QuantizedTensor.

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -271,6 +271,7 @@ def update_quantized(
         dst: QuantizedTensor,
         *,
         noop_flag: Optional[torch.Tensor] = None,
+        amax_reduction_group: Optional[Any] = None,  # pylint: disable=unused-argument
```
timmoon10: I strongly oppose this API change. amax reduction is very recipe-specific. It has different meanings for different recipes (FP8 DS might reduce over the TP+DP group, FP8 CS might only reduce over the TP group) and it has no meaning for other recipes (MXFP8 and FP8 block scaling). Moving it into the generic API will leak recipe-specific information, defeating the point of a generic API.

timmoon10 (dismissed):
This would be a design mistake. The amax reduction does not have a consistent meaning across recipes (including recipes where it doesn't make sense), and this change requires spilling out amax reduction logic into quantizer callsites (even where it doesn't make sense).

Can you go into more detail exactly why `torch.compile` doesn't work when quantizers have process groups? If we just want the quantizer to hold simple Python objects, maybe we can make the quantizer hold an `int` for the communicator ID. I envision something like:
```python
class Float8CurrentScalingQuantizer(Quantizer):

    _communicator_cache = {}

    @property
    def amax_reduction_group(self):
        if self._amax_reduction_group_id is None:
            return None
        return Float8CurrentScalingQuantizer._communicator_cache[self._amax_reduction_group_id]

    @property.setter
    def amax_reduction_group(self, comm):
        if comm is None:
            self._amax_reduction_group_id = None
        self._amax_reduction_group_id = id(comm)
        Float8CurrentScalingQuantizer._communicator_cache[self._amax_reduction_group_id] = comm
```
I'm not sure how this would interact with checkpointing though.

pggPL:
@timmoon10 I have changed the concept of this PR. Note that it also consist the changed from #3130

ksivaman (approved):
LGTM

pggPL:
Failures are unrelated

PR #3106, opened 2026-06-08
[Common] Reduce shared-memory bank conflicts in the colwise scaling path of the tuned NVFP4 kernel

Oleg-Goncharov:
# Description

This PR optimizes shared-memory accesses in the colwise scaling path by reducing bank conflicts, resulting in an average performance improvement of about 1%.

The colwise Y-lane mapping changed from:

`(thread_lane % 4 + warp) % 4`

image

to:

`(thread_lane / 2 + warp) % 4`

image

The intent is to change the shared-memory swizzle pattern used by the `sOut_tr` b64 stores. With the previous `%4` mapping, the bank pattern repeats every 4 lanes, which leads to an 8-way conflict pattern for this store layout. Using `/2` makes the mapping repeat every 8 lanes instead, reducing the conflict degree to 4-way for the same access pattern.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Modified the pattern of accessing shared memory

# Checklist:

### DIFF
--- a/transformer_engine/common/cast/nvfp4/specialized/quantize_transpose_nvfp4_tuned_1D.cuh
+++ b/transformer_engine/common/cast/nvfp4/specialized/quantize_transpose_nvfp4_tuned_1D.cuh
@@ -201,7 +201,7 @@ __device__ __forceinline__ void colwise_scaling(const IType *__restrict__ sIn_pt
   const int warp = threadIdx.x / THREADS_PER_WARP;
   const int thread_lane = threadIdx.x % THREADS_PER_WARP;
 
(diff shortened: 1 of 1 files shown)


ptrendx:
I agree that it would be good to provide some comment on the logic here, not necessarily in the code itself (that one would require a little bit more context on what you are trying to achieve here), but definitely in the PR description.

Oleg-Goncharov:
Good point, I expanded the PR description with the rationale and the updated mapping scheme.

ptrendx (approved):

PR #3108, opened 2026-06-09
[PyTorch] Add op-level activation offload opt-out API

lhb8125:
## Summary
Follow-up to #3047.

This PR adds an op-level activation offload policy for saved activation tensors so downstream fused grouped MLP users can opt individual TE ops out of activation offloading without changing the surrounding CPU offload context.

- add BasicOperation.set_activation_offloading(enabled)
- route activation offload marking through BasicOperation.maybe_mark_activation_offload
- use mark_not_offload for ops whose saved activations are opted out
- keep global CPU offload context checks and start_offload calls at their original call sites
- add unit coverage for the per-op mark/opt-out behavior

## Testing
- git diff --check
- python3.12 -m py_compile on the changed TE op files and tests/pytorch/test_fusible_ops.py

### DIFF
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -83,11 +83,6 @@ def test_basic_operation_activation_offloading_policy(monkeypatch):
 
     monkeypatch.setattr(
         cpu_offload,
-        "start_offload",
-        lambda *tensors: calls.append(("start", [id(t) for t in tensors])),
-    )
-    monkeypatch.setattr(
-        cpu_offload,
         "mark_activation_offload",
         lambda *tensors: calls.append(("mark", [id(t) for t in tensors])),
     )
@@ -97,18 +92,18 @@ def test_basic_operation_activation_offloading_policy(monkeypatch):
         lambda *tensors: calls.append(("skip", [id(t) for t in tensors])),
     )
 
-    op.maybe_mark_and_start_activation_offload(tensor, None, start=True)
-    assert calls == [("start", [tensor_id]), ("mark", [tensor_id])]
+    op.maybe_mark_activation_offload(tensor, None)
+    assert calls == [("mark", [tensor_id])]
 
     calls.clear()
     op.set_activation_offloading(False)
-    op.maybe_mark_and_start_activation_offload(tensor, start=True)
+    op.maybe_mark_activation_offload(tensor)
     assert calls == [("skip", [tensor_id])]
 
     calls.clear()
     op.set_activation_offloading(True)
-    op.maybe_mark_and_start_activation_offload(tensor, start=True, mark=False)
-    assert calls == [("start", [tensor_id])]
+    op.maybe_mark_activation_offload(tensor)
+    assert calls == [("mark", [tensor_id])]
 
 
 # Supported quantization recipes
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -107,8 +107,8 @@ def test_basic_operation_activation_offloading_policy(monkeypatch):
 
     calls.clear()
     op.enable_activation_offloading()
-    op.maybe_mark_and_start_activation_offload(tensor, start=True)
-    assert calls == [("start", [tensor_id]), ("mark", [tensor_id])]
+    op.maybe_mark_and_start_activation_offload(tensor, start=True, mark=False)
+    assert calls == [("start", [tensor_id])]
 
 
 # Supported quantization recipes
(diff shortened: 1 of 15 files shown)


transformer_engine/pytorch/ops/op.py
```diff
@@ -189,11 +189,51 @@ def __init__(self) -> None:
...
+
+        If activation offloading has been disabled for this op, mark the tensors so the
+        active offload context skips them. If mark is False, only start offload for tensors
+        that were already selected by the active offload context.
+        """
+        from ..cpu_offload import (  # pylint: disable=import-outside-toplevel
+            mark_activation_offload,
+            mark_not_offload,
+            start_offload,
+        )
```
ptrendx: Why can't you import it once (e.g. at the top of this file)? There is a non-zero CPU overhead from importing the already-imported module.

transformer_engine/pytorch/ops/op.py
```diff
@@ -189,11 +189,51 @@ def __init__(self) -> None:
...
+            return
+
+        if not self.activation_offloading:
+            mark_not_offload(*tensors)
+            return
+
+        if start:
+            start_offload(*tensors)
+        if mark:
+            mark_activation_offload(*tensors)
```
ptrendx: This would potentiall mark the tensors multiple times as all callsites are just leaving the default value here. Why do you combine these 2 functions rather than having 2 functions?

ptrendx (commented):
A general question about the motivation of this feature - I believe we already have a heuristics to not offload tensors which are too small, is that not enough?

CC @pggPL

lhb8125:
> A general question about the motivation of this feature - I believe we already have a heuristics to not offload tensors which are too small, is that not enough?
> 
> CC @pggPL

This is to support the selective offloading for Nemotron model training. If using fused MLP, MCore doesn't know which tensor is from expert_fc1 or moe_act or expert_fc2, so we need to expose an API to manually set offloading strategy for different ops. cc @timmoon10

transformer_engine/pytorch/ops/op.py
lhb8125: Good advice! We should leave `start_offload()` as it is.

transformer_engine/pytorch/ops/op.py
lhb8125: Nice catch!

ptrendx:
Ok, but does it actually matter to you that a specific tensor gets offloaded rather than getting the right amount of data to be offloaded?

xrennvidia:
> Ok, but does it actually matter to you that a specific tensor gets offloaded rather than getting the right amount of data to be offloaded?

Different activation tensors have different amount of data, selectively offloading activations is the way how we control the amount of data to be offloaded. If we offload too much, offloading latency can be exposed, if we offload too few, we get OOM. I think the essence of fine-grained offloading is to allow users to control the offloading of each tensor separately, the we can make the better perf tradeoffs.

ptrendx:
Well, but that could be controlled by a coarser mechanism. E.g. if we provided you with a way to say "offload up to X MBs per layer" and then choose the tensors out of all eligible in that layer wouldn't that also fulfill the same purpose, but potentially be easier to integrate/not require exposing the full control?

xrennvidia:
> Well, but that could be controlled by a coarser mechanism. E.g. if we provided you with a way to say "offload up to X MBs per layer" and then choose the tensors out of all eligible in that layer wouldn't that also fulfill the same purpose, but potentially be easier to integrate/not require exposing the full control?

I talked to @lhb8125 , he told me there is fraction config which controls the number of layers that do offloading. The layers that do offloading will offload all activations, other layers don't do offloading at all. And the method can have the following drawbacks:
1. Assuming there are 100 layers and the first 50 layers will offload 2x activation compared with baseline, if we reload the activations of 49th layer from the 100th layer's bprop and so on, the computing window is long enough but the reloaded tensors will accumulate and have a chance of running out of GPU memory.
2. If we reload the activations of 49th layer from the 50th layer's bprop and so on, the computing window will be only one layer, which may be not enough to overlap the reload of 2x activations.

Your idea is to config the number of bytes to offload per layer, it can fulfill the same purpose also. But what's the definition of a layer? Is it MoE, or the whole Transformer? This is ambiguous and could be confusing for users. I still do not know how we should clearly define this and expose this to users. I personally feel like using the module name like `expert_fc1`, `moe_act` is more straightforward.

tests/pytorch/test_fusible_ops.py
```diff
@@ -71,6 +71,42 @@
 # Supported devices
 _devices: list[torch.device] = [torch.device("cpu"), torch.device("cuda")]
 
+
+def test_basic_operation_activation_offloading_policy(monkeypatch):
```
timmoon10: Codex put this in such a weird and unmotivated place. No taste at all.

transformer_engine/pytorch/ops/op.py
timmoon10: I've found Claude sometimes does this kind of anti-pattern out of a misguided desire to minimize the impact of a code change. Human review is still important.

timmoon10 (approved):
I've made some changes so that we can use the same helper function to include and exclude tensors from offloading.

PR #3112, opened 2026-06-10
[JAX] Return max_logit from TE fused attn

### LINKED ISSUE #2945: Expose max_logit (and softmax aux statistics) from JAX fused attention higher-level APIs
## Summary
(coauthed by Claude)

The PyTorch fused attention bindings support an optional `return_max_logit` parameter that returns per-head maximum attention scores, but the JAX fused attention API (`transformer_engine.jax.attention.fused_attn`) has no equivalent mechanism. Additionally, the `softmax_aux` tensor (log-sum-exp statistics) computed during the forward pass is kept internal to the custom VJP machinery and never exposed to callers.

**Primary ask:** Expose `max_logit` (the per-head maximum attention score) through `transformer_engine.jax.attention.fused_attn`, matching the PyTorch API's `return_max_logit` parameter.

**Secondary ask:** Optionally expose `softmax_aux` (log-sum-exp statistics) for users who need them.

## Motivation

The `max_logit` value is already supported in the PyTorch frontend via `return_max_logit: bool` in `transformer_engine/pytorch/cpp_extensions/fused_attn.py`. Use cases include:

- Custom loss functions and reward signals that depend on attention statistics
- Numerical stability diagnostics and debugging
- Implementing custom backward passes outside of JAX's built-in autodiff
- Feature parity with the PyTorch frontend

## Current Behavior

**PyTorch** (supports `max_logit`):
In `transformer_engine/pytorch/cpp_extensions/fused_attn.py`, the forward function accepts `return_max_logit: bool` and returns `max_logit` as a separate output tensor when requested.

**JAX** (no `max_logit` support):
In `transformer_engine/jax/attention.py`, the public API `fused_attn` (line ~1394) returns only the output tensor. The internal forward rule `_fused_attn_fwd_rule` (line ~1272) calls `tex.fused_attn_fwd` which returns `(output, softmax_aux, rng_state)`, but only `output` is surfaced to the caller. The `softmax_aux` and `rng_state` are stored in the custom VJP context for the backward pass and discarded from the public return value.

```python
# Current JAX public API — no way to get aux stats
def fused_attn(qkv, bias, sequence_descriptor, seed, ...) -> jnp.ndarray:
    ...
    return output  # softmax_aux is only available internally for backward pass
```

## Expected Behavior

```python
from transformer_engine.jax.attention import fused_attn

# Primary: max_logit support
output, aux = fused_attn(q, k, v, ..., return_max_logit=True)
max_logit = aux["max_logit"]  # per-head max attention scores

# Secondary: softmax_aux (log-sum-exp stats)
output, aux = fused_attn(q, k, v, ..., return_aux=True)
softmax_aux = aux["softmax_aux"]  # shape [B, H, Sq, 1], float32 (log-sum-exp)
```

### 1. Add `return_max_logit` support to the JAX C++ extension layer

The underlying cuDNN kernels already support computing `max_logit` (as evidenced by PyTorch support). Wire this through `fused_attn_fwd` in `transformer_engine/jax/cpp_extensions/attention.py`.

### 2. Propagate through the custom VJP wrapper

Update `_fused_attn` and its forward/backward rules in `transformer_engine/jax/attention.py` to optionally return `max_logit` alongside `output`. Care is needed since changing the return signature affects `jax.custom_vjp` — one approach is to always compute `max_logit` when requested and pass it through the VJP context without requiring gradients for it.

### 3. Expose in the public `fused_attn` API

Add `return_max_logit: bool = False` (and optionally `return_softmax_aux: bool = False`) to the `fused_attn` signature at line ~1394. When enabled, return a tuple `(output, aux_dict)` instead of just `output`.

### 4. Current `fused_attn` signature for reference

```python
def fused_attn(
    qkv, bias, sequence_descriptor, seed,
    attn_bias_type, attn_mask_type, qkv_layout, softmax_type,
    scaling_factor, dropout_probability, is_training,
    max_segments_per_seq=1, window_size=None,
    context_parallel_strategy=CPStrategy.DEFAULT,
    context_parallel_causal_load_balanced=False,
    context_parallel_axis="", context_checkpoint_name="context",
    softmax_offset=None, stripe_size=None,
) -> jnp.ndarray
```

## Softmax Aux Details

The `softmax_aux` tensor in JAX contains log-sum-exp statistics with shape determined by cuDNN version:
- **cuDNN ≥ 9.6**: `[B, H, Sq, 1]` for BSHD layouts, `[B, Sq, H, 1]` for THD layouts
- **cuDNN < 9.6**: `[B, H, Sq, max_segments_per_seq]`
- **Always**: `float32`

Note: `softmax_aux` contains `log(Σ exp(x - max(x)))`, not the raw max logits. `max_logit` is a separate output tensor.

## References

- **PyTorch `return_max_logit` support:** [`transformer_engine/pytorch/cpp_extensions/fused_attn.py`](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/cpp_extensions/fused_attn.py)
- **JAX public API:** [`transformer_engine/jax/attention.py` `fused_attn` ~L1394](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/jax/attention.py)
- **JAX forward rule:** [`transformer_engine/jax/attention.py` `_fused_attn_fwd_rule` ~L1272](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/jax/attention.py)
- **JAX C++ extension forward:** [`transformer_engine/jax/cpp_extensions/attention.py` `fused_attn_fwd` ~L3359](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/jax/cpp_extensions/attention.py)

## Environment

- TransformerEngine version: main branch (HEAD)
- Framework: JAX

KshitijLakhani:
# Description

Return the max_logit from TE JAX fused attn to maintain parity with TE PyT fused attn.
Returning the softmax aux stats is a separate request and a bit mroe involved, so it'll be addressed separately. 
Exposing the max_logit gives sufficient telemetry insight to users, for starters.

Fixes #2945 

## Test timing added

-  Non-CP

```
   Level              Tests added    Approx. time added
  ━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━
   L0       41 (23 pass, 18 skip)                  ~23s
  ───────  ───────────────────────  ────────────────────
   L2       41 (23 pass, 18 skip)                  ~23s
```

 - CP
```

   Level                Tests added    Approx. time added
  ━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━
   L1       180 (74 pass, 106 skip)                 ~147s
  ───────  ─────────────────────────  ────────────────────
   L2         60 (16 pass, 44 skip)                  ~26s
```

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

# Checklist:

### DIFF
--- a/tests/jax/test_distributed_fused_attn.py
+++ b/tests/jax/test_distributed_fused_attn.py
@@ -522,7 +522,10 @@ class TestDistributedContextParallelSelfAttn:
             pytest.skip(f"Skipping {kv_groups=} not multiple of {data_shape=} or {tp_size=}")
 
         if return_max_logit:
-            runner.test_forward_with_max_logit(check_output=check_forward_output)
+            runner.test_forward(
+                return_max_logit=True,
+                check_output=check_forward_output,
+            )
         else:
             runner.test_backward()
         del os.environ["NVTE_FUSED_RING_ATTENTION_USE_SCAN"]
@@ -566,10 +569,12 @@ class TestDistributedContextParallelSelfAttn:
     ):
         """Check CP fused attention returns global per-head max_logit."""
         is_thd = qkv_layout.is_thd()
-        if window_size != (-1, -1) and not (
-            is_thd and cp_strategy == CPStrategy.RING and not use_scan_ring
-        ):
-            pytest.skip("SWA max-logit coverage is limited to THD Ring without scan.")
+        supports_swa = is_thd and (
+            cp_strategy == CPStrategy.ALL_GATHER
+            or (cp_strategy == CPStrategy.RING and not use_scan_ring)
+        )
+        if window_size != (-1, -1) and not supports_swa:
+            pytest.skip("CP SWA requires THD All-Gather or unrolled THD Ring.")
         # TODO: Evaluate cuDNN Max mismatches observed for striped multi-segment THD Ring GQA.
         if is_thd and cp_strategy == CPStrategy.RING and kv_groups > 1:
             pytest.skip("THD Ring GQA Max mismatches require further evaluation.")
--- a/tests/jax/test_distributed_fused_attn.py
+++ b/tests/jax/test_distributed_fused_attn.py
@@ -399,6 +399,75 @@ DISTRIBUTED_CONTEXT_SELF_ATTN_D256_LAYOUTS_MASKS_WINDOWS = [
     ),
 ]
 
+DISTRIBUTED_CONTEXT_SELF_ATTN_MAX_LOGIT_CASES = [
+    pytest.param(
+        QKVLayout.BSHD_BSHD_BSHD,
+        AttnMaskType.CAUSAL_MASK,
+        CPStrategy.ALL_GATHER,
+        (-1, -1),
+        None,
+        None,
+        False,
+        True,
+        id="AG-BSHD",
+    ),
+    pytest.param(
+        QKVLayout.THD_THD_THD,
+        AttnMaskType.PADDING_CAUSAL_MASK,
+        CPStrategy.ALL_GATHER,
+        (-1, -1),
+        64,
+        5,
+        False,
+        True,
+        id="AG-THD",
+    ),
+    pytest.param(
+        QKVLayout.BSHD_BSHD_BSHD,
+        AttnMaskType.CAUSAL_MASK,
+        CPStrategy.RING,
+        (-1, -1),
+        None,
+        None,
+        False,
+        True,
+        id="RING-BSHD-NO_SCAN",
+    ),
+    pytest.param(
+        QKVLayout.BSHD_BSHD_BSHD,
+        AttnMaskType.CAUSAL_MASK,
+        CPStrategy.RING,
+        (-1, -1),
+        None,
+        None,
+        True,
+        True,
+        id="RING-BSHD-SCAN",
+    ),
+    pytest.param(
+        QKVLayout.THD_THD_THD,
+        AttnMaskType.PADDING_CAUSAL_MASK,
+        CPStrategy.RING,
+        (-1, -1),
+        1,
+        5,
+        False,
+        False,
+        id="RING-THD-NO_SCAN",
+    ),
+    pytest.param(
+        QKVLayout.THD_THD_THD,
+        AttnMaskType.PADDING_CAUSAL_MASK,
+        CPStrategy.RING,
+        (-1, -1),
+        1,
+        5,
+        True,
+        False,
+        id="RING-THD-SCAN",
+    ),
+]
+
 
 class TestDistributedContextParallelSelfAttn:
     # TODO(KshitijLakhani): parametrize num_segments_per_seq for all CP tests
@@ -419,6 +488,8 @@ class TestDistributedContextParallelSelfAttn:
         window_size=None,
         stripe_size=None,
         num_segments_per_seq=None,
+        return_max_logit=False,
+        check_forward_output=True,
     ):
         if qkv_layout.is_thd():
             if not load_balanced and (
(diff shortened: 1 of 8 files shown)


transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -48,8 +48,8 @@ void PrepareFusedAttnForwardAuxTensors(NVTETensorPack *tensor_pack, const size_t
                                        const size_t bias_heads, const size_t q_max_seqlen,
                                        const size_t kv_max_seqlen, DType dtype,
                                        NVTE_Bias_Type bias_type, NVTE_Fused_Attn_Backend backend,
-                                       void *softmax_buf, void *rng_state_buf = nullptr,
-                                       void *bias_buf = nullptr,
+                                       void *softmax_buf, void *max_buf = nullptr,
```
jberchtold-nvidia: Can we rename `max_buf` to `max_logits_buf`

transformer_engine/jax/csrc/extensions/attention.cpp
```diff
@@ -65,8 +65,23 @@ void PrepareFusedAttnForwardAuxTensors(NVTETensorPack *tensor_pack, const size_t
...
+      max_aux_data.shape.ndim = 4;
+      max_aux_data.shape.data[0] = input_batch;
+      max_aux_data.shape.data[1] = attn_heads;
+      max_aux_data.shape.data[2] = q_max_seqlen;
+      max_aux_data.shape.data[3] = 1;
+      max_aux_data.dtype = static_cast<NVTEDType>(DType::kFloat32);
+      nvte_set_tensor_param(&max_aux, kNVTERowwiseData, &max_aux_data);
+    }
+
+    NVTETensor &rng_state_aux = tensor_pack->tensors[size++];
```
jberchtold-nvidia: Do we need to check the capacity of `tensor_pack->tensors` here?

transformer_engine/jax/cpp_extensions/attention.py
```diff
@@ -2690,14 +2849,17 @@ def half_q_no_mask_compute():
...
-                    return output_per_step, softmax_aux_per_step
+                    return output_per_step, softmax_aux_per_step, max_logit_per_step
 
                 def skip_compute():
                     output_per_step = jnp.zeros_like(q)
                     softmax_aux_per_step = jnp.full(
                         (batch, head, q.shape[1], 1), -jnp.inf, dtype=jnp.float32
                     )
-                    return output_per_step, softmax_aux_per_step
+                    max_logit_per_step = FusedAttnFwdPrimitive._empty_or_neg_inf_max_logit(
```
jberchtold-nvidia, line 2870: Why does this instance use empty max logits?

tests/jax/test_distributed_fused_attn.py
```diff
@@ -399,6 +399,12 @@ def test_softcap_score_mod_with_aux_params_backward(
     ),
 ]
```
jberchtold-nvidia, line 401: Can you run L1 CI on this PR? I don't see a CI run

transformer_engine/jax/cpp_extensions/attention.py
KshitijLakhani: That's a good question @jberchtold-nvidia  !
**TL;DR:** So, this is the no-compute branch for a fully causal masked Ring step and hence attention, softmax, isn't calculated in this branch and we need a neutral placeholder (-inf/0) because veery lax.cond branch must return the same shapes . 

**Let me try expanding on that:** 
With causal Ring attention, every rank cycles through KV blocks from all CP ranks. Some received KV blocks are entirely in the future relative to that rank’s query block and because causal attention must not attend to any token in such a block, TE skips the cuDNN attention call completely. So `skip_compute` does not run attention, softmax, or max-logit computation. It returns identity values so the JAX control flow and subsequent accumulation remain valid.

When max logits are requested, the helper returns [H] filled with -inf, so the skipped step cannot affect the running maximum. FWIW, it returns (0,) only when max-logit return is disabled.

tests/jax/test_distributed_fused_attn.py
KshitijLakhani: I had run the test suite locally only. Launched CI right now. Thanks for the reminder !

transformer_engine/jax/csrc/extensions/attention.cpp
KshitijLakhani: I was not too concerned about it here, because I think the max size is 10 and with max_logit we will have 5 tensors only: 1 softmax + 1 max + 1 RNG + 1 bias + 1 offset = 5 maximum so I wasn't concerned about exceeding.

However, that's a good point for the future when someone is packing in more tensors in here. Let me push a change not only for max_logit but also for the other 4 tensors packed in sothat it serves as a template for any tensors being packed in the future

transformer_engine/jax/csrc/extensions/attention.cpp
KshitijLakhani: Addressed in SHA 704957afba4d7e432d6332fe0c7e2f4b5bc382fd

transformer_engine/jax/csrc/extensions/attention.cpp
KshitijLakhani: Addressed in SHA 2e71bc8aefe7e3fa2ec02ae7609b075f0f93fe53

tests/jax/test_distributed_fused_attn.py
KshitijLakhani: CI looks good for max_logit tests. Only failures are outside of attention: 64353613 (TestDistributedLayernorm and distributed_test_base)
I verified the max_logit tests in the CI and they pass / skip as expected

jberchtold-nvidia (approved):
LGTM, thanks

cyanguwa:
@KshitijLakhani, do you mind just updating the PR description a bit to reflect the support we added for max_logit, e.g. non-CP + BSHD/SBHD/THD, CP + BSHD/SBHD/THD + Ring/AG?

PR #3107, opened 2026-06-09
Add entrypoint for flagos multi-backend plugin system

lxd-cumt:
# FlagOS Proposal: Plugin Architecture & Device-Agnostic Abstraction for TransformerEngine

##  Plugin System: Initialization-time Backend Loading

We propose a plugin architecture where TransformerEngine (TE) loads backend implementations at initialization time via an explicit plugin interface, while the actual multi-backend plugins reside in a separate repository (`TransformerEngine-Plugin-FL`).

TransformerEngine-Plugin-FL: https://github.com/lxd-cumt/TransformerEngine-Plugin-FL
### Current State

TE already has a prototype (`NVTE_ENABLE_PLUGIN=1` in `common/__init__.py`) that registers the original CUDA pybind module as `transformer_engine_torch_nv` and delegates to an external `load_plugins()` entry point.

### Proposed Design

- TE defines a stable plugin API contract (operator signatures, quantization interfaces, communication primitives).
- At `load_framework_extension()` time, if a plugin is present, TE dispatches backend calls through the plugin registry; otherwise it falls back to the native te implementation.
- The `TransformerEngine-Plugin-FL` repository is independently installable and contains multiple backend implementations for  diverse accelerators, and support more training scenarios.

PR #3113, opened 2026-06-10
Abstract CUDA hardcodes into configurable te_device_type / te_platform

lxd-cumt:
# FlagOS Proposal: Plugin Architecture & Device-Agnostic Abstraction for TransformerEngine

## Device-Type Abstraction: Replacing Hardcoded `"cuda"` References

The current TE PyTorch layer contains ~100 hardcoded `"cuda"` string literals and ~165 `torch.cuda.*` API calls. These span device placement (`device="cuda"`), autocast context (`device_type="cuda"`), device-type guards (`device.type == "cuda"`), and RNG state management (`torch.cuda.CUDAGraph`, `torch.cuda._lazy_call`). This makes TE non-functional on alternative accelerator platforms without invasive patching.

### Proposed Design

  1. **Soft abstraction** – A global `te_device_type()` / `te_platform()` accessor replaces ~200 literal `"cuda"` strings across the Python codebase.

  2. **Platform monkey-patch** – A vendor-provided `apply_patch()` hook runs at import time to directly remap `torch.cuda.*` APIs (e.g. `torch.cuda.device`, `torch.cuda.current_device`, `torch.cuda.current_stream`) to the vendor equivalents (e.g. `torch.other_vendor.*`).


lxd-cumt:
Thanks for the review! I've addressed the above Greptile comments.

lxd-cumt:
Should this PR be merged into the ```main``` branch, so it can follow future NVIDIA TE releases?

transformer_engine/__init__.py
```diff
@@ -10,6 +10,43 @@
 from importlib import metadata
 import transformer_engine.common
 
+import torch
```
ptrendx: We cannot import torch in this file - it is used both by pyTorch and JAX backends.

ptrendx: It should go into `transformer_engine/pytorch/__init__.py` instead.

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
```diff
@@ -1191,6 +1191,9 @@ def cp_p2p_bwd_flash_attn(
     dq, dk, dv = [torch.empty_like(x) for x in [q_part, k_part, v_part]]
     if fa_utils.v2_3_plus and not fa_utils.v2_7_0_plus:
         fa_backward_kwargs["window_size"] = (-1, -1)
+        # Fix: flash-attn 2.3.x ~ 2.6.x also needs rng_state for dropout
+        if not use_flash_attn_3 and rng_states is not None:
+            fa_backward_kwargs["rng_state"] = rng_states[cp_size - step - 1]
```
ptrendx: This change looks not connected to the rest of the PR - is it intended?

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -539,6 +540,9 @@ def get_attention_backend(
         if use_flash_attention:
             use_flash_attention = False
             logger.debug("Disabling FlashAttention for max_logit")
+        if use_fused_attention and qkv_format == "thd":
```
ptrendx: Changes in this file also seem disconnected from the rest of this PR and should be in its own PR.

transformer_engine/pytorch/module/linear.py
```diff
@@ -75,6 +75,7 @@
 )
 from ...debug.pytorch.debug_state import TEDebugState
 
+
```
ptrendx: Unnecessary change.

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -213,9 +215,11 @@ def make_empty(
         pin_memory: bool = False,
     ) -> Float8BlockwiseQTensor:
         """Construct quantized tensor with uninitialized data"""
+        if device is None:
+            device = torch.device(te_device_type())
```
ptrendx: This is already handled by the lines below.

ptrendx (commented):
Left some comments. Could we use this PR also as an opportunity to take a look at each of the cases where we hardcoded the device type and see if we can instead use a device from e.g. the input tensor instead?

transformer_engine/__init__.py
lxd-cumt: fixed, thanks

transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py
lxd-cumt: fixed

transformer_engine/pytorch/attention/dot_product_attention/utils.py
lxd-cumt: fixed

transformer_engine/pytorch/module/linear.py
lxd-cumt: fixed

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
lxd-cumt: fixed

lxd-cumt:
> Left some comments. Could we use this PR also as an opportunity to take a look at each of the cases where we hardcoded the device type and see if we can instead use a device from e.g. the input tensor instead?

Thanks for the review, I have addressed the comments.
This PR touches similar hardcoded device types in TransformerEngine, so we can align on a technical approach together. And if we plan to upstream this to main eventually, we'll need to handle cuda hardcoding across more files as well.

transformer_engine/__init__.py
```diff
@@ -10,6 +10,26 @@
...
+        _patches = import_module(f"{_nvte_plugin}.patches")
+        _patches.apply_patches()
+    except Exception as e:
+        import warnings
+
+        warnings.warn(
+            f"NVTE_PLUGIN={_nvte_plugin} but plugin patch apply failed: {e}",
+            RuntimeWarning,
+            stacklevel=1,
+        )
```
ptrendx, line 30: Is this needed in addition to #3107 or should it be removed once that PR is merged?

transformer_engine/pytorch/__init__.py
```diff
@@ -9,13 +9,48 @@
...
+if not hasattr(transformer_engine, "TE_DEVICE_TYPE"):
+    transformer_engine.TE_DEVICE_TYPE = "cuda"
+TE_DEVICE_TYPE = transformer_engine.TE_DEVICE_TYPE
+
+
+def te_device_type(default: str = "cuda") -> str:
+    try:
+        return transformer_engine.TE_DEVICE_TYPE
+    except Exception:
+        return default
```
ptrendx: Considering that we set this value just above, why do we need try/except here? And if there was some
chance that this value got deleted somehow, then it should probably be an error anyway rather than
a quiet ignore.

transformer_engine/pytorch/__init__.py
```diff
@@ -9,13 +9,48 @@
...
+if not hasattr(transformer_engine, "TE_PLATFORM"):
+    transformer_engine.TE_PLATFORM = torch.cuda
+TE_PLATFORM = transformer_engine.TE_PLATFORM
+
+
+def te_platform(default=torch.cuda):
+    try:
+        return transformer_engine.TE_PLATFORM
+    except Exception:
+        return default
```
ptrendx: Same comment here.

transformer_engine/pytorch/__init__.py
```diff
@@ -9,13 +9,48 @@
...
 assert torch_version() >= (2, 1), f"Minimum torch version 2.1 required. Found {torch_version()}."
 
 load_framework_extension("torch")
+
+# Device type and platform globals — live here because torch cannot be imported
+# at the top-level transformer_engine package (shared with JAX).
+# Only set defaults if the plugin hasn't already patched them.
+if not hasattr(transformer_engine, "TE_DEVICE_TYPE"):
+    transformer_engine.TE_DEVICE_TYPE = "cuda"
+TE_DEVICE_TYPE = transformer_engine.TE_DEVICE_TYPE
```
ptrendx: Actually I'm not sure why this code lives here. We should be able to just set TE_DEVICE_TYPE (and
platform) inside `transformer_engine/__init__.py` before loading of the patches. Also, do we actually
need this very generic patches interface or can we just set those values directly (plugin would just

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -1103,6 +1104,10 @@ def _is_fa3_supported(num_heads, num_gqa_groups, head_dim_qk, head_dim_v, qkv_dt
             logger.debug("Disabling FusedAttention for determinism reasons with post_scale_bias")
             use_fused_attention = False
             fused_attention_backend = None
+        if is_training and device_compute_capability >= (10, 0):
+            logger.debug("Disabling FusedAttention for determinism reasons on Blackwell")
+            use_fused_attention = False
+            fused_attention_backend = None
```
ptrendx: This looks unrelated.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -1501,7 +1508,9 @@ def get_alibi(
         _alibi_cache["_max_seqlen_q"], _alibi_cache["_max_seqlen_kv"] = max_seqlen_q, max_seqlen_kv
         _alibi_cache["_bottom_right_alignment"] = bottom_right_alignment
         bias_dtype = torch.float32 if bias_dtype is None else bias_dtype
-        _alibi_cache["_alibi_bias"] = bias.contiguous().to(dtype=bias_dtype, device="cuda")
+        _alibi_cache["_alibi_bias"] = bias.contiguous().to(
+            dtype=bias_dtype, device=te_device_type()
+        )
```
ptrendx: Here the bias should already be on the device, so probably we could just omit the device parameter
here entirely.

transformer_engine/pytorch/triton/permutation.py
```diff
@@ -49,10 +51,12 @@ def make_row_id_map(
         The [num_experts, num_experts + n_routed) items are the indices of the experts corresponding
         to the first n_routed row indices above.
     """
-    row_id_map = torch.empty((num_tokens, num_experts * 2 + 1), dtype=torch.int32, device="cuda")
+    row_id_map = torch.empty(
+        (num_tokens, num_experts * 2 + 1), dtype=torch.int32, device=te_device_type()
```
ptrendx: Use routing_map's device.

transformer_engine/pytorch/utils.py
```diff
@@ -549,12 +552,12 @@ def canonicalize_device(device: Optional[torch.device | str]) -> torch.device:
     if device is None:
         # Use default CUDA device
         device = torch.get_default_device()
-        if device.type != "cuda":
-            device = torch.device("cuda", torch.cuda.current_device())
+        if device.type != te_device_type():
+            device = torch.device(te_device_type(),te_platform().current_device())
```
ptrendx: Missing space after comma.

transformer_engine/__init__.py
lxd-cumt: This logic should be kept, as it is intended to replace ```torch.cuda```-related functionality with platform-specific APIs via patching. The previous PR only replaced the ```transformer_engine_torch``` package.
I think they should be merged separately.

lxd-cumt: Previous pr https://github.com/NVIDIA/TransformerEngine/pull/3401 replaces ops within ```transformer_engine_torch```, whereas this patch targets ```torch.cuda-related``` APIs repo-wide. Despite the similar plugin load mechanism, they address orthogonal concerns. I suggest retaining this.

transformer_engine/pytorch/__init__.py
lxd-cumt: `TE_PLATFORM` defaults to `torch.cuda`, which requires torch to be importable. The top-level `transformer_engine/__init__.py` is shared with
  JAX and does not import torch, so both `TE_DEVICE_TYPE` and `TE_PLATFORM` defaults are set here to keep them co-located. The functions
  `te_device_type()` and `te_platform()` are then registered onto the top-level package for use by other submodules.

  Regarding the generic patches interface: `apply_patches()` does more than setting these two values — it also monkey-patches `torch.cuda.*`
  APIs (e.g. `current_device`, `synchronize`, `get_device_properties`) for non-CUDA backends. A direct-set approach would not cover those.

transformer_engine/pytorch/__init__.py
lxd-cumt: fixed

transformer_engine/pytorch/__init__.py
lxd-cumt: fixed

transformer_engine/pytorch/attention/dot_product_attention/utils.py
lxd-cumt: deleted

transformer_engine/pytorch/attention/dot_product_attention/utils.py
lxd-cumt: fixed

transformer_engine/pytorch/triton/permutation.py
lxd-cumt: fixed

transformer_engine/pytorch/utils.py
lxd-cumt: fixed

lxd-cumt:
A few questions for discussion:

  1. **Target branch**: Should this PR target `main` instead of `release/v2.14`?

  2. **`te_device_type` / `te_platform` abstraction**: Is this approach acceptable? This implies that future development on `main` would need to use `te_device_type()` / `te_platform()` in place of hardcoded `"cuda"` hardstrings.

  3. **Megatron-LM**: If this approach is agreed upon, I plan to open a similar PR for Megatron-LM to abstract device-specific references there as well.

lxd-cumt:
> A few questions for discussion:
> 
> 1. **Target branch**: Should this PR target `main` instead of `release/v2.14`?
> 2. **`te_device_type` / `te_platform` abstraction**: Is this approach acceptable? This implies that future development on `main` would need to use `te_device_type()` / `te_platform()` in place of hardcoded `"cuda"` hardstrings.
> 3. **Megatron-LM**: If this approach is agreed upon, I plan to open a similar PR for Megatron-LM to abstract device-specific references there as well.

A new pull request (https://github.com/NVIDIA/TransformerEngine/pull/3513) has been submitted against the ```main``` branch and is ready for review.

PR #3116, opened 2026-06-10
TE EP integration to MoEBlock

tdophung:
# Description

Integrate the new TE EP dispatch + combine APIs in the MoEBlock, in replacement for the previous ragged-all-to-all communications and Triton permutation kernels. Router CUDA kernels + grouped GEMM is the same as before. 

Will rebase and squash the commits on this branch once about to merge
Will also change the JAX APIs if needed when TE EP JAX merge

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Remove previous multiprocess test script for MoE VJP in general
- Add new multiprocess test script for MoEBlock x TE EP
- Integrate TE EP dispatch + combine APIs into MoeBlock

# Checklist:


transformer_engine/jax/flax/moe.py
```diff
@@ -143,16 +119,27 @@ class _MoEBlock(TransformerEngineBase):
...
 
     # Dtypes / init / misc
     dtype: DType = jnp.float32
     kernel_init: Optional[Initializer] = None
     bias_init: Initializer = nn.initializers.zeros
-    expert_bias_init: Initializer = nn.initializers.zeros
     use_bias: bool = False
+    # Per-expert router bias added before the top-k. Only meaningful when
+    # score_function='sigmoid'.
+    use_expert_bias: bool = False
```
jberchtold-nvidia: nit SR: can we rename `use_bias` -> `use_ffn_bias` and `use_expert_bias` -> `use_expert_routing_bias`?

transformer_engine/jax/flax/moe.py
```diff
@@ -143,16 +119,27 @@ class _MoEBlock(TransformerEngineBase):
...
+    # the gate.
+    aux_loss_coeff: float = 0.0
+
+    # Fusion knob
+    apply_topk_weights_early: bool = False
+
+    # Minimum per-expert slot alignment fed to ``tex.ep_prepare``. Default 0
+    # uses the natural slot count; set to e.g. 128 to satisfy FP8 grouped-GEMM
+    # tile alignment.
+    align_size: int = 0
```
jberchtold-nvidia: Placeholder comment for me to fix this so align_size is inferred automatically based on the recipe and doesn't need to be specified by the user

transformer_engine/jax/flax/moe.py
```diff
@@ -240,13 +221,14 @@ def __call__(self, inputs: Array) -> Tuple[Array, Optional[Array]]:
...
+
         expert_bias = None
         if self.use_expert_bias:
             expert_bias = self.param(
                 "expert_bias",
-                nn.with_logical_partitioning(self.expert_bias_init, ("exp",)),
+                nn.with_logical_partitioning(self.bias_init, ("exp",)),
                 (self.num_experts,),
-                self.dtype,
+                jnp.float32,
```
jberchtold-nvidia, line 249: Is the router always in fp32 so this expert bias must also be? If so, can we add a small comment indicating this

transformer_engine/jax/moe.py
```diff
@@ -79,1070 +55,315 @@
...
-
-PRNGKey = Any
-Shape = Tuple[int, ...]
-DType = NewType("DType", jnp.dtype)
-Array = NewType("Array", jnp.ndarray)
+__all__ = ["moe"]
 
 
-__all__ = ["moe", "PermutationBackend"]
+def _with_sharding_constraint_cast_bwd(x: jnp.ndarray, sharding) -> jnp.ndarray:
```
jberchtold-nvidia, line 68: Why do we need this utility function? I haven't seen something like this required for our other VJPs

transformer_engine/jax/moe.py
```diff
@@ -79,1070 +55,315 @@
...
-            out_off_r,
-            recv_sz_r,
-            axis_name=ep_axis,
-        )
+# Registered as a pytree so jax.custom_vjp can flatten/unflatten it across
+# the fwd -> bwd boundary. ``cfg`` is the only static field (EpLayerConfig
+# is a frozen dataclass of ints); the rest are jnp.ndarray,
+# GroupedNoScaleTensor (already a pytree), or None when aux_loss_coeff == 0.
+@register_pytree_node_class
+@dataclass
```
jberchtold-nvidia: I think this tree_flatten was from my patch, but looking at the diff I think it'd be better to use the `@flax_struct.dataclass` you were using on the permutation dataclasses since that seems to auto-populate a default pytree flatten/unflatten for us

transformer_engine/jax/moe.py
```diff
@@ -1151,524 +372,135 @@ def _body_fwd(  # pylint: disable=unused-argument
...
+        mask_b = (recv_w_flat != 0).astype(d_intermediate.dtype)[:, None]
+        intermediate_unweighted = act_fn(gate_proj_out) * up_proj_out
+        d_recv_w_from_intermediate = jnp.sum(
+            d_intermediate * intermediate_unweighted * mask_b, axis=-1
+        ).astype(recv_w_flat.dtype)
+        d_intermediate = d_intermediate * w_b * mask_b
+    else:
+        d_recv_w_from_intermediate = jnp.zeros_like(recv_w_flat)
+
+    # Activation bwd. Mirror the fwd's fp32 promotion of silu+multiply
```
jberchtold-nvidia: Why is this dtype casting required? I don't recall us needing it for the non-MoE LNMLP block

transformer_engine/jax/moe.py
```diff
@@ -1687,170 +519,342 @@ def _moe_fwd_rule(  # pylint: disable=unused-argument
...
+    # buffer as ``[num_local_experts, num_ep * max_tokens_per_rank, hidden]``
+    # (see nccl_ep.cc::init kernel buffer sizing + the LL combine assertion
+    # at nccl_ep.cc:2185 which spells out the same layout). The natural
+    # dropless K-expanded count
+    # ``ceil((B/dp)*S*K / num_local_experts)`` does NOT match: it ignores
+    # the worst-case where all of one EP group's tokens land on a single
+    # local expert. We must size to that worst case or NCCL EP's HT kernel
+    # rejects the dispatch buffer with ``invalid argument``.
+    natural_spe = num_ep * max_tokens_per_rank  # = (B // dp_size) * S
+    # NCCL EP requires each expert-major output block to be at least
```
jberchtold-nvidia: Do we have a use-case for user-specified alignments beyond 128 currently? If NCCL EP requires an alignment of at least 128, and since an alignment of 128 is sufficient for all TE grouped GEMM types, would it make sense to instead hardcode `_ALIGN_SIZE = 128` as a constant at the top of the file for now to simplify this MoEBlock API.

We can always expand the API to support a user-specified align size in the future

transformer_engine/jax/moe.py
```diff
@@ -1687,170 +519,342 @@ def _moe_fwd_rule(  # pylint: disable=unused-argument
...
-    dp_size = 1
-    for ax in data_parallelism_axes:
-        dp_size *= mesh.shape[ax]
+        # ep must be innermost: ep_bootstrap forms NCCL EP comms from
+        # consecutive global ranks (dp_color = rank // ep_size), so the
+        # comm only stays within one model replica under (outer_dp, ep).
+        batch_pspec_axis = (*data_parallelism_axes, ep_axis)
+    ep3_spec = P(batch_pspec_axis, None, None)
+    ep2_spec = P(batch_pspec_axis, None)
+    x = jax.lax.with_sharding_constraint(x, NamedSharding(mesh, ep3_spec))
```
jberchtold-nvidia, line 541: Which axis name inputs are physical mesh axes and why can be logical axes? I see above `x = with_sharding_constraint_by_logical_axes(x, input_axes)` but here we directly use `jax.lax.with_sharding_constraint` which only supports mesh axes.

No need to make any changes for now, I just want to assess which are which and then we can discuss if it makes sense to support logical on some/all or if some are required to be physical axes. Thanks!

transformer_engine/jax/moe.py
```diff
@@ -1687,170 +519,342 @@ def _moe_fwd_rule(  # pylint: disable=unused-argument
...
+    # Sigmoid + K>1 normalises as `weights / (weights.sum + 1e-20)`; for
+    # tokens whose top-K sigmoid scores all underflow at bf16/fp32 the
+    # output is NaN at the selected positions. Those NaNs ride
+    # ep_dispatch -> recv_topk_weights -> combine and poison the per-token
+    # weighted sum, leaving entire output rows as NaN. Sanitize at the
+    # source so neither the fwd combine nor the bwd's manual
+    # `grad_pre_combine * w` sees them. Padded positions in sparse_probs
+    # are already zero (routing_map is False there); only the rare
+    # underflow path emits NaN.
+    sparse_probs = jnp.where(jnp.isnan(sparse_probs), 0, sparse_probs).astype(dtype)
```
jberchtold-nvidia: Is this NaN filtering a debugging artifact or something we need in the final version?

transformer_engine/jax/moe.py
tdophung: for MoE particularly, In our `_moe_bwd_rule`, `d_x` is built from two cotangent paths:

`d_x_from_dispatch = tex.ep_dispatch_bwd(...)` in bf16 if `x` is in bf16.
and
`d_x_from_gate = d_logits_2d @ gate_kernel^T`, where `d_logits_2d` comes from `tex.fused_topk_with_score_function_bwd`, which only ooutput the logits in fp32 (per alp in one of the weekly meetings + I checked the code for the router kernel). So `d_x_from_gate` is fp32.

Therefore, `d_x = d_x_from_dispatch + d_x_from_gate = bf16 + fp32 = fp32`, while our `x` could be bf16. So the cotangent that flows to bwd needs to be constrainted. Other VJPs don't have this issue because the dgrad will just be the same dtype as the activation dtype.

transformer_engine/jax/flax/moe.py
tdophung: yes I will add a comment

transformer_engine/jax/moe.py
tdophung: debugging artifact. Remopving

transformer_engine/jax/moe.py
tdophung: Yes, after confirming with Tim and Przemek, we should kjeep activation input in BF16 and activation output in GEMM's dtype. In latest commit, this should be BF16

transformer_engine/jax/moe.py
tdophung: added some comments

transformer_engine/jax/moe.py
tdophung: added in latest commits [f682d3f](https://github.com/NVIDIA/TransformerEngine/pull/3116/commits/f682d3fdc2042c4f1155d359b3ba8cb20b096610)

transformer_engine/jax/moe.py
```diff
@@ -1687,170 +541,337 @@ def _moe_fwd_rule(  # pylint: disable=unused-argument
...
+        # propagates into the combine output when the kernel's read
+        # pattern overlaps the padded region. Use jnp.where to overwrite
+        # padded positions with a literal 0 before combine.
+        w = recv_topk_weights[..., None].astype(expert_outputs.dtype)
+        mask_bool = (recv_topk_weights != 0)[..., None]
+        weighted = jnp.where(mask_bool, expert_outputs * w, jnp.zeros_like(expert_outputs))
+        output = tex.ep_combine_fwd(
+            cfg,
+            handle_mem,
+            weighted,
```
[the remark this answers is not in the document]
tdophung: I forgot to remove the sanitation here. in theory there should not be any NaNs to be consumed by downstream ops of this, as we trust the correctness of the dispatch and combine APIs. If there are NaNs here, we would like for it to surface, not hidden with sanitation.

Removed the sanitation in bwd in 893f3ff

transformer_engine/jax/ep.py
```diff
@@ -0,0 +1,311 @@
...
+    outer_axis = _ep_outer_axis()
+    if outer_axis is None:
+        if world_size != ep_size:
+            raise ValueError(
+                f"ep_bootstrap: world_size ({world_size}) > ep_size ({ep_size}) but neither"
+                " MeshResource.dp_resource nor fsdp_resource is set; name the outer axis so"
+                " EP-output tensors can shard across EP groups."
+            )
+        num_ep_groups = 1
+    else:
```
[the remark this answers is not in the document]
tdophung: All ep.py and ep_backend.cpp comments will be handled by Phuong in her PR: https://github.com/NVIDIA/TransformerEngine/pull/3036 
After that is merged, I will rebase this PR on top of it. For now, resolving all grep tilee comments on them

transformer_engine/jax/flax/moe.py
```diff
@@ -50,7 +49,7 @@
 Initializer = Callable[[PRNGKey, Shape, DType], Array]
 
 
-__all__ = ["PermutationBackend", "_MoEBlock"]
+__all__ = ["_MoEBlock"]
```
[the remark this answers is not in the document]
tdophung: This will be noted in release notes if needed.

tests/jax/test_te_ep_moe.py
```diff
@@ -0,0 +1,732 @@
...
+* No dedicated "Flax wrapper init+apply" smoke test: every config above
+  already calls ``MoEBlock`` (the Flax wrapper) end-to-end, so a
+  separate wrapper smoke would just duplicate ``test_forward[softmax]``
+  + ``test_backward[softmax]``.
+* No re-bootstrap-mismatch test: ``ep_bootstrap`` rejects a mismatched
+  signature unconditionally and is a one-line guard; covering it from
+  this suite would taint the per-process NCCL bootstrap cache for the
+  rest of the file with no real upside.
+
+FP8 / MXFP8 recipes are deferred — the ``quantizer_sets`` plumbing
```
jberchtold-nvidia: This seems agent-specific and we don't need this last paragraph in here

tests/jax/test_te_ep_moe.py
```diff
@@ -0,0 +1,732 @@
...
+            )
+            return jnp.mean(out.astype(jnp.float32) ** 2)
+
+        grads_ref = jax.jit(jax.grad(loss_fn))(
+            {k: jnp.asarray(v) for k, v in params_np.items() if k != "expert_bias"},
+            jnp.asarray(x_np),
+        )
+        grads_ref_np = {k: np.asarray(jax.device_get(v)) for k, v in grads_ref.items()}
+
+        for name in ("gate_kernel", "wi_0", "wi_1", "wo"):
```
jberchtold-nvidia, line 641: This checks the weight gradients but we also need to check the final propagated gradient that will be passed onto the next layer in backprop.

tdophung: added in [27fb270](https://github.com/NVIDIA/TransformerEngine/pull/3116/commits/27fb2709b7c43dd93c6f2a493bead630af191476)

transformer_engine/jax/moe.py
```diff
@@ -1151,524 +309,143 @@ def _body_fwd(  # pylint: disable=unused-argument
...
+        # across both factors before the activation bwd consumes it.
+        # Cast w_b so the multiply stays in d_intermediate.dtype and
+        # d_sorted_x (downstream into ep_dispatch_bwd) stays bf16.
+        w_b = recv_w_flat[:, None].astype(d_intermediate.dtype)
+        mask_b = (recv_w_flat != 0).astype(d_intermediate.dtype)[:, None]
+        intermediate_unweighted = act_fn(gate_proj_out) * up_proj_out
+        d_recv_w_from_intermediate = jnp.sum(
+            d_intermediate * intermediate_unweighted * mask_b, axis=-1
+        ).astype(recv_w_flat.dtype)
+        d_intermediate = d_intermediate * w_b * mask_b
```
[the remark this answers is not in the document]
tdophung: should be fixed in [c2869ed](https://github.com/NVIDIA/TransformerEngine/pull/3116/commits/c2869ede04fea8c58613c182f3b5c79da5c363c2)

transformer_engine/jax/moe.py
```diff
@@ -79,1070 +54,253 @@
...
+        # Fold the per-token combine weights into the FFN intermediate;
+        # the downstream wo GEMM is linear so this is equivalent to the
+        # late-weighting path, modulo elementwise op fusion gains. w_b is
+        # cast to intermediate.dtype so the multiply doesn't promote
+        # expert_outputs above the EP buffer's element width
+        # (ep_bootstrap rejects max_token_dtype != bf16, and the NCCL EP
+        # HT mega-buffer is sized for 2-byte slots accordingly).
+        w_b = recv_w_flat[:, None].astype(intermediate.dtype)
+        mask_b = (recv_w_flat != 0).astype(intermediate.dtype)[:, None]
+        intermediate = intermediate * w_b * mask_b
```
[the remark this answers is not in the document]
tdophung: should be fixed in [c2869ed](https://github.com/NVIDIA/TransformerEngine/pull/3116/commits/c2869ede04fea8c58613c182f3b5c79da5c363c2)

transformer_engine/jax/moe.py
```diff
@@ -1687,170 +473,337 @@ def _moe_fwd_rule(  # pylint: disable=unused-argument
...
+    # dropless K-expanded count
+    # ``ceil((B/dp)*S*K / num_local_experts)`` does NOT match: it ignores
+    # the worst-case where all of one EP group's tokens land on a single
+    # local expert. We must size to that worst case or NCCL EP's HT kernel
+    # rejects the dispatch buffer with ``invalid argument``.
+    natural_spe = num_ep * max_tokens_per_rank  # = (B // dp_size) * S
+    # NCCL EP requires each expert-major output block to be at least
+    # ``_ALIGN_SIZE`` (=128) tokens; see the constant's docstring.
+    slots_per_expert = ((natural_spe + _ALIGN_SIZE - 1) // _ALIGN_SIZE) * _ALIGN_SIZE
+    recv_pr = num_local_experts * slots_per_expert
```
jberchtold-nvidia: This is currently an overestimate on the worst-case memory when `local expert count > TopK's K`, compute will still be correct but we are using unnecessary memory.

For example, with `local_experts=16, topk=8, seqlen=4k`, this will allocate `16 * 4k` tokens in the buffer but the worst-case is really `8 * 4k` tokens because it's impossible for all tokens to be routed to all 16 local experts on a particular device since it's capped out at top K

Abhinav shared the following diff as a suggestion. Does this bound make sense? Want to make sure we can get a tight bound but not be too aggressive and underestimate the memory required
```
diff --git a/transformer_engine/jax/moe.py b/transformer_engine/jax/moe.py
--- a/transformer_engine/jax/moe.py
+++ b/transformer_engine/jax/moe.py
@@ -524,20 +524,13 @@ def _moe_fwd_rule(
     # Per-rank send capacity: B/num_procs rows x S tokens per rank.
     max_tokens_per_rank = (B // num_procs) * S
-    # Per-rank receive capacity. NCCL EP HT lays out the per-rank receive
-    # buffer as ``[num_local_experts, num_ep * max_tokens_per_rank, hidden]``
-    # and reserves the worst case independently for every local expert.
-    natural_spe = num_ep * max_tokens_per_rank
+
+    # Bound receive capacity by the total top-k assignments in one EP group.
+    natural_recv_pr = num_ep * max_tokens_per_rank * K
+    natural_spe = (
+        natural_recv_pr + num_local_experts - 1
+    ) // num_local_experts
+
     slots_per_expert = (
         (natural_spe + _ALIGN_SIZE - 1) // _ALIGN_SIZE
     ) * _ALIGN_SIZE
     recv_pr = num_local_experts * slots_per_expert
```

transformer_engine/jax/moe.py
```diff
@@ -862,22 +862,16 @@ def _moe_bwd_rule(
...
-        # all-NaN). Sanitize once here.
-        recv_w_clean = jnp.where(jnp.isnan(ctx.recv_topk_weights), 0, ctx.recv_topk_weights)
-        w = recv_w_clean[..., None].astype(grad_pre_combine.dtype)
-        mask_bool = (recv_w_clean != 0)[..., None]
+        # Bwd mirror of the fwd mask: grad_pre_combine and ctx.expert_outputs
+        # both carry NaN at padded slots (ep_dispatch_fwd leaves uninit
+        # memory in recv_tokens, the FFN and combine_bwd propagate), and
+        # IEEE NaN * 0 = NaN, so jnp.where is needed to overwrite padded
+        # positions with literal zeros before the sum reduction.
+        w = ctx.recv_topk_weights[..., None].astype(grad_pre_combine.dtype)
```
jberchtold-nvidia, line 861: Why do we need masking for NaNs here? I thought it's okay to have NaNs in the padded region because we never read from it

tdophung: I realized that it is okay in the forward pass but not the bwd pass because in the bwd pass, the weight gradient is produced by summing over rows in the expert group, which will be poisoned with the Nans in the padded region. Note that padded region here is the padding in each expert tokens group, not the overallocated long tail.

Ex:
```
  expert 0 group size = 4
  rows 0,1 = valid
  rows 2,3 = alignment padding
```
  Forward GEMM does:
```
  Y[row] = X[row] @ W_expert
```
  If X[2] = NaN, then Y[2] = NaN, but rows 0 and 1 are unaffected. Then `ep_combine` should only consume the valid rows by handle metadata, so padded-row NaNs do not poison valid output.

  Backward weight grad does a reduction over rows:
```
  d_wo = intermediate.T @ d_expert_output
       = sum over rows in this expert group
```
  With padded group size 4:
```
  d_wo = row0 contribution + row1 contribution + row2 contribution + row3 contribution
```
  If row2 is padding and contains NaN, then the reduction can become NaN. This is different from forward because the padded row no longer stays isolated; it participates in a sum that produces one shared weight gradient for the expert.

transformer_engine/jax/moe.py
tdophung: This seems valid, but need to account for the worst case  alignment  for each expert, not average alignment per expert. I fixed it in [9a13ef8](https://github.com/NVIDIA/TransformerEngine/pull/3116/commits/9a13ef8512b5537fd2487934526d7acb4c65b12b)

tdophung: Here is an example why ^ would fail:
```
  natural_recv_pr = num_ep * max_tokens_per_rank * K
  natural_spe = ceil(natural_recv_pr / num_local_experts)
  recv_pr = num_local_experts * align(natural_spe)
  ```
num_local_experts=4, align=4, natural_recv_pr=16.
The ^ formula will give:
  natural_spe = ceil(16 / 4) = 4
  recv_pr = 4 * align(4) = 16

  But a valid actual local distribution can have fewer active assignments than natural_recv_pr and still have the recv_pr required to hold the padded zones to be > 16

  actual counts = [5, 4, 4, 1]   # total = 14 <= natural_recv_pr
  padded zones  = [8, 4, 4, 4]   # total = 20 > recv_pr

  So even though natural_recv_pr is above the active total, dividing it evenly across experts loses the skew information.

File: transformer_engine/jax/moe.py (hunk 18 above)
Comment (tdophung) [reply transformer_engine/jax/moe.py:861]: So from this comment, I removed some more jnp.where in fwd passes in [a49dc6b](https://github.com/NVIDIA/TransformerEngine/pull/3116/commits/a49dc6bfeadb04398298a7ece354c66eb1c41240) and general anywhere where downstream from it, we do not do any reduction over the padded rows 
e626a7f

Comment (jberchtold-nvidia) [reply transformer_engine/jax/moe.py:861]: @tdophung 

> Backward weight grad does a reduction over rows:
  d_wo = intermediate.T @ d_expert_output
       = sum over rows in this expert group

This is referring to the grouped GEMM wgrad reduction, right? The backward wgrad grouped GEMM is "group aware" so it should ignore any padded regions and therefore shouldn't get poisoned with NaNs

File: transformer_engine/jax/cpp_extensions/ep.py [hunk 19]
```diff
@@ -125,8 +125,15 @@ def _ep_outer_axis():
...
+
+    A dp/fsdp axis that is sized 1 in the active mesh is treated as absent so
+    we don't pin EP-output specs to a degenerate axis that JAX may collapse.
     """
     gsr = global_mesh_resource()
+    if gsr.dp_resource is not None and get_mesh_axis_size(gsr.dp_resource) > 1:
+        return gsr.dp_resource
+    if gsr.fsdp_resource is not None and get_mesh_axis_size(gsr.fsdp_resource) > 1:
+        return gsr.fsdp_resource
     return gsr.dp_resource or gsr.fsdp_resource
```
jberchtold-nvidia, line 137: nit SR: with the newly added checks this final `return gsr.dp_resource or gsr.fsdp_resource` is unnecessary and can be simplified to `return None`, right?

transformer_engine/jax/moe.py
tdophung: Yes, this is referring to the grouped GEMM wgrad reduction.

 Grouped GEMM is “group-aware” only with respect to the `group_sizes` we pass. Here those sizes are TE EP’s caller-visible per-expert padded counts, not active counts. The padding has two forms: 1. type A is per-expert alignment padding, and 2. type B is the overallocated tail from reserving a worst-case recv buffer. `group_sizes` include type A padding but exclude type B.

  For example, if an expert has 5 real tokens and alignment 128, grouped GEMM sees group size 128. Forward/data-gradient GEMMs keep those extra rows row-local, so NaNs there do not affect valid rows. But wgrad is a reduction over the whole group:

  `d_wo[e] = intermediate[e, :group_size].T @ d_expert_output[e, :group_size]`

  Since `:group_size` includes the type-A padding rows, nans in those rows can participate in the reduction and poison the expert’s wgrad. That is why we still mask tensors feeding wgrad reductions, while removing masks  on paths where padded slot values stay row-local or are ignored by TE EP group_sizes-aware dispatch/combine.

jberchtold-nvidia: Thanks for the explanation! I understand it now. Yes, let's keep this masking in

jberchtold-nvidia (approved):
LGTM, thanks!

PR #3123, opened 2026-06-12
Avoid unpickling the extra state when not needed

ptrendx:
# Description

Avoids unpickling of the extra state if the recipe is stateless. Adds a guard prompting user to explicitly allow loading of the checkpoint when the unpickling is necessary.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Avoids unpickling of the stateless recipe extra state
- Adds a guard and environment variable for the delayed scaling recipes


transformer_engine/common/recipe/__init__.py
```diff
@@ -50,6 +50,20 @@ class Format(Enum):
     HYBRID = _FormatHelper(max_fwd=E4M3.max_fwd, max_bwd=E5M2.max_bwd)
 
 
+class CheckpointExtraStatePolicy(Enum):
```
timmoon10: It's uncomfortable how PyTorch-specific logic is making its way into `te.common`. Perhaps we should try to phrase this more generically so it's not about `torch.nn.Module.set_extra_state`, but about recipe statefulness. However, this won't work if we implement future stateful recipes and we want to add an enum value for non-pickle checkpoint formats.

Since the bug comes from a mistake in `te.pytorch`, it might be better for the hacky WAR to also live in `te.pytorch`. I see that `_extra_state.py` does some type inspection to create a map between recipes and policies:
https://github.com/ptrendx/TransformerEngine/blob/69a5ab8fd4bfeb188b40d8886faa155c9014aeaf/transformer_engine/pytorch/_extra_state.py#L37-L41
How about we just define the recipe-policy map directly? Each time we add a new stateful recipe, we can update the map and add any additional checkpointing logic there instead of touching `te.common`.

transformer_engine/pytorch/_extra_state.py
```diff
@@ -0,0 +1,168 @@
...
+from typing import Optional
+
+from ..common.recipe import CheckpointExtraStatePolicy, Recipe
+
+
+UNSAFE_PICKLE_EXTRA_STATE_ENV = "NVTE_ALLOW_UNSAFE_PICKLE_EXTRA_STATE"
+
+_RECIPE_MODULE = "transformer_engine.common.recipe"
+_RECIPE_KEY = "recipe"
+_DELAYED_STATE_KEYS = {
```
timmoon10: Nit: All of this is really a backward-compatibility hack for FP8 delayed scaling, and if we implement other delayed scaling recipes in the future we're not going to use pickles in checkpoints.
```suggestion
_FLOAT8_DELAYED_SCALING_STATE_KEYS = {
```

transformer_engine/common/recipe/__init__.py
```diff
@@ -50,6 +50,20 @@ class Format(Enum):
...
+    """How pickled PyTorch ``_extra_state`` should be handled in ``set_extra_state``.
+
+    Each recipe subclass must choose a policy so checkpoint loading can decide
+    whether unpickling is required. ``DYNAMIC`` means the recipe class alone is
+    not enough; callers must inspect the checkpoint payload shape inside
+    ``set_extra_state`` before deciding whether the pickle can be ignored.
+    """
+
+    STATELESS = "stateless"
+    STATEFUL = "stateful"
```
timmoon10: We may have stateful recipes in the future, but we've learned our lesson not to naively pickle. We should make clear that this particular enum value represents stateful recipes with unsafe pickling.
```suggestion
    STATEFUL_FP8_DELAYED_SCALING = "stateful_fp8_delayed_scaling"
```
Other possible names could be `STATEFUL_PICKLE` or `STATEFUL_UNSAFE`.

timmoon10 (commented):
Overall this seems like a reasonable fix, although I have some design suggestions and nits. FP8 delayed scaling still has pickling, but at least we can avoid it for more modern recipes.

ksivaman (approved):

PR #3135, opened 2026-06-17
[Common/PyTorch] Grouped-quantize kernels for 1D and 2D FP8 block-scaling

### LINKED ISSUE #2525: Blockwise (1x128 and 128x128) FP8 grouped quantization

denera:
# Description

Implements grouped-tensor quantize for the FP8 1D (1x128) and 2D (128x128) block-scaling recipes in row-wise (RW), column-wise (CW) and BOTH quantization directions. A single CUDA kernel launch walks 128x128 tiles across every tensor in the group, with each CTA decoding its owning tensor from the device-side GroupedTensor metadata with (N, R, K) shapes. Supports `SAME_BOTH_DIMS` (all tensors identical) and `VARYING_FIRST_DIM` (constant K, varying R) shape representations. 

Three kernels share the dispatcher in `group_quantize_blockwise_{1d,2d}`:
- `group_block_scaled_1d_rw_kernel` — RW-only dispatch; 8 threads/row, reads global memory directly into vec-16 registers; bypasses TMA because the shared memory roundtrip and `ptx::mbarrier` does not buy anything without re-use in CW path.
- `group_block_scaled_1d_tma_kernel` — CW-only and BOTH dispatch; TMA bulk-load fills shared memory input cache. BOTH runs RW pass first (8 threads/row, vec-16 read from shared memory) then CW pass (2 threads/column, 64-row register stage); CW-only skips the RW pass. CW path writes the transposed-FP8 tile to a shared memory transpose staging buffer, then drains to global memory.
- `group_block_scaled_2d_tma_kernel` — RW-only, CW-only and BOTH dispatch; TMA bulk-load fills shared memory input cache. Pass 1 stages 8 IVecs/thread in registers while computing the per-tile scalar amax. Pass 2 quantizes from registers, emits row-wise output, stages column-wise output to shared memory transpose staging buffer, then drains to global memory.

Kernels are gated to Hopper (sm_90) at the host dispatcher (cuBlasLt grouped GEMM supports FP8 block-scaling only on Hopper).

PR includes PyTorch integration into te.GroupedTensor only.

PyTorch integration into te.GroupedLinear and JAX integration deferred to a follow-up PRs.

Partially resolves #2525

## Performance

Benchmark on H200 with a sweep of grouped tensors in (N, M, K) shapes:
- N ∈ {4, 8, 16, 32, 64, 128} (# of device-local experts)
- M = 4096 @ N = 4 → M = 128 @ N = 128 (# of tokens/expert, scaling inversely with # experts)
- K ∈ {1024, 1792, 2048, 3584, 4096, 7168} (device-local shard of TP-hidden/intermediate-FFN dim)

Two shape families per config:
- **U_MoE** (uniform, `SAME_BOTH_DIMS`): all experts share the (M, K) shape
- **J_MoE** (jagged, `VARYING_FIRST_DIM`): per-expert M drawn from an imbalanced routing, common K

Buckets:
- **Small/Unsaturated (S):** R·K ≤ 32M elements (< 2048 tiles and < 15 waves on H200's 132 SMs)
- **Large/Saturated (L):** R·K > 32M elements (> 2048 tiles, SMs busy across many waves)

Bucket medians across 3 reps. Speedup is grouped vs the split-quantized fallback that loops over the grouped tensor and quantizes each constituent sequentially. % mono is grouped throughput relative to a single non-grouped FP8 block-scaling quantize on the equivalent monolithic (N·M, K) tensor.

| Bucket | Path    | Grouped (ms) | Split (ms) | Speedup | % memcpy tput | % mono tput |
|:------:|:--------|-------------:|-----------:|--------:|--------------:|------------:|
| S | 1D RW   |       0.019 |      0.084 |  4.50× |        64.9 % |     114.9 % |
| S | 1D CW   |       0.022 |      0.090 |  4.17× |        56.4 % |     112.4 % |
| S | 1D BOTH |       0.033 |      0.116 |  3.57× |        50.2 % |     102.8 % |
| S | 2D RW   |       0.018 |      0.076 |  4.19× |        65.9 % |     100.9 % |
| S | 2D CW   |       0.020 |      0.089 |  4.64× |        62.6 % |     126.3 % |
| S | 2D BOTH |       0.027 |      0.089 |  3.68× |        66.3 % |      98.8 % |
| L | 1D RW   |       0.058 |      0.198 |  2.06× |        87.1 % |     118.7 % |
| L | 1D CW   |       0.064 |      0.213 |  2.05× |        77.3 % |     118.5 % |
| L | 1D BOTH |       0.098 |      0.282 |  1.77× |        66.7 % |     108.4 % |
| L | 2D RW   |       0.054 |      0.178 |  1.99× |        87.7 % |     100.1 % |
| L | 2D CW   |       0.058 |      0.213 |  2.20× |        85.3 % |     135.9 % |
| L | 2D BOTH |       0.078 |      0.213 |  1.62× |        85.0 % |     102.4 % |

| # experts (N) | S bucket | L bucket |
|--------------:|---------:|---------:|
|             4 |    1.74× |    1.42× |
|             8 |    2.40× |    1.45× |
|            16 |    4.18× |    1.89× |
|            32 |    5.50× |    2.81× |
|            64 |   10.43× |    7.51× |
|           128 |   19.81× |    8.72× |

### Notes

- % of mono throughput is roughly consistent across buckets for every path, confirming no per-expert overhead in the new kernels.
- Greater-than-100% mono throughput cases come from TMA bulk-loads, register staging, and vec-16 reads that the non-grouped FP8 block-scaling kernels do not have.
- Speedup over split-quantize scales as expected with # of experts (roughly linearly in the unsaturated regime).
- S-bucket % memcpy is lower than L because launch and per-CTA setup are not amortized over a long bandwidth-bound steady state; absolute kernel times are still small (< 35 µs).

#### 1D CW load bank conflicts on ~35% of load wavefronts (reading from the shared memory input-cache)
- No possible stride padding or XOR swizzle to alleviate.
- TMA hardware swizzle with `CU_TENSOR_MAP_SWIZZLE_128B` has the right pattern but caps FP16/BF16 at 64-elements; does not fit the 128-element tile for FP8 block-scaling without doubling per-tile launch overhead (quadrupling for FP32).
- CW-only stays at 2 t/col. Going to 4 t/col would double the bank-conflict footprint (4 lanes per column at the same row stride instead of 2), and CW-only is not occupancy-bound (3 CTAs/SM regardless), so the restructure costs more than it saves.
- 1D BOTH uses 4 t/col with a 32-row reg_data per thread and two column passes per CTA. The RW pass's per-expert scale-offset arithmetic plus a 64-row reg_data crossed the 85-reg / 3-CTA threshold on sm_90; halving reg_data restores 4 CTAs/SM. The doubled column pass and extra XOR-reduce stage are cheap relative to the occupancy gain.

#### 1D BOTH reads the shared memory input-cache twice
- The RW (8 threads/row) and CW (2 threads/column) passes have different threading.
- Attempted to unify with 8 threads/row for both RW and CW. Caused bank conflicts on ~76% of store wavefronts (writing to the shared memory transpose buffer), reduced to ~43% with a XOR swizzle but not enough to beat separate RW/CW passes.
- Did not pursue the 2 threads/column unification; costs 40x more shfl ops than 8 threads/row attempt, plus a shared memory partial buffer and sync.

#### 2D CW/BOTH has bank conflicts on ~16% of store wavefronts (when writing to the shared memory transpose buffer)
- Already reduced from ~75% via a XOR swizzle, further reduction was not possible.
- Minimal impact (< 5%) on kernel time.

#### No TMA-store
- MXFP8 grouped quantize kernel leverages this by decomposing a 128x128 tile into 32-row sub-stages that each have their own independent 32x1 or 1x32 scale; shared memory footprint is based on a single sub-stage; can be quantized and TMA-stored independently; hides TMA-store of one stage under the compute of next stage.
- FP8 block-scaling 128-element scale-block spans the entire 128-row tile. Cannot decompose into independent sub-stages and pipeline the TMA-stores. Single non-pipelined TMA-store requires holding the transposed staging buffer for the entire tile until all work on tile is finished, blows up shared memory footprint, collapses occupancy to 2CTA/SM. The recipe itself is the roadblock.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/common/cast/dispatch/dequantize.cuh
+++ b/transformer_engine/common/cast/dispatch/dequantize.cuh
@@ -15,6 +15,7 @@
 
 #include "../../common.h"
 #include "../fp8/dequantize_fp8.cuh"
+#include "../fp8_blockwise/group_dequantize_fp8_blockwise.cuh"
 #include "../mxfp8/dequantize_mxfp8.cuh"
 #include "../mxfp8/group_dequantize_mxfp8.cuh"
 #include "../nvfp4/dequantize_nvfp4.cuh"
@@ -67,6 +68,11 @@ inline void group_dequantize_helper(const GroupedTensor &input, GroupedTensor *o
       } else {
         NVTE_ERROR("MXFP8 Grouped Dequantization is NOT supported by architectures < 10.0");
       }
+      break;
+    }
+    case NVTE_BLOCK_SCALING_1D:
+    case NVTE_BLOCK_SCALING_2D: {
+      fp8_blockwise::group_dequantize(&input, output, stream);
       break;
     }
     default:
--- a/transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
+++ b/transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
@@ -1,0 +1,1022 @@
+/*************************************************************************
+ * Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ *
+ * See LICENSE for license information.
+ ************************************************************************/
+
+/*! \file group_quantize_fp8_blockwise.cuh
+ *  \brief CUDA kernels to quantize grouped tensors with FP8 1D and 2D
+ *  block scaling. A single launch walks 128x128 tiles across every tensor
+ *  in the group, with each CTA decoding its owning tensor from the device-side
+ *  GroupedTensor metadata. Supports SAME_BOTH_DIMS and VARYING_FIRST_DIM.
+ */
+
+#ifndef TRANSFORMER_ENGINE_GROUP_QUANTIZE_FP8_BLOCKWISE_CUH_
+#define TRANSFORMER_ENGINE_GROUP_QUANTIZE_FP8_BLOCKWISE_CUH_
+
+#include <cuda.h>
+#include <cudaTypedefs.h>
+#include <cuda_runtime.h>
+#include <transformer_engine/transformer_engine.h>
+
+#include <vector>
+
+#include "../../common.h"
+#include "../../recipe/recipe_common.cuh"
+#include "../../transpose/cast_transpose.h"
+#include "../../util/cuda_runtime.h"
+#include "../../util/ptx.cuh"
+#include "../../utils.cuh"
+#include "../core/common.cuh"
+
+namespace transformer_engine {
+namespace dispatch {
+namespace fp8_blockwise {
+
+using transformer_engine::detail::FP8BlockwiseColumnwiseOption;
+using transformer_engine::detail::FP8BlockwiseRowwiseOption;
+
+constexpr int kTileDim = 128;
+constexpr int kThreadsPerWarp = 32;
+constexpr int kThreadsPerBlock = 256;
+constexpr int kNumWarps = kThreadsPerBlock / kThreadsPerWarp;
+
+// ---- Per-expert scale layout helpers --------------------------------------------
+//
+// cuBLAS grouped FP8 block-scaling GEMM expects each expert's scales to live
+// in a contiguous per-expert sub-block of the global scale buffer:
+//
+//   1D rowwise CW (op_rowwise=true)   per expert: (blocks_X, roundup(M_t, 4))    floats
+//   1D columnwise (op_rowwise=false)  per expert: (blocks_y_t, roundup(K, 4))    floats
+//   2D rowwise (op_rowwise=true)      per expert: (blocks_y_t, roundup(blocks_X, 4))
+//   2D columnwise (op_rowwise=false)  per expert: (blocks_X, roundup(blocks_y_t, 4))
+//
+// The grouped kernel writes the GLOBAL buffer in tile-stride order, but the
+// position assigned to each tile must map into the per-expert contiguous
+// sub-block. These helpers compute the per-expert cumulative byte/float
+// offset and the per-expert local stride. SAME_BOTH_DIMS lets us derive
+// without walking offsets; VARYING_FIRST_DIM walks `tensor_offsets_ptr` once
+// per block (only the writing thread).
+
+__device__ __host__ constexpr size_t kScaleColAlign = 4;
+
+// 2D columnwise per-expert scale offset. Per-expert layout is
+// (blocks_X, roundup(blocks_y_t, 4)). Two paths, picked at call sites:
+//   - SAME_BOTH_DIMS: direct formula (no walk, no reduction).
+//   - VARYING_FIRST_DIM: CTA-cooperative prefix sum. Each thread accumulates a
+//     partial over a strided subset of the tensors-before-this-one, a
+//     warp-shuffle reduces inside each warp, then all threads sum the per-warp
+//     partials read from a kNumWarps-element smem buffer to obtain the same
+//     total. The non-linear DIVUP_TO_MULTIPLE on each per-tensor blocks_y_t
+//     prevents a closed form.
+// ALL threads of the CTA must call this in lock-step (the cooperative path
+// contains __syncthreads()).
+template <bool kSameBothDims>
+__device__ __forceinline__ size_t compute_2d_cw_expert_offset(
+    const size_t tensor_id, const size_t blocks_X, const size_t common_first_dim_blocks,
+    const size_t tile_row_stride, const int64_t* __restrict__ tensor_offsets_ptr,
+    size_t* warp_partials_smem, const int tid, const int warp_id, const int lane) {
+  if constexpr (kSameBothDims) {
+    return tensor_id * blocks_X * DIVUP_TO_MULTIPLE(common_first_dim_blocks, kScaleColAlign);
+  }
+  size_t my_partial = 0;
+  for (size_t i = static_cast<size_t>(tid); i < tensor_id;
+       i += static_cast<size_t>(kThreadsPerBlock)) {
+    const size_t blocks_y_i =
+        static_cast<size_t>(tensor_offsets_ptr[i + 1] - tensor_offsets_ptr[i]) / tile_row_stride;
+    my_partial += DIVUP_TO_MULTIPLE(blocks_y_i, kScaleColAlign);
+  }
+  my_partial = warp_allreduce_sum(my_partial);
+  if (lane == 0) warp_partials_smem[warp_id] = my_partial;
+  __syncthreads();
+  size_t total = 0;
+#pragma unroll
+  for (int w = 0; w < kNumWarps; ++w) {
+    total += warp_partials_smem[w];
+  }
+  return total * blocks_X;
+}
+
+// 1D rowwise: per-expert layout (blocks_X, roundup(M_t, 4)).
+template <bool kSameBothDims>
+__device__ __forceinline__ size_t expert_scale_offset_1d_rowwise(
+    size_t tensor_id, size_t blocks_X, size_t common_first_dim_blocks, size_t tile_row_stride,
+    const int64_t* __restrict__ tensor_offsets_ptr) {
+  if constexpr (kSameBothDims) {
+    const size_t M = common_first_dim_blocks * kTileDim;
+    return tensor_id * blocks_X * DIVUP_TO_MULTIPLE(M, kScaleColAlign);
+  } else {
+    // Each M_i is enforced to be a multiple of kTileDim (=128), hence a
+    // multiple of kScaleColAlign (=4), so DIVUP_TO_MULTIPLE(M_i, 4) == M_i and
+    // sum_{i<tensor_id} M_i = tensor_offsets_ptr[tensor_id] / K. Computing the
+    // offset in O(1) avoids a per-CTA O(num_tensors) walk that otherwise
+    // dominates wall-clock for jagged routings at high expert counts.
+    const size_t K = tile_row_stride / kTileDim;
+    const size_t total_M_before = static_cast<size_t>(tensor_offsets_ptr[tensor_id]) / K;
+    return blocks_X * total_M_before;
+  }
+}
+
+// ---- Tensor-lookup helpers ----------------------------------------------------
+
+// Map a global tile-row index to its owning tensor. Delegates to the shared
+// `common::get_current_tensor_id` helper from `cast/core/common.cuh`. The
+// helper is parameterized by total `first_logical_dim` rather than per-tensor
+// block count, so we reconstruct it here for the SAME_BOTH_DIMS specialization
+// (VARYING_FIRST_DIM ignores it and uses `current_offset` + `offsets_ptr`).
+template <bool kSameBothDims>
+__device__ __forceinline__ size_t find_tensor_id_by_block_y(
+    const size_t block_y_global, const size_t num_tensors, const size_t common_first_dim_blocks,
+    const size_t tile_row_stride, const int64_t* __restrict__ tensor_offsets_ptr) {
+  constexpr auto shape_rep =
+      kSameBothDims ? ShapeRepresentation::SAME_BOTH_DIMS : ShapeRepresentation::VARYING_FIRST_DIM;
+  const size_t first_logical_dim = num_tensors * common_first_dim_blocks * kTileDim;
+  const size_t tensor_id = common::get_current_tensor_id<shape_rep, kTileDim>(
+      num_tensors, block_y_global * tile_row_stride, block_y_global, first_logical_dim,
+      /*last_logical_dim=*/0, tensor_offsets_ptr);
+  if constexpr (!kSameBothDims) {
+    // tensor_offsets_ptr carries cumulative element counts; tile_row_stride =
+    // kTileDim * K, so the per-tensor element span is divisible by
+    // tile_row_stride iff first_dim is a multiple of kTileDim.
+    if (tensor_id < num_tensors) {
+      const size_t span =
+          static_cast<size_t>(tensor_offsets_ptr[tensor_id + 1] - tensor_offsets_ptr[tensor_id]);
+      if (span % tile_row_stride != 0) {
+        NVTE_DEVICE_ERROR(
+            "Grouped FP8 block-scaling quantize: each tensor's first dimension must be a "
+            "multiple of 128 (VARYING_FIRST_DIM).");
+      }
+    }
+  }
+  return tensor_id;
+}
+
+// Per-tensor block-y base for VARYING_FIRST_DIM (in 128-row block units).
+__device__ __forceinline__ size_t tensor_block_y_base_from_offsets(
+    const size_t tensor_id, const int64_t* __restrict__ tensor_offsets_ptr,
+    const size_t tile_row_stride) {
+  return static_cast<size_t>(tensor_offsets_ptr[tensor_id]) / tile_row_stride;
+}
+
+// Per-vector amax. Uses bf16x2 `max.xorsign.abs` on sm_89+; FP32 fallback otherwise.
+template <typename IType, typename CType, int kVec>
+__device__ __forceinline__ CType compute_row_amax(const Vec<IType, kVec>& v) {
+#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 890)
+  if constexpr (std::is_same_v<IType, bf16>) {
+    static_assert(kVec % 2 == 0, "kVec must be even for packed bf16x2 amax");
+    const ptx::bf16x2* pairs = reinterpret_cast<const ptx::bf16x2*>(&v.data.elt[0]);
+    ptx::bf16x2 amax_x2{static_cast<bf16>(0.f), static_cast<bf16>(0.f)};
+#pragma unroll
+    for (int p = 0; p < kVec / 2; ++p) {
+      ptx::abs_max_2x(amax_x2, amax_x2, pairs[p]);
+    }
+    return static_cast<CType>(__hmax(__habs(amax_x2.x), __habs(amax_x2.y)));
+  }
+#endif
+  CType amax = 0.f;
+#pragma unroll
+  for (int e = 0; e < kVec; ++e) {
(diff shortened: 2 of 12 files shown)


transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+
+using transformer_engine::detail::FP8BlockwiseColumnwiseOption;
+using transformer_engine::detail::FP8BlockwiseRowwiseOption;
+
+constexpr int kTileDim = 128;
+constexpr int kThreadsPerWarp = 32;
+constexpr int kThreadsPerBlock = 256;
+constexpr int kNumWarps = kThreadsPerBlock / kThreadsPerWarp;
+
+// Align a dynamic-smem pointer to 128 bytes (TMA requirement).
```
Oleg-Goncharov: Could we reuse the existing `align_smem_ptr_per_TMA_requirements()` helper from `transformer_engine/cast/core/common.h` here?

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+// Follows the same convention as MXFP8 gemm_swizzled_scale_idx but applied to
+// FP32 1D/2D block-scaling colwise scales:
+//   i = column index (per-column scales for 1D; tile_x for 2D)
+//   j = row-block index (tile_y_global)
+//   num_tiles_X = DIVUP(total_row_blocks, TILE_DIM_X=4)
+__device__ __forceinline__ size_t swizzled_colwise_scale_idx(size_t i, size_t j,
+                                                             size_t total_row_blocks) {
+  using namespace transformer_engine::dispatch::mxfp8::swizzle;
+  const size_t num_tiles_X =
+      (total_row_blocks + GEMM_SWIZZLED_SCALE_TILE_DIM_X - 1) / GEMM_SWIZZLED_SCALE_TILE_DIM_X;
```
Oleg-Goncharov: We can also reuse the existing `DIVUP()` helper here (defined in `transformer_engin/common/common.h`).

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+  return gemm_swizzled_scale_idx(i, j, num_tiles_X);
+}
+
+// ---- Tensor-lookup helpers ----------------------------------------------------
+
+// Map a global tile-row index to its owning tensor by binary-searching
```
Oleg-Goncharov: We can also reuse the existing `get_current_tensor_id()` helper defined in `transformer_engine/cast/core/common.cuh`

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+
+  __shared__ CType warp_amaxes[kNumWarps];
+  __shared__ uint64_t tma_mbar;
+
+  const int tid = threadIdx.x;
+  const bool leading_thread = (tid == 0);
+
+  // ---- TMA async load of the input tile ----
+  if (leading_thread) {
+    ptx::mbarrier_init(&tma_mbar, 1);
```
Oleg-Goncharov, line 327: Since mbar resides in shared memory, a cross-proxy fence between the `async` and `generic` proxies needs to be issued here before `__syncthreads()` so that both the TMA engine and the threads observe mbar in the correct state. We can use `ptx::fence_proxy_async_shared_cta()` defined in `transformer_engine/common/util/ptx.cuh`.

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+    if (c + kVec <= K) {
+      in_vec[it].load_from(input_base + r_global * K + c);
+    } else if (c < K) {
+      in_vec[it].load_from_elts(input_base + r_global * K + c, 0, K - c);
+    } else {
+      in_vec[it].clear();
+    }
+
+    CType amax = compute_row_amax<IType, CType, kVec>(in_vec[it]);
+    amax = fmaxf(amax, __shfl_xor_sync(0xffffffff, amax, 1));
```
Oleg-Goncharov: Could we reuse the existing amax warp-reduction helpers (`warp_reduce_max()` or `reduce_max()`) from `transformer_engine/common/utils.cuh` here?

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+  IType(*smem)[kTileDim] = reinterpret_cast<IType(*)[kTileDim]>(smem_base);
+
+  __shared__ uint64_t tma_mbar;
+  const int tid = threadIdx.x;
+  const bool leading_thread = (tid == 0);
+
+  // ---- TMA async load of the input tile ----
+  if (leading_thread) {
+    ptx::mbarrier_init(&tma_mbar, 1);
+  }
```
Oleg-Goncharov, line 617: Similar to the above:
```suggestion
    ptx::fence_proxy_async_shared_cta();
  }
```

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+#pragma unroll
+    for (int it = 0; it < kRwIters; ++it) {
+      const int row_local = rw_thr_row + it * kRowsPerIterRW;
+      Vec<IType, kVec> in_vec;
+      in_vec.load_from(&smem[row_local][col_local]);
+
+      CType amax = compute_row_amax<IType, CType, kVec>(in_vec);
+      amax = fmaxf(amax, __shfl_xor_sync(0xffffffff, amax, 1));
+      amax = fmaxf(amax, __shfl_xor_sync(0xffffffff, amax, 2));
+      amax = fmaxf(amax, __shfl_xor_sync(0xffffffff, amax, 4));
```
Oleg-Goncharov: We can also reuse `reduce_max()` or `warp_reduce_max()` here.

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+
+    drain_smem_t_1d_to_gmem<OType, kSMemTRowStride>(smem_T, output_t_base, global_col_base,
+                                                    global_row_base, R_total, K, tid);
+  }
+#endif  // __CUDA_ARCH__ >= 900 && __CUDA_ARCH__ < 1000
+}
+
+// ----- Host-side dispatchers --------------------------------------------------------------------
+
+inline size_t align_up_to(size_t x, size_t a) { return ((x + a - 1) / a) * a; }
```
Oleg-Goncharov: We can reuse `DIVUP_TO_MULTIPLE()` defined in `transformer_engine/common/common.h`.

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+    info.common_first_dim_blocks = common_first_dim / kTileDim;
+    info.R_total = info.num_tensors * common_first_dim;
+  } else {
+    info.common_first_dim_blocks = 0;
+    info.R_total = output->logical_shape.data[0];
+    info.tensor_offsets_d = reinterpret_cast<const int64_t*>(output->tensor_offsets.dptr);
+    NVTE_CHECK(info.tensor_offsets_d != nullptr,
+               "VARYING_FIRST_DIM requires tensor_offsets to be set on the GroupedTensor.");
+  }
+  info.total_row_blocks = (info.R_total + kTileDim - 1) / kTileDim;
```
Oleg-Goncharov:
```suggestion
  info.total_row_blocks = DIVUP(info.R_total, kTileDim);
```

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+  info.blocks_X = (info.K + kTileDim - 1) / kTileDim;
```
Oleg-Goncharov:
```suggestion
  info.blocks_X = DIVUP(info.K, kTileDim);
```

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+  const bool same_both_dims = output->all_same_shape();
+  const bool varying_first_dim = (!output->all_same_first_dim()) && output->all_same_last_dim();
+  NVTE_CHECK(same_both_dims || varying_first_dim,
+             "Grouped FP8 block-scaling supports only SAME_BOTH_DIMS and VARYING_FIRST_DIM "
+             "shape representations.");
+
+  info.same_both_dims = same_both_dims;
+  info.num_tensors = output->num_tensors;
+  info.K = output->get_common_last_dim();
+  NVTE_CHECK(info.K % 16 == 0, "Last dim must be multiple of 16 (FP8 alignment).");
```
Oleg-Goncharov: If this is a TMA requirement, we can use the `TMA_GMEM_ALIGNMENT` constant defined in `transformer_engine/common/common.h`

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+             "Input and output must have same num_tensors.");
+
+  auto info = prepare_grouped_blockwise_launch(output);
+  if (info.R_total == 0 || info.K == 0) return;
+
+  using CType = float;
+  const float* noop_ptr =
+      (noop != nullptr) ? reinterpret_cast<const float*>(noop->data.dptr) : nullptr;
+
+  const size_t scale_stride_y = align_up_to(info.blocks_X, 4);
```
Oleg-Goncharov:
```suggestion
  const size_t scale_stride_y = DIVUP_TO_MULTIPLE(info.blocks_X, 4);
```

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+  // CW scales are stored [blocks_X, align4(total_row_blocks)] -- transposed to
+  // match the physically-transposed columnwise data the TN cuBLAS GEMM consumes.
+  const size_t scale_t_stride_y = align_up_to(info.total_row_blocks, 4);
```
Oleg-Goncharov:
```suggestion
  const size_t scale_t_stride_y = DIVUP_TO_MULTIPLE(info.total_row_blocks, 4);
```

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+             "Input and output must have same num_tensors.");
+
+  auto info = prepare_grouped_blockwise_launch(output);
+  if (info.R_total == 0 || info.K == 0) return;
+
+  using CType = float;
+  const float* noop_ptr =
+      (noop != nullptr) ? reinterpret_cast<const float*>(noop->data.dptr) : nullptr;
+
+  const size_t scale_stride_aligned_R = align_up_to(info.R_total, 4);
```
Oleg-Goncharov:
```suggestion
  const size_t scale_stride_aligned_R = DIVUP_TO_MULTIPLE(info.R_total, 4);
```

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+  const size_t scale_t_stride_aligned_K = align_up_to(info.K, 4);
```
Oleg-Goncharov:
```suggestion
  const size_t scale_t_stride_aligned_K = DIVUP_TO_MULTIPLE(info.K, 4);
```

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+  __shared__ CType warp_amaxes[kNumWarps];
+  __shared__ uint64_t tma_mbar;
+
+  const int tid = threadIdx.x;
+  const bool leading_thread = (tid == 0);
+
+  // ---- TMA async load of the input tile ----
+  if (leading_thread) {
+    ptx::mbarrier_init(&tma_mbar, 1);
+  }
```
Oleg-Goncharov, line 331:
```suggestion
    ptx::fence_proxy_async_shared_cta();
  }
```

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,838 @@
...
+
+// Computes the gemm-swizzled colwise scale index for FP8 block-scaling.
+// Follows the same convention as MXFP8 gemm_swizzled_scale_idx but applied to
+// FP32 1D/2D block-scaling colwise scales:
+//   i = column index (per-column scales for 1D; tile_x for 2D)
+//   j = row-block index (tile_y_global)
+//   num_tiles_X = DIVUP(total_row_blocks, TILE_DIM_X=4)
+__device__ __forceinline__ size_t swizzled_colwise_scale_idx(size_t i, size_t j,
+                                                             size_t total_row_blocks) {
+  using namespace transformer_engine::dispatch::mxfp8::swizzle;
```
vthumbe1503: I think we should rename the namespace for swizzle...given that we use the same constants for mxfp8, nvfp4, fp8 block scaling

denera: Grouped GEMM doesn't read FP8 block-scales in swizzled format. It requires a compact per-expert format instead, so I stripped out all the swizzle changes out of this PR.

transformer_engine/common/cast/dispatch/dequantize.cuh
```diff
@@ -69,6 +70,11 @@ inline void group_dequantize_helper(const GroupedTensor &input, GroupedTensor *o
       }
       break;
     }
+    case NVTE_BLOCK_SCALING_1D:
+    case NVTE_BLOCK_SCALING_2D: {
+      fp8_blockwise::group_dequantize(&input, output, stream);
```
vthumbe1503, line 75: We  would need group dequantize tests as well for fp8 blockwise.
Alternatively it can be seperated out into a different PR, since group dequant is not a priority for GroupedLinear integration

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
```diff
@@ -0,0 +1,1022 @@
...
+    info.common_first_dim_blocks = common_first_dim / kTileDim;
+    info.R_total = info.num_tensors * common_first_dim;
+  } else {
+    info.common_first_dim_blocks = 0;
+    info.R_total = output->logical_shape.data[0];
+    info.tensor_offsets_d = reinterpret_cast<const int64_t*>(output->tensor_offsets.dptr);
+    NVTE_CHECK(info.tensor_offsets_d != nullptr,
+               "VARYING_FIRST_DIM requires tensor_offsets to be set on the GroupedTensor.");
+  }
+  info.total_row_blocks = DIVUP(info.R_total, static_cast<size_t>(kTileDim));
```
vthumbe1503, line 805: We might want to have static persistent kernel similar to mxfp8 in case of cuda graphs in future.

With cuda graphs total_rows can be larger than the total sum of first dims. And so we would be overlaunching thread blocks.

Not a blocker for the current PR. Persistent kernel based optimization can be a future PR if necessary.

vthumbe1503 (commented):
LGTM! We should probably move group dequantize to a different PR or add a test for it in this PR.

vthumbe1503 (dismissed):

transformer_engine/common/cast/dispatch/dequantize.cuh
denera: PR already includes `test_grouped_tensor.py::test_group_dequantize_fp8_blockwise` so the dequantize **is** tested, but we're missing `test_group_dequantize_cudagraph_capturable`. I'll add it for parity with grouped MXFP8.

vthumbe1503: Do we have CPP test in common though, considering JAX might also need it?

denera: We do now. Just pushed the commit with the new tests, both C++ and PyTorch, passing on H100.

No tests on the JAX side. The PR doesn't include any JAX changes. I punted that to a separate PR later this week because JAX needs a lot more changes to integrate FP8BS. Not quite as drop-in ready as PyTorch GroupedTensor.

transformer_engine/common/cast/fp8_blockwise/group_quantize_fp8_blockwise.cuh
denera: Sure, I'll turn that into a separate PR. Would be a self-contained change with no framework integration impact so it should be something very minimal to review and merge later.

vthumbe1503 (approved):

PR #3136, opened 2026-06-21
[Common] Fix int32 overflow in multi_tensor_apply tensor sizes for numel > INT_MAX

### LINKED ISSUE #2918: [BUG] multi_tensor_apply: int32 overflow in TensorListMetadata::sizes causes illegal memory access for tensors with numel > INT_MAX
`multi_tensor_apply` silently truncates per-tensor sizes from 64-bit to 32-bit, causing illegal memory access when any input tensor has `numel() > INT_MAX (2,147,483,647)`.

  In transformer_engine/common/multi_tensor/multi_tensor_apply.cuh, `TensorListMetadataBase::sizes` is declared as `int sizes[...] (int32)`, but it is populated from `Tensor::numel() (size_t / int64)`:

```
  // multi_tensor_apply.cuh:24
  int sizes[depth_to_max_tensors[n - 1]];
  ...
  // multi_tensor_apply.cuh:68
  tl.sizes[loc_tensor_info] = tensor_lists[0][t]->numel();   // int64 -> int32, silent truncation
```

For a tensor with numel = 2,476,250,368 (e.g. an embedding of shape [19345706, 128]), this field becomes 2476250368 - 2^32 = -1,818,716,928. The resulting negative / bogus size is then consumed by downstream kernels (e.g. multi_tensor_l2norm_kernel) which compute element offsets from it, producing out-of-bounds global-memory accesses and the following error at the next CUDA sync:

RuntimeError: .../multi_tensor_apply.cuh:92 in function multi_tensor_apply:
CUDA Error: an illegal memory access was encountered

This is hit by any real-world use that feeds a tensor with numel > 2^31 to TE's multi_tensor utilities. In particular, megatron.training.utils.calc_params_l2_norm →
multi_tensor_applier(multi_tensor_l2norm, ...) crashes for any model containing a single parameter with >2.14B elements (common for large-vocab embeddings, tied output layers, over-encoding tables, etc.).

javierdejesusda:
# Description

`TensorListMetadataBase::sizes` in `multi_tensor_apply.cuh` is declared as `int` (int32) but populated from `Tensor::numel()`, which is 64-bit. A tensor with `numel() > INT_MAX` truncates to a negative size, and the consumer kernels then compute out-of-bounds element offsets from it, raising an illegal memory access at the next CUDA sync. This is hit by any model with a single parameter over 2.14B elements (large-vocab embeddings, tied output layers) feeding TE's multi-tensor utilities, e.g. through Megatron's `calc_params_l2_norm`.

Fixes #2918

## Type of change

## Changes

- Store `TensorListMetadataBase::sizes` as `int64_t` in `multi_tensor_apply.cuh`.
- Read each size into an `int64_t` (or the existing templated `index_t` in the Adam kernels) before the `n -= chunk_idx * chunk_size` subtraction in `adam.cu`, `compute_scale.cu`, `l2norm.cu`, `scale.cu`, and `sgd.cu`.
- Widen the `chunk_size` kernel argument to `int64_t` in every consumer that took it as `int` (the `AdamCapturable`/`AdamCapturableMaster` functors plus the non-Adam kernels) so the `chunk_idx * chunk_size` element offset is computed in 64-bit. The already-templated Adam kernels keep `index_t` chunk_size.

No public API change. The widening is confined to the multi_tensor metadata and the kernel index arithmetic.

## Testing

Verified statically on a box without a GPU: the size and chunk_size paths are now 64-bit, so values above INT_MAX no longer wrap negative. I could not run the runtime repro here, since reproducing the `numel > INT_MAX` path needs a multi-GB GPU allocation. Happy to add a guarded large-tensor regression test for `multi_tensor_l2norm` if maintainers want one gated behind sufficient device memory.

Signed-off-by: Javier de Jesus <javier.dejesusj9@gmail.com>

### DIFF
--- a/transformer_engine/common/multi_tensor/adam.cu
+++ b/transformer_engine/common/multi_tensor/adam.cu
@@ -382,7 +382,7 @@ struct AdamFunctor {
 
 template <typename T, typename FULL_T, typename MOMENT_T>
 struct AdamCapturableFunctor {
-  __device__ __forceinline__ void operator()(int chunk_size, volatile int *noop_gmem,
+  __device__ __forceinline__ void operator()(int64_t chunk_size, volatile int *noop_gmem,
                                              TensorListMetadata<4> &tl,  // NOLINT(*)
                                              const float beta1, const float beta2, const int *step,
                                              const int bias_correction, const float epsilon,
@@ -402,7 +402,7 @@ struct AdamCapturableFunctor {
     // int tensor_num = tl.start_tensor_this_launch + tensor_loc;
 
     int chunk_idx = tl.block_to_chunk[blockIdx.x];
-    int n = tl.sizes[tensor_loc];
+    int64_t n = tl.sizes[tensor_loc];
 
     T *g = reinterpret_cast<T *>(tl.addresses[0][tensor_loc]);
     g += chunk_idx * chunk_size;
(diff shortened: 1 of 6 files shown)


vthumbe1503:
LGTM. Jax unit  test failure is prexisting in main.
Pipeline: 55407557

vthumbe1503 (approved):

PR #3143, opened 2026-06-25
Graph Safe Current Scaling Support for GroupedLinear Module/Ops + Fix CUBLAS GGEMM heuristics

vthumbe1503:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


tests/pytorch/test_grouped_mlp.py
```diff
@@ -479,6 +478,20 @@ def test_grouped_linear_cuda_graph_safe(
             pytest.skip("Grouped GEMM CUDA-graph-safe path requires SM100+ (Blackwell)")
```
denera: This is going to blanket skip these tests on Hopper even though they could run with FP8 current scaling.

Is that intentional for the CUDA graphed execution tests? Is there a reason why the implementation is not graph-safe, or was this just an oversight?

tests/pytorch/test_grouped_mlp.py
```diff
@@ -288,11 +288,7 @@ def test_grouped_linear(
...
         if single_grouped_bias and not bias:
             pytest.skip("single_grouped_bias requires bias=True")
-        if (
-            single_grouped_weight
-            and quantized_weight
-            and quantization in ("fp8_delayed_scaling", "fp8_current_scaling")
-        ):
+        if single_grouped_weight and quantized_weight and quantization in ("fp8_delayed_scaling"):
             pytest.skip(
                 "single_grouped_weight does not support FP8 delayed/current scaling "
```
denera: The skip test now correctly continues for `fp8_current_scaling` but the skip message still includes `delayed/current` in the text. Should be corrected to just `delayed` instead since that's the new sole skip condition.

denera (changes requested):
LGTM except for two minor fixes/clarifications in the GroupedMLP tests.

tests/pytorch/test_grouped_mlp.py
vthumbe1503: Yes, this was an oversight. Relaxed the guard now. Thanks for catching it.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -1869,12 +1869,14 @@ void nvte_grouped_gemm_with_discrete_out(const NVTEGroupedTensor A, int transa,
...
+  // avg n hints should use A.
   gemm_config.avg_m =
-      config_.avg_m.value_or(transa ? compute_avg_last_dim(inputA) : compute_avg_first_dim(inputA));
+      config_.avg_m.value_or(transb ? compute_avg_last_dim(inputB) : compute_avg_first_dim(inputB));
   gemm_config.avg_n =
-      config_.avg_n.value_or(transb ? compute_avg_first_dim(inputB) : compute_avg_last_dim(inputB));
+      config_.avg_n.value_or(transa ? compute_avg_first_dim(inputA) : compute_avg_last_dim(inputA));
   gemm_config.avg_k =
-      config_.avg_k.value_or(transa ? compute_avg_first_dim(inputA) : compute_avg_last_dim(inputA));
+      config_.avg_k.value_or(transa ? compute_avg_last_dim(inputA) : compute_avg_first_dim(inputA));
```
timmoon10: This doesn't seem right to me:
- `avg_k` is now inconsistent between `nvte_grouped_gemm_with_discrete_out` and `nvte_grouped_gemm`/`nvte_grouped_gemm_with_discrete_inputA`.
- cuBLAS expects column-major matrices and the untransposed matmul looks like `A (m x k) x B (k x n) = C (m x n)`. [The dim logic for single GEMMs](https://github.com/NVIDIA/TransformerEngine/blob/4cd244eee50b2ac3b12ed2f77ac02964ca99f6c5/transformer_engine/common/gemm/cublaslt_gemm.cu#L329-L331) looks like:
```c++
m = transa ? A.flat_first_dim() : A.flat_last_dim();  // or C.flat_last_dim()
n = transb ? B.flat_last_dim() : B.flat_first_dim();  // or C.flat_first_dim()
k = transa ? A.flat_last_dim() : A.flat_first_dim();
// or k = transb ? B.flat_first_dim() : B.flat_last_dim();
```

It's odd that the grouped tensor avg logic is consistently swapping first-dim and last-dim, and I wonder if that is a bug. All this switching between row-major order and column-major order makes it hard to keep all this in my head.

vthumbe1503: Thanks for catching this inconsistency. The other two nvte APIs are actually also broken in terms of avg_k caclulation.

I have fixed the the other 2 nvte APIs as well. Will do a final perf check.. because it seems like we have been living with a bad cublas ggemm perf for a while in TE

timmoon10 (approved):
LGTM, pending CI and perf checks.

denera (approved):
LGTM as well!

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
vthumbe1503: Did a perf check on variety of sizes, and group imbalances.. and didnt find any difference in perf with and without the heuristics fix. So fortunately we werent affected.

PR #3152, opened 2026-06-29
[PyTorch][torch.compile] Make quantizers opaque value objects

pggPL:
# Description
Tensorless quantizers in TE (MXFP8, FP8 blockwise, FP8 current-scaling, NVFP4)
are fully described by a handful of plain, reproducible scalars — they hold no
live tensors and no process groups. This PR turns them into **opaque value
objects** so `torch.compile` can treat them as baked-in constants: two
quantizers with the same configuration become interchangeable, hashable, and
reconstructible inside an FX graph.

Quantizers that hold live state (delayed-scaling `Float8Quantizer`, which keeps
`scale`/`amax` tensors) and any user-defined quantizer keep the default
identity semantics, so the change is opt-in and backward compatible. On older
PyTorch builds without the opaque-object API the registration is a graceful
no-op.

Along the way this also un-breaks the existing `test_torch_compile.py` suite:
that file lived on `main` but was never wired into CI, and its
`test_autocast_nested_custom` case (nested `te.autocast` with multiple
`CustomRecipe` instances) was failing because of the `CustomRecipe` state-caching
bug fixed here. The file is now run in CI and passes.

## Type of change
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add opt-in value-object identity to the base `Quantizer`
  (`_value_fields` / `_value_key` / `__eq__` / `__hash__`). Returning `None`
  from `_value_fields()` (the default) keeps identity semantics.
- New module `transformer_engine/pytorch/dynamo.py` holding the
  `torch.compile` glue: `__fx_repr__`, value-key reconstruction and
  `register_value_opaque_quantizer` (gracefully no-op without PyTorch's
  opaque-object API).
- Register `MXFP8Quantizer`, `Float8BlockQuantizer`,
  `Float8CurrentScalingQuantizer` and `NVFP4Quantizer` as value opaque types
  (the deprecated `amax_reduction_group` is never part of the value).
- Fix `CustomRecipe` state caching in `TransformerEngineBaseModule.set_meta_tensor`:
  rebuild quantizers when the `CustomRecipe` *instance* changes (e.g. nested
  `te.autocast` regions) instead of reusing the first recipe's state, since
  every `CustomRecipe` shares the `CustomRecipeState` type but carries its own
  `qfactory`. This fixes the previously-failing `test_autocast_nested_custom`.
- Enable `tests/pytorch/test_torch_compile.py` in the `L0_pytorch_unittest` QA
  suite (it existed on `main` but was never run in CI), and add the quantizer
  value-object tests to it. Bringing it into CI required fixing the existing
  `CustomRecipe` torch.compile path: the `qfactory` now dispatches on
  `QuantizerRole.tensor_type` supplied by `ToyLinear.get_quantizer_roles`.
- Guard the value-object path against a stored amax reduction group: `__fx_repr__`
  already rejects any quantizer holding a process group, and `__eq__` / `__hash__`
  now raise too. The group is excluded from the value key, so a stored group would
  otherwise compare/hash equal to a groupless quantizer and let `torch.compile`
  reuse a graph that skips the reduction. Pass the group per quantize call instead.

# Checklist:


transformer_engine/pytorch/dynamo/quantizer_opaque.py
```diff
@@ -0,0 +1,139 @@
...
+    :class:`transformer_engine.pytorch.quantized_tensor.Quantizer`).
+    """
+    # Stamp the class so it can be recognized as value-opaque in dynamo-traced
+    # code (used to fall back to eager for unregistered quantizers).
+    setattr(cls, _VALUE_OPAQUE_FLAG, True)
+
+    # ``register_opaque_type`` requires ``__fx_repr__`` to already exist on the
+    # class, so attach it before registering.
+    if "__fx_repr__" not in cls.__dict__:
+        cls.__fx_repr__ = _quantizer_fx_repr
```
kshitij12345, line 113: If a Quantizer implements `__fx_repr__`, then it may skip the `_contains_process_group` check. I think `_contains_process_group` could be moved outside the `__fx_repr__`.

transformer_engine/pytorch/dynamo/quantizer_opaque.py
```diff
@@ -0,0 +1,139 @@
...
+    Raises ``TypeError`` if the quantizer stores a process group (e.g. a
+    non-``None`` deprecated ``amax_reduction_group``): live distributed state
+    must never be baked into the graph as a constant, so such a quantizer cannot
+    be used with ``torch.compile``. Pass the reduction group per quantize call
+    instead of storing it on the quantizer.
+    """
+    cls = type(self)
+    for name, value in vars(self).items():
+        if _contains_process_group(value):
+            raise TypeError(
```
kshitij12345: It would be good to add a test enforcing this error.

kshitij12345 (approved):
Overall LGTM, would be good to resolve the inline comments before merging.

tests/pytorch/test_torch_compile.py
```diff
@@ -384,3 +388,176 @@ def fn(inp):
...
+def test_quantizer_value_object(factory, other_kwargs):
+    """Value semantics + ``__fx_repr__`` round-trip via the production FX path."""
+    a, b = factory(), factory()
+    # Same config -> equal, same hash, interchangeable as a dict/set key.
+    assert a is not b
+    assert a == b
+    assert hash(a) == hash(b)
+    assert {a: "x"}[b] == "x"
+    # Different config -> not equal.
+    assert a != factory(**other_kwargs)
```
ptrendx: What exactly are we testing in this part?

tests/pytorch/test_torch_compile.py
```diff
@@ -384,3 +388,176 @@ def fn(inp):
...
+        q = MXFP8Quantizer(fp8_dtype=tex.DType.kFloat8E4M3)
+        q.amax_reduction_group = dist.group.WORLD
+        # Every value-materialization path must reject it (hash, eq, __fx_repr__).
+        with pytest.raises(TypeError):
+            hash(q)
+        with pytest.raises(TypeError):
+            q.__fx_repr__()
+    finally:
+        if created:
+            dist.destroy_process_group()
```
ptrendx, line 498: I'm not sure I understand the purpose of this test (and the logic needed to make it work). Is it here to make sure that if somebody accidentally adds the group to the quantizer then it would loudly fail e.g. in the CI?

transformer_engine/pytorch/dynamo/quantizer_opaque.py
```diff
@@ -0,0 +1,116 @@
...
+
+def _rebuild_quantizer(cls: type, items: Tuple[Tuple[str, Any], ...]) -> Any:
+    """Rebuild a tensorless quantizer of type *cls* from its value items.
+
+    Referenced by the ``__fx_repr__`` emitted for value-opaque quantizers; the
+    generated FX code calls this to materialize the quantizer constant.
+    """
+    # Bypass ``__init__`` and restore the value attributes directly: the value
+    # items already capture every value-defining field (including derived ones),
+    # and the constructors have heterogeneous signatures / side effects.
```
ptrendx, line 41: Huh? What are the issues with just calling the constructor?

transformer_engine/pytorch/dynamo/quantizer_opaque.py
```diff
@@ -0,0 +1,116 @@
...
+    versions without the opaque-object API it only attaches ``__fx_repr__``
+    (harmless), so Transformer Engine keeps importing and running in eager mode.
+
+    The quantizer class must already provide value ``__eq__`` / ``__hash__`` and
+    a non-``None`` ``_value_fields`` (see
+    :class:`transformer_engine.pytorch.quantized_tensor.Quantizer`).
+    """
+    # Stamp the class so it can be recognized as value-opaque in dynamo-traced
+    # code (used to fall back to eager for unregistered quantizers).
+    setattr(cls, _VALUE_OPAQUE_FLAG, True)
```
ptrendx: Is this value used anywhere else? If so, shouldn't it happen at the end of that function (since things can throw afterwards) once we know that everything is ok?

transformer_engine/pytorch/dynamo/quantizer_opaque.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # ``register_opaque_type`` requires ``__fx_repr__`` to already exist on the
+    # class, so attach it before registering.
+    if "__fx_repr__" not in cls.__dict__:
+        cls.__fx_repr__ = _quantizer_fx_repr
```
ptrendx, line 113: Then why aren'twe just adding that to the Quantizer class? I feel that I miss some context here as a lot of this file feels like a workaround to some issues that are not described.

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -408,6 +422,78 @@ def get_usages(self) -> Dict[str, bool]:
...
+
+        Returning ``None`` (the default) means the quantizer cannot be represented as
+        a value opaque object and keeps identity-based equality/hashing.
+        This also means that passing such a quantizer as an argument to a custom op
+        causes a graph break under torch.compile, since it cannot be baked into the
+        FX graph as a constant.
+        """
+        return None
+
+    def _check_value_has_no_process_group(self) -> None:
```
ptrendx, line 438: How often is this method called? Only once during torch.compile or does it need to be called every time during the graph launch to check the guards?

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -69,6 +70,9 @@ def copy(self) -> Float8BlockQuantizer:
 
         return quantizer
 
+    def _value_fields(self) -> Tuple[str, ...]:
+        return ("dtype", "block_len", "amax_epsilon", "force_pow_2_scales", "block_scaling_dim")
```
ptrendx: There are some ways of getting that list without hardcoding it again, e.g. using `inspect.get_annotations` in the cases like we have where we provide all of the data members up front or via vars.

transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -184,6 +186,16 @@ def __getstate__(self):
...
 
+    def _rebuild_derived_state(self) -> None:
+        """Restore the derived ``rht_matrix`` after value-key reconstruction.
+
+        ``rht_matrix`` is a ``torch.Tensor`` built from ``_with_random_sign_mask``
+        and the device, so it cannot be part of the (hashable) value key.
+        ``_rebuild_quantizer`` calls this hook to rebuild it; the ``lru_cache`` on
+        :func:`get_rht_matrix` makes an already-seen (flag, device) a cheap hit.
+        """
+        self.rht_matrix = get_rht_matrix(self._with_random_sign_mask, torch.cuda.current_device())
```
ptrendx: Using current device here is a little annoying since in principle it does not have to be the right device (which instead should be the input device).

pggPL: It's not problem of this PR, we use current device in constructor.

transformer_engine/pytorch/dynamo/quantizer_opaque.py
pggPL: Many things have different names in constructor than as parameters or are not present as args (like ".internal" or ".optimize_for_gemm")

tests/pytorch/test_torch_compile.py
pggPL: It was suggestion of @kshitij12345

tests/pytorch/test_torch_compile.py
pggPL: nothing, I can remove it it seems to be claude boilerplate

transformer_engine/pytorch/dynamo/quantizer_opaque.py
pggPL: We can do this, I wanted to have as much of torch.compile specific logic here

transformer_engine/pytorch/dynamo/quantizer_opaque.py
pggPL: good catch, thanks

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
pggPL: I see It's quite messy, but it is result of mess in quantizers. There are some fields called in the constructor (and arguments names sometimes does not match the field name). Some fields are set like .internal or .block_len are not arguments to constructor. And some of them are excluded on purpose like `amax_reduction_group` or `rht_matrix`.

transformer_engine/pytorch/quantized_tensor.py
pggPL: every time - I make it faster by just checking amax_reduction_group

tests/pytorch/test_torch_compile.py
kshitij12345: Context around the test - There was a possible code-path previously (now fixed) where having a ProcessGroup on the quantizer wouldn't trigger the unsupported TypeError (and fail later). This test verifies that the error would be raised with a clear message.

transformer_engine/pytorch/dynamo/quantizer_opaque.py
```diff
@@ -0,0 +1,114 @@
...
+from typing import Any, Dict, Tuple
+
+from ..constants import DType
+
+
+# Registration marks the class with this attribute rather than recording it in a
+# module-level set. It looks odd but is a deliberate workaround: the check must
+# stay traceable when it runs inside a torch.compile graph -- Dynamo can bake a
+# ``getattr`` on the opaque quantizer into a constant, but cannot evaluate
+# ``type(q) in some_set`` (no equality/hash rules for the opaque class object),
```
shino16: This sounds awkward to me, since `type(q)` is a regular Python class. Is a normal module-level set tested to be insufficient?

transformer_engine/pytorch/dynamo/quantizer_opaque.py
```diff
@@ -0,0 +1,114 @@
...
+            value = DType.cast(value)
+        object.__setattr__(obj, name, value)
+        field_names.add(name)
+    # The deprecated amax-reduction group is not a value field; initialize it to
+    # None so attribute access keeps working on the rebuilt quantizer.
+    if "with_amax_reduction" in field_names and not hasattr(obj, "amax_reduction_group"):
+        object.__setattr__(obj, "amax_reduction_group", None)
+    # Restore non-value derived state that ``__init__`` would normally build but
+    # that cannot live in the value key (e.g. NVFP4's ``rht_matrix`` tensor).
+    finalize = getattr(obj, "_rebuild_derived_state", None)
```
shino16, line 49: What about adding a no-op default `_rebuild_derived_state` in the Quantizer base class? It's just a minor style point though.

transformer_engine/pytorch/tensor/float8_tensor.py
```diff
@@ -386,6 +387,15 @@ def supports_only_rowwise_all_gather(self) -> bool:
         """
         return True
 
+    def _value_fields(self) -> Tuple[str, ...]:
+        # ``amax_reduction_group`` is intentionally excluded: it is a deprecated
+        # process group (not a value). If one is actually stored, ``__fx_repr__``
+        # raises so it can never be baked into a torch.compile graph.
```
shino16: This sounded confusing. Trying to restore `amax_reduction_group` should raise an error, but `__fx_repr__` (`_rebuild_quantizer`) has a safeguard.

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -1061,3 +1065,6 @@ def backward(
             )
             return dgrad, None
         return grad.view(ctx.shape), None
+
+
+register_value_opaque_quantizer(MXFP8Quantizer)
```
shino16: We can put this right after `MXFP8Quantizer` as other classes do.

transformer_engine/pytorch/dynamo/quantizer_opaque.py
shino16: Uh, this makes sense.
> We can do this, I wanted to have as much of torch.compile specific logic here
> https://github.com/NVIDIA/TransformerEngine/pull/3152#discussion_r3501991776

tests/pytorch/test_torch_compile.py
```diff
@@ -24,14 +24,18 @@
 from transformer_engine.common import recipe
 from transformer_engine.pytorch.constants import FP8FwdTensorIdx, FP8BwdTensorIdx
 from transformer_engine.pytorch.module.base import TransformerEngineBaseModule
+from transformer_engine.pytorch.quantization import QuantizerRole
 from transformer_engine.pytorch.ops.basic.basic_linear import BasicLinear
 from transformer_engine.pytorch.tensor.float8_tensor import Float8CurrentScalingQuantizer
-from transformer_engine.pytorch.quantization import QuantizerRole
```
shino16: Unintended diff?

tests/pytorch/test_torch_compile.py
```diff
@@ -24,14 +24,18 @@
...
+from transformer_engine.pytorch.tensor.nvfp4_tensor import NVFP4Quantizer
```
shino16: `NVFP4Quantizer` can be imported from `transformer_engine.pytorch` like other quantizers are.

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
ptrendx: Shouldn't we then try to clean the quantizers code instead of adding to the messiness? I think the quantizers should advertise properly all of the fields that the value object would need.
I don't think internal is a problem necessarily (although we could just as well have it in the quantizer constructor, it seems that right now it is very easily discoverable), since you also do not put it in this function, and I assumed what you want here are just the constructor arguments from this subclass?
The ones that we want to "deprecate" in a sense like the amax_reduction_group should be removed from the annotated fields, right?

ptrendx (dismissed):

ptrendx (approved):

PR #3153, opened 2026-06-29
[PyTorch][torch.compile] Add TensorProto mechanism

pggPL:
# Description

This PR introduces **`TensorSpec`** — a *data-free* description of a tensor (or quantized tensor) that captures everything needed to rebuild it without holding any storage: its logical `shape`/`dtype` and, for quantized tensors, the value-opaque `quantizer` that defines the buffer layout.

The key property is that `TensorSpec.create_tensor()` materializes a quantized tensor **purely in Python** — via `Quantizer.alloc_tensors` plus the storage's `__tensor_unflatten__` — so it traces under `torch.compile(fullgraph=True)` with no graph break, unlike `make_empty`, which goes through the opaque C++ `tex.create_empty_quantized_tensor`. This is the foundation for writing `torch.library` custom-op *fake implementations* of quantized ops; the consumers land in the follow-up Linear custom-op PR.

This builds on the value-opaque quantizer work, so a `TensorSpec` is itself safe to treat as a compile-time constant.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

**`dynamo/tensor_spec.py`** (new) — `TensorSpec` dataclass (`shape`, `dtype`, `quantizer`, `requires_grad`, `device`) with `is_quantized`, `update_usage()`, `inner_names()`, `create_metadata()`, `create_inner_tensors()`, `assemble()` and `create_tensor()`, plus a `to_tensor_spec()` helper that builds a spec from a plain `torch.Tensor`, a `QuantizedTensorStorage` or a `QuantizedTensor`. Exported from `transformer_engine.pytorch.dynamo`.

**`quantized_tensor.py`**
- Add the PyTorch wrapper-subclass flatten protocol (`__tensor_flatten__` / `__tensor_unflatten__`) to `QuantizedTensorStorage`.
- Declare the flat buffers on the field itself: `_scale_inv: Annotated[torch.Tensor, InnerTensor("fp8_scale_inv")]`. `__init_subclass__` collects these into `_INNER_TENSORS` in field order, so the attribute-to-constructor-kwarg mapping lives next to the attribute instead of in a parallel tuple.
- The flatten context carries the storage class itself, so `__tensor_unflatten__` needs no registry lookup.
- Add pure-Python, traceable allocation primitives to `Quantizer`: `inner_tensor_specs` (buffer geometry), `storage_metadata` (concrete class + non-tensor constructor kwargs), and the `alloc_tensors` / `create_metadata` built on top of them. The base implementations raise `NotImplementedError`, so a quantizer that does not implement them simply cannot be used with `TensorSpec`.
- Add a `shape` property that is valid on bare storages as well as wrapper tensors.

**Quantizers** — implement `inner_tensor_specs` and `storage_metadata` for `Float8CurrentScalingQuantizer`, `MXFP8Quantizer`, `Float8BlockQuantizer` and `NVFP4Quantizer`. The FP8 description mirrors the C++ allocation in `csrc/quantizer.cpp`, including the non-TN-capable-arch case where a single `_data` buffer backs both directions.

**Storage classes** — declare `InnerTensor` fields for `Float8TensorStorage`, `MXFP8TensorStorage`, `Float8BlockwiseQTensorStorage` and `NVFP4TensorStorage`.

**`module/base.py`** — override `nn.Module._apply` in `TransformerEngineBaseModule`. This is a *consequence* of the flatten protocol, not part of the new API: once a parameter implements it, `_apply` moves the parameter with `torch.utils.swap_tensors`, which exchanges its whole `__dict__`. Inner buffers ride across correctly, but state attached from the outside (`_high_precision_init_val` and its accessors, `main_grad`, user attributes) would be left behind on the discarded tensor. The override snapshots those attributes and re-attaches the ones the swap did not carry over, restoring the pre-PR behaviour of `.to()` / `.cuda()` / `.half()`.

**Tests**
- `tests/pytorch/test_torch_compile.py`: quantizer primitives under `FakeTensorMode`, storage flatten/unflatten round-trip, `TensorSpec` behaviour in eager and fake mode, `fullgraph=True` tracing, and `to_tensor_spec` round-trips — across FP8 current scaling, MXFP8, FP8 blockwise and NVFP4.
- `test_python_alloc_matches_cpp_make_empty` builds the same tensor twice, via `make_empty` (C++) and via the Python primitives, then checks structural parity (class, buffer set, per-buffer shape/dtype/device, logical shape/dtype, flatten context) **and** functional parity — the real quantize kernel writes bit-identical results into both — across quantizer families x rowwise/columnwise x wrapper/internal.
- `tests/pytorch/test_sanity.py`: attributes attached to a quantized parameter survive `nn.Module._apply` for `.cuda()`, `.cpu()` and `.half()`.

## Known limitations

- The flatten protocol covers the four storage classes listed above. `HybridQuantizedTensorStorage` and `IdentityTensorStorage` declare no `InnerTensor` fields, so flattening them raises rather than silently passing their buffers through the context; hybrid storage holds nested storages rather than flat buffers, which the current model does not express.
- `HybridQuantizer` and `IdentityQuantizer` are not registered as value-opaque quantizers.
- Single-device only; tensor/sequence-parallel shape effects are not modelled.

# Checklist:


transformer_engine/pytorch/tensor/nvfp4_tensor.py
```diff
@@ -345,6 +345,55 @@ def _canonicalized_amax_reduction_group(self) -> dist_group_type:
...
+                "quantizer": self,
+                "with_gemm_swizzled_scales": self.optimize_for_gemm,
+                "row_scaled_nvfp4": self.row_scaled_nvfp4,
+                "nvfp4_use_4over6": self.nvfp4_use_4over6,
+                "nvfp4_e4m3_max": self.nvfp4_e4m3_max,
+                "fake_dtype": fake_dtype,
+            },
+        }
+
+    def _describe_buffers(
```
ptrendx: I do not see any tests that would ensure that this and the actual C++ allocation produces compatible
results.

pggPL: Done — added `test_python_alloc_matches_cpp_make_empty`: builds the same quantized tensor via `make_empty` (C++) and via the Python primitives, then checks structural parity (class, buffer set, per-buffer shape/dtype/device, flatten context) and functional parity (the real quantize kernel writes bit-identical results into both, dequantize matches), across quantizer families × rowwise/columnwise × wrapper/internal. Formats not supported on local HW are covered by the same test on CI.

transformer_engine/pytorch/module/linear.py
```diff
@@ -83,6 +84,11 @@
 __all__ = ["Linear"]
 
 
+# Fields with this union may hold a *bare* ``QuantizedTensorStorage`` (see its
+# docstring), not just a plain / subclass tensor. The annotation is also
+# machine-read when the args bag becomes a torch.compile custom op (follow-up
+# PR): such fields get flatten/unflatten slots so a bare storage can cross the
```
shino16: Note: We will update this comment as part of the custom op PR.

shino16 (approved):

tests/pytorch/test_torch_compile.py
```diff
@@ -820,3 +830,381 @@ def fn(inp):
...
+    pytest.param(
+        _nvfp4,
+        (64, 128),
+        id="nvfp4",
+        marks=pytest.mark.skipif(
+            not nvfp4_available,
+            reason="NVFP4 is not available",
+        ),
+    ),
+]
```
ptrendx: Why is nvfp4 skipped when not available but not MXFP8?

pggPL: Good catch - the asymmetry was a leftover and is gone now. There is no `skipif` mark on any of the `_PROTO_QUANTIZERS` params anymore: the list is pure Python (shapes/dtypes/metadata) and works on any HW, and the tests that actually run a kernel gate every format uniformly through `_skip_if_dequantize_unsupported` (mxfp8 / nvfp4 / fp8 blockwise / fp8 current scaling).

tests/pytorch/test_torch_compile.py
```diff
@@ -820,3 +819,379 @@ def fn(inp):
...
+
+    ``dequantize()`` runs the real kernel on CUDA, so each format has its own
+    availability gate (mirrors the ``is_*_available`` checks in test_numerics).
+    """
+    if isinstance(q, MXFP8Quantizer):
+        if not mxfp8_available:
+            pytest.skip(reason_for_no_mxfp8)
+    elif isinstance(q, NVFP4Quantizer):
+        if not nvfp4_available:
+            pytest.skip("NVFP4 is not available")
```
ptrendx: Could we add (or take from somewhere else, since I think it is already defined) reason_for_no_nvfp4
or similar?

tests/pytorch/test_torch_compile.py
```diff
@@ -820,3 +819,379 @@ def fn(inp):
...
+        # eliminated as dead code -- forces the whole build path into the graph.
+        acc = x.new_zeros(())
+        for name in names:
+            acc = acc + getattr(t, name).float().sum()
+        return acc
+
+    x = torch.zeros(*shape, dtype=torch.bfloat16)
+    torch._dynamo.reset()
+    out = torch.compile(fn, fullgraph=True)(x)
+    assert out.shape == ()
```
ptrendx: There is already a test here that does the same thing with TensorProto, so why having both of them
in?

tests/pytorch/test_torch_compile.py
```diff
@@ -820,3 +819,379 @@ def fn(inp):
...
+    tensor with the right shape/dtype and fake inner buffers."""
+    q = factory()
+    proto = TensorProto(shape=shape, dtype=torch.bfloat16, quantizer=q, device=torch.device("cpu"))
+    with FakeTensorMode():
+        out = proto.create_tensor()
+    assert isinstance(out, QuantizedTensor)
+    assert tuple(out.shape) == tuple(shape)
+    assert out.dtype == torch.bfloat16
+    for name in proto.inner_names():
+        assert isinstance(getattr(out, name), FakeTensor)
```
ptrendx: Shouldn't this be a single test with 2 variants instead?

transformer_engine/pytorch/dynamo/tensor_proto.py
```diff
@@ -0,0 +1,164 @@
...
+import copy as _copy
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional, Tuple
+
+import torch
+from torch._prims_common import make_contiguous_strides_for
+
+
+@dataclass
+class TensorProto:
```
ptrendx: A more general question - considering that PyTorch went with Tensor/FakeTensor naming, shouldn't we
follow suit with QuantizedTensor/FakeQuantizedTensor rather than introducing a completely new name?

transformer_engine/pytorch/dynamo/tensor_spec.py
```diff
@@ -0,0 +1,164 @@
...
+        """Materialize the flat inner buffers (in :meth:`inner_names` order).
+
+        Under ``register_fake`` the ``torch.empty`` calls produce ``FakeTensor``s;
+        ``requires_grad`` is left default (managed by ``register_autograd``).
+        """
+        device = self.device if self.device is not None else torch.device("cuda")
+        if self.quantizer is None:
+            return [torch.empty(tuple(self.shape), dtype=self.dtype, device=device)]
+        inner = self.quantizer.alloc_tensors(tuple(self.shape), device=device)
+        return [inner[name] for name in self.inner_names()]
```
ptrendx, line 111: Why do we need to reorder them here? Shouldn't they be already in the right order coming from
the quantizer's method?

transformer_engine/pytorch/dynamo/tensor_proto.py
pggPL: Not really, because TensorProto represents both Tensor and QuantizedTensor. I changed the name to TensorSpec.

ptrendx (approved):

PR #3160, opened 2026-07-01
[Common][PyTorch] Add strided batched GEMM in BF16/MXFP8

yaox12:
# Description

This PR adds experimental cuBLASLt-backed strided batched GEMM support and a new fusible PyTorch operation exported as `transformer_engine.pytorch.ops.BatchedLinear`.

The main use case is applying one weight matrix per group without rearranging interleaved activations:

```
[S, B, G, D] @ [G, R, D] -> [S, B, G, R]
```

Both contiguous `[G, ..., D]` and `[..., G, D]` input layouts are supported through explicit leading dimensions and batch strides. MXFP8 scales are packed into the cuBLASLt per-batch layout internally before each GEMM.

## Dependency

This PR depends on #3327 for meta-device deferred initialization in the operation fuser and should be merged after it.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Add an experimental common API for cuBLASLt strided batched GEMM with independent matrix strides.
- Support high-precision and MXFP8 input pairs with high-precision output.
- Add `transformer_engine.pytorch.ops.BatchedLinear` with high-precision and MXFP8 forward and backward computation.
- Support primary MXFP8 weights, bias and `return_bias`, `save_original_input`, custom initialization and RNG tracking, and main-gradient accumulation.
- Add API documentation and PyTorch L0 tests.

### Current limitations

- A and B must both be high precision or both be MXFP8.
- MXFP8 operands must use contiguous `[G, ..., D]` or `[..., G, D]` layouts.
- Output quantization, fused GEMM epilogues, non-MXFP8 recipes, and tensor parallelism are not supported.

# Checklist


transformer_engine/pytorch/module/batched_linear.py
timmoon10, line 1: We should prefer implementing new modules in `te.ops` so that they work with the op fuser API (see [docs](https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/examples/op_fuser/op_fuser.html)). `te.module` is somewhat a legacy API.

timmoon10 (commented):
Mathematically, is this equivalent to `GroupedLinear` with uniform split sizes (`m_splits=[S*B] * G`)?

If so, are there perf reasons to support a dedicated kernel? I could imagine that batched GEMM is easier to optimize than grouped GEMM, but grouped GEMM should already be getting a good fraction of the peak compute throughput. I would like to make sure we have a compelling benefit before accepting a major maintenance burden (on the order of `Linear` and `GroupedLinear`, especially if we need to support all the weird Mcore-specific integrations).

transformer_engine/pytorch/module/batched_linear.py
yaox12: In fact I implemented it in `te.ops` at first, but later I found `te.ops` lacks `rng_state_tracker_function`, `init_method`, and especially `preserve_high_precision_init_val`, which requires a cross-module change to all ops that hold weights.  The current usage of `ops.GroupedLinear` doesn't encounter these problems because we first initialize the weights tensors in `modules.GroupedLinear`, then assign them to `ops.GroupedLinear` and then do the computation. That's to say `ops.GroupedLinear` doesn't really hold the weight tensors. But for `ops.BatchLinear`, we don't have a counter-part `modules.*`.

yaox12: > Why not using grouped GEMM instead?

The main reason is that we need a **strided** batched GEMM kernel because the A tensor (input activation) is contiguous in `[S, B, G, D]` where `G` is the batch/group dim.

yaox12: Also, codex found a potential deferred initialization bug of `te.ops`. https://github.com/NVIDIA/TransformerEngine/issues/3322

yaox12: Migrated it to `te.ops`.

PR #3161, opened 2026-07-01
 [PyTorch] Add optional caller-provided output/grad-input buffers to GroupedLinear and fused grouped MLP

phu0ngng:
## Description

Let callers have an option to supply preallocated buffers for the forward output and backward input gradient of `GroupedLinear` and the fused grouped MLP, instead of the module/op allocating them internally. The caller can pass a symmetric-memory buffer that is fed directly into TE EP's `combine`, eliminating the D2D copy from the token buffer into TE EP's internal staging buffer.

## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [ ] Breaking change
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- `GroupedLinear.forward`: optional `out` / `dgrad_out` buffers, wired through the legacy and GroupedTensor paths (incl. FP8). Validated via a shared `validate_or_alloc_output` helper (shape/dtype/device/contiguous/no-grad).
- Fusible ops: `Sequential.forward` takes an `op_kwargs` mapping (keyed by module or index) that routes per-op kwargs to the target op, used to pass output/grad-input buffers to grouped linear / grouped MLP.
- Fused grouped MLP: the GEMMs write the output and dgrad directly into the caller buffers for both nvfp4 and MXFP8, with no copy. The GEMM writes only the packed valid rows (per `m_splits`); any padded tail is left untouched. Buffers may be supplied independently.

## Dependency

The in-place MXFP8 path requires a cuDNN frontend change that exposes an optional output tensor on the grouped-GEMM quant wrapper: https://github.com/NVIDIA/cudnn-frontend/pull/338.

## Testing

- `test_grouped_linear_caller_output_buffers`: covers out/dgrad_out/both across the legacy and GroupedTensor paths _ buffer aliasing, bit-exact match vs. internal allocation, untouched padded tail, and `ValueError` on shape mismatch.
- `test_grouped_linear_caller_buffers` (ops) and `test_grouped_mlp_caller_buffers`: verify the fused op runs, the output aliases the buffer, and grads match internal allocation.

# Checklist:


transformer_engine/pytorch/ops/sequential.py
```diff
@@ -168,13 +172,29 @@ def forward(
         self,
         input: torch.Tensor,  # pylint: disable=redefined-builtin
         *extra_inputs: torch.Tensor,
+        output: Optional[torch.Tensor] = None,
+        grad_input: Optional[torch.Tensor] = None,
```
timmoon10: I'm skeptical of this API because it looks general, but is actually only supported with grouped linear and grouped MLP. How about something like the following:
```python
def forward(
    self,
    input: torch.Tensor,
    *extra_inputs: torch.Tensor,
    extra_kwargs: Optional[Sequence[Optional[dict[str, Any]]]] = None,
):
    ...

    if extra_kwargs is None:
        extra_kwargs = [None] * len(self)

    # Figure out what extra_kwargs go to each module group, and fail if a non-None extra_kwargs goes to a non-basic op

    for module_group in self._module_groups:
            xs = module_group(*xs, basic_op_kwargs=module_group_basic_op_kwargs)
```

transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -1422,7 +1431,19 @@ def fuser_forward(
...
+            # The kernel output is non-contiguous and must be materialized. When a
+            # caller buffer is given, materialize into it directly (no extra copy).
+            if output_buffer is None:
+                fc2_out = fc2_out.contiguous()
+
+        # Route the forward output into the caller-provided buffer. This is free
+        # when the value above was materialized into it; otherwise it is copied.
+        if output_buffer is not None:
+            output_buffer = validate_or_alloc_output(output_buffer, fc2_out_shape, dtype, device)
+            if output_buffer.data_ptr() != fc2_out.data_ptr():
```
YangFei1990: Under what case this condition can be false? It seems to be it is always true, I do not see we feed the user provided buffer to the kernel?

transformer_engine/pytorch/ops/sequential.py
phu0ngng: Thanks, Tim. Good call. The `output` and `grad_input` kwargs overpromised generality. Building on your suggestion, I went with a generic per op kwargs pass through, but keyed by module or index instead of a positional list:
```
seq(x, split_sizes, probs, split_sizes,
    op_kwargs={fc1: {GRAD_INPUT_BUFFER_KEY: g}, fc2: {OUTPUT_BUFFER_KEY: o}})
```
Sequential resolves each key to its group and basic op slot and forwards it as `basic_op_kwargs`, with no new dispatch path, and raises if a key targets a non-fusible op. This keeps the container kwarg-agnostic, so the buffers are now just op specific kwargs, while avoiding None padding and positional coupling.

Let me know what you think about this new design.

transformer_engine/pytorch/ops/fused/grouped_mlp.py
phu0ngng: You were right that it was always true for MXFP8, which uses CuTe GEMM.

I pushed the latest commit in which I reworked it so that it does not trigger extra copies: the caller buffer is passed straight into the CuTe GEMM via `d_tensor`, so the kernel writes output/dgrad in place for both nvfp4 and MXFP8. This needs a cuDNN FE change I made a PR for: https://github.com/NVIDIA/cudnn-frontend/pull/338 (until it lands and the bundled FE is bumped, the in-place MXFP8 path can't run). So for testing, we will need to check out this cuDNN FE manually.

phu0ngng: Hi, the cuDNN FE PR is merged. I will follow up to upgrade the cuDNN FE submodule in TE.

phu0ngng: Made a PR to upgrade cuDNN FE commit in TE submodule https://github.com/NVIDIA/TransformerEngine/pull/3179

phu0ngng:
pipeline 57614140

transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -879,6 +884,10 @@ def fuser_forward(
         fc1_op, activation_op, fc2_op = self.basic_ops
         fc1_ctx, _activation_ctx, fc2_ctx = basic_op_ctxs
 
+        # Caller-provided buffers: FC2 forward output and FC1 backward grad-input.
+        output_buffer = basic_op_kwargs[-1].get(OUTPUT_BUFFER_KEY)
+        fc1_ctx.dgrad_out = basic_op_kwargs[0].get(GRAD_INPUT_BUFFER_KEY)
```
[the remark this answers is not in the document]
YangFei1990: Downstream framework handles this by passing detached tensors, so the original buffer's attribute won't be override. Checking https://github.com/NVIDIA/Megatron-LM/pull/5735.

YangFei1990 (approved):

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -381,6 +420,7 @@ def _forward_grouped_tensor(
                         lambda j=i: weights[j].main_grad for i in range(num_gemms)
                     ]
             ctx.device = device
+            ctx.dgrad_out = dgrad_out
```
timmoon10, line 423: It's not good practice to store tensor in `ctx` and they prefer storing with `ctx.save_for_backward` (see [`save_for_backward` docs](https://docs.pytorch.org/docs/2.13/generated/torch.autograd.function.FunctionCtx.save_for_backward.html)). One additional complication is that we will need to mark these tensors so they don't participate in CPU offloading.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -441,6 +481,8 @@ def forward(
             skip_fp8_weight_update,
             save_original_input,
             debug,
+            out,
+            dgrad_out,
```
timmoon10: These args are tensors, so they shouldn't go in `non_tensor_args`. This is a requirement for future `torch.compile` support, which needs some special handling for tensors.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -719,6 +765,7 @@ def forward(
             ctx.debug = debug
             ctx.save_original_input = save_original_input
             ctx.input_quantizers = input_quantizers
+            ctx.dgrad_out = dgrad_out
```
timmoon10, line 768: `ctx` should not hold tensors directly, but should use `ctx.save_for_backward`.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -18,6 +18,39 @@
 from ..utils import canonicalize_dtype
 
 
+# Keys for passing caller-provided output and grad-input buffers through
+# basic_op_kwargs. See Sequential.forward.
+OUTPUT_BUFFER_KEY = "output"
+GRAD_INPUT_BUFFER_KEY = "grad_input"
```
timmoon10: These are specific to grouped linear and shouldn't be in these general-purpose utilities.

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -1111,6 +1119,8 @@ def fuser_forward_save_ctx(
             ctx.dtype = weight_param.dtype
         ctx.input_requires_grad = requires_grad[0]
         ctx.weight_requires_grad = requires_grad[0] and weight_param.requires_grad
+        # Caller-provided backward grad-input buffer.
+        ctx.dgrad_out = basic_op_kwargs[0].get(GRAD_INPUT_BUFFER_KEY)
```
timmoon10, line 1127: `ctx` should not hold tensors directly, but should use `ctx.save_for_backward`.

transformer_engine/pytorch/ops/fused/grouped_mlp.py
timmoon10, line 899: For safety, it would be nice check for invalid cases (FC1 has `out`, FC2 has `dgrad_out`) and error out.

transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -1434,8 +1443,23 @@ def fuser_forward(
...
 
+            if output_buffer is not None:
+                # Kernel writes directly into the caller buffer, so no copy.
+                output_buffer = validate_or_alloc_output(
+                    output_buffer, fc2_out_shape, dtype, device
+                )
+                fc2_quant_kwargs["d_tensor"] = output_buffer.as_strided(
+                    (in_shape[0], fc2_weight_shape[0], 1),
+                    (fc2_weight_shape[0], 1, in_shape[0] * fc2_weight_shape[0]),
+                )
```
timmoon10: Might as well always allocate. That way we don't need to have a different cuDNN function call depending on whether `out` is provided or not:

```suggestion
            output_buffer = validate_or_alloc_output(
                output_buffer, fc2_out_shape, dtype, device
            )
            fc2_quant_kwargs["d_tensor"] = output_buffer.as_strided(
                (in_shape[0], fc2_weight_shape[0], 1),
                (fc2_weight_shape[0], 1, in_shape[0] * fc2_weight_shape[0]),
            )
```
We'll also need to remove the conditional logic after the GEMM kernel.

transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -2063,8 +2088,21 @@ def fuser_backward(
...
 
+                if grad_input_buffer is not None:
+                    # Kernel writes directly into the caller buffer, so no copy.
+                    grad_input_buffer = validate_or_alloc_output(
+                        grad_input_buffer, in_shape, dtype, device
+                    )
+                    fc1_dgrad_kwargs["d_tensor"] = grad_input_buffer.as_strided(
+                        (out_shape[0], fc1_weight_shape[1], 1),
+                        (fc1_weight_shape[1], 1, out_shape[0] * fc1_weight_shape[1]),
+                    )
```
timmoon10: Better to just always allocate so that we can decouple `dgrad_out` logic from the kernel call.

tests/pytorch/test_grouped_mlp.py
```diff
@@ -1317,6 +1321,98 @@ def _run_case(single_grouped_weight: bool) -> tuple[torch.Tensor, ...]:
             torch.testing.assert_close(fc1_db_false, fc1_db_true, **bias_tols)
             torch.testing.assert_close(fc2_db_false, fc2_db_true, **bias_tols)
 
+    @pytest.mark.skipif(not mxfp8_available, reason=reason_for_no_mxfp8)
+    def test_grouped_mlp_caller_buffers(
```
timmoon10: We should test NVFP4 as well.

transformer_engine/pytorch/ops/sequential.py
```diff
@@ -196,3 +207,55 @@ def forward(
         if extra_outputs:
             return (x,) + tuple(extra_outputs)
         return x
+
+    def _resolve_op_kwargs(
+        self,
+        op_kwargs: Optional[dict[torch.nn.Module | int, dict[str, Any]]],
+    ) -> list[Optional[list[dict[str, Any]]]]:
```
timmoon10, line 214: Walking through the ops and manually reconstructing the op-to-group assignment seems very fragile. `OperationFuser` already keeps track of its `BasicOperation` s, so we can simplify:
```python
    def _resolve_op_kwargs(
        self,
        op_kwargs: Optional[dict[torch.nn.Module | int, dict[str, Any]]],
    ) -> list[Optional[list[dict[str, Any]]]]:
    """Map per-op kwargs onto each module group's basic-op kwargs list."""
        group_kwargs: list[Optional[list[dict[str, Any]]]] = [None] * len(self._module_groups)
        if not op_kwargs:
            return group_kwargs

        # Construct map from op id to kwarg dicts
        resolved: dict[int, dict[str, Any]] = {}
        for key, kwargs in op_kwargs.items():
            module = self[key] if isinstance(key, int) else key
            if not isinstance(module, BasicOperation):
                raise ValueError(f"Attempted to provide provide forward kwargs to {module.__class__.__name__}, but Sequential only allows providing forward kwargs to BasicOperation.")
            resolved[id(module)] = kwargs

        # Construct per-op kwargs for each module group
        for group_idx, module_group in enumerate(self._module_groups):
            if not isinstance(module_group, OperationFuser):
                continue
            group_kwargs[group_idx] = [None] * len(module_group._basic_ops)
            for idx, op in enumerate(module_group._basic_ops):
                if id(op) in resolved:
                    group_kwargs[group_idx][idx] = resolved[id(op)]
                    del resolved[id(op)]

        # Check if any kwarg dicts have not been assigned
        if resolved:
            raise ValueError(f"Attempted to provide forward kwargs to an op that is not in Sequential")

        return group_kwargs
```

transformer_engine/pytorch/module/grouped_linear.py
timmoon10: It seems we run into other problems with `save_for_backward`: https://github.com/NVIDIA/TransformerEngine/pull/3035#discussion_r3480468732

If I understand correctly, `save_for_backward` is intended to avoid leaking memory, while also allowing use-cases like multiple backward passes (not handled properly by TE). For the NCCL EP use-case, these buffers are externally-managed persistent worksapces, so we aren't very worried about the memory leak issue.

transformer_engine/pytorch/ops/fused/grouped_mlp.py
YangFei1990: Added assert for invalid cases

transformer_engine/pytorch/ops/sequential.py
YangFei1990: Done

timmoon10 (approved):
LGTM

PR #3162, opened 2026-07-01
[JAX] Add attention tutorials

KshitijLakhani:
# Description

Add attention tutorial for integrating TE into an existing framework

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

Add a page for non-CP attn tutorial
Add a page for CP attn tutorial

# Checklist:

### DIFF
--- a/docs/examples/jax/attention.rst
+++ b/docs/examples/jax/attention.rst
@@ -3,9 +3,128 @@
 
     See LICENSE for license information.
 
-JAX: Attention with TransformerEngine
-=====================================
+JAX: BSHD Attention with TransformerEngine
+==========================================
 
-**TODO — Coming soon.**
+This document walks through replacing a plain JAX implementation of BSHD
+attention with TransformerEngine's fused ``DotProductAttention``. The example
+uses `grouped-query attention (GQA) <https://arxiv.org/abs/2305.13245>`_ and
+sliding-window attention (SWA).
 
 `← Back to the JAX integration overview <../te_jax_integration.html>`_
+
+1. Baseline: native BSHD GQA + SWA
+----------------------------------
+
+The baseline keeps query, key, and value as separate BSHD tensors. GQA is modeled
+by repeating each KV head across a group of query heads, then applying a causal
+sliding-window mask before softmax.
+
+.. literalinclude:: attention.py
+   :language: python
+   :start-after: # ATTENTION_IMPORTS_START
+   :end-before: # ATTENTION_IMPORTS_END
+
+.. literalinclude:: attention.py
+   :language: python
+   :start-after: # ATTENTION_INPUTS_START
+   :end-before: # ATTENTION_INPUTS_END
+
+.. literalinclude:: attention.py
+   :language: python
+   :start-after: # ATTENTION_BASELINE_MODEL_START
+   :end-before: # ATTENTION_BASELINE_MODEL_END
+
+
+2. Transformer Engine ``DotProductAttention``
+----------------------------------------------
+
+The Transformer Engine version keeps the same separate BSHD inputs. The important arguments are
+``num_gqa_groups`` for GQA, ``attn_mask_type="causal"`` for autoregressive
+attention, and ``window_size`` for SWA.
+
+.. literalinclude:: attention.py
+   :language: python
+   :start-after: # ATTENTION_TE_MODEL_START
+   :end-before: # ATTENTION_TE_MODEL_END
+
+
+3. Single-GPU performance
+-------------------------
+
+``speedometer`` runs a JIT-compiled forward+backward loop with warmup for both
+implementations.
+
+.. literalinclude:: attention.py
+   :language: python
+   :start-after: # ATTENTION_SINGLE_GPU_BENCH_START
+   :end-before: # ATTENTION_SINGLE_GPU_BENCH_END
+
+.. raw:: html
+
+   <div style="background: #f5f5f5; border-left: 3px solid #9ca3af; padding: 4px 12px; font-size: 12px; color: #6b7280; margin-top: -16px;">
+      Output:
+   </div>
+
+.. container:: program-output
+
+   .. literalinclude:: attention.out
+      :language: text
+      :start-after: # SINGLE_GPU_OUTPUT_START
+      :end-before: # SINGLE_GPU_OUTPUT_END
+
+On a single GB200, this run is roughly **52x faster** for the fwd+bwd of this
+BSHD GQA + SWA example. This compares TE ``DotProductAttention`` against the
+native JAX baseline above, which materializes attention scores with XLA ops; it
+is not a comparison against ``jax.nn.dot_product_attention(...,
+implementation="cudnn")``.
+
+
+4. MLA-style head dimensions
+----------------------------
+
+In TE/JAX, the simple MLA-style attention case is represented by separate Q, K,
+and V tensors where Q/K and V use different per-head dimensions. Keep
+``qkv_layout="bshd_bshd_bshd"`` so TE can see the Q/K head dimension and the V
+head dimension separately.
+
+.. literalinclude:: attention.py
+   :language: python
+   :start-after: # ATTENTION_MLA_START
+   :end-before: # ATTENTION_MLA_END
+
+.. raw:: html
+
+   <div style="background: #f5f5f5; border-left: 3px solid #9ca3af; padding: 4px 12px; font-size: 12px; color: #6b7280; margin-top: -16px;">
+      Output:
+   </div>
+
+.. container:: program-output
+
+   .. literalinclude:: attention.out
+      :language: text
+      :start-after: # MLA_OUTPUT_START
+      :end-before: # MLA_OUTPUT_END
+
+
+Other attention knobs
+---------------------
+
+The examples above intentionally stay focused. Other ``DotProductAttention``
+features are enabled through the same module arguments:
+
+* Dropout: set ``attention_dropout > 0``, call with ``deterministic=False``, and
+  pass a Flax ``dropout`` RNG to ``apply``.
+* Bias: pass ``bias`` and set ``attn_bias_type`` when the selected fused kernel
+  supports that bias mode.
+* Sink attention: use ``softmax_type="off_by_one"`` or ``"learnable"``.
+* Determinism: set ``NVTE_ALLOW_NONDETERMINISTIC_ALGO=0`` before launching the
+  process if deterministic fused kernels are required.
+
+
+Next steps
+----------
+
+* `Context-parallel attention <attention_context_parallel.html>`_: packed THD
+  attention over a context-parallel mesh.
+* `← Hub <../te_jax_integration.html>`_
--- a/docs/examples/jax/attention.rst
+++ b/docs/examples/jax/attention.rst
@@ -110,8 +110,9 @@ head dimension separately.
 Other attention knobs
 ---------------------
 
-The examples above intentionally stay focused. Other ``DotProductAttention``
-features are enabled through the same module arguments:
+The examples above represent a subset of attention features. Other
+``DotProductAttention`` features can be enabled through the same module
+arguments as below:
 
 * Dropout: set ``attention_dropout > 0``, call with ``deterministic=False``, and
   pass a Flax ``dropout`` RNG to ``apply``.
--- a/docs/examples/jax/attention.rst
+++ b/docs/examples/jax/attention.rst
@@ -3,129 +3,47 @@
 
     See LICENSE for license information.
 
-JAX: BSHD Attention with TransformerEngine
-==========================================
+JAX: Attention with TransformerEngine
+=====================================
 
-This document walks through replacing a plain JAX implementation of BSHD
-attention with TransformerEngine's fused ``DotProductAttention``. The example
-uses `grouped-query attention (GQA) <https://arxiv.org/abs/2305.13245>`_ and
-sliding-window attention (SWA).
+Transformer Engine's JAX attention APIs support self-attention and
+cross-attention with MHA, GQA, and MQA. Inputs can use standard BSHD or packed
+THD layouts, with Q/K/V supplied separately or in packed forms.
+
+Common options include causal and padding masks, bias, dropout, sliding-window
+attention (SWA), attention sinks, and experimental ``score_mod`` callbacks for
+FlexAttention-style customization. The API also supports different Q/K and V
+head dimensions, as used by the attention operation after the projection
+stages in DeepSeek-style MLA.
+
+For long contexts, selected BSHD and THD configurations support context
+parallelism with Ring or AllGather collectives. Exact fused-kernel availability
+depends on the input shape, dtype, GPU, and feature combination; see the
+`JAX DotProductAttention API reference
+<../../api/jax.html#transformer_engine.jax.flax.DotProductAttention>`_ for the
+full interface. Choose the tutorial that matches how the sequence dimension is
+distributed in your model.
 
 `← Back to the JAX integration overview <../te_jax_integration.html>`_
 
-1. Baseline: native BSHD GQA + SWA
-----------------------------------
+Pick a tutorial
+---------------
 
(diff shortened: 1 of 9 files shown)


docs/examples/jax/attention.rst
```diff
@@ -3,9 +3,128 @@
...
     See LICENSE for license information.
 
-JAX: Attention with TransformerEngine
-=====================================
+JAX: BSHD Attention with TransformerEngine
+==========================================
 
-**TODO — Coming soon.**
+This document walks through replacing a plain JAX implementation of BSHD
+attention with TransformerEngine's fused ``DotProductAttention``. The example
```
jberchtold-nvidia: Thoughts on another sentence in here to mention TE/JAX attn has more features than shown in this tutorial?

Something like
> TransformerEngine's DotProductAttention supports a variety of features. The full documentation can be found here "link". The tutorial example uses `grouped-query attention (GQA) <https://arxiv.org/abs/2305.13245>`_ and
sliding-window attention (SWA).

docs/examples/jax/attention.rst
```diff
@@ -3,9 +3,128 @@
...
+
+   .. literalinclude:: attention.out
+      :language: text
+      :start-after: # SINGLE_GPU_OUTPUT_START
+      :end-before: # SINGLE_GPU_OUTPUT_END
+
+On a single GB200, this run is roughly **52x faster** for the fwd+bwd of this
+BSHD GQA + SWA example. This compares TE ``DotProductAttention`` against the
+native JAX baseline above, which materializes attention scores with XLA ops; it
+is not a comparison against ``jax.nn.dot_product_attention(...,
```
jberchtold-nvidia: > it is not a comparison against jax.nn.dot_product_attention(..., implementation="cudnn")

Does native JAX jax.nn.dot_product_attention with backend cudnn support SWA? My understanding is native JAX did not support SWA via cuDNN. If so, can you mention that as the reasoning for why we aren't comparing against jax.nn.dot_production_attention with cuDNN backend

docs/examples/jax/attention.rst
```diff
@@ -3,9 +3,128 @@
...
+head dimension separately.
+
+.. literalinclude:: attention.py
+   :language: python
+   :start-after: # ATTENTION_MLA_START
+   :end-before: # ATTENTION_MLA_END
+
+.. raw:: html
+
+   <div style="background: #f5f5f5; border-left: 3px solid #9ca3af; padding: 4px 12px; font-size: 12px; color: #6b7280; margin-top: -16px;">
```
jberchtold-nvidia: Do we need this raw HTML "output:" block? If the block is immediately below an existing code block I think it being the output is implied. If this is copied from our other TE/JAX docs, lmk and we can remove it from all the docs

docs/examples/jax/attention.rst
```diff
@@ -3,9 +3,128 @@
...
+      :end-before: # MLA_OUTPUT_END
+
+
+Other attention knobs
+---------------------
+
+The examples above intentionally stay focused. Other ``DotProductAttention``
+features are enabled through the same module arguments:
+
+* Dropout: set ``attention_dropout > 0``, call with ``deterministic=False``, and
```
jberchtold-nvidia: I like this section! Good overview of other attn features without being as overwhelming with options as our main docs are. Given attention inherently has so many config options our main docstrings are gonna be verbose, so this preview of features in the tutorial is nice!

docs/examples/jax/attention_context_parallel.rst
```diff
@@ -0,0 +1,145 @@
...
+the attention work.
+
+**Prerequisite:** this example requires four GPUs.
+
+`← Back to the JAX integration overview <../te_jax_integration.html>`_
+
+1. Packed THD inputs
+--------------------
+
+In the separate-QKV THD layout used here, Q/K/V are shaped
```
jberchtold-nvidia, line 42: Is there a performance difference between together QKV and QKV as separate tensors? If so, may be worth mentioning or at least mention we have support for all contiguous in one tensor too in case that is what a user has

docs/examples/jax/attention_context_parallel.py
```diff
@@ -0,0 +1,378 @@
...
+# ATTENTION_CP_INPUTS_END
+
+
+# ATTENTION_CP_MESH_START
+def build_cp_mesh():
+    """Use one JAX mesh axis for context parallelism over sequence."""
+
+    devices = np.asarray(jax.devices()[:cp_size])
+    mesh = Mesh(devices, axis_names=("cp",))
+    mesh_resource = MeshResource(cp_resource="cp")
```
jberchtold-nvidia, line 116: nit SR: can you make a code comment here saying "If you are using other parallelisms in your project, please fill out the other fields of MeshResource to inform TransformerEngine about which axes names you are using."

docs/examples/jax/attention_context_parallel.py
```diff
@@ -0,0 +1,378 @@
...
+    seq_desc,
+    *,
+    context_parallel_axis: str = "",
+    context_parallel_strategy: CPStrategy = CPStrategy.DEFAULT,
+    context_parallel_causal_load_balanced: bool = False,
+    stripe_size: int | None = None,
+):
+    """Call TE fused attention on separate THD Q, K, V tensors."""
+
+    return fused_attn(
```
jberchtold-nvidia, line 135: What is the use case for using the fused_attn function API vs the DotProductAttention layer? Afaiu the DotProductAttention layer did not "own" any weights, so I assumed `fused_attn` and the `DPA` layer were equivalent

docs/examples/jax/attention_context_parallel.rst
```diff
@@ -0,0 +1,145 @@
...
+
+Both examples use packed THD, causal masking, SWA, and dropout-free fused
+attention. The only strategy-specific difference is the CP strategy and stripe
+size. CP collectives depend on the compiler seeing the intended sharding, so the
+forward and forward+backward functions are compiled with explicit
+``in_shardings``; the forward+backward path also pins the gradient sharding. The
+timing loop follows the same forward+backward pattern as ``speedometer`` while
+keeping those sharding controls visible.
+
+.. literalinclude:: attention_context_parallel.py
```
jberchtold-nvidia, line 121: Should we add in the following entries to show where CP excels?
* JAX attention jax.nn.dpa: OOM
* TE DPA without CP: OOM

If you picked small/medium shapes for the example and they don't actually OOM without CP in these isolated examples, I think that's fine to keep as is without this suggestion

docs/examples/jax/attention.rst
```diff
@@ -3,9 +3,128 @@
 
     See LICENSE for license information.
 
-JAX: Attention with TransformerEngine
-=====================================
+JAX: BSHD Attention with TransformerEngine
+==========================================
```
cyanguwa: How about keeping "JAX: Attention with TransformerEngine" and then use "JAX: Context-Parallel Attention with TransformerEngine" for the CP file? We can walk through some of the features non-CP/CP supports before getting into specific examples. I think jberchtold suggested this, but also, the last section of this file seems to cover "other features" in attention as well.

docs/examples/jax/attention.rst
cyanguwa: We can use the table we had in recent presentations (or just turn that into words to use here).

docs/examples/jax/attention.rst
```diff
@@ -3,9 +3,128 @@
...
 
-**TODO — Coming soon.**
+This document walks through replacing a plain JAX implementation of BSHD
+attention with TransformerEngine's fused ``DotProductAttention``. The example
+uses `grouped-query attention (GQA) <https://arxiv.org/abs/2305.13245>`_ and
+sliding-window attention (SWA).
```
cyanguwa: Maybe add a link to SWA as well, or remove the one for GQA for consistency?

docs/examples/jax/attention.rst
```diff
@@ -3,9 +3,128 @@
...
 
 `← Back to the JAX integration overview <../te_jax_integration.html>`_
+
+1. Baseline: native BSHD GQA + SWA
```
cyanguwa: How about "1. Baseline: native-Jax BSHD GQA + SWA"? Also, how is the "unfused attn" path compare to this baseline? Should we use that path instead of reimplementing FlaxNativeGQAAttention here?

docs/examples/jax/attention.rst
```diff
@@ -3,9 +3,128 @@
...
+
+The baseline keeps query, key, and value as separate BSHD tensors. GQA is modeled
+by repeating each KV head across a group of query heads, then applying a causal
+sliding-window mask before softmax.
+
+.. literalinclude:: attention.py
+   :language: python
+   :start-after: # ATTENTION_IMPORTS_START
+   :end-before: # ATTENTION_IMPORTS_END
+
```
cyanguwa: Maybe add a single line before the next code block starts? Same for the next block.

docs/examples/jax/attention.rst
```diff
@@ -3,9 +3,128 @@
...
+      :end-before: # SINGLE_GPU_OUTPUT_END
+
+On a single GB200, this run is roughly **52x faster** for the fwd+bwd of this
+BSHD GQA + SWA example. This compares TE ``DotProductAttention`` against the
+native JAX baseline above, which materializes attention scores with XLA ops; it
+is not a comparison against ``jax.nn.dot_product_attention(...,
+implementation="cudnn")``.
+
+
+4. MLA-style head dimensions
```
cyanguwa: Maybe DeepSeek-style attention (MLA)?

docs/examples/jax/attention_context_parallel.rst
```diff
@@ -0,0 +1,145 @@
+..
+    Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+
+    See LICENSE for license information.
+
+JAX: Context-Parallel THD Attention with Transformer Engine
```
cyanguwa: How about "JAX: Context-Parallel Attention with Transformer Engine" and

"CP supports BSHD, THD with blah (link to our docs), and we are going to demonstrate a few popular user configs..."?

docs/examples/jax/attention_context_parallel.rst
```diff
@@ -0,0 +1,145 @@
...
+`← Back to the JAX integration overview <../te_jax_integration.html>`_
+
+1. Packed THD inputs
+--------------------
+
+In the separate-QKV THD layout used here, Q/K/V are shaped
+``[batch, seq, heads, dim]``, and the sequence dimension can pack several
+shorter segments. The ``SequenceDescriptor`` tells TE which tokens belong to
+which packed segment and which token slots are padding. This tutorial uses
+sixteen padded segments per sequence and a 64k sequence length.
```
cyanguwa: "sixteen padded segments per sequence, a 64k sequence length, and 2 sequences per batch."?

docs/examples/jax/attention_context_parallel.rst
```diff
@@ -0,0 +1,145 @@
...
+
+.. literalinclude:: attention_context_parallel.py
+   :language: python
+   :start-after: # ATTENTION_CP_IMPORTS_START
+   :end-before: # ATTENTION_CP_IMPORTS_END
+
```
cyanguwa, line 56: "The tensor inputs are created as follows."?

docs/examples/jax/attention_context_parallel.rst
cyanguwa: Can we add some speedup numbers as well, CP vs non-CP?

docs/examples/jax/attention_context_parallel.rst
```diff
@@ -0,0 +1,145 @@
...
+The JAX ``Mesh`` describes the physical devices. ``MeshResource`` tells TE which
+mesh axis is used for context parallelism.
+
+.. literalinclude:: attention_context_parallel.py
+   :language: python
+   :start-after: # ATTENTION_CP_MESH_START
+   :end-before: # ATTENTION_CP_MESH_END
+
+
+3. Fused THD attention call
```
cyanguwa: "Fused attention call"?

docs/examples/jax/attention.rst
KshitijLakhani: I am thinking of adding a sentence and API reference as @jberchtold-nvidia suggested
The table might change the scope of this documentation from samples to support matrix i feel. So that could be a separate effort that we can then link here in the samples once it is ready. Is that okay @cyanguwa ?

docs/examples/jax/attention.rst
KshitijLakhani: >  If this is copied from our other TE/JAX docs, lmk and we can remove it from all the docs

It is. This follows the existing output-block style used by the Dense, Getting Started documentation. I would prefer to keep it here for consistency; removing it would be better handled as a separate documentation-wide cleanup. Does that work @jberchtold-nvidia ?

docs/examples/jax/attention_context_parallel.rst
KshitijLakhani: So THD_THD_THD and THD_T2HD are supported whereas T3HD are not supported for CP. My understanding is that this is because typically we communicate the KV for CP while keeping the Q local, so unpacking the interleaving for T3HD makes it tricky.

Having said that, packed KV might be very slightly faster (I ran a couple experiments and it was ~1-2 %) so users should generally retain whichever supported layout their QKV projection naturally produces

A note might not be necessary for the same

docs/examples/jax/attention.rst
KshitijLakhani: The API does support SWA: https://docs.jax.dev/en/latest/_autosummary/jax.nn.dot_product_attention.html
I tried examples with the cuDNN backend separately to confirm that SWA is indeed supported

The motivation I had in mind for this sample/tutorial was to compare against native XLA backend and not the cuDNN backend (the comment here just helps clarify the same)

docs/examples/jax/attention.rst
KshitijLakhani: Sure I can do Baseline: native-xla BSHD GQA + SWA

I did think about this but the general flow of other sections seems to be: start with a native jax (non TE API) implementation and then show the user how to "port" it to TE so keeping that spirit in mind, I stayed away from using the TE unfused API

docs/examples/jax/attention_context_parallel.rst
KshitijLakhani: Okay let me add speed up numbers and tweak the config numbers to pick an appropriate config for the same

docs/examples/jax/attention.rst
KshitijLakhani: > How about keeping "JAX: Attention with TransformerEngine" and then use "JAX: Context-Parallel Attention with TransformerEngine" for the CP file? 

Sounds good

docs/examples/jax/attention_context_parallel.py
KshitijLakhani: The `DotProductAttention` layer currently does not expose `stripe_size`. Since the value must match the external THD load-balancing reorder, this tutorial uses the lower-level fused_attn API.
In a way it's good as we do have customers using both levels of the API anyways, so the diversification of the fused attn api usage in the tutorials helps

To address the root cause, I may have missed this plumbing when i added support for non default `stripe_size>1`. Earlier it would always default to 1 and we'd need no other value but when I added support for CP AG THD I introduced a configurable `stripe_size`. Ideally we should expose this to DPA API nontheless.

We can expose and validate `stripe_size` in `DotProductAttention` in a separate PR, then update this tutorial to use the higher-level layer if needed, but would be great for that to not be a blocker for the current PR's merge.

cyanguwa (dismissed):

cyanguwa (approved):

docs/examples/jax/attention.rst
jberchtold-nvidia: Follow-up sounds good

jberchtold-nvidia (approved):

PR #3165, opened 2026-07-02
Improve readability of dgrad overlap variable

### LINKED ISSUE #3164: Improve readability of dgrad_reduce_scatter_overlap variable name
The variable `dgrad_reduce_scatter_overlap = ["qkv_dgrad", "fc1_dgrad"]` is misleadingly named. By default both layers are placed in `layers_all_gather_overlap `with method="bulk". They are only moved to `layers_reduce_scatter_overlap` under two specific conditions:

The user explicitly provides a non-bulk method for these layers via ub_cfgs
with_cublasmp=True is set, since cuBLASMp doesn't support bulk overlap

Why the name is misleading:-
The name implies these dgrad layers always use reduce-scatter overlap, but the default is the opposite -- they default to all-gather overlap. What actually makes them special is that they are the only two GEMMs in the transformer block that support communication on both sides (all-gather on input, reduce-scatter on output), making their overlap strategy switchable -- unlike every other GEMM which has only one comm direction.

Prachi-kushwaha:
# Description
Rename `dgrad_reduce_scatter_overlap` to better reflect its purpose.

Fixes #3164

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

# Checklist:


denera (changes requested):
`layers_all_gather_overlap`, `layers_reduce_scatter_overlap` and `dgrad_reduce_scatter_overlap` are all internal variables that are never exposed to the user. The first two `layers_*_overlap` lists encode the default comm overlap behavior while `dgrad_reduce_scatter_overlap` keeps track of which DGRAD GEMMs can optionally be overlapped with reduce-scatter instead of bulk all-gather.

I don't think the new `layers_dgrad_ag_and_rs_overlap` name in this PR is any more descriptive of how the variable is actually used in our internal workflow.

More importantly, focusing on the variable name is glossing over the actual deficiency — documentation.

The only reason why anyone goes digging in this source code and looks at these internal variables is because we are not telling the user what the default configuration is for each layer. The real solution to this problem is to update the `initialize_ub()` docstring with that missing information. That would enable users to supply partial config dictionaries that only modify what they want done differently than the default behavior.

So I would recommend that you either change this PR or close and file a new one to supply that docstring update. That would be a very valuable change to review and merge for us.

denera:
Closing this PR because of no action/updates after review. Please re-open if you'd like to revisit it and address latest review comments. Thanks!

PR #3171, opened 2026-07-02
[PyTorch] Enable fused FP8 block-scaling path in GroupedLinear module

### LINKED ISSUE #2525: Blockwise (1x128 and 128x128) FP8 grouped quantization

denera:
# Description

This PR integrates the new grouped FP8 block-scaling quantization kernels into the fused GroupedTensor path in GroupedLinear.

Integration into fusible ops is deferred to a follow-up PR as it requires additional effort to **also** enable non-grouped FP8 block-scaling at the same time, considered out-of-scope for this PR.

## Notes
- DGRAD and WGRAD grouped GEMMs executing serially on the same stream appear to violate the workspace serialization under some specific graph replay conditions, leading to a deadlock in WGRAD. PR resolves the deadlock by allocating and caching TWO 32MiB cuBLAS workspaces in order to isolate DGRAD and WGRAD workspaces.
- PR enables pow_2 constrained scales for grouped FP8 block-scaling at the kernel level for numerical parity with the split-quantize path on SM90, but the FP8BS->MXFP8 scale-broadcast mechanism remains **disabled** on SM100+.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -36,6 +36,7 @@ from transformer_engine.pytorch.ops.fused import (
 )
 from transformer_engine.pytorch import (
     QuantizedTensor,
+    Float8BlockQuantizer,
     Float8CurrentScalingQuantizer,
     Float8Quantizer,
     MXFP8Quantizer,
@@ -58,6 +59,9 @@ from utils import (
 fp8_available, reason_for_no_fp8 = te.is_fp8_available(return_reason=True)
 mxfp8_available, reason_for_no_mxfp8 = te.is_mxfp8_available(return_reason=True)
 nvfp4_available, reason_for_no_nvfp4 = te.is_nvfp4_available(return_reason=True)
+fp8_block_scaling_available, reason_for_no_fp8_block_scaling = te.is_fp8_block_scaling_available(
+    return_reason=True
+)
 
 # Supported data types
 _dtypes: list[torch.dtype] = [torch.float32, torch.float16]
@@ -76,6 +80,11 @@ if mxfp8_available:
 if nvfp4_available:
     _quantization_list.append("nvfp4")
     _quantization_list.append("nvfp4_4over6")
+
+# GroupedLinear is currently the only basic op that supports FP8 block scaling.
+_grouped_linear_quantization_list: list[Optional[str]] = list(_quantization_list)
+if fp8_block_scaling_available:
+    _grouped_linear_quantization_list.append("fp8_block_scaling")
 
 
 @pytest.fixture(autouse=True, scope="function")
@@ -186,6 +195,18 @@ def make_reference_and_test_tensors(
         test = quantizer(test)
     elif quantization == "mxfp8":
         test = MXFP8Quantizer(fp8_dtype=te.DType.kFloat8E4M3)(test)
+    elif quantization == "fp8_block_scaling":
+        tensor_type = "input"
+        if quantizer_role is not None:
+            tensor_type = quantizer_role.tensor_type
+        # Match the Float8BlockScaling recipe defaults: 2D (128x128) blocks for
+        # weights, 1D (1x128) otherwise, with power-of-2 scales.
+        test = Float8BlockQuantizer(
+            fp8_dtype=te.DType.kFloat8E4M3,
+            rowwise=True,
+            columnwise=True,
+            block_scaling_dim=2 if tensor_type == "weight" else 1,
+        )(test)
     elif quantization in ("nvfp4", "nvfp4_row_scaled", "nvfp4_rht"):
         tensor_type = "input"
         if quantizer_role is not None:
@@ -2067,7 +2088,7 @@ class TestBasicOps:
 
     @pytest.mark.parametrize("bias", (False, True))
     @pytest.mark.parametrize("dtype", _dtypes)
-    @pytest.mark.parametrize("quantization", _quantization_list)
+    @pytest.mark.parametrize("quantization", _grouped_linear_quantization_list)
     @pytest.mark.parametrize("quantized_compute", (False, True))
     @pytest.mark.parametrize("quantized_weight", (False, True))
     @pytest.mark.parametrize("input_requires_grad", (False, True))
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -36,7 +36,6 @@ from transformer_engine.pytorch.ops.fused import (
 )
 from transformer_engine.pytorch import (
     QuantizedTensor,
-    Float8BlockQuantizer,
     Float8CurrentScalingQuantizer,
     Float8Quantizer,
     MXFP8Quantizer,
@@ -59,9 +58,6 @@ from utils import (
 fp8_available, reason_for_no_fp8 = te.is_fp8_available(return_reason=True)
 mxfp8_available, reason_for_no_mxfp8 = te.is_mxfp8_available(return_reason=True)
 nvfp4_available, reason_for_no_nvfp4 = te.is_nvfp4_available(return_reason=True)
-fp8_block_scaling_available, reason_for_no_fp8_block_scaling = te.is_fp8_block_scaling_available(
-    return_reason=True
-)
 
 # Supported data types
 _dtypes: list[torch.dtype] = [torch.float32, torch.float16]
@@ -80,11 +76,6 @@ if mxfp8_available:
 if nvfp4_available:
     _quantization_list.append("nvfp4")
     _quantization_list.append("nvfp4_4over6")
-
-# GroupedLinear is currently the only basic op that supports FP8 block scaling.
-_grouped_linear_quantization_list: list[Optional[str]] = list(_quantization_list)
-if fp8_block_scaling_available:
-    _grouped_linear_quantization_list.append("fp8_block_scaling")
 
 
 @pytest.fixture(autouse=True, scope="function")
@@ -195,18 +186,6 @@ def make_reference_and_test_tensors(
         test = quantizer(test)
     elif quantization == "mxfp8":
         test = MXFP8Quantizer(fp8_dtype=te.DType.kFloat8E4M3)(test)
-    elif quantization == "fp8_block_scaling":
-        tensor_type = "input"
-        if quantizer_role is not None:
-            tensor_type = quantizer_role.tensor_type
-        # Match the Float8BlockScaling recipe defaults: 2D (128x128) blocks for
-        # weights, 1D (1x128) otherwise, with power-of-2 scales.
-        test = Float8BlockQuantizer(
-            fp8_dtype=te.DType.kFloat8E4M3,
-            rowwise=True,
-            columnwise=True,
-            block_scaling_dim=2 if tensor_type == "weight" else 1,
-        )(test)
     elif quantization in ("nvfp4", "nvfp4_row_scaled", "nvfp4_rht"):
         tensor_type = "input"
         if quantizer_role is not None:
@@ -2088,7 +2067,7 @@ class TestBasicOps:
 
     @pytest.mark.parametrize("bias", (False, True))
     @pytest.mark.parametrize("dtype", _dtypes)
-    @pytest.mark.parametrize("quantization", _grouped_linear_quantization_list)
+    @pytest.mark.parametrize("quantization", _quantization_list)
     @pytest.mark.parametrize("quantized_compute", (False, True))
     @pytest.mark.parametrize("quantized_weight", (False, True))
     @pytest.mark.parametrize("input_requires_grad", (False, True))
--- a/tests/pytorch/test_grouped_linear.py
+++ b/tests/pytorch/test_grouped_linear.py
@@ -1539,20 +1539,6 @@ def _run_grouped_linear_path(
     out_features = weights[0].size(0)
     use_fp8 = fp8_recipe is not None
 
-    # Spy on the grouped quantize entry point so a predicate that silently
-    # declines the fused path fails the test instead of vacuously matching the
-    # legacy reference.
-    group_quantize_calls = 0
-    if enable_grouped_tensor_path and use_fp8:
-        real_group_quantize = tex.group_quantize
-
-        def _counting_group_quantize(*args, **kwargs):
-            nonlocal group_quantize_calls
-            group_quantize_calls += 1
-            return real_group_quantize(*args, **kwargs)
-
-        monkeypatch.setattr(tex, "group_quantize", _counting_group_quantize)
-
     x = x_base.detach().clone().requires_grad_(True)
     with quantized_model_init(enabled=fp8_model_params, recipe=fp8_recipe):
         grouped_linear = GroupedLinear(
@@ -1582,10 +1568,6 @@ def _run_grouped_linear_path(
     y.backward(dy)
     if delay_wgrad_compute:
         grouped_linear.backward_dw()
-
-    if enable_grouped_tensor_path and use_fp8:
-        monkeypatch.setattr(tex, "group_quantize", real_group_quantize)
-        assert group_quantize_calls > 0, "fused GroupedTensor path was not taken"
 
     outputs = [y, x.grad]
     for i in range(num_gemms):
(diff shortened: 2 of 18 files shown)


transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -474,6 +514,17 @@ def general_grouped_gemm_for_grouped_tensor(
...
+    def _is_fp8_blockwise(operand) -> bool:
+        if isinstance(operand, (list, tuple)):
+            return any(isinstance(t, Float8BlockwiseQTensorStorage) for t in operand)
+        if isinstance(operand, GroupedTensorStorage):
+            return isinstance(operand.quantizer, Float8BlockQuantizer)
+        return False
+
+    if _is_fp8_blockwise(A) or _is_fp8_blockwise(B):
+        # FP8 block-scaling requires split accumulator
+        use_split_accumulator = True
```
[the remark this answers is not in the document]
denera: `Float8BlockScaling` recipe itself already documents and enforces `use_split_accumulator=True` regardless of user selection because cuBLAS GEMM does not support the alternative for FP8 block-scaling. The unconditional silent override in `gemm.py` is defensive and aligned with the already documented intent of the recipe. Latest commit adds extra documentation to `gemm.py`.

transformer_engine/pytorch/ops/op.py
```diff
@@ -183,6 +183,9 @@ class BasicOperation(FusibleOperation, metaclass=abc.ABCMeta):
     num_extra_inputs: int = 0
     # Number of extra tensor outputs
     num_extra_outputs: int = 0
+    # Whether this operation supports the FP8 block-scaling recipe. Most basic
+    # ops have not been audited for it, so they opt in explicitly.
+    supports_float8_block_scaling: bool = False
```
timmoon10: Putting something so specific in the base class is quite hacky. As far as I can tell, this is only used to get around a recipe check in `reset_recipe_state`, when it would be better to just remove that check. 
```suggestion
```

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -778,9 +782,11 @@ def _is_graph_safe_path_supported(
         * FP8 per-tensor current scaling is backed by grouped current-scaling quantization
           (``tex.group_quantize``) and cuBLASLt grouped GEMM with per-batch scalar FP8 scaling,
           which are supported on Hopper (CC 9.0) and Blackwell (CC 10.x and 11.0).
-          Every other quantization recipe (fp8 delayed scaling, fp8 block scaling, ...)
-          falls back to the legacy flow because the corresponding grouped quantization kernels are
-          missing.
+        * FP8 block scaling runs the graph-safe flow on Hopper (CC 9.0) only, including
+          power-of-2 scales. On other architectures it falls back to the split-quantize
```
timmoon10: Nit: If power-of-2 doesn't matter, why bring it up? It seems like Claude struggled with this, and now it brings it up everywhere even when it won't be helpful to downstream users.

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -435,6 +436,45 @@ def _get_fp32_zeros_tensor(num_tensors: int, device: torch.device) -> torch.Tens
...
+    bytes of the workspace (cuBLAS emits a memset that zeros it before each
+    matmul). When two grouped matmuls in a replayed CUDA graph share one
+    workspace, that flag is aliased, and the second matmul's cooperative kernel
+    deadlocks on its second replay with cuBLAS 13.6 (and corrupts the last
+    expert's wgrad on cuBLAS < 13.6) -- the two matmuls are strictly stream-
+    ordered, so this is shared-workspace reuse, not concurrent co-scheduling.
+    Giving each its own persistent buffer removes the hazard; each slot is a
+    single persistent allocation, so CUDA-graph capture safety is preserved.
+    These are dedicated to the grouped path and are intentionally not shared with
+    the non-grouped GEMM workspace.
```
timmoon10: Wait, is this true? This is saying that the kernel will fail if the workspace has certain values. This means that it's not safe to allocate with `torch.emtpy` and we need to allocate with `torch.zero`.

tests/pytorch/test_grouped_mlp.py
```diff
@@ -61,6 +65,11 @@
     _quantization_list.append("nvfp4")
     _quantization_list.append("nvfp4_4over6")
 
+# GroupedLinear is currently the only basic op that supports FP8 block scaling.
+_grouped_linear_quantization_list: list[Optional[str]] = list(_quantization_list)
```
timmoon10: Why not just add FP8 block scaling to `_quantization_list`?

tests/pytorch/test_fusible_ops.py
```diff
@@ -73,6 +77,11 @@
     _quantization_list.append("nvfp4")
     _quantization_list.append("nvfp4_4over6")
 
+# GroupedLinear is currently the only basic op that supports FP8 block scaling.
+_grouped_linear_quantization_list: list[Optional[str]] = list(_quantization_list)
```
timmoon10: Have we tried just adding FP8 block scaling to `_quantization_list`? The quantization infrastructure should be general enough for it to work out of the box. If it works automatically without non-trivial changes, then I don't see a reason not to include it in the tests.

timmoon10 (changes requested):

transformer_engine/pytorch/cpp_extensions/gemm.py
denera: This is true and I have a deterministic reproducer for it, but it's not really about the `torch.empty` vs `torch.zero` allocation issue.

It is very specifically affecting only the 1Dx1D grouped FP8BS GEMM kernel (only used in WGRAD, while FWD and DGRAD use 1Dx2D and 2Dx1D), and only under graph **replay**. DGRAD GEMM is leaving behind stuff in the cuBLAS workspace that's corrupting/deadlocking the WGRAD GEMM, but it manifests only in the 2nd graph replay (first execution after capture is fine).

Zeroing out the workspace in between DGRAD and WGRAD GEMMs also resolves the issue, but I opted to utilize workspace isolation instead to avoid introducing extra work of zeroing out a shared workspace into the backward pass.

The catch is that the deterministic reproducer for this is still PyTorch-dependent. I cannot reproduce this in pure C++ using just the TE/common API. So at least by process of elimination, there seems to be an interaction I don't quite understand between graph capture/replay and PyTorch's allocation pool.

transformer_engine/pytorch/ops/basic/grouped_linear.py
denera: Split-quantize path supports power-of-2 constrained scales and disallowing that in the fused-quantize path means we either silently ignore the config flag for this in the quantizer or we introduce a bunch of ugly code branching. I wasn't a fan of either of those options, so I enabled support power-of-2 scales in the kernel because: 1) it's trivial, 2) makes the split- and fused-quantize paths numerically equivalent for every config, and 3) lays down the foundation for enabling grouped FP8BS on Blackwell+ using the same MXFP8 scale-broadcast trick we use for non-grouped FP8BS in a separate/future PR.

tests/pytorch/test_fusible_ops.py
denera: No, I haven't attempted this but you're right, if it's that simple, there's no reason to not try and include it if it works. I'll give it a try and see what happens.

transformer_engine/pytorch/ops/op.py
denera: Right, this becomes unnecessary if simply adding FP8BS into the `_quantization_list` works out of the box.

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -513,16 +567,10 @@ def general_grouped_gemm_for_grouped_tensor(
...
-    )
-    workspace_cublas = torch.empty(
-        get_cublas_workspace_size_bytes(),
-        dtype=torch.uint8,
-        device=device,
-    )
+    workspace_setup = _get_grouped_gemm_setup_workspace(device.index, num_tensors)
+    # dgrad and wgrad must use distinct persistent cuBLAS workspaces: sharing one
+    # deadlocks under CUDA-graph replay on cuBLAS 13.6 (see _get_grouped_cublas_workspace).
+    workspace_cublas = _get_grouped_cublas_workspace(device.index, 1 if is_discrete_out else 0)
```
vthumbe1503: For single_grouped_weight case, we would need single grouped tensor output instead discrete output even for wgrad.

So this logic wont give different workspaces for wgrad and dgrad in that case

tests/pytorch/test_fusible_ops.py
denera: `Float8BlockwiseQTensor` isn't propagating dtype recasts (e.g.`op.half()`) and does not support `.view()` reshape, so it looks like there's meaningful work that needs to be done before we can broadly enable FP8BS in fusible ops.

I'd suggest maybe scaling this PR back to strip out fusible ops support entirely, implement integration strictly in the GroupedLinear module, and shift the fusible ops FP8BS (both non-grouped and grouped) support to a separate PR. Is that acceptable?

CC: @vthumbe1503 @ptrendx

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -513,16 +561,14 @@ def general_grouped_gemm_for_grouped_tensor(
...
-        device=device,
-    )
+    workspace_setup = _get_grouped_gemm_setup_workspace(device.index, num_tensors)
+    # dgrad and wgrad must use distinct persistent cuBLAS workspaces: sharing one
+    # deadlocks under CUDA-graph replay on cuBLAS 13.6 (see _get_grouped_cublas_workspace).
+    # wgrad is the NT GEMM (transb=True); key the slot on the layout rather than on
+    # out-discreteness. A wgrad into a single grouped weight-grad (single_grouped_weight)
+    # has a GroupedTensor out, not a list, so an is_discrete_out proxy would collide it
+    # with dgrad on slot 0. fprop (TN) and dgrad (NN) can share slot 0; only NT is isolated.
+    workspace_cublas = _get_grouped_cublas_workspace(device.index, 1 if transb else 0)
```
vthumbe1503: Lets just  include layout(NN, TN, NT) as part of the function signature, so that we get different cached workspace for each layout. Although we face this problem only for wgrad now in TE. There are also some other reports of even fwd and dgrad conflicting with each other.

tests/pytorch/test_grouped_linear.py
```diff
@@ -1536,6 +1539,20 @@ def _run_grouped_linear_path(
     out_features = weights[0].size(0)
     use_fp8 = fp8_recipe is not None
 
+    # Spy on the grouped quantize entry point so a predicate that silently
+    # declines the fused path fails the test instead of vacuously matching the
+    # legacy reference.
+    group_quantize_calls = 0
```
vthumbe1503: This seems too much to be honest. We already have cuda graph test that verifies the graph safeness. If it is graph safe, it has to go through the group quantize route. So I dont think, counting the group_quantize calls and verifying is necessary

vthumbe1503 (dismissed):

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -435,6 +436,41 @@ def _get_fp32_zeros_tensor(num_tensors: int, device: torch.device) -> torch.Tens
...
+@functools.lru_cache(maxsize=None)
+def _get_grouped_gemm_setup_workspace(device: int, num_tensors: int) -> torch.Tensor:
+    """Persistent setup workspace (per-group pointer/dim arrays) for grouped-tensor GEMM.
+
+    This must not be allocated per call: under CUDA-graph capture a per-call
+    allocation's block returns to the shared capture pool as soon as the Python
+    reference dies, so the forward and backward graphs can alias the same block
+    and the captured GEMM's pointer/dimension arrays get overwritten at replay.
+    Consecutive GEMMs reusing one workspace are ordered by the stream, matching
+    how the non-grouped path shares its cached cuBLAS workspace.
```
timmoon10: Functionally I'm ok with caching the GGEMM workspace (it's a bit wasteful, but I don't expect the workspace to be that large anyways). This explanation is misleading though. It claims there's a bug when the forward and backward graphs alias the same workspace buffer, and the fix is to force aliasing the same buffer? The real bug is not in the workspace or the memory allocator, but in the cuBLAS kernel TMA descriptors. Caching the workspace is a hacky workaround until the cuBLAS fix is available.

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -435,6 +436,41 @@ def _get_fp32_zeros_tensor(num_tensors: int, device: torch.device) -> torch.Tens
...
+    """Persistent cuBLAS workspace for the grouped-tensor GEMM path, one per GEMM layout.
+
+    Two grouped-tensor GEMMs that share a single cuBLAS workspace can deadlock on the
+    second CUDA-graph replay: the grouped kernels interact through the shared workspace
+    and the second matmul hangs. It is deterministic per graph geometry and reproduces
+    on cuBLAS 13.5/13.6/13.7 (i.e. not fixed upstream as of 13.7); giving each GEMM its
+    own persistent workspace avoids it. The backward wgrad (NT) is the case seen in TE,
+    but fprop (TN) and dgrad (NN) have also been reported to conflict, so we isolate by
+    layout: TN, NN, and NT each get a distinct persistent workspace. Each is a single
+    persistent allocation, so CUDA-graph capture safety is preserved.
```
timmoon10: Now that we've gotten to the bottom of the bug (stale TMA descriptors), we should include that rather than this exploratory bug report. We should also note that the bug has been fixed in cuBLAS and that this hacky fix can be reverted in the future.

transformer_engine/pytorch/cpp_extensions/gemm.py
timmoon10: We've identified that the cuBLAS GGEMM kernel doesn't properly update the TMA descriptors under very specific circumstances, so if the workspace buffer changes between kernel calls then the second kernel might access the wrong buffer.

transformer_engine/pytorch/ops/basic/grouped_linear.py
timmoon10: My complaint is more about commment style, not functionality. Power-of-2 is a minor advanced config, and it should be documented where it is relevant. However, it seems the agent struggled with it at some point and now it brings it up as a top-level consideration everywhere even if it would be distracting to future users.

timmoon10 (dismissed):
This is functionally ready, although I would update the comments so future developers can understand the cuBLAS GGEMM bug with stale TMA descriptors. As it is, the comments make it seem the bug is in the memory allocator, which is wrong.

transformer_engine/pytorch/cpp_extensions/gemm.py
denera: Yeah, this comment is no longer relevant. It's left over from debugging when it wasn't yet clear whether it was the setup workspace or cuBLAS workspace causing the issue, but since then we've narrowed it down to stale descriptors in the cuBLAS workspace and the setup workspace does not need to be cached per GEMM layout (just number of tensors). I removed this comment entirely now.

transformer_engine/pytorch/cpp_extensions/gemm.py
denera: Yes, this was leftover from before cuBLAS devs diagnosed the issue. I updated the comment to reflect what the actual cause is.

transformer_engine/pytorch/ops/basic/grouped_linear.py
denera: There's only one place I see any mention of the power-of-2 in comments: https://github.com/NVIDIA/TransformerEngine/pull/3171/changes#diff-e52c6ddc8c0f4d20cb5aa832e92dd3794687bac1f828e367b526f17f524238a5R112-R113

I think there might have been more before because the PR originally did **not** support power-of-2 scales, and there had to be a lot of exceptions/branching to reconcile that with the fact that non-grouped/unfused FP8BS does (in fact it's the default, I believe).

Those comments/branching are removed now that grouped FP8BS has parity with non-grouped.

vthumbe1503 (approved):

PR #3172, opened 2026-07-02
[Pytorch][Bug] Requires Grad doesnt flow through Autograd boundaries for QuantizedTensor

vthumbe1503:
# Description

**Bugs**
- Requires Grad caching solution was introduced in this [PR](https://github.com/NVIDIA/TransformerEngine/pull/2559) that aimed at reducing CPU overheads in TE. Turns out when we pass QuantizedTensor across autograd boundaries, its corresponding requires_grad doesnt flow through. This is because requires_grad is set through grad_fn directly in C++ which bypasses our python cache. @pggPL originally caught this bug when running two back to back te.Linear layers with fp8_output=True, and found that the grads didnt flow.
- While debugging this bug, I also uncovered our Float8Blockwise Dequantize is not autograd aware similar to other QuantizedTensor.  This PR fixes that as well

**Alternative way of reducing CPU overheads for attribute access of custom torch Tensor:**
- The main root cause for requires_grad attribute access of a QuantizedTensor being slow was our custom torch function implementation. The implementation doesnt do anything and uses the _disabled_torch_function_impl. But the problem was presence of python implementation resulted in us crossing C++ to python boundary twice.

- When we do requires_grad access  of custom torch tensor it goes through this function in pytorch [C++](https://github.com/pytorch/pytorch/blob/7dc64495586d9ba7c8db82fa1f1588a9d3105390/torch/csrc/autograd/python_variable.cpp#L2911-L2912) which checks the presence of torch function and actually goes back to python to execute it.

- This can be avoided by assigning the sentinel value directly instead of defining a noop torch function
`    __torch_function__ = torch._C._disabled_torch_function_impl`
Torchao does the same for their custom MXFP8Tensor [here](https://github.com/pytorch/ao/blob/d1e0e2d8ee59cb2ccee19d303b20931e58820d8c/torchao/prototype/mx_formats/mx_tensor.py#L685-L686)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/test_quantized_tensor.py
+++ b/tests/pytorch/test_quantized_tensor.py
@@ -109,9 +109,9 @@ def to_float8_CS(
         quantizer.set_usage(rowwise=True, columnwise=False)
     return quantizer(tensor)
 
+
 def make_quantizer(quantization: str, device: torch.device = "cuda"):
-    """Construct a quantizer for the given quantization scheme.
-    """
+    """Construct a quantizer for the given quantization scheme."""
     if quantization in ("fp8", "fp8_delayed_scaling"):
         return Float8Quantizer(
             scale=torch.ones(1, dtype=torch.float32, device=device).squeeze(),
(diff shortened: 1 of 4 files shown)


timmoon10 (approved):
LGTM

PR #3178, opened 2026-07-06
grouped linear single param fixes

CarlosGomes98:
# Description

These changes fix convergence for grouped linear single param.
Issues encountered:

Without the TE fused ops path: causes the experts to get requires_grad=0 and the parent to not receive updates from the individual experts.

With TE fused ops: discards the high precision metadata, so the master weights were starting from dequantized mxfp8 weights, rather than the original high precision weights

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

These changes resolve the immediate issue, but cleaner medium- to long-term solutions would require broader changes.
In particular, the bug could likely have been avoided if Transformer Engine supported packing expert weights before quantization. The current flow quantizes each expert separately and then packs the resulting parameters, requiring the packing code to manually preserve and transfer FP8 initialization metadata. This stretches the metadata lifecycle and makes it easy to lose information needed to construct the optimizer’s high-precision master weights. Packing the high-precision weights first and quantizing the grouped parameter once would provide a cleaner and more robust ownership model.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/module/grouped_linear.py
+++ b/transformer_engine/pytorch/module/grouped_linear.py
@@ -80,6 +80,8 @@ class _GroupedWeightAutogradBridge(torch.autograd.Function):
 
     @staticmethod
     def forward(ctx, grouped_weight: torch.Tensor, fuse_wgrad_accumulation: bool):
+        # Backward only needs the parent's identity/attributes, not saved tensor values;
+        # use a weakref to avoid extending its lifetime or creating a reference cycle.
         ctx.grouped_weight_ref = weakref.ref(grouped_weight)
         ctx.fuse_wgrad_accumulation = bool(fuse_wgrad_accumulation)
         members = grouped_weight.quantized_tensors
--- a/transformer_engine/pytorch/module/grouped_linear.py
+++ b/transformer_engine/pytorch/module/grouped_linear.py
@@ -7,6 +7,7 @@
 from typing import Union, Optional, Callable, Tuple, List
 from itertools import chain
 import os
+from types import MethodType
 import warnings
 import weakref
 
@@ -66,6 +67,76 @@ from ...debug.pytorch.debug_quantization import DebugQuantizer
 from ...debug.pytorch.debug_state import TEDebugState
 
 __all__ = ["GroupedLinear"]
+
+
+def _get_high_precision_init_val(
+    grouped_weight: torch.Tensor,
+) -> Optional[torch.Tensor]:
+    """Return the pre-quantization initialization preserved on a grouped parameter."""
+    return getattr(grouped_weight, "_high_precision_init_val", None)
+
+
+def _clear_high_precision_init_val(grouped_weight: torch.Tensor) -> None:
+    """Release a grouped parameter's temporary high-precision initialization."""
+    if hasattr(grouped_weight, "_high_precision_init_val"):
+        del grouped_weight._high_precision_init_val
+
+
+class _GroupedWeightAutogradBridge(torch.autograd.Function):
+    """Connect cached per-GEMM views to their registered grouped parameter.
+
+    The legacy GroupedLinear autograd function consumes cached member tensors. Those
+    tensors are storage views, not registered parameters, so using them directly drops
+    the autograd edge to the grouped parent. This bridge makes the parent the autograd
+    input while preserving the member objects required by the grouped GEMM kernels.
+    """
+
+    @staticmethod
+    def forward(ctx, grouped_weight: torch.Tensor, fuse_wgrad_accumulation: bool):
+        ctx.grouped_weight_ref = weakref.ref(grouped_weight)
+        ctx.fuse_wgrad_accumulation = bool(fuse_wgrad_accumulation)
+        members = grouped_weight.quantized_tensors
+        if members is None:
+            members = grouped_weight.split_into_quantized_tensors()
+        return tuple(members)
+
+    @staticmethod
+    def backward(ctx, *member_grads):
+        grouped_weight = ctx.grouped_weight_ref()
+        if grouped_weight is None:
+            raise RuntimeError("Grouped weight was released before backward completed")
+
+        if ctx.fuse_wgrad_accumulation:
+            # GroupedLinear has already written each wgrad into a view of the parent's
+            # main_grad. Trigger the parent's accumulator/DDP hook with a zero dummy
+            # gradient and tell MCore not to add that dummy to main_grad.
+            if hasattr(grouped_weight, "grad_added_to_main_grad"):
+                grouped_weight.grad_added_to_main_grad = True
+            grouped_grad = torch.zeros(
+                tuple(grouped_weight.shape),
+                dtype=grouped_weight.dtype,
+                device=grouped_weight.device,
+            )
+        else:
+            members = grouped_weight.quantized_tensors
+            if members is None:
+                members = grouped_weight.split_into_quantized_tensors()
+            grouped_grad = torch.stack(
+                [
+                    (
+                        grad
+                        if grad is not None
+                        else torch.zeros(
+                            tuple(member.shape),
+                            dtype=grouped_weight.dtype,
+                            device=grouped_weight.device,
+                        )
+                    )
+                    for member, grad in zip(members, member_grads)
+                ],
+                dim=0,
+            )
+        return grouped_grad, None
 
 
 class _GroupedLinear(torch.autograd.Function):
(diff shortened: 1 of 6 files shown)


timmoon10 (commented):
If I understand, this PR fixes 3 bugs:
- The grouped linear module with a single grouped weight does not actually populate the param grad. It computes discrete weight grads, but it never ended up in the single weight param. The fix is to wrap the conversion from single grouped weights to discrete weights within an autograd function.
- The grouped linear module with a single grouped weight was not handling high-precision weight caching (used in Mcore DDP/FSDP). We were caching high-precision weights for the discrete weights, but they were not registered in the single weight param. The fix is to concat the discrete high-precision values and register it in the single weight param.
- When constructing a grouped linear op with `device="meta"`, it might incorrectly set some Mcore-specific param attrs before the params are initialized. Mcore does this when it constructs an internal fused to replace the grouped linear module implementation (see [here](https://github.com/NVIDIA/Megatron-LM/blob/509efe6c91f2dc1266e747184f947bafa6ec9e0c/megatron/core/transformer/moe/experts.py#L382)).

This is all hairy and messy, but the fixes seem reasonable. It seems that if you enable single grouped weights or any Mcore-specific optimizations, then grouped linear becomes extremely delicate. Users should avoid unless they know what they are doing.

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -218,7 +218,12 @@ def _apply_delay_wgrad_param_hooks(self) -> None:
         if not self.wgrad_store.delay_wgrad_compute():
             return
         if self.single_grouped_weight:
-            self.weight.skip_backward_post_hook = True
+            # A meta-device op may be created as a parameterless shell and have its
+            # grouped parent attached after construction. In that case there is no
+            # ``weight`` parameter to mark yet.
+            weight = self._parameters.get("weight")
```
zhongbozhu, line 240: without this this, will there be runtime errors in actual training? how do we trigger it?

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1904,6 +1994,28 @@ def _get_weight_tensors(self) -> List[Union[torch.Tensor, QuantizedTensorStorage
...
+                # The grouped parent is trainable, so make that state explicit on the
+                # bridged member outputs consumed by legacy GroupedLinear autograd.
+                for weight in weight_tensors:
+                    weight.requires_grad_(True)
+                if self.fuse_wgrad_accumulation:
+                    grouped_main_grad = getattr(grouped_weight, "main_grad", None)
+                    for index, weight in enumerate(weight_tensors):
+                        if grouped_main_grad is not None:
+                            weight.main_grad = grouped_main_grad[index]
+                        for attr in (
```
zhongbozhu, line 2018: why?

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1904,6 +1994,28 @@ def _get_weight_tensors(self) -> List[Union[torch.Tensor, QuantizedTensorStorage
...
+                            "grad_added_to_main_grad",
+                            "overwrite_main_grad",
+                            "zero_out_wgrad",
+                        ):
+                            if hasattr(grouped_weight, attr):
+                                setattr(weight, attr, getattr(grouped_weight, attr))
```
zhongbozhu, line 2024: why do we need this for loop to setattr?

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1490,9 +1567,22 @@ def make_grouped_weights(self, defer_init=False) -> None:
             and (weight_quantizers[0] is None or not weight_quantizers[0].internal)
         ):
             raise RuntimeError("Found internal quantizer with `single_grouped_weight=True`.")
+        grouped_parameter = torch.nn.Parameter(grouped_weights)
+        for member in grouped_parameter.quantized_tensors:
+            member.requires_grad_(grouped_parameter.requires_grad)
+        if all(value is not None for value in high_precision_init_vals):
+            _attach_high_precision_init_val(
```
zhongbozhu, line 1573: how does it avoid the dequantize exactly?

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -68,6 +69,66 @@
...
+            members = grouped_weight.split_into_quantized_tensors()
+        return tuple(members)
+
+    @staticmethod
+    def backward(ctx, *member_grads):
+        grouped_weight = ctx.grouped_weight_ref()
+        if grouped_weight is None:
+            raise RuntimeError("Grouped weight was released before backward completed")
+
+        if ctx.fuse_wgrad_accumulation:
```
zhongbozhu, line 100: There needs to be have a numerical test for both fuse_wgrad_accumulation and not fuse_wgrad_accumulation. Compared with discrete weights, with or without the fusion, the wgrad output should be extremely close to discrete.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1490,9 +1567,22 @@ def make_grouped_weights(self, defer_init=False) -> None:
...
+        grouped_parameter = torch.nn.Parameter(grouped_weights)
+        for member in grouped_parameter.quantized_tensors:
+            member.requires_grad_(grouped_parameter.requires_grad)
+        if all(value is not None for value in high_precision_init_vals):
+            _attach_high_precision_init_val(
+                grouped_parameter,
+                torch.stack(high_precision_init_vals, dim=0),
+            )
+            for weight in weights:
+                clear = getattr(weight, "clear_high_precision_init_val", None)
```
zhongbozhu: We need a less hacky implementation.

zhongbozhu (changes requested):

transformer_engine/pytorch/ops/basic/grouped_linear.py
CarlosGomes98: Yes. The real underlying issue is lack of explicit API support on the TE side for the MCore pattern of initializing ops with device=meta and then injecting the module weights later. Replicating the issue requires 

USE_TE_OPS=True
MOE_SINGLE_GROUPED_WEIGHT=True
NVTE_GROUPED_LINEAR_SINGLE_PARAM=1
DELAY_WGRAD_COMPUTE=True
NVTE_CUTEDSL_FUSED_GROUPED_MLP=1

The order is: 

1. TE GroupedLinear OP created by MCore with device meta
2. It created placeholder parameters, but `weight` is not created yet
3. This part of the code before the changes calls self.weight.skip_backward_post_hook = True, but weight does not exist yet.

In the current code, self._parameters.get("weight") will return None in this case, self.weight wont be accessed. In the "regular" case where device=meta is not used, it will work as normal.

transformer_engine/pytorch/module/grouped_linear.py
CarlosGomes98: This is how TE's initialization contract for high precision works. If the parameter has a method `get_high_precision_init_val`, then MCore uses that to initialize the optimizer master weights. If its not present, then it dequantizes the model weights.

https://github.com/NVIDIA/Megatron-LM/blob/eadbaa6189fbd272843b4e2bd75ae20984f891d4/megatron/core/optimizer/distrib_optimizer.py#L412

transformer_engine/pytorch/module/grouped_linear.py
CarlosGomes98: We can hide this away by directly using the _ methods I introduced in base.py, but its effectively doing the same thing.

I dont disagree with your general idea that a lot of this is hacky, but as far as I'm aware this is all we can do given the current TE API. Making it less hacky would require a non-trivial redesign and implementation work which I wouldnt want to add to this PR

transformer_engine/pytorch/module/grouped_linear.py
CarlosGomes98: This non-fused path will operate on the individual expert weights, whereas MCore is attaching distributed training flags to the grouped parent. So we need to pass these flags down to the expert weights.

transformer_engine/pytorch/module/grouped_linear.py
CarlosGomes98: Same as below:

This non-fused path will operate on the individual expert weights, whereas MCore is attaching distributed training flags to the grouped parent. So we need to pass these flags down to the expert weights.

transformer_engine/pytorch/module/grouped_linear.py
CarlosGomes98: Agreed, added a numerical test

CarlosGomes98:
Closing in favor of https://github.com/NVIDIA/TransformerEngine/pull/3225 and https://github.com/NVIDIA/TransformerEngine/pull/3224

PR #3186, opened 2026-07-07
[Common] Pass cu_seqlens and token-unit ragged offsets directly to cuDNN SDPA fprop

egilliam-nv:
# Description

cuDNN >= 9.24 SDPA (unified engine) accepts cumulative sequence lengths directly (`cu_seq_len_q/kv`) and can scale ragged offsets stored in coarser units (e.g. tokens) back to element offsets via a per-tensor ragged offset multiplier. This PR uses both features in the fused-attention forward passes to skip the conversion kernels that previously ran before every varlen call: `cu_seqlens_to_actual_seqlens` and (f16 THD only) `cu_seqlens_padded_to_offsets`.

**f16/bf16 forward** (cuDNN >= 9.24): binds the user's int32 `cu_seqlens` buffers as `CU_SEQ_LEN_Q/KV` for the padding mask, and the token-unit `cu_seqlens_padded` buffers as ragged offsets for Q/K/V/O/Stats with elements-per-token multipliers. Both conversion kernels are skipped and no conversion workspace is allocated.

**fp8/mxfp8 forward** (cuDNN >= 9.25 and cudnn-frontend >= 1.26): same for `cu_seqlens` (the fp8 path has no THD support, so only the one conversion kernel existed there). The frontend is header-only, so its version is a compile-time property; the gate is a constant-folded `CUDNN_FRONTEND_VERSION` check — all referenced symbols exist in 1.25 (`SDPA_fp8_attributes` is an alias of `SDPA_attributes`), so no preprocessor guards are needed and the fp8 direct path is simply dead code on 1.25 builds.

Notes for reviewers:
- `CU_SEQ_LEN_*` inputs pin implementation selection to the unified engine (the composite support surface rejects them), so the gates keep composite-served configs on the legacy path: training with dropout (the frontend rejects dropout together with generated stats on the unified engine, and TE always generates stats), and for f16 the direct path only engages for THD/ragged configs, which backend selection restricts to unified-capable hardware.
- The direct path keeps the true batch size instead of the quantized `max_b`: cuDNN reads the user's `[actual_b+1]` cu_seqlens buffers, so a quantized batch would read out of bounds. Token-dim bucketing is unchanged.
- The layout-group -> multiplier mapping is factored into `RaggedOffsetMultipliers` (utils.h), shared by the graph builder and the legacy conversion kernel so the two cannot drift. The kernel rewrite also removes a cross-thread read (`offsets_v[tid] = offsets_k[cu_seqlens_id]`) that raced for quantized-batch tail entries with interleaved layouts.
- Backward passes are unchanged (no backend support yet).
- `NVTE_FUSED_ATTN_DIRECT_SEQLENS=0` disables the new path (kept as an escape hatch for A/B testing; can be removed on request).

Validation against cuDNN 9.25, no-fallback design so passes prove the unified-engine path:
- f16: `test_dpa_softmax_thd` 15/15 with the direct path on and off (H100 + Blackwell), matching pre-change baseline; direct-vs-legacy fused outputs/grads match for all THD layouts (`thd_thd_thd`, `t3hd`, `th3d`, `thd_t2hd`, `thd_th2d`) x MHA/GQA x padding/padding_causal x pad_between_seqs {false, true} on H100 and Blackwell (14/14 each).
- fp8 (with cudnn-frontend 1.26 built from source): `test_dpa_fp8_vs_f16` padding configs — 56 passed on H100 (delayed + current scaling), 168 passed on Blackwell (adds MXFP8), zero failures.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- f16/bf16 fwd: pass `cu_seqlens` / `cu_seqlens_padded` directly to cuDNN (unified engine) on cuDNN >= 9.24, skipping both conversion kernels. (Also cudnn-frontend must be >= 1.25, but this is now required across the board by TransformerEngine so no need to check for this in the fp16 code.)
- fp8/mxfp8 fwd: pass `cu_seqlens` directly on cuDNN >= 9.25 + cudnn-frontend >= 1.26, skipping the conversion kernel
- Factor ragged-offset multipliers into a shared `RaggedOffsetMultipliers` helper used by both the graph builders and the legacy conversion kernel

# Checklist:

### DIFF
--- a/transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
+++ b/transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
@@ -94,6 +94,18 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
     NVTE_CHECK(is_padding, "Paged attention requires padding mask!");
   }
 
+  // Newer versions of cuDNN SDPA can accept sequence lengths directly as a cumulative
+  // tensor, and can accept ragged offsets in arbitrary units (such as tokens) instead
+  // of elements. Take advantage of this if possible to avoid 2 extra kernel calls.
+  const bool use_direct_seqlens =
+      transformer_engine::getenv<bool>("NVTE_FUSED_ATTN_DIRECT_SEQLENS", true) &&
+      cudnn_runtime_version >= 92400 &&
+      // This extra restriction is needed because cuDNN frontend doesn't yet allow
+      // the combination of dropout and stats generation for the fprop unified engine,
+      // so any such request would always get routed to the old composite SDPA engine
+      // (which doesn't support cu_seqlens). Remove this restriction when possible.
+      !is_dropout;
+
   // keep original batch size because cu_seqlens are created with [b+1] shape
   int64_t actual_b = b;
   if ((is_ragged_q || is_ragged_kv) && cudnn_runtime_version >= 90600) {
@@ -103,14 +115,26 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
     // so the check passes; ragged offset still provides variable-length boundaries.
     if (sm_arch_ != 120) {
       // replace batch size and maximum sequence lengths with maximum token counts
-      // for query and key/value so the graph is static within each quantization bucket
-      b = max_b;
+      // for query and key/value so the graph is static within each quantization bucket.
+      // With direct seqlens, keep the true batch size: cuDNN reads the user's
+      // [actual_b+1] cu_seqlens buffers, so a quantized batch would read out of bounds.
+      if (!use_direct_seqlens) {
+        b = max_b;
+      }
       s_q = is_ragged_q ? max_t_q : s_q;
       s_kv = is_ragged_kv ? max_t_kv : s_kv;
     }
   }
 
-  const DType ragged_offset_type = cudnn_runtime_version >= 90500 ? DType::kInt64 : DType::kInt32;
+  const DType ragged_offset_type =
+      use_direct_seqlens
+          ? DType::kInt32  // Direct cu_seqlens are given to us as int32, so keep it that way.
+          : (cudnn_runtime_version >= 90500 ? DType::kInt64 : DType::kInt32);
+
+  // Ragged offset multipliers (elements per token); shared with the legacy conversion
+  // kernel (cu_seqlens_padded_to_offsets) so the two paths cannot drift apart.
+  const RaggedOffsetMultipliers offset_mults(layout_group, h, hg, d_qk, d_v);
+
   bool generate_stats = true;  // Always return stats
   try {
     FADescriptor_v1 descriptor{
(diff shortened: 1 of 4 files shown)


transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -94,6 +94,18 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
     NVTE_CHECK(is_padding, "Paged attention requires padding mask!");
   }
 
+  // Newer versions of cuDNN SDPA can accept sequence lengths directly as a cumulative
+  // tensor, and can accept ragged offsets in arbitrary units (such as tokens) instead
+  // of elements. Take advantage of this if possible to avoid 2 extra kernel calls.
+  const bool use_direct_seqlens =
```
sudhakarsingh27: small nit: can we use `pass_cu_seqlens_tensor` or `use_cu_seqlens_directly` instead of `use_direct_seqlens` to be more explicit?

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -103,14 +117,26 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
...
+      }
       s_q = is_ragged_q ? max_t_q : s_q;
       s_kv = is_ragged_kv ? max_t_kv : s_kv;
     }
   }
 
-  const DType ragged_offset_type = cudnn_runtime_version >= 90500 ? DType::kInt64 : DType::kInt32;
+  const DType ragged_offset_type =
+      use_direct_seqlens
+          ? DType::kInt32  // Direct cu_seqlens are given to us as int32, so keep it that way.
```
sudhakarsingh27: nit: cu_seqlens* tensors are given to use directly...

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
```diff
@@ -437,18 +496,23 @@ void fused_attn_arbitrary_seqlen_fwd_impl(
     // Exit to request upper level API to allocate memory if needed
     // n.b. Care should be taken to align each of the added worksapce tensors to their type.
     // We do this by adding padding at the end of each separate allocation.
+    // With direct seqlens, no conversion workspace is needed: cuDNN consumes the
```
sudhakarsingh27: nit: When passing cu_seqlens* directly to cudnn sdpa, ...

sudhakarsingh27 (commented):
Just found some grammar nits, otherwise looks good to me

transformer_engine/common/fused_attn/fused_attn_f16_arbitrary_seqlen.cu
egilliam-nv: Done.

sudhakarsingh27 (approved):
LGTM!

PR #3189, opened 2026-07-08
[PyTorch][torch.compile] Make get_attention_backend traceable without graph breaks

pggPL:
# Description

This PR is part of an effort to enable `torch.compile` support for attention: it makes `get_attention_backend` traceable.

It works on current PyTorch as-is — there is no need to wait for the two related PRs I opened in PyTorch. Those only improve the remaining rough edge: pytorch/pytorch#189042 (bigger and much more important; among other things it handles symbolic scalar arguments, which with stock dynamo must stay non-symbolic in the backend-selection probe — with static scalars, the common case, everything in this PR already works) and pytorch/pytorch#189027 (small, and possibly not needed anymore since this PR reads env vars via `os.environ.get`, which stock dynamo already guards).

`get_attention_backend` currently graph-breaks under `torch.compile`, forcing backend selection to run as an eager island. This PR makes it fully traceable with `fullgraph=True`, with dynamo guards on the `NVTE_*` environment variables so that changing an env var triggers recompilation instead of silently reusing a stale backend selection.

## Type of change

- [x] Breaking change (see the FusedAttnBackend note below)
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Read `NVTE_*` env vars in `get_attention_backend` via `os.environ.get` instead of `os.getenv`: dynamo installs value guards on `os.environ` accesses, while `os.getenv` reads are unguarded (the stdlib-global source is skipped) and would bake a stale backend selection into the compiled graph.
- Wrap the `tex.get_fused_attn_backend` pybind call in a `torch.compiler.assume_constant_result` helper — the result depends only on the attention configuration, not on tensor values.
- Mark `get_device_compute_capability` and `get_cudnn_version` with `assume_constant_result` (pybind/CUDA property calls are not traceable, and their results are constant for a process).
- Use a no-op logger when compiling (`logging.Logger` methods graph-break in dynamo) and skip the debug-log blocks that evaluate `int()`/`str()` on pybind enum values during tracing. Backend-selection debug logs are unchanged in eager mode.
- Add `test_get_attention_backend_traceable` to `tests/pytorch/test_torch_compile.py`: compiles a function calling `get_attention_backend` with `fullgraph=True` (any graph break fails the test) and flips `NVTE_FUSED_ATTN`/`NVTE_UNFUSED_ATTN`/`NVTE_FLASH_ATTN` to verify the guards trigger recompilation and the result keeps matching eager.

- `FusedAttnBackend` (in `pytorch/cpp_extensions/fused_attn.py`) is now a python `IntEnum` generated at import time from `tex.NVTE_Fused_Attn_Backend.__members__`, replacing the previous str->pybind-enum dict, and all direct python-side uses of the pybind enum are replaced with it. Rationale: the pybind enum is not traceable by dynamo (C `__eq__`), and a pybind enum baked through `assume_constant_result` produces guards dynamo cannot evaluate when compared against module-level enum values (hard crash when cuDNN rejects a config). The transition follows the pattern used for `constants.DType`: explicit members pinned to the C values with an import-time sync assert, an `__eq__` override so mixed comparisons with `tex.NVTE_Fused_Attn_Backend` stay equivalent regardless of the pybind11 version, and a `cast()` classmethod; `fused_attn_fwd/bwd` normalize their backend argument through `cast()`, so callers still passing the pybind enum keep working. Name lookup `FusedAttnBackend["FP8"]` and `int(...)` behave as before. The remaining (minor) breaking surface is `isinstance` checks against the pybind enum and dict-API such as `.items()`/iteration; no such uses exist inside TE and none were found in Megatron-LM/NeMo.

Note: dynamo only guards `os.environ` keys that exist at trace time; reads of absent keys are not guarded yet (upstream PyTorch limitation).

# Checklist:


cyanguwa (dismissed):

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -324,6 +325,53 @@ def __eq__(self, other):
...
+    Stand-in for the "DotProductAttention" logger used when get_attention_backend
+    is traced by torch.compile. logging.Logger methods are not traceable by dynamo
+    (they cause graph breaks), while this class's no-op methods are inlined away.
+    """
+
+    def debug(self, *args, **kwargs):
+        """No-op."""
+
+    def warning(self, *args, **kwargs):
+        """No-op."""
```
shino16, line 342: We can add `info()` and `error()` too.

transformer_engine/pytorch/cpp_extensions/fused_attn.py
```diff
@@ -97,11 +98,65 @@
...
+    def __eq__(self, other: object) -> bool:
+        # ``FusedAttnBackend`` is an ``IntEnum`` while ``NVTE_Fused_Attn_Backend``
+        # is a pybind11 enum. Compare by integer value so the two enums stay
+        # equivalent regardless of the pybind11 version (the pybind ``__eq__``
+        # handles the reverse order).
+        if isinstance(other, NVTE_Fused_Attn_Backend):
+            return int(self) == int(other)
+        return int.__eq__(self, other)
+
+    def __ne__(self, other: object) -> bool:
```
shino16, line 143: Shouldn't this be omitted and be left to the default behavior?

shino16 (approved):

transformer_engine/pytorch/cpp_extensions/fused_attn.py
pggPL: Good instinct, but here it's load-bearing because `FusedAttnBackend` is an `IntEnum`. The "Python 3 auto-derives `__ne__` from `__eq__`" rule only applies to `object` subclasses — `int` defines its own `__ne__` slot, which **shadows** that default. So dropping this method makes `!=` resolve to `int.__ne__`, **not** the negation of our custom `__eq__`.

That reintroduces exactly the pybind-version dependence this class avoids: for `backend != nvte_pybind_enum`, `int.__ne__(int_val, enum)` returns `NotImplemented`, and whether the reflected `enum.__ne__(int_val)` gives the right answer depends on the pybind11 build. On a build whose enum doesn't compare against ints, `!=` falls back to identity and wrongly returns `True` even when the values are equal.

Minimal repro:
```python
from enum import IntEnum
class E(IntEnum):
    A = 1
    def __eq__(self, o): return int(self) == int(getattr(o, 'value', o))
    __hash__ = int.__hash__
    # no __ne__
class Strict:  # pybind enum that only compares to its own type
    value = 1
    def __eq__(self, o): return isinstance(o, Strict)
print(E.A == Strict())  # False? -> our __eq__ runs
print(E.A != Strict())  # -> True via int.__ne__, our __eq__ never consulted
```
The explicit `__ne__` delegating to `__eq__` keeps `!=` consistent with `==` regardless of pybind11 version, so I'd prefer to keep it. Happy to add a one-line comment noting the `int.__ne__` shadowing if that helps future readers.

cyanguwa (approved):

PR #3194, opened 2026-07-08
Work around nvcc 13.0 parse failure in group_quantize_fp8.cuh

xiuhu17:
## Scenario

Building with CUDA 13.0 (nvcc V13.0.88, sm100) fails in every TU that includes `group_quantize_fp8.cuh` (e.g. `activation/glu.cu`):

```
error: template argument 2 is invalid
error: 'IVecT' has not been declared
```

Root cause is an nvcc 13.0 cudafe++ bug: inside the macro-expanded `TRANSFORMER_ENGINE_*_SWITCH` block, the discarded `if constexpr` branch constant-folds the `constexpr size_t flat_nvec` template argument and emits an ill-formed two-token functional cast `unsigned long((8))` into the intermediate C++, which the host compiler then rejects. Reproduces with both g++-12 and g++-13 hosts; the identical code hand-expanded (no macro) compiles fine. Confirmed with a 29-line minimal repro.

## Fix

Replace the `using IVecT/OVecT = Vec<IType/OType, flat_nvec>` aliases (only used for `::BYTES` in the alignment checks) with the equivalent `constexpr size_t` byte counts (`Vec<T, N>::BYTES == N * sizeof(T)`, utils.cuh). Semantics-identical, host-side only, no kernel change. Verified: sm100 wheel builds clean on CUDA 13.0 and the NVFP4/FP8 paths run end-to-end.

### DIFF
--- a/transformer_engine/common/cast/fp8/group_quantize_fp8.cuh
+++ b/transformer_engine/common/cast/fp8/group_quantize_fp8.cuh
@@ -1100,12 +1100,15 @@ void group_quantize(const GroupedTensor *input, const Tensor *noop, GroupedTenso
                     const size_t rows_per_tensor = first_logical_dim / num_tensors;
                     if constexpr (SCALING_TYPE == ScalingType::ROWWISE) {
                       constexpr size_t flat_nvec = ROWWISE_FLAT_LOAD_SIZE_BYTES / sizeof(IType);
-                      using IVecT = Vec<IType, flat_nvec>;
-                      using OVecT = Vec<OType, flat_nvec>;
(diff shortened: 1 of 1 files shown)


vthumbe1503 (approved):
LGTM

ptrendx (approved):

xiuhu17:
@vthumbe1503 @ptrendx Thanks for reviewing! Could anyone help to merge the pr? Thanks!

ptrendx:
Merged. Thank you @xiuhu17 !

xiuhu17:
@ptrendx  Thanks for the help!

PR #3127, opened 2026-06-14
Expert Parallelism: common C API + NCCL EP backend

timmoon10:
This is a resubmission of https://github.com/NVIDIA/TransformerEngine/pull/3034, which was reverted in https://github.com/NVIDIA/TransformerEngine/pull/3126 due to CI issues.

Implementation is from @phu0ngng and it is already approved by @ptrendx and @timmoon10.

PR #3196, opened 2026-07-09
Install cuDNN frontend in GitHub builds

sudhakarsingh27:
## Summary

- Install the declared `nvidia-cudnn-frontend>=1.25.0` build dependency in every GitHub Actions build environment.
- Keep the workflow requirement aligned with `pyproject.toml`.

## Root cause

TransformerEngine now resolves cuDNN frontend headers from the installed Python distribution. The GitHub Actions builds use `--no-build-isolation`, so build-system requirements are not installed automatically and metadata generation fails with `PackageNotFoundError`.

## Validation

| Scenario | Result |
| --- | --- |
| Workflow YAML parsing | Passed with PyYAML |
| Diff hygiene | `git diff --check` passed |
| Package availability | Python 3.10 and 3.12 x86_64 wheels are published for the currently resolved 1.26.0 release |
| GitHub Actions | [Build run 28997378989](https://github.com/NVIDIA/TransformerEngine/actions/runs/28997378989) pending |

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,7 @@ jobs:
         run: |
           apt-get update
           apt-get install -y git python3.9 pip cudnn9-cuda-12
-          pip install cmake==3.21.0 pybind11[global] ninja
+          pip install cmake==3.21.0 pybind11[global] ninja "nvidia-cudnn-frontend>=1.25.0"
       - name: 'Checkout'
         uses: actions/checkout@v3
         with:
@@ -77,7 +77,7 @@ jobs:
           docker exec builder bash -c '\
             apt-get update && \
             apt-get install -y git python3.9 pip cudnn9-cuda-12 && \
-            pip install cmake torch ninja pydantic importlib-metadata>=1.0 packaging pybind11 numpy einops onnxscript && \
+            pip install cmake torch ninja pydantic importlib-metadata>=1.0 packaging pybind11 numpy einops onnxscript "nvidia-cudnn-frontend>=1.25.0" && \
             apt-get clean \
           '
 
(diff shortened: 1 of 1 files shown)


ptrendx:
I think there is an issue here that we shouldn't just WAR - the cudnn frontend should be installed as a dependency of TE, the fact that it does not do that is a problem.

.github/workflows/build.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         run: |
           apt-get update
           apt-get install -y git python3.9 pip cudnn9-cuda-12
-          pip install cmake==3.21.0 pybind11[global] ninja
+          pip install cmake==3.21.0 pybind11[global] ninja "nvidia-cudnn-frontend>=1.25.0"
```
timmoon10, line 26: This one shouldn't be necessary since `nvidia-cudnn-frontend` is listed as a dependency here: https://github.com/NVIDIA/TransformerEngine/blob/01bbe3eea2768bc15b99ba0752ff60b1317b6b85/pyproject.toml#L6

```suggestion
          pip install cmake==3.21.0 pybind11[global] ninja
```

The PyTorch and JAX builds do need to install cuDNN Frontend manually because they are installed with `pip install --no-deps`.

timmoon10:
It seems we are still experiencing build failures because NCCL EP added a NCCL dependency: https://github.com/NVIDIA/TransformerEngine/pull/3127

.github/workflows/build.yml
sudhakarsingh27: I think the problem is building TE with `pip install --no-build-isolation .` which might be disregarding `pyproject.toml`

vcherepanov-nv: I think requires list in pyproject.toml doesn't apply if we use --no-build-isolation, so installing nvidia-cudnn-frontend is necessary.

vcherepanov-nv (commented):
LGTM, thanks!

vcherepanov-nv (approved):

PR #3195, opened 2026-07-09
docs: add datatype and hardware support matrix

CodersAcademy006:
## Problem

There is no single place in the docs that states which low-precision formats
are supported on which GPU architecture. The information is spread across the
README, inline compute-capability checks, and test skip conditions. Several
issues ask for exactly this (#2512, #2304), and related threads keep hitting the
same confusion about SM120 support (#2668, #2255).

## Solution

Add `docs/support_matrix.rst`, a page that documents format support keyed by
compute capability, since that is the value Transformer Engine checks at runtime.
It covers BF16, FP8 per-tensor, FP8 block scaling, MXFP8, and NVFP4, plus the
default recipe selected per architecture.

The tables are derived directly from the runtime capability checks so docs and
behavior stay aligned:

- `_compute_fp8_support`, `_compute_mxfp8_support`, `_compute_nvfp4_support`,
  `_compute_fp8_block_scaling_support`, and `get_default_fp8_recipe` in
  `transformer_engine/pytorch/quantization.py`
- the BF16 check in `transformer_engine/pytorch/utils.py`

Two nuances are called out explicitly because they are common sources of
confusion:

- MXFP8 is not currently supported on compute capability 12.0 and higher; it is
  limited to 10.0 through 10.x.
- NVFP4 is reported as available on 10.0 and higher, but the default recipe uses
  stochastic rounding whose FP4 conversion instructions are architecture specific
  to 10.0 and 10.3. On 12.0 the default recipe raises an architecture-specific
  error and round-to-nearest is required.

The page is linked under the Getting Started section in `docs/index.rst`.

## Testing

Validated the reStructuredText structure with docutils. I was not able to run a
full Sphinx build locally because that requires a compiled Transformer Engine,
which needs a CUDA GPU I do not have access to. A reviewer running the docs build
in CI can confirm the page renders as expected.

## Limitations

This page documents the current state of the runtime capability checks. If those
checks change, the tables need to be updated to match. A note in the page states
this so the two stay in sync.

Closes #2512
Closes #2304
Refs #2668
Refs #2255

PR #3197, opened 2026-07-09
Add transformer_engine.test() installation smoke check

### LINKED ISSUE #170: Add: transformerengine.test()
When end user install as documented as:
```
pip install git+https://github.com/NVIDIA/TransformerEngine.git@stable
```

They can't run an TE test to make sure the software stack work well.
We should do as NumPy and add `transformerengine.test()` (equivalent of  `numpy.test()`).

This way they won't need to clone the repo and checkout the right branch, to test the installation.

CodersAcademy006:
## Problem

There is no way to verify a Transformer Engine installation without cloning the
repository and running the test suite. Users installing from PyPI or from source
have no quick sanity check that the package and its framework extensions loaded
correctly on their machine (#170).

## Solution

Add a top-level `transformer_engine.test()` entry point, similar to
`numpy.test()`. It runs a small set of smoke checks:

- Installation integrity, reusing the existing PyPI sanity check.
- For the PyTorch backend, when it is available: a minimal `te.Linear` forward
  pass on the current device, reporting the device compute capability and which
  low-precision formats it supports. Checks that need a GPU are skipped when no
  CUDA device is present, so the call is safe to run anywhere.

It prints the result of each check and returns `True` if every executed check
passed.

```python
import transformer_engine
transformer_engine.test()
```

Closes #170

### DIFF
--- a/transformer_engine/__init__.py
+++ b/transformer_engine/__init__.py
@@ -109,3 +109,82 @@ except FileNotFoundError as e:
             )
 
 __version__ = str(metadata.version("transformer_engine"))
+
+
+def test(verbose: bool = True) -> bool:
+    """Run smoke checks to verify the Transformer Engine installation.
+
+    Confirms the package is installed correctly and, when the PyTorch backend is
+    available, runs a minimal functional check on the current device. Checks that
+    need a GPU are skipped when no CUDA device is available, so the call is safe
+    to run anywhere as an installation sanity check.
+
+    Parameters
+    ----------
+    verbose : bool, default = True
+        Print the result of each check.
+
+    Returns
+    -------
+    bool
+        ``True`` if every executed check passed, ``False`` otherwise.
+    """
+    # ponytail: smoke test, not the full suite. Covers install integrity and a
+    # single PyTorch forward pass; deeper coverage stays in tests/ and qa/.
+    results = []
+
+    def _record(name: str, passed: bool, detail: str = "") -> None:
+        results.append(passed)
+        if verbose:
+            status = "PASS" if passed else "FAIL"
+            print(f"[{status}] {name}" + (f": {detail}" if detail else ""))
+
+    # Installation integrity (reuses the PyPI sanity check).
+    try:
+        from .common import sanity_checks_for_pypi_installation
+
+        sanity_checks_for_pypi_installation()
+        _record("installation", True, f"transformer_engine {__version__}")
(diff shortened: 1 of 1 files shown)


CodersAcademy006:
@ptrendx @phu0ngng same request as on #3195: open since July 9, no reviewer assigned, no human comment yet.

It adds a `transformer_engine.test()` entry point so that someone installing from PyPI or from source can confirm the package and its framework extensions actually loaded on their machine, without cloning the repo and running the full test suite, which is what #170 asks for. It is 79 lines in `transformer_engine/__init__.py` and adds no dependencies. If you would rather this lived under a different name, or as a console script instead of a top-level function, tell me which and I will move it.

ptrendx:
@fheinecke Could you take a look?

transformer_engine/__init__.py
```diff
@@ -109,3 +109,82 @@ def require_nccl_ep() -> None:
             )
 
 __version__ = str(metadata.version("transformer_engine"))
+
+
+def test(verbose: bool = True) -> bool:
```
fheinecke, line 114: What do you think about renaming this to `smoke_test` or `test_installation`? The current behavior (running a simple, single pass/fail test) doesn't quite match numpy's `test()`, which runs a full test suite.

fheinecke (commented):
This is fine if it fixes a specific problem that users are facing right now, but I think most of the actual checks are redundant. When importing this test function, the `transformer_engine.common` module [is also imported](https://github.com/NVIDIA/TransformerEngine/pull/3197/changes#diff-7c64219527bef4e5946611720c9ab9d5c018b252a501e00c6fb4394b23a47b6dR14), which [loads the required shared libraries](https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/common/__init__.py#L363-L386). If these raise an exception, then the import will fail before the test func can be called.

CodersAcademy006:
@fheinecke @ptrendx thanks for merging #3195 yesterday. This is its sibling and the last thing I have open here: `transformer_engine.test()`, so that someone installing from PyPI or from source can confirm the extension actually imports and runs before they take a training job down with it. It closes #170, is 79 lines, adds no dependency, and is green at 4/4.

@fheinecke you looked at it on 11 August and it has been quiet since. If the entry point is the sticking point I am happy to rename it, move it behind a console script, or fold it into the existing test layout instead, whichever fits the direction now that AGENTS.md is in the tree.

PR #3201, opened 2026-07-10
[PyTorch][torch.compile] Support for UnfusedDotProductAttention

pggPL:
# Description

Make the `UnfusedDotProductAttention` backend traceable by `torch.compile(fullgraph=True, mode="reduce-overhead")`, so the forward and backward can be captured into CUDA graphs without graph breaks.

Scope:
1. **bf16/fp16 path (compile-supported)**: register the TE softmax kernels and THD<->BSHD conversion helpers as `torch.library.custom_op`s with fake impls and autograd bindings; remove an unbacked-SymInt `.item()` from the hot path of `ConvertBSHDtoTHD`.
2. **FP8 is explicitly NOT supported under torch.compile**: with `fp8=True` (emulation) and/or `fp8_output=True` (Float8Tensor output, a tensor subclass that cannot cross a graph boundary) the backend runs as an eager island — the forward dispatches to a `torch._dynamo.disable`'d wrapper, the same mechanism `DotProductAttention` and `FusedAttention` use module-wide. FP8 attention always involves delayed scaling regardless of the recipe: S and dP are produced inside the kernel, so their amax cannot be known before quantization and they use delayed-scaling quantizers even under Float8CurrentScaling (see `DPA.init_fp8_metadata`) — and delayed scaling (`Float8Quantizer`, tensor scale/amax state) is not supported under torch.compile.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- `softmax.py` / `softmax.cpp`: `scaled_*_softmax_{forward,backward}` as custom ops; C++ backward kernels allocate a fresh output buffer instead of writing in-place into `output_grad` (custom ops and cudagraph trees forbid input aliasing).
- `utils.py`: `ConvertTHDtoBSHD` / `ConvertBSHDtoTHD` as custom ops; `num_tokens` passed by the caller instead of `cu_seqlens[-1].item()`.
- `backends.py`: `UnfusedDotProductAttention.forward` dispatches FP8 calls to an eager (dynamo-disabled) wrapper; the non-FP8 path is traced with no graph breaks.
- `tests/pytorch/test_torch_compile.py`: `test_unfused_dpa_torch_compile` (5 qkv layouts, fullgraph + reduce-overhead, fwd+bwd captured into CUDA graphs and replayed).

# Checklist:


cyanguwa (commented):
LGTM but please take a look at Greptile's few comments. Thanks.

cyanguwa (approved):

PR #3205, opened 2026-07-13
[PyTorch] Support FA3/FA4 with pad_between_seqs=True

xiaoyao0115:
## Description

Support FA3/FA4 with `pad_between_seqs=True` in THD attention, including non-CP, P2P CP, and all-to-all (A2A) CP.

This support requires attention output and Q/K/V gradients to be exactly zero at physical padding positions. FA3/FA4 can leave these positions unwritten, and CP partial-output merges can write them again. This PR adds the necessary padding cleanup and correctness checks.

The padding mask also handles reserved buffer capacity after the final padded sequence boundary. `num_tokens` is the full physical buffer capacity and may exceed `cu_seqlens_padded[-1]`; all positions beyond that boundary are treated as padding. For example, a 16-token buffer with `cu_seqlens=[0,3,8]` and `cu_seqlens_padded=[0,4,12]` includes tail padding at positions `[12,16)`, which map to a zero-length sentinel sequence instead of indexing beyond the sequence metadata.

## Changes

- Mask non-CP FlashAttention output padding and use an autograd identity to zero Q/K/V-gradient padding without copying forward inputs.
- Clean P2P output padding after merging partial outputs, and A2A output/gradient padding before inverse communication changes token order. Added cleanup is confined to FlashAttention.
- Make the shared THD padding mask safe for reserved buffer tails using device-side operations.
- Add exact-zero assertions and focused padding/CUDA Graph regression tests with an independently compacted reference. Query `get_attention_backend` to skip unsupported configurations while asserting that the requested FA version actually runs.

Padding cleanup must remain CUDA-graph-capturable in both non-CP and supported CP paths. The added in-tree capture/replay regression covers non-CP attention with raw CUDA Graph capture. CP capture uses TE's existing `make_graphed_callables` path; this PR does not extend raw-capture support for CP.

FA4 backend selection and varlen metadata plumbing already exist on the base branch. FA2 with inter-sequence padding and the unsupported padded A2A+P2P combination remain outside this change.

### DIFF
--- a/tests/pytorch/attention/test_attention.py
+++ b/tests/pytorch/attention/test_attention.py
@@ -1352,6 +1352,26 @@ def _run_dot_product_attention(
             return out, max_logit, (None, None, None, d_softmax_offset)
     if backend in ["FusedAttention", "FlashAttention"]:
         if qkv_format == "thd" and pad_between_seqs:
+            tensors_and_boundaries = [("out", out, cu_seqlens_q_after_pad)]
+            if is_training:
+                tensors_and_boundaries.extend(
+                    [
+                        ("dq", q.grad, cu_seqlens_q_after_pad),
+                        ("dk", k.grad, cu_seqlens_kv_after_pad),
+                        ("dv", v.grad, cu_seqlens_kv_after_pad),
+                    ]
+                )
+            for tensor_name, tensor, cu_seqlens_after_pad in tensors_and_boundaries:
+                for i in range(1, config.batch_size + 1):
+                    pad_range = (
+                        cu_seqlens_after_pad[i] - pad_len[i - 1],
+                        cu_seqlens_after_pad[i],
+                    )
+                    assert torch.count_nonzero(tensor[pad_range[0] : pad_range[1]]).item() == 0, (
+                        f"{backend} left nonzero values in {tensor_name} padding for sequence "
+                        f"{i - 1}"
+                    )
+
             out_orig = torch.Tensor([]).to(device="cuda", dtype=dtype)
             if is_training:
                 q_grad_orig = torch.Tensor([]).to(device="cuda", dtype=dtype)
--- a/tests/pytorch/attention/test_attention.py
+++ b/tests/pytorch/attention/test_attention.py
@@ -32,6 +32,7 @@ from transformer_engine.pytorch.attention.dot_product_attention.utils import (
     FlashAttentionUtils,
     _get_supported_versions,
     check_set_window_size,
+    get_thd_padding_mask,
 )
 from transformer_engine.pytorch.attention import RotaryPositionEmbedding
 import transformer_engine.pytorch.cpp_extensions as ext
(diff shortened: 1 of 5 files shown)


ioangatop:
hey @xiaoyao0115, great work I was just looking at this! do you think we can go ahead and merge it soon enough? cc @sudhakarsingh27

transformer_engine/pytorch/attention/dot_product_attention/backends.py
```diff
@@ -844,6 +844,33 @@ def _unalias_cu_seqlens(cu_q: torch.Tensor, cu_kv: torch.Tensor):
     return cu_q, cu_kv
 
 
+def _mask_thd_padding(tensor: torch.Tensor, padding_mask: torch.Tensor) -> torch.Tensor:
+    """Return ``tensor`` with invalid THD token rows zeroed."""
```
sudhakarsingh27, line 848: nit: token rows -> token positions

tests/pytorch/attention/test_attention.py
```diff
@@ -129,6 +130,124 @@ def test_flash_attention_supported_version_message():
...
+    """Check padded outputs and gradients, including buffer tails, in eager and raw graphs."""
+    if fa_version == 3 and (
+        device_compute_capability != (9, 0) or not FlashAttentionUtils.v3_is_installed
+    ):
+        pytest.skip("FlashAttention 3 on Hopper is required.")
+    if fa_version == 4 and (
+        not (10, 0) <= device_compute_capability < (12, 0)
+        or not FlashAttentionUtils.v4_is_installed
+    ):
+        pytest.skip("FlashAttention 4 on SM100/SM110 is required.")
```
sudhakarsingh27, line 163: these checks should already be handled by `utils.get_attention_backend` so probably not needed here

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -1706,11 +1706,15 @@ def _is_fa3_supported(num_heads, num_gqa_groups, head_dim_qk, head_dim_v, qkv_dt
...
 @torch.no_grad()
 def get_thd_padding_mask(num_tokens, cu_seqlens, cu_seqlens_padded):
-    """Identify inter-sequence padding in a flattened packed THD buffer."""
+    """Identify inter-sequence and tail padding in a flattened packed THD buffer."""
     rows = torch.arange(num_tokens, device=cu_seqlens_padded.device)
     sequence = torch.searchsorted(cu_seqlens_padded[1:], rows, right=True)
-    valid_end = cu_seqlens_padded[sequence] + cu_seqlens[sequence + 1] - cu_seqlens[sequence]
-    return rows >= valid_end
+    valid_ends = cu_seqlens_padded[:-1] + cu_seqlens[1:] - cu_seqlens[:-1]
+    # Tail rows map to sequence == batch_size. Append a zero-length sequence
```
sudhakarsingh27, line 1713: nit: Tail rows -> Tail token positions

sudhakarsingh27 (commented):
Thanks for this PR! It's very helpful

A few comments; 
1. Is the CUDA graph captureable requirement only for non-CP code? 
2. `get_thd_padding_mask` can accept input tensors that are longer than what cu_seqlens(_padded) show. It should be made more explicit

transformer_engine/pytorch/attention/dot_product_attention/backends.py
xiaoyao0115: thanks, i will make the changes

tests/pytorch/attention/test_attention.py
xiaoyao0115: okay

transformer_engine/pytorch/attention/dot_product_attention/utils.py
xiaoyao0115: okay~

xiaoyao0115:
> Thanks for this PR! It's very helpful
> 
> A few comments;
> 
> 1. Is the CUDA graph captureable requirement only for non-CP code?
> 2. `get_thd_padding_mask` can accept input tensors that are longer than what cu_seqlens(_padded) show. It should be made more explicit

Thanks for the review!
1. No—the padding cleanup should remain CUDA-graph-capturable for supported CP paths as well. The regression test added here exercises raw capture only for non-CP. Separately, I validated FA4 with pad_between_seqs=True on GB200 using TE’s make_graphed_callables for CP1, CP2 P2P, and CP2 A2A on the current commit. Forward/backward checks passed, including exact-zero output and Q/K/V-gradient padding. This PR does not extend raw-capture support for CP. I’ll clarify the distinction between the in-tree test coverage and the separate CP validation.
2. Agreed. `num_tokens` represents the full physical buffer capacity and may exceed cu_seqlens_padded[-1]. All positions beyond that final boundary are treated as padding.

PR #3210, opened 2026-07-14
Fix Cuda Graph based MOE Tests Hang in CI

vthumbe1503:
# Description

make_graphed_callables function returns the module replaced with the graphed module. Recent pytorch upgrade caused the graph replay to hang due to insufficient cleanup.

Irresspective, we were comparing the forward and backward results of the same graphed module being run twice, instead of actually running the non graphed and graphed module and comparing their results numerically.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/test_grouped_linear.py
+++ b/tests/pytorch/test_grouped_linear.py
@@ -1857,6 +1857,15 @@ def test_grouped_linear_fused_path_cuda_graph_safe(fp8_recipe, bias, monkeypatch
         params_dtype=dtype,
         device=device,
     )
+    reference_grouped_linear = GroupedLinear(
+        num_gemms,
+        in_features,
+        out_features,
+        bias=bias,
+        params_dtype=dtype,
+        device=device,
+    )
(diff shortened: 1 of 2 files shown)


vthumbe1503:
Pipeline: 58029914

timmoon10 (approved):
LGTM. Can you elaborate more on how "insufficient cleanup" was causing the hang?

Good catch that we weren't running the non-graphed implementation. It is awkward that `make_graphed_callables` changes the module forward so it always replays the graph, but this is also the behavior in [`torch.cuda.make_graphed_callables`](https://github.com/pytorch/pytorch/blob/d32b0a978a4725600d9eab99dcf824566b9103d9/torch/cuda/graphs.py#L1058).

vthumbe1503:
@timmoon10 I dug a bit deeper and actually it turns out the hang was due to CUDA graph replay compatibilty with CUBLAS GGEMM which exists currently. And it specifically happens on the second graph replay which was happening unintentionally in the test

PR #3212, opened 2026-07-15
Fix FusedAdam empty tensor handling

### LINKED ISSUE #3207: [PyTorch] FusedAdam silently skips updates when a parameter group ends with an empty tensor
### Describe the bug

`transformer_engine.pytorch.optimizers.FusedAdam` silently skips updates to
nonempty parameters when a parameter group ends with a zero-element parameter.
The optimizer step returns without an error, but the preceding nonempty parameter
remains unchanged. The equivalent `torch.optim.Adam` input updates correctly.

This matters for Megatron FSDP because it uses uneven parameter sharding. A valid
global parameter can therefore have a zero-element local shard on some
data-parallel ranks, making empty optimizer parameters common rather than an
invalid edge case. The resulting failure is rank-dependent and silently leaves
some shards stale.

Applications can work around this by detecting locally empty parameters and
excluding their gradients before every optimizer step. However, that adds
optimizer-specific filtering to every FSDP integration and complicates parameter
group, optimizer-state, and checkpoint bookkeeping. FusedAdam should accept a
valid empty local shard and treat its update as a no-op.

This appears related to #3202, which describes the underlying zero-numel handling
in `multi_tensor_apply`. This report captures a smaller two-parameter manifestation
that silently skips an otherwise valid update.

### Steps/Code to reproduce the bug

Run the following on a single GPU; no distributed setup is required:

```python
"""Reproduce TE FusedAdam skipping a parameter group ending in an empty tensor."""

import torch
import transformer_engine
from transformer_engine.pytorch.optimizers import FusedAdam


def run_optimizer(optimizer_type):
    """Run one Adam step over a nonempty parameter followed by an empty parameter."""
    parameters = [
        torch.nn.Parameter(torch.ones(4, device="cuda")),
        torch.nn.Parameter(torch.empty(0, device="cuda")),
    ]
    for parameter in parameters:
        parameter.grad = torch.ones_like(parameter)

    before = parameters[0].detach().clone()
    optimizer_type(parameters).step()
    return (parameters[0] - before).abs().max().item()


print(f"PyTorch: {torch.__version__}")
print(f"Transformer Engine: {transformer_engine.__version__}")
print(f"CUDA: {torch.version.cuda}")

te_change = run_optimizer(lambda params: FusedAdam(params, lr=1e-3))
torch_change = run_optimizer(lambda params: torch.optim.Adam(params, lr=1e-3))

print(f"TE FusedAdam max parameter change: {te_change}")
print(f"PyTorch Adam max parameter change: {torch_change}")

assert torch_change > 0, "The PyTorch Adam control did not update its nonempty parameter"
assert te_change > 0, (
    "TE FusedAdam silently skipped the nonempty parameter because the final parameter "
    "in the group was empty"
)
```

Observed output:

```text
PyTorch: 2.12.0a0+0291f960b6.nv26.04.48445190
Transformer Engine: 2.16.0+4220403e
CUDA: 13.2
TE FusedAdam max parameter change: 0.0
PyTorch Adam max parameter change: 0.0009999871253967285
AssertionError: TE FusedAdam silently skipped the nonempty parameter because the final parameter in the group was empty
```

The final assertion is expected to fail and demonstrates the bug.

### Expected behavior

FusedAdam should update every nonempty parameter in the group and ignore the empty
parameter as a no-op, matching `torch.optim.Adam`. It should not silently skip the
entire pending update because the final tensor is empty.

### Environment overview

- Environment location: Docker, NVIDIA NGC PyTorch 26.04 development container
- Transformer Engine: preinstalled in the container

### Environment details

- OS: Ubuntu 24.04.4 LTS
- Python: 3.12.3
- PyTorch: 2.12.0a0+0291f960b6.nv26.04.48445190
- Transformer Engine: 2.16.0+4220403e
- CUDA: 13.2
- cuDNN: 9.21

### Device details

- GPU: NVIDIA RTX A6000
- Driver: 570.211.01

wujingyue:
## Summary
- Filter zero-sized optimizer tensor slots through a shared helper before Adam/SGD call `multi_tensor_apply`
- Add a native `multi_tensor_apply` check so zero-chunk tensors fail with a clear error if they reach the launcher
- Add regression tests for an empty parameter at the end of Adam and SGD parameter groups

## Root cause
`multi_tensor_apply` drives launch bookkeeping from the per-tensor chunk loop. A zero-sized tensor has zero chunks, so if an empty tensor appears at the end of a group, pending work for earlier tensors can remain unlaunched and FusedAdam silently skips the update.

The generic `MultiTensorApply` wrapper still preserves caller-visible slot semantics. Adam and SGD opt into filtering because empty tensor slots are no-op optimizer updates.

Fixes NVIDIA/TransformerEngine#3207

## Validation
- `NVTE_FRAMEWORK=pytorch NVTE_CUDA_ARCHS=86 NVTE_WITH_NCCL_EP=0 NVTE_SKIP_SUBMODULE_CHECKS_DURING_BUILD=1 MAX_JOBS=4 NVTE_BUILD_THREADS_PER_JOB=1 python setup.py build_ext --inplace`
- `python -m py_compile transformer_engine/pytorch/optimizers/fused_adam.py transformer_engine/pytorch/optimizers/fused_sgd.py transformer_engine/pytorch/optimizers/multi_tensor_apply.py`
- `PYTHONPATH=/tmp/TransformerEngine pytest -q tests/pytorch/test_fused_optimizer.py::TestFusedAdam::test_empty_param_at_end_of_group tests/pytorch/test_fused_optimizer.py::TestFusedSGD::test_empty_param_at_end_of_group`
- All-empty sanity for `te.optimizers.FusedAdam` and `te.optimizers.FusedSGD`: both completed `step()` and `torch.cuda.synchronize()`
- Issue reproducer: `te 0.0009999871253967285`, `torch 0.0009999871253967285`
- `git diff --check`

### DIFF
--- a/transformer_engine/pytorch/optimizers/multi_tensor_apply.py
+++ b/transformer_engine/pytorch/optimizers/multi_tensor_apply.py
@@ -6,6 +6,18 @@
 from torch.distributed._tensor import DTensor
 
 
+def filter_empty_tensor_lists(tensor_lists):
+    """Remove aligned zero-sized tensor slots and return whether any slots remain."""
+    if any(len(tensors) != len(tensor_lists[0]) for tensors in tensor_lists):
+        raise RuntimeError("Expected aligned multi-tensor lists.")
+
+    keep_slot = [tensor.numel() > 0 for tensor in tensor_lists[0]]
+    for i, tensors in enumerate(tensor_lists):
+        tensor_lists[i] = [tensor for tensor, keep in zip(tensors, keep_slot) if keep]
+
+    return bool(tensor_lists[0])
+
+
 class MultiTensorApply:  # pylint: disable=too-few-public-methods
     """Multi-tensor apply entry."""
 
--- a/transformer_engine/pytorch/optimizers/multi_tensor_apply.py
+++ b/transformer_engine/pytorch/optimizers/multi_tensor_apply.py
@@ -6,18 +6,6 @@
 from torch.distributed._tensor import DTensor
 
 
-def filter_empty_tensor_lists(tensor_lists):
-    """Remove aligned zero-sized tensor slots and return whether any slots remain."""
-    if any(len(tensors) != len(tensor_lists[0]) for tensors in tensor_lists):
-        raise RuntimeError("Expected aligned multi-tensor lists.")
-
-    keep_slot = [tensor.numel() > 0 for tensor in tensor_lists[0]]
-    for i, tensors in enumerate(tensor_lists):
-        tensor_lists[i] = [tensor for tensor, keep in zip(tensors, keep_slot) if keep]
-
-    return bool(tensor_lists[0])
-
-
 class MultiTensorApply:  # pylint: disable=too-few-public-methods
     """Multi-tensor apply entry."""
 
@@ -29,6 +17,16 @@ class MultiTensorApply:  # pylint: disable=too-few-public-methods
             for j, t in enumerate(ts):
                 if isinstance(t, DTensor):
                     tensor_lists[i][j] = t._local_tensor
+
+        if any(len(tensors) != len(tensor_lists[0]) for tensors in tensor_lists):
+            raise RuntimeError("Expected aligned multi-tensor lists.")
+
+        keep_slot = [tensor.numel() > 0 for tensor in tensor_lists[0]]
+        for i, tensors in enumerate(tensor_lists):
+            tensor_lists[i] = [tensor for tensor, keep in zip(tensors, keep_slot) if keep]
+
+        if not tensor_lists[0]:
+            return None
 
         return op(self.chunk_size, noop_flag_buffer, tensor_lists, *args)
 
--- a/tests/pytorch/test_fused_optimizer.py
+++ b/tests/pytorch/test_fused_optimizer.py
@@ -166,6 +166,19 @@ class TestFusedAdam(TestFusedOptimizer):
 
             torch.testing.assert_close(ref_param, tst_param)
 
+    def test_empty_param_at_end_of_group(self):
+        tensors = [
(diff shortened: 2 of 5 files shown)


transformer_engine/pytorch/optimizers/multi_tensor_apply.py
```diff
@@ -6,6 +6,18 @@
 from torch.distributed._tensor import DTensor
 
 
+def filter_empty_tensor_lists(tensor_lists):
```
wujingyue: Should we call this from MultiTensorApply unconditionally? I'm not sure whether filtering is safe for every optimizer.

vthumbe1503: Yes, agreed, I dont think there is any harm applying this filter directly in the MultiTensorApply class instead of putting it in each optimizer. Could we make that change @wujingyue ?

vthumbe1503 (dismissed):
LGTM. Just one minor review comment

transformer_engine/pytorch/optimizers/multi_tensor_apply.py
vthumbe1503: Non blocking comment, given that we have just have optimizers exposed in TE

wujingyue: Done

wujingyue:
FYI, CI has an unrelated build error:

```
  nccl_ep.cc:23:10: fatal error: nccl_device.h: No such file or directory
     23 | #include <nccl_device.h>
        |          ^~~~~~~~~~~~~~~
  compilation terminated.
```

vthumbe1503 (approved):

vthumbe1503:
Pipeline: 58726828. Jax errors are pre-existing on main

PR #3213, opened 2026-07-15
Fix duplicate FP8 state updates during activation checkpoint recomputation

AlbertYang514:
## Summary

This change fixes FP8 first-module ownership across activation-checkpoint forward and recompute. The checkpoint frame reserves ownership for the original forward and reuses the captured state during recompute, preventing duplicate FP8 update ownership while preserving the expected single update.

## v2 edge-case handling

The revised candidate also covers three state-transition edges:

- nested autocast exit cannot revive first-module ownership that was consumed by the inner region;
- an exception in the original checkpointed forward restores ownership and invalidates the incomplete frame;
- recompute entry and body failures roll state back transactionally, allowing later checkpoint use in the same outer autocast.

Ownership merging is monotonic (available may become consumed, but consumed is not restored to available from an older snapshot). The nested-autocast path explicitly reacquires quantization state after restoring the outer autocast state rather than relying on object identity.

## Validation

- Existing PR #3213 tests: 8/8 passed.
- Focused nested-autocast and exception-recovery tests: 6/6 passed.
- Full local checkpoint audit suite: 49/49 passed.
- CPU state-transition probes passed for all added edge cases.

## Scope

This PR is limited to checkpoint/autocast state ownership and rollback. A residual SM120 RTC cast-transpose failure can still be reproduced with correct 1/1 ownership/update counts; its investigation and workaround belong to PR #3215 and are not part of this change.

### DIFF
--- a/tests/pytorch/test_fp8_checkpoint_state_edges.py
+++ b/tests/pytorch/test_fp8_checkpoint_state_edges.py
@@ -1,0 +1,222 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""Targeted FP8 checkpoint ownership and exception-state regressions."""
+
+from dataclasses import dataclass
+
+import pytest
+import torch
+
+import transformer_engine.pytorch as te
+from transformer_engine.common.recipe import DelayedScaling, Format
+from transformer_engine.pytorch.distributed import (
+    _ActivationRecomputeState,
+    activation_recompute_forward,
+    in_fp8_activation_recompute_phase,
+    is_fp8_activation_recompute_enabled,
+)
+from transformer_engine.pytorch.quantization import FP8GlobalStateManager
+
+
+fp8_available, reason_for_no_fp8 = te.is_fp8_available(return_reason=True)
+pytestmark = pytest.mark.skipif(not fp8_available, reason=reason_for_no_fp8)
+
+WIDTH = 16
+SEED = 20260722
+
+
+@dataclass
+class UpdateCounter:
+    forward: int = 0
+    backward: int = 0
+
+    def snapshot(self) -> tuple[int, int]:
+        return self.forward, self.backward
+
+    def delta(self, before: tuple[int, int]) -> tuple[int, int]:
+        return self.forward - before[0], self.backward - before[1]
+
+
+@pytest.fixture(autouse=True)
+def clean_fp8_state():
+    FP8GlobalStateManager.reset()
+    yield
+    qstate = FP8GlobalStateManager.quantization_state
+    assert qstate.autocast_depth == 0
+    assert not FP8GlobalStateManager.is_fp8_enabled()
+    assert not is_fp8_activation_recompute_enabled()
+    assert not in_fp8_activation_recompute_phase()
+
+
+@pytest.fixture
+def update_counter(monkeypatch) -> UpdateCounter:
+    counter = UpdateCounter()
+    original = FP8GlobalStateManager.reduce_and_update_fp8_tensors
+
+    def counted(_cls, forward=True):
+        if forward:
+            counter.forward += 1
+        else:
+            counter.backward += 1
+        return original(forward=forward)
+
+    monkeypatch.setattr(
+        FP8GlobalStateManager,
+        "reduce_and_update_fp8_tensors",
+        classmethod(counted),
+    )
+    return counter
+
+
+def make_linear() -> te.Linear:
+    return te.Linear(
+        WIDTH,
+        WIDTH,
+        bias=False,
+        params_dtype=torch.float32,
+        init_method=lambda tensor: torch.nn.init.normal_(tensor, mean=0.0, std=0.1),
+    ).cuda()
+
+
+def make_input() -> torch.Tensor:
+    return torch.randn(
+        WIDTH,
+        WIDTH,
+        dtype=torch.bfloat16,
+        device="cuda",
+        requires_grad=True,
+    )
+
+
+def assert_finite_nonzero(loss, inp, module) -> None:
+    assert torch.isfinite(loss)
+    assert inp.grad is not None
+    assert torch.isfinite(inp.grad).all()
+    assert torch.count_nonzero(inp.grad) > 0
+    for parameter in module.parameters():
+        assert parameter.grad is not None
+        assert torch.isfinite(parameter.grad).all()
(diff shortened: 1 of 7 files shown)


AlbertYang514:
Hi @ksivaman and @ptrendx gentle ping when either of you has a chance.

This PR has been open for about a week and is currently mergeable. It fixes duplicate FP8 global state updates under activation checkpoint recomputation, with regression coverage for reentrant, non-reentrant, per-layer, and nested checkpoint paths.

Could either of you please take a look, or point me to the current owner of the activation-recompute / FP8 state-management path? I’m happy to adjust or split the change if preferred.

AlbertYang514:
Hi @vthumbe1503 — I revised #3213 to cover nested-autocast ownership restoration and transactional recovery from original-forward, recompute-entry, and recompute-body exceptions.

The updated candidate passes the existing 8 tests, 6 focused edge tests, and all 49 cases in the local checkpoint audit suite. The residual SM120 RTC cast-transpose failure is still reproducible with correct 1/1 ownership/update counts and is being treated separately in #3215.

When convenient, could you please take another look at the revised ownership and rollback logic?

pggPL:
@AlbertYang514 sorry for delayed response. 

I started reviewing this PR and I think this issue may be solved together with other issue, I need one or few days for investigation and will reach out to you soon.

AlbertYang514:
@pggPL 
Thanks for taking a look. I tested the current NVIDIA/main (172bd93773ad6ee4ba44b460b7f10ef42fc89d57), which already contains #3284, using black-box regressions extracted from this PR without applying the #3213 production fix.
The original duplicate FP8 update ownership issue is still reproducible. The expected update count is forward=1, backward=1, but I see:
no checkpoint             fwd=1 bwd=1
non-reentrant/single      fwd=2 bwd=1
non-reentrant/per-layer   fwd=4 bwd=1
reentrant/single/1 layer  fwd=2 bwd=1
reentrant/single/3 layers fwd=2 bwd=3
reentrant/per-layer       fwd=4 bwd=3
nested reentrant          fwd=3 bwd=2
Losses and gradients remain finite; the failures are specifically duplicate ownership/update bookkeeping. I also still reproduce the nested-autocast ownership and exception/recompute rollback issues covered here.
For comparison, the #3284 inner-autocast/recompute metadata regression coverage passes on the same current main (6 passed), so #3284 appears to address the metadata stash/restore/recompute-region issue rather than the first-module update ownership issue here.
I'm also running an additional cross-check now: applying the #3213 production fix to the pre-#3284 main state and running #3284's regression coverage against it. I'll post the result once that check is complete.

AlbertYang514:
@pggPL
I finished the cross-check against #3284 as well.
The result indicates that the two fixes address distinct regressions:
Current main with #3284: #3284 regression tests pass (6/6), while the #3213 ownership regressions still fail (6/8 fail, with duplicate forward/backward updates).
Pre-#3284 main with only the #3213 production patch applied: all six #3284 regression tests fail, primarily with unbalanced recompute metadata stash/restore and KeyError: 'global_fp8_buffer_pos_fwd_recompute'.
So #3284 fixes the inner-autocast recompute-region / metadata bookkeeping issue, while #3213 fixes the first-module ownership / duplicate update and rollback issue. Neither fix subsumes the other.
For the reverse test I used pre-#3284 SHA 8260f49660cbadb78bc52c90449428c51625469d, applied only the #3213 production diff (396948571937518bd8f2d99044aba80070bec09a), and did not apply #3284 production changes.

AlbertYang514:
@pggPL 
I’ve updated #3213 onto current main with the post-#3284-compatible implementation.
The PR is now mergeable and the focused validation is green:
#3213 ownership regressions: 8 passed
#3213 edge/rollback tests:   6 passed
#3284 regressions:           6 passed
nearby checkpoint tests:     passed
All tested checkpoint configurations now perform exactly one forward and one backward FP8 update, while preserving #3284’s inner-autocast recompute-region behavior.
New head: aa415dbc868fa95230d0cc67fc1d1122c71f137a

pggPL:
Hi, I didn't mean #3284, this is indeed separate issue. 

We have hidden assumption that the FP8 layers are sequential and the backward is the reverse order of forward. This may not be true for some non-mainstream architectures I guess, but for most cases it is not the issue. I tried to substitute logic with `is_first_fp8_module` with using torch backward callback - this would make solving issue like this easier. But it has some problems also.


BTW I read your code with codex and my agent says
```
The four concrete P1 issues are:

  1. Megatron-Core CheckpointWithoutOutput becomes incompatible

     PR #3213 requires phase 1 and phase 2 to share an _ActivationRecomputeState. MCore constructs these contexts independently, so the recompute context receives state=None and raises:

     FP8 recompute state was not captured during a completed original forward

     Impact: a runtime failure in an existing integration.

  2. An inner-only FP8 autocast loses update ownership

     def body(x):
         with te.autocast(enabled=True):
             return layer(x)

     te.checkpoint(body, x, use_reentrant=True)

     FP8 is disabled when the checkpoint wrapper is entered, so the captured owner is False. During recompute, the wrapper opens an autocast with _recompute=True. The autocast inside body is therefore nested and
     does not initialize a new owner.

     Impact: no module backward performs the delayed-scaling update.

  3. A checkpoint reserves ownership without having a module that participates in backward

     When an outer FP8 autocast is active, PR #3213 uses:

     reserve_first_fp8_module=True

     before executing the checkpoint body and sets the global is_first_fp8_module=False. At that point, it does not know whether the checkpoint body contains an FP8 module that will participate in backward.

     If it does not, ownership is lost and a later FP8 module outside the checkpoint cannot claim it.

     Impact: the backward update is skipped even though the later module produced backward amax values.

  4. Monotonic set_autocast_state() breaks exact state restoration

     PR #3213 globally changes:

     set_autocast_state(snapshot)

     from exact restoration to behavior equivalent to:

     restored_owner = current_owner and snapshot_owner

     The LayerNormMLP internal checkpoint temporarily switches to its saved autocast state and then expects to restore the caller’s exact state. With the new monotonic semantics, a current False value cannot be
     restored to True.

     Impact: recompute may run with incorrect ownership, or ownership consumed during recompute may leak into the caller’s context.

  The first issue causes a direct runtime failure. The other three can silently skip or misassign the backward delayed-scaling update, so all four are correctness P1s.
```

pggPL:
I'm still investigating usage of `Variable._execution_engine.queue_callback` and need at least one more day for that

AlbertYang514:
@pggPL 
Thanks for the update! I really appreciate you looking into the queue_callback approach. No rush at all — I’ll wait for your findings.

PR #3215, opened 2026-07-15
Work around intermittent SM120 FP8 gradient corruption in RTC cast-transpose

AlbertYang514:
## Summary

This change retains a device-scope `__threadfence()` in the optimized RTC `cast_transpose` kernel as a validated workaround for intermittent FP8 gradient corruption observed on SM120. The evidence supports describing this as a timing/codegen-sensitive workaround; the exact root cause is still unknown.

## Observed issue

Under the affected checkpointed DelayedScaling workload, the forward loss remains normal, but backward can produce finite, extremely large gradients. The first affected layer and microbatch vary across runs. After the #3213 v2 checkpoint fix, the residual failure remains reproducible while FP8 ownership and update counts stay at the expected 1/1.

## Change

The current code inserts `__threadfence()` on the optimized RTC `cast_transpose` path. This device-scope fence is retained because it has been stable in the tested reproductions. It does not change the public API or intended mathematical operation.

## Investigation status

Nsight confirms that the optimized RTC cast-transpose kernel and the following dgrad GEMM execute on the same CUDA stream. Therefore, the original cross-kernel output-store/amax publication explanation is not supported by the execution model or the follow-up ablations. Seven tested fence placements were stable, including placements after the atomic operation, immediately before return, and under a thread-0-only condition. Replacing the fence with a system-scope atomic operation did not stabilize the no-fence variant, while a block-scope fence was also stable in the tested workload. Together, these results invalidate the earlier narrow publication-ordering explanation and instead point to a timing/codegen-sensitive effect. They do not identify the exact mechanism.

## Validation

- No-fence build: 5/10 failures in the fixed-seed sweep and 10/20 failures in the high-risk sweep; ownership/update counts remained 1/1.
- Device-fence build: 0/10 failures in the fixed-seed sweep and 0/20 failures in the high-risk sweep; ownership/update counts remained 1/1.
- Experimental block-fence build: 0/10 failures in the fixed-seed sweep and 0/20 failures in the high-risk sweep.
- A 100-step numerical comparison produced 1,000 finite records; configurations L/M/N/O were stepwise identical in the non-triggering workload.
- `memcheck`, `racecheck`, `initcheck`, and `synccheck` completed without a reported error; `initcheck` was also rerun without filtering. This is useful negative evidence, not proof that the kernel is free of races or memory-ordering defects.

## Relationship to #3213

The residual failure remains reproducible after #3213 v2 restores FP8 ownership/update counts to 1/1. This separates the observed residual phenomenon from the known duplicate-update pathology, but does not prove that every underlying mechanism is fully independent.

## Scope and limitations

The validation is specific to the tested SM120 hardware, CUDA 13 toolchain, and checkpointed DelayedScaling workload. The block-scope fence is experimental evidence only; this PR retains the device-scope fence. The root cause remains unresolved, and no claim is made for other architectures or workloads. The change does not alter public APIs or intended math.

### DIFF
--- a/tests/pytorch/test_delayed_scaling_dgrad_ordering.py
+++ b/tests/pytorch/test_delayed_scaling_dgrad_ordering.py
@@ -1,0 +1,120 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""Regression coverage for delayed-scaling cast/transpose ordering on SM120."""
+
+import math
+
+import pytest
+import torch
+import torch.nn.functional as F
+
+import transformer_engine.pytorch as te
+from transformer_engine.common.recipe import DelayedScaling, Format
+
+
+class _StressBlock(torch.nn.Module):
+    def __init__(self, width: int, kv_width: int, mlp_width: int) -> None:
+        super().__init__()
+
+        def linear(inp: int, out: int) -> te.Linear:
+            return te.Linear(
+                inp,
+                out,
+                bias=False,
+                params_dtype=torch.bfloat16,
+            )
+
+        self.norm = torch.nn.RMSNorm(width, dtype=torch.bfloat16, device="cuda")
+        self.q_proj = linear(width, width)
+        self.k_proj = linear(width, kv_width)
+        self.v_proj = linear(width, kv_width)
+        self.o_proj = linear(width, width)
+        self.gate_proj = linear(width, mlp_width)
+        self.up_proj = linear(width, mlp_width)
+        self.down_proj = linear(mlp_width, width)
+
+    def forward(self, value: torch.Tensor) -> torch.Tensor:
+        hidden = self.norm(value)
+        q_value = self.q_proj(hidden)
+        kv_value = torch.cat((self.k_proj(hidden), self.v_proj(hidden)), dim=-1)
+        attention = self.o_proj(torch.tanh(q_value + 0.125 * kv_value))
+        gate = F.silu(self.gate_proj(hidden))
+        mlp = self.down_proj(gate * self.up_proj(hidden))
+        return value + 0.03125 * (attention + mlp)
+
+
+class _StressNetwork(torch.nn.Module):
+    def __init__(self) -> None:
+        super().__init__()
+        self.layers = torch.nn.ModuleList([_StressBlock(2048, 1024, 4096) for _ in range(14)])
+
+    def forward(self, value: torch.Tensor) -> torch.Tensor:
+        for layer in self.layers:
+            value = te.checkpoint(layer, value, use_reentrant=True)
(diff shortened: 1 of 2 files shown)


tests/pytorch/test_delayed_scaling_dgrad_ordering.py
```diff
@@ -0,0 +1,120 @@
...
+            assert finite
+            assert math.isfinite(input_max)
+            assert final_norm < 0.1, (
+                f"gradient corruption at step={step}, microbatch={microbatch}: "
+                f"grad_norm={final_norm}, parameter_max={parameter_max}, "
+                f"input_max={input_max}"
+            )
+            assert input_max < 1e-4
+
+    assert 0.004 < final_norm < 0.006
```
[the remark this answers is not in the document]
AlbertYang514: Thanks — agreed. The final `0.004–0.006` band was narrower than necessary for this regression and could be sensitive to SM120 stepping or toolchain differences.

I widened only the terminal sanity bound to `0.001 < final_norm < 0.05`; the per-microbatch `final_norm < 0.1`, finite loss/gradient, input-gradient, anomaly-threshold, and all other assertions remain unchanged.

All three fixed seeds still pass on SM120, and Black, diff-check, syntax, license, and pre-commit checks pass.

AlbertYang514:
Hi @vthumbe1503 and @timmoon10, gentle ping when either of you has a chance.

This is a narrowly scoped SM120 RTC cast-transpose ordering fix: one device-scope fence before the DelayedScaling amax atomic, plus a targeted regression test.

It is currently mergeable, but has not received a human review yet. Could either of you please take a look, or point me to the current owner of this path?

transformer_engine/common/transpose/rtc/cast_transpose.cu
```diff
@@ -116,6 +116,8 @@ __global__ void __launch_bounds__(block_size) cast_transpose_optimized_kernel(
 
   // Reduce amax over block
   if (amax_ptr != nullptr) {
+    // Order the global output stores before publishing the amax update.
```
vthumbe1503, line 119: I am actually unclear why fp8 output store needs to happen before amax compute. Isnt the dgrad compute stream ordered anyway? Why would dgrad recieve incorrect fp8 values and threadfence helps it?

AlbertYang514: @vthumbe1503 — Your same-stream objection is correct, and the additional ablations do not support our original output-store/amax publication explanation. Seven fence placements were stable—including post-atomic, pre-return, and thread-0-only placements—while a system-scope atomic without the fence still failed, so I have rewritten the change as a timing/codegen-sensitive workaround. The current candidate retains the validated device-scope fence; the block-scope fence is documented only as experimental evidence. The exact root cause remains unknown, and the revised description now states that limitation explicitly.

PR #3224, opened 2026-07-20
Improve device-init grouped linear module with single grouped weight support 

zhongbozhu:
# Description

Need this Mcore PR to make it work: https://github.com/NVIDIA/Megatron-LM/pull/6000

Fixes numerical issues when using single weight for TE module grouped linear. Limit the single weight feature to the grouped tensor API instead of the legacy path.

E2E validation has been done and posted to the Megatron PR.

Performance:

Benchmarked using Qwen3.5 397B problem shape, tested the grouped linear FC1 FC2 layer time, including quantization kernels plus forward backward gemms. 

image

Note: needs to pay extra attention to whether bias grad and weight grad are generated properly. 

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1315,6 +1497,13 @@ class GroupedLinear(TransformerEngineBaseModule):
                        EXPERIMENTAL and subject to change. Gated by the
                        ``NVTE_GROUPED_LINEAR_SINGLE_PARAM`` environment variable: if the env var
                        is not set this argument is forced to ``False`` with a warning.
+    grouped_gemm_backend : {"legacy", "grouped_tensor"}, default = None
```
timmoon10: It's unintuitive that `grouped_gemm_backend="grouped_tensor"` might end up using split-quantize internally. What do you think about the following:

- `"split_tensors"`: Use split-quantize impl
- `"grouped_tensor"`: User promises split points are valid for grouped GEMM kernel. Use grouped tensor impl, or error out if not supported.
- `"prefer_grouped_tensor"`: User promises split points are valid for grouped GEMM kernel. Use grouped tensor impl, or fallback to split-quantize impl if not supported.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -86,6 +86,7 @@ def _maybe_dequantize(
     @staticmethod
     def _is_grouped_tensor_path_supported(
         *,
+        grouped_gemm_backend: str,
```
CarlosGomes98: If we accept raw str, lets make sure to document the allowed options in the docstring below

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1846,16 +2084,15 @@ def forward(
         is_grad_enabled = torch.is_grad_enabled()
         num_gemms = self.num_gemms
 
-        if FP8GlobalStateManager.fp8_graph_capturing():
```
zhongbozhu, line 2313: Note: this code block was deleted because it was duplicated

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -235,14 +236,108 @@ def _prepare_weights_for_grouped_tensor_gemm(
...
+                    saved_internal = weight_quantizer.internal
+                    weight_quantizer.internal = False
+                grouped_weight = tex.group_quantize(
+                    source,
+                    weight_quantizer,
+                    num_gemms,
+                    None,
+                )
+                if cache_weight:
+                    weight_quantizer.internal = saved_internal
```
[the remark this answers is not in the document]
zhongbozhu: no need

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -600,7 +744,24 @@ def forward(
...
+                "feature configuration does not support the grouped_tensor backend. Legacy "
+                "fallback is unavailable for single grouped parameters; use discrete parameters "
+                "or a supported grouped_tensor configuration."
+            )
+        if use_grouped_tensor_path:
+            if m_splits.device.type != "cuda":
+                raise ValueError(
+                    "The native grouped_tensor path requires CUDA m_splits. Pass a CUDA int64 "
+                    "tensor, or select grouped_gemm_backend='legacy'."
+                )
```
timmoon10, line 1149: I get that the h2d memcpy is suboptimal, but it's trivially easy to handle. Erroring out seems excessively rigid.
```suggestion
                m_splits = m_splits.to(device=device)
```
We need to handle the d2h case anyways when the user has specified `grouped_gemm_backend="grouped_tensor"`, but it's not supported and we fallback to split-quantize.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1379,6 +1580,13 @@ class GroupedLinear(TransformerEngineBaseModule):
                        EXPERIMENTAL and subject to change. Gated by the
                        ``NVTE_GROUPED_LINEAR_SINGLE_PARAM`` environment variable: if the env var
                        is not set this argument is forced to ``False`` with a warning.
+    grouped_gemm_backend : {"legacy", "grouped_tensor"}, default = None
```
timmoon10: These are misleading names.
- Legacy implies we are deprecating it, but in fact it is the default backend and it is the fallback when the grouped tensor impl is not supported.
- As mentioned in https://github.com/NVIDIA/TransformerEngine/pull/3224/changes#r3618050729, it's confusing that we might not actually perform the backend specified by the user. The user might put `grouped_gemm_backend="grouped_tensor"` and we'll actually do split-quantize because the cuBLAS version is old or something.

Really, this is not a backend config. It's a hint where the user promises to provide `m_splits` with a certain alignment. I'd suggest changing this to a bool like `enable_grouped_tensor_backend`. Alternatively if we want to generalize in case we add more backends in the future, we could have a string like `enabled_backends` (`enabled_backends="grouped_tensor,future_backend"`).

timmoon10 (commented):
The biggest change in this PR is that TE is abandoning any attempt to make `single_grouped_weight=True` a general feature. Things must be _exactly_ right, or we crash. Given how delicate and experimental this feature has been, I'm not opposed.

The second change is that users must opt-in to access the grouped GEMM kernel. This is also reasonable, since it has alignment requirements for `m_splits` and it's helpful having a way for users to accept that stricter contract.

We are experiencing many test failures. Given that `single_grouped_weight` is no longer a general feature, I think it's reasonable we move the corresponding tests to `test_grouped_linear.py` and `test_grouped_mlp.py`.

transformer_engine/pytorch/module/grouped_linear.py
zhongbozhu: But if the alignment is not provided in the first place, converting it to a device tensor also wouldn't work right, I am okay with another alignment check before adding this H2D copy.

transformer_engine/pytorch/module/grouped_linear.py
zhongbozhu: > and we'll actually do split-quantize because the cuBLAS version is old or something

Should we just raise error instead of silently doing this fallback? As a user, it's actually annoying to specify some optimization but it's not working because a fallback get triggered. 

I am okay with changing the TE interface into a bool flag like `enable_grouped_tensor_backend` instead of a str.

zhongbozhu: renamed

tests/pytorch/test_grouped_linear.py
```diff
@@ -1504,18 +1508,538 @@ def test_fp8_grouped_gemm(shape, accumulate):
...
+):
+    """Match CUDA m_splits and all meaningful parameter layouts against the legacy path."""
+    if not is_module_grouped_tensor_path_supported(
+        fp8_recipe,
+        torch.bfloat16,
+        single_grouped_weight=single_grouped_weight,
+    ):
+        pytest.skip("Recipe is not supported by the module GroupedTensor path on this system.")
+    if fuse_wgrad_accumulation and fp8_recipe is not None and fp8_recipe.float8_block_scaling():
+        # TODO: Re-enable after resolving cuBLASLt grouped FP8 block-scaling support for
```
zhongbozhu: TODO: I don't quite understand why this is the case, but without it, cublas will complain cannot find supported gemm. Unclear if it's related with TE integration or a cublas issue or if upgrading to the next cublas version can fix it. 
cc @denera @timmoon10

zhongbozhu: removing this and launch CI again to see if we should just modify cublas version guard

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -572,9 +572,18 @@ __global__ void __launch_bounds__(THREADS_PER_CHUNK) group_quantize_mxfp8_kernel
     const size_t scale_stride_colwise = DIVUP_TO_MULTIPLE(cols, scale_alignment_X_colwise);
 
     const size_t tensor_base = current_block.tensor_base;
-    const size_t tensor_base_for_scales = (is_single_tensor && num_tensors > 1)
-                                              ? static_cast<size_t>(offsets_ptr[tensor_id])
-                                              : tensor_base;
+    size_t tensor_base_for_scales = tensor_base;
+    size_t tensor_rows_for_scales = rows;
+    if constexpr (WITH_GEMM_SWIZZLED_SCALES && SHAPE_REP == ShapeRepresentation::SAME_BOTH_DIMS) {
```
zhongbozhu, line 577: Note: this is for single weight quantize for mxfp8. 

Before this change, the weight quantizer didn't have the `first_dims` because moe weights are uniform shape for both dimension. This will then trigger a CUDA illegal access because `offsets_ptr=nullptr`

transformer_engine/pytorch/csrc/type_converters.cpp
```diff
@@ -219,12 +219,22 @@ GroupedTensorWrapper GroupedTensorFromPyTorchGroupedTensor(py::handle tensor) {
   }
   auto ret = GroupedTensorWrapper(num_tensors, logical_shape, scaling_mode);
 
+  auto get_initialized_storage_shape = [&logical_shape](const at::Tensor &data) {
+    auto shape = getTensorShape(data);
+    if (data.numel() == 0) {
+      // PyTorch may use a null pointer for a valid zero-sized allocation. Keep an explicitly
+      // provided empty storage distinguishable from the uninitialized {nullptr, {0}} sentinel.
+      shape = logical_shape;
```
vthumbe1503: Aligned offline. I would rather prefer something like {0,0} here instead of logical_shape to make sure TE Common infrastructure understands the rowwise data to be empty tensor instead of None(not being provided). Since logical_shape would raise consistency issues. For columnwise_data logical_shape would infact not signify the right shape

zhongbozhu: fixed

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -1354,24 +1352,25 @@ __global__ void setup_grouped_gemm_kernel(
...
-                                   : (a_base + (a_offset * a_bits_per_elem) / 8);
-  B_ptrs[idx] = b_base + (b_offset * b_bits_per_elem) / 8;
-  C_ptrs[idx] =
-      has_c_multi_tensor ? c_multi_tensor_args.data_ptrs[idx] : (c_base + c_offset * c_elem_size);
-  D_ptrs[idx] =
-      has_d_multi_tensor ? d_multi_tensor_args.data_ptrs[idx] : (d_base + d_offset * d_elem_size);
+  A_ptrs[idx] = a_is_discrete
+                    ? a_multi_tensor_args.data_ptrs[idx]
+                    : (a_base == nullptr ? nullptr : a_base + (a_offset * a_bits_per_elem) / 8);
+  B_ptrs[idx] = b_base == nullptr ? nullptr : b_base + (b_offset * b_bits_per_elem) / 8;
```
vthumbe1503, line 1364: Do we need this?
```suggestion
  B_ptrs[idx] = b_base + (b_offset * b_bits_per_elem) / 8;
```

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -2447,31 +2688,31 @@ def backward_dw(self):
...
-                        if has_grad_biases[i] and bias_params[i].grad is None:
-                            bias_params[i].grad = grad_biases_[i].to(bias_params[i].dtype)
+                if self.use_grouped_tensor:
+                    raise RuntimeError(
+                        "Delayed wgrad unexpectedly produced bias gradients; the grouped-tensor "
+                        "path computes them during the main backward."
+                    )
+                bias_params = [getattr(self, f"bias{i}") for i in range(self.num_gemms)]
+                for i in range(self.num_gemms):
+                    if has_grad_biases[i] and bias_params[i].grad is None:
```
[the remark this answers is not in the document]
zhongbozhu: I do intend to raise error if advanced perf features are used but we are actually in a fallback path. Will improve the error messaging here.

tests/pytorch/test_grouped_linear.py
```diff
@@ -1208,12 +1292,18 @@ def _make_zero_tokens_grouped_tensor(logical_last_dim, is_a):
 
     if layout in ("TN", "NN"):
         weight_tensors = [torch.randn(n, k, dtype=dtype, device=device) for _ in range(z)]
-        if use_mxfp8:
-            grouped_A = _make_grouped_tensor_quantized_mxfp8(
-                weight_tensors,
+        if test_recipe is not None:
+            grouped_weight = torch.cat(weight_tensors, dim=0)
+            weight_quantizer = _make_quantizer(
+                tex.DType.kFloat8E4M3,
```
vthumbe1503: Can we use TE_DType instead of tex.DType here. That is the recommendation now for the public user facing APIs.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -604,14 +578,109 @@ def _prepare_weights_for_grouped_tensor_gemm(
...
+                    output=workspace,
+                )
+            elif update_workspace:
+                grouped_weight = tex.group_quantize(
+                    source,
+                    weight_quantizer,
+                    num_gemms,
+                    None,
+                    output=workspace,
+                )
```
vthumbe1503: Can we combine the conditions to one?

```suggestion
            elif skip_fp8_weight_update is not None or update_workspace:
                grouped_weight = tex.group_quantize(
                    source,
                    weight_quantizer,
                    num_gemms,
                    None,
                    noop_flag=skip_fp8_weight_update,
                    output=workspace,
                )
```

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -1293,11 +1479,16 @@ def _backward_grouped_tensor(
                 dtype=ctx.activation_dtype,
             )
 
-        grad_biases = [None] * N
         if ctx.use_bias:
             if dbias_packed is None:
                 dbias_packed = compute_grouped_dbias(dy_2d, base_split_offsets, N)
-            grad_biases = [dbias_packed[i].to(dtype=ctx.activation_dtype) for i in range(N)]
+            if ctx.single_grouped_bias:
```
vthumbe1503, line 1477: I guess we should eventually unified single_grouped_weight and single_grouped_bias into one parameter.

transformer_engine/pytorch/module/grouped_linear.py
```diff
@@ -446,7 +445,73 @@ def _split_quantize_and_bias(
     return outputs, grad_biases
 
 
-__all__ = ["GroupedLinear"]
+__all__ = ["GroupedLinear", "is_module_grouped_tensor_path_supported"]
+
+
+def is_module_grouped_tensor_path_supported(
```
vthumbe1503, line 451: Can we put this in pytorch/utils.py and even reuse for the ops. The function should be identical right?

Given that we eventually want to unify the module/ops code, would be better to not duplicate these pieces now.

zhongbozhu: Thanks for review, I made a decision to specifically not to do it because they are currently different implementations and I agree with @timmoon10 's proposal to de-duplicate it. But for now, the right thing to do is to keep them separate.

transformer_engine/pytorch/module/grouped_linear.py
zhongbozhu: sigh yes, but currently it's not and technically they can be totally orthogonal to each other as well, I have tested all the combinations already so it's fine

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
zhongbozhu: why not, it's nice to have (although the real fix is in `A_ptrs[idx]`)

vthumbe1503: Additional thread divergence. Anyway, first_dims and last_dims should already be in the descriptors and they will be zero and we wont read it from it anyway.

Same applies for a_base == null_ptr and and C and D as well

vthumbe1503: Real issue was we were computing a_is_discrete and c/d_is_discrete incorrectly

zhongbozhu: No harm keeping it here since, the root fix was:

Python-to-C++ conversion: A valid empty grouped storage has numel=0 and may have data_ptr=nullptr. It was converted as {nullptr, shape={0}}, which TE also uses to mean “storage is absent.” The converter now preserves the grouped tensor’s logical shape, so TE knows the empty storage exists and it won't early return in function `select_grouped_operand`.

Grouped GEMM setup: The code treated base_ptr == nullptr as meaning “discrete tensor list.” Therefore, valid all-empty grouped tensors were misclassified and read uninitialized discrete metadata. It now receives explicit a_is_discrete, c_is_discrete, and d_is_discrete flags.

vthumbe1503: So the null_ptr check originally was kepts as a way of determining discrete flags which was wrong as you mentioned. Now given that we already have the flags. Adding nullptr check is no longer needed.

There is no severe harm I agree. However, having it is extra code to read and extra compute(although negligible). Anyway compute_grouped_tensor_offsets function above is going to give a_offset, b_offset etc all zeros and we are going to nullptr even without the null check

Also, as I was seeing I see (for consistency perspective)
  int64_t a_offset = a_is_discrete ? 0 : compute_grouped_tensor_offset(A_meta, idx);
being done only for a and not for c and d.

transformer_engine/common/gemm/cublaslt_grouped_gemm.cu
```diff
@@ -1471,10 +1470,12 @@ __global__ void setup_grouped_gemm_kernel(
...
   };
 
-  if (a_scale_base) {
+  if (a_is_discrete) {
+    a_scale_inv_ptrs[idx] = a_multi_tensor_args.scale_inv_ptrs[idx];
+  } else if (a_scale_base) {
     fill_scale_ptr(a_scale_inv_ptrs, a_scale_base, A_meta, a_rowwise, a_scaling_mode);
   } else {
-    a_scale_inv_ptrs[idx] = a_multi_tensor_args.scale_inv_ptrs[idx];
+    a_scale_inv_ptrs[idx] = nullptr;
```
vthumbe1503, line 1478: This is not needed. Would increase divergence

vthumbe1503 (dismissed):
LGTM. My comments are nit. But we should clean those up

vthumbe1503 (approved):

PR #3229, opened 2026-07-21
[PyTorch] NCCL EP eager mode and drop-on-overflow policy

phu0ngng:
# Description
  Add two NCCL EP dispatch options to the PyTorch API:
  - **Eager mode**: sizes dispatch/combine outputs to the per-step recv-token total instead of the static `recv_capacity_per_rank`.
  - **Drop-on-overflow**: tokens past `max_recv_tokens_per_rank` are dropped and dispatch continues, instead of trapping.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes
  - `ep.h` / `ep_backend.cpp`: add `drop_on_overflow` group policy and wire `num_topk` + AUTO recv budget through the group config.
  - `pytorch/ep.py`: `ep_bootstrap(eager=...)` toggle captured at bootstrap; `EpBuffer` reads it to size outputs from the per-step recv-token total and exposes `num_recv_tokens`. Eager and zero-copy are mutually exclusive, and eager rejects caller-provided output buffers.
  - `pytorch/csrc/extensions`: plumb the new fields through the dispatch binding.

# Checklist:

### DIFF
--- a/transformer_engine/pytorch/ep.py
+++ b/transformer_engine/pytorch/ep.py
@@ -212,7 +212,7 @@ def is_symm_backed(t: torch.Tensor) -> bool:
 
 
 class EpBuffer:
-    """Per-microbatch EP layer state: handle_mem, token_counts, and shape/dtype config.
+    """Per-microbatch EP layer state: handle_mem, tokens_per_expert, and shape/dtype config.
     Use one EpBuffer per concurrently-in-flight call (e.g. per PP-1F1B microbatch).
     """
 
@@ -226,7 +226,7 @@ class EpBuffer:
         "num_local_experts",
         "payload_dtype",
         "device",
-        "token_counts",
+        "tokens_per_expert",
         "zero_copy",
         "eager",
         "total_recv_tokens",
@@ -262,7 +262,9 @@ class EpBuffer:
 
         size_bytes = tex.ep_handle_mem_size(self.top_k, self.alignment)
         self.handle_mem = torch.empty(int(size_bytes), dtype=torch.uint8, device=device)
-        self.token_counts = torch.empty(self.num_local_experts, dtype=torch.int64, device=device)
+        self.tokens_per_expert = torch.empty(
+            self.num_local_experts, dtype=torch.int64, device=device
+        )
         # Persistent tensor; keep resident if activation CPU offloading is on.
         mark_not_offload(self.handle_mem)
         # Per-step recv-token total (device int64 [1]), written by ep_prepare.
@@ -283,18 +285,18 @@ _LIB = "transformer_engine_ep"
 
 @torch.library.custom_op(
     f"{_LIB}::prepare",
-    mutates_args=("handle_mem", "token_counts", "total_recv"),
+    mutates_args=("handle_mem", "tokens_per_expert", "total_recv_tokens"),
     device_types="cuda",
 )
 def _prepare_op(
     handle_mem: torch.Tensor,
     top_k: int,
     topk_idx: torch.Tensor,
-    token_counts: torch.Tensor,
+    tokens_per_expert: torch.Tensor,
     alignment: int,
-    total_recv: torch.Tensor,
+    total_recv_tokens: torch.Tensor,
 ) -> None:
-    tex.ep_prepare(handle_mem, topk_idx, token_counts, top_k, alignment, total_recv)
+    tex.ep_prepare(handle_mem, topk_idx, tokens_per_expert, top_k, alignment, total_recv_tokens)
 
 
 @_prepare_op.register_fake
@@ -384,7 +386,7 @@ def _(*_args, **_kw):
 
 def ep_prepare(buffer: "EpBuffer", topk_idx: torch.Tensor) -> torch.Tensor:
     """AllGather the routing map; fills ``buffer.handle_mem`` and returns
-    ``buffer.token_counts`` (int64, shape [num_local_experts]). topk_idx must
+    ``buffer.tokens_per_expert`` (int64, shape [num_local_experts]). topk_idx must
     be int32 or int64.
 
     Always fills ``buffer.total_recv_tokens`` (device int64 [1]) with the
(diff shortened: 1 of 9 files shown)


transformer_engine/pytorch/ep.py
YangFei1990: With eager, do we still need to have `recv_capacity_per_rank` as a always required arg? I'm thinking to make it a kwarg, if user feed it, we mark eager to False, otherwise mark True. And we can get rid of eager kwarg.

transformer_engine/pytorch/ep.py
```diff
@@ -105,12 +111,29 @@ def ep_bootstrap(
     ``zero_copy`` opts the EP group into the symm-mem zero-copy IO path; pass
     ``True`` only when payload tensors are allocated via ``symm_mem_alloc``.
     Defaults to ``False``.
+
+    ``eager`` sizes the dispatch recv outputs from the per-step recv-token total
+    reported by ep_prepare, instead of the fixed ``recv_capacity_per_rank`` upper
+    bound. This requires a host sync each step, so it is not CUDA-graph
+    capturable. Mutually exclusive with ``zero_copy``. Defaults to ``False``.
+
+    ``max_num_topk`` is the upper bound on per-token top-k; it sizes NCCL EP
```
YangFei1990: Not sure I understand this, is just just the top_k? `upper bound on per-token top-k` is a bit confusing. If it is just topk, I would recommend to make it a arg as it is something constant that user can always provide. User does not need to be aware of the logic that only eager mode requires the topk.

transformer_engine/pytorch/ep.py
YangFei1990, line 587: Should we assert recv_tokens/recv_topk_weights None in eager mode? I see we have relevant assert in combine

transformer_engine/pytorch/ep.py
```diff
@@ -341,13 +386,26 @@ def _(*_args, **_kw):
...
+        buffer.handle_mem,
+        buffer.top_k,
+        topk_idx,
+        buffer.tokens_per_expert,
+        buffer.alignment,
+        buffer.total_recv_tokens,
     )
-    return buffer.token_counts
+    if buffer.eager:
+        buffer._host_total_recv_tokens = int(buffer.total_recv_tokens.item())
```
YangFei1990, line 412: Why do we need this `buffer._host_total_recv_tokens`? Can't we just call item whenever we need?

transformer_engine/pytorch/ep.py
```diff
@@ -539,36 +604,52 @@ def ep_dispatch(
...
+    recv_topk_weights, tokens_per_expert); tokens_per_expert is non-diff. See ``buffer.total_recv_tokens`` for
+    the per-step recv total.
     """
     _require_bf16("tokens", tokens)
     if topk_weights.dtype is not torch.float32:
         raise TypeError(
             f"topk_weights must be float32; got dtype={topk_weights.dtype}. "
             "Cast with topk_weights.float() before calling."
         )
+    skip_prepare = False
```
YangFei1990: This `skip_prepare` logic looks a bit awkward. Can we just move the `ep_prepare` from the autograd function to here always? Seems to me it is ok, in this way we also do not need to pass `buffer.total_recv_tokens` to autograd fn

transformer_engine/pytorch/ep.py
```diff
@@ -105,12 +111,29 @@ def ep_bootstrap(
...
+    global _BOOTSTRAPPED, _ATEXIT_REGISTERED, _EP_GROUP, _EAGER
     if _BOOTSTRAPPED:
         raise RuntimeError("ep_bootstrap was already called in this process")
     if ep_group.size() < 2:
         raise ValueError(f"ep_bootstrap requires ep_group.size() >= 2 (got {ep_group.size()}).")
+    if zero_copy and eager:
+        raise ValueError("ep_bootstrap: zero_copy and eager modes are mutually exclusive")
+    if eager and max_num_topk < 1:
+        raise ValueError("ep_bootstrap: eager mode requires max_num_topk >= 1")
+    if eager and drop_on_overflow:
```
YangFei1990: I wonder if we really need to assert this? Overflow should be irrelevant to eager right? When it is in eager mode, drop_on_overflow should be ignored automatically?

transformer_engine/pytorch/ep.py
phu0ngng: We can do that too. Let me update the design.

transformer_engine/pytorch/ep.py
phu0ngng: NCCL EP could work with the scenario in which there are multiple `num_topk` (in different layers), which we may not need.
We can simplify this to `num_topk` and make it mandatory.

transformer_engine/pytorch/ep.py
phu0ngng: I cached it so that it can be reused in the combine and bwd path without doing the D2H sync multiple times.

transformer_engine/pytorch/ep.py
phu0ngng: Already did so.

transformer_engine/pytorch/ep.py
phu0ngng: Removed.

transformer_engine/pytorch/ep.py
phu0ngng: For `drop_on_overflow`, we need `recv_capacity` which `eager` mode does not provide.

transformer_engine/pytorch/ep.py
```diff
@@ -217,10 +249,18 @@ def __init__(
         alignment = int(alignment)
         if alignment > 1 and (alignment & (alignment - 1)) != 0:
             raise ValueError(f"alignment must be 0, 1, or a power of two (got {alignment}).")
+        self.eager = _EAGER
```
YangFei1990, line 254: This adds a dependency that buffer creation needs to happen after bootstrap, can we add assertion for it?

YangFei1990 (dismissed):

transformer_engine/pytorch/ep.py
phu0ngng: Added

YangFei1990 (approved):

PR #3231, opened 2026-07-21
[JAX] Schedule EP dispatch/combine on XLA collective stream

phu0ngng:
# Description

Pin the TE-JAX Expert Parallelism dispatch/combine ops to XLA's high-priority collective stream so the scheduler serializes them with native collectives instead of overlapping.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/transformer_engine/jax/cpp_extensions/ep.py
+++ b/transformer_engine/jax/cpp_extensions/ep.py
@@ -19,16 +19,23 @@ from dataclasses import dataclass
 import jax
 import jax.numpy as jnp
 from jax import dtypes, ffi
-from jax.experimental.compute_on import compute_on
 from jax.sharding import NamedSharding, PartitionSpec
 
 import transformer_engine_jax
 from .base import BasePrimitive, register_primitive
 from ..sharding import global_mesh_resource, get_mesh_axis_size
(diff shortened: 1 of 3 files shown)


transformer_engine/jax/cpp_extensions/ep.py
```diff
@@ -19,12 +19,17 @@
 import jax
 import jax.numpy as jnp
 from jax import dtypes, ffi
+from jax.experimental.compute_on import compute_on
```
jberchtold-nvidia: What JAX version was this added in? Should we guard with something like
```
try:
   from jax.experimental.compute_on import compute_on
except ImportError:
   def compute_on(stream_name: str):
       """ compute_on shim for older JAX versions """
       # Make @compute_on annotation do nothing to the annotated function
       return lambda f: f
```

phu0ngng: Good suggestion for versioning.

The `import compute_on` is available since 2024, I think, so this check won't help here. 
I added a guard for JAX 0.10.0.

jberchtold-nvidia (approved):
LGTM pending CI, thanks!

PR #3235, opened 2026-07-22
Add stream-ordered CP gradient return primitive

foraxe:
# Description

Add one experimental CP=4 gradient-return primitive used by an opt-in Megatron
NVSHMEM context-parallel attention backend.

The primitive:

- scatters global dK/dV into symmetric owner buffers using Megatron's native
  two-chunk CP layout;
- publishes writer epochs with `cuStreamWriteValue32`;
- waits for all CP writers with `cuStreamWaitValue32`;
- accumulates four writer slots into the local owner's dK/dV;
- accepts peer tensor views supplied by the caller, so it does not add an
  NVSHMEM dependency to Transformer Engine.

This is intentionally narrow: three files and 122 additions. It does not add a
Transformer Engine attention backend or modify existing fused-attention
behavior.

## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- add a CUDA implementation of stream-ordered CP owner-gradient return;
- expose the operation through the PyTorch extension;
- dispatch the CUDA Driver stream-memory APIs through Transformer Engine's
  existing lazy driver loader, avoiding a hard `libcuda` link dependency.

## Validation

- Rebased the four-file patch onto current Transformer Engine `main`.
- Built the Python 3.12 / CUDA 13 PyTorch extension from the rebased source and
  verified that `nvshmem_cp_global_grad_return_execute` is present in the
  resulting extension.
- Rebuilt after review commit `f676f681`; `readelf` reports no `libcuda.so`
  dependency. Wheel SHA256:
  `7e6df6d2042fa14b6e9e72cad5e669aafa72e31809ad253a898f363d321f569f`.
- In the reviewed Transformer Engine 2.13 integration stack, ran four-GB200
  Megatron CP=4 training at S=262144, H=2560, 20 heads, head dimension 128,
  four layers, and vocabulary 157184.
- TE p2p and the NVSHMEM candidate matched all 73 checkpoint tensors with worst
  max-absolute difference `1.9073486328125e-6` (`atol=2e-6`, `rtol=0`).

For that single clean integrated comparison, median iteration time over updates
4-7 was 1707.45 ms for TE p2p and 1413.65 ms for the NVSHMEM candidate, 17.21%
lower. This is an integrated stack result, not a standalone speed claim for the
primitive in this PR.

## Checklist

- [x] The functionality is complete for the documented experimental contract
- [x] I have commented hard-to-understand ownership and synchronization logic
- [ ] I have made corresponding user documentation changes
- [x] My changes generate no new build warnings observed in local validation
- [ ] I have added an in-tree unit test
- [ ] The full existing test suite passes locally

The end-to-end correctness test currently lives in the dependent Megatron PR,
which owns symmetric allocation and peer tensor construction. The current-main
validation is a compile/export check; end-to-end runtime validation used the
reviewed Transformer Engine 2.13 integration branch.

## Stack relationship

- [Dao-AILab/flash-attention#2722](https://github.com/Dao-AILab/flash-attention/pull/2722)
  provides two-section causal backward over shared K/V.
- [NVIDIA/Megatron-LM#5969](https://github.com/NVIDIA/Megatron-LM/pull/5969)
  owns the opt-in backend, symmetric allocation, and peer tensor views passed
  to this primitive.

### DIFF
--- a/build_tools/pytorch.py
+++ b/build_tools/pytorch.py
@@ -97,7 +97,9 @@ def setup_pytorch_extension(
         cxx_flags.append("-DUSE_NCCL")
 
     library_dirs = []
-    libraries = []
+    # The CP gradient-return primitive uses stream-ordered CUDA Driver API
+    # writes and waits for its symmetric epoch protocol.
+    libraries = ["cuda"]
     if bool(int(os.getenv("NVTE_ENABLE_NVSHMEM", 0))):
         assert (
             os.getenv("NVSHMEM_HOME") is not None
--- a/build_tools/pytorch.py
+++ b/build_tools/pytorch.py
@@ -97,9 +97,7 @@ def setup_pytorch_extension(
         cxx_flags.append("-DUSE_NCCL")
 
     library_dirs = []
-    # The CP gradient-return primitive uses stream-ordered CUDA Driver API
-    # writes and waits for its symmetric epoch protocol.
-    libraries = ["cuda"]
+    libraries = []
     if bool(int(os.getenv("NVTE_ENABLE_NVSHMEM", 0))):
         assert (
             os.getenv("NVSHMEM_HOME") is not None
--- a/transformer_engine/pytorch/csrc/extensions.h
+++ b/transformer_engine/pytorch/csrc/extensions.h
@@ -121,6 +121,14 @@ std::vector<py::object> fused_attn_bwd(
     const std::optional<at::Tensor> cu_seqlens_kv_padded, py::handle s_quantizer,
     py::handle dp_quantizer, py::handle dqkv_quantizer, bool cuda_graph);
 
+std::vector<at::Tensor> nvshmem_cp_global_grad_return_execute(
+    at::Tensor dk_global, at::Tensor dv_global, at::Tensor key, at::Tensor value,
+    at::Tensor grad_key_return, at::Tensor grad_value_return,
+    at::Tensor grad_committed_epoch,
+    const std::vector<at::Tensor> &peer_grad_key_returns,
+    const std::vector<at::Tensor> &peer_grad_value_returns,
+    const std::vector<at::Tensor> &peer_grad_committed_epochs, int cp_size, int rank);
+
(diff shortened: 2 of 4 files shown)


sudhakarsingh27:
@foraxe, could you elaborate what's the motivation behind this. I see that this is a part of a bigger effort here https://github.com/NVIDIA/Megatron-LM/pull/5969 but why are we then hardcoding this for only CP=4, why not for CP=2 or 8?

foraxe:
Hi @sudhakarsingh27 

  Thanks for asking. The motivation is to **reduce Megatron CP attention communication overhead**
- In Megatron: We try to [Reduce CP attention communication overhead](https://github.com/NVIDIA/Megatron-LM/pull/5969), by using 
  - symmetric K/V storage and direct peer access in forward, 
  - direct owner gradient return in backward.
- This TE PR try to [Reduce CP **backward** dK/dV redistribution overhead](https://github.com/NVIDIA/TransformerEngine/pull/3235) through: 
  -  direct peer-buffer writes and stream-ordered GPU synchronization.

------

  CP=4 came from the production workload we were targeting and the hardware used for validation. The end-to-end stack was tested at CP=4, S=262144 on four GB200 GPUs, including training correctness against TE p2p and integrated timing. I added the CP=4 check as a fail-closed validation boundary, rather than implying that the operation fundamentally requires CP=4.

  That said, I agree that hardcoding the experimental integration constraint in the TE primitive is unnecessarily restrictive. The ownership rule itself is generic: for CP size N, owner o receives native sequence chunks o and (2 * N - 1 - o), and the implementation already sizes its peer lists,  epoch vectors, and accumulation loop from cp_size.

---

  I will update this PR to make the low-level primitive CP-size-generic and add focused CP=2/4/8 tests for its scatter, synchronization, and reduction contract. The dependent Megatron backend can remain explicitly gated to CP=4 until its full forward/backward path is validated at CP=2 and CP=8.
  This keeps the TE API general while being precise about what has end-to-end performance and correctness evidence today.

PR #3236, opened 2026-07-22
[All] Bump minimum supported cuDNN version to 9.12

cyanguwa:
# Description

This PR raises the minimum supported cuDNN backend version from 9.3 to **9.12.0**, enforces it at
configure, compile, and run time across all three TE components (C++ common core, PyTorch
extension, JAX extension), and removes the version gates that are now dead below the new
floor.

9.12.0 is over a year old, and we encourage users to update their cuDNN version to the latest if possible.

In the future, when updating the cuDNN floor, please pay attention to these places.
- The gate and its message in `transformer_engine/common/cmake/cuDNN.cmake`, which runs at configure time.
- The `kMinCudnnVersion` and its `static_assert` message in `transformer_engine/common/cudnn_min_version.h`, which checks the headers at build time.
- The `MIN_CUDNN_VERSION` in `transformer_engine/common/__init__.py` and the `NVTE_CHECK` in `transformer_engine/common/cudnn_utils.cpp`, which are the run-time checks.
- The requirements lists in `README.rst` and `docs/installation.rst`.
- The cuDNN pip install in the `te_ci` branch for core build (not a public repo).
- The `.github/workflows/minimum_versions.yml` in GitHub CI installs the newest cudnn9 package so no need for updating.

## Type of change

- [x] Infra/Build change
- [x] Code refactoring

## Changes

**Minimum version and enforcement**

- Declare the minimum in two places that are kept in sync: `kMinCudnnVersion = 91200` in the
  new `transformer_engine/common/cudnn_min_version.h`, and `MIN_CUDNN_VERSION = (9, 12, 0)`
  in `transformer_engine/common/__init__.py`.
- **Configure time (CMake):** `FATAL_ERROR` in `common/cmake/cuDNN.cmake`, failing in seconds
  rather than mid-compile. This covers the core library, which is always built through CMake;
  the two framework extensions are built by setuptools and rely on the `static_assert` below.
- **Compile time:** `static_assert` on `CUDNN_VERSION` in `cudnn_min_version.h`. It is split out
  from `cudnn_utils.h` so the extensions can assert the header version without pulling in the
  cuDNN frontend headers. Included by the core lib (via `cudnn_utils.h`), the PyTorch extension
  (`pytorch/csrc/extensions/misc.cpp`) and the JAX extension (`jax/csrc/extensions/attention.cpp`).
- **Run time (C++):** `CreateCuDNNHandle` in `common/cudnn_utils.cpp` validates
  `cudnnGetVersion()`. Every cuDNN handle in the core lib is created through `HandleManager`,
  so this covers all cuDNN-backed paths, including direct C-API users with no Python involved.
- **Run time (Python):** `check_cudnn_version()` is called from `load_framework_extension()`,
  which covers both framework extensions from a single place and fails at import with a clear
  message rather than deep inside a kernel launch. It runs before the `NVTE_PLUGIN` swap, so it
  always sees the real pybind module.

**Dead version gates removed** (every removed condition was below 9.12, so it was statically
true or unreachable at the new floor)

- C++ core: `fused_attn/config_and_params.cpp`, `fused_attn/fused_attn.cpp`,
  `fused_attn/fused_attn_f16_arbitrary_seqlen.cu`, `normalization/common.cpp`
- JAX: `cpp_extensions/attention.py`, `cpp_extensions/normalization.py`,
  `csrc/extensions/attention.cpp`
- PyTorch: `attention/dot_product_attention/utils.py`,
  `attention/dot_product_attention/context_parallel.py`
- Tests: skip guards below 9.12 dropped from 6 JAX and PyTorch test files. Guards for features
  that require *higher* versions (9.14, 9.18, 9.21, 9.23, 9.25) are intentionally kept.

**Refactoring and cleanup**

- Deduplicate `decode_cudnn_version()` into `common/__init__.py`; `jax/cpp_extensions/misc.py`
  and `pytorch/utils.py` had identical copies and now delegate to it.
- Replace the JAX score_mod `ScoreModCudnnHandleCache` with the shared
  `cudnnExecutionPlanManager`. It was a hand-rolled duplicate of `HandleManager` that called
  `cudnnCreate` directly, making it the one cuDNN path in the tree that bypassed the version
  check. score_mod now uses the same handle mechanism as fused attention and normalization:
  a thread-local, per-device cache, so all three share one cuDNN handle per (thread, device).
- Remove the now-unreachable `CUDNN_MAJOR_VERSION EQUAL 8` branch from `cuDNN.cmake`.
- Remove the unused `<ATen/cudnn/Handle.h>` include from `pytorch/csrc/common.h`.
- Mark `nvte_get_runtime_num_segments` deprecated. It worked around cuDNN before 9.3, which
  could not handle an actual sequence length of 0, and now has no in-tree callers. Also fixes
  its docblock, whose brief described an unrelated function.

**Documentation**

- `README.rst` and `docs/installation.rst`: cuDNN 9.3 → 9.12.
- `docs/examples/attention/attention.ipynb`: this file needs an overhaul; will do in another PR.

# Checklist:


KshitijLakhani (commented):
IIUC, the minimum version check does not mean that we cannot build with <9.11 right ? It just means that older versions (features) are just unsupported, right ? So the 9.11 is merely a lowest tested/supported version and not a hard runtime/build floor right ?
If my understand is right we do not need to worry about compilation checks during build and these: 
https://github.com/NVIDIA/TransformerEngine/blob/65bdbe317b6480bbe335f2ba3d61df6d75cd47e2/transformer_engine/jax/csrc/extensions/attention.cpp#L186 ?

https://github.com/NVIDIA/TransformerEngine/blob/65bdbe317b6480bbe335f2ba3d61df6d75cd47e2/transformer_engine/jax/cpp_extensions/attention.py#L602

https://github.com/NVIDIA/TransformerEngine/blob/65bdbe317b6480bbe335f2ba3d61df6d75cd47e2/transformer_engine/common/fused_attn/fused_attn.cpp#L364

https://github.com/NVIDIA/TransformerEngine/blob/65bdbe317b6480bbe335f2ba3d61df6d75cd47e2/tests/pytorch/attention/test_attention.py#L1035


------------------------
However, I think these below two might needs anotehr look right ?
- Wouldn't the notebook here need a refresh  : `docs/examples/attention/attention.ipynb` ?
- Also, would we want to refresh the assert in here: `tutorial_generation_gemma_with_te.ipynb` ?

sbhavani:
For min version, maybe we could match with a previous PyTorch NGC container (e.g. cuDNN 9.12.0.46 for 25.08) from [frameworks support matrix](https://docs.nvidia.com/deeplearning/frameworks/support-matrix/index.html#framework-matrix-2025).

cyanguwa:
I'll do a bit more cleanup throughout the code to support this update, and will update to cuDNN 9.12.

sbhavani (dismissed):

cyanguwa:
> IIUC, the minimum version check does not mean that we cannot build with <9.11 right ? It just means that older versions (features) are just unsupported, right ? So the 9.11 is merely a lowest tested/supported version and not a hard runtime/build floor right ? 

This PR makes 9.12 a hard floor, rather than a "lowest tested" marker. Users will not be able to build against pre-9.12 headers or run against pre-9.12 runtimes. I don't think we can go the "lowest tested" version route while not maintaining/testing older logics.

sbhavani (dismissed):

cyanguwa:
Failures of L0/L1 cpp tests on all arches, and L1/L2 Jax distributed tests on B200s are not related to this PR.

transformer_engine/common/cmake/cuDNN.cmake
```diff
@@ -38,6 +38,12 @@ if(CUDNN_INCLUDE_DIR)
     else()
         set(CUDNN_VERSION
             "${CUDNN_VERSION_MAJOR}.${CUDNN_VERSION_MINOR}.${CUDNN_VERSION_PATCH}")
+        # Keep in sync with kMinCudnnVersion in transformer_engine/common/cudnn_min_version.h.
```
sbhavani, line 41: is it an issue that this check doesn't fire when lines 2-5 mark cuDNN as found and return?

sbhavani (approved):

transformer_engine/common/cmake/cuDNN.cmake
cyanguwa: Yes, if there's already a cuDNN build in the environment, CMake will go through Lines 2-5 and not here. When the version falls below 9.12, TE will fail at build time (instead of config time here), so it's still safe.

I did play around with adding more code here but in the end I decided to leave it as is, because the version check still fires pretty early on (compile time). Thanks.

PR #3244, opened 2026-07-23
[Pytorch] Add support for row-wise quanted input for grouped gemm

YangFei1990:
# Description

To support the NCCL EP FP8 dispatch, this PR introduce a capability to allow grouped gemm to take input that is already quantized in row-wise. 
- For forward, we directly pass the row-wise quantized for forward grouped gemm and dequant + requant with col-wise to save for backward.
- For backward, we pass the row-wise quantized for dgrad computation and dequant + requant with col-wise for wgrad

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:


transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -1296,11 +1303,42 @@ def _fuser_forward_grouped_tensor(
 
         # Flatten to 2D so the first dim is the total token count.
         original_shape = list(input_.size())
-        x = maybe_dequantize(input_, dtype).reshape(-1, self.in_features)
-        total_tokens = x.size(0)
+        prequantized_mxfp8_input = (
+            with_quantized_compute
+            and isinstance(input_, GroupedTensor)
+            and isinstance(input_quantizers[0], MXFP8Quantizer)
+            and isinstance(input_.quantizer, MXFP8Quantizer)
```
phu0ngng: Should we also assert `input_quantizers[0] == input_.quantizer`?

transformer_engine/pytorch/ops/_common.py
```diff
@@ -66,6 +71,128 @@ def maybe_dequantize(
...
+    grouped_x: GroupedTensorStorage,
+    quantizer: MXFP8Quantizer,
+    num_groups: int,
+    split_sizes: torch.Tensor,
+    dtype: torch.dtype,
+    *,
+    with_columnwise: bool,
+    tensor_offsets: Optional[torch.Tensor] = None,
+) -> None:
+    """Prepare a rowwise-only MXFP8 grouped input for grouped GEMM (in place).
```
phu0ngng: Looks like this helper actually "prepares" colwise besides rowwise. I think we should make the function name and docstring clearer and more consistent.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -66,6 +71,128 @@ def maybe_dequantize(
...
+    Supports inputs that arrive already rowwise-quantized (e.g. FP8 token
+    dispatch): the rowwise data is fed to the forward GEMM as-is, while the
+    columnwise copy needed by the wgrad GEMM cannot be derived from the
+    rowwise data (per-block scales differ per direction), so it is
+    manufactured by dequantizing the rowwise data and requantizing
+    columnwise-only. Rowwise scales must arrive in compact (unswizzled)
+    format; they are converted to the GEMM-swizzled layout afterwards.
+    TODO: optimize and fuse the round-trips requant
+    """
+    if grouped_x.rowwise_data is None:
```
phu0ngng: Should we validate `grouped_x.rowwise_scale` as well?

transformer_engine/pytorch/ops/_common.py
```diff
@@ -66,6 +71,128 @@ def maybe_dequantize(
...
+        raise ValueError("Pre-quantized MXFP8 grouped input is missing rowwise data.")
+    if grouped_x._with_gemm_swizzled_scales:
+        raise NotImplementedError(
+            "Pre-quantized MXFP8 grouped input must have scales in compact format."
```
phu0ngng: compact -> unswizzled.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -66,6 +71,128 @@ def maybe_dequantize(
...
+
+    # Convert rowwise scales to the GEMM-swizzled layout. The grouped GEMM
+    # reads activation scales as one (total_tokens, cols) matrix, so the
+    # single-tensor swizzle applies. Swizzling allocates a new scale buffer;
+    # the original compact scales are left untouched.
+    total_tokens, cols = grouped_x.logical_shape
+    scale_shape = (
+        round_up_to_nearest_multiple(total_tokens, 128),
+        round_up_to_nearest_multiple(cols // 32, 4),
+    )
```
phu0ngng: Do we need to do this while we already have 128 per-expert-alignment in Dispatch? The total tokens should already be divisible by 128.

For hidden size, there will be an assertion in Dispatch to expect % 512 = 0. 

Here, I think we can add assertions instead.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -66,6 +71,128 @@ def maybe_dequantize(
...
+        with_gemm_swizzled_scales=tensor._with_gemm_swizzled_scales,
+        row_scaled_nvfp4=tensor.row_scaled_nvfp4,
+        nvfp4_use_4over6=tensor.nvfp4_use_4over6,
+        nvfp4_e4m3_max=tensor.nvfp4_e4m3_max,
+    )
+
+
+def prepare_prequantized_mxfp8_grouped_input(
+    grouped_x: GroupedTensorStorage,
+    quantizer: MXFP8Quantizer,
```
phu0ngng: Actually, is there a GroupedQuantizer in TE/PyT? If so, we probably need to use it instead.

transformer_engine/pytorch/ops/basic/grouped_linear.py
YangFei1990: I think `==` comparison is too strict. It will compare all value fields, I think things like `columnwise_usage` or `optimize_for_gemm` could mismatch? Maybe we should explicit assert dype to be `E4M3`? But let me know if the dispatch output tensor's quantizer can match all value fields.

transformer_engine/pytorch/ops/_common.py
YangFei1990: I do not see a GroupedQuantizer in Pytorch. I see what used is `tex.group_quantize`. Let me know if I missed it. Also can you help me understand the benefit of GroupedQuantizer and how it compared with current Pytorch approach?

tests/pytorch/test_grouped_mlp.py
```diff
@@ -436,6 +454,189 @@ def test_grouped_linear(
...
 
+    @staticmethod
+    def _make_rowwise_mxfp8_wire_input(
+        x_hp: torch.Tensor,
+        group_size: int,
+        split_sizes: torch.Tensor,
+    ) -> "GroupedTensor":
+        """Rowwise-only MXFP8 GroupedTensor with compact scales (FP8 dispatch wire format)."""
+        wire_quantizer = MXFP8Quantizer(
+            fp8_dtype=tex.DType.kFloat8E4M3, rowwise=True, columnwise=False
```
vthumbe1503: Recommendation is to use TE_DType instead of tex.DType when constructing quantizers now, could we please do that here?

transformer_engine/pytorch/ops/_common.py
```diff
@@ -66,6 +72,199 @@ def maybe_dequantize(
...
+        fp8_dtype=quantizer.dtype,
+        rowwise_data=grouped_x.rowwise_data.view(total_tokens, cols),
+        rowwise_scale_inv=grouped_x.scale_inv.view(scale_shape),
+        columnwise_data=None,
+        columnwise_scale_inv=None,
+        quantizer=quantizer,
+        requires_grad=False,
+        with_gemm_swizzled_scales=False,
+    )
+    tex.swizzle_scales_for_gemm_(tmp)
```
zhongbozhu: It feels like we are swizzling as if the activation is a dense tensor, which is only valid is the activation has tokens-per-expert padded to 128 multiple or higher (like 256). 

How are we handling the swizzle of columnwise scales?

transformer_engine/pytorch/ops/_common.py
```diff
@@ -66,6 +72,199 @@ def maybe_dequantize(
...
+        scale_inv_offsets=tensor.scale_inv_offsets,
+        columnwise_scale_inv_offsets=tensor.columnwise_scale_inv_offsets,
+        with_gemm_swizzled_scales=tensor._with_gemm_swizzled_scales,
+        row_scaled_nvfp4=tensor.row_scaled_nvfp4,
+        nvfp4_use_4over6=tensor.nvfp4_use_4over6,
+        nvfp4_e4m3_max=tensor.nvfp4_e4m3_max,
+    )
+
+
+def prepare_prequantized_mxfp8_input_for_gemm(
```
zhongbozhu: This function needs to be heavily unit tested like here: https://github.com/NVIDIA/TransformerEngine/blob/98524e5ddb52eb27639a126cb5474db6fb2f1adc/tests/pytorch/mxfp8/test_mxfp8_group_quantize_graph_safe.py#L333

transformer_engine/pytorch/ops/_common.py
```diff
@@ -66,6 +72,199 @@ def maybe_dequantize(
...
+            f"Pre-quantized MXFP8 grouped input has FP8 dtype {grouped_x.quantizer.dtype}, "
+            f"but the op's input quantizer expects {quantizer.dtype}."
+        )
+
+    # Manufacture columnwise data for the wgrad GEMM: dequantize the rowwise
+    # wire data and requantize columnwise-only.
+    dbias = None
+    dequantized = None
+    if with_columnwise or with_dbias or with_dequantized:
+        dequantized = tex.group_dequantize(grouped_x, TE_DType[dtype]).rowwise_data.view(
```
zhongbozhu: Does group dequantize work with already swizzled scales? Would be better if we put an assert here.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -66,6 +72,199 @@ def maybe_dequantize(
...
+            f"Pre-quantized MXFP8 grouped input has {grouped_x.scale_inv.numel()} rowwise "
+            f"scales, but expected {math.prod(scale_shape)} for shape {scale_shape}."
+        )
+    tmp = MXFP8Tensor(
+        shape=(total_tokens, cols),
+        dtype=dtype,
+        fp8_dtype=quantizer.dtype,
+        rowwise_data=grouped_x.rowwise_data.view(total_tokens, cols),
+        rowwise_scale_inv=grouped_x.scale_inv.view(scale_shape),
+        columnwise_data=None,
```
zhongbozhu: Why is colwise data None if we have `if with_columnwise:` above?

transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -919,7 +921,16 @@ def fuser_forward(
         # Tensor properties
         fc1_weight_shape = (fc1_op.out_features, fc1_op.in_features)
         fc2_weight_shape = (fc2_op.out_features, fc2_op.in_features)
-        input_ = input_.reshape(-1, fc1_weight_shape[1])
+        if isinstance(input_, GroupedTensor):
```
zhongbozhu, line 1001: Overall I think it makes sense to limit the scope of mxfp8 alltoall to the TE fused ops module because it assumes using grouped kernels already so it's fine if we enforce shape alignment on top of it and we can safely ignore the zero padding of mxfp8 scaling factors. However, was NCCL-EP designed to work with TE fused ops only? Sounds like we should insert more assertions elsewhere if TE fused ops module is not used.

zhongbozhu (changes requested):

transformer_engine/pytorch/ops/fused/grouped_mlp.py
zhongbozhu: cc @timmoon10

YangFei1990: NCCL EP itself is designed to be compatible with both fused and unfused path. We would like to make it general and in mcore I built many support for both paths. When mcore is ready for the grouped tensor path (which I think you are working on?), we can adopt NCCL EP to that path as well.

For mxfp8 specifically, the major connected interface is dispatch (combine takes bf16 input). TE EP dispatch will return `GroupedTensor` with only rowwise quantized, I think it should work with both paths?

cc @phu0ngng

transformer_engine/pytorch/ops/_common.py
YangFei1990: We have an assert above in line 177

transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -1659,12 +1666,36 @@ def fuser_backward(
...
             and fc2_grad_output_quantizer_matches
         ):
-            grouped_fc2_dy = grad_output
+            if prequantized_mxfp8_grad:
+                # Rowwise-only MXFP8 grad output (e.g. FP8 token dispatch): reuse
+                # the rowwise data for the dgrad GEMM and manufacture FC2's
+                # columnwise copy for wgrad. ``scale_bias`` needs the
+                # high-precision grad below, so it takes the dequantized tensor
+                # instead of the fused dbias.
+                grouped_fc2_dy = grouped_storage_from_grouped_tensor(grad_output)
```
vthumbe1503: There is already copy method defined in GroupedTensorStorage class to achieve this functionality of creating a shallow storage copy. Could we please use that and remove this method?
```suggestion
                grouped_fc2_dy = grad_output.copy()
```

vthumbe1503: Same applies in all other places where this method is used,

transformer_engine/pytorch/ops/_common.py
YangFei1990: When we have `colwise_quantizer.optimize_for_gemm = True`, the colwise data should be already swizzled in the quantize kernel. We need to manually swizzle for rowwise data because it is generated from NCCL EP dispatch.

transformer_engine/pytorch/ops/_common.py
YangFei1990: Yes this is the precondition, that we require the input dimension (both row and col) to be 128 multiple or higher. While I built in this way, please let me know if we should handle padding here.

transformer_engine/pytorch/ops/_common.py
```diff
@@ -66,6 +72,199 @@ def maybe_dequantize(
...
+        columnwise_scale_inv=None,
+        quantizer=quantizer,
+        requires_grad=False,
+        with_gemm_swizzled_scales=False,
+    )
+    tex.swizzle_scales_for_gemm_(tmp)
+    grouped_x.scale_inv = tmp._rowwise_scale_inv.view(-1)
+    grouped_x._with_gemm_swizzled_scales = True
```
vthumbe1503: I think the better way should be to directly use the grouped_swizzle API which handles variable dims as well.

This would work only if all per expert dims are padded to 128.

transformer_engine/pytorch/ops/_common.py
vthumbe1503: I am uncomfortable with the complexity of this function. 

I think we should define two functions, one for the `prepare_rowwise_mxfp8_input` for forward pass and `preparing_colwise_mxfp8_input` for the backward pass.

To be honest I dont think `prepare_rowwise_mxfp8_input` is even needed, Simply calling grouped_swizzle API on the activation should be enough which can be done directly before the GEMM in grouped_linear.py and grouped_mlp.py. However, the perf of the variable swizzle kernel is not being well studied. So it would be good to evaluate that before making the change.

For colwise, I am assuming the rowwise dequant + quant is well studied in terms of convergence. And so having a sperate function call for that is ideal which also takes care of the dbias computation

transformer_engine/pytorch/ops/_common.py
vthumbe1503: @zhongbozhu I think the group_quantize + swizzle fusion generates the swizzled scales for columnwise.

zhongbozhu: what you are doing is essentially this right? 

`group_quantize_colwise_only_swizzle_fused( group_dequantize( mxfp8_input_compact_scales ) )` 

It would be better if we have a fused kernel for this process, is this already planned?

transformer_engine/pytorch/ops/_common.py
zhongbozhu: sounds good

transformer_engine/pytorch/ops/_common.py
zhongbozhu: Thanks, would be better if this kernel is also fused. 

Nit: In case we don't have the kernel yet, I would still prefer having a `tex.` API calling unfused kernels in C++, and replace its implementation with a fused kernel when it's ready.

transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -1206,36 +1216,27 @@ def fuser_forward(
...
-                tensor_offsets=input_.tensor_offsets,
-                offsets=input_.offsets,
-                scale_inv_offsets=input_.scale_inv_offsets,
-                columnwise_scale_inv_offsets=input_.columnwise_scale_inv_offsets,
-                with_gemm_swizzled_scales=input_._with_gemm_swizzled_scales,
-                row_scaled_nvfp4=input_.row_scaled_nvfp4,
-                nvfp4_use_4over6=input_.nvfp4_use_4over6,
-                nvfp4_e4m3_max=input_.nvfp4_e4m3_max,
-            )
+            grouped_fc1_x = input_.copy()
```
phu0ngng, line 1234: Just to confirm, this is not a deep copy, right?

vthumbe1503: Correct, this is a shallow copy. Internal buffers remain the same

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -663,6 +663,70 @@ py::object group_dequantize(const py::handle &input, transformer_engine::DType o
   return py::reinterpret_borrow<py::object>(out_py);
 }
 
+py::object group_requantize_columnwise_and_swizzle_rowwise_(
+    py::handle grouped_x, py::handle columnwise_quantizer, const size_t num_tensors,
+    std::optional<at::Tensor> first_dims, DType otype, std::optional<at::Tensor> tensor_offsets,
+    bool return_dequantized) {
```
vthumbe1503:
```suggestion

py::object group_requantize(
    py::handle grouped_x, py::handle quantizer, const size_t num_tensors,
    std::optional<at::Tensor> first_dims, DType otype, std::optional<at::Tensor> tensor_offsets,
    bool return_dequantized) {
```

I suggest to make this function generic.

The intent of this function as follows
1. If quantizer.rowwise_usage is true and grouped_x has rowwise data/scales, then grouped_swizzle
2. If quantizer.columnwise_usage is true and grouped_x has columnwise_data/scales, then grouped_swizzle
3. If quantizer.rowwise_usage is true and grouped_x doesnt have rowwise_data/scales, then dequant + requant with swizzle fusion
4. If quantizer.columnwise_usage is true and grouped_zx doesnt have columnwise_data/scales, then dequant + requant with swizzle fusion

3 --> is a unrealistic case and we can throw an error in that case as well

Also, we should make sure to gather all swizzling directions once(rowwise, colwise or both rowwise/colwise) and call grouped_swizzle once.

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -1325,11 +1331,46 @@ def _fuser_forward_grouped_tensor(
...
+            # (total_tokens, in_features) layout; just validate the shape.
+            if input_.dim() != 2 or input_.size(-1) != self.in_features:
+                raise ValueError(
+                    "GroupedTensor input must have shape (total_tokens, "
+                    f"{self.in_features}), but got {tuple(input_.size())}."
+                )
+            total_tokens = input_.size(0)
+        else:
+            x = maybe_dequantize(input_, dtype).reshape(-1, self.in_features)
+            total_tokens = x.size(0)
```
timmoon10: Adding a special case for MXFP8 input is quite hacky. If we try to understand the code, this entire code section is converting the input (previously assumed to be BF16, but now may be grouped MXFP8) to the format needed by GGEMM. We can lift all of this into a clean helper function:
```python
def _convert_input_to_grouped_tensor(input_, ...):

    # Do nothing if input is already in expected format
    if input_is_in_expected_format:
        return input_

    # Fast requantize impls
    if input_is_mxfp8 and compute_is_mxfp8:
        return tex.group_requantize...(x)
    if fancy_future_fused_impl_is_available:
        return tex.fancy_future_fused_impl(...)

    # Fallback: dequantize if needed and group quantize
    x = maybe_dequantize(input_)
    grouped_x = tex.group_quantize(x, ...)
    return grouped_x
```

transformer_engine/pytorch/csrc/extensions/cast.cpp
YangFei1990: I'm not sure if we should make this function generic, the purpose of it is to handle the mxfp8 output from dispatch, and we are also talking about make this a fused kernel with dispatch's permutation, that is the major reason to refactor it from python into c++ layer. cc @phu0ngng

vthumbe1503: So the thing is, for now it is now serving the purpose of 
mxfp8 dispatch --> mxfp8 gemm ready input

But it can be this as well in the future
mxfp8 dispatch --> nvfp4 gemm ready input

Now if your intent is to retain the quantization between input and output
like mxfp8 dispatch --> mxfp8 gemm ready input
or  nvfp4 dispatch --> nvfp4 gemm ready input
I am ok with that as well. But that should mentioned in the comment.

group_requantize_columnwise_and_swizzle_rowwise_ --> This as a name seems too specific and encodes too much information in the name of the function which isnt needed. Quantizer already has the information with optimize_for_gemm=True/False which tells whether to swizzle or not and which direction to swizzle. So just keeping the name as "group_requantize" should suffice.

As far as fused kernels are concerned, the special case where fusion is available, only that case can be replaced with the fused kernel

In general, I want to differentiate between use-case of a function and intent of the function. Use-case is dispatch --> gemm-ready handling. But the Intent of the function is  --> requantize to the best of the ability. And that means if you have already quantized data along a direction, then to make gemm ready you just need to swizzle it. However if you dont have quantized data along a direction, then to make it gemm ready dequant + quant + swizzle(based on quantizer config)

vthumbe1503: Currently in your function 1 and 4 that I mentioned above is handled. We can throw error for 2 and 3 as well based on quantizer config and grouped_x.quantizer config.  So I am not asking to implement the generic function. But keep the interface and name of the function generic. So that in future if we want to implement new feature we dont have to change the name and signature of the function

transformer_engine/pytorch/ops/basic/grouped_linear.py
YangFei1990: There are four sites that these checks are involved, fwd_grouped_linear (this one), bwd_grouped_linear, fwd_grouped_mlp, bwd_grouped_mlp, the paths have common parts, but also differ a lot, e.g. quantize entry point, handle of dbias, NVFP4 path, fallback paths, I feel if we create a single helper function to handle all cases, the function itself might be complicated with a lot of conditional branches. What are your recommendations?

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -1700,7 +1758,34 @@ def _fuser_backward_grouped_tensor(
             fuse_bgrad = isinstance(grad_output_quantizer, MXFP8Quantizer) or (
                 isinstance(grad_output_quantizer, Float8BlockQuantizer) and ctx.input_requires_grad
             )
-            if has_bias and not self._scale_bias and fuse_bgrad:
+            if prequantized_mxfp8_grad:
+                # Rowwise-only MXFP8 grad output (e.g. FP8 token dispatch): reuse the
+                # rowwise data for the dgrad GEMM and manufacture the columnwise copy
+                # for wgrad. Bias grads are reduced from the dequantized grad below,
+                # which is only kept when there is a bias.
+                grouped_dy = grad_output.copy()
```
zhongbozhu, line 1771: It will be better if we let the `group_requantize_columnwise_and_swizzle_rowwise_` API accepts regular 2D inputs or dY, convert to grouped tensor in C++ for less CPU overhead.

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -1325,11 +1331,46 @@ def _fuser_forward_grouped_tensor(
...
-        if with_quantized_compute:
+        if prequantized_mxfp8_input:
+            # Rowwise-only MXFP8 input (e.g. FP8 token dispatch): feed the
+            # rowwise data to the forward GEMM as-is, manufacture the
+            # columnwise copy needed by the wgrad GEMM, and swizzle the
+            # rowwise scales for the GEMM.
+            grouped_x = input_.copy()
+            if weight_requires_grad:
+                tex.group_requantize_columnwise_and_swizzle_rowwise_(
+                    grouped_x,
```
zhongbozhu: same comment, it would be better if we just pass in plain torch tensor into this API (please refer to our API design for `grouped_quantize`, and the grouped tensor descriptor is created in C++ for less overhead

transformer_engine/pytorch/csrc/extensions/cast.cpp
YangFei1990: I see. Renamed to `group_requantize` and added one additional assert. Currently the function only takes rowwise_data and column quantizer, but we can extend in the future.

transformer_engine/pytorch/ops/basic/grouped_linear.py
vthumbe1503: Agreed that It would reduce CPU overheads, however we would loose generality of the function. If we want to use it for NVFP4 later, then we would need to pass amax as well. I think for now, it is better to consume grouped tensor

vthumbe1503 (dismissed):
LGTM. Pending CI.

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -1703,8 +1733,20 @@ def _fuser_backward_grouped_tensor(
...
+            # (total_tokens, out_features) layout; just validate the shape.
+            if grad_output.dim() != 2 or grad_output.size(-1) != self.out_features:
+                raise ValueError(
+                    "GroupedTensor grad output must have shape (total_tokens, "
+                    f"{self.out_features}), but got {tuple(grad_output.size())}."
+                )
+            dy_2d = None
+            total_tokens = grad_output.size(0)
+        else:
+            dy_2d = grad_output.reshape(-1, self.out_features)
```
vthumbe1503, line 1748: This wont work for BF16 GroupedTensor generated from Dispatch.

YangFei1990: Dispatch won't generate bf16 GroupedTensor, it will be a plain torch tensor. Only MXFP8 will be GroupedTensor

vthumbe1503: We have been considering outputting GroupedTensor even in GroupedLinear and ScaledActivation in future and in that case we might need to support bf16 GroupedTensor.

But i guess for now it is ok to not support it,

vthumbe1503 (approved):

PR #3251, opened 2026-07-24
[CI] Improve build time dependency resolution

fheinecke:
# Description

There are a few issues with how build-time dependencies are resolved for native code build paths today:
* Installed python packages are not checked when looking up some dependencies
* The CUDA toolkit root is not auto-discovered when it's built via python deps
* Paths to libraries and headers installed via python packages are not plumbed to cmake

The goal here is to allow all build-time requirements to be installed via wheel. This fixes many of the issues blocking TE wheel build isolation.

This change aims to be 100% backwards compatible.

Fixes # (issue)

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- Shared libraries and headers installed via wheel are now provided to cmake
- CUDA toolkit root built via python wheels is now provided to cmake
- NVCC installed via wheel is now used when compiling the JAX extension

# Checklist:

### DIFF
--- a/build_tools/build_ext.py
+++ b/build_tools/build_ext.py
@@ -19,6 +19,8 @@ import setuptools
 
 from .utils import (
     cmake_bin,
+    cuda_home_path,
+    cuda_version,
     debug_build_enabled,
     found_ninja,
     get_frameworks,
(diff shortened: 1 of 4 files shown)


build_tools/build_ext.py
```diff
@@ -61,6 +63,31 @@ def _build_cmake(self, build_dir: Path, install_dir: Path) -> None:
...
+        if discovered_cuda_home is not None:
+            configure_command += [
+                f"-DCUDAToolkit_ROOT={discovered_cuda_home}",
+                # CUDA wheels use `lib`, while nvcc's host linker expects `lib64`.
+                f"-DCMAKE_CUDA_FLAGS=-L{discovered_cuda_home}/lib",
+            ]
+
+            if cuda_full_version := cuda_version():
+                cuda_major_version = cuda_full_version[0]
+                cuda_lib_dir = discovered_cuda_home / "lib"
```
fheinecke: TODO this could be lib or lib64 (this is the cause of the CI failures)

build_tools/jax.py
```diff
@@ -117,6 +121,12 @@ def setup_jax_extension(
     if nccl_ep_enabled():
         cxx_flags.append("-DNVTE_WITH_NCCL_EP")
 
+    kwargs = {}
+    if (discovered_nccl_lib_path := nccl_lib_path()) is not None:
+        kwargs["extra_objects"] = [str(discovered_nccl_lib_path)]
```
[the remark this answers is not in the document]
fheinecke: This is conflating an existing issue with this PR. Today, TE does not support loading nccl from site-packages the way it does other libraries. This is likely an oversight, but it should be fixed in a separate PR.

The only impact is that if TE is built from source against NCCL installed via wheel, it will error at runtime if nccl is only available as a python package (because of the above existing issue), instead of at build time (which this PR now allows for).

build_tools/utils.py
```diff
@@ -175,6 +175,90 @@ def found_pybind11() -> bool:
     return False
 
 
+@functools.lru_cache(maxsize=None)
+def cuda_home_path() -> Optional[Path]:
```
timmoon10: We have somewhat inconsistent search logic:
- `cuda_home_path`: prioritize `CUDA_HOME`, otherwise look in `PYTHONPATH` for `nvidia/*-cu13/bin` (assumed to hold NVCC).
- `cuda_toolkit_include_path`: prioritize `CUDA_HOME`, then look in `PATH` for NVCC, then in `/usr/local/cuda/include`.
- `nvcc_path`: look in `cuda_home_path()`, then `PATH`, then Python libs, then `/usr/local/cuda`

It seems `cuda_home_path` and `cuda_toolkit_include_path` are primarily looking for NVCC, so could we consolidate logic?
- `nvcc_path`: prioritize `CUDA_HOME`, then look in `PYTHONPATH`, then Python libs, then `PATH`, then `/usr/local/cuda`
- `cuda_home_path`: look in `nvcc_path()`
- `cuda_toolkit_include_path`: look in `cuda_home_path()`

timmoon10 (dismissed):
Overall LGTM. I have a suggestion to consolidate logic, but it's not required if this PR is blocking more important improvements.

ksivaman (approved):
LGTM

timmoon10 (approved):

fheinecke:
CI failures are due to the triton increase memory usage bug, not this PR

PR #3252, opened 2026-07-24
Enable runtime resolution of CUDA header path for NVRTC

fheinecke:
# Description

NVRTC relies on having CUDA toolkit headers available at runtime for JIT compilation. The following directories are searched:
1. `${NVTE_CUDA_INCLUDE_DIR}`
2. `${CUDA_HOME}`
3. `${CUDA_DIR}`
4. The **build-time** header path
5. `/usr/local/cuda`

The build-time path may coincidentally line up with the runtime path, but may be incorrect when building TE on another machine or environment, or when building TE with build isolation enabled. This enables searching site-packages at runtime for the headers, allowing them to be resolved in more environments. If checked paths don't exist, then this will fall back to the existing behavior.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Enable runtime resolution of CUDA toolkit headers in site-packages for NVRTC

# Checklist:

### DIFF
--- a/transformer_engine/common/util/cuda_runtime.cpp
+++ b/transformer_engine/common/util/cuda_runtime.cpp
@@ -8,6 +8,8 @@
 
 #include <cublasLt.h>
 
+#include <dlfcn.h>
+
 #include <filesystem>
 #include <fstream>
 #include <mutex>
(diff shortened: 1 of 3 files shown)


fheinecke:
Failing test is due to runner scheduling, should be unrelated to this

transformer_engine/common/util/cuda_runtime.cpp
```diff
@@ -26,6 +27,83 @@ namespace {
...
+
+std::filesystem::path python_cuda_directory() {
+  using Path = std::filesystem::path;
+
+  // Find the Python package root from the installed Transformer Engine package. Do not
+  // assume that the root is named site-packages or dist-packages since valid installs
+  // may use an arbitrary target directory.
+  Path te_package_directory = shared_library_directory();
+  while (true) {
+    if (te_package_directory.filename() == "transformer_engine") {
```
ptrendx, line 70: We should also break (and raise an error) if we reach `/` directory (just in case there is something strange going on and our root directory is not named transformer_engine).

fheinecke: It does return early [here](https://github.com/NVIDIA/TransformerEngine/pull/3252/changes#diff-6710d29b3eeaf4928b1796bcb425bd476c172867bebe4f4c14a49194f681ebc2R74-R78), but does not currently throw.

I'm happy to make this throw if you feel strongly about it, but this would break [the fallback to today's search behavior](https://github.com/NVIDIA/TransformerEngine/pull/3252/changes#diff-6710d29b3eeaf4928b1796bcb425bd476c172867bebe4f4c14a49194f681ebc2R234-R235). A directory named `transformer_engine` might not be in the shared lib's path in either of these scenarios:
* Editable installs (`pip install -e .` where the repo root is at `~/repos/TransformerEngine` places the library at `~/repos/TransformerEngine/libtransformer_engine.so`)
* Installing only the shared library ([this docs page](https://github.com/NVIDIA/TransformerEngine/blob/main/docs/api/c/index.rst) implies that this is supported)

The second case is covered by `L0_cpuunittest`, which would break if this function were to throw an exception. I can certainly fix this, but it would change one of the contracts that the test covers.

ptrendx: ok, I see - did not notice that it is already handled. I also agree that not throwing is a better choice to enable fallback. Thanks!

ptrendx (approved):

PR #3254, opened 2026-07-24
Use a temp directory for CMake build

fheinecke:
# Description

CMake currently always uses `./build/cmake` as a build directory when `NVTE_CMAKE_BUILD_DIR` is not set. This causes a couple of issues:
* Builds can collide with each other
* Generated files can become stale and require cleaning prior to every build
* Build files are not cleaned up upon success nor failure

This creates a temporary directory and cleans it up afterwards. **Note:** this makes it harder to inspect the build directory, because it will now be cleaned up after every run. For this use case I would recommend setting `NVTE_CMAKE_BUILD_DIR="${PWD}/build/cmake" to retain the current behavior. As a result of moving to a temp directory, object files will now not be reused between builds as well.

## Type of change

- [x] Infra/Build change
- [ ] Code refactoring

## Changes

- CMake extension builds are now isolated

# Checklist:


fheinecke:
The failing job is a scheduling timeout for a GPU-specific runner that was having some issues next week. Going to assume that it's unrelated.

ptrendx:
Hmm, this has a potential of being a little disruptive (especially considering that the time needed to build TE is quite long). Could there maybe be some middle ground where e.g. the `-e` builds (develop and in-place) could use the in-place build directory and the regular installation would use the temp instead?

fheinecke:
Good catch - added

ptrendx (approved):

PR #3257, opened 2026-07-25
test(pytorch): cover QuantizedTensor view NotImplementedError

andrewwhitecdw:
Add a unit test for the previously uncovered error path in QuantizedTensor.__torch_dispatch__ when .view() is called on the base QuantizedTensor class.

This complements the fix in #3256 by ensuring the NotImplementedError is raised with the correct, interpolated message.

### DIFF
--- a/tests/pytorch/test_quantized_tensor.py
+++ b/tests/pytorch/test_quantized_tensor.py
@@ -716,7 +716,9 @@ class TestQuantizedTensor:
     def test_view_not_implemented(self) -> None:
         """QuantizedTensor base class does not support tensor views."""
         qt = QuantizedTensor((128, 128), torch.bfloat16)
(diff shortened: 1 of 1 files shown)


andrewwhitecdw:
Acknowledged the code review finding about the `match` pattern dependency.

**Status:** PR #3256 (which adds the missing `f` prefix to the error message) is **already merged** as of 2026-07-27. Verified that `transformer_engine/pytorch/quantized_tensor.py:711` now reads:

```python
raise NotImplementedError(f"{cls.__name__} class does not support tensor views")
```

With the f-string in place, the `pytest.raises(match="QuantizedTensor class does not support tensor views")` pattern correctly matches the interpolated exception message. This test is valid and ready for review.

ptrendx (approved):

PR #3265, opened 2026-07-27
[PyTorch] Add architecture gate to NVFP4 split_quantize RHT path

### LINKED ISSUE #3219: GroupedLinear NVFP4 with RHT fails at runtime on sm_120/sm_121
`GroupedLinear` under the NVFP4 recipe fails on sm_120/sm_121. `split_quantize` with RHT-enabled NVFP4 quantizers dispatches unconditionally to the grouped Hadamard-transform kernels (`nvte_group_hadamard_transform_*`), which are implemented with SM100-family instructions (tcgen05/TMEM) and have no sm_120/121 implementation. There is no architecture check before the dispatch, so the call fails:

```
RuntimeError: .../group_row_cast_col_hadamard_transform_cast_fusion.cu:1276 in function group_row_col_rht_gemm_ntt_w_sfc: CUDA Error: invalid argument
```

The single-tensor path already handles this: `NVFP4Quantizer` checks `is_eligible_for_rht_cast_fusion` (arch band 100..110) and falls back to the generic unfused RHT kernels, so plain `Linear` works on sm_120/121. Only the split/grouped entry point lacks the check.

Repro: script below. GB10 (DGX Spark), sm_121a, CUDA 13.1, driver 580.95.05, TE main @ 9d92fa05.

Proposed fix: add the same arch gate to `split_quantize_nvfp4_impl` and fall back to per-tensor quantize on non-SM100 archs. I have this working on sm_121 hardware and can open a PR.

<details>
<summary>test_split_quantize_rht_sm12x.py</summary>

```python
# Reproducer: split_quantize with RHT-enabled NVFP4 quantizers on sm_120/121
# (the GroupedLinear input/grad-output quantization path).
# Before the fix: dispatches to grouped Hadamard-transform kernels implemented
# with SM100-family instructions -> runtime failure on sm_120/121.
# After the fix: falls back to per-tensor quantization (generic unfused RHT
# path); outputs are well-formed and rowwise dequant matches the input at the
# NVFP4 quantization floor.
import sys
import torch
import transformer_engine.pytorch  # noqa: F401 (loads the torch extension)
from transformer_engine.pytorch.tensor.nvfp4_tensor import NVFP4Quantizer

def main():
    torch.manual_seed(0)
    n_splits, rows_per_split, cols = 4, 128, 256
    x = torch.randn(n_splits * rows_per_split, cols, dtype=torch.bfloat16, device="cuda")
    quantizers = [
        NVFP4Quantizer(rowwise=True, columnwise=True, with_rht=True, with_post_rht_amax=True)
        for _ in range(n_splits)
    ]
    outs = tex.split_quantize(x, [rows_per_split] * n_splits, quantizers)
    torch.cuda.synchronize()

    ok = True
    for i, y in enumerate(outs):
        d = y.dequantize(dtype=torch.float32)
        ref = x[i * rows_per_split:(i + 1) * rows_per_split].float()
        cos = torch.nn.functional.cosine_similarity(
            d.flatten().unsqueeze(0), ref.flatten().unsqueeze(0)
        ).item()
        nr = (d.norm() / (ref.norm() + 1e-30)).item()
        print(f"split {i}: rowwise dequant cos {cos:.6f} norm-ratio {nr:.6f}")
        ok &= cos > 0.98 and abs(nr - 1.0) < 0.05
    print("PASS" if ok else "FAIL")
    return 0 if ok else 1

if __name__ == "__main__":
    try:
        sys.exit(main())
    except Exception as e:
        print(f"ERROR: {type(e).__name__}: {e}")
        sys.exit(2)
```

</details>

davidkny22:
# Description

`GroupedLinear` under the NVFP4 recipe fails at runtime on sm_120/sm_121. `split_quantize` with RHT-enabled quantizers goes straight to `split_quantize_nvfp4_impl_with_rht_helper`, which calls the grouped Hadamard transform kernels. Those are SM100 only, so the launch fails:

```
RuntimeError: .../group_row_cast_col_hadamard_transform_cast_fusion.cu:1276 in function group_row_col_rht_gemm_ntt_w_sfc: CUDA Error: invalid argument
```

The single-tensor path does not have this problem. `NVFP4Quantizer` checks `is_eligible_for_rht_cast_fusion`, which gates on `100 <= sm_arch <= 110`, and uses the unfused RHT kernels otherwise. This adds the same check to `split_quantize_nvfp4_impl` and quantizes the splits per tensor when it fails. That is slower than a fused grouped kernel would be, but it makes the configuration work.

I repeated the band instead of calling `is_eligible_for_rht_cast_fusion` because that function also checks shape alignment, and reusing it would change SM100 behavior. Happy to factor the band into a shared helper if you prefer.

Fixes #3219

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- `transformer_engine/pytorch/csrc/extensions/cast.cpp`: architecture gate and per-tensor fallback in `split_quantize_nvfp4_impl`.
- `tests/pytorch/nvfp4/test_nvfp4_group_quantize.py`: test that RHT split_quantize matches per-tensor quantization, on either dispatch route.

# Checklist:

No documentation changes needed. On the last box, see below.

# Testing done

GB10 (DGX Spark), sm_121a, CUDA 13.1, driver 580.95.05, main @ f1d5f8d5222b built with `NVTE_CUDA_ARCHS=121a`. cpplint, black and clang-format are clean on the changed files.

The new test fails on all three parametrizations without the fix and passes with it. It is not gated on architecture, so it exercises the grouped kernels on SM100 and the fallback elsewhere.

The rest of `tests/pytorch/nvfp4/test_nvfp4_group_quantize.py` goes from 274 failures to 120. The 154 that start passing are the RHT cases with `optimize_for_gemm=False`, which is what this patch covers.

The 120 that remain fail before the patch too, for a different reason. They are all `optimize_for_gemm=True`: swizzled scale factor emission is gated on the same architecture band, so the outputs here carry compact scale factors, but the test swizzles the reference whenever `optimize_for_gemm` is set.

`tests/pytorch/nvfp4/test_nvfp4_rht_quantize_exact.py` is 16 failures and 284 passes both before and after, so the single-tensor path is unaffected. Those 16 are small numeric mismatches already on main.

The reproducer in #3219 fails before the patch and passes after: per-split rowwise dequant cos 0.995463 to 0.995507, norm ratio 0.999135 to 1.001009.

cc @zhongbozhu @cael-ling

### DIFF
--- a/transformer_engine/pytorch/csrc/extensions/cast.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/cast.cpp
@@ -19,6 +19,7 @@
 #include "../extensions.h"
 #include "common.h"
 #include "common/common.h"
+#include "common/util/cuda_runtime.h"
 #include "common/util/system.h"
 #include "pybind.h"
 #include "transformer_engine/multi_tensor.h"
@@ -1630,6 +1631,25 @@ void split_quantize_nvfp4_impl(const TensorWrapper &input,
 
   // CUDA stream
   auto stream = at::cuda::getCurrentCUDAStream();
+
+  // The RHT helper dispatches to grouped Hadamard transform kernels that are
+  // implemented for the SM100 family only. On other architectures, where
+  // NVFP4Quantizer::is_eligible_for_rht_cast_fusion is false as well, quantize
+  // each split on its own instead. That takes the generic unfused RHT path.
+  const int sm = transformer_engine::cuda::sm_arch();
+  const bool grouped_rht_supported = sm >= 100 && sm <= 110;
+  if (quantizer.with_rht && !grouped_rht_supported) {
+    NVTE_CHECK(input.dtype() == DType::kBFloat16, "RHT is only supported for bfloat16 input");
+    NVTE_SCOPED_GIL_RELEASE({
+      for (size_t i = 0; i < num_tensors; ++i) {
+        if (input_list[i].numel() == 0) {
+          continue;
+        }
+        quantizers[i]->quantize(input_list[i], output_list[i], std::nullopt);
+      }
+    });
+    return;
+  }
 
   // Perform multi-tensor quantization
   NVTE_SCOPED_GIL_RELEASE({
--- a/transformer_engine/pytorch/csrc/extensions/cast.cpp
+++ b/transformer_engine/pytorch/csrc/extensions/cast.cpp
@@ -1632,33 +1632,30 @@ void split_quantize_nvfp4_impl(const TensorWrapper &input,
   // CUDA stream
   auto stream = at::cuda::getCurrentCUDAStream();
 
-  // The RHT helper dispatches to grouped Hadamard transform kernels that are
-  // implemented for the SM100 family only. On other architectures, where
+  // The grouped Hadamard transform kernels are implemented for the SM100 family
+  // only. On other architectures, where
   // NVFP4Quantizer::is_eligible_for_rht_cast_fusion is false as well, quantize
   // each split on its own instead. That takes the generic unfused RHT path.
   const int sm = transformer_engine::cuda::sm_arch();
   const bool grouped_rht_supported = sm >= 100 && sm <= 110;
-  if (quantizer.with_rht && !grouped_rht_supported) {
-    NVTE_CHECK(input.dtype() == DType::kBFloat16, "RHT is only supported for bfloat16 input");
-    NVTE_SCOPED_GIL_RELEASE({
-      for (size_t i = 0; i < num_tensors; ++i) {
-        if (input_list[i].numel() == 0) {
-          continue;
-        }
-        quantizers[i]->quantize(input_list[i], output_list[i], std::nullopt);
-      }
-    });
-    return;
-  }
 
   // Perform multi-tensor quantization
   NVTE_SCOPED_GIL_RELEASE({
     if (quantizer.with_rht) {  // Quantize row-wise data, RHT+quantize column-wise data
       // Check that config is supported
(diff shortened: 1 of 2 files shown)


transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -1631,6 +1632,25 @@ void split_quantize_nvfp4_impl(const TensorWrapper &input,
...
+      for (size_t i = 0; i < num_tensors; ++i) {
+        if (input_list[i].numel() == 0) {
+          continue;
+        }
+        quantizers[i]->quantize(input_list[i], output_list[i], std::nullopt);
+      }
+    });
+    return;
+  }
+
```
ptrendx: Could we move this code inside the NVTE_SCOPED_GIL_RELEASE block?

davidkny22:
Moved it inside the existing `NVTE_SCOPED_GIL_RELEASE` block. That also drops the duplicated bfloat16 check and the early return, so the diff is smaller.

Rebuilt on sm_121a: the added test passes, `test_nvfp4_group_quantize.py` is unchanged at 120 failed and 453 passed, and the reproducer still passes.

The failing CI jobs look unrelated to this change. They all stop at `fatbinary fatal: Could not open input file 'flash_attn.compute_75.cubin'` in `common/fused_attn`, and the same failure is on the other open PRs including ones that only touch `common`.

ptrendx (approved):

ptrendx:
There has been some issue with the CI. I retried it on our end.

PR #3267, opened 2026-07-27
Single group mxfp8 grouped mlp

sraman-rgb:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:

### DIFF
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -3558,10 +3558,7 @@ class TestSequentialModules:
         """GroupedLinear + scaled activation + GroupedLinear"""
 
         # Split sizes
-        if group_size == 1:
-            split_sizes = [split_alignment]
-        else:
-            split_sizes = [split_alignment * i for i in range(group_size)]
+        split_sizes = [split_alignment * (i) for i in range(group_size)]
         random.shuffle(split_sizes)
         split_sizes = torch.tensor(split_sizes, dtype=torch.int, device=device)
 
@@ -3751,24 +3748,6 @@ class TestSequentialModules:
             device=device,
             activation="scaled_srelu",
         )
-
-    @pytest.mark.parametrize("bias", (False, True))
-    def test_grouped_mlp_single_group_mxfp8(
-        self,
-        *,
-        bias: bool,
-        device: torch.device = "cuda",
-    ) -> None:
-        """Single-group GroupedLinear + ScaledSwiGLU + GroupedLinear with MXFP8."""
-
-        self.test_grouped_mlp(
-            group_size=1,
-            bias=bias,
-            dtype=torch.bfloat16,
-            quantization="mxfp8",
-            device=device,
-        )
-
 
 class TestCustomOps:
     """Test with ops that are defined externally"""
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -3558,7 +3558,10 @@ class TestSequentialModules:
         """GroupedLinear + scaled activation + GroupedLinear"""
 
         # Split sizes
-        split_sizes = [split_alignment * (i) for i in range(group_size)]
+        if group_size == 1:
+            split_sizes = [split_alignment]
+        else:
+            split_sizes = [split_alignment * i for i in range(group_size)]
         random.shuffle(split_sizes)
         split_sizes = torch.tensor(split_sizes, dtype=torch.int, device=device)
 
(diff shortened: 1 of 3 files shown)


transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -975,24 +1081,81 @@ def fuser_forward(
...
+                    )
+                )
+            (
+                split_sizes,
+                split_points,
+                base_split_offsets,
+                fc1_x_tensor_offsets,
+                fc2_x_tensor_offsets,
+                fc2_out_tensor_offsets,
+            ) = self._single_group_split_metadata_cache[metadata_key]
```
timmoon10: It seems extremely delicate to maintain a cache for the single group case, which should be considered as a temporary hack. Is the overhead of `tex.splits_to_offsets_multi` really bad enough to justify this complexity?

transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -1077,8 +1240,23 @@ def fuser_forward(
         ):
             grouped_fc1_weight._with_gemm_swizzled_scales = False
 
+        # The canonical shared-expert TELinear saves its original BF16 input and
+        # creates the columnwise MXFP8 representation just before FC1 wgrad.
+        # Preserve that policy for the single-group fused path so fprop only
+        # pays for the rowwise representation required by the Rubin kernel.
```
timmoon10: Do we have a canonical shared-expert `TELinear`? Isn't this whole approach a hack to get access to the fused GEMM+activation kernels? I'd say that the canonical implementation is the dense linear forward, where we generate row-wise and column-wise data at the same time. We should justify why we are deferring generating the column-wise data, especially since it will increase memory usage (we have to cache BF16 values instead of FP8/FP4).

transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -975,24 +1081,81 @@ def fuser_forward(
...
-
         # Extract per-row activation probabilities from the middle op.
         scales = basic_op_extra_inputs[1][0]
+        unit_activation_scale = bool(
+            getattr(self.basic_ops[1], "_grouped_mlp_unit_activation_scale", False)
+        )
+        if unit_activation_scale and num_groups != 1:
+            raise ValueError(
+                "Unit activation scaling is only supported for a single-group grouped MLP."
+            )
```
timmoon10: This secret attr should not be loadbearing. It should be safe for us to ignore it, just like it's ignored in the unfused impl.
```suggestion
        unit_activation_scale = bool(
            getattr(self.basic_ops[1], "_grouped_mlp_unit_activation_scale", False)
        )
        if unit_activation_scale and num_groups != 1:
            unit_activation_scale = False
```

tests/pytorch/test_fusible_ops.py
timmoon10, line 1: It would be better to put these fusion-specific tests in `test_grouped_mlp.py`.

timmoon10 (commented):
Mostly fine as a hacky expedient. Remember that the single-group approach is a dead end and we should not put much more effort into it. The right approach is to implement a dense MLP fused op.

transformer_engine/pytorch/ops/fused/grouped_mlp.py
sraman-rgb: yes currently 10% of shared expert GEMM time is swizzle time (GEMM is 40us, and swizzle is 4us)

transformer_engine/pytorch/ops/fused/grouped_mlp.py
sraman-rgb: Sorry, I was referring to the existing dense TELinear shared-expert implementation as the baseline. Like shared expert MLP in MCore creates -> TEColumnParallelLinear / TELinear -> dense GEMM. But that doesnt allow us to use kernel fusion, because only groupedGEMM kernels are integrated. The dense TELinear baseline generates rowwise quant during forward, and saves the entire tensor. I wanted to ensure that it is similar to that for comparison. I wanted to change it, but forgot. Thanks for catching this

tests/pytorch/test_fusible_ops.py
sraman-rgb: changed

vthumbe1503 (dismissed):

vthumbe1503 (approved):

vthumbe1503 (approved):
LGTM. Although I agree with Tim, grouped_mlp has now become grouped_mlp+dense_mlp. We probably should consider migrating it to a seperate fused op soon.

PR #3270, opened 2026-07-28
[Common][PyTorch] EP dispatch with unfused MXFP8 quantization

phu0ngng:
# Description

This PR adds MXFP8 support to the **dispatch op** of the NCCL EP path. The dispatch op is used in two places, and MXFP8 applies to both:
  - **Dispatch forward**  bfloat16 tokens are quantized to MXFP8 internally and dispatched to the target experts; recv is returned as a per-expert `GroupedTensor`.
  - **Combine backward**  the result-grad is scattered back to expert positions through the same (reverse) dispatch op, quantized to MXFP8, returning the expert-output grad as a per-expert `GroupedTensor`.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

#### PyTorch frontend (`transformer_engine/pytorch/ep.py`, `distributed.py`, `csrc/extensions/ep.cpp`)**
  - The dispatch op quantizes bfloat16 tokens to MXFP8 internally when the buffer's `dispatch_quant_recipe` is set (`MXFP8BlockScaling` only for now); dispatch-forward recv is returned as a per-expert `GroupedTensor`. A pre-quantized input is rejected.
  - Combine backward reuses the dispatch op to scatter the result-grad: it quantizes the grad to MXFP8 and returns the expert-output grad as a per-expert `GroupedTensor`. Combine forward is unchanged (high-precision).
  - Recv data and block scales share a single caller-supplied (optionally symm-mem-backed) buffer, sliced into data-then-scale regions; the same convention is used for the combine backward grad buffer.

  #### Common backend (`common/ep/ep_backend.cpp`, `include/.../ep.h`, `comm_window.h`)**
  - Backend and public headers extended to carry block-scale buffers/windows through the dispatch primitive.

  #### NCCL EP submodule**
  - Bumped `3rdparty/nccl-extensions` to the revision providing block-scaled dispatch.

  #### Tests (`tests/cpp_distributed/test_ep.cu`, `tests/pytorch/distributed/run_ep.py`, `run_test_ep.sh`)**
  - Added C++ distributed coverage for the MXFP8 dispatch path.
  - Added PyTorch MXFP8 test passes for dispatch forward (normal, zero-copy, eager IO modes) and combine backward, gated behind a dedicated `NVTE_EP_MXFP8_PASS` run since the grouped path pins the per-expert alignment process-wide.

# Checklist:


transformer_engine/pytorch/ep.py
```diff
@@ -243,6 +250,7 @@ def __init__(
         alignment: int = 0,
         payload_dtype: torch.dtype = torch.bfloat16,
         device: Optional[torch.device] = None,
+        dispatch_quant_recipe: Optional["Recipe"] = None,
```
phu0ngng: Let me separate the recipe for dispatch fwd and combine bwd, in case users want to use different quantization for fwd and bwd in the future.

.gitignore
YangFei1990, line 1: Is this intentional?

transformer_engine/pytorch/ep.py
```diff
@@ -442,32 +474,86 @@ class _EpDispatch(torch.autograd.Function):
...
+        token_counts: Optional[torch.Tensor] = None,
+        rows: int = 0,
+        payload_dtype: torch.dtype = torch.bfloat16,
     ):
-        """Dispatch fwd; prepare must have run into ``handle_mem`` beforehand."""
+        """Dispatch fwd; prepare must have run into ``handle_mem`` beforehand. When scales are set
+        (MXFP8 for now), ``tokens`` is the quantized tensor kept as the autograd operand so grad
+        reaches the pre-quant input. Recv outputs are carved/allocated here: a caller may supply
+        ``recv_tokens`` / ``recv_topk_weights``, else they are sized to ``rows``."""
+        from .quantized_tensor import QuantizedTensor
```
YangFei1990: nit: why not import at top?

transformer_engine/pytorch/ep.py
```diff
@@ -243,6 +251,8 @@ def __init__(
         alignment: int = 0,
         payload_dtype: torch.dtype = torch.bfloat16,
         device: Optional[torch.device] = None,
+        dispatch_fwd_quant_recipe: Optional["Recipe"] = None,
```
YangFei1990, line 266: Do we expect user/mcore to provide a recipe? I think `recipe` is specific concept in TE that we do not want to let end user to touch. Ideally, mcore can provide something dtype/quant config, e.g.
- `ep_dtype`: if we allow single dtype for all
- `dispatch_dtype`, `combine_dtype`: if we want to allow different dtypes for dispatch and combine (you know combine can be tricky)
- `dispatch_fwd_dtype`, `dispatch_bwd_dtype`, `combine_fwd_dtype`, `combine_bwd_dtype`: fine-grained full config

I would recommend the middle option, as combine and dispatch might want to use different recipe, but between dispatch they might want to use same. The arge can take things like `mxfp8`/`nvfp4`, and TE will automatically pick the recipe

transformer_engine/pytorch/ep.py
```diff
@@ -283,6 +295,8 @@ def __init__(
         mark_not_offload(self.total_recv_tokens)
         # Host mirror of total_recv_tokens, set by ep_prepare in eager mode.
         self._host_total_recv_tokens: Optional[int] = None
+        # Set by ep_prepare; dispatch/combine read the routing it cached in handle_mem.
+        self.prepared = False
```
YangFei1990: Seems this is only in the raw functions for assertion? What is the usage?

transformer_engine/pytorch/ep.py
```diff
@@ -578,6 +707,116 @@ def _alloc_io(shape, dtype: torch.dtype, device, zero_copy: bool) -> torch.Tenso
     return torch.empty(*shape, dtype=dtype, device=device)
 
 
+def _quantize(x: torch.Tensor, recipe):
```
YangFei1990: I feel it would be better to find a better home for these quantization functions, seems to me they do not EP specific, or the requirement is implicit.

transformer_engine/pytorch/ep.py
```diff
@@ -442,32 +474,86 @@ class _EpDispatch(torch.autograd.Function):
...
+        hidden = tokens_data.shape[-1]
+        device = tokens_data.device
+        zero_copy = tex.ep_get_zero_copy()
+
+        recv_scale_inv = None
+        if is_scaled:
+            if tokens._fp8_dtype != tex.DType.kFloat8E4M3:
+                raise NotImplementedError("EP dispatch supports only E4M3 MXFP8 tokens for now.")
+            # recv data + scales share one buffer (data then scales); carve or allocate it here.
+            recv_tokens, recv_scale_inv = _scale_alloc_io(
```
YangFei1990, line 508: What if `recv_tokens` is not None in this case? Basically how do we handle user allocated symm buffer with MXFP8 dispatch? Mcore has persistent buffers allocated, would be good to have TE expose the scale allocate APIs for Mcore can call as well.

transformer_engine/pytorch/ep.py
```diff
@@ -587,46 +826,69 @@ def ep_dispatch(
...
-            buffer.zero_copy,
-        )
-    if recv_topk_weights is None:
-        recv_topk_weights = _alloc_io((rows,), torch.float32, buffer.device, buffer.zero_copy)
+
+    tokens_scale_inv = None
+    if buffer.dispatch_fwd_quant_recipe is not None:
+        # The grouped recv's packed scales only match the grouped layout when each expert slot is
+        # 128-aligned.
+        if buffer.alignment <= 0 or buffer.alignment % 128 != 0:
```
YangFei1990: Is this a requirement for NCCL EP? The downstream group gemms have their own asserts

transformer_engine/pytorch/distributed.py
```diff
@@ -1911,6 +1918,41 @@ def _get_symm_mem_pool(device: torch.device, backend: str = "NCCL"):
     return _SYMM_MEM_POOL
 
 
+def release_symm_mem_pool(device: Optional[torch.device] = None) -> None:
```
YangFei1990: This is great, shall we also call it in `ep_finalize`?

transformer_engine/pytorch/distributed.py
```diff
@@ -1911,6 +1918,41 @@ def _get_symm_mem_pool(device: torch.device, backend: str = "NCCL"):
...
+    _SYMM_MEM_POOL_BACKEND = None
+    _SYMM_MEM_POOL_DEVICE = None
+    torch_cached = _SYMM_MEM_POOL_TORCH_CACHED
+    _SYMM_MEM_POOL_TORCH_CACHED = False
+    # A pool from torch's get_mem_pool is also cached in the private module dict
+    # symm_mem._symm_mem_pools; drop that reference so the segments' refcount reaches zero
+    # and their NCCL windows deregister. Fail loudly if this internal has changed shape,
+    # otherwise the windows would silently leak past destroy_process_group().
+    if torch_cached:
+        pools = getattr(symm_mem, "_symm_mem_pools", None)
```
YangFei1990, line 1946: Do we release all the pools, not only the one we created?

transformer_engine/pytorch/ep.py
```diff
@@ -578,6 +707,116 @@ def _alloc_io(shape, dtype: torch.dtype, device, zero_copy: bool) -> torch.Tenso
...
+                f"scaled output buffer too small: need {data_bytes + scale_bytes} bytes "
+                f"(data + scales), got {flat.numel()}."
+            )
+        data = flat[:data_bytes].view(data_dtype).reshape(rows, data_cols)
+        scale_inv = (
+            flat[data_bytes : data_bytes + scale_bytes].view(scale_dtype).reshape(rows, scale_cols)
+        )
+        return data, scale_inv
+    data = _alloc_io((rows, data_cols), data_dtype, device, zero_copy)
+    scale_inv = _alloc_io((rows, scale_cols), scale_dtype, device, zero_copy)
```
YangFei1990, line 785: Currently symm pool based allocation/deallocation will break cuda graph. We should either
- Reuse the current persistent buffer design
- Push PT to fix graph issue and migrate all symm allocation to pool based

YangFei1990 (changes requested):

.gitignore
phu0ngng: Nope. Let me remove this change.

transformer_engine/pytorch/ep.py
phu0ngng: Right now, the design is that when `recv_tokens` is not None, TE will check if the size is big enough for both quantized data and scale, then carve out the buffer for both.

transformer_engine/pytorch/ep.py
phu0ngng: This is not a requirement from NCCL EP. 
You are right that it's the job of the GEMM ops to enforce their alignment requirements.

transformer_engine/pytorch/ep.py
phu0ngng: For assertion only, but actually not very useful, so I will remove it.

transformer_engine/pytorch/ep.py
phu0ngng: Removed the helper and inlined the impl.

transformer_engine/pytorch/distributed.py
phu0ngng: `release_symm_mem_pool` is only required before `destroy_process_group()`. 
User may call `ep_finalize` without `destroy_process_group()`.

transformer_engine/pytorch/distributed.py
phu0ngng: Yes, all tensors on that pool on the current device.

transformer_engine/pytorch/ep.py
phu0ngng: Did 1 + added an assertion to raise an error if symm pool + cuda Graph.

transformer_engine/pytorch/ep.py
```diff
@@ -521,55 +604,104 @@ def forward(  # type: ignore[override]
...
+        if ctx.bwd_quant_recipe is None:
+            grad_expert_out = ctx.grad_out
+            if grad_expert_out is None:
+                grad_expert_out = _alloc_io(
+                    ctx.expert_out_shape, ctx.expert_out_dtype, ctx.device, tex.ep_get_zero_copy()
+                )
+            torch.ops.transformer_engine_ep.combine_bwd(handle_mem, g_result, grad_expert_out)
+        else:
+            if tex.ep_get_zero_copy():
+                raise NotImplementedError(
```
YangFei1990: What is the limitation here?

transformer_engine/pytorch/distributed.py
YangFei1990: Hmm why is that? I did not have it in mcore but it seems working (could be luck)

YangFei1990: Also just see input device is never used, shall we remove?

transformer_engine/pytorch/ep.py
YangFei1990: Discussed offline, decision is to let mcore handle recipe creation at the moment.

transformer_engine/pytorch/distributed.py
phu0ngng: On why
> release_symm_mem_pool is only required before destroy_process_group().

I think `destroy_process_group()` deletes the NCCL communicator that is used to create the symmem, so the cached symmem-s become limbo. 

Are you calling `destroy_process_group()` in MCore? Why?

transformer_engine/pytorch/ep.py
phu0ngng: Stale. Let me remove it.

transformer_engine/pytorch/distributed.py
YangFei1990: Mcore shutdown will call `destroy_process_group`

phu0ngng: Good to know.
Anyway, I found out that `release_symm_mem_pool` is idempotent, so I added it to `ep_finalize` as you suggested.

YangFei1990 (approved):

YangFei1990 (changes requested):
Need to further align on API contracts before merging

YangFei1990 (dismissed):

YangFei1990 (approved):

PR #3271, opened 2026-07-28
[Common] Ensure quantization kernels handle noop properly

kainzhong:
# Description

Fixes a few places where quantization treats the noop flag wrong.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- MXFP8 & Group MXFP8 now ignore noop with `IS_DBIAS` since the reduction kernel will run unconditionally, so if we skip quantization with `IS_DBIAS` the results would be wrong.
- MXFP8's `zero_scales_kernel` now will ignore noop using the same condition as how the quantization ignores it (otherwise if we skip zeroing but not skip quantization the result could be wrong)
- Blockwise FP8's `group_block_scaled_1d_tma_kernel` and `group_block_scaled_2d_tma_kernel` now ignore the noop flag when dbias workspace is nullptr (I suppose this means `IS_DBIAS`) for the same reason
- Pass noop flags to FP8 quantization
- Add tests to check if noop flag is honored

# Checklist:

### DIFF
--- a/tests/pytorch/test_quantized_tensor.py
+++ b/tests/pytorch/test_quantized_tensor.py
@@ -940,9 +940,6 @@ def test_skip_quantization_with_noop_flag(
     Test if the quantization honors the noop flag and skips quantization.
     This test only verifies that the kernel skips quantization where it's expected to do so.
     It doesn't verify otherwise (kernel correctly quantizes when noop is false) since that is already covered by other tests.
-
-    Parametrized over output usage and shape so that every kernel reachable from
-    dispatch/quantize.cuh with a noop flag gets exercised, not just the ones recently fixed.
     """
     if usage == "columnwise" and quantization in ("fp8", "fp8_delayed_scaling"):
         pytest.skip("Delayed scaling does not support columnwise-only quantization")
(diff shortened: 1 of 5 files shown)


tests/pytorch/test_quantized_tensor.py
```diff
@@ -928,3 +928,91 @@ def test_mxfp8_dequantize_columnwise_only_quantized_separately(
...
+def test_skip_quantization_with_noop_flag(
+    quantization: str, usage: str, shape: Tuple[int, int]
+) -> None:
+    """
+    Test if the quantization honors the noop flag and skips quantization.
+    This test only verifies that the kernel skips quantization where it's expected to do so.
+    It doesn't verify otherwise (kernel correctly quantizes when noop is false) since that is already covered by other tests.
+
+    Parametrized over output usage and shape so that every kernel reachable from
+    dispatch/quantize.cuh with a noop flag gets exercised, not just the ones recently fixed.
```
ptrendx: Not needed comment.

tests/pytorch/test_quantized_tensor.py
```diff
@@ -928,3 +928,91 @@ def test_mxfp8_dequantize_columnwise_only_quantized_separately(
...
+    def clone_quantized_output():
+        output_buffers = {}
+        for attr in (
+            "_data",
+            "_transpose",
+            "_rowwise_data",
+            "_columnwise_data",
+            "_scale_inv",
+            "_rowwise_scale_inv",
+            "_columnwise_scale_inv",
```
ptrendx, line 990: amax values should also be included here.

tests/pytorch/test_quantized_tensor.py
kainzhong: Removed

tests/pytorch/test_quantized_tensor.py
kainzhong: Added

ptrendx (approved):

PR #3273, opened 2026-07-29
Minimize the memory usage of the fused cross entropy kernel

ptrendx:
# Description

The changes introduced in #3193, while fixing the numerical issue with the fused cross entropy kernel, introduced additional memory requirement. This PR aims to greatly reduce that additional memory usage while preserving the numerics and optimizing runtime performance.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- The original algorithm computed the full derivative in the forward pass and then passed it to backward pass in order to be multiplied by the incoming gradients. In this PR the thing saved to backward is instead the statistics needed to compute the derivative. This lowers the memory consumption and the number of bytes needed to be read by the backward kernel.
- The new flow contains 2 paths - single-GPU with a single kernel for forward computation and the multi-GPU one with 2 kernels, similar to the existing path.

# Checklist:


tests/pytorch/test_parallel_cross_entropy.py
```diff
@@ -205,8 +210,8 @@ def pack_hook(tensor):
     with torch.autograd.graph.saved_tensors_hooks(pack_hook, lambda tensor: tensor):
         loss = parallel_cross_entropy(logits, target, 0.0, False, None)
 
-    assert len(saved_tensors) == 1
-    assert saved_tensors[0].dtype == torch.float32
+    assert len(saved_tensors) == 4
+    assert saved_tensors[0].dtype == torch.bfloat16
```
ptrendx: Either we should check all of the dtypes or none of them.

transformer_engine/pytorch/cross_entropy.py
```diff
@@ -99,49 +124,62 @@ def parallel_cross_entropy(
...
-    """
-    Cross Entropy loss with optional distributed reduction.
+    """Cross entropy loss with optional distributed reduction.
 
-    The input tensor can be in BF16/FP32, and loss and gradient calculations happen in
-    FP32. The returned loss is always in FP32.
+    Loss and derivative calculations use FP32 arithmetic for BF16 and FP32
+    inputs. Instead of saving a full FP32 derivative, this function saves an
+    input-typed saved-input buffer and per-row FP32 softmax maximum/denominator
+    statistics. The returned loss is FP32.
```
ptrendx: This comment talks about the internals of how the function works, which is not needed in the
user-facing docstrings.

transformer_engine/pytorch/cross_entropy.py
```diff
@@ -99,49 +124,62 @@ def parallel_cross_entropy(
...
-    get equal shards along the V dimension.
+    By default, the saved input is a private contiguous copy and ``inp`` is
+    preserved. This safe mode necessarily has both the caller's input and the
+    copy live during forward. With ``overwrite_input=True``, the original
+    contiguous, non-overlapping input is used as the saved buffer and is
+    overwritten with its gradient during backward. Callers must not read or
+    otherwise reuse that input after starting backward.
+
+    The saved input buffer is consumed by backward, so repeated backward
+    passes on the same result are not supported.
```
ptrendx: This comment has both the information useful to the caller (like the fact that when using
the overwrite input option they cannot reuse the input buffer) as well as the internal
implementation details. It needs to be rewritten to include only information relevant to the
user of the API.

transformer_engine/pytorch/triton/cross_entropy.py
```diff
@@ -30,116 +28,126 @@ def cross_entropy_forward(
...
-
     BLOCK_SIZE = min(MAX_FUSED_SIZE, triton.next_power_of_2(V))
 
-    # unreduced loss
-    loss_1d = torch.zeros(n_rows, dtype=torch.float32, device=_input.device)
-
-    # tensor to hold this rank's m/d/X_y values
-    m_d_X_y = torch.zeros(n_rows * 3, dtype=torch.float32, device=_input.device)
-
+    target = target.contiguous().reshape(-1)
```
ptrendx: The reshape is not needed here and introduces CPU overhead.

janbernloehr:
Thanks for the fix. We validated commit 41157db8 on a downstream H100 TP=1 BF16 workload with 4096 × 128256 logits.
- Baseline: 1.96 GiB FP32 CE allocation, OOM.
- overwrite_input=False: 1002 MiB BF16 allocation, still OOM.
- overwrite_input=True: no additional CE allocation; 5/5 iterations passed.

With destructive reuse enabled, peak memory matched the previous baseline exactly, with no throughput regression. The numerical fix from #3193 remained active.
This suggests that the default safe mode can still OOM for large padded vocabularies. Please consider documenting this distinction and supporting the overwrite_input=True path in downstream callers where the input is not reused after backward. A `release_v2.18` backport would also be valuable for users still on that release.
These results were obtained against 41157db8; since the current PR head has changed, it would be useful to rerun this workload on the latest commit.

pggPL:
Torch.compile support triton kernels and adding overwrite_input=True can lead to silent incorectness under compile.
May be worth to force `no_torch_dynamo` in that case.

pggPL:
Otherwise LGTM

ptrendx:
@pggPL Addressed the torch.compile comment

pggPL (approved):
LGTM

pggPL:
@cspades this will need proper integration in MCore.

PR #3274, opened 2026-07-29
[PyT] Add per-sequence attention policies to packed THD attention

desh2608:
# Description

Add per-sequence attention policies to packed THD attention.

A scalar `attn_mask_type`, `window_size`, and `bottom_right_diagonal` apply one policy to an entire packed invocation. Mixed offline/causal training instead needs individual packed sequences to use different policies while preserving a single caller-facing API.

The new `thd_attention_policies` API accepts an ordered list of complete policies:

```python
policies = [
    {
        "sequence_ids": offline_sequence_ids,
        "mask_type": "padding",
        "window_size": (-1, -1),
        "bottom_right_diagonal": True,  # optional
    },
    {
        "sequence_ids": causal_sequence_ids,
        "mask_type": "padding_causal",
        "window_size": (128, 0),
    },
]

output = transformer_layer(
    hidden_states,
    cu_seqlens_q=cu_seqlens,
    cu_seqlens_kv=cu_seqlens,
    max_seqlen_q=max_seqlen,
    max_seqlen_kv=max_seqlen,
    thd_attention_policies=policies,
    thd_attention_policy_dispatch="auto",
)
```

Every policy must contain `sequence_ids`, `mask_type`, and `window_size`. `bottom_right_diagonal` is optional and falls back to the forward-level or module-level setting. Mask types with explicit causal alignment retain their existing semantics: `padding_causal` forces top-left alignment and `padding_causal_bottom_right` forces bottom-right alignment.

Sequence IDs are one-dimensional, strictly ascending CUDA `int32` or `int64` tensors. The policy tensors must be disjoint and together cover every packed sequence exactly once. Repeated mask types are allowed, so separate sequence subsets can use the same mask with different windows or diagonal alignment. Policy IDs are validated once before CUDA graph capture and then treated as immutable.

The mixed-policy argument is mutually exclusive with scalar `attn_mask_type` and `window_size` arguments.

## API accessibility

Both mixed-policy options are exposed at every relevant PyTorch layer:

```text
TransformerLayer.forward
  -> MultiheadAttention.forward
    -> DotProductAttention.forward
```

- `thd_attention_policies`
- `thd_attention_policy_dispatch`

Callers using `TransformerLayer` or `MultiheadAttention` therefore have the same automatic/grouped control as direct DPA callers.

## Runtime dispatch

`thd_attention_policy_dispatch="auto"` is intentionally padded-first. For each nonempty policy, Transformer Engine queries its ordinary backend selector with the complete runtime `AttentionParams` and `pad_between_seqs=True`:

- If an eligible backend supports that representation, the policy retains the original packed Q/K/V storage and represents sequences owned by other policies as inter-sequence padding.
- Otherwise, the policy's logical Q/K/V tokens are compacted before an ordinary scalar-mask THD attention call and its output is restored to the original physical layout.
- Different policies in one invocation may select different representations; their disjoint full-layout outputs are combined.
- A uniform policy uses the existing scalar fast path in automatic mode.

`thd_attention_policy_dispatch="grouped"` explicitly compacts every policy and bypasses padded-backend probing. The override also applies to a uniform one-policy batch; the scalar fast path does not bypass the requested dispatch.

### Inter-sequence-padding representation

The padding path:

- Keeps Q/K/V in their original packed layout.
- Derives logical and physical cumulative sequence lengths for each policy.
- Calls the existing scalar-mask DPA path with `pad_between_seqs=True`.
- Masks inactive forward lanes before combining policy outputs.
- Uses an identity-forward autograd guard to zero inactive Q/K/V gradient lanes when a backend leaves padding lanes undefined.
- Avoids policy-dependent host synchronization and supports CUDA graph capture.

### Grouped-compaction representation

The grouped path:

- Compacts the logical Q/K/V tokens for each policy.
- Invokes the existing scalar THD attention path with `pad_between_seqs=False`.
- Restores each policy's output to the original physical THD layout.
- Supports unequal Q/KV lengths, pre-existing physical padding, and per-policy windows and diagonal alignment.

## Backend-selection caching

Mixed policies alternate among a small set of `AttentionParams`. Their complete backend-selection results are stored in a bounded cache and reused by the recursive scalar attention calls. The cache accounts for backend-control environment state and is invalidated when ordinary DPA backend selection is marked for refresh.

In a controlled same-node H100 microbenchmark, the steady-state selector calls per mixed forward changed from 4 to 0 for automatic dispatch and from 2 to 0 for grouped dispatch. The directly measured selector CPU time removed was 242.0 us/forward and 169.6 us/forward, respectively. Total host submission time decreased by 170.6 us (8.1%) for automatic dispatch and 109.5 us (5.3%) for grouped dispatch. These are DPA microbenchmark results, not full-model throughput estimates.

## Supported scope

The new API supports:

- Packed THD self-attention and cross-attention.
- Direct DPA, `MultiheadAttention`, and `TransformerLayer` callers.
- FP16 and BF16 forward and backward.
- Mixed and uniform policies, including repeated mask types.
- Padding, causal-padding, and bottom-right causal-padding masks.
- Per-policy sliding windows and diagonal alignment.
- Different Q and KV sequence lengths.
- Existing physical padding between sequences.
- Automatic padded-first routing and an explicit grouped override.
- Empty packed batches with connected zero gradients.
- CUDA graph capture on the supported inter-sequence-padding path.

It is currently restricted to padding mask types and does not support:

- Non-THD layouts.
- Explicit attention masks.
- Attention bias or ALiBi.
- KV caching.
- Score modification.
- FP8 attention or max-logit output.
- Context Parallelism.

## Type of change

- [ ] Documentation change
- [ ] Bug fix
- [x] New feature
- [ ] Breaking change
- [ ] Infra/Build change
- [ ] Code refactoring

## Validation

Focused tests cover:

- FP16/BF16 output and Q/K/V gradient parity against independent scalar-mask attention.
- Mixed, uniform, causal, full-attention, and bottom-right policies.
- Repeated mask types with distinct windows and policy ordering.
- Per-policy diagonal alignment with unequal Q/KV lengths.
- Existing inter-sequence padding.
- Cross-attention with unequal Q/KV lengths.
- Per-policy backend selection and combined padded/grouped outputs.
- Automatic and grouped dispatch parity.
- Grouped override accessibility through DPA, MHA, and TransformerLayer.
- Uniform-policy grouped semantics.
- Backend-selection cache reuse and backend-control invalidation.
- Invalid policy schema, sequence coverage, and scalar/mixed conflicts.
- Empty batches.
- CUDA graph capture and replay on Hopper/FA3.

| Platform | Focused tests |
|---|---:|
| H100 / SM90 | 34 passed |
| A100 / SM80 | 33 passed, 1 expected skip |

The A100 skip is the Hopper/FA3-only CUDA graph case. Python compilation, the repository's Black configuration, and whitespace checks pass for all changed files.

## End-to-end 4B ASR profile

A dedicated one-node/eight-GPU profile ran the 4B/A770M model on ASR-only data for 100 optimizer steps per mode. All-causal, all-full-attention, and 50%-causal modes ran sequentially in the same exclusive allocation on each architecture. The table reports the steady-state mean over iterations 7-100 (`n=94`).

| Hardware | All full attention | All causal | 50% causal | 50% causal vs full |
|---|---:|---:|---:|---:|
| H100, 8 GPUs | 1,585.954 ms | 1,589.284 ms | 1,772.807 ms | +11.782% |
| A100, 8 GPUs | 1,860.790 ms | 1,877.415 ms | 2,110.247 ms | +13.406% |

All-causal and all-full attention are within 1% end to end on both systems. The heterogeneous path is slower because policy grouping and multiple backend invocations are paid in each audio layer. Since A100 and H100 used different cluster-local containers, the cross-architecture ratio should not be interpreted as a pure silicon comparison.

# Checklist

- [x] I have read and followed the contributing guidelines.
- [x] The functionality is complete.
- [x] I have commented the hard-to-understand areas.
- [x] I have updated the public API documentation.
- [x] My changes generate no new warnings in the focused validation.
- [x] I have added tests covering the new behavior.
- [ ] The full repository unit-test suite was run locally.

The focused GPU suites, backend-selection tests, end-to-end 4B profiles, Python compilation, formatting, and whitespace checks pass as described above.


transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1090,6 +1090,99 @@ def get_quantizer_roles(
...
+            group_output = run_scalar_policy(
+                query_layer.index_select(0, group_token_indices),
+                key_layer.index_select(0, group_token_indices),
+                value_layer.index_select(0, group_token_indices),
+                group_cu_seqlens,
+                group_max_seqlen,
+                group_max_seqlen,
+                is_causal=is_causal,
+            )
+            output = output.index_copy(0, group_token_indices, group_output)
```
cyanguwa: The miscellaneous kernels generated in this approach might degrade the overall performance quite a bit. TE supports an option called `pad_between_seqs=True`, which allows attention backends to skip through the padded tokens and only process the "intended" tokens/sequences. In your case, I think we can perform two attention calls, one viewing the causal sequences as the "intended" sequences (and the rest "pad tokens"), and the other one viewing the non-causal sequences as the "intended" sequences. Could you please take a look at this [note](https://github.com/NVIDIA/TransformerEngine/blob/a7aec214eb5c3969984a40c3accb6d66987d8f25/transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py#L1416) and this [test](https://github.com/NVIDIA/TransformerEngine/blob/a7aec214eb5c3969984a40c3accb6d66987d8f25/tests/pytorch/attention/test_attention.py#L1075) see if it makes sense for your use case?

For example, we have a batch of 7 sequences, [111aaa222bb3344cccc], where we would like to perform "padding_causal" on the numbered sequences, {111, 222, 33, 44}, and "padding" on the lettered ones, {aaa, bb, cccc}, we can do something like this:
```
mask_type='padding_causal'
cu_seqlens=[0,3,6,8,10]
cu_seqlens_padded=[0,6,11,13,19]
out_causal = self.forward(
    q, k, v,  # full batch, not split
    qkv_format='thd',
    attn_mask_type=mask_type,
    cu_seqlens_q=cu_seqlens,
    cu_seqlens_kv=cu_seqlens,
    cu_seqlens_q_padded=cu_seqlens_padded,
    cu_seqlens_kv_padded=cu_seqlens_padded,
) # shape [19, h, d] with 0s on pad positions

mask_type='padding'
cu_seqlens=[0,3,5,9]
cu_seqlens_padded=[3,9,15,19]
out_non_causal = self.forward(
    q, k, v,  # full batch, not split
    qkv_format='thd',
    attn_mask_type=mask_type,
    cu_seqlens_q=cu_seqlens,
    cu_seqlens_kv=cu_seqlens,
    cu_seqlens_q_padded=cu_seqlens_padded,
    cu_seqlens_kv_padded=cu_seqlens_padded,
) # shape [19, h, d] with 0s on pad positions

out = out_causal + out_non_causal
```

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1362,8 +1456,147 @@ def forward(
...
+            Per-sequence causal policy for packed THD self-attention. When supplied,
+            it must be a CUDA ``torch.bool`` tensor of shape ``[batch_size]``, where
+            ``batch_size = cu_seqlens_q.numel() - 1``. ``True`` selects causal
+            attention for that sequence and ``False`` selects full-context
+            attention. Transformer Engine groups the two policies and launches the
+            existing scalar-mask attention path once per non-empty group, then
+            restores the original packed token order. It therefore shares the
+            surrounding encoder computation but is not a single fused-attention
+            kernel. Initial support is limited to FP16/BF16 THD self-attention
+            without attention dropout, context parallelism, cacheing, FP8,
```
cyanguwa: Nit: cacheing -> caching

But more importantly, can we rename the argument to something more generic and extensible, for example, `attn_mask_type_per_seq`? We can have the type as a dict, for example, users can pass in {"padding": torch.Tensor, "padding_causal": xxx, "padding_causal_bottom_right": xxx, xxx} to specify which SeqIDs need to run with "padding", and which with "padding_causal", etc. The tensor needs to be on-device.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1362,8 +1456,147 @@ def forward(
...
+                    f"got {thd_sequence_is_causal.numel()} values for "
+                    f"{cu_seqlens_q.numel() - 1} sequences."
+                )
+            if not all(
+                tensor.ndim == 3 and tensor.shape[0] == query_layer.shape[0] for tensor in qkv
+            ):
+                raise ValueError(
+                    "thd_sequence_is_causal requires THD Q, K, and V with the same token count."
+                )
+            if int(cu_seqlens_q[-1].item()) != query_layer.shape[0]:
```
cyanguwa: Please try to avoid GPU-CPU syncs (e.g. `item()`). I think if we pursue the approach suggested above, we can achieve CUDA graph and torch.compile compatibility.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1362,8 +1456,147 @@ def forward(
...
+                )
+            effective_bottom_right_diagonal = (
+                self.bottom_right_diagonal
+                if bottom_right_diagonal is None
+                else bottom_right_diagonal
+            )
+            if effective_bottom_right_diagonal is True:
+                raise ValueError(
+                    "thd_sequence_is_causal requires the standard top-left causal diagonal."
+                )
```
cyanguwa, line 2275: Some of the runtime checks can stay here, but others, regarding what's supported and what's not, can probably go to [get_attention_backend](https://github.com/NVIDIA/TransformerEngine/blob/a7aec214eb5c3969984a40c3accb6d66987d8f25/transformer_engine/pytorch/attention/dot_product_attention/utils.py#L382). For example, the feature only supports THD, non-CP, etc. If you pursue the `out=out_causal+out_non_causal` approach above, the support matrix will expand naturally I think, for example, with self/cross-attention, bottom_right_diagonal=T/F. Also, `pad_between_seqs=T` already has some constraints in `get_attention_backend` which can be reused for this feature.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
cyanguwa, line 2435: The entire `if thd_sequence_is_causal is not None` code block can sit here, to take advantage of the existing mask/window_size checks, as well as be in the `self.prepare_forward_ctx` context for FP8 (if FP8 is ever available for this feature).

desh2608:
@cyanguwa Thanks for your detailed suggestions! I followed the advice for the new implementation in https://github.com/NVIDIA/TransformerEngine/pull/3274/changes/a88c9b64b291bccea2373f15a9bcf90626508afb. However, it seems that `pad_between_seqs` is not supported in FlashAttention 2?

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
desh2608: Done.

desh2608:
@cyanguwa @KshitijLakhani I made the changes following @cyanguwa's suggestions. The main issue is that the `pad_between_seqs` option is only supported for FA3, so I have kept the previous "grouping" implementation for FA2. This design follows some existing arch-based implementations in the repo. Here is a pictorial summary along with some benchmark numbers. Overall, the two implementations achieve loss parity and roughly similar iteration times when used in a real training setup. 

image

cyanguwa:
Thanks for the pictorial summary :) @KshitijLakhani will help guide you through the rest of the PR.

Regarding the benchmarks, should we not expect more savings from `pad_between_seqs` in C, given that FA3 is a bit faster than FA2 and then the `pad_between_seqs` approach should also provide some benefits over `grouped`? Also, you might want to benchmark the FusedAttention backend by the way. It uses cuDNN and supports THD on sm90+, and `pad_between_seqs=True`.

desh2608:
> Also, you might want to benchmark the FusedAttention backend by the way. It uses cuDNN and supports THD on sm90+, and `pad_between_seqs=True`.

I previously ran this comparison on H100: `grouped` averaged 1802.289 ms and `pad-between` averaged 1779.596 ms over ~100 train steps.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1357,6 +1679,8 @@ def forward(
         qkv_layer: Optional[torch.Tensor] = None,
         kv_layer: Optional[torch.Tensor] = None,
         qkv_interleave_dim: int = -3,
+        attn_mask_type_per_seq: Optional[Dict[str, torch.Tensor]] = None,
+        window_size_per_mask_type: Optional[Dict[str, Tuple[int, int]]] = None,
```
KshitijLakhani: The current approach tightly couples the attn_mask_type and the SW size. Is this intentional ? The current structure will never allow two separate "padding" regions to have different SW.  So although the feature is described as “per-sequence policy,” it is currently closer to - One sequence group for each distinct mask type.
I'm recommending : 

```
 per_seq_policies = [
      {
          "sequence_ids": ids_a,
          "mask_type": "padding",
          "window_size": (-1, -1),
      },
      {
          "sequence_ids": ids_b,
          "mask_type": "padding_causal",
          "window_size": (128, 0),
      },
      {
          "sequence_ids": ids_c,
          "mask_type": "padding_causal",
          "window_size": (512, 0),
      },
  ]
```

What do you think @cyanguwa ?

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1676,7 +2052,11 @@ def forward(
             )
 
             # checks for attention mask
-            if attn_mask_type is None:
+            if attn_mask_type_per_seq is not None:
+                # This scalar is used only for shared shape/window validation.
+                # Each policy call below uses its dictionary key.
+                attn_mask_type = next(iter(attn_mask_type_per_seq))
```
KshitijLakhani: This normalizes the global fallback window using whichever policy happens to appear first in dictionary order. Then each policy receives that already-normalized window unless it has an explicit override.
I'm wondering if this can result in errors. For e.g.: 

Suppose the caller provides:
window_size = (2, 2)
attn_mask_type_per_seq = {
      "padding_causal": ids_a,
      "padding": ids_b,
  }
Because the first policy is causal, the shared validation may normalize:
(2, 2) → (2, 0)
The later padding policy then inherits (2, 0) rather than the original (2, 2). That changes its semantics based on dictionary ordering.

KshitijLakhani: I suggest the following flow.

  #### 1. Preserve the unmodified fallback window

  default_window_size = (
      self.window_size if window_size is None else window_size
  )

  Do not normalize this using any one mask type.

  #### 2. Resolve and validate each policy separately

  After normalizing the mask names and window_size_per_mask_type keys, construct a resolved window dictionary:

```
  resolved_window_sizes = {}

  for mask_type in attn_mask_type_per_seq:
      policy_window_size = default_window_size

      if (
          window_size_per_mask_type is not None
          and mask_type in window_size_per_mask_type
      ):
          policy_window_size = window_size_per_mask_type[mask_type]

      resolved_window_sizes[mask_type] = check_set_window_size(
          mask_type,
          policy_window_size,
      )

  This gives every policy its own validated window:

  {
      "padding_causal": (2, 0),
      "padding": (2, 2),
  }
```

  The result no longer depends on dictionary order.

  #### 3. Pass the resolved window to each scalar attention call

  In both mixed-THD implementations—the padding path and grouped/compaction path—the policy call can simply use:
```

  policy_window_size = resolved_window_sizes[mask_type]

  policy_output = self.forward(
      ...,
      attn_mask_type=mask_type,
      window_size=policy_window_size,
      ...
  )
```

The recursive scalar forward() may validate the window again, but that validation will be irrelevant/unchanged because the window has already been normalized for the correct mask type.

  #### 4. Keep the existing scalar logic only for the scalar path
  The ordinary path can retain:

```
  if attn_mask_type is None:
      attn_mask_type = self.attn_mask_type
      attn_mask_type,
      window_size,
  )
```

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1753,6 +2133,80 @@ def forward(
                         seqlens_kv = cu_seqlens_kv[1:] - cu_seqlens_kv[:-1]
                     max_seqlen_kv = int((seqlens_kv.max().item() + 63) // 64 * 64)
 
+            if attn_mask_type_per_seq is not None:
```
KshitijLakhani: Thanks for adding checks from L2136 - 2190 - these are important 
A short comment at the top of this block would help

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
desh2608: Yeah this is a good suggestion. For our current use case, we always need a fixed left context for the causal attention, which is why I initially implemented it this way. But your suggestion should make it more flexible. I will make this change if @cyanguwa agrees.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
desh2608: Good catch. Will change accordingly.

desh2608: This was indeed a bug. I have fixed it in [this commit](https://github.com/NVIDIA/TransformerEngine/pull/3274/changes/fff75e60c2a558c166b483bb8842391eb9efdbb0) and also added a unit test.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
desh2608: However, based on how `window_size` is currently initialized in `DotProductAttention` it seems to be tightly coupled with `attn_mask_type` already?

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
KshitijLakhani: @desh2608 I think this resolves the bug but can still be cleaner. 

Since every attn_mask_type_per_seq key has already been normalized and validated as a padding-family mask, do we still need to assign the first dictionary key to scalar attn_mask_type? It seems cleaner to keep scalar mask validation conditional on attn_mask_type_per_seq is None, and similarly run the THD "padding" in attn_mask_type assertion only for the scalar path. Each mixed-policy recursive call will validate its actual mask type independently. This removes the arbitrary representative key and makes the scalar versus mixed-policy control flow clearer.

### Ordinary scalar-mask path.
 ```
 # Checks for attention mask and sliding window.
  if attn_mask_type_per_seq is None:
      if attn_mask_type is None:
          attn_mask_type = self.attn_mask_type
      else:
          attn_mask_type = attn_mask_type.replace(",", "_")
          if attn_mask_type == "causal_padding":
              attn_mask_type = "padding_causal"

      assert attn_mask_type in AttnMaskTypes, (
          f"Attention mask type {attn_mask_type} is not supported!"
      )

      if window_size is None:
          window_size = self.window_size
      window_size = dpa_utils.check_set_window_size(
          attn_mask_type,
          window_size,
      )

      if bottom_right_diagonal is None:
          bottom_right_diagonal = self.bottom_right_diagonal
      if attn_mask_type in {"causal", "padding_causal"}:
          bottom_right_diagonal = False
      if bottom_right_diagonal is None or attn_mask_type in {
          "causal_bottom_right",
          "padding_causal_bottom_right",
      }:
          bottom_right_diagonal = True
```
###  Mixed-policy path. Mask keys were normalized and validated earlier.
```
  else:
      default_window_size = (
          self.window_size if window_size is None else window_size
      )

      resolved_window_sizes = {}
      for mask_type in attn_mask_type_per_seq:
          policy_window_size = default_window_size
          if window_size_per_mask_type is not None:
              policy_window_size = window_size_per_mask_type.get(
                  mask_type,
                  default_window_size,
              )

          resolved_window_sizes[mask_type] = (
              dpa_utils.check_set_window_size(
                  mask_type,
                  policy_window_size,
              )
          )

      window_size_per_mask_type = resolved_window_sizes
      window_size = default_window_size
```

  Then change the THD mask assertion from:

```
  assert "padding" in attn_mask_type, (
      "Attention mask type must be padding or padding_causal "
      "for qkv_format=thd!"
  )
```

  to:

```
  if attn_mask_type_per_seq is None:
      assert "padding" in attn_mask_type, (
          "Attention mask type must be padding or padding_causal "
          "for qkv_format=thd!"
      )

```


Thoughts on this @desh2608 ?

File: transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py [hunk 8]
```diff
@@ -1323,6 +1337,314 @@ def get_quantizer_roles(
...
+                checkpoint_core_attention=checkpoint_core_attention,
+                fast_zero_fill=fast_zero_fill,
+                pad_between_seqs=pad_between_seqs,
+                bf16_backward=bf16_backward,
+                num_splits=num_splits,
+            )
+        if not nonempty_mask_types:
+            raise ValueError("attn_mask_type_per_seq must assign at least one sequence.")
+
+        if dpa_utils.is_thd_mask_type_padding_supported():
```
KshitijLakhani: This is a bit of problem because it decides the physical representation (padding/grouping) without consulting the complete backend-selection logic. The calling function `_forward_thd_mask_types()` is itself called pretty early in `DPA forward()`, even before the `get_attention_backend()` which is towards the very end, just before `fused`/`flash`/`unfused` attention have been called

KshitijLakhani: This is a bit of problem because it decides the physical representation (padding/grouping) without consulting the complete backend-selection logic. The calling function _forward_thd_mask_types() is itself called pretty early in DPA forward(), even before the get_attention_backend() which is towards the very end, just before fused/flash/unfused attention have been called

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -379,6 +379,22 @@ def _get_fused_attn_backend(
     )
 
 
+def is_thd_mask_type_padding_supported() -> bool:
```
KshitijLakhani: There's a slight issue in this check, I think
So, this function is basically checking for SM90 and FA3 installed and FlashAttention enabled and FA3 enabled. This can produce an inconsistent case. For e.g. :

- FA3 absent, FusedAttention supports padding
This function's gate returns false, so TE uses grouped compaction: pad_between_seqs=False
The recursive call to get_attention_backend() when forward() is called may still select FusedAttention, but it will use FusedAttention on compacted tensors. So it is not that FusedAttention cannot run at all. Rather, TE fails to use FusedAttention’s pad_between_seqs=True capability.

- It can also fail : FA3 installation may pass this function's coarse gate, while the policy’s complete parameters disable FA3 and no other padded backend is available. TE can then fail instead of falling back to grouped execution.

KshitijLakhani: I suggest keeping the current per-policy scalar attention approach, but restructuring the mixed-policy dispatch so that representation selection reuses the existing backend capability checks.

  ### Proposed change


- Remove is_thd_mask_type_padding_supported() as the dispatch gate. It currently treats FA3 availability as a proxy for inter-sequence-padding support, even though FusedAttention may also support pad_between_seqs=True.

- For a mixed batch, resolve each policy’s mask and window, then construct candidate AttentionParams for that policy with:

```
  attn_mask_type=mask_type
  window_size=resolved_window_sizes[mask_type]
  pad_between_seqs=True
```

**_I'd suggest you reuse the existing `get_attention_backend()` for this check rather than introducing a reduced feature-specific selector._**
 
- Partition policies based on that query:

```
padded_policies = {}
grouped_policies = {}

selection = get_attention_backend(padded_params)

if selection_has_backend(selection):
    padded_policies[mask_type] = sequence_ids
else:
    grouped_policies[mask_type] = sequence_ids
```

The outer pad_between_seqs value should not disable this query. It describes the original input layout. When TE preserves only one policy’s sequences in their original positions, the other policies become physical gaps, so that inner policy call correctly uses pad_between_seqs=True.

- Continue reusing the existing helpers for padded and grouped

```
  if padded_policies:
      padded_output = self._forward_thd_mask_types_with_padding(
         ...,
          attn_mask_type_per_seq=padded_policies,
      )
  if grouped_policies:
      grouped_output = self._forward_thd_mask_types_grouped(
          ...,
          attn_mask_type_per_seq=grouped_policies,
      )
```
  The padded helper continues to call scalar attention with:
`  pad_between_seqs=True`
  The grouped helper compacts the selected tokens and calls scalar attention with:
`  pad_between_seqs=False`


 - Combine the two full-sized, disjoint outputs:

```
  return padded_output + grouped_output
```
NOTE: This requires the sequence-ID validation to guarantee exact, disjoint coverage.

- I was thinking a bit about this but I do not think a preliminary pad_between_seqs=False backend query is necessary for the smaller implementation. The grouped helper’s recursive self.forward() will run the normal backend selector and raise the standard error if no compact backend is available.

- I would also rename the helper descriptions from “FA3 path” and “FA2 path” to “inter-sequence-padding path” and “compacted/grouped path,” since either path may ultimately select FusedAttention.

KshitijLakhani (changes requested):

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
desh2608: Discussed offline with @KshitijLakhani and we aligned that we should add this flexibility now that we are allowing a per-sequence policy.

desh2608: Hi @KshitijLakhani and @cyanguwa, I have updated the API as discussed so that it now supports per-sequence mask policy with mask type and context window.

desh2608:
Ran some benchmarking again with the latest changes:
image

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1596,8 +1934,77 @@ def forward(
...
+                    mask_type = "padding_causal"
+                if mask_type not in AttnMaskTypes or "padding" not in mask_type:
+                    raise ValueError(
+                        f"Mixed THD policies support only padding mask types, got {mask_type!r}."
+                    )
+                policy_window_size = dpa_utils.check_set_window_size(
+                    mask_type,
+                    policy["window_size"],
+                )
+                policy_bottom_right_diagonal = bottom_right_diagonal
```
KshitijLakhani: `bottom_right_diagonal` is stored internally per policy, but looks like it is derived from the global argument and mask type and ussers cannot specify it independently per policy. Is my understanding right ?
Now, this may be intentional, particularly because the causal mask names encode much of the alignment behavior. Still, since mask and sliding window are now policy-specific, could you please confirm whether bottom-right alignment can ever legitimately vary between policies. If so, it belongs in the policy object too.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1596,8 +1934,77 @@ def forward(
             interleave sits; must be -3 (e.g. ``bs3hd``) or -2 (e.g. ``bsh3d``,
             Megatron-style). This is an explicit knob rather than shape inference,
             since e.g. ``h == 3`` would make the shapes ambiguous.
+        attn_mask_type_and_window_size_per_seq_policies: Optional[List[Dict[str, Any]]],
```
KshitijLakhani: Choose a more generic name like : `thd_attention_policies` or earlier suggested `attn_policies_per_seq` so that it is not perceived as a policy for swa and attn masks only

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1596,8 +1934,77 @@ def forward(
...
+            default = None
+            Per-sequence policies for packed THD attention. Each list item must contain
+            ``"sequence_ids"``, ``"mask_type"``, and ``"window_size"``. Sequence IDs
+            are supplied as a one-dimensional, strictly ascending CUDA integer tensor;
+            all policy tensors must be disjoint and together contain every sequence
+            exactly once. A list may contain multiple policies with the same mask type
```
KshitijLakhani: Are we validating/verifying these mentioned conditions in the comments anywhere in code?: 
 - Strictly ascending IDs within a policy.
  - Duplicates within one policy.

KshitijLakhani: Since policy metadata is expected to be static, could we validate the sequence IDs once before CUDA graph capture and then treat them as immutable? 
I'm thinking that for each policy, we should check that IDs are strictly ascending, and the sorted concatenation of all policy IDs should equal arange(batch_size). Now the issues is that these checks will synchronize when their result is converted to a Python boolean IIUC, but that should be acceptable as a one-time preparation/warm-up cost. They should not execute inside the captured forward path. A small validation step (or cached validation before capture though need to think a bit through this one's detils) may be the cleanest approach.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1323,6 +1337,329 @@ def get_quantizer_roles(
...
+            )
+        if not nonempty_policies:
+            raise ValueError("Mixed THD policies must assign at least one sequence.")
+
+        padded_policies, grouped_policies = self._partition_thd_mask_policies(
+            nonempty_policies,
+            attention_params_kwargs,
+        )
+        padded_output = None
+        if padded_policies:
```
KshitijLakhani, line 1865: The routing currently treats padded-backend availability as a preference: if FA3 or Fused Attention can run the policy with pad_between_seqs=True, padded execution is always selected. This makes sense and I agree with it as an initial policy, but `get_attention_backend()` only determines backend eligibility for that representation; it does not compare its end-to-end performance against compacting and using FA2/FA4.
It is more of a what is supported call rather than which is more performant call.

So, I'm wondering, to give the users of this feature better guidance, could you please document that _mixed THD dispatch is intentionally padded-first_ .
Users can however, indirectly force grouped execution by disabling FA3 and Fused Attention while leaving FA2/FA4 enabled, but those environment variables affect all attention calls. This is documentation/comment should give enough guidance to the user to maybe experiment the different routes padded and grouped, (assuming they have them available) and then if they find that our approach of choosing padded is less performant for their case they can force TE to use grouped via the env vars.


**Note for the future:** I have not done any benchamarking on this but @desh2608 if you do any benchmarking and we learn that grouped remains materially faster for important workloads, it may be worth providing a direct padded/grouped preference or override rather than relying solely on global env var backend controls.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
```diff
@@ -1323,6 +1337,329 @@ def get_quantizer_roles(
...
+                continue
+            padded_attention_params = dpa_utils.AttentionParams(
+                **attention_params_kwargs,
+                batch_size=policy["sequence_ids"].numel(),
+                attn_mask_type=policy["mask_type"],
+                window_size=policy["window_size"],
+                bottom_right_diagonal=policy["bottom_right_diagonal"],
+                pad_between_seqs=True,
+            )
+            *_, available_backends = dpa_utils.get_attention_backend(padded_attention_params)
```
KshitijLakhani: So seems like for the attention params, we have an uncache probe here, which is unlike what we do otherwise in DPA (see lines 2535 in this file) 
It looks something like this: 
```
                if (
                    _attention_backends["attention_params"] is None
                    or attention_params != _attention_backends["attention_params"]
                ):
```


`_partition_thd_mask_policies()` calls `get_attention_backend()` directly for every policy at line 1359, bypassing the normal `_attention_backends cache` used by the usual forward path. The subsequent recursive `self.forward()` can then perform backend selection again when it executes that policy. Was this repeated selection intentional @desh2608 ?  
I'm concerned  this maybe fine initially, but could you measure the CPU overhead for the latest implementation ? If it is substantial/materical engouth , perhaps the policy probe results could be cached by AttentionParams or reused during execution ? Please evaluate this

tests/pytorch/attention/test_mixed_thd_attention.py
```diff
@@ -0,0 +1,582 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
KshitijLakhani, line 1: Could the new tests reuse more of the existing attention test infrastructur? For e.g. DPA/config construction, THD input generation, backend forcing and comparison helpers?

I am less conflicted about whether the tests remain in a dedicated file or we find an easy/clean way to integrate into the `test_attention` file instead, however, since the feature now has enough cases to justify a separate file, I'm happy to consider that possibility. My main maintenance concern is avoiding duplicated
setup/reference machinery. Trying to focus on mainainability.

While you perform this restructure, please weight the possibility of just having this as a separate function/class in the fused attention test file itself if it relatively clean to do so, else stick wih a separate file and only address the re-use comment i have bove

KshitijLakhani (changes requested):
General comments:
1. The new argument/feature currently appears to be exposed directly through `DotProductAttention`; I do not see corresponding plumbing through `MultiheadAttention` or `TransformerLayer`. Would be good if you could add that too - I'm guessing we wouldn't necessarily want to restrict users of this feature to the `DPA` API only. Ideally, would be good if you could just plumb it in this PR, however, if you are in a rush, I am happy to have this addressed in a separate PR. Please add a TODO so that we do not forget, in case you plan to address it separately

2. I might have missed this but if this PR does not explicitly reject CP for this policy based approach, we should. Users should no try to use it with CP. A very small / lightweight CP rejection test would help add confidence to this rejection code.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
desh2608: I had benchmarked both in my training setup, and the iteration times are as follows:
| Method  | Attn. backend | Iteration time (ms) |
|---------|---------------|---------------------|
| Grouped | FA2           | 1669.9              |
| Grouped | FA3           | **1548.2**              |
| Padded  | FA3           | 1635.7              |

Given that grouped + FA3 seems faster, I will add an optional override so users can select this implementation.

tests/pytorch/attention/test_mixed_thd_attention.py
desh2608: Updated to reuse the existing infrastructure in https://github.com/NVIDIA/TransformerEngine/pull/3274/changes/9f01ef863650617e2c633a2d7c58efedb81c43cf. I kept the new tests in a separate file since there are many cases.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
desh2608: Added in https://github.com/NVIDIA/TransformerEngine/pull/3274/changes/9f01ef863650617e2c633a2d7c58efedb81c43cf along with tests.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
desh2608: Added in https://github.com/NVIDIA/TransformerEngine/pull/3274/commits/2dca0089afe1f509ed59e113421f6bbead14679c

desh2608:
> 2. I might have missed this but if this PR does not explicitly reject CP for this policy based approach, we should. Users should no try to use it with CP. A very small / lightweight CP rejection test would help add confidence to this rejection code.

Added in https://github.com/NVIDIA/TransformerEngine/pull/3274/commits/2dca0089afe1f509ed59e113421f6bbead14679c

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
desh2608: No this was an oversight. Fixed in https://github.com/NVIDIA/TransformerEngine/pull/3274/changes/0ced9fc91e5aef6c39df097a563e5d23591198c3. The CPU overhead was ~200us per call.

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
desh2608: Changed in https://github.com/NVIDIA/TransformerEngine/pull/3274/changes/156069f1d5114a6aa2b32a5bc74f9e47ab133af6

desh2608:
> 1. The new argument/feature currently appears to be exposed directly through `DotProductAttention`; I do not see corresponding plumbing through `MultiheadAttention` or `TransformerLayer`. Would be good if you could add that too - I'm guessing we wouldn't necessarily want to restrict users of this feature to the `DPA` API only. Ideally, would be good if you could just plumb it in this PR, however, if you are in a rush, I am happy to have this addressed in a separate PR. Please add a TODO so that we do not forget, in case you plan to address it separately

Added plumbing in https://github.com/NVIDIA/TransformerEngine/pull/3274/changes/156069f1d5114a6aa2b32a5bc74f9e47ab133af6

transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py
desh2608: For our use cases right now, we don't need policy-specific variation in `bottom_right_diagonal`, but I can imagine in some cross-attention/local attention use cases it may be useful. I will include it as per-policy field in `thd_attention_policies`.

desh2608: Added in https://github.com/NVIDIA/TransformerEngine/pull/3274/changes/d46d82125b90c906459b03f38f366a29b8635b65.

tests/pytorch/attention/test_attention.py
```diff
@@ -1149,7 +1115,47 @@ def test_dpa_qkv_layout_thd_declarative(dtype, model_configs, model, qkv_layout)
     test_dpa_qkv_layout_thd(dtype, model_configs, model, qkv_layout, declarative_packed=True)
 
 
-def _run_dot_product_attention(
+def make_cu_seqlens(seqlens: torch.Tensor) -> torch.Tensor:
```
desh2608, line 1152: I extracted these into public methods so they can be re-used in `test_mixed_thd_attention.py`.

tests/pytorch/attention/test_mixed_thd_attention.py
```diff
@@ -0,0 +1,885 @@
...
+)
+from test_attention import (
+    ModelConfig,
+    compare_and_assert,
+    dtype_tols,
+    make_cu_seqlens,
+    make_dot_product_attention,
+    reset_rng_states,
+    run_dot_product_attention,
+)
```
desh2608, line 22: We are now re-using the same setup as in `test_attention.py`. This file only contains cases specific to mixed attention.

desh2608:
@KshitijLakhani FYI one of the tests was failing on H100 due to a backend selection bug which is now fixed in https://github.com/NVIDIA/TransformerEngine/pull/3274/changes/98f375f0cc2b9c97092919a7db6ed6e6209fc801.

transformer_engine/pytorch/attention/dot_product_attention/utils.py
```diff
@@ -1370,12 +1370,17 @@ def _is_fa3_supported(num_heads, num_gqa_groups, head_dim_qk, head_dim_v, qkv_dt
...
+        use_flash_attention
+        and (window_size[0] != -1 or window_size[1] not in [-1, 0])
+        and not bottom_right_diagonal
+        and max_seqlen_q != max_seqlen_kv
+    ):
+        logger.debug(
+            "Disabling FlashAttention as it only supports sliding window with bottom right"
+            " diagonal alignment for cross-attention"
+        )
+        use_flash_attention = False
```
KshitijLakhani, line 1383: This fix follows the existing convention of using max_seqlen_q != max_seqlen_kv for cross-attention-like cases. For variable-length THD, can individual Q/KV lengths differ while their global maxima are equal? @cyanguwa 
If so, the selector cannot distinguish that case today. I believe that is a broader existing limitation rather than something this PR needs to solve, but it may be worth a follow-up test or issue.

Not blocking the PR on this for now

KshitijLakhani (approved):
LGTM !
Thanks for implementing this feature, Desh and taking it to the finish line :)

PR #3277, opened 2026-07-29
[JAX] EP Dispatch with overflow detection option

phu0ngng:
# Description

Adds a way to size EP recv buffers below the worst case and still stay correct. `ep_prepare` / `ep_dispatch` now expose `total_recv_tokens`, the per-rank pre-drop recv-slot demand, so callers can tell exactly when a step's routing exceeds `recv_capacity_per_rank`; no need to conservatively provision every rank for the theoretical maximum. 

Bootstrap also gains an opt-in`drop_on_overflow` policy so an occasional overflowing step is handled gracefully and keeps running (the default behavior is unchanged). 

Together these let users pick a tighter, cheaper recv capacity and monitor `total_recv_tokens` to confirm it holds or to drive their own capacity policy.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- **Bootstrap:** `ep_bootstrap` gains a `drop_on_overflow` flag, plumbed through `set_ep_bootstrap_params` (pybind) and `NVTEEpGroupConfig.drop_on_overflow`.
  - **New output:** `ep_prepare` now returns `(token_counts, total_recv_tokens, handle_mem)`; `ep_dispatch`'s
    `custom_vjp` primal correspondingly returns `total_recv_tokens` as a non-differentiable output.
  - **MoE:** `moe()` and the Flax `_MoEBlock` now return`(output, aux_loss, total_recv_tokens)`; the extra output is non-differentiable
  - **Teardown:** adds `ep_finalize()` (and `reset_ep_config`) to tear down the EP
    communicator so a process can re-`ep_bootstrap` with a different config.

# Checklist:


transformer_engine/jax/cpp_extensions/ep.py
```diff
@@ -908,7 +919,9 @@ def shardy_sharding_rule(*args):
 
 @_on_collective_stream
 def ep_prepare(cfg: EpLayerConfig, topk_idx):
-    """Exchange routing metadata for ``cfg``; return ``(token_counts, handle_mem)``."""
+    """Exchange routing metadata for ``cfg``; return
+    ``(token_counts, total_recv_tokens, handle_mem)``. ``total_recv_tokens`` is
+    the per-rank pre-drop recv-slot total (includes tokens dropped on overflow)."""
```
jberchtold-nvidia, line 924: When drop-on-overflow aka overflow-detection is disabled, this will now return a "trt" tensor that has an uninitialized value, right? Can we instead update this so `ep_prepare` will return `(token_counts, None, handle_mem)` instead when drop-on-overflow=False?

phu0ngng: No, it actually has the `total_recv_tokens` value i.e., the sum of token_counts. 

The only difference here is that when `drop-on-overflow=False` (which is the default), as soon as there is an overflow, the kernel will trap and the program will crash.

jberchtold-nvidia: Oh I see, then the current implementation looks good. The PR LGTM once CI passes

jberchtold-nvidia (approved):

PR #3281, opened 2026-07-29
[Common] Fix NVFP4 stochastic rounding on architectures without cvt.rs

davidkny22:
# Description

NVFP4 quantization with stochastic rounding returns all-zero data on sm_120 and sm_121.

The four e2m1 stochastic rounding helpers gate on `ARCH_HAS_STOCHASTIC_ROUNDING`, which is `ArchSpecific<100>` and `ArchSpecific<103>`, matching where the `cvt.rs` instruction exists. The other branch is `NVTE_DEVICE_ERROR`, which is `printf` plus `assert(0)`. Release builds define `NDEBUG`, so the assert is compiled out, the helper returns its zero-initialized output, and each thread prints:

```
.../common/util/ptx.cuh:953 in function mul_cvt_bf16_to_fp4_8x_stochastic_rounding (thread (0,0,0), block (1,0,0)): FP4 cvt PTX instructions are architecture-specific. Try recompiling with sm_XXXa instead of sm_XXX.
```

`NVFP4BlockScaling` sets `stochastic_rounding` on gradient quantization by default and `_compute_nvfp4_support()` reports NVFP4 for every device at compute capability 10.0 or above, so on these parts the default recipe trains on zeroed gradients while the NVFP4 GEMMs themselves work. The message is also not actionable here: the build already used `NVTE_CUDA_ARCHS=121a`, and `cvt.rs` does not exist at any feature level on this family.

This replaces that branch with a software e2m1 stochastic rounder. It takes 8 random bits per element, which is what `cvt.rs.satfinite.e2m1x4.f32` takes from its `rbits` operand, adds `u * step` where `step` is the grid step at that magnitude, and truncates onto the grid, so a value rounds up with probability equal to its fractional position between its two neighbours. The `if constexpr` is unchanged, so architectures with `cvt.rs` keep the asm path and nothing else in the helpers moves.

The edges follow the PTX ISA text for `.satfinite`, under `cvt`, Floating Point Notes: for `.e2m1x4` destinations, "NaN results are converted to positive MAX_NORM", and an input above MAX_NORM in absolute value gives "sign-preserved MAX_NORM of the destination format". So NaN gives +6 and overflow gives ±6.

A fallback rather than refusing stochastic rounding on these architectures: `_compute_nvfp4_support()` already advertises NVFP4 here and the default recipe turns SR on, so a capability gate would replace a silent wrong answer with a hard failure of the default recipe on hardware that otherwise runs it. The fallback keeps recipe parity.

The fallback is not bit identical to `cvt.rs` and cannot be. `cvt.rs` adds `rbits` to the mantissa bits the conversion discards and rounds on the carry out; this adds a scaled uniform to the value and truncates. Both are stochastic rounding at the same per-element entropy, and their round-up probabilities agree to the 1/256 that 8 bits resolve. There is no `cvt.rs` behaviour on sm_120 or sm_121 to differ from, and stochastic rounding is not reproducible across architectures in any case.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- `transformer_engine/common/util/ptx.cuh`: add `stochastic_round_fp4_e2m1` and use it in the non-`cvt.rs` branch of `mul_cvt_bf16_to_fp4_4x_with_stochastic_rounding`, `mul_cvt_fp32_to_fp4_4x_with_stochastic_rounding` and `mul_cvt_bf16_to_fp4_8x_stochastic_rounding`.
- `transformer_engine/common/transpose/quantize_transpose_vector_blockwise_fp4.cu`: the same in `cvt_fp32_to_fp4_4x_with_stochastic_rounding`.
- `tests/pytorch/nvfp4/test_nvfp4_sr_quantize.py`: test that every stochastically rounded value sits on one of the two E2M1 grid values that bracket the exact one, and that two draws differ.

# Checklist:

No documentation changes needed. On the last box, see below.

# Testing done

GB10 (DGX Spark), sm_121a, CUDA 13.1, driver 580.95.05, main @ 0ee1320e84f0, built with `NVTE_CUDA_ARCHS=121a`. clang-format 18.1.6, cpplint 1.6.0 and black 24.4.2 are clean on the changed files, and there are no compiler warnings in the build.

The new test is not gated on architecture, so it exercises `cvt.rs` where the instruction exists and this branch here. All four parametrizations fail without the patch and pass with it, on separate unpatched and patched builds of the same tree, twice each. Together they reach three of the four helpers: `cvt_fp32_to_fp4_4x_with_stochastic_rounding` for FP32 input, `mul_cvt_bf16_to_fp4_8x_stochastic_rounding` for BF16 with 1D scaling, and `mul_cvt_bf16_to_fp4_4x_with_stochastic_rounding` for BF16 with 2D. I could not reach `mul_cvt_fp32_to_fp4_4x_with_stochastic_rounding` from any NVFP4 entry point on this hardware; it is the FP32 twin of the same columnwise loop and is fixed the same way.

`tests/pytorch/nvfp4/test_nvfp4_sr_quantize.py` is 6 failed, 22 passed, 4 skipped with the patch, in 37 seconds. Without it the same file does not finish: I stopped it at 61 minutes, since the error path prints once per thread and this file quantizes 8192x8192 tensors 50 times per case.

The 6 that fail are `test_group_stochastic_rounding_quantization_versus_reference` with `use_tex_split_quantize=True`, and they are not this path. They fail at `group_row_cast_col_hadamard_transform_cast_fusion.cu:1276 in function group_row_col_rht_gemm_ntt_w_sfc: CUDA Error: invalid argument`, which is the grouped RHT dispatch reported in #3219. `tex.split_quantize` with RHT-enabled NVFP4 quantizers raises the same error on an unpatched build of this tree.

On one 256x256 BF16 tensor, dequantized cosine against the input is 0.000000 before and 0.990767 after, against 0.995495 for round to nearest on the same tensor. Stochastic rounding sitting slightly below round to nearest on a single draw is the expected direction.

If there is anything else you want checked on sm_121, I can run it.

cc @Oleg-Goncharov @ptrendx

### DIFF
--- a/transformer_engine/common/util/ptx.cuh
+++ b/transformer_engine/common/util/ptx.cuh
@@ -600,6 +600,35 @@ __device__ __forceinline__ void mul_cvt_4x(fp4e2m1x4 &out, const Tx2 &in01, cons
   out = fp4e2m1x4(make_float4(x0, x1, x2, x3));
 }
 
+// Software stochastic rounding onto the e2m1 grid, for architectures without
+// cvt.rs. Takes 8 random bits per element, which is what
+// cvt.rs.satfinite.e2m1x4.f32 takes from its rbits operand. Follows satfinite
+// for the edges: NaN becomes positive MAX_NORM and larger magnitudes clamp to
+// MAX_NORM with the sign kept. The result is exactly representable in e2m1, so
+// packing it is lossless.
+__device__ __forceinline__ float stochastic_round_fp4_e2m1(const float x, const uint32_t rbits8) {
+  constexpr float max_norm = 6.0f;
+  if (isnan(x)) {
+    return max_norm;
+  }
+  const float u = static_cast<float>(rbits8 & 0xFFu) * (1.0f / 256.0f);
+  const float a = fabsf(x);
+  // Grid step at |x|: 0.5 below 2, 1 in [2, 4), 2 in [4, 6].
+  const float step = (a >= 4.0f) ? 2.0f : ((a >= 2.0f) ? 1.0f : 0.5f);
+  const float t = fmaf(u, step, a);
+  float q;
+  if (t >= max_norm) {
+    q = max_norm;
+  } else if (t >= 4.0f) {
+    q = 4.0f;
+  } else if (t >= 2.0f) {
+    q = 2.0f + floorf(t - 2.0f);
+  } else {
+    q = floorf(t * 2.0f) * 0.5f;
+  }
+  return copysignf(q, x);
+}
+
 __device__ __forceinline__ fp4e2m1x4 mul_cvt_bf16_to_fp4_4x_with_stochastic_rounding(
     const uint64_t in_4x, const float2 scale, const uint32_t rbits) {
   uint16_t out_4x = 0;
@@ -633,9 +662,14 @@ __device__ __forceinline__ fp4e2m1x4 mul_cvt_bf16_to_fp4_4x_with_stochastic_roun
         : "=h"(out_4x)
         : "l"(in_4x), "l"(reinterpret_cast<const uint64_t &>(scale)), "r"(rbits));
   } else {
-    NVTE_DEVICE_ERROR(
-        "FP4 cvt PTX instructions are architecture-specific. "
-        "Try recompiling with sm_XXXa instead of sm_XXX.");
+    // mul.f32x2 above applies scale.x to the even elements and scale.y to the odd ones.
+    const bf16 *vals = reinterpret_cast<const bf16 *>(&in_4x);
+    const float q0 = stochastic_round_fp4_e2m1(static_cast<float>(vals[0]) * scale.x, rbits);
+    const float q1 = stochastic_round_fp4_e2m1(static_cast<float>(vals[1]) * scale.y, rbits >> 8);
+    const float q2 = stochastic_round_fp4_e2m1(static_cast<float>(vals[2]) * scale.x, rbits >> 16);
+    const float q3 = stochastic_round_fp4_e2m1(static_cast<float>(vals[3]) * scale.y, rbits >> 24);
+    const fp4e2m1x4 packed(make_float4(q0, q1, q2, q3));
+    out_4x = *reinterpret_cast<const uint16_t *>(&packed);
   }
   return *reinterpret_cast<fp4e2m1x4 *>(&out_4x);
 }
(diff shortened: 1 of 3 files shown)


ptrendx:
Thanks @davidkny22 for the contribution! I will take a look either today or early next week.

bvolpato (approved):
Thank you for addressing this. I built this exact revision for `sm_120` and tested it on an RTX 5070 Ti. With current main, default `NVFP4BlockScaling()` produced finite output but zero input and weight gradients. With this PR, both gradients are fully nonzero, architecture-specific error spam is gone, and empirical stochastic-rounding distributions match expected 8-bit probabilities.

One non-blocking issue remains on SM120: fused `split_quantize` + RHT still fails all six grouped cases with `CUDA Error: invalid argument` in `group_row_col_rht_gemm_ntt_w_sfc`. Same failure reproduces on current main and occurs during round-to-nearest setup, before stochastic rounding, so it does not appear introduced by this change and should not block approval. Still, this patch does not make every NVFP4 path healthy on SM120; grouped/RHT likely needs follow-up.

davidkny22:
Thanks for testing this on hardware I don't have!

That grouped failure is a different bug and #3265 fixes it. Same signature as #3219: `split_quantize` hands RHT quantizers to the grouped Hadamard kernels, which are SM100 only, so the launch fails. The fix gates on `100 <= sm_arch <= 110` and falls back to per-tensor, so your 5070 Ti would take the same route sm_121 does. It's approved, just stuck behind broken CI.

If you do apply it, I'd love to know whether those six cases pass on sm_120.

ptrendx:
@davidkny22 Why didn't you go for trying to emulate the cvt.rs instruction exactly? Performance-wise it should actually be similar I believe.

davidkny22:
I went with the float-domain add-and-truncate because it reuses the grid-step math already in the file for round-to-nearest, instead of writing a second bit-level path that has to handle bf16 vs fp32 mantissa widths and the different truncation boundary per magnitude band (step doubles at 2 and 4). Felt like more places for a subtle bug for the same statistical result. Also there's no `cvt.rs` on sm_120/121 to match bit-for-bit anyway, so exact could only mean matching the algorithm, not any real output on this hardware.

Happy to build the bit-carry version if there's a reason you want it.

transformer_engine/common/util/ptx.cuh
```diff
@@ -600,6 +600,35 @@ __device__ __forceinline__ void mul_cvt_4x(fp4e2m1x4 &out, const Tx2 &in01, cons
...
+  float q;
+  if (t >= max_norm) {
+    q = max_norm;
+  } else if (t >= 4.0f) {
+    q = 4.0f;
+  } else if (t >= 2.0f) {
+    q = 2.0f + floorf(t - 2.0f);
+  } else {
+    q = floorf(t * 2.0f) * 0.5f;
+  }
```
ptrendx: This probably would be faster if rewritten as branchless code, could you try it @davidkny22?

davidkny22: Happy to. Rewrote it branchless in 6c93e41e. I ran an exhaustive sweep covering all 2^32 float bit patterns (every NaN payload, infinity, and denormal included) at 16 rbits values, plus every float with magnitude in [1, 8) at all 256 rbits values, and confirmed the output is bit-identical to the previous version. In an isolated microbenchmark on sm_121 it runs 1.34x faster.

ptrendx (approved):

PR #3296, opened 2026-07-31
[PTQ] Store FP32 global scaling factors (absmax or scale_inv) for all quantized activations and weights.

cspades:
# Description

<table>
  <tr>
    <td align="center" width="25%"><b>agieval</b><br/>agieval</td>
    <td align="center" width="25%"><b>arc_challenge</b><br/>arc_challenge</td>
    <td align="center" width="25%"><b>arc_easy</b><br/>arc_easy</td>
    <td align="center" width="25%"><b>gsm8k</b><br/>gsm8k</td>
  </tr>
  <tr>
    <td align="center" width="25%"><b>humaneval</b><br/>humaneval</td>
    <td align="center" width="25%"><b>lambada_standard</b><br/>lambada_standard</td>
    <td align="center" width="25%"><b>mmlu</b><br/>mmlu</td>
    <td align="center" width="25%"><b>winogrande</b><br/>winogrande</td>
  </tr>
</table>

Source: https://github.com/NVIDIA/Megatron-LM/issues/5660#issuecomment-5581231117

See issue for context: https://github.com/NVIDIA/Megatron-LM/issues/5660
Related to: https://github.com/NVIDIA/Megatron-LM/pull/6183

- Extended the `Quantizer.calibrate()` API to support decayed calibration and model buffering.
  - The public API `autocast(calibrating: bool, calibration_config: QuantizationCalibrationConfig)` (both arguments, kept the older one for legacy / backwards compatibility with decay defaulting to 0) turns on the feature, and by default it is deactivated.
  - `calibrate` will first look for calibration metadata according to the quantization recipe, but if that doesn't exist yet it will just take the non-quantized `Tensor` argument and compute its quantization metadata, _**since this is the original behavior of `calibrate`**_.
    - It is super un-performant to recompute the quantization metadata, usually it is fused into a TE kernel. So during training and inference, this path where we cannot find the quantization metadata should never be run.
  - `calibrate` now has implementations for all of the recipes that use FP32 global scaling factors, and the decayed calibration update is implemented in the base class `Quantizer._update_calibration_value`. It'll also buffer them automatically, which is a change in behavior but the extra buffering feature isn't a big deal for performance/memory (and frankly what's the point of storing calibration data natively in TE if we never export it), refer to my results below!
- To enable users to mine PTQ calibration data from their pre-training datasets, buffer scaling factors for quantized activations and weights for checkpointing in Megatron and inference in vLLM, Tensor-RT, ModelOpt, etc.
  - This is an exported common state artifact, it can be saved but (should) not (be) loaded, and will not appear in the model state dict. No risk of surprise overwriting of weight scaling factors, for instance.
  - Buffer names are linked to TE modules and are named pretty comprehensively: `f"{tensor_name}_tensor_{metadata_name}_{recipe}_te_ptq_calibrated"`
- `QuantizationCalibrationConfig` only has a single argument at the moment but in the future could hypothetically have way more, it's just that now we only buffer FP32 global scaling factors so it's simply just `activation_scale_decay`.
  - `activation_scale_decay` controls the decay of past max-accumulated activation scaling factors when your model or dataset is in-flight during training. Intends to capture steady-state maxima.
    - `New AbsMax = Max(Old AbsMax * Decay, Observed AbsMax)`
    - Setting it to 0 implies only taking the most recent scaling factor, while setting it to 1 implies taking an absmax over the entire history of your model and/or dataset.
    - You can do quite a lot of hacky customization on the MCore side, i.e. setting LR=0 means you're basically just calibrating on your calibration dataset with a frozen model.
  - Tiny performance loss to compute the accumulated max, but almost unnoticeable for both dense and MoE (GroupedLinear).
  - Tiny memory overhead to store the accumulated activation scaling factors. Since it's just 4 bytes per activation (FP32 global scale), it's not a lot at all.
    - Weights do not need to be accumulated, so the weight buffer points to the literal row-wise scaling factor in the quantized tensor storage and doesn't take any memory. (Set row-wise usage to True, on the user to do it.)
  - We do not save blockwise scaling factors (or `CustomRecipe` and `HybridQuantizer`, which will error out), those are usually computed during inference or using more advanced calibration techniques.
    - We do store the row-wise NVFP4 scaling factors, but in MCore I just max-reduce it to a scalar as well.
- `TEAutocastState` is another new data-class that may concern you since it changes the API of `get_autocast_state` and `set_autocast_state` which are public interfaces for the global quant state.
  - That being said, currently Megatron-LM and other NeMo repositories are not directly using this, and it gets really messy with checkpointing if I have to reset `QuantizationCalibrationConfig` so I think now's a good time to stop returning `tuple`s...

## Testing

- Really basic unit tests, tell me if you want to see more.
- Performance is only affected when decay is activated, and memory is only increased when buffering activations.
```
# Main / No scaling factor checkpointing.

[2026-07-27 13:24:57.339692] iteration       16/15258789 | consumed samples:         2048 | elapsed time per iteration (ms): 6282.9 | throughput per GPU (TFLOP/s/GPU): 2147.5 | learning rate: 7.864316E-08 | global batch size:   128 | lm loss: 1.207198E+01 | loss scale: 1.0 | grad norm: 20.134 | number of skipped iterations:   0 |number of nan iterations:   0 |

[Rank 0] (after 2 iterations) memory (MB) | allocated: 69607.20 | max allocated: 99633.89 | reserved: 72394.00| max reserved: 100310.00

# Scaling factor & absmax checkpointing. (--buffer-transformer-engine-calibration-metadata)

[2026-07-27 13:45:56.610449] iteration       16/15258789 | consumed samples:         2048 | elapsed time per iteration (ms): 6282.0 | throughput per GPU (TFLOP/s/GPU): 2147.8 | learning rate: 7.864316E-08 | global batch size:   128 | lm loss: 1.207239E+01 | loss scale: 1.0 | grad norm: 20.133 | number of skipped iterations:   0 | number of nan iterations:   0 |

[Rank 0] (after 2 iterations) memory (MB) | allocated: 69607.26 | max allocated: 99633.94 | reserved: 72394.00 | max reserved: 100310.00

# Decaying Activation Scaling Factors & Absmax (--transformer-engine-calibration-decay 0.99)

[2026-07-27 13:51:03.183943] iteration       16/15258789 | consumed samples:         2048 | elapsed time per iteration (ms): 6486.2 | throughput per GPU (TFLOP/s/GPU): 2080.2 | learning rate: 7.864316E-08 | global batch size:   128 | lm loss: 1.207229E+01 | loss scale: 1.0 | grad norm: 20.133 | number of skipped iterations:   0 |number of nan iterations:   0 |

[Rank 0] (after 2 iterations) memory (MB) | allocated: 69607.32 | max allocated: 99634.00 | reserved: 72396.00| max reserved: 100310.00
```
For future reference, I also considered buffering MXFP8 / NVFP4 / FP8-Blockwise factors, but there is a memory and massive performance hit, and generally inference computes these dynamically:
```
# Decaying Activation Scaling Factors + Blockwise Activation Scales (i.e. `_rowwise_scale_inv`)
# (Just FYI why we don't want to export blockwise scales, they significantly affect perf and memory, and aren't really necessary for calibration.)

[2026-07-27 13:57:09.949572] iteration       16/15258789 | consumed samples:         2048 | elapsed time per iteration (ms): 6857.8 | throughput per GPU (TFLOP/s/GPU): 1967.5 | learning rate: 7.864316E-08 | global batch size:   128 | lm loss: 1.207223E+01 | loss scale: 1.0 | grad norm: 20.134 | number of skipped iterations:   0 |number of nan iterations:   0 |

[Rank 0] (after 2 iterations) memory (MB) | allocated: 71158.36 | max allocated: 101184.57 | reserved: 74394.00 | max reserved: 101934.00 
```

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/examples/pytorch/mnist/main.py
+++ b/examples/pytorch/mnist/main.py
@@ -76,7 +76,10 @@ def calibrate(model, device, test_loader, fp8):
     with torch.no_grad():
         for data, target in test_loader:
             data, target = data.to(device), target.to(device)
-            with te.autocast(enabled=fp8, calibrating=True):
+            with te.autocast(
+                enabled=fp8,
+                calibration_config=te.QuantizationCalibrationConfig(),
+            ):
                 output = model(data)
 
 
--- a/transformer_engine/pytorch/module/_common.py
+++ b/transformer_engine/pytorch/module/_common.py
@@ -63,7 +63,7 @@ def _update_scale_buffers(
     for buffer_name, scale in scale_updates.items():
         if scale is None:
             continue
-        if buffer_name.startswith("input"):
+        if activation_scale_decay > 0.0:
             observed_scale = scale.detach().float()
             scale_buffer = scale_buffers.get(buffer_name)
             if scale_buffer is not None and scale_buffer.shape != observed_scale.shape:
@@ -71,10 +71,6 @@ def _update_scale_buffers(
                     "Quantized scaling-factor buffer shape changed from "
                     f"{tuple(scale_buffer.shape)} to {tuple(observed_scale.shape)}"
                 )
-            if activation_scale_decay == 0.0:
-                # If not using scale decay, just buffer the current scaling.
-                scale_buffers[buffer_name] = observed_scale
-                continue
             if scale_buffer is None:
                 # Initialize the rolling activation scaling factor.
                 # Requires CUDA graph warmup step.
@@ -89,8 +85,8 @@ def _update_scale_buffers(
                 out=scale_buffer,
             )
         else:
-            # Keep a reference to the current weight metadata without
-            # allocating or copying a separate buffer.
+            # Without scale history, keep a reference to the current metadata
+            # without allocating or copying a separate buffer.
             # Requires CUDA graph warmup step.
             scale_buffers[buffer_name] = scale.detach()
 
--- a/transformer_engine/pytorch/module/_common.py
+++ b/transformer_engine/pytorch/module/_common.py
@@ -6,14 +6,79 @@
 
 import dataclasses
 import queue
-from typing import Any, Callable, List, Optional, Tuple, Union
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import torch
 
 from .. import cpp_extensions as tex
 from ..constants import TE_DType
 from ..export import is_in_onnx_export_mode
+from ..tensor.utils import get_quantization_recipe_name
 from ..utils import get_default_init_method
+
+
+def _get_scale_buffer_info(
+    tensor_name: str,
+    tensor: Any,
+    quantizer: Any,
+) -> Optional[Tuple[str, Optional[torch.Tensor]]]:
+    """Get the calibration buffer name and value for a quantized tensor."""
+    recipe = get_quantization_recipe_name(quantizer)
+    if not recipe:
+        return None
+    if recipe == "fp8_current_scaling":
+        metadata_name = "scale_inv"
+        metadata = getattr(tensor, "_scale_inv", None)
+    else:
+        metadata_name = "amax_rowwise"
+        metadata = getattr(
+            tensor,
+            "_amax_rowwise",
+            getattr(quantizer, "amax", None),
+        )
+    buffer_name = f"{tensor_name}_tensor_{metadata_name}_{recipe}_te_ptq_calibrated"
+    return buffer_name, metadata
+
+
+def _update_scale_buffers(
+    scale_buffers: Dict[str, Optional[torch.Tensor]],
+    scale_updates: Dict[str, Optional[torch.Tensor]],
+    activation_scale_decay: float = 0.0,
+) -> None:
+    """Merge observed scaling factors into checkpoint buffers."""
+    for buffer_name, scale in scale_updates.items():
+        if scale is None:
+            continue
+        if buffer_name.startswith("input"):
+            observed_scale = scale.detach().float()
+            scale_buffer = scale_buffers.get(buffer_name)
+            if scale_buffer is not None and scale_buffer.shape != observed_scale.shape:
+                raise RuntimeError(
+                    "Quantized scaling-factor buffer shape changed from "
+                    f"{tuple(scale_buffer.shape)} to {tuple(observed_scale.shape)}"
+                )
+            if activation_scale_decay == 0.0:
+                # If not using scale decay, just buffer the current scaling.
+                scale_buffers[buffer_name] = observed_scale
+                continue
+            if scale_buffer is None:
+                # Initialize the rolling activation scaling factor.
+                # Requires CUDA graph warmup step.
+                scale_buffer = torch.zeros_like(observed_scale)
+                scale_buffers[buffer_name] = scale_buffer
+            # Track a decaying maximum so early-training activation
+            # outliers do not permanently determine the inference scale.
+            scale_buffer.mul_(activation_scale_decay)
+            torch.maximum(
+                scale_buffer,
+                observed_scale,
+                out=scale_buffer,
+            )
+        else:
+            # Keep a reference to the current weight metadata without
+            # allocating or copying a separate buffer.
+            # Requires CUDA graph warmup step.
+            scale_buffers[buffer_name] = scale.detach()
 
 
 def set_quantizer_amax_reduction_group(quantizer, amax_reduction_group) -> None:
--- a/transformer_engine/pytorch/module/_common.py
+++ b/transformer_engine/pytorch/module/_common.py
@@ -26,16 +26,30 @@ def _get_scale_buffer_info(
     recipe = get_quantization_recipe_name(quantizer)
     if not recipe:
         return None
-    if recipe == "fp8_current_scaling":
+
+    if recipe == "fp8_delayed_scaling":
+        metadata_name = "amax"
+        metadata = getattr(quantizer, "amax", None)
+    elif recipe == "fp8_current_scaling":
         metadata_name = "scale_inv"
         metadata = getattr(tensor, "_scale_inv", None)
-    else:
+    elif recipe == "nvfp4":
+        metadata_name = "amax"
+        metadata = getattr(tensor, "_amax_rowwise", None)
+    elif recipe == "nvfp4_rowwise":
         metadata_name = "amax_rowwise"
-        metadata = getattr(
-            tensor,
-            "_amax_rowwise",
-            getattr(quantizer, "amax", None),
-        )
+        metadata = getattr(tensor, "_amax_rowwise", None)
+    elif recipe == "mxfp8":
+        # MXFP8 only exposes blockwise E8M0-encoded inverse scales, not a
+        # global FP32 scaling factor suitable for PTQ checkpoint export.
+        return None
+    elif recipe == "fp8_block_scaling":
+        # FP8 block scaling only exposes blockwise inverse scales, not a
+        # global FP32 scaling factor suitable for PTQ checkpoint export.
+        return None
+    else:
+        raise ValueError(f"Unsupported quantization recipe {recipe!r}")
+
     buffer_name = f"{tensor_name}_tensor_{metadata_name}_{recipe}_te_ptq_calibrated"
     return buffer_name, metadata
 
--- a/transformer_engine/pytorch/module/_common.py
+++ b/transformer_engine/pytorch/module/_common.py
@@ -31,6 +31,17 @@ def _supports_calibration_decay(quantizer_type: type) -> bool:
     )
 
 
+def _is_in_activation_recompute_phase() -> bool:
+    """Whether a forward is running during activation recomputation."""
+    from ..distributed import in_fp8_activation_recompute_phase
+
+    if in_fp8_activation_recompute_phase():
+        return True
+    # Special hidden PyTorch AutoGrad identifier for activation recompute.
+    current_graph_task_id = getattr(torch._C, "_current_graph_task_id", None)
+    return current_graph_task_id is not None and current_graph_task_id() != -1
+
+
 def _resolve_calibration_quantizer(tensor: Any, quantizer: Any) -> Any:
     """Get the quantizer that owns calibration state for a tensor."""
     quantizer = getattr(tensor, "_quantizer", None) or quantizer
--- a/transformer_engine/pytorch/module/_common.py
+++ b/transformer_engine/pytorch/module/_common.py
@@ -5,6 +5,8 @@
 """Internal function used by multiple modules."""
 
 import dataclasses
+import functools
+import inspect
 import queue
 from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
@@ -17,13 +19,27 @@ from ..tensor.hybrid_tensor import HybridQuantizer
 from ..utils import get_default_init_method
 
 
+@functools.lru_cache(maxsize=None)
+def _supports_calibration_decay(quantizer_type: type) -> bool:
+    """Whether a quantizer's calibrate override accepts calibration_decay."""
+    try:
+        parameters = inspect.signature(quantizer_type.calibrate).parameters
+    except (TypeError, ValueError):
+        return False
+    return "calibration_decay" in parameters or any(
+        parameter.kind == inspect.Parameter.VAR_KEYWORD for parameter in parameters.values()
+    )
+
+
 def _resolve_calibration_quantizer(tensor: Any, quantizer: Any) -> Any:
     """Get the quantizer that owns calibration state for a tensor."""
     quantizer = getattr(tensor, "_quantizer", None) or quantizer
     return getattr(quantizer, "parent_quantizer", quantizer)
 
 
-def _get_scale_buffer_info(tensor_name: str, quantizer: Any) -> Dict[str, torch.Tensor]:
+def _get_calibration_metadata_buffers(
+    tensor_name: str, quantizer: Any
+) -> Dict[str, torch.Tensor]:
     """Get checkpoint-buffer aliases from quantizer calibration state."""
     if quantizer is None:
         return {}
@@ -33,7 +49,7 @@ def _get_scale_buffer_info(tensor_name: str, quantizer: Any) -> Dict[str, torch.
     return {
         # Standard naming for PTQ calibration data.
         f"{tensor_name}_tensor_{metadata_name}_{recipe}_te_ptq_calibrated": value
-        for metadata_name, value in getattr(quantizer, "_calibration_state", {}).items()
+        for metadata_name, value in quantizer._calibration_state.items()
     }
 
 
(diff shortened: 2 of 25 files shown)


transformer_engine/pytorch/module/_common.py
```diff
@@ -6,16 +6,91 @@
...
+        metadata = getattr(quantizer, "amax", None)
+    elif recipe == "fp8_current_scaling":
+        metadata_name = "scale_inv"
+        metadata = getattr(tensor, "_scale_inv", None)
+    elif recipe == "nvfp4":
+        metadata_name = "amax"
+        metadata = getattr(tensor, "_amax_rowwise", None)
+    elif recipe == "nvfp4_rowwise":
+        metadata_name = "amax_rowwise"
+        metadata = getattr(tensor, "_amax_rowwise", None)
```
ptrendx, line 41: Those should probably be merged into one.

ptrendx:
@cspades, there is already a `calibrate` API available in the quantizers (and invoked by modules), even though it only really works for the delayed scaling. Why couldn't that API be reused and extended in this PR instead of introducing calling new APIs in the modules?

cspades:
> already a calibrate API available in the quantizers (and invoked by modules)

@ptrendx It can! I didn't know that was there. Apologies for the late reply have been working on other tasks.

That being said, wanted to share that this max-reduce kind of works (vs. ModelOpt checkpoints, and I think Xiaokang did this with NVFP4 since that was the main topic of the issue):
Screenshot 2026-08-17 at 11 25 29 AM

but can definitely merge/fold this into the calibrate API. Will report back.

transformer_engine/pytorch/module/_common.py
```diff
@@ -6,16 +6,91 @@
...
+                    f"{tuple(scale_buffer.shape)} to {tuple(observed_scale.shape)}"
+                )
+            if scale_buffer is None:
+                # Initialize the rolling activation scaling factor.
+                # Requires CUDA graph warmup step.
+                scale_buffer = torch.zeros_like(observed_scale)
+                scale_buffers[buffer_name] = scale_buffer
+            # Track a decaying maximum so early-training activation
+            # outliers do not permanently determine the inference scale.
+            scale_buffer.mul_(activation_scale_decay)
```
shangxiaokang, line 83: 'scale_buffer' might be Nan in some step when amax = 0, then, 
```
S_enc = 448 * 6 / global_amax   # 2688 / 0 = Inf
S_dec = block_amax * S_enc / 6  # 0 * Inf = NaN
```

There will be an error rat https://github.com/NVIDIA/Megatron-LM/pull/6183/changes#diff-debd55cccb8837189e626c048e61fa3ef0b63f15883fd299427f0de0e0a25bc6R387-R390.

cspades: Yup, I'll filter these out.

transformer_engine/pytorch/module/layernorm_mlp.py
```diff
@@ -557,12 +567,35 @@ def _forward(
...
+        should_calibrate = calibration_buffers is not None
+        if should_calibrate:
+            if fc1_input_calibration_quantizer is not None:
+                if _supports_calibration_decay(type(fc1_input_calibration_quantizer)):
+                    fc1_input_calibration_quantizer.calibrate(
+                        ln_out_total,
+                        calibration_decay=transformer_engine_calibration_decay,
+                    )
+                else:
+                    fc1_input_calibration_quantizer.calibrate(ln_out_total)
```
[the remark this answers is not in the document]
cspades: Yup yup fixing...

transformer_engine/pytorch/module/_common.py
cspades:
```
        if observed_value is None or torch.isnan(observed_value).any():
            # Un-initialized scale. Ignore it.
            return
```
done

transformer_engine/pytorch/module/_common.py
```diff
@@ -17,6 +19,49 @@
...
+
+def _is_in_activation_recompute_phase() -> bool:
+    """Whether a forward is running during activation recomputation."""
+    from ..distributed import in_fp8_activation_recompute_phase
+
+    if in_fp8_activation_recompute_phase():
+        return True
+    # Special hidden PyTorch AutoGrad identifier for activation recompute.
+    current_graph_task_id = getattr(torch._C, "_current_graph_task_id", None)
+    return current_graph_task_id is not None and current_graph_task_id() != -1
```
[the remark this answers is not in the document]
cspades: Huh, you just told me to fix it. https://github.com/NVIDIA/TransformerEngine/pull/3296#discussion_r3994280676

This is completely intended. While double-decay is meaningless at `0.0` and barely noticeable at `0.99`, we should indeed only calibrate on the forward pass (and we are, there is no backward calibration) assuming recomputed activations are going to be the same in the backward pass. Furthermore, inference (which uses this PTQ calibration) has a source-of-truth being the forward / row-wise activations and forward / row-wise weights. (And even then, we only calibrate FP32 globals, which have no orientation or difference in FWD/BWD.)

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -564,14 +565,57 @@ def create_metadata(
...
+
+    def _update_calibration_value(
+        self,
+        metadata_name: str,
+        observed_value: Optional[torch.Tensor],
+        *,
+        calibration_decay: float,
+    ) -> None:
+        """Merge an observation into quantizer-owned calibration state."""
+        if observed_value is None or torch.isnan(observed_value).any():
```
ptrendx, line 587: This makes the calibration not graph-safe, which is fine I think, but this should be included in the documentation (also, the PR suggests that it is graph-safe by modifying the graph infrastructure).

transformer_engine/pytorch/module/_common.py
ptrendx, line 42: Why do we need to use this hidden attribute here?

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -564,14 +565,57 @@ def create_metadata(
...
+            # Un-initialized scale. Ignore it.
+            return
+        observed_value = observed_value.detach()
+        calibration_state = self._calibration_state
+        calibration_value = calibration_state.get(metadata_name)
+        if calibration_decay > 0.0:
+            if calibration_value is not None and calibration_value.shape != observed_value.shape:
```
ptrendx, line 594: We do have recipes like the nvfp4 rowwise scaling where the shape will change throughout the calibration in the case of variable shaped inputs. Not sure how to handle such cases with calibration to be honest - we should probably just disallow calibration for those recipes.

transformer_engine/pytorch/tensor/hybrid_tensor.py
```diff
@@ -157,6 +157,9 @@ def copy(self) -> "HybridQuantizer":
         quantizer.optimize_for_gemm = self.optimize_for_gemm
         return quantizer
 
+    def calibrate(self, tensor: torch.Tensor, *, calibration_decay: float = 0.0) -> None:
+        """HybridQuantizer calibrate() has not yet been implemented."""
```
ptrendx, line 161: Should this error out instead?

transformer_engine/pytorch/tensor/mxfp8_tensor.py
```diff
@@ -132,9 +132,16 @@ def is_quantizable(self, inp: torch.Tensor) -> bool:
...
+    def calibrate(self, tensor: torch.Tensor, *, calibration_decay: float = 0.0) -> None:
+        """Calibrate an MXFP8 tensor."""
+        # NOTE(@cspades): Currently, PTQ calibration requirements don't need
+        # non-global / blockwise scaling factors, which are usually computed
+        # on-the-fly during inference. Implement this interface for future
+        # applications of MXFP8 calibration.
+
+    def get_quantization_recipe_name(self) -> str:
+        """Get the stable name of the quantization recipe."""
+        return "mxfp8"
```
ptrendx, line 144: Hmmm. Shouldn't it be added to the recipe instead? It seems that this is a parallel way of figuring out what the recipe is to what we already have. Maybe that is a better way, but I would still prefer to see a single one rather than two.

transformer_engine/pytorch/quantization.py
```diff
@@ -385,6 +385,37 @@ def is_nvfp4_available(return_reason: bool = False) -> Union[bool, Tuple[bool, s
...
+    ----------
+    transformer_engine_calibration_decay : float, default = 0.0
+        Decay applied to buffered activation scaling factors before incorporating
+        each new observation. With zero decay, only the latest value is retained.
+    """
+
+    transformer_engine_calibration_decay: float = 0.0
+
+    def __post_init__(self) -> None:
+        if self.transformer_engine_calibration_decay < 0.0:
```
ptrendx, line 402: Probably 0 < decay < 1?

examples/pytorch/mnist/main.py
```diff
@@ -76,7 +76,10 @@ def calibrate(model, device, test_loader, fp8):
     with torch.no_grad():
         for data, target in test_loader:
             data, target = data.to(device), target.to(device)
-            with te.autocast(enabled=fp8, calibrating=True):
+            with te.autocast(
+                enabled=fp8,
+                calibration_config=te.QuantizationCalibrationConfig(),
```
ptrendx, line 81: If the intended practice is to get rid of calibrating in the future and instead use only the config as the way to enable calibration then that should be added to the docs somehow. Personally, I'm not necessarily sure we would want to do that and then this example should still have calibrating=True in this call as well.

transformer_engine/pytorch/module/_common.py
```diff
@@ -17,6 +19,49 @@
...
+@functools.lru_cache(maxsize=None)
+def _supports_calibration_decay(quantizer_type: type) -> bool:
+    """Whether a quantizer's calibrate override accepts calibration_decay."""
+    try:
+        parameters = inspect.signature(quantizer_type.calibrate).parameters
+    except (TypeError, ValueError):
+        return False
+    return "calibration_decay" in parameters or any(
+        parameter.kind == inspect.Parameter.VAR_KEYWORD for parameter in parameters.values()
+    )
```
ptrendx, line 31: I don't think this function is needed - I don't believe custom calibration was a thing before this PR anyway, so we don't need to worry about breaking it.

transformer_engine/pytorch/module/_common.py
ptrendx: Also, if we really do use it then why don't we put it in the in_fp8_activation_recompute_phase in the first place?

transformer_engine/pytorch/module/_common.py
```diff
@@ -17,6 +19,49 @@
...
+    """Get checkpoint-buffer aliases from quantizer calibration state."""
+    if quantizer is None:
+        return {}
+    recipe = quantizer.get_quantization_recipe_name()
+    if not recipe:
+        return {}
+    return {
+        # Standard naming for PTQ calibration data.
+        f"{tensor_name}_tensor_{metadata_name}_{recipe}_te_ptq_calibrated": value
+        for metadata_name, value in quantizer._calibration_state.items()
```
ptrendx, line 61: Why couldn't we have the recipe name also in the calibration state? And why wouldn't this function be a function of the quantizer instead? This string that is being built here basically becomes our API (since it lands in the checkpoint if I understand correctly), so it should not be hidden in a random underscore function inside an underscore module.

ptrendx: Also, this string should be documented somewhere for people trying to get those calibrated values and putting them in the inference engines.

ptrendx: And one more thing - based on the usage, this function and the one above could be merged.

transformer_engine/pytorch/quantized_tensor.py
```diff
@@ -564,14 +565,57 @@ def create_metadata(
             "nontensor_kwargs": meta["nontensor_kwargs"],
         }
 
-    def calibrate(self, tensor: torch.Tensor) -> None:
-        """Calibrate quantizer state
-
-        Updates quantization state as if quantizing a tensor, but
-        without actually performing the quantization.
+    def calibrate(self, tensor: torch.Tensor, *, calibration_decay: float = 0.0) -> None:
```
ptrendx, line 568: Why don't we just pass the full calibration config here instead of just the decay? If we add more fields there it will be awkward to change this API again.

transformer_engine/pytorch/tensor/float8_blockwise_tensor.py
```diff
@@ -243,10 +243,16 @@ def is_quantizable(self, inp: torch.Tensor) -> bool:
...
-    def calibrate(self, tensor: torch.Tensor) -> None:
-        # NOTE: This interface is specific to requirements like delayed scaling
-        # where state from an estimator influences distribution parameters.
-        pass
+    def calibrate(self, tensor: torch.Tensor, *, calibration_decay: float = 0.0) -> None:
+        """Float8BlockQuantizer does not yet support calibration."""
+        # NOTE(@cspades): Currently, PTQ calibration requirements don't need
+        # non-global / blockwise scaling factors, which are usually computed
+        # on-the-fly during inference. Implement this interface for future
+        # applications of blockwise FP8 calibration.
```
ptrendx, line 251: So if somebody specifically asked for a calibration in this case we should probably issue a warning or at least info saying that this does not actually make sense.

transformer_engine/pytorch/quantized_tensor.py
cspades: Oh, I do want everything to be graph-safe, I added this lazily in response to the NaN bug.

I think we can just torch.isnan mask into zeros, run through this function, and trivialize the update, assuming that we have a calibration buffer initialized already (since let's say, previous values were not NaN, just this new observation was).

Good catch!

transformer_engine/pytorch/quantized_tensor.py
cspades: Makes sense, I think I did at least informally support the rowwise absmax (which I end up just [max-reducing into a scalar prior to checkpointing as a way to normalize NVFP4 calibration](https://github.com/NVIDIA/Megatron-LM/pull/6183/changes#diff-debd55cccb8837189e626c048e61fa3ef0b63f15883fd299427f0de0e0a25bc6R301)), but I didn't really consider changing shape during runtime.

I think the Megatron side was when I decided to walk back on even the fanciness of globally-checkpointing an FP32 vector that doesn't take too much memory in the checkpoint. We can just make this calibration function only work on scalars and reduce the amax vector into a scalar here probably.

transformer_engine/pytorch/module/_common.py
cspades: Does `in_fp8_activation_recompute_phase` also cover non-FP8 / general activation recomputation?

I am biased to use Torch native detection mechanisms and increase the recall of this function, because if we screw it up, we'll be double-decaying and colwise-writing our calibration values.

ptrendx: But is this hidden thing a native detection mechanism that we would like to use here? It being internal and hidden means that it could change at any point.

Issue #3054, opened 2026-05-28
[Common] Move heavy kernel compilation to NVRTC


CarlosGomes98:
# Reduce TE build time by migrating template-heavy kernels to NVRTC

## Problem

Build time for Transformer Engine is currently very long.

## Hypothesis

This is dominated by compilation of heavily templated CUDA kernels. Each `.cu` translation unit expands template products over (dtype × layout × hidden-size × ...) axes at build time, producing large object files (multi-MB each) and long per-TU `nvcc` invocations. Many of these template axes are runtime-discoverable, so the work could be deferred to NVRTC and amortized across only the kernel variants actually invoked.

The pattern is already in use under [`transformer_engine/common/transpose/rtc/`](../tree/main/transformer_engine/common/transpose/rtc) (cast-transpose, transpose, swap-first-dims, cast-transpose-fusion).

This issue tries to model the impact of the this migration.

### Measurement methodology

All measurements are on a single workstation: 32 logical CPUs, 125 GB RAM, NVIDIA driver + `nvcc 13.1` (CUDA 13.1), host compiler `g++ 12` (forced via `-DCMAKE_CUDA_HOST_COMPILER=/usr/bin/g++-12` to dodge a gcc-13 ICE). Build target is `libtransformer_engine.so` (the C++ common library, framework wrappers excluded).

Two configurations were measured:

- **sm_89** (initial measurement): `-DCMAKE_CUDA_ARCHITECTURES=89`, no extra flags.
- **sm_100a + fast-math activations**: `-DCMAKE_CUDA_ARCHITECTURES=100a -DNVTE_BUILD_ACTIVATION_WITH_FAST_MATH=ON`

Clean full build of the common library:

```bash
# After cmake configure of build/cmake/ (the location pip uses)
/usr/bin/time -f "wall=%es cpu_user=%Us cpu_sys=%Ss max_rss=%MKB" \
  .venv/bin/ninja -C build/cmake transformer_engine
```

Per-TU breakdown comes from parsing the ninja log with a script made by claude.

### sm_89 — current state

```
wall    = 110.89 s
cpu_user = 1624.70 s          # sequential-equivalent compile work
cpu_sys  =   71.75 s
max_rss  = 4063 MB            # peak across all parallel nvcc jobs
outputs  = 79 TUs + 1 link
sequential compile total = 1797.1 s
effective parallelism = 1797.1 / 110.89 = 16.2× (50.6% of 32 cores)
```

[...]

PR #3301, opened 2026-08-03
NVRTC NVFP4 quantization kernels

CarlosGomes98:
# Description

Migrate NVFP4 quantization kernels to NVRTC. #3054 

These kernels represent a large chunk of TE's build time for sm100 + in the cast.cu TU.

Migrating them to NVFP4 I record a 70% reduction in build time, due to the quantize_4over6 template fan-out being removed.

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Introduce nvfp4 nvrtc support. Keeps support for non-nvrtc path by extracting out the kernel body.
- Support for Architecture family specification for NVRTC
- A few additional util files due to the usual pain point of not being able to include some host-side libs in nvrtc.

# Checklist:


transformer_engine/common/cast/nvfp4/core_nvfp4.cuh
```diff
@@ -94,6 +115,7 @@ __device__ __forceinline__ float compute_global_encode_scaling_factor_FP4(const
   return global_encode_scale;
 }
 
+#if !defined(__CUDACC_RTC__)
```
ptrendx, line 115: If you are not including this in the NVRTC build then how do you handle the stochastic rounding?

transformer_engine/common/include/transformer_engine/nvfp4/4over6.h
```diff
@@ -0,0 +1,26 @@
+/*************************************************************************
```
ptrendx, line 1: Could we have that in some directory rather than in the top-level include directory?

transformer_engine/common/cast/nvfp4/quantize_4over6_nvfp4.cuh
```diff
@@ -614,6 +626,23 @@ __global__ void __launch_bounds__(kThreads)
...
+                           const float *amax_rowwise, const float *amax_colwise, const size_t rows,
+                           const size_t cols, const size_t scale_stride,
+                           const size_t scale_stride_t, const float *noop) {
+  quantize_4over6_body<USE_2D_QUANTIZATION, RETURN_IDENTITY, RETURN_TRANSPOSE, ROW_SCALED_NVFP4,
+                       Cfg, E4M3_MAX, IType>(input, output, output_t, scales, scales_t,
+                                             amax_rowwise, amax_colwise, rows, cols, scale_stride,
+                                             scale_stride_t, noop);
+}
+
+#if !defined(__CUDACC_RTC__)
```
ptrendx: Could we instead split this file so that the kernel is in its own header included in this file?

transformer_engine/common/cast/nvfp4/rtc_dispatch.cuh
```diff
@@ -0,0 +1,103 @@
...
+ ************************************************************************/
+
+/*! \file rtc_dispatch.cuh
+ *  \brief Host-side NVRTC dispatch for the NVFP4 4over6 quantize kernel.
+ */
+
+#ifndef TRANSFORMER_ENGINE_CAST_NVFP4_RTC_DISPATCH_CUH_
+#define TRANSFORMER_ENGINE_CAST_NVFP4_RTC_DISPATCH_CUH_
+
+#if !defined(__CUDACC_RTC__)
```
ptrendx: Considering this entire file is excluded, why do we even try to pass it to nvrtc? Also why is it .cuh extension? It does not seem to actually need to be CUDA header, but rather regular c++ header, no?

transformer_engine/common/util/ptx.cuh
```diff
@@ -22,6 +23,23 @@
...
+#else
+// NVRTC build: common.h drags in host-only headers (cuDNN etc.) and cannot be
+// compiled by NVRTC.
+#ifndef FP4_TYPE_SUPPORTED
+#define FP4_TYPE_SUPPORTED (CUDA_VERSION >= 12080)
+#endif
+#include <cuda_bf16.h>
+#if FP4_TYPE_SUPPORTED
+#include <cuda_fp4.h>
+#endif  // FP4_TYPE_SUPPORTED
```
ptrendx, line 30: Snippets like these should just live outside the RTC macro rather than being duplicated in 2 code paths.

transformer_engine/common/util/rtc.cpp
```diff
@@ -26,22 +27,80 @@ namespace {
...
+
+/*! \brief Resolve the static build's Blackwell target policy for NVRTC */
+inline ArchSpecificity resolve_arch_specificity(ArchSpecificity specificity, int sm_arch) {
+  if (specificity != ArchSpecificity::BlackwellSpecific) {
+    return specificity;
+  }
+
+  // Keep this matrix synchronized with NVTE_SPECIFIC_ARCHS in CMakeLists.txt.
+  // The runtime NVRTC version is the JIT equivalent of CUDAToolkit_VERSION.
+  switch (sm_arch) {
```
ptrendx: We don't need to actually repeat the logic of the specific archs from cmake here - since we are building those kernels at runtime on the target machine, we can just always compile for the "a" variant of the specific chip that we run on (for 90+, since that's where this became a thing). The cmake compiles to "a" or "f" in order to limit the compilation targets, here we don't have this problem.

ptrendx: That way we will also have less maintenance since we will not need this switch at all.

transformer_engine/common/common.h
```diff
@@ -680,41 +678,7 @@ TRANSFORMER_ENGINE_TYPE_NAME(__nv_fp4_e2m1)
...
-  static constexpr float max = 0x1.FEp127;
-};
-
-template <>
-struct TypeExtrema<fp16> {
-  // Hex float format of 1.(10 bits of 1) * 2 ^ 15
-  static constexpr float max = 0x1.FFCp15;
-};
-
+// Host-only fallback for types without an explicit specialization above.
```
ptrendx: Please put that together with the rest of things in type_extrema.h (just guard it), as it belongs there.

transformer_engine/common/utils.cuh
```diff
@@ -24,10 +24,21 @@ using uint8_t = unsigned char;
 using uint16_t = unsigned short int;  // NOLINT(*)
 using uint32_t = unsigned int;
 using uint64_t = unsigned long long int;  // NOLINT(*)
+using int8_t = signed char;
+using int16_t = short int;                 // NOLINT(*)
+using int32_t = int;                       // NOLINT(*)
+using int64_t = long long int;             // NOLINT(*)
+using intptr_t = long long int;            // NOLINT(*)
+using uintptr_t = unsigned long long int;  // NOLINT(*)
```
ptrendx: Why do we need those?

transformer_engine/common/cast/nvfp4/core_nvfp4.cuh
CarlosGomes98: This particular kernel rejects stochastic rounding

```
  NVTE_CHECK(!quant_config->stochastic_rounding,
             "NVFP4 4over6 quantization does not support stochastic rounding.");
``` 

We can do a guard like this, or move this get_rbits to a separate header which we then include in the other kernels which use it.

transformer_engine/common/cast/nvfp4/quantize_4over6_nvfp4.cuh
CarlosGomes98: Agreed. I split this out into its own header to minimize the code that we pass to rtc. This naturally resolved a lot of the other comments you had regarding the amount of RTC guards I added.

transformer_engine/common/cast/nvfp4/rtc_dispatch.cuh
CarlosGomes98: True, this was not actually being passed in the header list for nvrtc, so there was no reason for the .cuh

transformer_engine/common/util/ptx.cuh
CarlosGomes98: Done, with a few other examples I found of this

transformer_engine/common/util/rtc.cpp
CarlosGomes98: That makes a lot of sense - I was too focused on trying to exactly replicate the compile-time rules. Now should compile always to "a"

transformer_engine/common/utils.cuh
CarlosGomes98: int16_t, int32_t, and int64_t are needed because NVRTC now has to parse ptx.cuh, which uses them, while avoiding `<cstdint>`. The rest were leftovers from the mxfp8 work, I've removed them.

transformer_engine/common/cast/nvfp4/rtc_dispatch.h
```diff
@@ -0,0 +1,103 @@
...
+  const std::string itype_name = rtc_type_name<IType>();
+  const std::string mode_name = rtc_mode_name(Cfg::mode);
+
+  // Cache key encodes everything that varies the compiled kernel.
+  const std::string kernel_label =
+      std::string("quantize_4over6,itype=") + itype_name +
+      ",use2d=" + (USE_2D_QUANTIZATION ? "1" : "0") + ",id=" + (return_identity ? "1" : "0") +
+      ",t=" + (return_transpose ? "1" : "0") + ",rowscaled=" + (row_scaled_nvfp4 ? "1" : "0") +
+      ",mode=" + mode_name + ",fastmath=" + (Cfg::err_use_fast_math ? "1" : "0") +
+      ",e4m3max=" + std::to_string(E4M3_MAX);
```
ptrendx: I'm a little worried about the CPU overhead of those cache key creation and lookups. Could you
benchmark how long does this launching function take? You should at least use reserve function of the std::string,
but generally we should think about using just structs as the keys to the cache rather than serializing them.

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -19,6 +19,8 @@
 extern "C" {
 #endif
 
+#include <transformer_engine/nvfp4/4over6.h>
```
ptrendx, line 22: Do we need to include that in the top-level header? People that need it can just include the
corresponding header directly.

transformer_engine/common/CMakeLists.txt
```diff
@@ -589,6 +604,16 @@ make_string_header_from_file(normalization/rmsnorm/rtc/rmsnorm_bwd_kernel.cu
 target_include_directories(transformer_engine PRIVATE
                            "${CMAKE_CURRENT_BINARY_DIR}/string_headers")
 
+# Default OFF: the NVFP4 4over6 quantize kernel (384-way fanout) is compiled via
+# NVRTC at runtime. Set ON to additionally statically instantiate it as the
+# fallback selected when NVTE_DISABLE_NVRTC=1.
+option(NVTE_BUILD_LEGACY_STATIC_NVFP4
```
ptrendx, line 615: Flag should be added to documentation.

transformer_engine/common/include/transformer_engine/transformer_engine.h
CarlosGomes98: I added it here because the function QuantizationConfig::set_nvfp4_4over6_mode requires an enum which I moved to the device-safe header transformer_engine/nvfp4/4over6.h. So this file itself needs the include

transformer_engine/common/cast/nvfp4/rtc_dispatch.h
CarlosGomes98: Makes sense, I will try to do some benchmarking and come back with the results soon

CarlosGomes98: I modified it to use reserve() as well as find() and end() instead of count() in the lookup.

For the benchmarking, I timed 1000 launches of the static compile (main) vs nvrtc, isolating the launch time with something like

```cpp
  cudaStreamSynchronize(stream);  // Start with an empty stream

  start = steady_clock::now();

  for (int i = 0; i < batch_size; ++i) {
      launch_kernel();             // Asynchronous CUDA launch
  }

  stop = steady_clock::now();

  cudaStreamSynchronize(stream);  // Wait for execution, outside timed region

  time_per_launch = (stop - start) / batch_size;
  ```
  
Static: 1.79 µs
NVRTC: 2.03 µs

So about 14% increase.

I have a hard time judging if this is acceptable. If it's not, I think we could improve this by caching the launch handle after the first launch. This will add some code complexity so I'll wait until we are sure we want to optimize this before giving it a shot.

As for the structs, that makes sense to me as well, but since it affects other kernels as well I would argue for leaving it for a separate PR.

ptrendx (approved):

PR #3304, opened 2026-08-03
[JAX] Migrate EP collective-stream annotation to region-based compute_on

phu0ngng:
# Description

JAX rewrote `jax.experimental.compute_on.compute_on` into a region-based transform. This migrates `_on_collective_stream`, which uses `compute_on` to the new signature.

## Type of change
- [ ] Infra/Build change
- [ ] Code refactoring

## Changes
- `_on_collective_stream` calls `compute_on(compute_type="gpu_stream:collective", out_memory_spaces=Device)` on a nullary thunk that closes over the static `EpLayerConfig`/`PartitionSpec` args, since region-based `compute_on` abstract-evals every argument, then invokes the annotated call.
- Region-based `compute_on` moves `_xla_stream_annotation` onto the async wrapper XLA generates, so `test_z_dispatch_combine_on_collective_stream` checks each EP call is reachable from a collective-annotated wrapper.
- Region-based `compute_on` and `jax.memory.Space` exist since JAX 0.10.1, the current collective-stream floor, so no min-version bump is needed.

# Checklist:


phu0ngng:
Pipeline #60891724 with the nightly JAX image.

jberchtold-nvidia (dismissed):
LGTM, thanks!

phu0ngng:
pipeline #61310617 with the nightly JAX image.

transformer_engine/jax/cpp_extensions/ep.py
jberchtold-nvidia, line 34: Can we update `is_collective_stream_supported`'s definition to attempt to import `compute_on` in a try/catch, and return False if an import error occurs? Since this `compute_on` is experimental, it seems JAX may freely rename this function without warning, so any future JAX versions could cause TE to break until we update TE

If we instead tried to import `compute_on` as part of the `is_collective_stream_supported` check, we'd just skip the compute_on logic but TE would still work and import correctly. We could log a warning in this case

We could even go a step further and wrap the body of `_on_collective_stream` in a try/catch, so if something subtler happens like JAX keeps `compute_on` as the same function name but changes the params, we would try/catch then log a warning and fallback to the original function without the `compute_on` annotation.

Since this can fail at import-time, it affects all of TE usage even for modules unrelated to TE EP, so I think try/catch'ing at least the first import case would be helpful

What do you think?

transformer_engine/jax/moe.py
```diff
@@ -751,6 +751,8 @@ def _body(*args):
             num_local_tokens=(B, S),
             out_partition_spec=out_partition_spec,
         )
+    # output of MLP should be sharded the same way as the activation input
+    output = with_sharding_constraint_by_logical_axes(output, input_axes)
```
jberchtold-nvidia, line 755: `input_axes` and `out_partition_spec` should be equivalent here, right? If so, can we simplify this be replacing `out_partition_spec` above with `input_axes`

jberchtold-nvidia: Can be in follow-up

transformer_engine/jax/cpp_extensions/ep.py
jberchtold-nvidia: Can be in follow-up

jberchtold-nvidia (approved):

PR #3315, opened 2026-08-04
[Common/PyTorch] Grouped weighted-SwiGLU MXFP8 kernel

cael-ling:
# Description

Add `nvte_group_scaled_swiglu`, which fuses the scaled SwiGLU activation with column-wise MXFP8 quantization over grouped (MoE) tensors:

```
saved FC1 output [N, 2H]  --(weighted SwiGLU)-->  [N, H]  --(colwise MXFP8)-->  FC2 wgrad input

scaled_swiglu(x) = ( silu(x[:, :H]) * x[:, H:] ) * prob[:, None]
```

Also add `nvte_group_scaled_clamped_swiglu` (`tex.group_scaled_clamped_swiglu`), the same kernel with the clamped-SwiGLU activation, matching the semantics in `gated_mxfp8.cuh`:

```
act  = min(x[:, :H], limit)
gate = clamp(x[:, H:], -limit, limit) + glu_linear_offset

scaled_clamped_swiglu(x) = ( clamped_silu(act) * gate ) * prob[:, None]

clamped_silu(a) = a * sigmoid(alpha * a)    # alpha = 1 reduces this to silu(a)
```

`limit` is required and must be positive; `alpha` defaults to 1.702 and `glu_linear_offset` to 1.0.

## Performance (GB200)

`N=98304`, compact scales, eager timing. `BW` is the achieved bandwidth against the minimum DRAM traffic the path must move, so it says how close each variant is to being purely bandwidth bound.

### Expert-count sweep

| experts | H=2048 fused | BW | vs compiled | H=4096 fused | BW | vs compiled |
|---|---|---|---|---|---|---|
| 2  | 164.4 us | 6.16 TB/s | 1.69x | 320.1 us | 6.33 TB/s | 1.71x |
| 4  | 164.0 us | 6.17 TB/s | 1.69x | 320.5 us | 6.32 TB/s | 1.71x |
| 8  | 164.8 us | 6.15 TB/s | 1.68x | 322.2 us | 6.29 TB/s | 1.70x |
| 16 | 166.0 us | 6.10 TB/s | 1.67x | 323.5 us | 6.26 TB/s | 1.70x |
| 32 | 167.0 us | 6.06 TB/s | 1.66x | 325.2 us | 6.23 TB/s | 1.69x |
| 64 | 168.0 us | 6.03 TB/s | 1.65x | 326.4 us | 6.21 TB/s | 1.69x |

Going from 2 to 64 experts costs about 2%, which is the per-expert metadata indexing.

### Shape sweep, 8 experts

| N × H | fused | BW | vs compiled |
|---|---|---|---|
| 8192×1024   |  31.1 us | 1.36 TB/s | 3.36x |
| 8192×2048   |  31.3 us | 2.70 TB/s | 2.96x |
| 8192×4096   |  38.5 us | 4.39 TB/s | 2.31x |
| 8192×8192   |  62.8 us | 5.38 TB/s | 1.57x |
| 32768×1024  |  38.1 us | 4.43 TB/s | 2.50x |
| 32768×2048  |  62.5 us | 5.40 TB/s | 1.63x |
| 32768×4096  | 114.7 us | 5.89 TB/s | 1.63x |
| 32768×8192  | 219.0 us | 6.17 TB/s | 1.68x |
| 98304×1024  |  87.1 us | 5.81 TB/s | 1.62x |
| 98304×2048  | 166.1 us | 6.10 TB/s | 1.67x |
| 98304×4096  | 323.9 us | 6.25 TB/s | 1.70x |
| 98304×8192  | 638.8 us | 6.34 TB/s | 1.72x |

### 16 experts, all variants

| variant | H=2048 | H=4096 | vs fused |
|---|---|---|---|
| fused (`nvte_group_scaled_swiglu`) | 166.7 us | 323.5 us | 1.00x |
| unfused-compiled (3 elementwise ops fused into 1 kernel) | 277.8 us | 548.5 us | **1.67x / 1.70x** |
| unfused-te-op (`ScaledSwiGLU` + `tex.group_quantize`) | 314.6 us | — | 1.89x |

### Clamped

The baseline is the clamped expression, since the plain one computes less arithmetic:

| variant | H=2048 | H=4096 | vs fused-clamped |
|---|---|---|---|
| fused-clamped (`nvte_group_scaled_clamped_swiglu`) | 178.2 us | 342.3 us | 1.00x |
| unfused-compiled-clamped | 297.0 us | 589.5 us | **1.78x / 1.82x** |

Over the same shape sweep as above, 8 experts, against the clamped baseline:

| N × H | fused-clamped | BW | vs plain fused | vs compiled-clamped |
|---|---|---|---|---|
| 8192×1024   |  30.7 us | 1.37 TB/s | −1.3% | 3.43x |
| 8192×2048   |  30.5 us | 2.77 TB/s | −2.6% | 3.19x |
| 8192×4096   |  39.7 us | 4.25 TB/s | +3.3% | 2.65x |
| 8192×8192   |  66.7 us | 5.06 TB/s | +6.4% | 1.58x |
| 32768×1024  |  39.5 us | 4.27 TB/s | +3.7% | 2.44x |
| 32768×2048  |  66.2 us | 5.10 TB/s | +6.0% | 1.60x |
| 32768×4096  | 121.8 us | 5.54 TB/s | +6.2% | 1.64x |
| 32768×8192  | 233.7 us | 5.78 TB/s | +6.7% | 1.69x |
| 98304×1024  |  92.0 us | 5.50 TB/s | +5.6% | 1.64x |
| 98304×2048  | 177.6 us | 5.71 TB/s | +6.9% | 1.68x |
| 98304×4096  | 342.7 us | 5.91 TB/s | +5.8% | 1.72x |
| 98304×8192  | 689.5 us | 5.88 TB/s | +7.9% | 1.72x |

## Numerics

The activation uses `tanh.approx.f32` via the identity `2*silu(x) = x*(1 + tanh(x/2))`, one transcendental per element instead of the two the exact sigmoid needs. The compensating factor of 2 is folded into the per-token `prob` during shared-memory staging, so it costs one multiply per row rather than per element.

The output has at most 3 mantissa bits, so the approximation is far below one FP8 ULP. Its only visible effect is on the shared e8m0 block scale: a block whose amax sits within the approximation error of a power-of-two boundary can pick a different exponent than the CPU reference. Measured rate is 2 flips in 361472 blocks, and the gtest budget is 4 per tensor.

## Reproduce

```bash
# expert-count sweep with absolute throughput
python benchmarks/benchmark_group_scaled_swiglu.py \
    --hidden 2048 --num-groups 2 4 8 16 32 64 --scales compact --loop eager

# full variant matrix: compact/swizzled scales x eager/graph timing
python benchmarks/benchmark_group_scaled_swiglu.py --num-groups 16

# clamped
python benchmarks/benchmark_group_scaled_swiglu.py \
    --variants fused-clamped unfused-compiled-clamped

# over-allocated / empty-expert cases
python benchmarks/benchmark_group_scaled_swiglu.py \
    --num-groups 8 64 --occupancy 1.0 0.5 0.125 0.03125 --variants fused fused-clamped

# shape sweep (one invocation per shape)
python benchmarks/benchmark_group_scaled_swiglu.py \
    --tokens 8192 --hidden 8192 --num-groups 8 --scales compact --loop eager

# correctness
./tests/cpp/build/operator/test_operator --gtest_filter='*GroupedScaledSwigluMXFP8*'
```

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


transformer_engine/common/include/transformer_engine/activation.h
```diff
@@ -85,6 +85,27 @@ void nvte_silu(const NVTETensor input, NVTETensor output, cudaStream_t stream);
...
+ *  compact or in the cuBLAS GEMM-swizzled layout; the swizzled layout additionally
+ *  requires F divisible by 128 and, for multiple experts, VARYING_FIRST_DIM.
+ *
+ *  \param[in]     input     Grouped input tensor [T, 2F] ([act|gate]).
+ *  \param[in]     prob      Per-token weights, at least T elements, in the same
+ *                           dtype as \p input.
+ *  \param[in,out] output    Grouped output tensor [T, F] (columnwise MXFP8).
+ *  \param[in]     stream    CUDA stream used for the operation.
+ */
+void nvte_group_swiglu_quantize(const NVTEGroupedTensor input, const NVTETensor prob,
```
vthumbe1503: Can we rename this? Quantization details are already encoded in GroupedTensor. It ll be consistent with the other API naming convention
```suggestion
void nvte_group_scaled_swiglu(const NVTEGroupedTensor input, const NVTETensor prob,
```

transformer_engine/pytorch/csrc/extensions.h
```diff
@@ -352,6 +352,12 @@ py::object group_quantize(const at::Tensor &tensor, py::handle quantizer, const
                           std::optional<at::Tensor> tensor_offsets,
                           std::optional<at::Tensor> noop_flag);
 
+py::object group_swiglu_quantize(const at::Tensor &input_2f, const at::Tensor &prob,
```
vthumbe1503: Similar comment. 
```suggestion
py::object group_scaled_swiglu(const at::Tensor &input_2f, const at::Tensor &prob,
```

vthumbe1503 (commented):
Thanks for the PR @cael-ling. Could you also please publish the perf numbers for this kernel?

cael-ling:
Hi @vthumbe1503, thanks for your review, rename work has been performed, benchmark results have been added in the description, please take a look, thanks!

transformer_engine/common/include/transformer_engine/activation.h
cael-ling: Resolved

transformer_engine/pytorch/csrc/extensions.h
cael-ling: Resolved

transformer_engine/pytorch/csrc/quantizer.cpp
```diff
@@ -94,6 +94,13 @@ void check_grouped_metadata_tensor(const at::Tensor& metadata_tensor, const char
   NVTE_CHECK(metadata_tensor.is_contiguous(), metadata_name, " must be contiguous.");
   NVTE_CHECK(static_cast<size_t>(metadata_tensor.numel()) == expected_len, metadata_name,
              " must have length ", expected_len, ".");
+  // The offsets are built and consumed by kernels on the current device's stream, and the
+  // grouped data is allocated there too, so metadata on another device would be read
+  // through a foreign pointer.
+  const int current_device = static_cast<int>(at::cuda::current_device());
+  const int metadata_device = static_cast<int>(metadata_tensor.get_device());
+  NVTE_CHECK(metadata_device == current_device, metadata_name, " must be on the current CUDA ",
+             "device (cuda:", current_device, "), but is on cuda:", metadata_device, ".");
```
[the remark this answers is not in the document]
cael-ling: Resolved

transformer_engine/pytorch/csrc/extensions/cast.cpp
```diff
@@ -397,6 +399,138 @@ py::object group_quantize(const at::Tensor &tensor, py::handle quantizer, const
...
+             "group_scaled_swiglu only supports MXFP8 quantizers.");
+  NVTE_CHECK(input_2f.is_cuda(), "group_scaled_swiglu input must be a CUDA tensor.");
+  // Both operands are handed to the kernel as raw pointers over a densely packed
+  // range, so a strided view would be read as if it were contiguous.
+  NVTE_CHECK(input_2f.is_contiguous(), "group_scaled_swiglu input must be contiguous.");
+  NVTE_CHECK(prob.is_contiguous(), "group_scaled_swiglu prob must be contiguous.");
+  NVTE_CHECK(prob.device() == input_2f.device(),
+             "group_scaled_swiglu prob must be on the same device as the input.");
+  NVTE_CHECK(prob.numel() >= static_cast<int64_t>(T),
+             "group_scaled_swiglu prob must have at least T elements.");
```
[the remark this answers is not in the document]
vthumbe1503: Could we fix this @cael-ling ?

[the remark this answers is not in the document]
vthumbe1503: @cael-ling I fixed the merge conflict and updated with this change as well

transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh
```diff
@@ -0,0 +1,596 @@
...
+  asm("tanh.approx.f32 %0, %1;" : "=f"(tanh_ah) : "f"(half_alpha * x));
+  return fmaf(x, tanh_ah, x);
+}
+
+// clamp(g, -limit, limit) for limit > 0, which the binding enforces.
+__device__ __forceinline__ float clamp_symmetric(const float g, const float limit) {
+  float clamped;
+  asm("min.xorsign.abs.f32 %0, %1, %2;" : "=f"(clamped) : "f"(g), "f"(limit));
+  return clamped;
+}
```
Oleg-Goncharov: Can't we reuse the existing operators from `common/util/math.h`, or define these there if needed?

transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh
```diff
@@ -0,0 +1,596 @@
...
+    //
+    // Every branch produces twice the activation, to pair with the halved prob.
+    float act_x2;
+    if constexpr (std::is_same_v<ParamOP, ClampedSwiGLUParam>) {
+      gate_elt = clamp_symmetric(gate_elt, p.limit) + p.glu_linear_offset;
+      if constexpr (OP == &clamped_silu<fp32, fp32>) {
+        act_x2 = clamped_silu_approx_x2(fminf(act_elt, p.limit), half_alpha);
+      } else {
+        act_x2 = 2.0f * OP(act_elt, p);
+      }
```
Oleg-Goncharov: Maybe we could avoid dividing by 2 and then multiplying by 2 again by modifying those operators? This would make the code a bit cleaner here and avoid the need to special-case `if constexpr (OP == &clamped_silu<fp32, fp32>)`
And the lines from R187 to R191 similarly

transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh
```diff
@@ -0,0 +1,596 @@
...
+  }  // if constexpr (is_single_tensor)
+#endif  // (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 1000)
+}
+
+}  // namespace group_scaled_swiglu_kernel
+
+// Host launcher: grouped scaled SwiGLU -> columnwise MXFP8.
+//   input  : GroupedTensor [T, 2F] ([act|gate]) in a floating input dtype.
+//   prob   : Tensor [T] per-token weights, in the input (model) dtype.
+//   output : GroupedTensor with columnwise_data / columnwise_scale_inv for [T, F].
```
Oleg-Goncharov: Nit: could we use the usual tensor dimension naming, `[N, H]`, instead of `[T, F]`?

transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh
```diff
@@ -0,0 +1,596 @@
...
+  const size_t out_last_logical_dim = output->logical_shape.data[1];  // F
+  const size_t in_last_logical_dim = input->logical_shape.data[1];    // 2F
+
+  NVTE_CHECK(in_last_logical_dim == 2 * out_last_logical_dim,
+             "group_scaled_swiglu input last dim must be 2x the output last dim ([act|gate]).");
+  NVTE_CHECK(input->logical_shape.data[0] == first_logical_dim,
+             "group_scaled_swiglu input/output must share the token dimension T.");
+
+  const size_t T = first_logical_dim;
+  const size_t F = out_last_logical_dim;
```
Oleg-Goncharov: Nit: similar to above, could we use the usual tensor dimension naming, `[N, H]`, instead of `[T, F]` here and throughout the code below?

transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh
```diff
@@ -0,0 +1,596 @@
...
+  // Single-tensor schedule: one virtual work grid over [T, F].
+  const size_t work_blocks_Y = DIVUP(T, static_cast<size_t>(CHUNK_DIM_Y));
+  const size_t work_blocks_X = DIVUP(F, static_cast<size_t>(CHUNK_DIM_X));
+
+  NVTE_CHECK(T % 128 == 0, "group_scaled_swiglu requires T divisible by 128.");
+
+  const size_t sm_num = static_cast<size_t>(transformer_engine::cuda::sm_count());
+  const size_t static_grid_size = sm_num * TunableConfig::STATIC_PERSISTENT_BLOCKS_PER_SM;
+  NVTE_CHECK(static_grid_size > 0, "Static persistent grid size must be greater than zero.");
+  const dim3 grid(static_grid_size);
```
Oleg-Goncharov, line 456: Maybe not necessary for this PR, but something we could consider in a follow-up PR if/when performance becomes important. There is [PR #3337 Common: Improved performance of Group MXFP8 kernels](https://github.com/NVIDIA/TransformerEngine/pull/3337), which improves performance by selecting specific **kernel launch configurations** depending on the GroupCase (`SAME_BOTH_DIMS`, `VARYING_FIRST_DIM`, etc.). Similar **launch-configuration** tuning could potentially help here as well.

transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh
```diff
@@ -0,0 +1,596 @@
...
+
+using namespace dispatch::common;
+
+// Reuse the same tiling as group_quantize_mxfp8 so the scheduler/TMA math match.
+struct TunableConfig {
+  static constexpr uint CHUNK_DIM_Y = 128;
+  static constexpr uint CHUNK_DIM_X = 128;
+  static constexpr uint THREADS_PER_CHUNK = 128;
+  static constexpr uint STATIC_PERSISTENT_BLOCKS_PER_SM = 24;
+};
```
Oleg-Goncharov, line 54: Maybe not necessary for this PR, but something we could consider in a follow-up PR if/when performance becomes important. There is [PR #3337 Common: Improved performance of Group MXFP8 kernels](https://github.com/NVIDIA/TransformerEngine/pull/3337), which improves performance by selecting specific **kernel parameters** depending on the GroupCase (`SAME_BOTH_DIMS`, `VARYING_FIRST_DIM`, etc.). Similar tuning could potentially help here as well.

Oleg-Goncharov (commented):
LGTM overall. If performance becomes important here, there may still be room for additional optimization similar to what was done in the unmerged [PR #3337 — Common: Improved performance of Group MXFP8 kernels](https://github.com/NVIDIA/TransformerEngine/pull/3337), for example by tuning the kernel launch configuration for different GroupCases.

It would also be useful to include absolute performance numbers (e.g. throughput) for the gated kernels, in addition to the `Group` vs. `Non-Group` comparison. Ideally, these numbers would cover different shapes, different numbers of tensors in a group (e.g. 2, 4, 8, 16, 32, 64), and cases where the tensor allocation is much larger than the actual amount of data to be processed. For example, the allocation could be `8192x8192`, while only 1024 rows need to be processed, meaning that some tensor IDs effectively correspond to empty tensors.

That would help us understand how close these kernels are to the expected performance and whether further optimization is actually needed.

transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh
cael-ling: Done, `silu_approx_x2` and `clamped_silu_approx_x2` now live in `common/util/math.h` next to `silu` and `clamped_silu`, with the same `(val, const ParamOP&)` signature, and the gate clamp is there as `clamp_symmetric`.

transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh
cael-ling: The `if constexpr` is gone, which was the real point, but I got there by making the
approximate operators first-class rather than by removing the factor of two, because that
factor turns out not to be a round trip.

Both operators now live in `math.h` with the standard `(val, const ParamOP&)` signature, so
`swiglu_grouped.cu` passes them as `OP` directly (`silu_approx_x2<fp32, fp32>` /
`clamped_silu_approx_x2<fp32, fp32>`) and the whole body collapses to:

```cpp
if constexpr (std::is_same_v<ParamOP, ClampedSwiGLUParam>) {
  gate_elt = clamp_symmetric(gate_elt, p.limit) + p.glu_linear_offset;
}
float elt = OP(act_elt, p) * gate_elt * half_prob;

File: transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh (hunk 7 above)
Comment (cael-ling) [reply transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh]: Done

File: transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh (hunk 8 above)
Comment (cael-ling) [reply transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh]: Done throughout: the kernel and its host launcher, the `nvte_group_scaled_swiglu` /
`nvte_group_scaled_clamped_swiglu` doc blocks, the dispatch layer, the pybind layer, the
tests and the benchmark. The pybind keyword argument went from `input_2f` to `input_2h` for the same reason.

File: transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh (hunk 10 above)
Comment (cael-ling) [reply transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh:54]: Agreed as a follow-up. One data point on this struct specifically: I did sweep the tile shape here — `CHUNK_DIM_X` at 64 / 128 / 256, together with `CHUNK_DIM_Y` and prefetch depth — and 128 was a clear measured optimum. 256 regressed about 9% and also broke the small-`H` cases. So the headroom is probably not in these constants; selecting them per `GroupCase` as #3337 does is untried and orthogonal to that sweep.

File: transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh (hunk 9 above)
Comment (cael-ling) [reply transformer_engine/common/cast/mxfp8/group_scaled_swiglu_mxfp8.cuh:456]: Agreed as a follow-up.

**Comment by cael-ling** [comment]
Thanks @Oleg-Goncharov,I have addressed your review comments point‑by‑point. The absolute numbers have been added in PR description, and they also answer @vthumbe1503's earlier request.

At `8192x8192` with 8 experts the fused kernel runs 62.8 us at 5.38 TB/s, 1.57x the `torch.compile` baseline, and it's insensitive to both expert count (2/4/8 within 1.5%).

File: transformer_engine/pytorch/csrc/extensions/cast.cpp [hunk 11]
```diff
@@ -426,6 +428,136 @@ py::object group_quantize(const at::Tensor &tensor, py::handle quantizer, const
...
+                 " must be on the same device as the input.");
+    }
+  };
+  check_metadata_device(first_dims, "first_dims");
+  check_metadata_device(last_dims, "last_dims");
+  check_metadata_device(tensor_offsets, "tensor_offsets");
+
+  // Allocate the output and launch on the operands' device rather than on whatever
+  // torch.cuda.set_device last selected.
+  at::cuda::CUDAGuard device_guard(input_2h.device());
```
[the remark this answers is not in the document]
cael-ling: group_quantize is not modified by this PR — zero changed lines; the hunk header naming it is git's nearest-preceding-signature context, since this PR appends after that function ends.

The guard here is a function-local RAII object in group_scaled_swiglu_impl, not a shared helper, so group_quantize could not have inherited it.

An earlier revision did put the device requirement in the shared check_grouped_metadata_tensor, which group_quantize and two other grouped entry points go through. I backed that out in 8572e3aa because it silently tightens the contract for four existing call sites this PR does not otherwise touch — and a subsequent Greptile review scored the result 5/5, reasoning that the issues were "guarded in the new binding, while the prior current-device metadata rejection no longer exists in the shared validator." This review asks for the opposite of what that one approved.

transformer_engine/common/util/math.h
```diff
@@ -83,6 +83,51 @@ __device__ inline OType clamped_silu(const IType val, const ClampedSwiGLUParam&
   return qgelu_with_alpha<OType, float>(cval, p.alpha);
 }
 
+// clamp(val, -limit, limit). Requires limit > 0: the single-instruction form below
+// derives the lower bound from the upper one's sign.
+__device__ inline float clamp_symmetric(const float val, const float limit) {
+#if (defined __CUDA_ARCH__) && (__CUDA_ARCH__ >= 860)
+  float clamped;
+  asm("min.xorsign.abs.f32 %0, %1, %2;" : "=f"(clamped) : "f"(val), "f"(limit));
```
vthumbe1503, line 96: Is this PTX instruction faster? We use the same logic for clamped swiglu as well and probably this function can be used there as well. Nothing more needs to be done in this PR. Just wanted to know, so we can use this function for clamped swiglu in future if perf demands it.

tests/pytorch/test_grouped_tensor.py
```diff
@@ -637,6 +637,107 @@ def test_group_quantize_precomputed_offsets(self, output_dbias: bool) -> None:
...
+        num_tensors = 3
+        last_dim = 256
+        split_sizes_list = [128, 384, 512]
+        total_tokens = sum(split_sizes_list)
+
+        input_2h = torch.randn(total_tokens, 2 * last_dim, dtype=torch.bfloat16, device="cuda")
+        prob = torch.rand(total_tokens, dtype=torch.bfloat16, device="cuda")
+        first_dims = torch.tensor(split_sizes_list, dtype=torch.int64, device="cuda")
+
+        quantizer = MXFP8Quantizer(fp8_dtype=tex.DType.kFloat8E4M3)
```
vthumbe1503: Could we use TE_Dtype instead of tex.DType here? tex.DType is not intended to be used as user facing.

tests/pytorch/test_grouped_tensor.py
```diff
@@ -637,6 +637,107 @@ def test_group_quantize_precomputed_offsets(self, output_dbias: bool) -> None:
...
+                wide[:, : 2 * last_dim], prob, quantizer, num_tensors, first_dims
+            )
+
+        strided_prob = torch.rand(2 * total_tokens, dtype=torch.bfloat16, device="cuda")[::2]
+        with pytest.raises(RuntimeError):
+            tex.group_scaled_swiglu(input_2h, strided_prob, quantizer, num_tensors, first_dims)
+
+    @pytest.mark.parametrize("optimize_for_gemm", [False, True])
+    @pytest.mark.skipif(not mxfp8_available, reason=reason_for_no_mxfp8)
+    def test_group_scaled_clamped_swiglu_shapes(self, optimize_for_gemm: bool) -> None:
```
vthumbe1503, line 791: I would replace this test with verify the exact correctness of group_scaled_swiglu and group_scaled_clamped_swiglu against a more thorough pytorch based reference implementation. Pytorch reference outputs can then be quantized and compared against the grouped_scaled_* outputs

vthumbe1503 (commented):
I just have one minor concern in pytorch test. We should verify results against a correct pytorch reference implementation. Apart from that LGTM

tests/pytorch/test_grouped_tensor.py
vthumbe1503: @cael-ling I have updated the branch as well. Since there was a recent fix in the pytorch reference implementation of clamped swiglu present in test_fusible_ops.py. You can reuse that reference implementation

tests/pytorch/test_grouped_tensor.py
cael-ling: Yes, switched to the public te.DType.kFloat8E4M3

tests/pytorch/test_grouped_tensor.py
cael-ling: Added test_group_scaled_swiglu_matches_reference, parametrized over clamped x optimize_for_gemm. Reference is the test_fusible_ops.py one character for character, including your torch.where form from #3424, so the grouped kernel is pinned to the canonical ScaledSwiGLU / ScaledClampedQGeGLU. Each expert's reference comes from that expert's own rows and is quantized with the same quantizer instance the kernel used.

transformer_engine/common/util/math.h
cael-ling: Yes, should be faster.

cael-ling:
@vthumbe1503 Thanks for the review, all three addressed in 66aa446d.

Oleg-Goncharov (commented):
Thank you @cael-ling for addressing those items. The kernel part LGTM. We may follow up with a PR to improve kernel job management for better performance, similar to [PR #3337 [Common] Improved performance of Group MXFP8 kernels](https://github.com/NVIDIA/TransformerEngine/pull/3337).

Oleg-Goncharov (approved):
LGTM.

PR #3276, opened 2026-07-29
[Pytorch] [NCCL EP] Allow zero tokens for an EP rank in eager mode

YangFei1990:
# Description

For eager mode, there could be case where certain EP ranks do not receive any token, based on the routing strategy. This PR relax the constraint to allow such behavior.

Fixes # (issue)

## Type of change

- [ ] Documentation change (change only to the documentation, either a fix or a new content)
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

- [x] I have read and followed the [contributing guidelines](https://github.com/NVIDIA/TransformerEngine/blob/main/CONTRIBUTING.rst)
- [x] The functionality is complete
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #3317, opened 2026-08-05
[Common][PyTorch] Fuse the RHT into grouped NVFP4 quantize on non-SM100 architectures

davidkny22:
# Description

Follows #3265. That PR made `split_quantize` with RHT-enabled NVFP4 quantizers fall back to quantizing each split on its own outside the SM100 family, since the grouped Hadamard transform cast-fusion kernels are SM100 only. This makes that fallback fused.

The fallback runs two launches per split plus a separate `HadamardTransformKernel` that writes the transformed tensor to global memory for the quantize pass to read back. At G=64 with 256-row splits that is 320 launches per `split_quantize` call and 11.125 bytes of traffic per input element.

`group_quantize_transpose` already has most of what a fused version needs. `MultiAmaxCastTransposeFusionArgs` carries `colwise_amax_list`, `output_colwise_data_list`, `output_colwise_scale_inv_list` and `output_colwise_scale_stride`, all marked "unused for rowwise only scaling", and the transposed store path is present but commented out with `TODO(zhongbo): add back when transpose is supported`. This completes that path with per-split direct stores instead of the shared memory staging it replaces, folds the 16-point random Hadamard transform into the columnwise read as an in-register butterfly, and adds `nvte_group_quantize_with_colwise_rht` for the PyTorch dispatch to call. The same G=64 case becomes 24 launches and 3.125 bytes per element.

Direct stores make the transposed staging buffers dead, so they go, along with the commented-out stores that named them and two declarations (`tid_Y_t`, `out_mem_rowwise_data`) that only fed those buffers. Dynamic shared memory drops from 25728 to 20608 bytes per 128-thread CTA, 3 resident CTAs per SM to 4.

On GB10 (sm_121a) this is 2.22x median on `split_quantize` eager, 2.00x median of GPU work under graph capture, over 14 shapes. At `te.GroupedLinear` level it is 1.00x to 1.17x of a forward and backward step, median 1.07x.

The fused path is taken only when every split is a multiple of 128 rows. A split boundary is resolved once per 128-row chunk in the transposed direction, so a ragged split would write into the wrong buffer. Ragged cases decline to the existing per-split path. For MoE this means expert capacity has to be padded to 128 tokens to benefit.

The new launches are chunked by input bytes rather than by tensor count, at 8 MiB, so the amax pass prefetches for the quantize pass instead of being evicted from L2 before it reads the same data. Chunking by tensor count instead puts 128 MiB against a 24 MiB L2 at G=64, which costs most of the win.

## Architecture behaviour

The transposed path is reachable only through the new entry point. `group_quantize_transpose` refuses a columnwise output unless `nvfp4_colwise_rht` is set, and only `nvte_group_quantize_with_colwise_rht` sets it, so generic `nvte_group_quantize` keeps today's refusal on every architecture. The new entry point additionally refuses on `sm >= 100 && sm <= 110`, the same band `split_quantize_nvfp4_impl` uses, so the SM100 family cannot reach the new code even through the C API.

No existing entry point changes behaviour on any architecture.

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- `transformer_engine/common/cast/nvfp4/group_quantize_transpose_nvfp4.cuh`: complete the transposed output path with per-split direct stores of columnwise data and scale factors, add the in-register 16-point RHT behind a `COLWISE_RHT` template switch, bind the columnwise pointers on the per-stage tensor switch, gate the transposed path on the columnwise RHT entry point and the architecture band, require 128-row-aligned splits, and drop the now-dead transposed staging buffers and the two declarations that only fed them.
- `transformer_engine/common/cast/cast.cu`: `nvte_group_quantize_with_colwise_rht`.
- `transformer_engine/common/include/transformer_engine/cast.h`: its declaration and docs.
- `transformer_engine/common/common.h`: two internal quantization config fields, `nvfp4_colwise_rht` and `nvfp4_rht_sign_mask_t`, not exposed through the attribute API.
- `transformer_engine/pytorch/csrc/extensions/cast.cpp`: use the grouped path inside the existing non-SM100 branch when the quantizers carry post-RHT amax, both usages, no 2D quantization, no row-scaled NVFP4 and no 4over6, and every split is a multiple of 128 rows. Empty splits are filtered out of the grouped launch. Launch pairs chunk at 8 MiB of input.
- `tests/pytorch/nvfp4/test_nvfp4_group_quantize.py`: fused versus per-split equality across aligned, mixed, empty and ragged splits.

# Checklist:

No documentation changes needed beyond the header docs on the new entry point. On the last box: the added tests pass and nothing regresses, but `test_nvfp4_group_quantize.py` has 120 pre-existing failures on main and `test_nvfp4_group_quantize_graph_safe.py` has 528, unchanged by this PR. Numbers and cause below.

# Testing done

GB10 (DGX Spark), sm_121a, CUDA 13.0, driver 580.95.05. Built and gated at main @ `d35eedf5f6d1cfba64b5e1dd4c8bfc3fc8214750`, the #3265 merge commit, then re-verified as `git apply --check` clean against `af1ed441255ec66e3555fd485dddb4344d7c495a`, current main tip at submission, which adds one unrelated commit (#3276, NCCL EP) touching none of these six files. Built with `NVTE_CUDA_ARCHS=121a`. clang-format 18.1.6, cpplint 2.0.2 and black 24.4.2 clean on the changed files. No warning lines from any touched file, verified by rebuilding the affected translation units with `ninja -v` rather than trusting the `pip install` log, which hides per-file compiler output.

Both arms live in one binary. `NVTE_NVFP4_DISABLE_GROUPED_RHT` is read with `std::getenv` at the dispatch on every call, so the A/B is one build and one process on identical operands. The new test uses it to compare the two routes, and it is the way out of the one-ulp difference below for a caller that needs output identical to the per-split path.

## Tests

`test_rht_split_quantize_grouped_matches_unfused` compares the two routes byte for byte over 12 cases: aligned, mixed-aligned, empty-front, empty-back, empty-mixed and two ragged split patterns, at two widths. It would have caught the empty-split bug I hit building this, where a leading zero-row split has `has_data()` true with a null data pointer and aborted the grouped launch. It does not assert which route ran; the launch counts below do that.

`tests/pytorch/nvfp4/test_nvfp4_group_quantize.py` goes 120 failed, 453 passed, 270 skipped to 120 failed, 465 passed, 270 skipped, the 12 being the new cases. `test_nvfp4_group_quantize_graph_safe.py` is 528 failed, 60 passed either way. Those failures are pre-existing on main: the `optimize_for_gemm=True` cases, where swizzled scale factor emission is gated on the same architecture band so outputs carry compact scale factors while the test swizzles the reference.

## Numerics

17 case classes compare the fused route against the per-split route byte for byte: aligned, ragged, mixed, empty-front, empty-middle, empty-back, and 64-way splits. Each case runs the fallback twice and the fused route once; the second fallback run is the control, since scale buffers are allocated at `roundup(rows, 128)` and neither route writes the padding, so an uncontrolled comparison would report phantom mismatches that move with allocator state.

16 of 17 are identical over the full buffer and over the defined region, on rowwise data, columnwise data, both scale planes and both amaxes.

One is not. A single byte differs, `columnwise_data[195, 78]` of one split, reproducibly, with both arms individually deterministic. It is reduction order rather than a race. `HadamardTransformKernel` evaluates the 16x16 transform with `mma_m16_n16_k16_b16_b16_b16_noacc`, bf16 operands and fp32 accumulate; the fused path evaluates the same transform as a scalar fp32 butterfly. Both multiply by exactly representable values, so the only difference is the order of the fp32 additions in a 16-term reduction. For that element the input chunk spans 3.4375 down to 1.4e-07, the sum cancels heavily, the two orders land one bf16 ulp apart (-0.1000976562 against -0.1005859375), and the element sits on the FP4 rounding boundary for its block, so it moves one FP4 code. Block amax, both scale planes and the rowwise plane are unaffected.

Rate over 1,073,741,824 columnwise elements across 5 shapes and 4 seeds each: zero further instances. Counting the one known case against everything compared, 1 in 1.1e9, bounded to one code step, no measurable norm or sign bias.

Identical on every tested input except that one-ulp reduction-order difference, which can move a single FP4 code. Not bit identical, and it cannot be while the two paths reduce in different orders.

## Launch counts

Kernel launches per `split_quantize` call, from the profiler:

| case | fallback | fused |
|---|---|---|
| G=64, [256] x64, cols 2048 | 320 | 24 |
| G=8, [1024] x8, cols 4096 | 40 | 24 |
| G=8, [96,160,32,224,...], cols 512 | 40 | 40, declined |

## Performance, split_quantize

Clocks locked at 3003 MHz, 9 windows, 20 iterations per window, medians, arms alternating window by window so they share clock history. Both arms also captured into CUDA graphs, which removes per-launch submission cost from both sides. Milliseconds per call.

| G | rows/split | cols | fallback | fused | eager | fallback graph | fused graph | graph |
|---|---|---|---|---|---|---|---|---|
| 1 | 16384 | 4096 | 3.695 | 1.577 | 2.34x | 3.667 | 1.583 | 2.32x |
| 2 | 8192 | 4096 | 3.562 | 1.596 | 2.23x | 3.652 | 1.587 | 2.30x |
| 8 | 2048 | 4096 | 2.655 | 1.527 | 1.74x | 2.656 | 1.483 | 1.79x |
| 8 | 1024 | 2048 | 0.465 | 0.328 | 1.42x | 0.429 | 0.349 | 1.23x |
| 8 | 2048 | 14336 | 12.955 | 5.856 | 2.21x | 13.078 | 5.766 | 2.27x |
| 32 | 512 | 2048 | 1.228 | 0.740 | 1.66x | 0.908 | 0.684 | 1.33x |
| 32 | 512 | 4096 | 2.019 | 1.462 | 1.38x | 1.716 | 1.384 | 1.24x |
| 64 | 256 | 2048 | 1.775 | 0.706 | 2.51x | 1.269 | 0.670 | 1.89x |
| 64 | 256 | 4096 | 2.648 | 1.394 | 1.90x | 1.912 | 1.320 | 1.45x |
| 64 | 128 | 2048 | 1.336 | 0.461 | 2.90x | 0.878 | 0.301 | 2.92x |
| 64 | 128 | 4096 | 1.800 | 0.659 | 2.73x | 1.287 | 0.605 | 2.13x |
| 64 | 512 | 4096 | 4.052 | 2.917 | 1.39x | 3.326 | 2.753 | 1.21x |
| 128 | 128 | 2048 | 2.768 | 0.940 | 2.95x | 1.753 | 0.603 | 2.91x |
| 128 | 128 | 4096 | 3.511 | 1.311 | 2.68x | 2.546 | 1.208 | 2.11x |

Median 2.22x eager, range 1.38x to 2.95x. Median 2.00x of GPU work, range 1.21x to 2.92x. No shape regresses in either column. On a gate-only build, where both arms are the same fallback code, the same harness reads 1.00x median on 14 of 14 shapes, range 1.00x to 1.03x.

The gap between the columns is per-launch submission cost the fallback pays on 320 to 640 launches, removed by graph capture.

## Performance, te.GroupedLinear

Forward and backward at the `te.GroupedLinear` level, steady-state trainer settings (`is_first_microbatch=False`, `fuse_wgrad_accumulation=True`). Expert counts and per-expert token counts come from published MoE configurations at `tokens x top_k / experts`. Recipe `NVFP4BlockScaling(disable_stochastic_rounding=True)`, since the SR kernel uses a `cvt.rs` instruction sm_121 does not have. 7 windows, 5 iterations, medians. Milliseconds per step.

| shape | G | rows/expert | in | out | fallback | fused | eager | graph |
|---|---|---|---|---|---|---|---|---|
| Mixtral-8x7B fc1 | 8 | 2048 | 4096 | 28672 | 117.17 | 100.21 | 1.17x | 1.18x |
| Mixtral-8x7B fc2 | 8 | 2048 | 14336 | 4096 | 59.66 | 50.99 | 1.17x | 1.18x |
| Qwen3-235B-A22B fc1 | 128 | 256 | 4096 | 3072 | 88.06 | 82.68 | 1.07x | 1.03x |
| DeepSeek-V3 fc1, EP=8 | 32 | 1024 | 7168 | 4096 | 73.49 | 68.44 | 1.07x | 1.09x |
| GPT-OSS-120B fc1 | 128 | 128 | 2880 | 5760 | 104.91 | 102.55 | 1.02x | 0.99x |
| Qwen3 fc1, ragged router | 128 | 32 to 1120 | 4096 | 3072 | 90.04 | 90.27 | 1.00x | 1.00x |

Median 1.07x, range 1.00x to 1.17x. Noise floor on a gate-only build with both arms identical: 1.00x eager on all six shapes, 0.98x to 1.01x graph.

The ragged case is the `%128` decline, and it reads 1.00x with identical launch counts on both arms. GPT-OSS declines in the forward for an unrelated reason: `split_quantize` already refuses NVFP4 kernel fusion when the input's inner dimension is not a multiple of 128, and 2880 is not, so only its backward fuses.

The share of the step is why the module number is smaller than the kernel number: on Mixtral fc1 the RHT quantize path is 25.1% of device time in the fallback and 14.2% after; on Qwen3 fc1 the GEMM is roughly 85% of the step and the RHT path 7.4%, which caps anything done here at about 1.08x on that shape.

# Open questions

1. Is the fused transform meant to match `HadamardTransformKernel` bit for bit? The one-ulp difference is inherent to evaluating the same transform in a different reduction order, so if exact match is required this needs a different approach.
2. Should `kChunkBytes` be derived from the device's L2 size rather than fixed at 8 MiB? I only have one L2 size to size it against.
3. Is the `%128` alignment requirement acceptable for how grouped quantize gets called in practice, or is ragged-split support worth pursuing?

cc @zhongbozhu @Oleg-Goncharov

### DIFF
--- a/tests/pytorch/nvfp4/test_nvfp4_group_quantize.py
+++ b/tests/pytorch/nvfp4/test_nvfp4_group_quantize.py
@@ -274,3 +274,77 @@ def test_rht_split_quantize_matches_per_tensor_reference(quantize_mode: str) -> 
         split_sections=split_sections,
         with_rht=True,
     )
+
+
+@pytest.mark.skipif(not recipe_available, reason=reason_for_no_recipe)
+@pytest.mark.parametrize("N", [512, 2048])
+@pytest.mark.parametrize(
+    "split_sections",
+    [
+        [128, 128, 128, 128],
+        [256, 512, 128, 384],
+        [0, 256, 256, 0],
+        [0, 128, 0, 256, 128, 0, 384, 128],
+        [192, 320],
+        [64, 64, 192, 192],
+    ],
+    ids=[
+        "aligned",
+        "mixed_aligned",
+        "empty_ends",
+        "empty_mixed",
+        "unaligned",
+        "small_unaligned",
+    ],
+)
+def test_rht_split_quantize_grouped_matches_unfused(monkeypatch, split_sections, N) -> None:
+    # split_quantize folds the RHT into the columnwise pass in one grouped launch when
+    # every split is a multiple of 128 rows, and quantizes each split on its own
+    # otherwise. Empty splits are dropped from the grouped launch. Both routes have to
+    # produce the same bytes.
+    x = torch.randn((sum(split_sections), N), dtype=torch.bfloat16, device="cuda")
+
+    def run(disable_grouped: bool):
+        monkeypatch.setenv("NVTE_NVFP4_DISABLE_GROUPED_RHT", "1" if disable_grouped else "0")
+        quantizers = [
+            NVFP4Quantizer(
+                fp4_dtype=te.DType.kFloat4E2M1,
+                rowwise=True,
+                columnwise=True,
+                with_rht=True,
+                with_post_rht_amax=True,
+            )
+            for _ in split_sections
+        ]
+        return [
+            {
+                "rowwise_data": out._rowwise_data.view(dtype=torch.uint8).clone(),
+                "columnwise_data": out._columnwise_data.view(dtype=torch.uint8).clone(),
+                "amax_rowwise": out._amax_rowwise.clone(),
+                "amax_columnwise": out._amax_columnwise.clone(),
+                "rowwise_scale_inv": out._rowwise_scale_inv.clone(),
+                "columnwise_scale_inv": out._columnwise_scale_inv.clone(),
+            }
+            for out in tex.split_quantize(x, split_sections, quantizers)
+        ]
+
+    torch.manual_seed(0)
+    unfused = run(True)
+    fused = run(False)
+
+    x_splits = torch.split(x, split_sections)
+    for i, rows in enumerate(split_sections):
+        if rows == 0:
+            continue
+        for key in ("rowwise_data", "columnwise_data", "amax_rowwise", "amax_columnwise"):
+            torch.testing.assert_close(fused[i][key], unfused[i][key], atol=0.0, rtol=0.0)
+        # Scale buffers are allocated with padded shapes and neither route writes the
+        # padding, so compare only the region both of them define.
+        for key, columnwise in (("rowwise_scale_inv", False), ("columnwise_scale_inv", True)):
+            valid = get_nvfp4_scale_shape_no_padding(x_splits[i].shape, columnwise)
+            torch.testing.assert_close(
+                fused[i][key][: valid[0], : valid[1]],
+                unfused[i][key][: valid[0], : valid[1]],
+                atol=0.0,
+                rtol=0.0,
+            )
--- a/tests/pytorch/nvfp4/test_nvfp4_group_quantize.py
+++ b/tests/pytorch/nvfp4/test_nvfp4_group_quantize.py
@@ -376,8 +376,9 @@ def _grouped_kernel_names(split_sections, N, columnwise_per_split):
 
 
 def _launched_grouped_kernel(names):
-    return any("GroupHadamardAmaxTma" in n or "group_quantize_transpose_nvfp4_kernel" in n
-               for n in names)
+    return any(
+        "GroupHadamardAmaxTma" in n or "group_quantize_transpose_nvfp4_kernel" in n for n in names
+    )
 
 
 @pytest.mark.skipif(not recipe_available, reason=reason_for_no_recipe)
--- a/tests/pytorch/nvfp4/test_nvfp4_group_quantize.py
+++ b/tests/pytorch/nvfp4/test_nvfp4_group_quantize.py
@@ -348,3 +348,57 @@ def test_rht_split_quantize_grouped_matches_unfused(monkeypatch, split_sections,
                 atol=0.0,
                 rtol=0.0,
             )
+
+
+def _grouped_kernel_names(split_sections, N, columnwise_per_split):
+    # Kernel names, not output bytes: TE allocates every split's columnwise buffer
+    # regardless of that split's own columnwise_usage flag, and rowwise output does
+    # not depend on columnwise_usage, so a heterogeneous list that wrongly takes the
+    # grouped path can still produce byte-identical rowwise output to the per-split
+    # path on some shapes. Which kernel launched is the property that actually
+    # distinguishes the two routes; see the launch-count table in the PR body.
+    x = torch.randn((sum(split_sections), N), dtype=torch.bfloat16, device="cuda")
+    quantizers = [
+        NVFP4Quantizer(
+            fp4_dtype=te.DType.kFloat4E2M1,
+            rowwise=True,
+            columnwise=columnwise_per_split[i],
+            with_rht=True,
+            with_post_rht_amax=True,
+        )
+        for i in range(len(split_sections))
+    ]
+    torch.cuda.synchronize()
+    with torch.profiler.profile(activities=[torch.profiler.ProfilerActivity.CUDA]) as prof:
+        tex.split_quantize(x, split_sections, quantizers)
+        torch.cuda.synchronize()
+    return {e.key for e in prof.key_averages() if e.count > 0}
+
+
+def _launched_grouped_kernel(names):
+    return any("GroupHadamardAmaxTma" in n or "group_quantize_transpose_nvfp4_kernel" in n
+               for n in names)
+
+
+@pytest.mark.skipif(not recipe_available, reason=reason_for_no_recipe)
+def test_rht_split_quantize_grouped_kernel_engages_when_uniform() -> None:
+    # Positive control for the test below: with every split asking for the same
+    # usage, the grouped path is eligible and must be the one that runs.
+    split_sections = [128, 128, 128, 128]
+    names = _grouped_kernel_names(split_sections, 256, [True] * len(split_sections))
+    assert _launched_grouped_kernel(names)
+
+
+@pytest.mark.skipif(not recipe_available, reason=reason_for_no_recipe)
+def test_rht_split_quantize_declines_grouped_on_mismatched_quantizers() -> None:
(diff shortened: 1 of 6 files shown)


davidkny22:
Confirmed and fixed in the latest commit.

The exact mechanism Greptile named was not quite right (`rht_matrix_random_sign_mask_t` is memoized via `functools.lru_cache` on `(with_random_sign_mask, device)` in `get_random_sign_mask_for_rht`, so every quantizer on the same device already gets the same value there), but the underlying concern is real and broader than that one field.

`rowwise_usage` and `columnwise_usage` are the concrete case: they are the one axis `_validate_grouped_quantizer_list` in `grouped_linear.py` explicitly allows to vary per expert, and `NVFP4Quantizer` never defines `_value_fields()`, so that same validator's uniformity check is a no-op for every NVFP4-specific field, including these two. Traced it with a profiler: a heterogeneous quantizer list does engage `GroupHadamardAmaxTmaKernel` and `group_quantize_transpose_nvfp4_kernel` today, using only `quantizers[0]`'s settings for the whole launch.

Fix requires every quantizer in the list to agree with the first on the fields the grouped launch actually reads (`with_post_rht_amax`, `rowwise_usage`, `columnwise_usage`, `row_scaled_nvfp4`, `nvfp4_4over6_mode`, `stochastic_rounding`, the RHT sign mask), mirroring the scaling-mode check already in this function. A mismatch now falls through to the existing per-split loop.

Added a regression test asserting on which kernel launches, not on output bytes, since a heterogeneous `columnwise_usage` list does not actually produce diverging rowwise bytes on the shapes I tried (rowwise computation does not depend on it, and TE allocates every split's columnwise buffer regardless of the flag), so a byte-comparison test would have passed either way. Full suite unaffected: 120/467/270 on `test_nvfp4_group_quantize.py` (was 120/465/270, the +2 is the new test), 528/60 on the graph-safe file, both unchanged from before.

PR #3318, opened 2026-08-05
[PyTorch] Advance FusedAdam step counter for empty param groups

### LINKED ISSUE #1986: FusedAdam step counter desynchronizes across DP ranks with empty param_groups
**Describe the bug**

In `FusedAdam`, the step counter (`group["step"]`) for a parameter group becomes desynchronized across Data Parallel (DP) ranks if that group is empty on some ranks. The optimizer logic skips the step increment for any group where `len(group["params"]) == 0`.

This creates a critical issue when resuming from a checkpoint. A rank that holds parameters for a specific group (e.g., RMSNorm params in a `no_weight_decay` group) may load its optimizer state from another DP rank where that group was empty. This causes it to load an incorrect, stale `step` value. An incorrect `step` leads to a wrong `bias_correction` term (`1 - beta1 ** step`), causing training instability or divergence upon resumption.

Image

The image below illustrates this problem in a `PP=2, EP=4, DP=8` setup after 2640 steps. The step counter for the `no weight` decay group is `1` on several ranks (orange) because they contain no such parameters, while it is correctly `2641` on others.

Within the same data parallel group `(pp=0, ep=2)`, the rank `dp=6` (global rank 6) has the correct step `2641`. However, rank `dp=2` (global rank 2) has an incorrect step of `1`. When resuming, rank 6 might load the state from rank 2, inheriting the wrong step and breaking the `bias_correction` calculation.

**Suggested Fix:**
The step increment logic should be moved *before* the check for an empty parameter list to ensure it always executes.
```python
# In FusedAdam's step() method
for group in self.param_groups:
    # Increment step for all groups on all ranks first
    if "step" in group:
        group["step"] += 1
    else:
        group["step"] = 1
    
    # Then, skip the kernel launch for empty groups
    if len(group["params"]) == 0:
        continue
    
    # ... rest of the logic
```

adityasingh2400:
Fixes #1986

## Root cause

`FusedAdam.step()` opens its param-group loop with an early skip for groups that hold no parameters:

```python
for group in self.param_groups:
    if len(group["params"]) == 0:
        continue
    device = group["params"][0].device
    ...
    if "step" in group:
        group["step"] += ...
    else:
        group["step"] = ...
```

The step counter is updated after that skip, so an empty group never gets one. That is harmless for a group that is empty everywhere, but a group is often empty on only some data-parallel ranks. A `no_weight_decay` group holding just RMSNorm parameters is the usual case, and with pipeline or expert parallelism the ranks that own none of those parameters see an empty group while their peers do not.

`step` lives in `param_groups`, so `state_dict()` serializes it and it goes into the checkpoint. The ranks where the group was empty write `step = null` while the ranks where it was populated write the true iteration count, which is exactly what the counter table in the issue shows for a `PP=2, EP=4, DP=8` run at iteration 2640. On resume, a rank that loads its optimizer shard from a rank where the group was empty picks up the stale value, and `bias_correction` computes `1 - beta1 ** step` from a step that has nothing to do with how far training actually got.

## Fix

Move the counter update above the empty-group skip so every group advances on every rank, then skip the kernel work for empty groups as before. This is the change suggested in the issue.

One detail the issue does not cover: with `capturable=True` the first update creates `group["step"]` as a device tensor and takes the device from `group["params"][0]`, which an empty group does not have. The new code falls back to the device of `self._dummy_overflow_buf`, the optimizer's own scratch buffer, which is allocated on CUDA in `__init__`. Nothing else in the loop moved, so populated groups execute exactly the same sequence as before.

## Verification

I do not have a GPU, so I could not run the TE test suite. Two things I did do.

The control flow itself is checked with a standalone CPU script that reproduces the loop head, before and after, on a real `torch.optim.Optimizer` so that `param_groups` and `state_dict()` behave as they do in TE. The kernel launch plays no part in the defect and is omitted:

```python
import torch


class LoopHead(torch.optim.Optimizer):
    def __init__(self, params, fixed):
        super().__init__(params, {"lr": 1e-3})
        self.fixed = fixed

    def step(self):
        for group in self.param_groups:
            if self.fixed:
                # post-fix ordering
                if "step" in group:
                    group["step"] += 1
                else:
                    group["step"] = 1
                if len(group["params"]) == 0:
                    continue
            else:
                # ordering on main
                if len(group["params"]) == 0:
                    continue
                if "step" in group:
                    group["step"] += 1
                else:
                    group["step"] = 1


def run(fixed, num_steps=3):
    populated = torch.nn.Parameter(torch.zeros(4))
    optim = LoopHead([{"params": [populated]}, {"params": []}], fixed=fixed)
    for _ in range(num_steps):
        optim.step()
    return (
        [g.get("step") for g in optim.param_groups],
        [g.get("step") for g in optim.state_dict()["param_groups"]],
    )


for label, fixed in (("main", False), ("fixed", True)):
    print(label, run(fixed))
```

Output:

```
main  ([3, None], [3, None])
fixed ([3, 3], [3, 3])
```

The `None` in the checkpoint on `main` is the `null` step reported in the issue.

The regression test in this PR is the GPU version of the same property. `test_empty_param_group_advances_step` builds a `FusedAdam` over one populated group and one empty group, steps three times, and asserts that both groups report the same step in `param_groups` and in `state_dict()`. It is parametrized over `capturable` so the tensor-valued counter and the new device fallback are both exercised. On `main` the test fails at the first assertion on the empty group with a `KeyError` for `step`.

The changed files were formatted with the repository's pinned `black` 24.4.2 and the pre-commit arguments, and both are unchanged by it.

### DIFF
--- a/tests/pytorch/test_fused_optimizer.py
+++ b/tests/pytorch/test_fused_optimizer.py
@@ -166,6 +166,31 @@ class TestFusedAdam(TestFusedOptimizer):
 
             torch.testing.assert_close(ref_param, tst_param)
 
+    @pytest.mark.parametrize("capturable", [False, True])
+    def test_empty_param_group_advances_step(self, capturable):
+        # An empty param group must advance its step counter like a populated one.
+        # The same group can be empty on one data-parallel rank and populated on
+        # another, and "step" is part of the checkpoint, so a group that stops
+        # counting makes a resumed run apply a stale bias correction.
+        param = torch.nn.Parameter(torch.rand(4, dtype=torch.float, device="cuda"))
+        tst_optim = self.fused_optim(
+            [{"params": [param]}, {"params": []}], capturable=capturable, **self.options
+        )
+
+        num_steps = 3
+        for _ in range(num_steps):
+            param.grad = torch.rand_like(param)
+            tst_optim.step()
+
+        populated_step, empty_step = (int(g["step"]) for g in tst_optim.param_groups)
+        assert populated_step == num_steps
+        assert empty_step == populated_step
+
+        # The counter is checkpointed through the param groups, so the empty group
+        # has to round trip with the same value as the populated one.
+        checkpoint = tst_optim.state_dict()
+        assert int(checkpoint["param_groups"][1]["step"]) == num_steps
+
     def test_empty_param_at_end_of_group(self):
         tensors = [
             torch.ones(4, dtype=torch.float, device="cuda"),
--- a/transformer_engine/pytorch/optimizers/fused_adam.py
+++ b/transformer_engine/pytorch/optimizers/fused_adam.py
@@ -557,12 +557,11 @@ class FusedAdam(torch.optim.Optimizer):
             loss = closure()
 
         for group in self.param_groups:
-            if len(group["params"]) == 0:
-                continue
-            device = group["params"][0].device
-            bias_correction = 1 if group["bias_correction"] else 0
-            beta1, beta2 = group["betas"]
-
+            # Advance the step counter before skipping empty groups. A param group can be
+            # empty on some data-parallel ranks and populated on others, so incrementing
+            # only for populated groups desynchronizes "step" across the ranks that share
+            # an optimizer state shard. A rank then resumes from a checkpoint written by a
+            # rank where the group was empty and applies a stale bias correction.
             # assume same step across group now to simplify things
             # per parameter step can be easily support by making it tensor, or pass list into kernel
             if "step" in group:
@@ -570,9 +569,24 @@ class FusedAdam(torch.optim.Optimizer):
                     1 if not self.capturable else (self._dummy_overflow_buf != 1).to(torch.int)
                 )
             else:
-                group["step"] = (
-                    1 if not self.capturable else torch.tensor([1], dtype=torch.int, device=device)
+                # Empty groups have no parameter to take the device from, so fall back to
+                # the device of the optimizer's own scratch buffer.
+                step_device = (
+                    group["params"][0].device
+                    if len(group["params"]) > 0
+                    else self._dummy_overflow_buf.device
                 )
+                group["step"] = (
+                    1
+                    if not self.capturable
+                    else torch.tensor([1], dtype=torch.int, device=step_device)
+                )
+
+            if len(group["params"]) == 0:
+                continue
+            device = group["params"][0].device
+            bias_correction = 1 if group["bias_correction"] else 0
+            beta1, beta2 = group["betas"]
 
             # create lists for multi-tensor apply
             p_main_of_fp8_model = []


ptrendx (approved):

adityasingh2400:
Flagging that the three red build jobs all died before any test ran, so none of them is exercising this change.

```
JAX      failure   failed step: Build
PyTorch  failure   no failed step, Build never completed, Sanity check never ran
All      failure   no failed step, same shape
Core     still in progress
```

`JAX` is the only one with a real failed step, and it is a CUDA link error rather than anything Python:

```
fatbinary fatal : Could not open input file 'flash_attn.compute_75.cubin'
error: subprocess-exited-with-error
```

`PyTorch` and `All` are stranger. Both report `conclusion=failure` with **no failed step at all**, their `Build` step still showing `in_progress` and `Sanity check` never reached. `PyTorch` started at 00:24:27 and completed at 02:45:31, so it sat in `Build` for 2h21m before the job was marked failed. That reads like a lost runner or a timeout during the docker build rather than a test result.

This PR is 21 lines of Python in `fused_adam.py` plus a test. It cannot make a `.cubin` fail to link or a container build hang for two hours, and in the `PyTorch` job the sanity check that would import it never executed.

Not asking for anything, just did not want the three reds to read as coming from the change while `blossom-ci` and `te-ci` are still pending. Happy to rebase onto current main if you think a fresh run would help.

adityasingh2400:
Update on the run you triggered, since `Core` has now finished too.

```
JAX      failure    step 6 Build      died after 3m28s
PyTorch  failure    step 8 Build      still in_progress when killed, Sanity check never started
All      failure    step 8 Build      same shape
Core     cancelled  step 6 Build      cancelled at exactly 6h
```

All four died in Build, so still nothing here has run a test.

I also found the line just above the fatbinary error I quoted earlier, which points at the runner rather than at the code:

```
[  1%] Building CUDA object CMakeFiles/transformer_engine.dir/fused_attn/flash_attn.cu.o
fatbinary fatal   : Could not open input file 'flash_attn.compute_75.cubin'
sccache: Compiler killed by signal 1
gmake[2]: *** [CMakeFiles/transformer_engine.dir/build.make:76: ... flash_attn.cu.o] Error 254
```

The compiler was killed, so the cubin was never written. That is at 1 percent of the build, compiling `transformer_engine/common/fused_attn/flash_attn.cu`. This PR only touches `transformer_engine/pytorch/optimizers/fused_adam.py` and `tests/pytorch/test_fused_optimizer.py`, both Python, so it cannot reach that compile.

adityasingh2400:
Correcting my last comment. I said nothing here had run a test. That was true of the four `Build` jobs, but `te-ci` has since reported and it did run the suite, so this change is now exercised.

`test_fused_optimizer.py` is green everywhere it ran:

```
A100   28 passed,  7 skipped
L40    30 passed,  5 skipped
B200   35 passed
H100   no pytest failures in the log
```

The red `te-ci` jobs are failing elsewhere. On A100 and L40 almost all of it is one environment assert, 590 of 680 on A100 and 594 on L40:

```
assert arch // 10 in [9, 10, 11, 12], "Unsupported compute capability. Supported: 9.x, 10.x, 11.x, 12.x"
E   AssertionError: Unsupported compute capability. Supported: 9.x, 10.x, 11.x, 12.x
/usr/local/lib/python3.12/dist-packages/flash_attn/cute/interface.py:1353
```

A100 is 8.0 and L40 is 8.9, so both land on 8 and fail that check. The installed `flash_attn` cute path does not cover those two cards. B200 is different, 41 failures across `test_grouped_mlp.py`, `test_attention.py` and `test_kv_cache.py`.

I could not account for the H100 and H100 debug statuses. Their uploaded logs show no pytest failures at all, so whatever turned them red is outside the part I can read, and I did not want to guess.

ptrendx:
Rerunning CI with the updated branch.

adityasingh2400:
Flagging that the rerun did not finish. `PyTorch`, `All` and `Core` were all cancelled at 00:26:46, so the pytorch suite never completed on `95a1af2`.

`JAX` failed earlier, at 00:00:17, and it is a build error rather than a test one:

```
gmake[2]: *** [CMakeFiles/transformer_engine.dir/build.make:76:
CMakeFiles/transformer_engine.dir/fused_attn/flash_attn.cu.o] Error 254
```

This PR only touches `transformer_engine/pytorch/optimizers/fused_adam.py` and `tests/pytorch/test_fused_optimizer.py`, so that CUDA build is not coming from here. `JAX C++` and `JAX Python` both passed on the same commit.

Happy to do anything on my side that helps, otherwise it just needs another run once the build is sorted.

PR #3320, opened 2026-08-05
[Pytorch] Introduce Extra Tensor channels to enable resnet like connections within TE Sequential

vthumbe1503:
# Description

Fixes # (issue)

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Change A
- Change B

# Checklist:


docs/examples/op_fuser/op_fuser.rst
```diff
@@ -113,43 +113,115 @@ quantized compute.
...
-extra inputs or produce extra outputs. For example, ``AddExtraInput``
-will add an extra input tensor to the intermediate tensor and
-``MakeExtraOutput`` will return the intermediate tensor as an extra
-output. When calling a ``Sequential`` that contains any of these
-branching operations, the extra inputs should be passed in as
-arguments and the extra outputs will be returned.
+The operation fuser supports limited branching behavior. While the
+operations must be in sequential order, basic operations may declare
+extra tensor inputs and outputs. By default, an extra tensor slot has
+no channel assigned and is part of the public ``Sequential`` interface:
```
ptrendx: There should be some explanation what a channel is before we say "no channel assigned".
It could be as simple as saying that the extra inputs/outputs may optionally specify
a channel (and then proceed with the rest of the text).

docs/examples/op_fuser/op_fuser.rst
```diff
@@ -113,43 +113,115 @@ quantized compute.
...
+output slot and a later input slot connects them internally instead.
 
 .. code-block:: python
 
     import torch
     import transformer_engine.pytorch as te
 
-    # Construct MLP with residual connection
-    fc1 = te.ops.Sequential(
+    # Keep a residual connection inside one Sequential.
```
ptrendx: I would actually keep both of these examples here.

docs/examples/op_fuser/op_fuser.rst
```diff
@@ -113,43 +113,115 @@ quantized compute.
...
+    # Channels supply all later extra inputs internally.
+    y = moe(x, router_probs)
+
+The following conditions apply to extra tensor channels:
+
+- A producer must appear before all of its consumers. Backward edges
+  and cycles are not supported.
+- A channel has exactly one producer, but its output may fan out to
+  multiple consumers.
+- Every named output channel must have at least one consumer, and the
```
ptrendx: This limitation that there has to be at least one consumer in the named channel seems
arbitrary to me. If we do not strictly need this behavior then we shouldn't have that
as it would introduce friction when somebody needs to refactor the code using those
named channels by splitting the sequential - now they also need to remove the channel
names. In fact, I would expect people to generally want to name their extra outputs and
inputs even if they would not be reused inside the sequential. That could also enable
us to accept and return the dictionary rather than a list (which would make it less
fragile).

ptrendx (commented):
Additional review comments from Codex:
```
  1. [High] Internal channel outputs lose the signal that their gradient is required.
     transformer_engine/pytorch/ops/fuser.py:170 only calls requires_grad_ for public outputs at line 194. A fresh tensor created by an internal producer inside
     torch.autograd.Function.forward therefore arrives at its consumer with requires_grad=False. Existing operations such as transformer_engine/pytorch/ops/basic/
     swiglu.py:468 and ScaledSReLU use that flag to decide whether to compute the extra-input gradient. They consequently return None, silently dropping the
     gradient to a differentiable producer such as the documented router-probability dispatch. The new tests do not expose this because MakeExtraOutput returns the
     original input tensor.

  2. [Medium] Channel routing violates the declared iterable output contract.
     transformer_engine/pytorch/ops/fuser.py:140 applies len() and indexing to a producer’s extra outputs, but transformer_engine/pytorch/ops/op.py:92 permits any
     Iterable[Iterable[Tensor]]. A custom Dispatch returning a generator works under the old flattening logic but now fails when a later channel consumer executes.

  3. [Medium] A standalone operation call permanently prevents later channel configuration.
     transformer_engine/pytorch/ops/op.py:598 constructs a temporary OperationFuser, while transformer_engine/pytorch/ops/fuser.py:509 permanently locks every
     attached operation. Calling an operation once through its normal forward, then placing it into a channel-connected Sequential, makes either setter raise even
     though the temporary fuser no longer exists.

  4. [Medium] The tests do not exercise two major routing branches.
     tests/pytorch/test_fusible_ops.py:460 registers only a forward fusion, so backward remains unfused and never exercises the same-fusion skip at
     transformer_engine/pytorch/ops/fuser.py:323. The multi-output test at tests/pytorch/test_fusible_ops.py:624 only checks duplicate-name rejection; its custom
     operation never runs. Thus mixed bound/unbound slot ordering, filtered autograd returns, and the modified two-input GroupedLinear(scale_bias=True) behavior
     remain unproved.

  ## Suggested repairs

  - For finding 1: Preserve the gradient-requirement flag on every extra output before classifying it as public or internal, or carry equivalent explicit per-slot
    metadata. Add a producer that creates a fresh tensor and verify gradient propagation through ScaledSwiGLU or ScaledSReLU.

  - For finding 2: Materialize and validate each operation’s extra outputs as a tuple immediately after fuser_forward; store that tuple for later consumers and
    lifetime tracking.

  - For finding 3: Make transient fusers created by BasicOperation.forward non-locking, while persistent Sequential fusers retain immutable routing. Add a call-
    then-bind regression test.

  - For finding 4: Add a joint/backward fused residual operation and assert fusion selection plus input/parameter/channel gradients. Add a successful multi-input/
    multi-output routing test and a GroupedLinear(scale_bias=True) channel case.

```

I would like you to also take a look at the tests - multiple of them duplicate each other (e.g. test_channel_fan_out_accumulates_grad is a stronger duplicate of test_internal_extra_tensor_channel_fanout).

docs/examples/op_fuser/op_fuser.rst
vthumbe1503: Done

docs/examples/op_fuser/op_fuser.rst
vthumbe1503: Done

docs/examples/op_fuser/op_fuser.rst
vthumbe1503: Originally I wanted to restrict the channel naming as a way to just do internal routing of tensors to reduce possibility of errors and the friction was kind of intentional. But I see your point of making it more seamless for user in future to construct a big a sequential op. If they want to refactor a code from 1 to 2 below

1. single sequential having internal routing  
2. Two sequentials with one sequential passing extra output as extra input to another sequential

This can indeed be a problem since there might be some use-cases just supporting 2 but not 1.


And so I have removed that restriction. However, supporting extra_input and extra_output as dictionaries would be a problem from backwards compatibility perspective. Also, I want to restrict the scope of this PR. And allowing for dict based extra_input and extra output can be a seperate PR.


~~I have one extra requirement from named extra input channel added currently. If two different extra inputs share the same channel name, and is not internally connected to extra output of a previous op. Caller/User should still provide the extra_input two times.~~



~~This is done so that user's code doesnt have to change while naming an input channel vs not naming it. Also as you can see introducing extra_input dict is also going to make this tricky from backward compatibility perspective.~~

transformer_engine/pytorch/ops/basic/grouped_linear.py
```diff
@@ -147,11 +147,11 @@ def __init__(
         delay_wgrad_compute: bool = False,
         scale_bias: bool = False,
     ) -> None:
-        super().__init__()
-
+        # Decide before BasicOperation.__init__ sizes _extra_input_channels.
         self._scale_bias: bool = scale_bias and bias
         if self._scale_bias:
             self.num_extra_inputs = 2
+        super().__init__()
```
timmoon10, line 214: Is this working correctly? I recall having trouble setting attrs in `torch.nn.Module` before calling `__init__` since it does special handling for params and modules: https://github.com/pytorch/pytorch/blob/99ee33c9719e0ceeb9cbda00e4cdd12a76c9ae45/torch/nn/modules/module.py#L1973

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -118,7 +123,20 @@ def forward(
...
+            for idx in basic_op_idxs:
+                for input_idx, source in enumerate(fuser._basic_op_extra_input_sources[idx]):
+                    if source is None:
+                        continue
+                    producer_idx, output_idx = source
+                    if producer_idx in basic_op_idxs:
+                        # fused op will wire the channel itself internally
+                        continue
+                    producer_outputs = extra_outputs[producer_idx]
+                    basic_op_extra_inputs[idx][input_idx] = producer_outputs[output_idx]
```
timmoon10: Nit: The "Forward op" comment originally described the entire code block with `op.fuser_forward`, but it's become meaningless now that it's so far away.
```suggestion
            # Resolve internal channel inputs from outputs of
            # earlier basic ops. When a fusion contains both producer and
            # consumer, leave the consumer slot unset so the fused op can
            # wire the channel itself
            for idx in basic_op_idxs:
                for input_idx, source in enumerate(fuser._basic_op_extra_input_sources[idx]):
                    if source is None:
                        continue
                    producer_idx, output_idx = source
                    if producer_idx in basic_op_idxs:
                        # fused op will wire the channel itself internally
                        continue
                    producer_outputs = extra_outputs[producer_idx]
                    basic_op_extra_inputs[idx][input_idx] = producer_outputs[output_idx]

            # Prepare args for op forward
```

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -139,24 +157,30 @@ def forward(
...
+                if len(ys) != num_extra_outputs:
+                    raise RuntimeError(
+                        f"Expected op {idx} to generate {num_extra_outputs} extra outputs, "
+                        f"but got {len(ys)}"
+                    )
+                for output_idx, y in enumerate(ys):
+                    if (
+                        set_output_requires_grad
+                        and idx >= fuser.first_op_requiring_backward
+                        and (y.is_floating_point() or y.is_complex())
```
timmoon10: I too look forward to the day when complex-valued operations become mainstream in DL. FFT and eigendecomposition are cool. Until then, we can avoid some CPU overhead:
```suggestion
                        and y.is_floating_point()
```

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -224,29 +251,49 @@ def backward(
...
             if all(not basic_op_ctxs[idx].requires_grad for idx in basic_op_idxs):
                 dx = None
                 break
 
-            # Backward op
+            # Backward op. Supply gradients accumulated from every consumer of
+            # each internal channel.
+            for idx in basic_op_idxs:
+                for output_idx, channel in enumerate(basic_op_extra_output_channels[idx]):
+                    if basic_op_extra_output_is_internal[idx][output_idx]:
```
timmoon10: This is fine, but isn't `basic_op_extra_output_is_internal` redundant? If the channel is unused, `channel_grads.get(channel)` will return `None`, and we handle that case gracefully.

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -339,10 +400,93 @@ def __init__(
...
+        for op_idx, op in enumerate(basic_ops):
+            for input_idx, channel in enumerate(op._extra_input_channels):
+                if channel is None:
+                    self._external_extra_input_slots.append((op_idx, input_idx))
+                    continue
+                producer = channel_producers.get(channel)
+                # If no producer for named channel, this is a public input.
+                if producer is None:
+                    self._external_extra_input_slots.append((op_idx, input_idx))
+                    continue
```
timmoon10: We should error out if the channel is missing a producer. Handling this edge case adds complexity. It also makes the contract less intuitive without providing any benefit.

tests/pytorch/test_fusible_ops.py
```diff
@@ -437,6 +438,417 @@ def test_extra_tensors(self, size: int = 16) -> None:
...
+
+        consumer.set_extra_input_channel(0, None)
+        extra = torch.rand_like(x)
+        with pytest.raises(RuntimeError, match="Construct a new OperationFuser"):
+            model(x, extra)
+
+        model = te_ops.Sequential(producer, consumer)
+        y, route = model(x, extra)
+        torch.testing.assert_close(y, x + extra)
+        torch.testing.assert_close(route, x)
```
timmoon10: We shouldn't reach this:
```suggestion
```

tests/pytorch/test_fusible_ops.py
```diff
@@ -437,6 +438,417 @@ def test_extra_tensors(self, size: int = 16) -> None:
...
+        y, route = model(x, extra)
+
+        torch.testing.assert_close(y, 2 * x + extra)
+        torch.testing.assert_close(route, x)
+        dy = torch.rand_like(y)
+        y.backward(dy)
+        torch.testing.assert_close(x.grad, 2 * dy)
+        torch.testing.assert_close(extra.grad, dy)
+
+    def test_external_named_extra_inputs_remain_separate(self, size: int = 16) -> None:
```
timmoon10: We should remove this test if we require that input channels are valid.

tests/pytorch/test_fusible_ops.py
```diff
@@ -437,6 +438,417 @@ def test_extra_tensors(self, size: int = 16) -> None:
...
+            ops = [producer1, producer2, consumer]
+        else:
+            producer = _DualExtraOutput()
+            producer.set_extra_output_channel(0, "route")
+            producer.set_extra_output_channel(1, "route")
+            ops = [producer, consumer]
+        with pytest.raises(ValueError, match="multiple producers"):
+            OperationFuser(ops)
+
+    def test_named_extra_output_without_consumer_is_public(self, size: int = 16) -> None:
```
timmoon10: This is fine, but somewhat redundant with `test_mixed_channel_outputs_are_public`.

docs/examples/op_fuser/op_fuser.rst
```diff
@@ -151,6 +151,139 @@ arguments and the extra outputs will be returned.
...
+
+    # Dispatch's extra input has no channel, so the caller passes router_probs.
+    # Channels supply all later extra inputs internally, while Dispatch's
+    # extra outputs are still returned in their original order.
+    y, m_splits, probs, routing_map = moe(x, router_probs)
+
+Channels cannot connect operations in different ``OperationFuser``
+instances. In particular, an ordinary PyTorch module inside a
+``Sequential`` splits the fusible operations on either side into
+separate fusers. The following channel connection is therefore not
```
timmoon10, line 244: It would be nice if `Sequential` could handle channels across `OperationFuser`s, but the implementation would be quite hairy and not worth it for the current effort.

docs/examples/op_fuser/op_fuser.rst
```diff
@@ -151,6 +151,139 @@ arguments and the extra outputs will be returned.
...
+"""""""""""""""""""""
+
+Extra inputs and Extra outputs may optionally specify a channel. Assigning
+the same channel name to an extra output and one or more later extra
+inputs routes the tensor internally within the same
+``OperationFuser``. An extra input connected to an earlier producer is
+removed from the public ``Sequential`` arguments because the channel
+supplies it.
+Extra outputs remain in the public ``Sequential`` return value,
+including outputs that are also consumed through a channel.
```
timmoon10: If you read pendantically, then this technically describes our current behavior where we ignore unmatched input channels. However, this is subtle and non-obvious. Better to error out if the input channel is invalid.

```suggestion
Branching operations can also route their extra inputs and outputs within the same ``Sequential`` via named channels. Extra output tensors with a specified channel can be consumed by other operations, in addition to being returned from the ``Sequential``. Extra input tensors with a specified channel are accessed internally instead of being provided as arguments to ``Seqential``.
```

transformer_engine/pytorch/ops/basic/grouped_linear.py
vthumbe1503: Yes, this seems to be working fine and the tests are passing. Maybe this was a bug in previous pytorch version? Or maybe this happens if you set special attributes like nn.parameter, nn.module etc?

tests/pytorch/test_fusible_ops.py
vthumbe1503: Why not? The intent of the test is to make sure that we are able to construct a new sequential with the old ops used for the previous sequential, even after the channel wiring between producer/consumer ops have changed.

Every OperationFuser, when initialized, also maintains a version for the channel-wiring being done. The moment channel wiring changes, old OperationFuser instances become unusable and any newly created OperationFuser should still work fine. And that is what is being tested

timmoon10: Ah, good point. I thought the `pytest.raises` would kill the test, but acutally there's non-trivial stuff afterwards.

tests/pytorch/test_fusible_ops.py
vthumbe1503: Done

transformer_engine/pytorch/ops/fuser.py
vthumbe1503: Removed it

transformer_engine/pytorch/ops/fuser.py
vthumbe1503: Agreed. It was indeed very uncomfortable to add this requirement. My only reason originally to allow for unnamed extra input channels was to keep it symmetric with exra_output channel convention. But there is intentional assymetry there. And erroring out for makes sense to me

tests/pytorch/test_fusible_ops.py
vthumbe1503: Yep test_mixed_channel_outputs_are_public is a superset. removed this one

docs/examples/op_fuser/op_fuser.rst
vthumbe1503: Update: Tim has a valid point below of extra_input's complexity. For the named extra_input which is not connected to any producer op, we are simply raising an error. Its users responsibility to unname the channel if it is already named(by setting it to None) and is not connected to any internal producer.

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -507,6 +702,12 @@ def __call__(
         *extra_inputs: torch.Tensor,
         basic_op_kwargs: Optional[list[dict[str, Any]]] = None,
     ) -> torch.Tensor | tuple[torch.Tensor, ...]:
+        if self.has_stale_op_channels():
+            raise RuntimeError(
+                "Extra tensor channels changed after this OperationFuser captured "
+                "its routing. Construct a new OperationFuser."
+            )
```
ptrendx: Do we need to do this on every call? We could just make it impossible to change them with
the API itself (e.g. have the fuser mark the ops as finalized when it takes them).

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -139,24 +161,53 @@ def forward(
...
+                            raise RuntimeError(
+                                f"Op {idx} extra output {output_idx} is public, "
+                                f"but {type(op).__name__} returned None"
+                            )
+                        if needed_outside_fusion:
+                            raise RuntimeError(
+                                f"Op {idx} extra output {output_idx} is required by an "
+                                "operation outside its forward fusion, "
+                                f"but {type(op).__name__} returned None"
+                            )
```
ptrendx, line 196: This feels like something that we should not need to do every time we call this
autograd function - those would not change unless a different fusion was used.

ptrendx (approved):

vthumbe1503:
Pipeline: 62579701

PR #3325, opened 2026-08-07
Prototype NVFP4 with FP8 UE5M3 block scales

timmoon10:
# Description

This is a proof-of-concept for NVFP4 with FP8 UE5M3 block scales. Quantization is supported natively and GEMMs go through the cuDNN Frontend kernels (see https://github.com/NVIDIA/cudnn-frontend/pull/545).

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

## Changes

- Support NVFP4 tensors with UE5M3 scales
- Support UE5M3 scales in NVFP4 quantize/dequantize kernels
- Support UE5M3 scales in NVFP4 quantizers

# Checklist:


transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -188,6 +190,496 @@ def _validate_native_gemm_output_quantizer(quantization_params):
...
+
+    if bias is not None:
+        assert bias.dim() == 1 and bias.shape[0] == M, (
+            f"cuDNN MX GEMM expects a ({M},) bias, but got {tuple(bias.shape)}."
+        )
+        # cuDNN checks the stride literally, so (1, N) rather than reshape's (1, 1).
+        bias = bias.contiguous().as_strided((M, 1), (1, M))
+
+    # Prepare for output
+    out = validate_or_alloc_output(out, out_shape, out_dtype, device)
```
kainzhong: You can remove this line now since you've done this at line 530 just before "Trivial cases"

transformer_engine/common/include/transformer_engine/recipe.h
```diff
@@ -14,7 +14,10 @@
 #include "transformer_engine.h"
 
 #ifdef __cplusplus
+#define NVTE_NVFP4_SCALE_DTYPE_DEFAULT = kNVTEFloat8E4M3
 extern "C" {
+#else
+#define NVTE_NVFP4_SCALE_DTYPE_DEFAULT
 #endif
```
timmoon10, line 18: Won't this break ABI compatibility if accessing as C functions? It's also quite hacky and nonstandard.

We should check, but I believe these are advanced functions that are only used for our Mcore DDP integration. If this is true, I think it's fine to break backward compatibility and just add the scale dtype as a normal function arg.

transformer_engine/pytorch/cpp_extensions/gemm.py
```diff
@@ -188,6 +190,507 @@ def _validate_native_gemm_output_quantizer(quantization_params):
...
+    alpha = alpha if alpha is not None else 1.0
+    # cuDNN's general GEMM path (grouped_gemm_quant_wrapper_sm100) doesn't support accumulation
+    assert (
+        accumulate is False
+    ), "cuDNN GEMM currently does not support accumulation for this operation."
+    assert beta in (0.0, None), "beta must be zero or None if not accumulate"
+
+    # cuDNN's grouped quant kernel requires M to be divisible by 256 so we need to pad it
+    N_padded = ceil_div(N, 256) * 256
+    if N_padded != N:
```
timmoon10: All of this padding logic is unnecessary since unaligned data goes to the wgrad kernel.

transformer_engine/common/include/transformer_engine/recipe.h
tdophung: yesh it does break ABI. Since I did not want to create separate function for each (like a `_v2`) with the extra scale_dtype arg, I resorted to this thinking that it is a nvfp4-plus-devel branch specifically anyways so it should be ok to assume that people that intentionally uses it (Mcore) would always pass in the scale_dtype.

But i think you are right, that it is hacky/deceiving, so I'll just make it a normal arg

transformer_engine/common/include/transformer_engine/recipe.h
```diff
@@ -374,32 +374,34 @@ void nvte_nvfp4_2d_compute_partial_amax(const NVTETensor inp, NVTETensor amax, s
  *  \param[in]     start_offset    Starting element offset in the flattened tensor.
  *  \param[in]     block_len       Tile dimension (must be 16 for NVFP4 2D).
  *  \param[in]     stream          CUDA stream used for the operation.
+ *  \param[in]     scale_dtype     NVFP4 scale storage type (E4M3 or UE5M3).
  */
 void nvte_nvfp4_2d_partial_cast(const NVTETensor inp, NVTETensor out, const NVTETensor scale,
                                 const NVTETensor global_scale, size_t h, size_t w,
                                 size_t scale_stride_h, size_t scale_stride_w, size_t start_offset,
-                                size_t block_len, cudaStream_t stream);
+                                size_t block_len, cudaStream_t stream, const NVTEDType scale_dtype);
```
timmoon10: Nit: Since we're breaking backward compatibility anyways, it would be nicer to keep our convention where `stream` is the last arg.

transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -812,14 +879,33 @@ def fuse_grouped_mlp_ops(
...
+
+    # Fused kernels are only supported for MXFP8 and NVFP4
+    if recipe is None:
         return ops
-    # NVFP4 fused grouped MLP uses graph-safe grouped quantize, which currently requires RHT.
-    if recipe.nvfp4() and recipe.disable_rht:
+    if recipe.custom():
+        # Check if custom recipe explicitly enables fusion
+        if not getattr(recipe, "enable_cutedsl_fused_grouped_mlp", False):
+            return ops
```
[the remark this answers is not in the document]
timmoon10: This is an explicit opt-in, so I think it's reasonable to expect the user knows what they're doing.

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -121,8 +127,11 @@ void quantize_fwd_helper(const NVTETensor input, NVTETensor output,
                 (dtype == DType::kBFloat16 && rows % 32 == 0 && cols % 32 == 0),
             "Row-scaled NVFP4 transpose quantization requires BF16 input and dimensions that are "
             "multiples of 32.");
-        nvfp4::compute_rowwise_amax(*input_tensor, noop_tensor, output_tensor, stream);
-        if (output_tensor->has_columnwise_data()) {
+        if (output_tensor->amax.dptr != nullptr) {
```
ptrendx: You check against that condition just before.

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -298,8 +313,11 @@ void quantize_bwd_helper(const NVTETensor grad, const NVTETensor input, NVTETens
                 (dtype == DType::kBFloat16 && rows % 32 == 0 && cols % 32 == 0),
             "Row-scaled NVFP4 transpose quantization requires BF16 input and dimensions that are "
             "multiples of 32.");
-        nvfp4::compute_rowwise_amax(*grad_tensor, noop_tensor, output_tensor, stream);
-        if (output_tensor->has_columnwise_data()) {
+        if (output_tensor->amax.dptr != nullptr) {
```
ptrendx: Same.

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -121,8 +127,11 @@ void quantize_fwd_helper(const NVTETensor input, NVTETensor output,
                 (dtype == DType::kBFloat16 && rows % 32 == 0 && cols % 32 == 0),
             "Row-scaled NVFP4 transpose quantization requires BF16 input and dimensions that are "
             "multiples of 32.");
-        nvfp4::compute_rowwise_amax(*input_tensor, noop_tensor, output_tensor, stream);
-        if (output_tensor->has_columnwise_data()) {
+        if (output_tensor->amax.dptr != nullptr) {
+          nvfp4::compute_rowwise_amax(*input_tensor, noop_tensor, output_tensor, stream);
+        }
```
timmoon10: https://github.com/NVIDIA/TransformerEngine/pull/3325/changes#r3850769689:
```suggestion
        nvfp4::compute_rowwise_amax(*input_tensor, noop_tensor, output_tensor, stream);
```

transformer_engine/common/cast/dispatch/quantize.cuh
```diff
@@ -298,8 +313,11 @@ void quantize_bwd_helper(const NVTETensor grad, const NVTETensor input, NVTETens
                 (dtype == DType::kBFloat16 && rows % 32 == 0 && cols % 32 == 0),
             "Row-scaled NVFP4 transpose quantization requires BF16 input and dimensions that are "
             "multiples of 32.");
-        nvfp4::compute_rowwise_amax(*grad_tensor, noop_tensor, output_tensor, stream);
-        if (output_tensor->has_columnwise_data()) {
+        if (output_tensor->amax.dptr != nullptr) {
+          nvfp4::compute_rowwise_amax(*grad_tensor, noop_tensor, output_tensor, stream);
+        }
```
timmoon10: https://github.com/NVIDIA/TransformerEngine/pull/3325/changes#r3850773240:
```suggestion
        nvfp4::compute_rowwise_amax(*grad_tensor, noop_tensor, output_tensor, stream);
```

transformer_engine/common/include/transformer_engine/recipe.h
```diff
@@ -373,81 +373,88 @@ void nvte_nvfp4_2d_compute_partial_amax(const NVTETensor inp, NVTETensor amax, s
...
  *  \param[in]     start_offset    Starting element offset in the flattened tensor.
  *  \param[in]     block_len       Tile dimension (must be 16 for NVFP4 2D).
+ *  \param[in]     scale_dtype     NVFP4 scale storage type (E4M3 or UE5M3).
  *  \param[in]     stream          CUDA stream used for the operation.
  */
 void nvte_nvfp4_2d_partial_cast(const NVTETensor inp, NVTETensor out, const NVTETensor scale,
                                 const NVTETensor global_scale, size_t h, size_t w,
                                 size_t scale_stride_h, size_t scale_stride_w, size_t start_offset,
-                                size_t block_len, cudaStream_t stream);
+                                size_t block_len, NVTEDType scale_dtype, cudaStream_t stream);
```
ptrendx: Why do we need that API break? We have scale tensor there which has dtype inside it, then why do we need to pass it also as an additional argument?

ptrendx: In general, we should think very carefully here why we would do this (and if really needed then I would much rather have v2 variants rather than break the APi).

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -23,18 +23,19 @@ extern "C" {
...
-  kNVTEInt64 = 3,       /*!< 64-bit integer */
-  kNVTEFloat32 = 4,     /*!< 32-bit float */
-  kNVTEFloat16 = 5,     /*!< 16-bit float (E5M10) */
-  kNVTEBFloat16 = 6,    /*!< 16-bit bfloat (E8M7) */
-  kNVTEFloat8E4M3 = 7,  /*!< 8-bit float (E4M3) */
-  kNVTEFloat8E5M2 = 8,  /*!< 8-bit float (E5M2) */
-  kNVTEFloat8E8M0 = 9,  /*!< 8-bit float (E8M0) */
-  kNVTEFloat4E2M1 = 10, /*!< 4-bit float (E2M1) */
-  kNVTENumTypes         /*!< Number of supported types */
+  kNVTEByte = 0,         /*!< Byte */
```
ptrendx, line 26: Could we move those comments more to the right, so that when we add even more types only the new type the diff would be easier to assess?

tests/cpp/operator/test_cast_nvfp4_transpose.cu
```diff
@@ -1477,3 +1577,142 @@ INSTANTIATE_TEST_SUITE_P(
         }
         return name;
     });
+
+#if CUDA_VERSION >= 13040 && FP4_TYPE_SUPPORTED
+INSTANTIATE_TEST_SUITE_P(
```
ptrendx, line 1576: Why do we only test it on a single 256x256 shape?

tests/cpp/test_common.cu
```diff
@@ -49,6 +49,9 @@ bool areShapesEqual(const NVTEShape &s1, const NVTEShape &s2) {
 }
 
 size_t typeToNumBits(DType type) {
+  if (type == DType::kFloat8UE5M3) {
+    return 8;
+  }
```
ptrendx: Huh? Why do we need to specialcase it?

tests/cpp/test_common.cu
```diff
@@ -374,6 +378,14 @@ Tensor::Tensor(const std::string& name,
...
       // Block scaling factors
       auto [rowwise_scale_meta, colwise_scale_meta] = get_scales(flattened_shape, tensor_.scaling_mode());
+      if (scaling_mode == NVTE_NVFP4_1D_SCALING) {
+        NVTE_CHECK(scale_dtype == DType::kFloat8E4M3 ||
+                   scale_dtype == DType::kFloat8UE5M3);
+        rowwise_scale_meta.type = scale_dtype;
+        rowwise_scale_meta.type_size_bits = typeToNumBits(scale_dtype);
+        colwise_scale_meta.type = scale_dtype;
+        colwise_scale_meta.type_size_bits = typeToNumBits(scale_dtype);
+      }
```
ptrendx: This should be part of the get_scales function, no?

tests/pytorch/nvfp4/test_nvfp4_quantize_exact.py
```diff
@@ -240,6 +241,65 @@ def check_quantization_nvfp4_versus_reference(
     torch.testing.assert_close(qx_amax, ref_amax, atol=0.0, rtol=0.0)
 
 
+@pytest.mark.skipif(not recipe_available, reason=reason_for_no_recipe)
+@pytest.mark.parametrize("return_transpose", [False, True], ids=["rowwise", "with_columnwise"])
+@pytest.mark.parametrize("use_4over6", [False, True], ids=["standard", "4over6"])
+def test_disable_second_level_scale_uses_only_block_scale(
```
ptrendx, line 259: Not really for this test, but for this file - we need to actually test E5M# here - this includes creating the reference implementation just like we have for regular nvFP4 recipe.

tests/pytorch/test_fusible_ops.py
```diff
@@ -133,16 +136,19 @@ def maybe_skip_quantization(
...
                 pytest.skip("FP8 block scaling requires dims that are divisible by 128")
-        elif quantization in ("nvfp4", "nvfp4_row_scaled", "nvfp4_4over6", "nvfp4_rht"):
+        elif quantization in nvfp4_variant_names:
             if math.prod(dims[:-1]) % 16 != 0 or dims[-1] % 16 != 0:
                 pytest.skip("NVFP4 GEMMs require dims that are divisible by 16")
+            if quantization in ("nvfp4_ue5m3", "nvfp4_rht_ue5m3") and (
+                math.prod(dims[:-1]) % 64 != 0 or dims[-1] % 64 != 0
+            ):
+                pytest.skip(
+                    "cuDNN FE NVFP4-UE5M3 GEMMs produce incorrect values with 32x32 tensors"
```
ptrendx: Multiple issues here:
 - generally speaking this is a problem that we do not support tensors not divisible by 64 here
 - the test should not actually be skipped here but instead test that the execution fails with a proper error message telling users what the issue is - otherwise there could be a case where they get silent wrong answer.

tests/pytorch/test_grouped_mlp.py
```diff
@@ -261,16 +264,19 @@ def maybe_skip_quantization(
...
                 pytest.skip("MXFP8 GEMMs require dims that are divisible by 32")
-        elif quantization in ("nvfp4", "nvfp4_row_scaled", "nvfp4_4over6", "nvfp4_rht"):
+        elif quantization in nvfp4_variant_names:
             if math.prod(dims[:-1]) % 16 != 0 or dims[-1] % 16 != 0:
                 pytest.skip("NVFP4 GEMMs require dims that are divisible by 16")
+            if quantization in ("nvfp4_ue5m3", "nvfp4_rht_ue5m3") and (
+                math.prod(dims[:-1]) % 64 != 0 or dims[-1] % 64 != 0
+            ):
+                pytest.skip(
+                    "cuDNN FE NVFP4-UE5M3 GEMMs produce incorrect values with 32x32 tensors"
```
ptrendx: Same comment as in the nongrouped case.

tests/pytorch/test_recipe.py
```diff
@@ -564,7 +564,7 @@ def expected_use_4over6(tensor_type):
 
     def expected_e4m3_max(tensor_type):
         if not expected_use_4over6(tensor_type):
-            return 448
+            return 0
```
ptrendx, line 567: Why? Is the result of this function something else than e4m3 max? If so, the name of the function should change too.

transformer_engine/common/util/pybind_helper.h
```diff
@@ -23,7 +23,9 @@
       .value("kBFloat16", transformer_engine::DType::kBFloat16)                                    \
       .value("kFloat8E4M3", transformer_engine::DType::kFloat8E4M3)                                \
       .value("kFloat8E5M2", transformer_engine::DType::kFloat8E5M2)                                \
+      .value("kFloat8E8M0", transformer_engine::DType::kFloat8E8M0)                                \
```
ptrendx, line 26: Why do we put it in the middle of this enum type? That would be breaking for places that have this enum as int.

transformer_engine/pytorch/tensor/storage/grouped_tensor_storage.py
```diff
@@ -850,6 +898,8 @@ def make_grouped_tensor(
                 )
                 columnwise_amax = torch.empty(num_tensors, dtype=torch.float32, device=device)
         elif compatible_recipe.float8_block_scaling():
+            scale_inv_dtype = DType.kFloat32
```
ptrendx, line 904: Earlier in this file amax buffers are always created, whereas it should check whether the second level scale should be disabled.

transformer_engine/pytorch/ops/fused/grouped_mlp.py
```diff
@@ -1355,47 +1451,65 @@ def fuser_forward(
...
+                    # Kernel has large numerical error with bias
+                    pass
+                elif self._cudnn_act_func == "swiglu":
+                    kernel_impl = "gemm_act_rht_quant"
+            elif fc2_input_quantizer.with_post_rht_amax:
+                # Use GEMM + act + RHT + amax kernel if available
+                if self.grouped_gemm_act_hadamard_kernel() is None:
+                    # Kernel is not available
+                    pass
+                elif self._cudnn_act_func in ("swiglu", "situlu"):
```
ptrendx: situglu

ptrendx (commented):
One more comment from Codex:
  [Medium] The new unaligned GEMM fallback is never tested.

     Production routes N % 256 != 0 to the wgrad wrapper at /tmp/te-pr3325/transformer_engine/pytorch/cpp_extensions/gemm.py:578. The test helper skips every corresponding case at /tmp/te-pr3325/tests/pytorch/nvfp4/
     test_nvfp4_gemm_exact.py:699, including all apparently targeted nonaligned tuples.

     Fix: retain targeted nonaligned cases and assert that grouped_gemm_wgrad_wrapper_sm100 executes.

transformer_engine/common/include/transformer_engine/recipe.h
timmoon10: We did think about maintaining backward compatibility, but I'm not sure if it's worth the effort since these are advanced functions used specifically for Mcore DDP integration: https://github.com/NVIDIA/TransformerEngine/pull/3325#discussion_r3779295465

transformer_engine/common/util/pybind_helper.h
timmoon10: This matches the dtype order in the C API: https://github.com/timmoon10/TransformerEngine/blob/58dfc41046b7068dffc29eba80ecf65005988009/transformer_engine/common/include/transformer_engine/transformer_engine.h#L25
Also, we explicitly specify the value for each Python enum, so the order doesn't matter.

tests/pytorch/test_recipe.py
timmoon10: There are two changes to how we're interpreting `e4m3_max`:
- It is now an _override_ for the maximum scale dtype value. A value of zero means we use the actual dtype max. This is only used for 4over6, which is not supported with e5m3.
- We can no longer assume that the NVFP4 scales are e4m3. I agree it would be better to rename the variable to something like `scale_max` so it's no longer married to e4m3. However, I'm trying to limit the amount this PR touches 4over6 logic.

transformer_engine/common/recipe/__init__.py
zhongbozhu, line 492: If it's just one UE5M3 toggle, we can actually extend it to `NVFP4BlockScaling` here so that it automatically inherits the Mcore `--fp4-param-gather` feature. I agree that real hybrid recipes like mixing with mxfp8 should continue to live in CustomRecipe structure.

timmoon10: You're right that basic UE5M3 support just requires setting `recipe.fp8_format = Format.UE5M3`. This can already be done in Mcore by setting `config.fp8_recipe_attrs = {"fp8_format": Format.UE5M3}` (once https://github.com/NVIDIA/Megatron-LM/pull/6341 merges).

However, the real practical benefit is that UE5M3 allows us to change the tensor scaling. We skip tensor scaling in the forward pass, while keeping current scaling in the backward pass. This is all a research question and I figure the recipe recommendations might change at any moment, so I'm resistant to hard-coding it in TE. `CustomRecipe` is a much better approach for unstable research recipes.

transformer_engine/common/include/transformer_engine/transformer_engine.h
```diff
@@ -83,11 +84,12 @@ enum NVTETensorParam {
    *  its values are populated during quantization.
    */
   kNVTERowScaledNVFP4 = 8,
-  /*! Global E4M3 scale bound used by an NVFP4 tensor.
+  /*! Global scale-bound selector used by an NVFP4 tensor.
    *
    *  This is part of the tensor data contract. Downstream dequantization and
    *  GEMM scale consumers must use the same bound used during quantization.
    *  Standard NVFP4 uses 448; 4over6 may use 256 for map-to-4 headroom.
+   *  For UE5M3 scales, these settings map to 114688 and 65536, respectively.
```
timmoon10:
```suggestion
```

tests/cpp/operator/test_cast_nvfp4_transpose.cu
timmoon10: Expanded to match shapes for other NVFP4 tests.

tests/pytorch/test_fusible_ops.py
timmoon10: Good point. My thought was that this wasn't really the main point of this test, it's more the job of the linear op tests, but this is bad practice in any case.

transformer_engine/common/include/transformer_engine/recipe.h
ptrendx: Those are part of our API, so even if we believe that only MCore uses it at the moment we should not break them since there may be some other customer of them that we do not know about. Public API stability is very important for the developer experience.

transformer_engine/common/recipe/__init__.py
zhongbozhu: I think for nvfp4 recipe, upgrading to ue5m3 even for non-hybrid recipes should be a no-brainer, because e2m1 has a sign bit, and e4m3 doesn't need a sign bit to begin with, so switch to ue5m3 for the nvfp4 recipe should also be needed since it should improve the numerics generally.

transformer_engine/common/gemm/cublaslt_gemm.cu
tdophung, line 391: for jax, alpha is an XLA device buffer, not on CPU, so this seem to be what causes the crashes here: https://gitlab-master.nvidia.com/dl/transformerengine/transformerengine/-/jobs/421075125. when we try to dereference this alpha ptr on the host. This is fine on Pytorch side though.

See jax side:
https://github.com/NVIDIA/TransformerEngine/blob/abf0ebb079b666ea4a37af9f5b9a36adb854d0db/transformer_engine/jax/csrc/extensions/gemm.cpp#L318

timmoon10: Hm, we are ping-ponging. https://github.com/NVIDIA/TransformerEngine/pull/3325/changes/adef2dc83164df4767c12324b0e184876411e199 changed to always call`nvte_nvfp4_compute_per_tensor_scale` and https://github.com/NVIDIA/TransformerEngine/pull/3325/changes/b3415c19c579ad756f6f2eaac00a82d3e4f8b71d reverted that change. The problem:
- Our cuBLAS integration assumes alpha is a GPU pointer:
https://github.com/NVIDIA/TransformerEngine/blob/8589b15c38da1c202deb38b91fcaeddab5f85ddb/transformer_engine/common/gemm/cublaslt_gemm.cu#L552-L554
- PyTorch is passing in a CPU alpha pointer. If we don't call`nvte_nvfp4_compute_per_tensor_scale`, cuBLAS fails to read alpha on GPU.
- JAX is passing in a GPU alpha pointer. If we call `nvte_nvfp4_compute_per_tensor_scale`, it segfaults trying to read alpha on CPU. 

The root cause is that we have a secret contract that not providing the amax implies alpha is a GPU pointer. This is unintuitive and unmaintainable. I think the right design is to make the contract explicit, perhaps by adding an config option to specify that alpha and beta are on device: https://github.com/NVIDIA/TransformerEngine/blob/8589b15c38da1c202deb38b91fcaeddab5f85ddb/transformer_engine/common/include/transformer_engine/gemm.h#L31

Alternatively, we could see if there's a nice way to query if the pointer is CPU or GPU, but I don't like that this will probably add non-trivial overhead.

tdophung: That makes sense to me. I did not look closer into why we had this contract at first, was just noticing that if no-amax is provided, then alpha is a GPU ptr. Having an attritbute would be way clearer

transformer_engine/jax/csrc/extensions/gemm.cpp
```diff
@@ -325,6 +325,7 @@ Error_Type GemmV2FFI(cudaStream_t stream, Buffer_Type lhs, Buffer_Type lhs_scale
   transformer_engine::MatmulConfigWrapper matmul_config;
   matmul_config.set_use_split_accumulator(config.use_split_accumulator);
   matmul_config.set_sm_count(num_math_sm);
+  matmul_config.set_alpha_beta_on_device(is_nvfp4_scaling(config.scaling_mode));
```
jberchtold-nvidia, line 328: @tdophung can you check with Ilia if E5M3 has been added as an XLA type yet? If so, can you add an assert in this file that the scale dtype cannot be E5M3 since we don't support it yet.

If it hasn't been added, then this PR is fine as is

transformer_engine/common/include/transformer_engine/gemm.h
```diff
@@ -54,6 +54,8 @@ enum NVTEMatmulConfigAttribute {
   kNVTEMatmulConfigUseSplitAccumulator = 5,
   /*! Number of streaming multiprocessors to use in GEMM kernel. */
   kNVTEMatmulConfigSMCount = 6,
+  /*! Whether alpha and beta are device pointers. Default: false. */
+  kNVTEMatmulConfigAlphaBetaOnDevice = 7,
```
timmoon10:
```suggestion
  /*! Whether alpha and beta are device pointers. Default: false.
   *
   *  Known bugs: only supported with NVFP4, NVFP4 amaxes are ignored when alpha and beta are device pointers.
   * 
   *  \todo Generalize to all tensor formats
   *  \todo Correctly handle NVFP4 amaxes when alpha and beta are device pointers
   */
  kNVTEMatmulConfigAlphaBetaOnDevice = 7,
```

transformer_engine/common/gemm/cublaslt_gemm.cu
```diff
@@ -356,14 +370,11 @@ void cublas_gemm(const Tensor *inputA, const Tensor *inputB, Tensor *outputD,
   const bool gelu = pre_gelu_out != nullptr;
   const bool use_fp8 = is_fp8_dtype(param.Atype) || is_fp8_dtype(param.Btype);
   const bool use_fp4 = is_fp4_dtype(param.Atype) || is_fp4_dtype(param.Btype);
+  const bool nvfp4_tensor_scaling =
+      is_nvfp_scaling(inputA->scaling_mode) && is_nvfp_scaling(inputB->scaling_mode);
 
   // Update scaling factors with NVFP4 tensor scales
```
timmoon10, line 376:
```suggestion
  // Update scaling factors with NVFP4 tensor scales
  // TODO Fix bug where amax is ignored when alpha and beta are on device
```

transformer_engine/jax/csrc/extensions/gemm.cpp
jberchtold-nvidia: E5M3 doesn't yet exist in JAX/XLA, resolving

jberchtold-nvidia (dismissed):
Approving for code owner permissions for the two TE/JAX files. Thanks!

negvet:
Since `disable_second_level_scale` does not belong to the recipe object, what do you think about adding `nvfp4_ue5m3_single_level_input_factory` to `quantizer_factory_zoo.py` as an experimental recipe which sets `disable_second_level_scale=True` for X/X.T?

timmoon10:
> Since `disable_second_level_scale` does not belong to the recipe object, what do you think about adding `nvfp4_ue5m3_single_level_input_factory` to `quantizer_factory_zoo.py` as an experimental recipe which sets `disable_second_level_scale=True` for X/X.T?

That's not a bad idea, although I'd like to add some more NVFP4 variants (beyond the scope of this PR) first.

ptrendx (approved):

PR #3327, opened 2026-08-07
[PyTorch] Fix deferred initialization in fusible ops

### LINKED ISSUE #3322: [Bug] Deferred init issue in `te.pytorch.ops`
**Describe the bug**

`OperationFuser` caches Parameter objects during construction at [fuser.py:357](https://github.com/NVIDIA/TransformerEngine/blob/a7aec214eb5c3969984a40c3accb6d66987d8f25/transformer_engine/pytorch/ops/fuser.py#L357). On the first forward, it calls `reset_recipe_state()` and then `pre_first_fuser_forward()` at [fuser.py:461](https://github.com/NVIDIA/TransformerEngine/blob/a7aec214eb5c3969984a40c3accb6d66987d8f25/transformer_engine/pytorch/ops/fuser.py#L461), but never refreshes the cached parameters before passing them into autograd at [fuser.py:544](https://github.com/NVIDIA/TransformerEngine/blob/a7aec214eb5c3969984a40c3accb6d66987d8f25/transformer_engine/pytorch/ops/fuser.py#L544).

The following existing ops replace meta Parameters with new CUDA Parameter objects during deferred initialization:

- `BasicLinear`
- `Bias`
- `LayerNorm`
- `RMSNorm`
- `GroupedLinear`

I reproduced the problem for all five on an H100 using BF16 deferred initialization. Forward successfully materializes the parameters on CUDA, but backward fails with:

```text
RuntimeError: Function _OperationFuserAutogradFunctionBackward returned an
invalid gradient at index 1 - expected device meta but got cuda:0
```

Quantized deferred initialization has the same problem: `BasicLinear` correctly materializes its weight as a CUDA `Float8Tensor`, but backward still returns a CUDA gradient for the stale meta Parameter. Therefore, `reset_recipe_state()` must remain before `pre_first_fuser_forward()`.

`ops.Linear` has an additional issue. It registers top-level `weight` and `bias` aliases at [linear.py:143](https://github.com/NVIDIA/TransformerEngine/blob/a7aec214eb5c3969984a40c3accb6d66987d8f25/transformer_engine/pytorch/ops/linear.py#L143). After deferred initialization, the internal basic-op parameters are CUDA tensors, while `op.weight` and `op.bias` still reference the old meta tensors. Refreshing the fuser cache fixes the autograd mismatch, but these aliases must also be synchronized.

The current tests do not cover this path: `test_linear` defaults to `device="cuda"` at [test_fusible_ops.py:1106](https://github.com/NVIDIA/TransformerEngine/blob/a7aec214eb5c3969984a40c3accb6d66987d8f25/tests/pytorch/test_fusible_ops.py#L1106), and there is no meta-device deferred-init case.

In short, this should be treated as a pre-existing, general `OperationFuser` deferred-initialization bug. The generic fix is to refresh the flattened basic-op parameter cache immediately after all first-forward materialization hooks, plus separately refresh the public parameter aliases in `ops.Linear`.

Please list *minimal* steps or code snippet for us to be able to reproduce the bug.

A  helpful guide on on how to craft a minimal bug report  http://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports. 

**Expected behavior**

**Environment overview (please complete the following information)**

 - Environment location: [Bare-metal, Docker, Cloud(specify cloud provider - AWS, Azure, GCP, Collab)]
 - Method of Transformer Engine install: [pip install or from source]. Please specify exact commands you used to install.
 - If method of install is [Docker], provide `docker pull` & `docker run` commands used

**Environment details**

Otherwise, please provide:
- OS version
- PyTorch version
- Python version
- Transformer Engine version
- CUDA version
- CUDNN version

**Device details**
- GPU model

**Additional context**

denera:
# Description

Ops replace their params when materializing them on the first fuser forward pass, which invalidated the fuser's cached params and ops.Linear's parameter aliases. Backward then failed with a meta-vs-cuda device mismatch.

This PR makes basic ops version their parameter registrations, and the fuser treats version change as a cache invalidation. `te.ops.FusedOperation` and `te.Sequential` also now have a `reset_parameters()` method to explicitly materialize parameters the same way it's done for standard `torch.nn.Modules` and TE custom modules.

New tests added for deferred initialization with the affected ops.

Fixes #3322

## Type of change

- [ ] Infra/Build change
- [ ] Code refactoring

# Checklist:

### DIFF
--- a/tests/pytorch/test_fusible_ops.py
+++ b/tests/pytorch/test_fusible_ops.py
@@ -649,6 +649,254 @@ class TestFuser:
             assert x.grad.dtype == model_dtype
             assert op.weight.grad.dtype == model_dtype
 
+    _deferred_init_op_types = (
+        "basic_linear",
+        "bias",
+        "layer_norm",
+        "rmsnorm",
+        "grouped_linear",
+        "grouped_linear_single_param",
+        "linear",
+        "sequential",
+    )
+
+    @staticmethod
+    def _make_deferred_init_op(
+        op_type: str,
+        *,
+        size: int,
+        num_groups: int,
+        dtype: torch.dtype,
+        device: torch.device | str,
+    ) -> torch.nn.Module:
+        """Construct an op for the deferred initialization tests"""
+        kwargs = {"device": device, "dtype": dtype}
+        if op_type == "basic_linear":
+            return te_ops.BasicLinear(size, size, **kwargs)
+        if op_type == "bias":
+            return te_ops.Bias(size, **kwargs)
+        if op_type == "layer_norm":
+            return te_ops.LayerNorm(size, **kwargs)
+        if op_type == "rmsnorm":
+            return te_ops.RMSNorm(size, **kwargs)
+        if op_type == "grouped_linear":
+            return te_ops.GroupedLinear(num_groups, size, size, bias=True, **kwargs)
+        if op_type == "grouped_linear_single_param":
+            return te_ops.GroupedLinear(
+                num_groups,
+                size,
+                size,
+                bias=True,
+                single_grouped_weight=True,
+                single_grouped_bias=True,
+                **kwargs,
+            )
+        if op_type == "linear":
+            return te_ops.Linear(size, size, bias=True, **kwargs)
+        if op_type == "sequential":
+            return te_ops.Sequential(
+                te_ops.LayerNorm(size, **kwargs),
+                te_ops.Linear(size, size, bias=True, **kwargs),
+            )
+        raise ValueError(f"Unsupported op type ({op_type})")
+
+    @staticmethod
+    def _check_materialized_params(
+        op: torch.nn.Module,
+        device: torch.device | str,
+        *,
+        expect_grads: bool = True,
+    ) -> dict[str, torch.nn.Parameter]:
+        """Check that an op's params are on the device, and return them"""
+        params = dict(op.named_parameters())
+        assert params
+        for name, param in params.items():
+            assert param.device.type == device, f"{name} was not materialized on {device}"
+            if expect_grads:
+                assert param.grad is not None, f"{name} did not get a grad"
+                assert param.grad.device.type == device, f"{name} got a grad on {param.grad.device}"
+
+        # Fused ops must not alias stale params
+        for module in op.modules():
+            if isinstance(module, te_ops.Linear):
+                assert module.weight is module.basic_ops[module._linear_idx].weight
+                assert module.bias is module.basic_ops[module._bias_idx].bias
+        return params
+
+    @pytest.mark.parametrize("op_type", _deferred_init_op_types)
+    @pytest.mark.parametrize("dtype", _dtypes)
+    @pytest.mark.parametrize("input_requires_grad", (False, True))
+    def test_deferred_param_init(
+        self,
+        monkeypatch,
+        *,
+        op_type: str,
+        size: int = 32,
+        num_groups: int = 2,
+        dtype: torch.dtype,
+        input_requires_grad: bool,
+        device: torch.device = "cuda",
+    ) -> None:
+        """Test ops constructed on the meta device
+
+        Ops replace their params when materializing them on the first
+        forward pass. See
+        https://github.com/NVIDIA/TransformerEngine/issues/3322.
+
+        """
+        if op_type == "grouped_linear_single_param":
+            # Materializing params also changes how many there are
+            monkeypatch.setenv("NVTE_GROUPED_LINEAR_SINGLE_PARAM", "1")
+
+        # Construct operation on meta device
+        op_kwargs = {"size": size, "num_groups": num_groups, "dtype": dtype}
+        op = self._make_deferred_init_op(op_type, device="meta", **op_kwargs)
+        for param in op.parameters():
+            assert param.device.type == "meta"
+
+        # Forward and backward pass
+        in_shape = (size, size)
+        extra_inputs = []
+        if op_type.startswith("grouped_linear"):
+            in_shape = (size * num_groups, size)
+            extra_inputs.append(torch.tensor([size] * num_groups, dtype=torch.int, device=device))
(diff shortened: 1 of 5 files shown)


tests/pytorch/test_fusible_ops.py
```diff
@@ -649,6 +649,140 @@ def test_pyt_autocast(
...
+        # Skip invalid configurations
+        in_shape = (size, size)
+        if quantization is None:
+            pytest.skip("Quantization scheme is not specified")
+        maybe_skip_quantization(quantization, dims=in_shape, device=device, dtype=dtype)
+
+        # Construct operation on meta device
+        recipe = make_recipe(quantization)
+        with te.quantized_model_init(enabled=quantized_weight, recipe=recipe):
+            op = te_ops.Linear(size, size, bias=True, device="meta", dtype=dtype)
```
vthumbe1503: Should we extend this test for GroupedLinear as well?

transformer_engine/pytorch/ops/fuser.py
```diff
@@ -423,6 +438,20 @@ def maybe_fuse_ops(
...
+        # quantized. Ops are initialized top-down so fused ops can sync
+        # params aliased from their basic ops.
+        is_first_forward = self.recipe_type is None
+        if is_first_forward:
+            for op in self._basic_ops:
+                op.reset_recipe_state(recipe=recipe)
+            for op in self._ops:
+                op.pre_first_fuser_forward()
+
+            # Cache params now that ops have initialized them
```
vthumbe1503: If  somebody calls register_parameter on the ops in a later iteration, this means it would not get cached in fuser's parameter list

I know it is a weird use-case, and maybe nobody does this. But was this always intended, @timmoon10 ?

vthumbe1503 (commented):
LGTM, have some question and a minor suggestion for adding grouped_linear test case as well for quantized_model_int case

tests/pytorch/test_fusible_ops.py
```diff
@@ -649,6 +649,140 @@ def test_pyt_autocast(
...
+        for param in op.parameters():
+            assert param.device.type == "meta"
+
+        # Forward and backward pass
+        in_shape = (size, size)
+        extra_inputs = []
+        if op_type.startswith("grouped_linear"):
+            in_shape = (size * num_groups, size)
+            extra_inputs.append(torch.tensor([size] * num_groups, dtype=torch.int, device=device))
+        x = torch.randn(in_shape, dtype=dtype, device=device, requires_grad=True)
```
ptrendx: We should test both requires_grad = True and False

tests/pytorch/test_fusible_ops.py
```diff
@@ -649,6 +649,140 @@ def test_pyt_autocast(
...
+        # Check that params have been materialized
+        assert op.weight is op.basic_ops[op._linear_idx].weight
+        assert op.bias is op.basic_ops[op._bias_idx].bias
+        if quantized_weight:
+            assert isinstance(op.weight, QuantizedTensor)
+        for name, param in op.named_parameters():
+            assert param.device.type == device, f"{name} was not materialized on {device}"
+            assert param.grad is not None, f"{name} did not get a grad"
+            assert param.grad.device.type == device, f"{name} got a grad on {param.grad.device}"
+        assert x.grad is not None and x.grad.device.type == device
```
ptrendx, line 873: In addition to that there should be some testing that the resulting output/gradients actually match the op that did not go through this materialization step.

timmoon10 (changes requested):
Are we sure there is actually a bug? I see this PR adds logic to defer some `OperationFuser` setup to the first forward pass. However, `OperationFuser` is an internal class that is only instantiated by `Sequential` in the first forward pass:
https://github.com/NVIDIA/TransformerEngine/blob/32f7096a355ca803e7151da0d0f4aa65ac4ba46b/transformer_engine/pytorch/ops/sequential.py#L184

It's perfectly possible that different fusions are available with quantized/unquantized params, so trying to "fix" the `OperationFuser` is dangerous. If a change is detected, better to destroy and create a new one.

tests/pytorch/test_fusible_ops.py
```diff
@@ -649,6 +649,140 @@ def test_pyt_autocast(
...
+                te_ops.Linear(size, size, bias=True, **kwargs),
+            )
+        else:
+            raise ValueError(f"Unsupported op type ({op_type})")
+        for param in op.parameters():
+            assert param.device.type == "meta"
+
+        # Forward and backward pass
+        in_shape = (size, size)
+        extra_inputs = []
```
vthumbe1503, line 759: Isnt the normal flow for deferred initialization to call reset_parameters before doing the first forward pass?

transformer_engine/pytorch/ops/fuser.py
denera: This gap actually gets resolved neatly with the new parameter versioning. Registering a new param bumps the version, which automatically invalidates the pre-existing cache and forces the fuser to re-cache and re-fuse.

tests/pytorch/test_fusible_ops.py
```diff
@@ -649,6 +649,254 @@ def test_pyt_autocast(
...
+        # Check that params have been materialized
+        params = self._check_materialized_params(op, device)
+        if quantized_weight:
+            weights = [param for name, param in params.items() if "weight" in name]
+            assert weights
+            for weight in weights:
+                assert isinstance(weight, QuantizedTensor)
+        assert x.grad is not None and x.grad.device.type == device
+
+    def test_param_replaced_after_first_forward(
```
vthumbe1503, line 875: I dont think this is a use-case @denera 

The way this test     `def test_deferred_param_init_explicit(`
calls reset_parameters before calling the first forward is the realistic use-case. Can we make all the tests follow this use-case. 

And I think the fix in fuser is not needed as @timmoon10 pointed out. Since the cache_parameters that is going to be called will be after the parameters are materialized using reset_parameters() and so it will be right.


I think we should modify this PR to have an example in pytorch's example folder, showing how deferred initialization works and some test cases in this file for that.

Issue #3331, opened 2026-08-08
THD + CP (p2p): pad_between_seqs auto-detect is blind to tail padding → nondeterministic forward, silently wrong attention (dropped real keys, exact-zero real rows), nondeterministic gradients


JackRao123:
## Summary

With `qkv_format="thd"`, `attn_mask_type="padding_causal"`, context parallelism (`cp_comm_type="p2p"`), and a sequence whose **real** length is not divisible by `2*cp_size` (tail-padded — e.g. real 698 padded to 704 at CP4, the common case for packed THD training), the `pad_between_seqs` auto-detect in `DotProductAttention.forward` compares `cu_seqlens_padded[:-1]` with `cu_seqlens[:-1]` — deliberately ignoring tail padding after the last sequence ([`dot_product_attention.py` @ v2.16, L1519–1529](https://github.com/NVIDIA/TransformerEngine/blob/v2.16/transformer_engine/pytorch/attention/dot_product_attention/dot_product_attention.py#L1519-L1529)).

The call therefore takes the `pad_between_seqs=False` fast path in the CP p2p orchestration, which approximates per-ring-step seqlens as `cu_seqlens // cp_size` ([`context_parallel.py` @ v2.16, L762–773](https://github.com/NVIDIA/TransformerEngine/blob/v2.16/transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py#L762-L773), `cp_p2p_fwd_prepare_qkv`). That uniform-average approximation mislabels each rank's chunk-boundary rows whenever `real % (2*cp) != 0`. In short: the detect conflates "no padding **between** sequences" with "no padding at all" — an equivalence that only holds when CP is off, because CP wraps tail padding into a rank's local buffer and divides boundaries by `cp_size`.

**Affects:** every released TE we checked (2.16.0, 2.17, 2.17.1). `main` is only *incidentally and partially* unaffected — see below.

[...]

PR #3332, opened 2026-08-08
[PyTorch] Regression test + docs for THD CP tail-padding mislabeling (pad_between_seqs)

JackRao123:
# Description

Regression test + documentation for the THD + context-parallel tail-padding defect reported in #3331.

In short: with `qkv_format="thd"` + CP `p2p`, a sequence whose real length is not divisible by `2*cp_size` (tail-padded — the common case for packed THD training) is mishandled whenever the `pad_between_seqs=False` fast path is taken, because that path approximates per-ring-step seqlens as `cu_seqlens // cp_size` and thereby mislabels chunk-boundary rows. Consequences (proven at element level in #3331): nondeterministic forward rows, silently wrong attention (real keys dropped, real query rows exact-zero), nondeterministic gradients on all CP ranks.

`main` routes callers that pass *distinct* padded/unpadded cu_seqlens tensors to the exact path (incidental effect of the #2898 auto-detect rewrite), but nothing pins that behavior, and the docstring/comment still describe tail padding as a safe `False` signal — which is incorrect under CP.

Refs #3331 (this PR is the regression test + docs; the deeper routing question — e.g. whether THD+CP should always take the exact per-rank path, and LSE aux hardening as defense in depth — is left for discussion on the issue, per the asks there).

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue) — regression coverage for a released bug

## Changes

- **`tests/pytorch/attention/run_cp_thd_tail_padding.py`** (new): distributed runner. Single packed sequence, real length 698 (not divisible by `2*cp` for cp∈{2,4}), tail-padded; FusedAttention backend, `padding_causal`, THD, CP `p2p`. Three arms per pad mode (`pad_between_seqs=True` explicit, and auto-detect with distinct padded tensors): (1) forward bitwise determinism over 25 iterations, (2) fwd+bwd bitwise determinism over 10 iterations, (3) correctness of out/dq/dk/dv vs an unpadded no-CP reference on all real rows.
- **`tests/pytorch/attention/test_cp_thd_tail_padding.py`** (new): pytest wrapper launching the runner at world sizes 2 and 4 via `run_distributed`.
- **`dot_product_attention.py`**: docstring for `pad_between_seqs` and the auto-detect comment now state that tail padding is *not* benign under CP, and that tail-padded THD+CP callers must pass `pad_between_seqs=True` explicitly or supply distinct padded cu_seqlens tensors.

# Validation (H100×4, torch 2.11.0+cu128, cuDNN 9.25.0.15)

| run | TE 2.17.1 (affected release) | TE main (this branch) |
|---|---|---|
| CP4 | **FAIL** — auto arm: forward nondeterministic 24/24 iters; fwd+bwd nondeterministic; out/dq/dk/dv vs reference max \|Δ\| up to 1.41 | **PASS** (both modes) |
| CP2 | **FAIL** — auto arm: deterministic mis-attention vs reference, max \|Δ\| up to 0.95 | **PASS** (both modes) |

The explicit `pad_between_seqs=True` arm passes on 2.17.1 as well, confirming the documented workaround.

One environment note: TE 2.17.1's THD backward fails wholesale with cuDNN 9.19.0 (the version torch 2.11.0+cu128 pins) — `CUDNN_STATUS_BAD_PARAM` at `fused_attn_f16_arbitrary_seqlen.cu:934` (reshape attribute), independent of this defect and reproducible on a plain no-CP THD backward; cuDNN 9.25.0.15 resolves it. May be worth a minimum-cuDNN note for the THD backward path.

# Checklist:

- [x] New and existing unit tests pass locally with my changes — new test validated on H100×4 (see table); it exercises a path no existing test covers (existing THD CP tests pad every sequence, so internal boundaries always move and the auto-detect never sees the tail-only case)

### DIFF
--- a/tests/pytorch/attention/run_cp_thd_tail_padding.py
+++ b/tests/pytorch/attention/run_cp_thd_tail_padding.py
@@ -1,0 +1,226 @@
+# Copyright (c) 2022-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+#
+# See LICENSE for license information.
+
+"""Regression test for THD + context-parallel (p2p) tail padding.
+
+GitHub issue: https://github.com/NVIDIA/TransformerEngine/issues/3331
+
+A single packed sequence whose REAL length is not divisible by 2*cp_size is
+tail-padded to a multiple of 2*cp_size (the common case for packed THD
+training). The pad_between_seqs=False CP path approximates per-ring-step
+seqlens as cu_seqlens // cp_size, which mislabels chunk-boundary rows in
+this case, producing (a) nondeterministic forward outputs (uninitialized
+softmax-LSE aux rows merged into a real row's scale factor), (b) silently
+wrong attention (real keys dropped from every query's attention, real query
+rows receiving exact-zero output), and (c) nondeterministic gradients on all
+CP ranks.
+
+This test pins the exact-path semantics — reached via explicit
+pad_between_seqs=True, or via auto-detect with padded cu_seqlens supplied as
+tensors distinct from their unpadded counterparts — along two axes:
+
+  1. determinism: forward and fwd+bwd outputs are bitwise-stable across
+     iterations on identical inputs;
+  2. correctness: the reassembled CP output and gradients match the no-CP
+     reference on all real (non-padding) rows within bf16 tolerance.
+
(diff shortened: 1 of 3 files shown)


JackRao123:
Closing for now — we'll carry the workaround on our side (explicit pad_between_seqs=True). The analysis in #3331 stands; happy to reopen if there's maintainer interest in the regression test.

PR #3337, opened 2026-08-10
[Common] Improved performance of Group MXFP8 kernels

Oleg-Goncharov:
# Description

This PR optimizes grouped MXFP8 quantization by introducing layout-specific kernel configurations and work scheduling strategies.

Groups with uniform shapes or a common last dimension use direct CTA-to-chunk mapping, while layouts requiring per-tensor TMA descriptors use per-tensor persistent scheduling. The kernel processes larger chunks through multiple stages, reducing scheduling and tensor-mapping overhead for large grouped workloads.

The change also ensures that dynamically updated TMA descriptors are acquired before use, handles graph-safe logical capacity without out-of-bounds accesses, and preserves correct dBias workspace behavior across multi-chunk execution.

## Type of change

- [ ] Infra/Build change
- [x] Code refactoring

## Changes

- Extend grouped TMA storage with per-tensor row, column, and offset metadata.
- Introduce layout-specific grouped MXFP8 kernel configurations and compile-time traits.
- Use direct CTA-to-chunk mapping for `SAME_BOTH_DIMS` and `VARYING_FIRST_DIM` layouts.
- Use per-tensor persistent scheduling for `VARYING_LAST_DIM` and `VARYING_BOTH_DIMS` layouts.
- Process `128x128` and `128x256` chunks through multiple X/Y stages.
- Acquire dynamically updated TMA descriptors before they are consumed by the quantization kernel.
- Fixed shared-memory reuse races and incomplete completion handling (details in [PR#3417 [Common] Fix TMA synchronization in quantization kernels](https://github.com/NVIDIA/TransformerEngine/pull/3417)).

### Runtime Speedup Factors on GB300

image

# Checklist:


transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -31,53 +31,169 @@ namespace group_quantize_kernel {
...
+  if constexpr (shape_rep == ShapeRepresentation::SAME_BOTH_DIMS) {
+    NVTE_CHECK(first_logical_dim % num_tensors == 0,
+               "SAME_BOTH_DIMS requires an equal integral row count per tensor.");
+    config.same_both_rows = first_logical_dim / num_tensors;
+    config.work_blocks_X = DIVUP(last_logical_dim, CHUNK_DIM_X);
+    config.work_blocks_Y = DIVUP(config.same_both_rows, CHUNK_DIM_Y);
+    NVTE_CHECK(config.work_blocks_X > 0 && config.work_blocks_Y > 0,
+               "SAME_BOTH_DIMS requires non-empty tensors.");
+    NVTE_CHECK(config.work_blocks_Y <= 65535 && num_tensors <= 65535,
+               "SAME_BOTH_DIMS direct mapper exceeds CUDA grid Y/Z limits.");
```
ptrendx: Those errors are not really actionable by the user - what do those values correspond to?

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -31,53 +31,169 @@ namespace group_quantize_kernel {
...
+    NVTE_CHECK(config.work_blocks_Y <= 65535,
+               "VARYING_FIRST_DIM direct mapper exceeds the CUDA grid Y limit.");
+
+    // Tensor boundaries are TILE_DIM_Y-aligned, so each CTA directly owns one configured chunk
+    // in the vertically stacked tensor.
+    config.grid = dim3(config.work_blocks_X, config.work_blocks_Y);
+  } else {
+    NVTE_CHECK(num_tensors <= MAX_SUPPORTED_TENSOR_DESCRIPTORS,
+               "Number of tensors in a group is larger than the MAX number of supported "
+               "descriptors (64).");
```
ptrendx: Not really needed in this PR, but at some point we will need to lift this limitation.

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -31,53 +31,162 @@ namespace group_quantize_kernel {
...
+    config.work_blocks_X = DIVUP(last_logical_dim, CHUNK_DIM_X);
+    config.work_blocks_Y = DIVUP(config.same_both_rows, CHUNK_DIM_Y);
+    NVTE_CHECK(config.work_blocks_X > 0 && config.work_blocks_Y > 0,
+               "SAME_BOTH_DIMS requires non-empty tensors.");
+
+    // Direct tensor-local mapper: one CTA owns one chunk.
+    config.grid = dim3(config.work_blocks_X, config.work_blocks_Y, num_tensors);
+  } else if constexpr (shape_rep == ShapeRepresentation::VARYING_FIRST_DIM) {
+    config.work_blocks_X = DIVUP(last_logical_dim, CHUNK_DIM_X);
+    config.work_blocks_Y = DIVUP(first_logical_dim, CHUNK_DIM_Y);
```
vthumbe1503, line 159: Do we have to get rid of persistent kernel for varying first dims?

If sum(first_dims) is considerably lesser than first_logical_dim we will overlaunch a lot of thread blocks. Although, it is an early return for overlaunched blocks, but do we have some perf numbers for those cases as well?

Oleg-Goncharov: Yes, these cases are already covered by the benchmarks in the table. For `VARYING_FIRST_DIM`, I specifically tested cases where the tensor allocation is much larger than the amount of data actually processed, e.g. `4096x4096`, `8192x8192`, or `16384x16384` allocations with only `1024` or `2048` active rows.

So this does include scenarios with a significant overlaunch due to `sum(first_dims) << first_logical_dim`. Despite the early-return blocks, the new implementation still shows very good performance compared to the previous version for `VARYING_FIRST_DIM`.

`VARYING_BOTH_DIMS` does behave worse in this regard. However, since this scenario is not currently used in production, I think we can leave that case as-is for now.

ptrendx:
@Oleg-Goncharov I see that in the commit addressing the PR review you removed the NVTE_CHECK calls that guarded against y/z block coordinates to be larger than 65k, but there are no other changes. Could you comment why just removing this check was enough?

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -31,53 +31,162 @@ namespace group_quantize_kernel {
...
 
-struct TunableConfig {
+struct DefaultCastConfig {
+  static constexpr uint TILE_DIM_Y = 128;
+  static constexpr uint TILE_DIM_X = 128;
   static constexpr uint CHUNK_DIM_Y = 128;
   static constexpr uint CHUNK_DIM_X = 128;
   static constexpr uint THREADS_PER_CHUNK = 128;
-  // Launch static persistent grid as (SM_count * STATIC_PERSISTENT_BLOCKS_PER_SM, 1, 1).
+  static constexpr uint PREFETCH_STAGES = 1;
```
tdophung: Please keep the comment, I think it is helpful

Oleg-Goncharov:
> @Oleg-Goncharov I see that in the commit addressing the PR review you removed the NVTE_CHECK calls that guarded against y/z block coordinates to be larger than 65k, but there are no other changes. Could you comment why just removing this check was enough?

@ptrendx You are right, simply removing the checks was not sufficient, since the direct mapper still relied on the limited Y/Z grid dimensions. I have now linearized the virtual (column block, row block, tensor) grid into grid.x for `SAME_BOTH_DIMS`, and (column block, row block) for `VARYING_FIRST_DIM`. The kernel decodes the original coordinates from `blockIdx.x`, preserving the CTA count and the original X to Y to tensor order. This removes the dependency on the 65k Y/Z limits while keeping the potentially large work grid in CUDA’s larger X dimension.

transformer_engine/common/cast/mxfp8/group_quantize_mxfp8.cuh
```diff
@@ -856,35 +1133,17 @@ void group_quantize(const GroupedTensor *input, const GroupedTensor *activations
...
-  size_t work_blocks_X = 0;
-  size_t work_blocks_Y = 0;
-
-  if (is_single_tensor) {
-    work_blocks_Y = DIVUP(first_logical_dim, static_cast<size_t>(CHUNK_DIM_Y));
-    work_blocks_X = DIVUP(last_logical_dim, static_cast<size_t>(CHUNK_DIM_X));
-  } else {
-    NVTE_CHECK(num_tensors <= MAX_SUPPORTED_TENSOR_DESCRIPTORS,
-               "Number of tensors in a group is larger than "
-               "the MAX number of supported descriptors (64).");
```
tdophung, line 869: Is removing this check intentional?

Oleg-Goncharov: Yes, for this PR, we decided not to check whether we hit this limit. We'll increase the maximum number of supported tensor descriptors at some point. Please see Przemek's comment above.

ptrendx: I think there was some miscommunication then - I assume you are talking about my comment "Not really needed in this PR, but at some point we will need to lift this limitation.". I did not mean that the error message is not needed here, it should still be there. Instead, what I meant is that we should lift this limitation but that lifting is not really needed to be done in this PR. Sorry for the confusion!

Oleg-Goncharov: Ah, got it - thank you for clarifying. I understood the original comment as meaning that the check itself was not needed in this PR. I’ll restore it.

ptrendx (approved):

